PR #3869, opened 2018-09-10
bug: prevent removing full project by mistake due to build or generate paths

pimlie:
## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Description
When you set `options.buildDir` or `options.generate.dir` to a path equal or lower then `options.rootDir`, you will completely remove your project when you build or generate your project as these paths are indiscriminately removed with `fs-extra.remove` in `builder/builder.js` and `builder/generator.js` . I have added a test in options.js which will throw an error when this would be the case.

I feel I need to clarify for everybody here that I did **not** make that mistake while playing around and that because I did **not** make that mistake I did **not** lose my whole project and will **not** have to revert to a day old backup :blush: :cry: 

## Checklist:
- [x] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests passed.

PR #3893, opened 2018-09-15
feat: add tests to check for changed files

pimlie:
As discussed in #3869

Make sure that if we are building or generating only files in buildDir and generate.dir are changed. If files in another location would also be changed due to a new config option, those locations should be guarded in `lib/common/options.js` so you cant set them lower then rootDir or srcDir.

## Checklist:

### DIFF
--- a/test/unit/basic.generate.test.js
+++ b/test/unit/basic.generate.test.js
@@ -62,7 +62,7 @@ describe('basic generate', () => {
     let changedFileFound = false
     const paths = listPaths(generator.nuxt.options.rootDir, pathsBefore)
     paths.map((item) => {
-      console.log(changedFileName, item.path)
+      console.log(changedFileName, item.path) // eslint-disable-line no-console
       if (item.path === changedFileName) {
         changedFileFound = true
       } else {
--- a/test/unit/basic.generate.test.js
+++ b/test/unit/basic.generate.test.js
@@ -62,6 +62,7 @@ describe('basic generate', () => {
     let changedFileFound = false
     const paths = listPaths(generator.nuxt.options.rootDir, pathsBefore)
     paths.map((item) => {
+      console.log(changedFileName, item.path)
       if (item.path === changedFileName) {
         changedFileFound = true
       } else {
--- a/test/unit/basic.generate.test.js
+++ b/test/unit/basic.generate.test.js
@@ -1,10 +1,10 @@
-import { existsSync } from 'fs'
+import { existsSync, writeFile } from 'fs'
 import http from 'http'
-import { resolve } from 'path'
+import { resolve, sep } from 'path'
 import { remove } from 'fs-extra'
 import serveStatic from 'serve-static'
 import finalhandler from 'finalhandler'
-import { Builder, Generator, getPort, loadFixture, Nuxt, rp } from '../utils'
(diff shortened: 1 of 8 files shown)


pimlie:
Yay, all checks passed. I did a forced push with the double 'add debug logs' commits to try to remove / hide some commits as I was remotely debugging the CI failures. Either my pc is much slower or much faster then the CI's as locally the writeFile just finished long before running the tests ;)

test/utils/build.js
```diff
@@ -15,4 +24,17 @@ export const buildFixture = function (fixture, callback, hooks = []) {
...
   }, 120000)
+
+  test('Check changed files', () => {
+    expect.hasAssertions()
+
+    // When building Nuxt we only expect files to changed
+    // within the nuxt.options.buildDir
+    Object.keys(pathsBefore).forEach((key) => {
+      const paths = listPaths(nuxt.options[`${key}Dir`], pathsBefore[key])
+      paths.map((item) => {
```
manniL: `forEach` instead of `map` if you don't need a return value.

test/unit/basic.generate.test.js
```diff
@@ -43,6 +53,24 @@ describe('basic generate', () => {
...
 
+  test('Check changed files', () => {
+    // When generating Nuxt we only expect files to change
+    // within nuxt.options.generate.dir, but also allow other
+    // .nuxt dirs for when tests are runInBand
+    const allowChangesDir = resolve(generator.nuxt.options.generate.dir, '..', '.nuxt')
+
+    let changedFileFound = false
+    const paths = listPaths(generator.nuxt.options.rootDir, pathsBefore)
+    paths.map((item) => {
```
manniL: `forEach` instead of `map` if you don't need a return value.

TheAlexLichter (changes requested):

pimlie:
It seems appveyor really screws up the timeout time. I was confident that using a high resolution time would be a quick fix for the waitfor test but as many times before I was wrong. Rounding the time to a whole number fixes the test in this case, but we will have to wait and see if timeouts on appveyor are off by more than 0.5ms. If so we can skip the test again, or maybe file an report with appveyor? Spent way too much on this already so actually I wouldnt mind to understand why this is at all possible ;)

test/unit/utils.test.js
```diff
@@ -16,10 +16,15 @@ describe('utils', () => {
...
-    expect(Date.now() - s >= 100).toBe(true)
+  test('waitFor', async () => {
+    const delay = 100
+    const s = process.hrtime()
+    await Utils.waitFor(delay)
+    const t = process.hrtime(s)
+    // Node.js makes no guarantees about the exact timing of when callbacks will fire
+    // HTML5 specifies a minimum delay of 4ms for timeouts
+    // although arbitrary, use this value to determine our lower limit
+    expect((t[0] * 1e9 + t[1]) / 1e6).not.toBeLessThan(delay - 4)
```
pi0: Crazy boy 😂😋

pi0 (approved):
LGTM

TheAlexLichter (approved):

test/unit/utils.test.js
clarkdo: Good idea

clarkdo (approved):
LGTM

PR #3901, opened 2018-09-16
update PostCSS to version 7.0.2

nathanchase:


aldarund:
dupe of https://github.com/nuxt/nuxt.js/pull/3901

PR #3906, opened 2018-09-17
load mjs files as javascript/auto

### LINKED ISSUE #3892: [edge] Uncaught ReferenceError: require is not defined when using @nuxtjs/apollo
### Version

nuxt-edge

### Reproduction link

[https://github.com/Vlaoff/apollonuxt](https://github.com/Vlaoff/apollonuxt)

### Steps to reproduce

I get the error when I launch nuxt and open it in the browser

### What is expected ?

I should not have this error

I think it's a problem with the babel/wepack configuration.

### Additional comments?

It does not occur on nuxt 1.4.2, only on the nuxt-edge 2.0.0-xxx (could not select the correct version on cmty.app)

I don't know if the problem is from @nuxtjs/apollo  or nuxt. But as it does not occur on the nuxt stable branch, I submit it here 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7745">#c7745</a>)</em></sub></div>

aldarund:
## Types of changes

## Description
Resolves #3892 

## Checklist:

### DIFF
--- a/lib/builder/webpack/base.js
+++ b/lib/builder/webpack/base.js
@@ -173,6 +173,10 @@ export default class WebpackBaseConfig {
         })
       },
       {
+        test: /\.mjs$/,
+        type: 'javascript/auto'
+      },
+      {
         test: /\.css$/,
         oneOf: perfLoader.poolOneOf('css', styleLoader.apply('css'))
       },


pi0:
I think mjs spec strictly forbids use of require. Maybe we can offer users to extend webpack config instead of changing it to something non standard. Or Simply they can use js if require is needed.

pi0:
Another note: We don't know what optimization and behavior would webpack have for mjs type in the future. If we behave .mjs as js/auto (Same as js) then what's the point of setting mja extension instead of js?

aldarund:
@pi0 there no require in the mjs file itself. But it endsup in generated code.
```
 function(e, t, n) {
    "use strict";
    (function(e) {
        t.a = function(e, t) {
            return e instanceof t
        }
    }
    ).call(this, require("./../../process/browser.js"))
}
```

Fix was stolen from here : https://github.com/aws-amplify/amplify-js/issues/686#issuecomment-387710340

clarkdo (commented):
@pi0 is correct, webpack now is using `javascript/esm` for `.mjs` file, it's more strictly, dynamic modules (non-esm, i. e. CommonJs) can only imported via default import, everything else (including namespace import) emit error

If we use `javascript/auto`, `.mjs` functionality will be limited, because it will be treated as a `js` and lost the `.mjs` features (Although it's still WIP) 😄

aldarund:
@clarkdo  @pi0 digged a bit depper,  ye, you are right. I`m not sure why and how issue with nuxt/apollo happens because i dont see any require inside graphql mjs files. But updating graphql to 14 fixes the problem. I`l open a pr to vue-cli-appollo-plugin with version bump then :)

PR #3726, opened 2018-08-15
Minor consistency tweaks

galvez:
@Atinux @clarkdo just a few places `function` seemed to be uselessly used (and in files where `() =>` was _also_ used)... changed and ran the tests, ditto! No biggie :)

Just my daily [pomodoro](https://en.wikipedia.org/wiki/Pomodoro_Technique) studying Nuxt.

Also noticed a few places where `[].concat` and `Array.prototype.concat` were used for the same purpose. Just thought it was cleaner to standardize on referencing the actual prototype method.

PR #3916, opened 2018-09-19
feat: add lint check for app (on fixture builds)

pimlie:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Description
In #3726 @clarkdo suggested to not refactor the syntax in the app folder until we have some eslint in place. As I think it would be quite difficult to fully lint all the tags in those templates, my suggestion for this would be to run the lint on the builded test fixtures and loosen the lint rules where necessary.

If you agree this is a good (or good enough) solution I will fix the linting issues as well.

Just for fun, running `yarn lint:app` currently results in:
```js
✖ 36176 problems (35740 errors, 436 warnings)
  33092 errors and 436 warnings potentially fixable with the `--fix` option.
```
## Checklist:
- [ ] All new and existing tests passed.


aldarund:
What about https://www.npmjs.com/package/eslint-plugin-lodash-template ? Maybe it could work fine

pimlie:
I couldnt get it to work because its seems to be aimed at linting a html template instead of a vue component.

clarkdo:
@pimlie It's a good idea and you're right `eslint-plugin-lodash-template` only works for HTML syntax unfortunately.

@Atinux @pi0 Are we going to refactor the `templates` ? If not, I think we can merge this one after the release 2.

pimlie:
Looking at the errors I actually already found a possible bug. Didnt look into it further yet but for the `with-config` fixture it seems the `loading.duration` property is empty and therefor the generated nuxt-loading component has a parse error: `duration: ,`

manniL:
@clarkdo @Atinux If we don't refactor/change templates, this PR is great. However on long term we might consider writing an own linting plugin for js lodash templates (in case we keep this way of templates for long)

pimlie:
@manniL In case we keep using this way of templates, couldnt this also be more of an `and ` situation instead of an `or`? I agree that having a `eslint-plugin-lodash-template-vue` plugin would be nice as template readibility is probably more important then correctly linted output of those templates. Therefore we probably will need to loosen the lint rules a bit. But looking at the nuxt-loading issue above, having a lint check on the generated templates as well seems to provide at least some safe guards against templating errors.

manniL:
Of course both solutions will provide the best coverage! If we'd write an own lint plugin, it won't replace this solution 👌

Atinux:
We can add this, and when templates refactored (to reconsider since it's not a big priority) we could change it.

pimlie:
This PR is ready for review again. I have fixed only the lint errors I was quite sure about (and have ignored nuxt-loading because of my other PR). Most errors that still exist are `no-unused-var` errors, but I am not 100% sure this is because they are really not used or there is a test case missing which would use that var.
Taking this into account, the same could apply to const / let uses. I have checked them to my best ability, but please review these changes carefully.

For better maintainability I have added an `isTest` prop to `nuxt.options`. With this prop we can selectively add `eslint-disable` comments in the templates where needed (eg all serialize, devalue and JSON.stringify statements).

I have added a special eslint formatter which merges all the messages for a template and adds an occurence count of the file and the message. It can also display the source code where the error occured (run `ESLINT_SHOWSRC=1 yarn test:app`).

This is an example of one of the errors that still exist:
image

pimlie:
I will close this issue as its not likely this will ever be merged due to the global name PR. Or let me rephrase that, I will not try to merge the changes of that PR into this one as I like to stay sane and re-linting will most likely be much quicker. I guess there was some miscommunication about this between us.

In the mean time I have created this [`eslint-multiplexer`](https://github.com/pimlie/eslint-multiplexer) package which provides a generic solution for linting generated app templates. When the time comes we could use that to help with re-linting the app templates again.

galvez:
@pimlie excellent work -- we'll make sure to revisit this.

PR #3927, opened 2018-09-20
refactor: config Postcss preset and plugins together

### LINKED ISSUE #3944: Merge PostCSS config

If I pass my own config to `build.postcss`, it will replace the default plugins along with their settings.
If I keep the default plugins by setting them to an empty object, like the [docs](https://nuxtjs.org/api/configuration-build#postcss) show, I keep the plugins but not their settings.
In fact, the docs mention defaults but the code example isn't what's found in the source. If I want to **add** to Nuxt's default PostCSS setup, I shouldn't be expected to know that a certain file in the source contains the defaults I'm gonna have to copy and paste into `nuxt.config.js`. Unless I'm missing something very obvious, there's no elegant way of adding Tailwind for example while keeping [these](https://github.com/nuxt/nuxt.js/blob/e9bb9e7236eb0c9643daee8e557e6e254a08db8e/lib/builder/webpack/utils/postcss.js#L20-L50) defaults.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7806">#c7806</a>)</em></sub></div>

clarkdo:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

## Description

TODO:

- [ ] Consider how to config `postcss.plugins` and attach defaults

Fix #3944

## Checklist:

- [ ] All new and existing tests passed.


lib/builder/webpack/utils/postcss.js
```diff
@@ -115,7 +115,7 @@ export default class PostcssConfig {
         this.preset = config.preset
         delete config.preset
       }
-      _.defaults(config, this.defaultConfig)
+      _.defaultsDeep(config, this.defaultConfig)
```
pi0: This is not just for examples @clarkdo right?

clarkdo: Yeah, this is only impacting change. How do you think ?

pi0: If required we should make it as a Hotfix :)

clarkdo: With this we can support : 

```js
    postcss: {
      plugins: [ ... ],
      preset: { ... }
    }
```

clarkdo: Sure. let's consider it after 2

TheAlexLichter (approved):
That's a great way to merge defaults with possibly added plugin :clap: 
Thanks for the effort!

lib/builder/webpack/utils/postcss.js
```diff
@@ -115,11 +121,14 @@ export default class PostcssConfig {
         this.preset = config.preset
         delete config.preset
       }
-      _.defaults(config, this.defaultConfig)
-
-      this.loadPlugins(config)
+      if (Array.isArray(config.plugins)) {
```
aldarund: When plugins can/should be array?

clarkdo: If plugins is array,  will just treat it as option plugins of postcss-loader and override all default plugins.
As it can be array before, so this part is for back compatibility.

aldarund: Ah, i see, i got that it will override if its array, i was wondering why it can be array :) Didnt saw it as array before. But make sense.

aldarund (approved):

galvez (approved):

pi0 (approved):

Issue #3858, opened 2018-09-07
exposed some data


fengjiel:
### Version

[v1.4.2](https://github.com/nuxt.js/releases/tag/v1.4.2)

### Reproduction link

[http://no-link](http://no-link)

### Steps to reproduce

I think I found an unsafe factor when I typed '$nuxt' into the console that exposed some data, like the store, all the router and configurations, or whether it could be encapsulated like the vueJs

### What is expected ?

No relevant data is displayed

### What is actually happening?

when I typed '$nuxt' into the console that exposed some data

### Additional comments?

I don't know if this is a loophole. Forgive my poor English,

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7701">#c7701</a>)</em></sub></div>

manniL:
Related: https://github.com/nuxt/nuxt.js/issues/3132#issuecomment-377366437

manniL:
This reference (to the root instance) is needed to emit the `triggerScroll` event which is important for the `scrollBehavior` of the app.

If someone finds a better way to provide such an interface while not leaking that much data, we'd surely accept a PR for it.

But as everyone can simply set breakpoints in the browser and inspect variables, code ("obfuscated") and state, this shouldn't be a severe issue.

aldarund:
I dont think its issue.  Everything that is on client can be debugged, there no real way to protect what u pass to client

tolgap:
Exactly. It is the programmer's responsibility to not leak anything clientside. Any secrets should be handled serverside and returned clientside without leaking those secrets.

Nothing is stopping a visitor on your site to just run breakpoints on variables and exposing those secrets otherwise 🙂.

pimlie:
To be fair, normally I am the one who says 'obscurity is not security' but this time I disagree with the above opinions that we shouldnt hide it _because_ there are other ways to get the data. The fact that there are other ways doesnt mean we should make it easy for wrong-doers.
And yes, to obscure certainly doesnt mean its secure but obscuring is still a completely valid technique to at least lower the attack surface. 

We could go in to lengthy discussions about this, but that would take much more of everyone's time then just fixing this issue. We can use the amazing power of javascript scopes to only expose the nuxt features on the window object we really need :grinning: 

```js
// App.js
      window.nuxtTriggerScroll = () => {
        this.$nextTick(() => {
[...]

PR #3522, opened 2018-07-11
feat: configurable global name

JuliaNeumann:
First of, many thanks for Nuxt, it's a great project!!

This PR is related to / a partial solution for #1792, though the motivation is a bit different: The main goal is to be able to run multiple Nuxt apps on the same page (in a microservice / microfrontend setting), and this is a step towards it, as it makes it possible to set a custom string in the config which will be used instead of the default "nuxt" in the global variables & HTML wrappers.

I'd be happy to improve this based on any feedback you provide, also if you think a different approach should be taken here. But it would be awesome to see a feature like this included in Nuxt, as it would help both with the security concerns in the issue mentioned above and with the encapsulation needed for a microfrontend application.

PR #3930, opened 2018-09-21
fix: dont expose $nuxt in production

pimlie:
Exposing $nuxt is a potential security issue as it provides api access to the whole website

## Types of changes

## Description
See #3858 for discussion why exposing $nuxt in production should be avoided

## Checklist:


aldarund:
What about when someone want to run e2e test on production build, so needs $nuxt ?

pimlie:
Not sure if I understand you correctly, all the unit/e2e tests already run in production mode. See https://github.com/nuxt/nuxt.js/blob/dev/test/utils/index.js#L28

aldarund:
@pimlie i mean if end user want to run e2e test on their own site and want $nuxt for it

pimlie:
Then I guess they should use the same 'trick' as I added for the e2e tests by assinging `window.$nuxt` in a layout: https://github.com/nuxt/nuxt.js/blob/e2de318259953fa5c62dc664326d1f5b8732c391/test/fixtures/basic/layouts/default.vue

aldarund:
Yes, but then this should be documented.  Or maybe it just should be an option in config that will be disabled by default and when enabled $nuxt will be exposed,but not sure if it worth it

pimlie:
Ah ok sorry, now I understand. Maybe it would be an idea to do that by adding a `NODE_ENV=testing`? I think currently nuxt only differentiates between dev and production, Testing could then be the same as production but eg with $nuxt exposed?

pimlie:
@aldarund Thanks for the pointers, this is indeed a better solution :+1:

pi0:
I'm not against this enhancement but we have to consider:

- Vue dev tools depend on `window.$nuxt` to detect nuxt.js production websites. So we can expose a less exposed version of `window.$nuxt` in production instead of `window.nuxt`
- Some projects depend on `window.$nuxt` for example to manually update the progress bar. That's not a valid approach but we should mention it as a breaking change when merged and released.
- `window.$nuxt` sometimes makes production debugging easier.  So we may add an option inside nuxt.config to make it configurable (To any arbitrary one like `window.$nuxtSecureObj` or disable at all)
- As an attacker point of view, he/she can still access this reference using debugger tools.

My suggestion is to first complete works on #3522. (configurable global names)

PR #3940, opened 2018-09-22
fix: reload in case of missing chunk ( e.g. site updated)

### LINKED ISSUE #3389: [edge] Deployment of new version breaking site for live users. ( Loading chunk {n} failed error)
### Version

### Reproduction link

[None](None)

### Steps to reproduce

### What is expected ?

User see some cryptic error 

aldarund:
## Types of changes

## Description
Resolves: #3389
 Simplest solution for now, just reload window with current path in case of loading chunk error. Suggestions how to handle better welcome :)

## Checklist:


lib/app/client.js
```diff
@@ -140,6 +140,9 @@ async function loadAsyncComponents (to, from, next) {
     next()
   } catch (err) {
     err = err || {}
+    if (/^Loading chunk (\d)+ failed\./.test(err.message) && process.client) {
```
pi0: Is process.client necessary here?

lib/app/client.js
```diff
@@ -140,6 +140,9 @@ async function loadAsyncComponents (to, from, next) {
...
+      window.location.replace(location.href)
```
pi0: With this improvement, we may make behavior more configurable. Sometimes persisting store data is needed for example before redirect or something like that.

My samplet suggestion is something like scroll behavior. But serialization is not actually a flexible idea. We can think about possible ways.

lib/app/client.js
aldarund: not really as i understand, was just trying to be super safe

lib/app/client.js
aldarund: Ye, making it configurable sounds fine. I could make it same way as scrollBehaviour if there no better idea.

lib/app/client.js
Atinux: Indeed, we are already in app/client.js

Atinux:
Actually, I believe this could be the default behaviour for Nuxt to avoid displaying error.

One idea is to call the errorHandler before redirecting or some kind of hook so user could take control of it instead of reloading. WDYT?

aldarund:
Yes, some kind of control\override would be good. One way would be same way as with scrollbehaviour with serailizing function. Is there examples of similar things with hooks?

pimlie:
What about a nuxt.config.js option and/or a build parameter to keep the chunk files for the last X number of builds? Currently the dist folders (.nuxt/dist and dist/_nuxt) are indiscriminately removed when building or generating, but maybe we should add support to keep the chunks so we dont break 'running' apps in between site updates. Maybe even also add a method for indicating the chunk is still 'fresh' or out of date (probably a bit difficult to do this fully agnostic but i.e. a replaceable boolean string in the chunk files), so we can either show a message like 'a new version of this page is available, please reload' or just reload fully automatically on the next navigation instead.

This way we 'intelligently' fix the issue of missing chunks that we caused ourselves by updating our site without breaking the user-experience. And then we also know that in other cases its most probably best to really display a temporary or hard error.

pimlie:
@aldarund Maybe I am a bit slow on this monday, but why wouldnt my suggestion solve the problem **at all**? 

It seems to me that if you break down the issue this PR tries to fix there are three main possible reasons why an app running in a users browser cant find a chunk:
1. it doesnt exists anymore because a new version has been published
2. it never existed
3. there is a temporary lookup failure (eg network)

So why not try to handle each case the best we can while we are at it?

aldarund:
@pimlie ok, not at all, but the problem will still happen with your solution when deployments happen without previous sources, which is quite common case that certainly would happen . For what u are talking there is issues like this -> https://github.com/nuxt/nuxt.js/issues/3078
This pr intended to fix situation 1 ( and maybe 3) , without going too deep, because proper solution will require a lot more things.

husayt:
This is probably most important outstanding fix at the moment for Nuxt. As this issue is the only one I know which causes broken page for everyone on new deploy. Just need to make sure we don;t get infinite reload problem.

Thanks @aldarund

pi0:
> Or just replace the old chunks with a stub that does the page reload for us

This would be possible using a connect middleware to respond for all 404 requests matching `/_nuxt/*.js` with a window.reload. But it won't work for CDN deployments.

aldarund:
@husayt i have tested and was unable to get infinite reload. If page loaded directly ( e.g. via reload or direct navigation) it will throw app initialization error in console https://i.imgur.com/CWAFfFC.png ( thats current behaviour even without this PR )

usulix:
As a senior developer on a team with 2 current production Nuxt apps, I wanted to chime in here. 

first - Thanks for the great framework and for the hard work. I am working to get corporate to fund some support for the project :-)

I think @aldarund has a simple and workable fix, and I put it in my error.php file to handle the situation, but beyond that... when you ask the infinite reload question, for me it becomes an end-user question. It depends a lot on hosting and even more on deployment process. It also depends on what I want my end-users to experience. 

My personal opinion is that a Nuxt internals fix for this might be the wrong path, because I'm not sure the project can completely anticipate all the crazy things I might do as an end user nor all the jacked up ways I might deploy. Maybe this is a place where documentation needs to be clear and implementation is left up to the user.

just my 2 cents...

aldarund:
@usulix well, there wont be case when it will be infinite reload. If such errors happens and reload happens - app will fail initialiaze and no more logic will be working at all.

usulix:
@aldarund , thanks for the education on Discord... checks in every way I tried to emulate.. first chunk failed error in loaded app.. gets redirect.. on page refresh, I get an app load error and no infinite redirect.

fuzzybaird:
I would love to see this one through to dev :) Its a real problem for us in Prod and this implementation looks like it would be a reliable fix.

axetroy:
How about reloading the manifest and re-get the page again?

ahus1:
I've posted a cut-and-paste workaround with a reload-once-strategy to #3389 that works by adding it to the error page template (no internals need to be changed) - maybe this helps some of the users who commented here.

pi0:
@ahus1 I exactly thought about that. But there is a potential pitfall that using custom error layouts, users have to provide their own _missing chunk handler solution_.

ahus1:
@pi0 the workaround is only a workaround, but AFAIK this is what is possible with the versions released today. I am really looking forward for this PR to be completed, merged and released. Then the workaround form the error layout can be removed again.

manniL:
Should we still implement this after #4323? :thinking:

pi0:
@manniL While #4323 removed perf and runtime issues still missing chunks after a reload happen and such workaround may be added as a built-in functionality to auto-reload or gracefully handle this error.

manniL:
@pi0 Agreed :+1:

fuzzybaird:
I agree with @pi0, #4323 looks like it does not exactly address this issue. I guess where I am confused is why this implementation of #3940 is bad specifically? I have had over 30 failed chunks on production from deploying to our CDN since this PR was proposed. I would be happy to come up with an alternate solution to this if someone can guide me as to what they think a better path would be.

manniL:
What's the status? ☺️

aldarund:
@pi0 @manniL so do we go to for this simple solution or add more things like allow to override its behaviour etc?

main2018:
What should I do, just upgrade nuxt?

manniL:
@main2018 Not released yet, so either wait for 2.4 or use `nuxt-edge` (uninstall nuxt, install nuxt-edge, that's it :) )

main2018:
@manniL 3Q very much indeed，This problem has been bothering me for several days.

A-Tione:
> @main2018 Not released yet, so either wait for 2.4 or use `nuxt-edge` (uninstall nuxt, install nuxt-edge, that's it :) )

ecently, I have also encountered this problem. Have you solved it now?

PR #3965, opened 2018-09-25
feat: abstract minify and use value for all modes

### LINKED ISSUE #3961: Nuxt 2 minifier no longer removing spaces between display:inline-block elements
### Version

[v2.0.0](https://github.com/nuxt.js/releases/tag/v2.0.0)

### Reproduction link

[https://github.com/guillermovs/nuxt2-display-inline](https://github.com/guillermovs/nuxt2-display-inline)

### Steps to reproduce

Install repo, run `npm run dev` and visit `http://localhost:3000`

### What is expected ?

The two `<section>` elements in the centre of the page should be shown inline (it worked prior to Nuxt 2).

A space is introduced between the `inline-block` elements, breaking them up into two lines.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7826">#c7826</a>)</em></sub></div>

### LINKED ISSUE #3964: Abstract generate.minify further and use values for all modes

https://github.com/nuxt/nuxt.js/issues/3961

We currently use the `generate.minify` values only for `generate` mode. It should be worth it to put it into build and use it for `SPA`, `SSR` and `generate` mode.

https://github.com/nuxt/nuxt.js/blob/b69cabef88a68de2376e29ee0995691e2e983c33/lib/common/nuxt.config.js#L140-L159

This is no breaking change as long as we might keep the defaults (`minify: true`).

Related code part: 
https://github.com/nuxt/nuxt.js/blob/b69cabef88a68de2376e29ee0995691e2e983c33/lib/builder/webpack/client.js#L56

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7827">#c7827</a>)</em></sub></div>

TheAlexLichter:
## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Description

Currently, we use `generate.minify` for configuring the `html-minifier` in `generate` mode.
With this PR we can use `build.minify` for *all* modes to fine-tune the minifier.

The value of `build.minify` will fall back to the old values in the respective modes to avoid a breaking change.

Resolves #3961, resolves #3964

## Checklist:
- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/799)


lib/builder/generator.js
```diff
@@ -223,7 +223,32 @@ export default class Generator {
       return false
     }
 
-    if (this.options.generate.minify) {
+    // Legacy workaround for obsolete options.generate.minify
+
+    if (typeof this.options.build.minify === 'undefined') {
+      this.options.build.minify = {
```
clarkdo: I think this can be default value of build.minify directly.

lib/builder/webpack/client.js
```diff
@@ -47,13 +47,19 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
   plugins() {
     const plugins = super.plugins()
 
+    // Legacy workaround for obsolete options.generate.minify
+
+    if (typeof this.options.build.minify === 'undefined') {
```
clarkdo: If we make above value as deault value, this part can be removed

clarkdo (commented):
Another thought is that if build.minify has some confusion with build.optimization.minifier?
How about htmlMinify or some name else?

lib/builder/generator.js
manniL: The problem is that we'd possibly introduce a breaking change because this value is not equal to `minify: true` or the resulting default values (which was the value for SSR/SPA layout minification).

manniL:
I agree on changing the name to `htmlMinify`! That fits better and doesn't cause as much confusion

galvez (approved):
@manniL just need to fix the unit test, otherwise LGTM

lib/builder/generator.js
clarkdo: I think there actually no difference to users because they don’t need change anything and also wont influence their project features, it’s just some specific options of minifier, let’s ask @pi0 and @Atinux opinion.

manniL: In the past things like script injection broke because of minification (where closing tags were removed), so I can't judge which problems the changed setting might cause. :see_no_evil: 

But it shouldn't be up to me :relaxed:

lib/common/options.js
```diff
@@ -209,6 +209,13 @@ Options.from = function (_options) {
     delete options.render.gzip
   }
 
+  // Legacy generate.minify
+  if (options.generate.minify) {
```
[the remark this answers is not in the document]
manniL: How do you mean that? :thinking:

lib/builder/generator.js
clarkdo: Yep, but current default options are more conservative than true, right? It should only fix sth instead of breaking 😅

pi0 (dismissed):

lib/builder/webpack/client.js
```diff
@@ -47,13 +47,18 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
   plugins() {
     const plugins = super.plugins()
 
+    // Legacy workaround for obsolete options.generate.minify
+    if (typeof this.options.build.htmlMinify === 'undefined') {
```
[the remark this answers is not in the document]
clarkdo: @pi0 How do you think make that option object as default value, then this part can be removed, more details you can see our discussion above.

lib/builder/generator.js
```diff
@@ -223,9 +223,34 @@ export default class Generator {
       return false
     }
 
-    if (this.options.generate.minify) {
+    // Legacy workaround for obsolete options.generate.minify
+
+    if (typeof this.options.build.htmlMinify === 'undefined') {
+      this.options.build.htmlMinify = {
+        collapseBooleanAttributes: true,
```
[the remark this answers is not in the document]
clarkdo: If it’s default value in nuxt.config, it’s solved automatically

lib/builder/webpack/client.js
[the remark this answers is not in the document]
pi0: Agree about setting defaults :)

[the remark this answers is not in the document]
manniL: I'll go for this path then :ok_hand:

manniL:
Refactored the changes to use the former `generate.minify` as default value for `build.htmlMinify` and added a warning concerning overrides

lib/common/options.js
```diff
@@ -209,6 +209,14 @@ Options.from = function (_options) {
     delete options.render.gzip
   }
 
+  // Legacy: Override build.htmlMinify with generate.minify if present
+  if (this.options.generate.minify) {
+    this.options.build.htmlMinify = this.options.generate.minify
```
clarkdo: This part still have the issue we just discussed, I think it should only be executed in generation

manniL: But how can we distinguish between `generated` and "normal" SSR?

clarkdo: Unfortunately we can’t, so maybe just warning default value willl be applied but not override build options? @pi0 How do you think?

pi0: That would be a SOFT BREAKING CHANGE :D Breaking change but warn to fix it :D But i agree.

manniL: That's why I was unsure if we should really proceed this way. Also because can't oversee how much damage the new "defaults" can cause for SSR HTML minification :thinking:

manniL:
To avoid breaking changes, the minification settings will be overridden in the `generator` module when `generate.minify` is present (but not in `options`). 

We can then introduce the minification *improvements* that are set as default values of `build.htmlMinify` in 2.1 without risking a breaking change.

lib/builder/generator.js
```diff
@@ -223,9 +223,17 @@ export default class Generator {
       return false
     }
 
+    let minificationOption = this.options.build.htmlMinify
+
+    // Legacy: Override minification options with generate.minify if present
     if (this.options.generate.minify) {
+      minificationOption = this.options.generate.minify
+      consola.warn('generate.minify has been deprecated. Use build.htmlMinify instead!')
```
clarkdo: If we all agree to do this way, I suggest adding a message that this option will be removed in 2.1

manniL: You mean `generate.minify`? Though it's not super clean, I wouldn't remove the override in the next minor version. Otherwise, we don't follow semver.

clarkdo: Yeah, what I mean is just making warning message more clear

manniL: Would `...deprecated and will be removed in the next major version` be sufficient?

clarkdo: Yep, good

manniL:
I have no clue why the CI fails. Will investigate later on. 
If **anybody** has an idea or so, feel free!

manniL:
I had to remove `sortAttributes` from the list of defaults because it broke the normal templates.
I'm confused though: Does minification happen *before* or *after* injecting the template  (or *both*)? I've tried to trace it down and saw replacing the template happening quite early. Unfortunately hadn't time to dig in deeper.

pi0 (approved):

aldarund (approved):

atinux (approved):
Good for me, could you add a PR to nuxt/docs?

manniL:
Will add a doc PR, yup!

Issue #4002, opened 2018-09-28
Inconsistent builds


pi0:
### Meta

- Hotfix: #4003
- Related PR: vuejs/vue-loader#1409

### Version

[v2.0.0](https://github.com/nuxt.js/releases/tag/v2.0.0)

### Reproduction link

dev

### Steps to reproduce

- `./bin/nuxt build examples/hello-world`
- `NODE_ENV=production ./bin/nuxt build examples/hello-world`

### What is expected?

Results should be the same. 

### What is actually happening?

Chunk names are different. ~~Also things like `s.options.__file` should be stripped but it is not.~~

### Additional comments?

Normal build:

```
Hash: 77622da25e0d579af1b6
Version: webpack 4.20.2
Time: 8670ms
Built at: 2018-09-28 12:43:39
                  Asset       Size  Chunks             Chunk Names
87c9ae15a64580abbac1.js   29.5 KiB       0  [emitted]  app
4767e3f41bc29ddaa1e5.js  494 bytes       1  [emitted]  pages/about
7d029e4429ef3a4557bf.js  437 bytes       2  [emitted]  pages/index
67347ecf5af03458ef7b.js    133 KiB       3  [emitted]  commons.app
ce430f2c9428f9c92feb.js    2.2 KiB       4  [emitted]  runtime
               LICENSES  422 bytes          [emitted]
 + 3 hidden assets
Entrypoint app = ce430f2c9428f9c92feb.js 67347ecf5af03458ef7b.js 87c9ae15a64580abbac1.js

Hash: 3a2a3ea1bb1bb0a04c93
Version: webpack 4.20.2
Time: 669ms
Built at: 2018-09-28 12:43:39
             Asset     Size  Chunks             Chunk Names
server-bundle.json  147 KiB          [emitted]
Entrypoint app = server-bundle.js server-bundle.js.map
```

When setting `NODE_ENV=production`: 

```
Hash: 1e074dabd64eeb326435
Version: webpack 4.20.2
Time: 7456ms
Built at: 2018-09-28 12:44:03
                  Asset       Size  Chunks             Chunk Names
964f935295162117ed63.js   29.4 KiB       0  [emitted]  app
06cd332c6aa0e8b0c1f0.js  467 bytes       1  [emitted]  pages/about
efb705924dd2b7ecdb9b.js  410 bytes       2  [emitted]  pages/index
67347ecf5af03458ef7b.js    133 KiB       3  [emitted]  commons.app
[...]

PR #4003, opened 2018-09-28
fix(cli): apply default value for NODE_ENV environment variable

pi0:
## Types of changes

## Description
Related to #4002. Some tools (Including the current version of `vue-loader`) implicitly use `process.env.NODE_ENV` to align their options. This change only sets the best value for `NODE_ENV` if not already set so it won't break anything. 

- The reason for including before any `require()` is that some modules may need this env during evaluation.
- The reason for not refactoring this logic inside `bin/nuxt` is binaries portability. (nuxt-start directly is being called) 

## Checklist:

- [ ] All new and existing tests pass.

### DIFF
--- a/bin/nuxt-build
+++ b/bin/nuxt-build
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+
 const parseArgs = require('minimist')
 const consola = require('consola')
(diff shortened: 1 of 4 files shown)


TheAlexLichter (approved):

manniL:
Related (though closed already): https://github.com/nuxt/nuxt.js/issues/2998

aldarund (approved):

clarkdo (approved):

PR #4012, opened 2018-09-29
feat: configurable global name 

### LINKED ISSUE #1792: Hiding references to __nuxt in the scripts and the template ID
I've been reading the source code and found out that `window.__nuxt__` and the id for the App component `#__nuxt` are hardcoded. For security reasons I would like to hide what technology is powering my site, do you have plans on having those as options or should I work on my own fork? 

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt/nuxt.js/issues/c1605">#c1605</a>)</em></sub></div>

galvez:
Picking up on https://github.com/nuxt/nuxt.js/pull/3522 which's been abandoned for some time.

## Types of changes

## Description
Permits running multiple Nuxt apps with different names on the same page by introducing a new `globalName` configuration option, along with granular global options (`globals`) based on `globalName` by default.

Resolves #1792

## Checklist:


pi0:
I suggest instead of using a single `globalName`, add a configuration to `nuxt.config` to allow customizing different global names. This helps to avoid workarounds like `._on<%= _.capitalize(globalName) %>Loaded` and also the ability to only customize it when needed.

Config example:

```js
globals: {
 rootId: `__nuxt`,
 vuePrototype: `$nuxt`,
 global: `window.$nuxt`,
 state: `window.__NUXT__`,
 readyCbs: `window._nuxtReadyCbs `,
 onLoaded: `window._onNuxtLoaded`,
}
```

In the example above, I've extracted `window.` parts into the configuration too. So we can have even more low-level control over where exactly global callbacks and instance is injected!

lib/app/App.js
```diff
@@ -100,8 +100,8 @@ export default {
           return resolvedLayouts[_layout]
         })
         .catch((e) => {
-          if (this.$nuxt) {
-            return this.$nuxt.error({ statusCode: 500, message: e.message })
+          if (this.$<%= globalName %>) {
+            return this.$<%= globalName %>.error({ statusCode: 500, message: e.message })
```
pi0: Why we need to make this internal variable customizable?

lib/app/index.js
```diff
@@ -135,7 +135,7 @@ async function createApp (ssrContext) {
     store[key] = app[key]
     <% } %>
     // Check if plugin not already installed
-    const installKey = '__nuxt_' + key + '_installed__'
+    const installKey = '__<%= globalName %>_' + key + '_installed__'
```
pi0: I suggest to something like `globals.plugin` or `globals.pluginPrefix` instead of directly using `globalName`

lib/app/index.js
```diff
@@ -153,8 +153,8 @@ async function createApp (ssrContext) {
   <% if (store) { %>
   if (process.client) {
     // Replace store state before plugins execution
-    if (window.__NUXT__ && window.__NUXT__.state) {
-      store.replaceState(window.__NUXT__.state)
+    if (window.<%= globals.state %> && window.<%= globals.state %>.state) {
```
pi0: Instead of `state.state` can't we use something like `context`?

lib/common/options.js
```diff
@@ -132,6 +132,23 @@ Options.from = function (_options) {
...
+    ? options.globalName.toLowerCase()
+    : 'nuxt'
+
+  // Call globalName option function setters
+  for (const globalNameOption in options.globals) {
+    if (typeof options.globals[globalNameOption] === 'function') {
+      options.globals[globalNameOption] = options.globals[globalNameOption](options.globalName)
+    }
+  }
+  for (const globalCallbackOption in options.globals.callbacks) {
```
pi0: What if flatten `globals.callbacks` into `globals.readyCallback`, etc. This way we can eliminate this for loop. (If more root keys like this are added we have to repeat this logic)

lib/common/options.js
```diff
@@ -132,6 +132,23 @@ Options.from = function (_options) {
     options.store = true
   }
 
+  // Set the name for global variables - should contain only letters
```
pi0: I suggest moving this into a getter function inside the Builder module. This way, modules have a chance to override options. Also, we don't mutate options anymore.

lib/app/index.js
galvez: I like this suggestion, will change ASAP.

lib/common/options.js
galvez: Thanks, I'll work something out.

lib/common/options.js
galvez: @pi0 I was a little uneasy about it because I feel `callbacks.something` is cleaner than `somethingCallback`, given we've been following an hierarchical style, so it made sense to try and group these things semantically. But as long as it's only a couple of callbacks, I see no harm. Will change before we merge.

pi0 (approved):
👌 Sexy

TheAlexLichter (approved):
Really well done :clap:

manniL:
The only thing we might want to consider is that this might break `Nuxt` detection for dev tools, right? :thinking:

aldarund (approved):

qm3ster (approved):

pi0:
Waiting for CI

axetroy:
also ``nuxt-progress`` and ``nuxt-link-active`` class name for element

monsterboom:
nice~But the documentation need to be updated~I just set 'globalName' in 'nuxt.config.js', but the 
progress dom cant take effect.

manniL:
@monsterboom Yeah, that's right. (https://github.com/nuxt/docs/issues/842)

Make sure you are on version 2.2!

monsterboom:
@manniL yes,i'm sure, i mean, the dom '<div id="__nuxt"></div>' is changed to my 'globalName', but '<div class="nuxt-progress"></div>' is not changed.  I  npm install lastest just now. I checked package.json and node_modules and the nuxt version was 2.2.0 . Another problem, when i set 'globalName', the browser console a error: 'Cannot read property 'constructor' of undefined' .

pi0:
>  When i set 'globalName', the browser console a error: 'Cannot read property 'constructor' of undefined'

What value do you use for `globalName`?

monsterboom:
the value just my project name "epweb". i found the git tag 2.2.0, the file nuxt-loading.vue is not be changed like in the b1f7123 , https://github.com/nuxt/nuxt.js/blob/v2.2.0/lib/app/components/nuxt-loading.vue

sam3d:
@monsterboom I've found this issue appears to go away when I disable Vue DevTools (as per #4296)

lib/app/App.js
```diff
@@ -53,10 +53,13 @@ export default {
...
   created () {
     // Add this.$nuxt in child instances
-    Vue.prototype.$nuxt = this
+    Vue.prototype.<%= globals.nuxt %> = this
     // add to window so we can listen when ready
     if (typeof window !== 'undefined') {
-      window.$nuxt = this
+      window.<%= globals.nuxt %> = this
+      <% if (globals.nuxt !== '$nuxt') { %>
+      window.$nuxt = true
```
Roriz: What do you think in apply `pluginPrefix` instead of only `true`?
Will still be a truly and external resources can find the Vue instance (like a vuedevtools).

edum18:
Hello, I have multiple separated apps and need to implement what was discussed here. I've read [docs](https://nuxtjs.org/api/configuration-global-name/) but can't understand how to show two apps on the same page. I'll be very grateful if anyone can write a specific example

PR #4014, opened 2018-09-29
feat: use understandable file names for analyze mode

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/bin/nuxt-build
+++ b/bin/nuxt-build
@@ -51,6 +51,10 @@ if (argv.analyze && typeof options.build.analyze !== 'object') {
   options.build.analyze = true
 }
 
+if (options.build.analyze) {
+  consola.warn('Notice: Please do not deploy bundles built with analyze mode, it\'s only for analyzing purpose')
+}
+
 // Silence output when using --quiet
 if (argv.quiet) {
   options.build.quiet = !!argv.quiet
--- a/bin/nuxt-build
+++ b/bin/nuxt-build
@@ -51,10 +51,6 @@ if (argv.analyze && typeof options.build.analyze !== 'object') {
   options.build.analyze = true
 }
 
-if (options.build.analyze) {
-  consola.warn('Notice: Please do not deploy bundles built with analyze mode, it\'s only for analyzing purpose')
-}
-
 // Silence output when using --quiet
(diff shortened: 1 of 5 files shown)


lib/builder/webpack/client.js
```diff
@@ -15,6 +15,16 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
     super(builder, { name: 'client', isServer: false })
   }
 
+  getFileName(...args) {
+    if (this.options.build.analyze) {
+      const key = args[0]
+      if (['app', 'chunk'].includes(key)) {
+        return '[name].js'
```
pi0: What's the behaviour of making a production build with `--analyze` flag? Does it remove hashes? If yes I think it is a slightly breaking change for some users. (Not ME :D)

clarkdo: It's not for production, just for `yarn build --analyze`.

If I want to watch bundle size and use `yarn build --analyze`, the file name in report is [name] should be more understandable, I suppose that nobody will use `yarn build --analyze` to generate production bundles, it's just for generate report.

pi0: I think added this feature for a while. Not sure why and by was reverted 😆  Anyway I like it and just wanted to mention that fact. And also a really nice idea about the warning.

pi0 (approved):

bin/nuxt-build
```diff
@@ -51,6 +51,10 @@ if (argv.analyze && typeof options.build.analyze !== 'object') {
   options.build.analyze = true
 }
 
+if (options.build.analyze) {
+  consola.warn('Notice: Please do not deploy bundles built with analyze mode, it\'s only for analyzing purpose')
```
pi0: This may be shown for `nuxt-generate` command too. Maybe refactor it into Builder or just duplicate.

lib/builder/webpack/client.js
clarkdo: Thanks for the reminding, I also remembered it's a legacy feature, so I marked pr as a bug fix. 😸

bin/nuxt-build
clarkdo: Good idea

PR #4018, opened 2018-09-30
feat(builder): allow to customize or disable minimizer plugins

pi0:
## Types of changes

## Description

**Background:** CSSNano is currently applied twice. Once with PostCSS loader and once with optimizeCSSAssetsPlugin. This is normally not harmful and good indeed but for incompatible CSS assets like Framework7 is problematic and currently there is not any way to disable or customize options.

This enhancement allows toggling or providing more options to both terser and optimizeCSS plugins. Also adds support for setting `false` value to `minimize` to disable them at all (for testing purposes and more webpack options compatibility)

## Checklist:

- [x] I have updated the documentation accordingly. (PR: nuxt/docs#805)

### DIFF
--- a/lib/builder/webpack/client.js
+++ b/lib/builder/webpack/client.js
@@ -100,7 +100,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
     const config = super.extendConfig(...arguments)
 
     // Add minimizer plugins
-    if (!this.options.dev && config.optimization.minimizer === undefined) {
+    if (config.optimization.minimize && config.optimization.minimizer === undefined) {
       config.optimization.minimizer = []
 
       // https://github.com/webpack-contrib/terser-webpack-plugin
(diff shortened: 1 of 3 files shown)


lib/builder/webpack/client.js
```diff
@@ -99,31 +99,36 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
...
-          filename: 'LICENSES'
-        },
-        terserOptions: {
-          output: {
-            comments: /^\**!|@preserve|@license|@cc_on/
-          }
-        }
-      })
-      config.optimization.minimizer.push(terserJsPlugin)
+      if (this.options.build.terser) {
```
clarkdo: So if minifier configed, teaser config is useless. Do we still need this check?

clarkdo: How about enable teaser if minifie enabled and no specified minifier.
Then this check can be removed

pi0: It is to disable terser only. `optimize.minify` is a webpack option indeed.

https://webpack.js.org/configuration/optimization

PR #4021, opened 2018-09-30
feat(cli): support async nuxt.config.js

### LINKED ISSUE #4005: Async data inside nuxt.config.js

Currently there is no nice way how to use async data inside nuxt.config.js. 
For example getting some module configuration from remote server, getting head data, list of plugins that need to be enabled for current build etc.

This would allow to use same Nuxt app for multiple similar sites but with slight difference of configuration that is changeable from remote admin panel.

Something like this (quite pseudocode) 
```js
// nuxt.config.js

const getHeadConfig = async () => {
  let { data } = await axios.get('/my/head/config/url')
  return {
    ...data,
    title: 'hurrdurr'
  }
}

module.exports = {
  mode: 'universal',

  // Get conf from remote with axios
  loading: axios.get('/my/loading/config/url'),

  // Use some method to get data from remote
  head: getHeadConfig()
}

```

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7860">#c7860</a>)</em></sub></div>

TheAlexLichter:
## Types of changes

## Description

With this PR, the `nuxt.config.js` export can be `async` now and will be `await`ed properly.
Resolves #4005 

## Checklist:
- [ ] I have updated the documentation accordingly. (PR: #TODO)

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -41,7 +41,7 @@ export default async function build() {
     process.exit(0)
   }
 
-  const options = loadNuxtConfig(argv)
+  const options = await loadNuxtConfig(argv)
 
   // Create production build when calling `nuxt build`
   options.dev = false
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -1,6 +1,6 @@
 import parseArgs from 'minimist'
 import consola from 'consola'
-import { loadNuxtConfig } from '../common/utils'
+import { loadNuxtConfig, runAsyncScript } from '../common/utils'
 
 export default async function dev() {
   const { Nuxt } = await import('@nuxt/core')
@@ -50,9 +50,9 @@ export default async function dev() {
     process.exit(0)
   }
 
-  const config = () => {
+  const config = async () => {
     // Force development mode for add hot reloading and watching changes
-    return Object.assign(loadNuxtConfig(argv), { dev: true })
+    return Object.assign(await loadNuxtConfig(argv), { dev: true })
   }
 
   const errorHandler = (err, instance) => {
@@ -61,15 +61,15 @@ export default async function dev() {
   }
 
   // Start dev
-  (function startDev(oldInstance) {
+  async function startDev(oldInstance) {
     let nuxt, builder
 
     try {
-      nuxt = new Nuxt(config())
+      nuxt = new Nuxt(await config())
       builder = new Builder(nuxt)
-      nuxt.hook('watch:fileChanged', (builder, fname) => {
+      nuxt.hook('watch:fileChanged', async (builder, fname) => {
         consola.debug(`[${fname}] changed, Rebuilding the app...`)
-        startDev({ nuxt: builder.nuxt, builder })
+        await startDev({ nuxt: builder.nuxt, builder })
       })
     } catch (err) {
       return errorHandler(err, oldInstance)
@@ -96,5 +96,7 @@ export default async function dev() {
         // Handle errors
         .catch(err => errorHandler(err, { builder, nuxt }))
     )
-  })()
+  }
+
+  await runAsyncScript(startDev)
 }
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -38,7 +38,7 @@ export default async function generate() {
     process.exit(0)
   }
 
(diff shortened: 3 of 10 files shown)


manniL:
This seemed a little bit too easy though :thinking:

galvez (dismissed):
@manniL lovely :)

bin/nuxt-build
```diff
@@ -40,49 +40,51 @@ if (argv.help) {
...
+      .then(close)
+      .catch(err => consola.fatal(err))
+  } else {
+    // Build + Generate for static deployment
+    new Generator(nuxt, builder)
+      .generate({ build: true })
+      .then(close)
+      .catch(err => consola.fatal(err))
+  }
+})()
```
pi0: The result of this promise is unhandled. We can call it like this:

```js
build()
  .then(() => { process.exit(0) } )
  .catch(err => { consola.fatal(err); process.exit(1) })
```

bin/common/utils.js
```diff
@@ -49,6 +48,15 @@ exports.loadNuxtConfig = (argv) => {
     if (options.default) {
       options = options.default
     }
+
+    if (typeof options === 'function') {
+      try {
+        options = await options()
```
pi0: Usage of top-level `async/await` will break `nuxt-legacy` for Node 6 support. We have to decide to deprecate that package or use normal `Promise`.

bin/nuxt-build
manniL: Right! I guess same goes for the other bin scripts?

pi0: Yep :) Another idea is to create a wrapper util to refactor this logic. Something like `runScript`.

Impl:

```js
async function runScript(fn) {
  try {
   await fn();
   process.exit(0);
  } catch(e) {
consola.fatal(err)
consola.fatal(err)
 }
}
```

bin/common/utils.js
pi0: Alternative idea: Use webpack to transpile `nuxt-legacy/bin` scripts :D

manniL: Though all major cloud platforms support Node 8, I wouldn't like to see `nuxt-legacy` deprecated yet. So either Promises (which hurt the readability a little) or webpack (which introduces another build step, but only for the core-team and not for the users).

Webpack seems to be the better choice, doesn't it?

bin/nuxt-build
manniL: But there is the top-level async problem again :see_no_evil:

manniL: Ah, no. We can avoid it with promises as well.

bin/common/utils.js
pi0: Using babel or babylon to this is as simple as something like this:

```
yarn add --dev babel-cli
yarn babel bin -x ".js," -d packages/nuxt-legacy/bin
```

bin/common/utils.js
```diff
@@ -31,6 +31,15 @@ const getLatestHost = (argv) => {
   return { port, host, socket }
 }
 
+exports.runAsyncScript = (fn) => {
+  fn().then(() => { process.exit(0) })
+    .catch((err) => {
+      consola.fatal('Failed to run async Buxt script!')
```
pi0: Typo + Script is not always build :)

bin/common/utils.js
```diff
@@ -32,9 +31,20 @@ const getLatestHost = (argv) => {
...
 }
 
+exports.runAsyncScript = async (fn) => {
+  try {
+    await fn()
+    process.exit(0)
+  } catch (err) {
+    consola.fatal('Failed to run async Nuxt script!')
+    consola.fatal(err)
+    process.exit(1)
```
aldarund: Is it required? Is even second fatal executed? There is a hook to exit on fatal
https://github.com/nuxt/nuxt.js/blob/dev/bin/nuxt#L12-L19

manniL: Good catch!

pi0 (dismissed):
Except for yarn.lock, LGTM 👍

manniL:
Tests fixed! cc @pi0 @clarkdo

manniL:
@pi0 Updated to latest dev ☺️

pi0 (approved):

PR #4051, opened 2018-10-04
feat: refactors to convert nuxt into a monorepo

clarkdo:
## Types of changes

## Description

Changes summary: https://github.com/nuxt/nuxt.js/pull/4051#issuecomment-429611188

## Checklist:


galvez:
@clarkdo looking good!

pi0:
Seems we have lots' of conflicts now :(

clarkdo:
@pi0 Could you help me sync this branch with dev branch? The confilcts should be solved automatically

galvez (approved):

TheAlexLichter (approved):

pi0:
Changes summary: 

- Using Lerna and Yarn workspaces
- Prefer named exports for all packages
- Expose both `main` and `module` fields for each package (Fully ESM)
- Created specific packages for nuxt _distributions_:
  - `nuxt`
  - `nuxt-start`
  - `nuxt-legacy`
- Split nuxt into sub-packages:
  - `@nuxtjs/app`
  - `@nuxtjs/builder`
  - `@nuxtjs/core`
  - `@nuxtjs/common`
- Add `options.build.template` (replaces `options.nuxtAppDir`) so we can finally have a way to implement non-web/default templates (Weex, NativeScript, Framework7, etc) (Maybe react one day :D)

pi0:
TODO:

- [ ] We need a build step for development again! Either implement watch mode for `scripts/package.js` or Make `scripts/nuxt` working without dist files! (ESM bug?)
- [ ] Tests failing
- [ ] Review by all of core team members

pi0:
Some changes: module field and `src` directory will not be in published npm packages. Mainly because of the lack of proper ES support ATM. It also reduced packages size and complexity to handle.

Also currently we don't have much coverage for `@nuxtjs/cli` package (Like before) because Jest can't instrument `spawn`ed processes. I think having coverage can be achieved easily because of new CLI low-level command exports (See `packages/cli/src/index.js` and `packages/cli/commands`)

scripts/package.js
```diff
@@ -120,34 +120,68 @@ export default class Package extends EventEmitter {
...
-    }, this.pkg))
+    }, this.pkg)
+
+    if (_watch) {
+      // Watch
+      const watcher = watch(config)
+      watcher.on('event', (event) => {
+        switch (event.code) {
+          // The watcher is (re)starting
+          case 'START': return this.logger.debug('Watching for changes')
```
manniL: Could we "reference" the event codes somehow? It's unlikely that they'll break but it'd be *safer* to have a Map or a lookup object 🤔

packages/builder/src/builder.js
```diff
@@ -1,22 +1,23 @@
 import path from 'path'
 import fs from 'fs'
-
 import pify from 'pify'
```
manniL: @pi0 Can we replace it with `utils.promisify` or do we need it for the legacy build? 🤔

pi0: I'm on it ;)

pi0: @manniL Just created and published [pify-native](https://github.com/blindmedia/pify-native). But it seems that `pify` is way faster. So let's keep using it for now :) (https://github.com/sindresorhus/pify/issues/41#issuecomment-429988506)

manniL: Alright, thanks for the info @pi0!

galvez:
> because Jest can't instrument spawned processes

@pi0 I was eventually able to get it to work, that is, capture the output of buffered executables. I'll try and get it wired up after this PR gets in. Reading through.

pi0:
Locally merged with 9c1e0d174319f209bbfa4d34cfd2e17e82f45537

PR #4079, opened 2018-10-10
fix(app): don't change serialized head content

btkostner:
## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Description

I have a script that I am putting inline. This regex changes some of the code and breaks it.

```js
// nuxt.config.js

const compatibarPath = resolve(__dirname, './public/scripts/compatibar.js')
let compatibarFile = readFileSync(compatibarPath, 'utf-8')

export default {
 head: () => ({
   script: [
      { innerHTML: compatibarFile }
    ],

    __dangerouslyDisableSanitizers: ['script']
  })
}
```

## Checklist:

Maybe requires documentation if there are edge cases. I'm assuming there are some weird stuff with new JS syntax that caused this code to be included in the first place.

This removes code and will break other tests if it fails. If you guys still want tests, I can create some pretty basic ones.


manniL:
You could use an `appTemplate` for that and disable inline-script minification.

I'm not convinced that breaking change is needed 🤔

btkostner:
The problem with using an appTemplate is that I have to keep it in sync with nuxt or risk missing out on features and bug fixes.

Secondly, if for some reason you had some text in your description like this:
```
something with weird technical documentation about `:testing() { return 2 }`.
```
will end up showing as:
```
something with weird technical documentation about `:function() { return 2 }`.
```
because the regex isn't scoped to where the original problem was.

Overall, I think magic regex to fix an edge case bug is not a great solution, and this would be better without any magic and just documenting the edge cases.

btkostner:
The edge case I know of happens to be with the different ways you can declare a function in an object and how it gets serialized in the app.

```js
// This breaks
{
  head: {
    title: 'testing',
    titleTemplate (title) { return `${title} testing` }
  }
}
```

```
 ERROR  Failed to compile with 1 errors                                                                                                      09:47:25

 error  in ./.nuxt/App.js

Syntax Error: SyntaxError: /home/btkostner/tmp/nuxt-head/.nuxt/App.js: Unexpected token, expected "," (12:64)

  10 | 
  11 | export default {
> 12 |   head: {"title":"testing","titleTemplate":titleTemplate(title) { return `${title} testing` },"meta":[],"link":[],"style":[],"script":[]},
     |                                                                 ^
  13 |   render(h, props) {
  14 |     const loadingEl = h('nuxt-loading', { ref: 'loading' })
  15 |     const layoutEl = h(this.layout || 'nuxt')


 @ ./.nuxt/index.js 14:0-27 110:15-18
 @ ./.nuxt/client.js
 @ multi webpack-hot-middleware/client?name=client&reload=true&timeout=30000&path=/__webpack_hmr ./.nuxt/client.js
```

```js
// This works
{
  head: {
    title: 'testing',
    titleTemplate: (title) => `${title} testing`
  }
}
```

```js
// This works
{
  head: {
    title: 'testing',
    titleTemplate: function () { return `${title} testing` }
  }
}
```

galvez:
@btkostner I see that you [already touched this](https://github.com/nuxt/nuxt.js/commit/2fd1b229cbe6a91ec5c5ab64445badecacce765d) before. I've updated your branch with latest `dev`, pinging @pi0 @clarkdo @Atinux for a final review.

galvez:
ping @manniL @pi0 @clarkdo -- closable?

manniL:
The points are valid. We should further discuss this and try to minimize these "hacks" as far as we can.

galvez:
Picked up work/research on this in #4558 -- thanks @btkostner!

PR #4109, opened 2018-10-13
src/ -> lib/

galvez:
@clarkdo @pi0 @manniL perhaps we should maintain the `lib` convention?

See https://github.com/zeit/next.js/pull/5341/files


clarkdo:
What’s the reason for lib?

galvez:
@clarkdo @pi0 I just have a percetion that `lib` is the most commonly used convention for anything meant to be imported, internally or not. The related next.js PR was just an example.

PR #4142, opened 2018-10-18
Throw an error when layout property is not string

### LINKED ISSUE #4035: Throw an error when `layout` is no string or fn

The `layout` property should only allow strings and function. If you set the layout to an `array` (for example), no error is shown. Only a Warning (coming from Vue directly): 

```
[Vue warn]: Avoid using non-primitive value as key, use string/number value instead.

(found in <Root>)
```

A decent error message would be great to see here!

To get started, the source code for setting and evaluating the layout is [here](https://github.com/nuxt/nuxt.js/blob/dev/lib/app/App.js#L82)

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7888">#c7888</a>)</em></sub></div>

suusan2go:
## Types of changes

## Description

Resolves: #4035 

Avoid using non-string value as layout property.

## Checklist:

### DIFF
--- a/packages/app/template/App.js
+++ b/packages/app/template/App.js
@@ -83,6 +83,9 @@ export default {
     <% } %>
     <% if (splitChunks.layouts) { %>
     setLayout (layout) {
+      <% if (debug) { %>
+      if(layout && typeof layout !== 'string') throw new Error('[nuxt] Avoid using non-string value as layout property.')
+      <% } %>
       if (!layout || !resolvedLayouts['_' + layout]) layout = 'default'
       this.layoutName = layout
       let _layout = '_' + layout
@@ -110,6 +113,9 @@ export default {
     }
     <% } else { %>
     setLayout(layout) {
+      <% if (debug) { %>
+      if(layout && typeof layout !== 'string') throw new Error('[nuxt] Avoid using non-string value as layout property.')
+      <% } %>
       if (!layout || !layouts['_' + layout]) {
         layout = 'default'
       }


pi0:
Am I missing something or we should support layout as a function for [dynamic layouts](https://github.com/nuxt/nuxt.js/blob/dev/examples/dynamic-layouts/pages/index.vue#L10)?

manniL:
@pi0 Dang, you are correct! Almost forget that use case :see_no_evil: 

@suusan2go Could you do a check for string or function? Sorry!

suusan2go:
@pi0 @manniL 
I think check for function is unnecessary.
because if layout is a function,  it seems to call `layout = layout(app.context)` before `setLayout`.
https://github.com/nuxt/nuxt.js/blob/dev/packages/app/template/server.js#L123-L128
https://github.com/nuxt/nuxt.js/blob/dev/packages/app/template/client.js#L455-L464

Although I checked the `examples/dynamic-layouts` and it works perfect.

pi0:
@suusan2go. Thanks for checking that. I just wanted to to endure that it's happening.

pi0 (approved):

manniL:
Well, I fast to fast with my conclusion there 🙈 @suusan2go Thank you for looking into it!

The last thing I'd consider is maybe only throwing an `Error` if you are in `debug` mode 🤔

suusan2go:
@manniL Thank you for your comment! I changed my code to throw an error only in debug mode.

TheAlexLichter (approved):

PR #4147, opened 2018-10-18
In renderer if ssr rendering fails, fall back to spa rendering with warning

juno-yu:
Background:
creating a module as in https://codesandbox.io/s/qkxvnmv104 to take params for general redirect.

Under universal mode generate mode,
`let APP = await this.bundleRenderer.renderToString(context)` would fail, 
as bundleRenderer comes from `{ createBundleRenderer } from 'vue-server-renderer'`, in which having assumption that the route is having a template finally .

more background:
https://github.com/nuxt-community/redirect-module/issues/3

### DIFF
--- a/lib/core/renderer.js
+++ b/lib/core/renderer.js
@@ -306,7 +306,7 @@ export default class Renderer {
     const spa = context.spa || (context.res && context.res.spa)
     const ENV = this.options.env
 
-    if (this.noSSR || spa) {
+    const spaRenderer = async () => {
       const {
         HTML_ATTRS,
         BODY_ATTRS,
@@ -341,8 +341,18 @@ export default class Renderer {
       return { html, getPreloadFiles }
     }
 
+    if (this.noSSR || spa) {
+      return spaRenderer()
+    }
+
     // Call renderToString from the bundleRenderer and generate the HTML (will update the context as well)
-    let APP = await this.bundleRenderer.renderToString(context)
+    let APP
+    try {
+      APP = await this.bundleRenderer.renderToString(context)
+    } catch(_) {
+      consola.warn(`ssrRenderer fails, fall back to spaRenderer for path ${url}.`);
+      return spaRenderer()
+    }
 
     if (!context.nuxt.serverRendered) {
       APP = '<div id="__nuxt"></div>'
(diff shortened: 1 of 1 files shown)


lib/core/renderer.js
```diff
@@ -341,8 +341,18 @@ export default class Renderer {
...
+    if (this.noSSR || spa) {
+      return spaRenderer()
+    }
+
     // Call renderToString from the bundleRenderer and generate the HTML (will update the context as well)
-    let APP = await this.bundleRenderer.renderToString(context)
+    let APP
+    try {
+      APP = await this.bundleRenderer.renderToString(context)
+    } catch (_) {
```
pi0: The real error shouldn't be thrown away. It makes debugging much harder.

lib/core/renderer.js
```diff
@@ -306,7 +306,7 @@ export default class Renderer {
     const spa = context.spa || (context.res && context.res.spa)
     const ENV = this.options.env
 
-    if (this.noSSR || spa) {
+    const spaRenderer = async () => {
```
pi0: Good, refactor idea. But why it should be async? Also, can we rename it to `renderSpa`?

lib/core/renderer.js
```diff
@@ -341,8 +341,18 @@ export default class Renderer {
...
+      return spaRenderer()
+    }
+
     // Call renderToString from the bundleRenderer and generate the HTML (will update the context as well)
-    let APP = await this.bundleRenderer.renderToString(context)
+    let APP
+    try {
+      APP = await this.bundleRenderer.renderToString(context)
+    } catch (_) {
+      consola.warn(`ssrRenderer fails, fall back to spaRenderer for path ${url}.`)
```
pi0: I consider this a breaking change. Because Nuxt behaviour regarding SSR errors will be completely different

pi0 (commented):
Good path. But we need to assume more aspects. I'm going to work on a complete renderer refactor. Will take your PR into the account.

juno-yu:
Sure, and a more thorough solution for my proposed case might be to get the route considered by vue-server-renderer. Maybe let's get this closed first

PR #4155, opened 2018-10-19
fix: check styleResources for existence

aldarund:
## Types of changes

## Description
Check styleResource for its existence, so it dont fail silently without giving clue to user why its not working.

## Checklist:


packages/builder/src/builder.js
```diff
@@ -497,6 +498,18 @@ export default class Builder {
       }
     }))
 
+    // Check styleResource exist
```
manniL: Maybe [`options.js`](https://github.com/nuxt/nuxt.js/blob/dev/packages/common/src/options.js) is a better place for the check?

TheAlexLichter (changes requested):

packages/builder/src/builder.js
aldarund: Maybe. But why check for plugins.js is here and not in options then  ? :)

pi0: Wecare trying to scan file system which is only specific to builder ( l mean not required for production nuxt start) So here is probably better place and helps decoupling packages.

TheAlexLichter (dismissed):

pi0 (dismissed):

galvez (changes requested):
@aldarund 

```
$ ack "throw new Error" | grep -i not
builder/src/builder.js:445:          throw new Error(`Could not compile template ${src}: ${err.message}`)
builder/src/builder.js:478:        throw new Error(`Plugin not found: ${p.src}`)
core/src/module.js:41:      throw new Error('Template src not found:' + src)
core/src/resolver.js:86:    throw new Error(`Cannot resolve "${path}" from "${resolvedPath}"`)
```

To maintain consistency, can you change it to read `Style src not found:`?

aldarund:
@galvez not quite follow you. Style != Style Resource. Style would be misleading because it could be confused with css section. And it same style message as with plugin not found. And its not quite src because it could be glob pattern

galvez (dismissed):

manniL:
Conflicts again 🙈

pi0 (approved):

packages/webpack/src/builder.js
```diff
@@ -60,6 +65,18 @@ export default class WebpackBuilder {
       }
     }
 
+    // Check styleResource existence
+    const styleResources = this.context.options.build.styleResources
```
pi0: @clarkdo As this was a loader, I've just kept in the place as original PR was. What do you think?

Issue #4050, opened 2018-10-04
TypeScript TypeDefs Suggestion for Nuxt 2.x


kevinmarrec:
### What problem does this feature solve?

Developing with both Nuxt & TypeScript, I lately have been likely frustrated using the Nuxt API around pages using TypeScript decorators, so here is what I want to suggest :

Provide Nuxt TypeDefs to be able to use (and get AutoCompletion on) the Nuxt Pages API (https://nuxtjs.org/api/ => "Pages") around TypeScript decorators for class-style Vue component (`vue-class-component` https://github.com/vuejs/vue-class-component ).

Example (Autocompletion on `head` API property through MetaInfo typedef of `vue-meta`): 

![Screenshot_5.png](https://cmty.s3.eu-west-1.amazonaws.com/71995937/56da2480-c7cd-11e8-a9cb-39caeaa4d4c1.png)

`nuxt-class-component` would be not needed anymore, unless you want to put your Nuxt logic in your `export default class` without Editor Autocompletion and having trouble with TypeScript implict any !

### What does the proposed changes look like?

For now I ended up with these definitions, according the official Nuxt API Documentation:

![Screenshot_7.png](https://cmty.s3.eu-west-1.amazonaws.com/71995937/fafec0b0-c7ce-11e8-a9cb-39caeaa4d4c1.png)

As you may have figured out, `Context` is declared as `any` type, it's just a question of time to define it properly according to Nuxt API Context (https://nuxtjs.org/api/context ).

I would need to know if what I did is relevant, I heard about `types` which were maybe present on 1.x but disappeared since 2.x release (#795), so I wasn't sure if my work might have already been done somewhere, maybe it was about another type of types ?

Then I'll be glad to finish my work and propose a PR.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7907">#c7907</a>)</em></sub></div>

kevinmarrec:
Little update to let people what's going on here :

I will take time to build the type definition of the Nuxt `Context`, refering to https://nuxtjs.org/api/context.

Then I'll make a PR.

If you want to have Nuxt ComponentOptions TypeScript support right now in your project and you don't care (at least for now) having the Context declared as `any` type, you can simply add a `nuxt.d.ts` in a new `@types` folder in your Nuxt project :

```js
// name-of-your-project/@types/nuxt.d.ts
[...]

PR #4164, opened 2018-10-21
feat(ts): provide type definitions

kevinmarrec:
See #4050

**Progress**
- [x] Provide full ComponentOptions typedefs
- [x] Provide `$nuxt` typedefs
- [x] Provide `process` typedefs (`process.server`, `process.client`, ...) => Need to use a `global.d.ts` within user's project workspace
- [x] Test types
- [x] Decide with Core Team where should be placed these types in the new Nuxt monorepo architecture


galvez:
@kevinmarrec thanks for doing this!

I've added the WIP label and edited the title to match our log style. 
Please ping us when you are done.

pi0:
Thanks, @kevinmarrec for this PR. Same as @galvez please ping when we were good to go.

kevinmarrec:
@galvez @pi0, About Component "options", I just need to improve definitions for `Context` attributes which are functions : `redirect`, `error`, `beforeNuxtRender`.

For `redirect` & `error` I would need to be sure :
1) a) `redirect` can be called these ways `redirect(status)`, `redirect(status, path)`, `redirect(status, path, query)` ?
b) `path` is a string but what about `status` (number ?) and query ? (object ?) 
c) The function return `void` ?
2) `error` can be called this way : `error(params)` with params being an Object which must have `statusCode: number` and `message: string`

For `beforeNuxtRender`, I never used it and I would need help to properly define the function definition.

Then, I was considering providing definitions for the `this.$nuxt` thing, I know there is a `this.$nuxt.$loading.start()` thing, what are the files I would need to read and understand to know what's goin on around the `$nuxt` object ?

kevinmarrec:
@pi0 Thanks for your *hints* ! And about abstract type for Error-Like objects, I don't know.. Anyway we could just add one more interface in my definitions like I did for the `Transition` one.

kevinmarrec:
@pi0 https://github.com/nuxt/nuxt.js/blob/dev/packages/app/template/server.js#L78 Does this `path` is a third property of the error object that can take the `error` function ?

I'm also confused with https://github.com/nuxt/nuxt.js/blob/dev/packages/app/template/client.js#L39 , can `statusCode` be a string or is it only for rendering the default `error` component ?

kevinmarrec:
Can anyone can give me an example of how can be use the `Components` thing inside the function `fn` given to `beforeNuxtRender` within `Context` ?

Maybe should we update https://github.com/nuxt/nuxt.js/blob/cf6b0df45f678c5ac35535d49710c606ab34787d/test/fixtures/basic/pages/special-state.vue ?

I'm confused about the `type` of this `Components` param/property.

kevinmarrec:
Assumming https://github.com/nuxt/nuxt.js/blob/dev/packages/app/template/client.js#L553, I'm confused about the `Component.options`.

I thought that `Component` could be of type `ComponentOptions<Vue>` but around Vue typings there is nowhere a `options` property defined. It it something defined at runtime ?

@pi0 @galvez If you have some seconds or minutes to help me to understand what is this `Components` all about ?

husayt:
This is great. But I think it will be even better to merge `nuxt-class-component` and `nuxt-property-decorator` into this:
https://github.com/nuxt-community/nuxt-class-component
https://github.com/nuxt-community/nuxt-property-decorator
For meaningful typescript project you need all three. They need to merge to avoid duplications. So I think they all belong either here or in one community library.

kevinmarrec:
(Sorry, Mobile phone missclick for the closing/reopening PR thing)
Well @husayt thanks for your point of view but these packages are not really maintained/updated and I'm not sure why. And anyway these typedefs mentionned in this PR are related to Nuxt API itself so it will be published around Nuxt package namespace. I don't think the packages you mentionned will be needed anymore, but I may bé wrong.

kevinmarrec:
@husayt These are external packages and users are free to use them or not.

Here is my thoughts :

`vue-property-decorator` provides `Inject`, `Provide`, `Model`, `Prop`, `Watch` with better support than `nuxt-property-decorator`, and the library is maintained.

Then, the only thing changing between these two libraries is that `vue-property-decorator` uses `vue-class-component`, meanwhile `nuxt-property-decorator` uses `nuxt-class-component`.

The difference between the class components is that `nuxt-class-component` is an extended version of `vue-class-component`, which allows to "bridge gaps" and do weird things like defining a `layout` method in your class, without autocompletion, without any errors.

It's only a way to use the Nuxt API (layout, meta, asyncData, etc...) in TypeScript Component Class as methods.

Anyway my conclusion is that this discussion is ouf of the scope of this PR, here it's about typings, feels free to open a new issue about a more general Nuxt TypeScript topic if you want do discuss further.

kevinmarrec:
PR Status : More changes to come this week with the incoming help of @iwata (see discussion on the related #4050 issue.

hartmut-co-uk:
Hi @kevinmarrec I've locally had added following as well to `ComponentOptions`:
```
key?: Function
```

seems I've had forgotten to leave a comment where I've had to put it in place - but I think it's been related either to `watchQuery` or maybe to $refs of nuxt pages (the page components) - couldn't find any official documentation after a quick search - but it definitely was required on my project.

hartmut-co-uk:
also for other reference: https://github.com/wagerfield/nuxt-typescript/issues/6

hartmut-co-uk:
Hi, I'll try to find out - I think it might be the key|index of the page component reference - might have todo with the router? Maybe one of the core members can help out? @galvez @pi0 

e.g. I've got
```
  @Component({
    name: 'AccountPage',
    watchQuery: ['q'],
    key: (to) => to.fullPath
  })
  export default class extends Vue {
```

hartmut-co-uk:
Update: I tracked it down to following issue, see the 2 comments and the example pointed out by @Atinux 
https://github.com/nuxt/nuxt.js/issues/1255#issuecomment-361952492

kevinmarrec:
@hartmut-co-uk Cool !! Seems to be dealing with `<nuxt-child>` component.
https://github.com/nuxt/nuxt.js/blob/3235b037a075c4927c56c8891d7d80c2a4fe54d7/examples/routes-transitions/pages/users.vue#L25

Will dive into this later, I'll keep you updated here.

hartmut-co-uk:
correct, thanks!
Great job by the way! 🙇 
Really looking forward to the improved TS support!

kevinmarrec:
TODO :

- [ ] Use TypeScript **indexed access types** to define the `Context.params` & `Context.query` through `vue-router` typings
- [ ] Rename `Error` interface to `ErrorParams`
- [ ] Test that properties within `ErrorParams` interface are optional, and update the definitions if everything works
- [ ] Define the `Component` type to finish the definition of `beforeNuxtRender` within Context
- [ ] Improve `transition` type for JavaScript Hooks methods https://nuxtjs.org/api/pages-transition => https://vuejs.org/v2/guide/transitions.html#JavaScript-Hook
- [ ] Find how is used the `key` component option not mentionned in Nuxt documentation, and implement a definition for it

kevinmarrec:
@hartmut-co-uk I found https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt.js#L47

Definition is now trivial for the `key` property.

kevinmarrec:
@pi0 What are these weird things with `Merge branch 'dev' into typescript-component-options-definitions` ? I don't really get how rebases work between 2 remotes (my fork & official nuxt)

kevinmarrec:
Alright, I think I have finished everything, the current PR status is now listening for improvements (maybe missing parts ?) or reviews.

pi0:
@kevinmarrec Is it safe to move types into `@nuxt/types` and require it for `nuxt` package?

kevinmarrec:
@pi0 You want to put this in a @nuxt/types repo ? I would not recommend that.

TypeScript need your package types to be accessible like `node_modules/nuxt/types` OR `node_modules/@types/nuxt` to be referenced in `tsconfig.json` to make TypeScript aware.

I did how `vue`, `vue-router`, `vue-apollo`, `vuex` and more have done.

aldarund:
@kevinmarrec shouldnt there also be a process.client\server typedefs?  https://github.com/nuxt/nuxt.js/issues/2823

kevinmarrec:
@aldarund True !
Can your or someone enumerate what is added to the `process` object ?
As far as I searched into Nuxt source code I found :
- client
- server
- static
- browser
- mode

kevinmarrec:
@aldarund I wasn't able to make working the workaround of @qm3ster suggested in https://github.com/nuxt/nuxt.js/issues/2823.

I tried different things but couldn't be able to extend the `Process` interface of `NodeJS` 😥

aldarund:
process.browser deprecated, imo no need to provide typedef for it, since there wasn't one before this PR.
As for workaround i cant help here :( i dont use typescript.

distributions/nuxt/package.json
```diff
@@ -39,9 +39,12 @@
     }
   ],
   "files": [
-    "dist"
+    "dist",
+    "types/index.d.ts",
```
pi0: This typings are directly related to `@nuxt/vue-app` package. If we can't reference typings from a dependency package into main `index.d.ts` (It seems it should be published under `@types`), I suggest at least renaming it into `vue-app.d.ts` and importing in `index.d.ts` :D 	@kevinmarrec

pi0:
@kevinmarrec Also can you please add some fixtures that we can pass through tslint?

kevinmarrec:
@pi0 Responding to your review comment here as my answer is quite big :

I didn't find any documentation on the web that would approve that it's possible to reference typings from a dependency package :/

There is how I tell to TypeScript which packages' types I want :
screenshot_4

The idea was to have `nuxt` or `@types/nuxt` at the end, if we reflect with Nuxt 2.2.0.

But as Nuxt (2.3.0) is now a mono repo, and everything related to a Vue app is under package `vue-app`, I don't know how looks like `node_modules` with the monorepo behavior.
Is there a `@nuxt` folder now in `node_modules` ?

Cause maybe the thing we want is saying to TypeScript "I want `@nuxt/vue-app` for my Vue project, I don't care about other incoming typings for other packages cause I don't need them"

kevinmarrec:
@pi0 For the testing part, I will try to do something like the `test` folder here : https://github.com/vuejs/vue/tree/dev/types , a smaller one here : https://github.com/vuejs/vue-router/tree/dev/types

distributions/nuxt/package.json
kevinmarrec: (See my comment in the PR before reading this)

Best thing seems to move the types under `@nuxt/vue-app`, and people will enable these typings by refering `types: ['@nuxt/vue-app']` in their `tsconfig.json`.

NickBolles:
Just throwing it out there that I ran into the [internals ](https://nuxtjs.org/api/internals), for programmatic use, not having types, it doesn't look like this PR has them either. 

At the very least we should consider where they should live, if not define them in this PR. I can try to build some out as I go but I'm a little strapped for time lately.

kevinmarrec:
@NickBolles I think this PR will cover only the `vue-app` package.

I guess We'll then extend Nuxt typings to other packages like `core`, `vue-renderer`, `build`, `generator`, etc ...

But for example, for Nuxt modules, we can't code them in TypeScript for now, as far as I remember, so providing type for `core/module.js` is maybe not relevant yet if you want editor autocompletion around the Nuxt "options" that your module can take. But it would be still relevant if you want to extend Nuxt core parts, which is out of the scope I wanted to impose in the first instance.

pi0:
@kevinmarrec About monorepo, you are right. The package will be located into `node_modules/@nuxt/vue-app` directory. 

For reference idea, i was talking about using [`/// <reference types="" />`](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) tag. 

About the link that you provided (For `vue` repository), That sounds like backing to `@nuxt/types` (Centerizing all typings into a seperate package with proper namespaces)

BTW if you agree, please move typings to the `vue-app` package. We mention it in the docs that users should add `@nuxt/vue-app` into `tsconfig.json`. 

Summerizing we can have different ways to publish those typings that each has it's own cons/pros. I assume all of them are supported. Users just need to list path in `tsconfig.json`

1. Put relevant typings inside each sub-package (`vue-app` for this PR) That allows us easier maintenance.
2. Put everything into `@nuxt/types`. Which can be published into `@types/nuxt` too
3. Publish typings directly via `distributions/nuxt` wich i think is the best idea.

Considering the point that packages are published using _independent versioning_, i highly recommand the first approach + adding fixture.

pi0:
@kevinmarrec For process, i agree wiith you. They could be better inside `process.env.*`. For legacy support we can keep both them until next major release. You can just ignore it for this PR to get merged as soon as possible. And then making a PR (If you want me to do this or prefer do it yourself) that adds `process.env.server`/`process.env.client` + docs + typings for it :D /cc @aldarund what do you think?

kevinmarrec:
@pi0 Yep, first approach is the one I prefer too :)

I just moved them, but I've just seen I did weird stuff around my PR. I merged the `dev` branch into mine so that it's up to date. Cause I couldn't be able to rebase, I have weird conflicts. But there is now differents files changed since this merge 😱.

So I don't know what to do to correct the PR now, any ideas ?

For the `process` thing, I should be able to it myself BUT for typings, I think there is no one if we do the `process.env` thing, there is the type of the `env` property :

```ts
interface ProcessEnv {
  [key: string]: string | undefined;
}
```

It's a dictionnary, same as `Context.env`.
But as it seems that it needs to return a string, there is a problem now, as `process.client`, `process.server`, ... are boolean.

packages/vue-app/package.json
```diff
@@ -5,9 +5,12 @@
   "license": "MIT",
   "files": [
     "dist",
-    "template"
+    "template",
+    "types/index.d.ts",
```
pi0: Listing `types` would be enough (As by-convention we use a separate directory for `tests`.

kevinmarrec: Oh alright, I've just understood that `vue`, `vue-router`, etc... were listing each file to not include `tests` in the final published NPM package.

Indeed, listing `types` would be enough !

I will update this at the same time as the `tests`, but where should be the `test`/`tests` folder then ?

pi0: Currently, test fixtures are inside root `test/fixtures` so I suggest creating testable ts project for linting inside `test/fixtures/typescript`. I know this is not the best place but there are plans for moving each package tests into its own.

kevinmarrec: Alright, I'll dive into it soon (this week). If I'm facing any problems I'll post them in the main PR thread.

kevinmarrec: Maybe could we let tests inside `vue-app` and in root `package.json` have the script `test:types` to be `tsc -p packages/vue-app/types/test`

kevinmarrec: Tests really need to be "close" to the typings tested

pi0: @kevinmarrec As I said that's handled by Lerna. Types are also symlinked into `[root]/node_modules/@nuxt/vue-app/types/index`. And listing `@nuxt/vue-app` in tsconfig, _should_ be enough. If you encounter problems just commit and ping me :)

kevinmarrec:
@aldarund I don't know yet but what I can said it's that we shouldn't touch to `process`.
What was the reflection which resulted to use `process.client` over `context.isClient` ?

pi0:
@kevinmarrec Context is not available everywhere. (For for example `require` calls outside of `default export` that need `process.browser` guard)

kevinmarrec:
Well we could let as it is now if we find the solution to extend `NodeJS.Process` interface, but I still didn't succeed to make it working.

And I don't know if it's recommended to use `process` as a global state where we can put things to be accessed everywhere. Did you find any other projects using this "way" ?

aldarund:
@kevinmarrec next.js use process.browser :)  But it seems they have same issue -> https://github.com/zeit/next.js/issues/2177

kevinmarrec:
@aldarund ahah 😁, well I don't know, but we need to find something 😂

aldarund:
@kevinmarrec sapper even written in TS but still they provide process.browser

packages/vue-app/package.json
kevinmarrec: @pi0 Lerna handle this gently when running `tsc`, for example to test `Transition type`

```ts
import { Transition } from '@nuxt/vue-app'

// Pass
const test1: Transition = {
  name: 'pass'
}

// Fail
const test2: Transition = {
  fail: 'fail'
}
```

BUT it's a pain to code de tests, I only have autocompletion within editor if I do 
```ts
import { Transition } from '../../packages/vue-app'
```

kevinmarrec:
@aldarund The only solution I found to have `process.server`, `process.client`, `process.static` and `process.browser` working in my Nuxt TypeScript project is to make a `global.d.ts` with
```ts
declare namespace NodeJS {
  interface Process {
    browser: boolean;
    client: boolean;
    server: boolean;
    static: boolean;
  }
}
```

AND it only works with the `global.d.ts` within your project, I tried different ways to put it around external package types, it won't work.

kevinmarrec:
@pi0 Look my last commit ;) This Test suite should be enough.

pi0:
@kevinmarrec 

1. Rebased. Pull your remote branch.
2. Tests are passing an honestly I can't wait to see this improvement being merged :P (I'm sure too many are waiting too) so please confirm if your changes are stable to be merged
3. Besides of _unit like_ type tests, I wanted to have a nuxt project fixture to test things on a real project. We can do it later.
4. Typescript setup & structure need more detailed steps. (Including `global.d.ts`) If interested we can add typescript section to the nuxt docs. (/cc @manniL)
5. ❤️

pi0 (approved):

kevinmarrec:
@pi0 

1. Cool, but could you tell me what were your git steps ? Cause if it was a rebase it should have rewrote every single commit after `dev`, as I did once days ago but it was so ugly in Git History that I decided to take the `merge` approach.

2. It's ready ! Every interface/type definition here matches the Nuxt API documentation. And in the worse case I really did a mistake, I'll be not far to either update it quickly or provide support on what I've done.

3. That's a good point, we'll for sure do it.

4. Yeah a Typescript section to the nuxt docs could be awesome, count me in to help on that.

5. ❤

pi0:
@kevinmarrec I didn't locally _rebaed_ simply used Github interface :)

kevinmarrec:
@pi0 Alright :D

hartmut-co-uk:
👏 🏆

hartmut-co-uk:
are there any plans to add a typescript option/setup to 'create-nuxt-app' with the type defs now available?
Found https://github.com/nuxt/create-nuxt-app/issues/23
(and again referring to https://github.com/wagerfield/nuxt-typescript -> nuxt module https://github.com/wagerfield/nuxt-typescript/blob/master/index.js)

Atinux:
Thank you so much for your work @kevinmarrec, as well as @husayt & @hartmut-co-uk!

Do mind joining our Discord server and ping me so I can add you as contributors and see more solutions like this one about Nuxt & TypeScript?
https://discordapp.com/invite/VApZF5W

husayt:
@Atinux thanks for invite. I am already in discord, although not able to follow all the updates there, as so much is going there.

On a good note, I have converted HackerNews demo to Typescript and am going to publish it soon. I realised there were no realistic examples showcasing best practices of using Typescript with Nuxt, that should be filling this gap.  Also it will help to validate type definitions committed here. Overall, I found using nuxtjs with Typescript takes Nuxt experience to next level and once tried one can't go back to plain JS/ES.

Shall I add it as a branch in HN's repo or as a separate repo? If the latter could you please create a HackerNews NuxtTS repo and give me an access there.

NickBolles:
@husayt I'm curious to see what your best practices are. Ping me when you have it public?

husayt:
[Here](https://codesandbox.io/s/1q478mzyvj)  it is  codesandbox with Typescript version of Nuxt HN.

@pi0 @Atinux Shall I add it as a branch in HN's repo or as a separate repo? If the latter could you please create a HackerNews NuxtTS repo and give me an access there.

@NickBolles These are not claimed to be best practices, but rather a first attempt of putting  together working practical typescript sample, so that we can together arrive to best practices.

That is why I invite everyone to try and provide feedback. You will notice some warnings, some hacks which we should aim to replace/fix with proper solution.

To really showcase value of typescript I aim to add type definitions from news api and show how that clarifies the code and helps to avoid bugs.

kevinmarrec:
@husayt Nice work !! 🎉🚀

I'm working on a new incoming PR to add typescript support (no more typescript module needed !) With a "typescript" option in nuxt.config.js. it will be highly inspired by https://github.com/nuxt/nuxt.js/pull/4406, but with some changes and from another POV

husayt:
@kevinmarrec That will be great. One of the reasons for this exercise was trying to have a shared playground for trying out nuxt with typescript. Will be great if we can resolve workarounds and warnings. Also will be interesting to see suggestions 
* for index.d.ts
* for use of `nuxt-property-decorator`
*for  use of typescript module vs [nuxt-typescript](https://github.com/wagerfield/nuxt-typescript) (direct TS support suggested by @kevinmarrec will be even better)

It seems classes extended from `Vue` don't see methods defined by type Nuxt definitions. What might be causing this?

kevinmarrec:
@husayt `index.d.ts` should not be needed, it's maybe needed on Codesandbox, as for example, VSCode load automatically load types from `node_modules`.

The only declaration files needed should be the one where you extend the `NodeJS.Process` interface, to be able to use `process.server`, `process.client`, `process.browser` & `process.static`. (and another declaration file for TSX-only users).

Using `nuxt-property-decorator` or only `vue-property-decorator` is optional, and it's a choice. My POV on `nuxt-property-decorator` is that I don't like the hooks within the class with no autocompletion. (It may answer your question)

Thanks to me and contributors, there are now type definitions, which let you being able to use this "hooks" as Component options, with autocompletion and typechecking.

husayt:
>which let you being able to use this "hooks" as Component options,

If you mean `@Component({ scrollToTop: true })`, that is good.
But having an option to declare these as class methods/properties and still have auto-completion would be nice and that is what `nuxt-property-decorator`  is trying to solve. It is all about options: js/ts/tsx, html/pug, style/less/css/sass and etc.

PR #4170, opened 2018-10-22
feat: allow scrollToTop to be explicitly disabled

homerjam:
This enables us to manually override scroll position. For instance to offset scroll position before a transition.

## Types of changes

## Checklist:


manniL:
Hey @homerjam 

This would be a breaking change, so it won't be merged until Nuxt 3.0

However, you can adapt the `scrollBehavior` in your `nuxt.config.js` through the `router.scrollBehavior` key on your own :relaxed:

homerjam:
Hi @manniL 

Thanks, I'm using that for now.

My mistake I thought that this wouldn't be a breaking change as I thought scrollToTop was undefined by default. Perhaps there's a better way in that case, something more explicit. To clarify my use case is for regular (or parent) page components as opposed to child components.

Feel free to close if you feel this could be improved. Though I'd be interested to know if anyone else has come across a need for this.

clarkdo (commented):
@homerjam Could you add a unit test for it ? 😸

manniL:
@homerjam It actually isn't a breaking change. Seems like I mixed something up there 🙈 Sorry! Kudos to @clarkdo for the follow up ;)

homerjam:
I will when I can! Something along the lines of scrolling the page then checking the scroll position hasn't changed after a page navigation right?

homerjam:
PR in the works

galvez:
This is continuing in #4564 -- thanks @homerjam!

PR #4175, opened 2018-10-23
fix(app): lint templates

pimlie:
## Types of changes

## Description

This PR is for the actual linting of the app templates. Currently the errors below still exist, I expect that all `no-unused-vars` (except `interopDefault`) could actually be removed, but as I was not 100% sure I'd like your input on that.

image

## Checklist:


packages/app/template/utils.js
```diff
@@ -117,7 +118,6 @@ export async function getRouteData(route) {
 }
 
 export async function setContext(app, context) {
-  const route = (context.to ? context.to : context.route)
```
pimlie: Please check this removal with extra care.

packages/app/template/client.js
```diff
@@ -337,21 +337,21 @@ async function render (to, from, next) {
       Component._dataRefresh = false
       // Check if Component need to be refreshed (call asyncData & fetch)
       // Only if its slug has changed or is watch query changes
-      if (this._pathChanged && this._queryChanged || Component._path !== _lastPaths[i]) {
+      if ((this._pathChanged && this._queryChanged) || Component._path !== _lastPaths[i]) {
```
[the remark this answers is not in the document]
pimlie: Hehe, we actually did the same test more or less :taco: But I tought there could be a small chance this was an unnoticed bug and the wanted behaviour in this line was actually different then the default preferences.

galvez (approved):

packages/common/src/utils.js
```diff
@@ -389,3 +389,18 @@ const DYNAMIC_ROUTE_REGEX = /^\/(:|\*)/
...
+
+const WHITESPACE_REPLACEMENTS = [
+  [/[ \t\f\r]+\n/g, '\n'], // strip empty indents
+  [/{\n{2,}/g, '{\n'], // strip start padding from blocks
+  [/\n{2,}([ \t\f\r]*})/g, '\n$1'], // strip end padding from blocks
+  [/\n{3,}/g, '\n\n'], // strip multiple blank lines (1 allowed)
+  [/\n{2,}$/g, '\n'] // strip blank lines EOF (0 allowed)
+]
+
+export const stripWhitespace = function stripWhitespace(string) {
```
pi0: GREAT!

pi0 (approved):
You rock @pimlie. However i think there should be better ways of not messing up templates with `<%= isTest ? .... %>` but get along with it for now :)

pimlie:
Thanks for the quick review & merge. 

Most of the isTest's are for when serialize and/or devalue is used, I looked into pretty formatting those JSON structures but tought that would be to cumbersome to do because I couldnt find a formatter which supports a `startIdentation`. [stringify-object](https://github.com/yeoman/stringify-object) comes close, but then still you would need to do a `.split('\n').join('\n' + ' '.repeat(startIndent))` afterwards to have them correctly indented.

PR #4179, opened 2018-10-24
feat: check whether port is used before server creation

### LINKED ISSUE #4178: Conflicts with applications running first on the same port
### Version

[v2.0.0](https://github.com/nuxt.js/releases/tag/v2.0.0)

### Reproduction link

[https://github.com/ohjihwan-nbt/default-nuxt-structure](https://github.com/ohjihwan-nbt/default-nuxt-structure)

### Steps to reproduce

1. Create node application that makes HTTP request of port 4000 and execute
- Execute with node command after creating the following file

```javascript
const http = require ('http')

const server = http.createServer ()
server.on ('request', (req, res) => {
   res.end ('OK')
}}

server.listen (4000)
```

2. Create and run the Nuxt application that makes http request on port 4000
- Run `npm run dev` after the repository clone: https://github.com/ohjihwan-nbt/default-nuxt-structure

3. Check the results after requesting http: // localhost: 4000 from the browser
- Check 'OK' string response created in node application executed in # 1

### What is expected ?

I expect that the nuxt application will not be processed after the following error message after running the application.

Error: listen EADDRINUSE ::: 4000

After running `npm run dev`, I see the following message.

READY Listening on http: // localhost: 4000

### Additional comments?

I think there should be a check to see if the port is in use elsewhere before appServer is listening and also think it is correct to show a message such as ERROR rather than READY. 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8022">#c8022</a>)</em></sub></div>

fureweb-com:
## Types of changes

## Description

this commit resolves #4178

## Checklist:

### DIFF
--- a/packages/core/src/nuxt.js
+++ b/packages/core/src/nuxt.js
@@ -136,6 +136,15 @@ export default class Nuxt {
     this.readyMessage = null
   }
 
+  async isPortInUse(port) {
+    return await new Promise((resolve, reject) => {
+      const tester = require('net').createServer()
+        .once('error', err => (err.code == 'EADDRINUSE' ? resolve(true) : reject(err)))
+        .once('listening', () => tester.once('close', () => resolve(false)).close())
+        .listen(port)
+    })
+  }
+
   listen(port, host, socket) {
     return this.ready().then(() => new Promise((resolve, reject) => {
       if (!socket && typeof this.options.server.socket === 'string') {
@@ -176,10 +185,14 @@ export default class Nuxt {
             return reject(err)
           }
 
+          let { address: host, port } = server.address();
+
+          const isPortInUse = await this.isPortInUse(port);
+          if(isPortInUse) throw `Error: listen EADDRINUSE ${host}:${port}`;
+
           let listenURL
 
           if (!socket) {
-            ({ address: host, port } = server.address())
             if (host === '127.0.0.1') {
               host = 'localhost'
             } else if (host === '0.0.0.0') {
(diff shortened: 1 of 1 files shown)


pimlie:
Why can't you just listen for an error event on the existing server instance? Like the Node docs describe here: https://nodejs.org/api/net.html#net_server_listen

fureweb-com:
> Why can't you just listen for an error event on the existing server instance? Like the Node docs describe here: https://nodejs.org/api/net.html#net_server_listen

Sorry about that. Honestly I don't know well about net api, so I just use that way.

packages/core/src/nuxt.js
```diff
@@ -136,6 +136,15 @@ export default class Nuxt {
     this.readyMessage = null
   }
 
+  isPortInUse(port) {
+    return new Promise((resolve, reject) => {
+      const tester = require('net').createServer()
+        .once('error', err => (err.code === 'EADDRINUSE' ? resolve(true) : reject(err)))
+        .once('listening', () => tester.once('close', () => resolve(false)).close())
```
galvez: @fureweb-com seems we've got a problem with this line. It's not enough time for the port to be cleared and reused, as the failing tests suggest.

galvez (changes requested):

packages/core/src/nuxt.js
galvez: Also probably a good idea to follow @pimlie's idea.

fureweb-com:
When I ran it on MacOS, it was Okay.
I think this problem is not related to nuxt.js so I close this issue. sorry for the confusion.

PR #4180, opened 2018-10-24
chore: test utils cleanup

galvez:
Refactor `test/utils` to better separate Nuxt-specific imports/exports and generally clean up these files' headers. Also removed a comment mentioning a `createSportsSelectionView` which doesn't appear to be related to Nuxt at all.


packages/common/src/index.js
```diff
@@ -1,4 +1,7 @@
+import * as _Utils from './utils'
+
 export { default as Modes } from './modes'
 export { default as NuxtConfig } from './nuxt.config'
 export { default as Options } from './options'
 export * from './utils'
+export const Utils = _Utils
```
clarkdo: Now we export an object `Utils` includes all functions in `./utils` and also all functions on `index`, it looks kind of confusing, we'd better keep only one behaviour.

galvez: @clarkdo gotcha, makes sense. I'll modify.

galvez:
@clarkdo updated, tests passing :)

pi0 (dismissed):
LGTM

pi0 (approved):

PR #4193, opened 2018-10-25
chore: packaging and development improvements

pi0:
## Types of changes

## Description

- [x] Eliminate the build step that temporary required for #4051
- [x] Refactors on builder/webpack to relax dependencies and fix problems with ESM
- [x] Eliminate 'lodash/*' import warns
- [x] Use const instead of var for dist

## Checklist:

### DIFF
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -79,6 +79,10 @@ export default class NuxtCommand {
     return imports.generator()
   }
 
+  importWebpack() {
+    return imports.webpack()
+  }
+
   async getNuxt(options) {
     const { Nuxt } = await this.importCore()
     return new Nuxt(options)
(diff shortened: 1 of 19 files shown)


scripts/dev
```diff
@@ -0,0 +1,28 @@
...
+import consola from 'consola'
+
+import Package from './package.js'
+
+async function main() {
+  // Read package at current directory
+  const rootPackage = new Package()
+  const workspacePackages = rootPackage.getWorkspacePackages()
+
+  // Create a dev-only entypoint to the src
```
manniL: Small typo there 🙊 
```suggestion
  // Create a dev-only entrypoint to the src
```

packages/webpack/src/index.js
```diff
@@ -1 +1 @@
-export { default } from './builder'
+export { WebpackBuilder } from './builder'
```
clarkdo: Should be `as BundleBuilder` ?

pi0: @clarkdo No different. Explicit names help reducing namespace conflicts (Like importing two different builders). Will change it :)

clarkdo: I mean https://github.com/nuxt/nuxt.js/pull/4193/files#diff-4f235fc89b94513b32c2cc8b5896cc5aR92 is using `BundleBuilder`, but export name is `WebpackBuilder` 😄

pi0: You are right! Good catch :D

packages/cli/src/command.js
```diff
@@ -79,20 +79,25 @@ export default class NuxtCommand {
...
+
   async getNuxt(options) {
     const { Nuxt } = await this.importCore()
     return new Nuxt(options)
   }
 
   async getBuilder(nuxt) {
     const { Builder } = await this.importBuilder()
-    return new Builder(nuxt)
+    const { BundleBuilder } = await this.importWebpack()
```
clarkdo: Good way to separate the concerns !

clarkdo (approved):

PR #4195, opened 2018-10-25
feat(cli): add imports mock and cleanup

pimlie:
## Types of changes

## Description

Added a mock for `./src/imports` and did some cleanup possible due to new insights :)

## Checklist:


packages/cli/test/unit/dev.test.js
```diff
@@ -1,5 +1,4 @@
-import { consola } from '../utils'
-import { mockNuxt, mockBuilder, mockGetNuxtConfig } from '../utils/mocking'
+import { consola, mockNuxt, mockBuilder, mockGetNuxtConfig } from '../utils'
 
 describe.skip('dev', () => {
```
pi0: @pimlie Can we enable `dev` tests now?

pi0: Enabled to see CI results 💃

packages/cli/test/utils/mocking.js
```diff
@@ -4,16 +4,16 @@ import Command from '../../src/command'
 jest.mock('../../src/imports', () => {
   return {
     core: jest.fn().mockImplementation(() => ({
-      Nuxt: () => {}
+      Nuxt: function () {}
```
pi0: @pimlie Just curious why this was needed? :D

pi0 (approved):

packages/cli/test/utils/mocking.js
pimlie: It wasnt really, except maybe for a clearer (future) understanding of the code. Arrow functions can't be used as a constructor, as `Nuxt` etc are mocks for class constructors I thought it would be better to explicitly indicate that.
Also the only reason it wasnt needed because there are no tests which use these standard mocks directly, all dev tests use mockNuxt and mockBuilder to explicitly set an implementation.

PR #4198, opened 2018-10-25
feat: custom pushAssets function to control http2 push headers

### LINKED ISSUE #4011: http2 push headers - should be a function

Now there is no configuration to decide what resources we want to push to the client.
So if it is returning client we are pushing all the resources even if he has it.

If the ```http2: push``` will be a function we could decide if we want to push and what.
It will be huge client optimalization.

Now we have something similar in ```shouldPrefetch``` and ```shouldPreload```

Besides ```file``` and ```type``` we need ```request``` to check for cookies or some other headers and ```response``` to set some output.

This is just example how we could use new implementation.
(isReturningVisitor is some custom logic that check for client cookies)

```js
render: {
    http2: (req, res, file, type) => {
       return isReturningVisitor(req) === false && type === 'script'
    }
  },
```

Than for example [nginx](https://www.nginx.com/blog/nginx-1-13-9-http2-server-push/) can read the push headers and push the resources.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7870">#c7870</a>)</em></sub></div>

awronski:
## Types of changes

## Description
Resolves: #4011 

Adds `pushAssets(req, res, publicPath, preloadFiles)` option to `http2`.
Function returns array of links. Now we can conditionally decide which assets we want to push.

Undocumented filter function `shouldPush` was marked as `deprecated`

## Checklist:

### DIFF
--- a/packages/core/src/middleware/nuxt.js
+++ b/packages/core/src/middleware/nuxt.js
@@ -44,30 +44,18 @@ export default async function nuxtMiddleware(req, res, next) {
     if (!error && this.options.render.http2.push) {
       // Parse resourceHints to extract HTTP.2 prefetch/push headers
       // https://w3c.github.io/preload/#server-push-http-2
-      const pushAssets = []
       const preloadFiles = getPreloadFiles()
-      const { shouldPush } = this.options.render.http2
+      const { shouldPush, pushAssets } = this.options.render.http2
       const { publicPath } = this.resources.clientManifest
 
-      preloadFiles.forEach(({ file, asType, fileWithoutQuery }) => {
-        // By default, we only preload scripts or css
-        /* istanbul ignore if */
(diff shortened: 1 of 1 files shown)


pi0 (approved):
LGTM. As this is almost a refactor i think we can add tests later.

awronski:
@pi0 I am thinking about removing `publicPath` from arguments of `pushAssets(..)`. Because user has it in `nuxt.config.js`. What do you think?

pi0:
@awronski This value _may_ be different. It should be read from client manifest to follow the value passed at compile time.

clarkdo:
Doc and release note need to be updated

awronski:
@clarkdo I will update docs

PR #4202, opened 2018-10-25
refactor core into sub-packages

pi0:
## Types of changes

## Description
- New Packages:
  - `@nuxt/core` : module/nuxt/resolver
  - `@nuxt/server`: jsdom/server/resources
  - `@nuxt/vue-renderer`: renderer/spa-meta
- Decoupled `listen` and `showReady`  from `nuxt` into `nuxt.server` (With aliases for BW support)
- Created a new `Hookable` class that `Nuxt` class inherits from it
- Renamed `@nuxt/app` to `@nuxt/vue-app` for consistency.
- Refactors for `@nuxt/config`
- Increase coverage 

## Checklist:


packages/core/src/nuxt.js
```diff
@@ -1,43 +1,43 @@
...
-    )
-    this.resolveAlias = this.resolver.resolveAlias.bind(this)
-    this.resolvePath = this.resolver.resolvePath.bind(this)
+    // Deprecated hooks
+    this._deprecatedHooks = {
+      'render:context': 'render:routeContext' // #3773
+    }
 
+    // Add Legacy aliases
+    defineDeprecateGetter(this, 'nuxt.renderer', this.server, 'nuxt.searver')
```
galvez: @pi0 `nuxt.server`

galvez (dismissed):

packages/core/src/nuxt.js
pi0: Good catch! Thanks :)

packages/server/src/middleware/error.js
```diff
@@ -4,7 +4,7 @@ import consola from 'consola'
 
 import Youch from '@nuxtjs/youch'
 
-export default function errorMiddleware(err, req, res, next) {
+export default ({ resources, options }) => function errorMiddleware(err, req, res, next) {
```
galvez: @pi0 is this really meant to be an arrow function that returns a function?

pi0: @galvez Yes :) It is now a middleware factory that is made using a context. Previously we were using `bind(this)` for middleware which is not a good idea (Increases coupling)

galvez: @pi0 coolio

packages/core/package.json
```diff
@@ -9,33 +9,17 @@
   "main": "dist/core.js",
   "dependencies": {
     "@nuxt/common": "^2.2.0",
+    "@nuxt/server": "^2.2.0",
```
pimlie: In the light of all the split packages, should core really have a dependency on server? If you use express or another server from create-nuxt-app you shouldn't need to install the default nuxt server as well I guess

packages/generator/src/generator.js
```diff
@@ -153,7 +153,7 @@ export default class Generator {
     }
 
     // Render and write the SPA template to the fallback path
-    const { html } = await this.nuxt.renderRoute('/', { spa: true })
+    const { html } = await this.nuxt.server.renderRoute('/', { spa: true })
```
pimlie: Same as above, to me it doesnt sound logical that generator uses the server to renderRoutes. Could it maybe use renderer directly?

packages/core/src/nuxt.js
```diff
@@ -1,43 +1,43 @@
...
-    this.hook = this.hook.bind(this)
+    // Assign options and apply defaults
+    this.options = Options.from(options)
 
     // Create instance of core components
-    this.moduleContainer = new ModuleContainer(this)
-    this.renderer = new Renderer(this)
     this.resolver = new Resolver(this)
+    this.moduleContainer = new ModuleContainer(this)
+    this.server = new Server(this)
```
pimlie: Extending on my other comments, if you compare how you use @nuxt/server with how you'd use express then with express nuxt is a property of the server and not the otherway around (or two way). E.g. with this implementation of nuxt/server as a separate package my first thought would be that it would also be possible/needed to set something like `this.server = express()` to change the server type.
Or in other words, if core/nuxt and server are this tightly integrated why go through the trouble of splitting into separate packages ;)

pimlie (commented):
Maybe I missed some discussion about the why of this, if so then disregard my comments ;)

packages/core/src/nuxt.js
galvez: > Or in other words, if core/nuxt and server are this tightly integrated why go through the trouble of splitting into separate packages ;)

I have to say, I'm still finding it a bit too tightly coupled for the separation to make sense. But I believe it's a direction worth taking.

pi0:
@pimlie I agree with you. But unfortunately, at least with current Nuxt API, we can't easily break things up. For sure this just a part of improvements like @galvez said. Also, nuxt was designed to have a sexy API. While discussing with @Atinux, we need to keep `nuxt.render` without warns.

About reason of introducing `@nuxt/server`, it's more about separating of concerns. At least things like showReady are refactored out of the core :)

PS: I'm trying to decouple builder/generator and renderer as well as using an event bus.

pimlie:
@pi0 Thanks for taking the time to respond to my remarks, I know you are very busy already so its really appreciated. Hopefully you were aware I was playing the Devil's Advocate a bit as I fully support the changes in this pr :)

galvez (dismissed):

galvez (approved):

clarkdo (approved):

PR #4208, opened 2018-10-26
refactor(common): nuxt.config.js and dynamic server options (latest process.env)

galvez:
Shouldn't be needed after https://github.com/nuxt/nuxt.js/pull/4207#issuecomment-433521714


galvez:
@pi0 or maybe not? will check later, gtg now :)

aldarund:
Tests failing so not so easy :)

pi0:
@pimlie is our cli hero :D

pimlie:
It seems to me this could be merged after changing the test. The failing test is a direct result of the expectation that ./utils.js set a default for host/port. If we want to improve that by moving it to nuxt.config.js like all other defaults then we just cant test for that here.

galvez:
@pimlie ha, you're totally right -- I had a feeling about this too but couldn't look into it at the time. Will update.

galvez:
So I've stopped at trying to load a fresh version of NuxtConfig because we need to pick up fresh `process.env` values. Suggestions welcome :)

galvez:
Back later!

pimlie:
Why and where is the fresh config needed? Its a bit hard to follow your thought process with all the fix descriptions, but if you are trying to fix the failing test which checks `env-host` then you will never succeed in that ;)

These are all strict unit tests which (try) to only test one (class) method at the time (and only stuff from this package). As setting the host from env is now moved from this package to nuxt/common the `env-host` test can be removed here I guess (and it should be added to a nuxt/common test).

pimlie:
Ah ok, thanks. If we have that problem in nuxt/common then its probably easier to just call `jest.resetModules()` or (sometimes that doesnt work) `delete require.cache[...path.to.nuxt-config]` as it seems it'd only be a test issue and not a real world issue.

pi0 (approved):

PR #4212, opened 2018-10-27
refactor: move config files to @nuxt/config

galvez:
Following #4208 and recent discussion in #core.


pi0:
@galvez As we may also move some config related utils into this package, what's your idea to keep configs in a sub-directory of `packages/config/src` ? like `src/config` or `src/nuxt.config`?

galvez:
@pi0 yeah, I rearranged things a bit now, we had the same thought :) Still WIP

galvez:
@pi0 how do I generate the `dist` dir for a new package?

pi0 (approved):

PR #4231, opened 2018-10-30
feat: modern build

clarkdo:
## Types of changes

## Description

- [x] SPA modern build
- [x] SSR modern rendering
- [x] CLI command
- [x] Test
- [ ] Documentation
- [ ] Release Note

## Checklist:


packages/webpack/src/config/plugins/vue/modern.js
```diff
@@ -0,0 +1,92 @@
...
+        const tempFilename = path.join(this.targetDir, htmlPath, `legacy-assets-${htmlName}.json`)
+        await fs.mkdirp(path.dirname(tempFilename))
+        await fs.writeFile(tempFilename, JSON.stringify(data.body))
+        cb()
+      })
+    })
+  }
+
+  applyModern(compiler) {
+    const ID = `vue-cli-modern-bundle`
```
pi0: vue-cli?

clarkdo: The plugin is from vue-cli, I’m wondering what we should do, maybe changing it and prepending credit comment?

pi0: I think keeping credit/license in the code should be enough. You know it is somehow unusual to see a `vue-cli` related error when debugging a Nuxt app 😆

packages/webpack/src/config/modern.js
```diff
@@ -0,0 +1,31 @@
+import _ from 'lodash'
```
manniL: Please use single imports for lodash :raised_hands:

clarkdo: Good catch

packages/vue-renderer/src/renderer.js
```diff
@@ -1,26 +1,34 @@
...
     // Will be set by createRenderer
-    this.bundleRenderer = null
-    this.spaMetaRenderer = null
+    this.renderer = {
+      ssr: null,
+      modern: null,
+      spa: null
+    }
+
+    this.modernBrowsers = Object.keys(ModernBrowsers)
```
pi0: Can't we precompute this and put what we cant into JSON file? (To reduce startup time)

packages/webpack/src/config/base.js
```diff
@@ -25,11 +26,20 @@ export default class WebpackBaseConfig {
     this.loaders = this.options.build.loaders
   }
 
+  get colors() {
+    return {
+      client: 'green',
+      server: 'orange',
+      modern: 'blue'
```
pi0: cyan look better i think :D

pi0 (approved):

PR #4239, opened 2018-10-31
feat(cli): lock project to prevent simultaneous builds

pimlie:
## Types of changes

## Description

This pr adds a lock method for nuxt cli commands. The lock is used to prevent simultaneous builds for a project, so only one build can run at once.
The locking also helps automation, if you develop a long running nuxt command which is triggered by a cronjob you can rely on locking to prevent running that command more then once.

To make sure that the locking doesnt cause problems on exotic filesystems (CI's?), I have added a common `--no-lock` cli option to disable the locking.

## Checklist:

Did not write tests yet, will do after initial approval


packages/cli/src/utils.js
```diff
@@ -18,6 +18,16 @@ const _require = esm(module, {
 const getRootDir = argv => path.resolve(argv._[0] || '.')
 const getNuxtConfigFile = argv => path.resolve(getRootDir(argv), argv['config-file'])
 
+export function getLockPath(dir) {
+  return path.resolve(dir || process.cwd())
```
pi0: Current nuxt repositories are not configured to gitignore such lock file. So we should either use `node_modules` (rootDir/node_modules) or `.nuxt` (buildDir). Which i suggest the second one as there may be multi projects with the same rootDir.

packages/cli/src/commands/dev.js
```diff
@@ -9,26 +9,36 @@ export default {
...
 
+    const config = await getConfig()
+
+    if (argv.lock) {
+      await this.lock(config.srcDir || config.rootDir, { autoUnlock: false })
+    }
+
     // Start dev
-    async function startDev(oldInstance) {
+    const startDev = async (config, oldInstance) => {
```
pi0: Why changed to arrow function? named functions help to debug easier

packages/cli/src/command.js
```diff
@@ -61,8 +69,54 @@ export default class NuxtCommand {
...
+      })
+      .then(() => {
+        if (this.forceExit) {
+          const exitTimeout = setTimeout(() => {
+            let msg = `The command 'nuxt ${this.name}' finished but Nuxt did not exit after ${forceExitAfterSeconds}s\n`
+            msg += 'This is most likely not caused by a bug in Nuxt\n'
+            msg += 'Make sure to wait for all timers you set and stop all listeners, also check any plugin, module, etc you import\n'
+            msg += 'If you are developping a custom Nuxt command, call this.disableForceExit() in your run method to prevent the forced exit\n'
+            msg += 'Force exiting'
+            foldLines(msg, maxCharsPerLine).split('\n').forEach(line => consola.warn(line))
```
pi0: :D Nice trick

packages/cli/src/commands/dev.js
pimlie: When using an arrow function we can use `this.` to access `NuxtCommand` instead of having to use the first parameter of the run method. At first I wanted to make `startDev` a class method as well, ~~but there is no mechanism~~ wait while writing this I realised because of the from method you implemented we can just have `dev` export a NuxtCommand instance and make startDev a class method. What do you think about that?

packages/cli/src/utils.js
pimlie: My original plan was to use `buildDir` indeed, but I didnt want to wait with locking until nuxt was initialised. And `buildDir` / `.nuxt` could not exist yet and didnt want to go through the trouble of having to mkdirp that folder.
In a normal project the lock dir is on the same level as the project dir with the current implementation, so if you have `/var/www/projects/my-huxt-project/` which is a git repo then the lock folder will be `/var/www/projects/my-huxt-project.lock/` (which is outside the git repo).

packages/cli/src/commands/dev.js
pi0: I see your reasons. I think keeping it here would be better to keep NuxtCommand as simple as possible.

packages/cli/src/utils.js
pi0: @pimlie That would be something worst (creating something outside of project). What if we move lock creating logic intro `core/Nuxt` class? This way programmatic usage would benefit from this feature and also we have a chance to easily use `this.options` to create buildDir.

pimlie: Hmm, moving to `core/Nuxt` sounds a bit like you suggest you could only ever have one lock within your nuxt project. I'd like to keep the idea open for now that you could use multiple locks.

We could move locking methods to `common`, add a `options.build.lock` property and implement it by default for the builder on `buildDir`?

pimlie: Another solution would be to (optionally) factor the `getNuxtConfig` out of the `core/Nuxt` constructor. Especially now since the config is a separate package anyway, we could implement similar logic as with NuxtCommand, so a `class NuxtConfig` with a static `NuxtConfig.from(options)`. That way we could check in core/Nuxt whether defaults have been applied or not and in cli you dont have to instantiate core/Nuxt just to find the buildDir.

pimlie:
@pi0 You probably wont like this implementation either, but I wanted to show it before we move the lock to core or common. If we move it, I think we should add `build.lock` and `generate.lock` properties so you can differentiate between those.

pimlie:
I will close this so you dont have to, feel free to take any idea/implementation of this pr you like :+1:

PR #4243, opened 2018-10-31
feat(cli): use cjs dist stub in package

pimlie:
## Types of changes

## Description

As discussed on discord, the cli package could use a cjs stub so we dont have to use `-r esm` all the time. The other packages dont need this because they dont provide a point of entry
and are thus not directly invoked during development. They are either called through the cli or through jest when testing (and jest didnt seem to like all the esm stubs).

## Checklist:

### DIFF
--- a/scripts/dev
+++ b/scripts/dev
@@ -5,6 +5,26 @@ import consola from 'consola'
 
 import Package from './package.js'
 
+const useCjs = new Set([
+  '@nuxt/cli'
+])
+
+const stub = {
+  es: `export * from '../src/index'`,
+  cjs: `const esm = require('esm')
+
+const _require = esm(module, {
+  cache: false,
+  cjs: {
+    cache: true,
+    vars: true,
+    namedExports: true
+  }
+})
+
+module.exports = _require('../src/index')
+`}
+
 async function main() {
   // Read package at current directory
   const rootPackage = new Package()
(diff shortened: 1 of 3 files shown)


scripts/dev
```diff
@@ -5,6 +5,26 @@ import consola from 'consola'
 
 import Package from './package.js'
 
+const useCjs = new Set([
+  '@nuxt/cli'
+])
+
+const stub = {
+  es: `export * from '../src/index'`,
```
pi0: Why don't just add a shebang to the first line all files:
```
#! /usr/bin/env node -r esm
```

test/unit/cli.test.js
```diff
@@ -9,7 +9,7 @@ const rootDir = resolve(__dirname, '..', 'fixtures/cli')
 const url = route => 'http://localhost:' + port + route
 
 const nuxtBin = resolve(__dirname, '../../packages/cli/bin/nuxt.js')
-const spawnNuxt = (command, opts) => spawn('node', ['-r', 'esm', nuxtBin, command, rootDir], opts)
+const spawnNuxt = (command, opts) => spawn('node', [nuxtBin, command, rootDir], opts)
```
pi0: We can directly spawn `nuxtBin` when having shebang :)

pi0: PS: I think there was another related unit test too

scripts/dev
pimlie: Does that work for you on a mac? When I do that on linux it says:
```
/usr/bin/env: ‘node -r esm’: No such file or directory
```

test/unit/cli.test.js
pimlie: Thanks, will change that. Better indeed to keep testing as close to real world usage as much as possible :)

scripts/dev
pi0: Don't want to add more complexity

pi0 (approved):

pimlie:
@pi0 The failing test on azure is again a timing issue in the component tests I added for `nuxt-loading.vue`. Maybe I should redo those tests to not rely on waitFor, but would appreciate some suggestions how to do that differently ;)

Maybe we should skip those component tests for now on Azure?

PR #4248, opened 2018-11-01
feat(cli): add help command

galvez:
Following convention used in `vue` and other tools like `kubectl` (Kubernetes).

```
nuxt help <command>
```

@pi0 still need to add tests, but welcome early feedback as usual :)


packages/cli/src/commands/help.js
```diff
@@ -0,0 +1,24 @@
+import consola from 'consola'
+import * as commands from '../commands'
+import NuxtCommand from '../command'
+
+export default {
+  name: 'dev',
```
pimlie: Help?! :sos:

packages/cli/src/command.js
```diff
@@ -9,7 +9,7 @@ export default class NuxtCommand {
     this.name = name || ''
     this.description = description || ''
     this.usage = usage || ''
-    this.options = Object.assign({}, options)
+    this.options = Object.assign({}, options || {})
```
pi0: According to the ECMA specs, we don't need this :D (null and undefined are allowed and will be ignored)

image

galvez: @pi0 coolio, will update soon :)

packages/cli/src/commands/help.js
```diff
@@ -0,0 +1,24 @@
...
+  description: 'Shows help for <command>',
+  usage: 'help <command>',
+  async run(cmd) {
+    const argv = cmd.getArgv()._
+    const name = argv[0] || null
+    const _commands = { ...commands }
+    if (name in _commands) {
```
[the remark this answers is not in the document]
galvez: @pi0 yeah, I was hoping this would be another stepping stone :) For now it's good enough for core commands.

packages/cli/src/run.js
```diff
@@ -33,14 +33,7 @@ async function listCommands(_commands) {
...
   const defaultCommand = 'dev'
-
-  const cmds = new Set([
-    defaultCommand,
-    'build',
-    'start',
-    'generate'
-  ])
-
+  const cmds = new Set(Object.keys(commands))
```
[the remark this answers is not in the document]
galvez: @pi0 good call.

packages/cli/src/commands/help.js
galvez: @pimlie good catch!

pi0 (approved):

PR #4251, opened 2018-11-01
feat(cli): add cli --edge option

pimlie:
## Types of changes

## Description

This PR adds a common `--edge` cli option so you can easily test your project with nuxt-edge packages. If the nuxt-edge packages are not yet installed, there is an interactive install included for the edge packages.

There are no tests included yet as this feature might trigger some discussion first ;)

## Checklist:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -16,6 +16,7 @@
     "consola": "^2.0.0-1",
     "esm": "^3.0.84",
     "minimist": "^1.2.0",
+    "prompts": "^1.2.0",
     "wrap-ansi": "^4.0.0"
   },
   "publishConfig": {
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -55,6 +55,8 @@ export default class NuxtCommand {
       this.showHelp()
     }
 
+    this.useEdge = argv.edge
+
     return argv
   }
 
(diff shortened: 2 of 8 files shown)


pimlie:
@galvez Thanks for your thoughts, I agree that hopefully the use-case is small for this option. But to me its also about ease-of-use and DX experience in general. Hopefully it happens less and less, but in the past there have been serious bugs in Nuxt which were solved in edge but due to a variety of reasons not yet released. With this pr we could just say, run with `--edge` to check if bug X is removed or until a new version has been released.

galvez:
@pimlie my main concern is with adding complexities that are highly specific to the main loading mechanism. Perhaps we could add an `edge` _command_ instead, that encapsulates all the wrangling needed separately? What do you think @pi0 @Atinux?

pimlie:
Other subjects I imagined we could talk about are:
- should we abstract cli interaction?
- installing packages:
  - this could be abstracted more, but I wanted to take it one step at a time. (e.g. in the past it was mentioned that nuxt/cli should be aware of modules which support cli commands and install those modules automatically if needed so nuxt/cli should probably have a method of installing packages)
  - should installing packages really be interactive or should we just do it? (interactive could be an issue with automation when nuxt is indefinitely waiting for user input)
    - if so, should we provide a uniform interface for that?
    - if not, should we maybe just log the yarn command that people need to copy / paste themselves and exit if not all needed packages are installed?
  - is it ok to just spawn yarn/npm or do you prefer to use e.g. npm's api (yarn doesnt seem to have such an api and thought therefore better to keep it similar as we have a preference for yarn anyway)
  - should we support npm's --no-save? (yarn doesnt have a similar option, dont want to link it here but see their issue 1743)

galvez:
@pimlie we decided to archive this for now, as the preferred method for switching Nuxt will take place at the distribution level. We're however trying to take some goodies from this PR into the new CLI commands architecture. We'll address your other comments after work resumes in https://github.com/nuxt/nuxt.js/pull/4215.

galvez:
Thanks again for your effort with this!

PR #4259, opened 2018-11-02
feat: improve DX

pi0:
## Types of changes

## Description

- Remove showReady from Core/Server
- Added a fancy banner inside CLI
- Listener refactored out of the Server
- Disable clearConsole
- Migrate to WebpackBar 3
-  Pause logs while building
- Wrap all console logs into consola

Fancy badge:

image

Extend support: (Via options.cli.badgeMessages)

image

Error stacks:

image

Colorized commands:

image

image

Fancier fatal error handler:

image

More informative messages:

image

No more clear screens. Each dev-reload has its own log:

image

## Checklist:


packages/server/src/listener.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    // Call server.listen
+    this.server = await new Promise((resolve, reject) => {
+      const s = this._server.listen(listenArgs, error => error ? reject(error) : resolve(s))
+    })
+
+    // Enable destroy support
+    enableDestroy(this.server)
+    pify(this.server.destroy)
```
clarkdo: `server-destroy` is a wrapper a `server.close` but not supports `async`, I think this part cannot work well, if you want I can create async version `server-destroy`

pi0: It should pass `cb` (which is generated by pify for making promise) directly into `close` call so works :)

clarkdo: Right, ignore me

clarkdo: Since lots of codes in nuxt are using `promisify`, are we going to replace ?

Issue #3994, opened 2018-09-27
Long hot reload due to SASS processing


LordGuard:
### Version

[v2.0.0](https://github.com/nuxt.js/releases/tag/v2.0.0)

### Reproduction link

[https://nuxtjs.org](https://nuxtjs.org)

### Steps to reproduce

Upgrade a large project to Nuxt 2.0.0.

### What is expected ?

Fast hot reload

### What is actually happening?

After updating the project to nuxt 2.0 hot reload started to work very slowly. At any change, the sass-loader processes styles in all .vue files. It takes more than 30 seconds.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7850">#c7850</a>)</em></sub></div>

AndrewBogdanovTSS:
Yeah, I've started to get same issues recently, especially on relatively big vue files (+100 lines of code). It sometimes takes more than a minute to recompile and sometimes the process just shut down with a memory overflow error

LordGuard:
Still up to date. Sass is re-processed in all files even if the change did not affect the styles at all.

AndrewBogdanovTSS:
Is it possible to do a partial recompile of only those parts that were changed?

XavierLeTohic:
Running `nuxt dev` it takes 122s before the app become available. 
Node-sass is extremely slow, we are customizing element-ui sass variables and using a theme.

manniL:
Related: https://github.com/webpack-contrib/sass-loader/issues/296

manniL:
After digging through a lot of material I don't see much we can do. `sass-loader` does not cache imports/partials and will recompile everything. You could try https://github.com/yibn2008/fast-sass-loader or enabling the experimental options `parallel` and `cache` (https://nuxtjs.org/api/configuration-build#parallel).

Also things like `sourcemap` compilation slows down the build as well.

LordGuard:
> After digging through a lot of material I don't see much we can do. `sass-loader` does not cache imports/partials and will recompile everything. You could try https://github.com/yibn2008/fast-sass-loader or enabling the experimental options `parallel` and `cache` (https://nuxtjs.org/api/configuration-build#parallel).
> 
> Also things like `sourcemap` compilation slows down the build as well.

But why did it work much faster before nuxt 2.0?

manniL:
[...]

PR #4302, opened 2018-11-09
chore: warn when styleResources is used without the community module

TheAlexLichter:
related: #3994 

As `style-resources-loader` causes huge performance problems and we cannot easily replace or remove it as it'd be a breaking change, we should put out a warning and a reference to the drop-in replacement (https://github.com/nuxt-community/style-resources-module).

The module will set `build.styleResources` to an empty object if used, so this warning will only pop up if you are not using the module.

## Types of changes
- [x] Warning

### DIFF
--- a/packages/webpack/src/builder.js
+++ b/packages/webpack/src/builder.js
@@ -77,8 +77,10 @@ export class WebpackBundler {
     // Check styleResource existence
     const styleResources = this.context.options.build.styleResources
     if (styleResources && Object.keys(styleResources).length) {
-      consola.warn('Using styleResources without the nuxt-style-resources-module is not suggested and can lead to severe performance issues')
-      consola.warn('Please use https://github.com/nuxt-community/style-resources-module')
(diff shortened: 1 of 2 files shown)


packages/webpack/src/builder.js
```diff
@@ -76,6 +76,10 @@ export class WebpackBundler {
 
     // Check styleResource existence
     const styleResources = this.context.options.build.styleResources
+    if (styleResources && Object.keys(styleResources).length) {
+      consola.warn('Using styleResources without the nuxt-style-resources-module is not suggested and can lead to severe performance issues')
```
pi0: ٔNot sure. But maybe combining two warns into a single one looks better :)

pi0 (dismissed):

packages/webpack/src/builder.js
manniL: Agreed!

pi0 (approved):

PR #4303, opened 2018-11-09
chore: warn on unknown mode

TheAlexLichter:
Warn if mode is neither `spa` nor `universal`.

## Types of changes
- [x] Warning

### DIFF
--- a/packages/cli/src/utils/index.js
+++ b/packages/cli/src/utils/index.js
@@ -56,10 +56,6 @@ export async function loadNuxtConfig(argv) {
   options.mode =
     (argv.spa && 'spa') || (argv.universal && 'universal') || options.mode
 
-  if (options.mode && !['spa', 'universal'].includes(options.mode)) {
-    consola.warn(`Unknown mode: ${options.mode}`)
-  }
-
   // Server options
   options.server = defaultsDeep({
     port: argv.port || undefined,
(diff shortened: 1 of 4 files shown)


packages/cli/src/utils/index.js
```diff
@@ -56,6 +56,10 @@ export async function loadNuxtConfig(argv) {
   options.mode =
     (argv.spa && 'spa') || (argv.universal && 'universal') || options.mode
 
+  if (options.mode && !['spa', 'universal'].includes(options.mode)) {
```
pi0: The way that `mode` is designed is like a defined preset. We can add more modes using `options.modes`.  So i suggest moving this check after resolving `modePreset`: https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L240

manniL: I felt that I missed something like that. Thanks for the heads up! 🙈

aldarund (approved):
nice idea. recently in discord #russian channel someone asked that he used mode: 'static' and was wondering why it dont generate static html files :)

pi0 (approved):

AndrewBogdanovTSS:
@aldarund I was also caught in this pitfall when I was trying out new `modern` mode. For some reason I decided that it should be specified like `mode: 'modern'` and since I didn't get any warnings/nor errors during the build - I was convinced that it's the right way to go, although nothing been working in the browser. It just was an endless wait for the response

Issue #4296, opened 2018-11-08
When globalName is set to something other than "nuxt", there is an Uncaught TypeError


sam3d:
### Version

[v2.2.0](https://github.com/nuxt.js/releases/tag/v2.2.0)

### Reproduction link

[https://codesandbox.io/s/430mvo2zo0](https://codesandbox.io/s/430mvo2zo0)

### Steps to reproduce

1. Create the most basic `nuxt.js` app that can be created (with a single `pages/index.vue`)
2. In the `nuxt.config.js` set the following:

```javascript
export default {
    globalName: "root"
}
```

3. The web console will now throw the following error:

```
VM63:1 Uncaught TypeError: Cannot read property 'constructor' of undefined
    at setTimeout (<anonymous>:1:123)
```

### What is expected ?

No error is thrown when a different `globalName` is set

### What is actually happening?

The TypeError (shown above) is thrown

### Additional comments?

This happens on any `globalName` that isn't `nuxt`

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8110">#c8110</a>)</em></sub></div>

sam3d:
Ah, okay. This appears to happen on interaction with Vue DevTools, otherwise it doesn't occur. It likely isn't an error with Nuxt.

galvez:
Thanks guys, I'll take a look at this tonight.

galvez:
Fixed via #4305

PR #4305, opened 2018-11-09
fix: dev tools detection of custom global name

galvez:
See #4296 -- apparently we need to maintain the reference to the Vue object.

### DIFF
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -60,7 +60,7 @@ export default {
     if (typeof window !== 'undefined') {
       window.<%= globals.nuxt %> = this
       <% if (globals.nuxt !== '$nuxt') { %>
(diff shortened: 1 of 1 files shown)


pi0:
@galvez Here is the check. So for more security can't we _emulate_ such object? 

image

packages/vue-app/template/App.js
```diff
@@ -60,7 +60,7 @@ export default {
     if (typeof window !== 'undefined') {
       window.<%= globals.nuxt %> = this
       <% if (globals.nuxt !== '$nuxt') { %>
-      window.$nuxt = true
+      window.$nuxt = { $root: { constructor: this.$root.constructor } }
```
galvez: Something like this @pi0?

pi0 (approved):

PR #4313, opened 2018-11-11
feat: add option to push minification at the end

TheAlexLichter:
Currently, we have no real way to control the order of the loaded postCSS plugins while saving the defaults (as they are provided as `object` to get merged with the defaults). Until Nuxt 3.0 we can't change the array behavior as well because we can't introduce breaking changes.

This leads to problems where CSS minification won't succeed as plugins are used after `cssnano` is executed.

The option `pushCssNano` will move the `cssnano` plugin to the end of the plugin array, so it'll be executed **last**

## Types of changes

## Checklist:

### DIFF
--- a/packages/webpack/src/config/utils/postcss.js
+++ b/packages/webpack/src/config/utils/postcss.js
@@ -129,6 +129,11 @@ export default class PostcssConfig {
         config = merge({}, this.defaultConfig, config)
         this.loadPlugins(config)
       }
+      if (config.options && config.options.pushCssNano) {
+        const cssnanoIndex = config.plugins.findIndex(p => p.postcssPlugin.toString() === 'cssnano')
+        const cssnanoPlugin = config.plugins.splice(cssnanoIndex, 1)
+        config.plugins.push(cssnanoPlugin[0])
+      }
       return config
     }
   }


packages/webpack/src/config/utils/postcss.js
```diff
@@ -129,6 +129,11 @@ export default class PostcssConfig {
         config = merge({}, this.defaultConfig, config)
         this.loadPlugins(config)
       }
+      if (config.options && config.options.pushCssNano) {
```
pi0: This option would be better to move under `build` with default value inside `config` package :)

manniL: Where would you put t exactly? I thought using an options object inside postcss is a good idea (`build.postcss.options`) 🤔

pi0: oh my bad! this `config.options` is already scoped 👍

pi0 (approved):

manniL:
Adding a default to the config src wouldn’t be a bad idea anyway, for clarity/visibility and to omit the existence check

manniL:
Closing this in favor of #4318

PR #4314, opened 2018-11-11
feat(cli): improvements and external commands

galvez:
An attempt at implementing the _universal approach_ from https://github.com/nuxt/rfcs/issues/11

(edited by @pi0)

## Main objectives

- Alias `nuxt-custom ...` commands with `nuxt custom ...` shortcut. Allowing a universal way for pluggable nuxt module commands. 
- Expose `NuxtCommand` class (`@nuxt/cli`) for command authors

## Other changes

- Internal refactors and test improvements
- Prevent `process.exit` as much as possible (Thanks to @pimlie tip)
- Refactors on `NuxtCommand` for making it ready for external commands
	- This class could be finally moved into module-utils package
- Refactors on built-in commands
- `-v` alias for version


packages/cli/src/list.js
```diff
@@ -1,13 +1,16 @@
 import chalk from 'chalk'
 import NuxtCommand from './command'
+import { getLocalCommands, loadLocalCommand } from './local'
```
pi0: An small naming suggestion, we can call these commands `builtIn` or `core` :)

packages/cli/src/local.js
```diff
@@ -0,0 +1,30 @@
+import { resolve, join, parse } from 'path'
+import { readdirSync, existsSync } from 'fs'
+import { requireModule } from './utils'
+import NuxtCommand from './command'
+
+const filterCommands = (dir) => {
+  return readdirSync(dir).filter(c => c.endsWith('.js'))
+}
+
+export function getLocalCommands() {
```
pi0: These utilities seem are going to be a standard way for a _group_ of commands. What do you think of introducing `CommandGroup` interface class with public methods including `exists(cmd)` / ``get(cmd)`? (Useful for modules that expose more than one command). PS: We need to somehow add a namespace support so that 3rd party modules are restricted to only use their own. (Except builtIn ones)

test/fixtures/with-commands/commands/my-custom-command.js
```diff
@@ -0,0 +1,23 @@
+import consola from 'consola'
+
+export default {
+  name: 'my-custom-command',
+  description: 'My Custom Command',
+  usage: 'my-custom-command <foobar>',
```
pi0: Double emphasize that we should consider command namespacing with command object schema :)

packages/cli/src/list.js
galvez: @pi0 not really, `NuxtCommand` represents the built-in ones, `local` has commands local _to the project_ only. In my othe branch, I've evolved this into NuxtCommand subclasses, `LocalNuxtCommand` (for locally defined commands) and `ExternalNuxtCommand` (for commands defined inside NPM packages).

packages/cli/src/local.js
galvez: Hmm, take a look at:

https://github.com/nuxt/nuxt.js/tree/feat/module-commands/packages/cli/src/command
https://github.com/nuxt/nuxt.js/blob/feat/module-commands/packages/cli/src/run.js

WDYT? That's from my other branch (evolved from this one).

If you're OK with this PR, I can address these changes in that next one :)

test/fixtures/with-commands/commands/my-custom-command.js
galvez: @pi0 gotcha, will make this change.

packages/cli/src/run.js
```diff
@@ -1,31 +1,23 @@
...
-    process.argv.splice(2, 1)
-  } else {
+  const cmd = process.argv[2] || 'dev'
+  try {
+    NuxtCommand.ensure(cmd)
+  } catch (notFoundError) {
     if (process.argv.includes('--help') || process.argv.includes('-h')) {
-      listCommands().then(() => process.exit(0))
-      return
+      return listCommands().then(process.exit)
```
galvez: @pi0 this is because process.exit() runs with status 0 by default :)

packages/cli/bin/nuxt-cli.js
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../dist/cli.js').run()
+require('../dist/cli.js').run().then()
```
pi0: Why this is required?

packages/cli/src/run.js
pi0: I think this should be `then(() => process.exit(0)).catch(() => process.exit(1)` (Currently it is evaluated in parallel to listCommands)

pi0:
Nice idea to use [command-exists](https://www.npmjs.com/package/command-exists) @galvez

packages/cli/src/run.js
galvez: @pi0 A reference is passed to then(), there's no immediate call there :) The difference is in providing an existing function and passing a wrapper to an existing function.

galvez: Good call about adding `catch()` though!

pi0: Oops, you are right. But I suggest for code clarity avoid direct passing reference.

packages/cli/bin/nuxt-cli.js
galvez: @pi0 oh, I assumed it would be -- but you're right, Node won't exit until the last promise has finished running.

pi0: Actually, it would be better to ensure process exit here too. Like nuxt/opencollective#4

galvez: @pi0 indeed -- this has happened to me before, totally forgot about it.

packages/cli/src/command.js
```diff
@@ -29,6 +28,11 @@ export default class NuxtCommand {
   }
 
   run() {
+    if (this.argv.help || this.argv.h) {
+      this.showHelp()
+    } else if (this.argv.version || this.argv.v) {
+      this.showVersion()
+    }
     return this.cmd.run(this)
```
pimlie: @galvez @pi0 I noticed in the preliminary documentation that the example there also used a `process.exit(0)`. But the node [docs](https://nodejs.org/dist/latest-v11.x/docs/api/process.html#process_process_exit_code) advise against using process.exit() if possible. Is it an idea to use the [unreffed timeout trick](https://github.com/nuxt/nuxt.js/pull/4239/files#diff-f24635a18a996e327cd2c5124f2e440dR104) to detect whether the run command exited nicely? That way we adhere more to the official Node.js docs/advise instead of promoting the use of forced exits even in non-error conditions.

galvez: I like that @pimlie -- worth looking into @pi0 :)

pi0: @pimlie Yes, I agree. Honestly, I was surprised too when saw `process.exit` codes inside `NuxtCommand` 😆 We can limit process exiting logic to `cli/run` command only or use better techniques there like the one you mentioned.

galvez: Sorry @pi0 -- I had no idea :) Thanks for clearing it up.

pi0: Now, we only have process.exit in `.catch` case of cli binary :)

galvez (commented):
Thanks @pi0 -- love the changes, LGreatTM :)

PR #4321, opened 2018-11-12
fix: handle errors that are not Error instances

aldarund:
## Types of changes

## Description
There several issues currently with error handling.
If string error thrown in fetch - it will crash inside nuxt sources both on ssr and sp
If object error thrown in fetch - it will be displayed as [Object object] on spa and will be completely hidden on ssr
And during dev errors is printed twice with debug
On nuxt-error error object that is passed expected to be Object type, not Error, so passing any error instance there will throw error in console

Yes, its not good to throw not Errors but it can happen in client code or in some library code, so we should properly handle such cases.

## Checklist:


packages/vue-app/template/utils.js
```diff
@@ -273,6 +273,13 @@ export function getQueryDiff(toQuery, fromQuery) {
   return diff
 }
 
+export function normalizeError(err) {
+  return {
+    message: err.message || (typeof err === 'string' ? err : JSON.stringify(err)),
```
pi0: `JSON.stringify` may cause circular serialization problems. We are also in the browser context here and can't use big deps like `devalue`. Currently out of idea just mentioned that problem :D

pi0 (dismissed):
Except mentioned stringify problem, LGTM

packages/vue-app/template/utils.js
aldarund: @pi0 but we use devalue in browser context :) But its not much usable for this usecase because it can produce not too readable js. I also though about https://bundlephobia.com/result?p=stringify-object@3.3.0 but as you said its still another dependency...

pi0: @aldarund Sorry I was meaning that we don't bundle it with client dists. (Only used in template strings)

aldarund: @pi0 i added catch for serialization problems and it that case just display object name. That should be pretty rare case

packages/vue-app/template/utils.js
```diff
@@ -273,6 +273,23 @@ export function getQueryDiff(toQuery, fromQuery) {
   return diff
 }
 
+export function normalizeError(err) {
+  let message
+  if (!(err.message || typeof err === 'string')) {
+    try {
+      message = JSON.stringify(err, null, 2)
+    } catch (e) {
+      message = `Non stringified object ${err.constructor.name} `
```
pi0: @aldarund Combining this type of message with an actual error may cause more confusions for developers (DX) They start thinking, hmm so what's that _non stringified_ thing! What if we just say `[ErrorName]`?

pi0 (approved):

packages/vue-app/template/client.js
```diff
@@ -150,9 +145,7 @@ async function loadAsyncComponents(to, from, next) {
     // Call next()
     next()
   } catch (err) {
-    const error = err || {}
-    const statusCode = (error.statusCode || error.status || (error.response && error.response.status) || 500)
-    this.error({ statusCode, message: error.message })
+    this.error(err)
     this.<%= globals.nuxt %>.$emit('routeChanged', to, from, error)
```
jorisbontje: This change causes:
```
ReferenceError: Can't find variable: error
```

manniL: Could you create a new issue with a reproduction repo for that?

jorisbontje: If line 153 used to define `error`, which is then removed in this PR, but line 156/149 still references to this `error` variable, it can only go wrong.

I'll try to reproduce and create a new issue with minimal reproduction repo.

manniL: Oh, I see, you are right @jorisbontje 🙈 
Would you submit a PR?

aldarund: @manniL @jorisbontje https://github.com/nuxt/nuxt.js/pull/4442

manniL: Nvm, already there ☺️ 
https://github.com/nuxt/nuxt.js/pull/4442

Thanks for the info!

PR #4322, opened 2018-11-12
refactor: resolve styleResources when not null

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/webpack/src/builder.js
+++ b/packages/webpack/src/builder.js
@@ -81,16 +81,17 @@ export class WebpackBundler {
         'Using styleResources without the nuxt-style-resources-module is not suggested and can lead to severe performance issues.',
         'Please use https://github.com/nuxt-community/style-resources-module'
       )
-    }
-    Object.keys(styleResources).forEach(async (ext) => {
-      await Promise.all(wrapArray(styleResources[ext]).map(async (p) => {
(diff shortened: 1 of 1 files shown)


manniL:
Not sure if I get this PR right :thinking: 

The warning will be emitted and `styleResources` will be handled anyway (to avoid a breaking change).
Your PR changes that behavior (if `styleResources` are empty) :thinking:

TheAlexLichter (changes requested):
See comment

clarkdo:
If styleResources is undefined or empty，the loop is unnecessary since it just runs an empty look, I think this way is easier to understand and reduce the empty loop

clarkdo:
Got you

manniL:
That should, in theory, never happen. `styleResources` is `{}` by default. The user must have set it to `undefined` or `null` to reproduce this error.

TheAlexLichter (changes requested):
You might want to move the actual loop in the `if` statement and add the or in line 79 instead https://github.com/nuxt/nuxt.js/blob/f60cb39c968ed622fdcaa64fb602fe24df479ad2/packages/webpack/src/builder.js#L79.

clarkdo:
Make sense, close it then

PR #4331, opened 2018-11-14
fix(core): stable route sorting

### LINKED ISSUE #4188: Node v11 breaks route sorting
### Version

[v2.2.0](https://github.com/nuxt.js/releases/tag/v2.2.0)

### Reproduction link

[https://github.com/nuxt/nuxt.js](https://github.com/nuxt/nuxt.js)

### Steps to reproduce

Install node v11 and run `yarn test`

### What is expected ?

Tests finish without errors

Unit tests which check exact position of routes in route array fail

### Additional comments?

Node v11 uses v8 7.0 which uses a stable TimSort algorithm instead of an unstable QuickSort.

Relevant links:
- [Node v11 release page](https://github.com/nodejs/node/releases/tag/v11.0.0)
- [Node v11 medium article](https://medium.com/@nodejs/october-brings-node-js-10-x-to-lts-and-node-js-11-to-current-ae19f8f12b51)
- [v8 v7.0 release page](https://v8.dev/blog/v8-release-70)
- [v8 blog array_sort](https://v8.dev/blog/array-sort)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8029">#c8029</a>)</em></sub></div>

pimlie:
## Types of changes

## Description

This PR fixes route sorting in Node v11 (fix #4188).

The problem was caused because the route sorting in Nuxt was inherently unstable due to default ordering to be dictated by the order in which elements where compared. This caused issues in v11 because quicksort compares `arr[0]` with `arr[1]` but timsort does `arr[0]` with `arr[arr.length - 1]`. 
E.g. if you look at the `dynamic-routes` test then with quicksort it compared `parent` to `test` but in timsort it compared `test` to `parent`. As the default behaviour of route sorting was to return `1`, this meant that if there was no 'special' case `a` was always sorted after `b` regardless of its actual value (hence the inherently unstable). This PR changes this behaviour to alphabetically sort a and b (with `localeCompare`) if there is no 'special' case.

I have also moved route sorting outside the forEach loop, there is no need to sort the routes array on every intermediate step and it probably saves a little bit of time to only sort it once.

## Checklist:

### DIFF
--- a/packages/common/src/utils.js
+++ b/packages/common/src/utils.js
@@ -256,6 +256,56 @@ function cleanChildrenRoutes(routes, isChild = false) {
   return routes
 }
 
+const DYNAMIC_ROUTE_REGEX = /^\/(:|\*)/
+
+const sortRoutes = function sortRoutes(routes) {
+  routes.sort((a, b) => {
+    if (!a.path.length) {
+      return -1
+    }
+    if (!b.path.length) {
+      return 1
+    }
+    // Order: /static, /index, /:dynamic
+    // Match exact route before index: /login before /index/_slug
+    if (a.path === '/') {
+      return DYNAMIC_ROUTE_REGEX.test(b.path) ? -1 : 1
+    }
+    if (b.path === '/') {
+      return DYNAMIC_ROUTE_REGEX.test(a.path) ? 1 : -1
+    }
+
+    let i
+    let res = 0
+    let y = 0
+    let z = 0
+    const _a = a.path.split('/')
+    const _b = b.path.split('/')
+    for (i = 0; i < _a.length; i++) {
+      if (res !== 0) {
+        break
+      }
+      y = _a[i] === '*' ? 2 : _a[i].includes(':') ? 1 : 0
+      z = _b[i] === '*' ? 2 : _b[i].includes(':') ? 1 : 0
+      res = y - z
+      // If a.length >= b.length
+      if (i === _b.length - 1 && res === 0) {
+        // sort alphabetically unless * found
+        res = _a[i] === '*' ? -1 : a.path.localeCompare(b.path)
+      }
+    }
+    return res === 0 ? (_a[i - 1] === '*' && _b[i] ? 1 : -1) : res
+  })
+
+  routes.forEach((route) => {
+    if (route.children) {
+      sortRoutes(route.children)
+    }
+  })
+
+  return routes
+}
+
 export const createRoutes = function createRoutes(files, srcDir, pagesDir) {
   const routes = []
   files.forEach((file) => {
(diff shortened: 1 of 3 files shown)


packages/common/src/utils.js
```diff
@@ -256,6 +256,56 @@ function cleanChildrenRoutes(routes, isChild = false) {
...
+      if (res !== 0) {
+        break
+      }
+      y = _a[i] === '*' ? 2 : _a[i].includes(':') ? 1 : 0
+      z = _b[i] === '*' ? 2 : _b[i].includes(':') ? 1 : 0
+      res = y - z
+      // If a.length >= b.length
+      if (i === _b.length - 1 && res === 0) {
+        // sort alphabetically unless * found
+        res = _a[i] === '*' ? -1 : a.path.localeCompare(b.path)
```
clarkdo: Should `a.path.localeCompare(b.path)` be `b.path.localeCompare(a.path)` ?

Since `'snake_case_route'.localeCompare('another_route-id')` is `1`, but we need `another_route-id`is in front of `snake_case_route` as alphabetical order.

clarkdo (dismissed):
LGTM

packages/common/src/utils.js
pimlie: The tests succeeds in v11 actually but fails in v10. It is because the last line in sortRoutes also has a fallback to sort by default with -1 regardless of value, I missed that because I tested with `dynamic-routes` and it seems that test doesnt test that code path. Currently trying to fix the dynamic-routes test to include that ;)

clarkdo: `(_a[i - 1] === '*' && _b[i] ? 1 : -1)` could be `(_a[i - 1] === '*' && _b[i] ? 1 : _a[i - 1].localeCompare(_b[i - 1]))` ?

TheAlexLichter (dismissed):

packages/common/src/utils.js
pimlie: If we use that then `/test/projects` is sorted before `/test`, if I use `a.path.localeCompare(b.path)` then `/test/songs/toto` is sorted before `/test/users`, so I think we should only compare alphabetically if _a.length === _b.length

pimlie:
Azure fails (again, intermittently) on the components test. The route sorting should be stable now between v10 and v11. Maybe we can add v11 testing to CI before merging this to make sure it does?

clarkdo:
I will trigger azure rebuild.
I think we can merge it asap and I’ll add v11 into CI shortly.

clarkdo (approved):

PR #4332, opened 2018-11-14
ci: add platforms and latest node in azure

clarkdo:


aldarund (approved):

clarkdo:
@pi0 has some other ideas about this or, let’s hold it for now 🤥

galvez (approved):

pi0:
As asked and get confirmation from @clarkdo, temporary closing this PR so we can address some concerns and make CI/Tests faster and stable:

- Prevent doing _duplicate_ platform tests when possible
- Reduce azure job queue
- Reduce fail chances
- More focus on docker based workflows and probably adding GH Actions
- Use `puppeteer-core` and docker image with pre-installed chrome for E2E tests

Issue #4334, opened 2018-11-15
Automatically resolve scoped module slash in Windows when specified in transpile block


amoshydra:
### What problem does this feature solve?

Related to: https://github.com/nuxt/docs/pull/985

---

On Windows, the filepath that received by JavaScript uses backslash `\`.

At first it might be tempting to use
```js
build: {
  transpile: [
    '@scoped/packageA',
    '@scoped/packageB',
  ],
},
```

However, this will fail on Windows.

Related code:
```js
!modulesToTranspile.some(module => module.test(file))
```
https://github.com/nuxt/nuxt.js/blob/3702dfe5dcb5c4fdeb03fcd3025f5a5ae39232be/packages/webpack/src/config/base.js#L176

Which translate to:
```js
/@scoped\/packageA/.test('C:\\projects\\node_modules\\@scoped\\packageA\\src\\index.js')
```

and evaluated to: `false`

### What does the proposed changes look like?

We could convert filepath for user before testing, so that user can continue using one forward slash for scoped package.

```js
const normalizePathForRegex = module => path.join(module).replace(/\\/, '\\\\')
```

```js
  ...
  options.build.transpile = [].concat(options.build.transpile || [])
    .map(module => module instanceof RegExp ? module : new RegExp(normalizePathForRegex(module)));
  ...
)
```

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8142">#c8142</a>)</em></sub></div>

amoshydra:
Digging throught the test file, it seems like, user are expected to write RegExp friendly string too

https://github.com/nuxt/nuxt.js/blob/dev/test/unit/basic.dev.test.js#L28-L31
```js
transpile: [
  'vue\\.test\\.js',
  /vue-test/
],
```

Perhaps, this shouldn't be the case too?

**Suggestion:**
User should be able to write normal string that represent a filepath and have their provided string sanitized for regex later.

In conjunction to the above feature request, if user provide a plain string, nuxt should
1. convert given path to os friendly file path
2. sanitize string for RegExp
3. convert the sanitized string to RegExp
4. use this RegExp for testing

amoshydra:
Resolved by #4335

PR #4335, opened 2018-11-15
feat: Process transpile path

amoshydra:
## Types of changes

## Description

This PR aimed to resolve issue #4334 by processing the plain string entry of `options.build.transpile`.

The processing includes:
1. convert given path to os friendly file path (using path.normalize)
2. sanitize string for RegExp (using lodash escapeRegExp)
3. convert the sanitized string to RegExp (original)
4. use this RegExp for testing (original)

However, doing this conflict with one of the test case that uses escaped RegExp string as entry. 
**input**: `'vue\\.test\\.js'`
**test**: `node_modules/vue.test.js`
This will no longer work (thus breaking change) because we will now be expecting user to write plain string instead.

Unfortunally this test is also depended on the OS as it uses `path.normalize` which its behaviour is depended on OS.

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -2,6 +2,7 @@ import path from 'path'
 import fs from 'fs'
 import defaultsDeep from 'lodash/defaultsDeep'
 import defaults from 'lodash/defaults'
+import escapeRegExp from 'lodash/escapeRegExp'
 import pick from 'lodash/pick'
 import isObject from 'lodash/isObject'
 import consola from 'consola'
@@ -320,7 +321,12 @@ export function getNuxtConfig(_options) {
 
   // include SFCs in node_modules
   options.build.transpile = [].concat(options.build.transpile || [])
-    .map(module => module instanceof RegExp ? module : new RegExp(module))
+    .map(module => (module instanceof RegExp)
+      ? module
+      : new RegExp(
+        escapeRegExp(path.join(module))
+      )
+    )
 
   if (options.build.quiet === true) {
     consola.level = 0
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -324,7 +324,7 @@ export function getNuxtConfig(_options) {
     .map(module => (module instanceof RegExp)
       ? module
       : new RegExp(
-        escapeRegExp(path.join(module))
+        escapeRegExp(path.normalize(module))
       )
     )
 
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -324,7 +324,11 @@ export function getNuxtConfig(_options) {
     .map(module => (module instanceof RegExp)
       ? module
       : new RegExp(
-        escapeRegExp(path.normalize(module))
+        escapeRegExp(
+          path.normalize(
+            module.replace(/\\/g, '/')
+          )
+        )
       )
     )
 
--- a/test/unit/basic.dev.test.js
+++ b/test/unit/basic.dev.test.js
@@ -62,11 +62,11 @@ describe('basic dev', () => {
 
   test('Config: build.transpile', () => {
     expect(transpile('vue-test')).toBe(true)
-    expect(transpile(path.join('node_modules/test.js'))).toBe(false)
-    expect(transpile(path.join('node_modules/vue-test'))).toBe(true)
-    expect(transpile(path.join('node_modules/vue.test.js'))).toBe(true)
-    expect(transpile(path.join('node_modules/test.vue.js'))).toBe(true)
-    expect(transpile(path.join('node_modules/@scoped/packageA/src/index.js'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/test.js'))).toBe(false)
+    expect(transpile(path.normalize('node_modules/vue-test'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/vue.test.js'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/test.vue.js'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/@scoped/packageA/src/index.js'))).toBe(true)
   })
 
   test('Config: build.filenames', () => {
--- a/test/unit/basic.dev.test.js
+++ b/test/unit/basic.dev.test.js
@@ -28,7 +28,7 @@ describe('basic dev', () => {
         },
         transpile: [
           '@scoped/packageA',
-          'vue\\.test\\.js',
+          'vue.test.js',
           /vue-test/
         ],
         loaders: {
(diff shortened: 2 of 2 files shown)


test/unit/basic.dev.test.js
```diff
@@ -60,10 +62,11 @@ describe('basic dev', () => {
...
     expect(transpile('vue-test')).toBe(true)
-    expect(transpile('node_modules/test.js')).toBe(false)
-    expect(transpile('node_modules/vue-test')).toBe(true)
-    expect(transpile('node_modules/vue.test.js')).toBe(true)
-    expect(transpile('node_modules/test.vue.js')).toBe(true)
+    expect(transpile(path.normalize('node_modules/test.js'))).toBe(false)
+    expect(transpile(path.normalize('node_modules/vue-test'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/vue.test.js'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/test.vue.js'))).toBe(true)
+    expect(transpile(path.normalize('node_modules/@scoped/packageA/src/index.js'))).toBe(true)
```
clarkdo: We should also mock windows path, like 'node_modules\\\\@scoped\\\\packageA\\\\src\\\\index.js'

packages/config/src/options.js
```diff
@@ -320,7 +321,12 @@ export function getNuxtConfig(_options) {
 
   // include SFCs in node_modules
   options.build.transpile = [].concat(options.build.transpile || [])
-    .map(module => module instanceof RegExp ? module : new RegExp(module))
+    .map(module => (module instanceof RegExp)
+      ? module
+      : new RegExp(
+        escapeRegExp(path.normalize(module))
```
clarkdo: `normalize` won't parse windows style path into POSIX, so it may still not work in windows.

test/unit/basic.dev.test.js
amoshydra: Hi @clarkdo, I am not sure how I can mock windows path here and still have the test run correctly on both POSIX and Windows due to the single direction POSIX to Windows normalization by `path.normalize`.

Also, in all case (I suspect?) babel will always pass native filepath into the exclude function automatically.

packages/config/src/options.js
amoshydra: @clarkdo,

I think you meant to say "it may still not work in POSIX"?

Here is my proposal for this issue: e85a201
>  Convert to POSIX path before normalisation
>  
>  path.normalize can convert POSIX to Windows but not Windows to POSIX.
>  This replacement is added to ensure that the path is always in POSIX
>  form before normalisation occur.

test/unit/basic.dev.test.js
amoshydra: > Also, in all case (I suspect?) babel will always pass native filepath into the exclude function automatically.

I meant to imply that: Since native filepath is used, we should be able to cover both case for 
Windows and POSIX using (ci.azure) and (ci/circleci: test-unit).

Is that correct?

clarkdo: Actually not, exclude file path will be `c:\\project\\node_modules\\@Scoped\\packageA\\src\\index.js`.

So we may also need to normalize `file` in `exclude`.

That's also why I mentioned we'd better to mock windows path, like:

```js
transpile('node_modules\\\\@scope\\\\module')
```

clarkdo: If we don't consider compatibility, we can only normalize `file` in `exclude`.

packages/config/src/options.js
clarkdo: Please see above comment 😄

test/unit/basic.dev.test.js
amoshydra: Hi @clarkdo, sorry I may not have understood on what you meant by compatibility, I am still unclear about why we need to normalize the filepath (`file`) in the `exclude` function.

## Clarifications:
1. Just to be clear, when we talk about `normalize`, this is what we mean right?
    > to convert a given file path to a new plain string with the OS's separator
2. I do understand that the filepath in the exclude function will be `"c:\\project\\node_modules\\@scoped\\packageA\\src\\index.js"` in Windows.
However, I am not sure about why we should do `transpile('node_modules\\\\@scope\\\\module')` in the test.
    1. Firstly, this will always fail the test for POSIX system, since our entry in options.build.transpile will be normalized to test for `/`.

    2. This will always fail for Windows too since we are testing `/@scope\\module/` against `"@scope\\\\module"`
    because:
    `/@scope\\module/.test("@scope\\\\module") === false`

    3. Using 4 backslash seems to imply that you want to convert the string to a RegExp? i.e. 
    `new RegExp('@scope\\\\module').source === /@scope\\module/.source`

    Therefore, the more accurate behavior will be to do `transpile(path.normalize('node_modules/@scope/module'))` instead? as this will create the same behavior as how webpack passes filepath to the `exclude` function.




## My thoughts:
https://github.com/nuxt/nuxt.js/blob/3702dfe5dcb5c4fdeb03fcd3025f5a5ae39232be/packages/webpack/src/config/base.js#L173-L176

According to my understanding, `exclude` function will receive a `filepath` as parameter. The file path will be constructed depending on the OS. On Windows it will always be separated with `\` on POSIX it will be seperated with `/`.

`modulesToTranspile` will contain RegExps using the file path we specified in the `options.build.transpile` array. These RegExps' file path will be normalize according to the OS.

Since e85a201, regardless of the format that user specify (POSIX or Windows) they will always get normalized to the OS's format. This will match well with the logic in exclude function.

## Illustration of logic:
#### Windows

| input                | moduleToTranspile    | exclude's param |
| :--                  | :--                  | :--             |
| `"@scoped/packageA"`  | `/@scoped\\packageA/` | `"c:\\project\\node_modules\\@scoped\\packageA\\src\\index.js"`
| `"@scoped\\packageA"` | `/@scoped\\packageA/` | `"c:\\project\\node_modules\\@scoped\\packageA\\src\\index.js"`

#### POSIX

| input                | moduleToTranspile    | exclude's param |
| :--                  | :--                  | :--             |
| `"@scoped/packageA"`  | `/@scoped\/package/` | `"/project/node_modules/@scoped/packageA/src/index.js"`
| `"@scoped\\packageA"` | `/@scoped\/package/` | `"/project/node_modules/@scoped/packageA/src/index.js"`

## Tested
Both of these cases are written as part of the test case too
https://github.com/nuxt/nuxt.js/blob/e85a201be01ca8dcd8e59f4fcc326cf2b501b323/test/unit/basic.dev.test.js#L29-L34

and both passed the unit test ran in Windows (in [ci.azure](https://dev.azure.com/nuxt/nuxt.js/_build/results?buildId=612&view=logs)) and POSIX environment (in [circleci](https://circleci.com/gh/nuxt/nuxt.js/15562?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link))

---

Thank for you reading through my long comment. :)

clarkdo: Thanks you the clarifications, it works fine after converting POSIX commit.

The `\\\\` is because github will escape the `\` in quote, so I add duplicate ones in whole comment, sorry.

clarkdo (approved):

PR #4338, opened 2018-11-15
fix: transpile server not work

clarkdo:
## Types of changes

## Checklist:


packages/config/src/options.js
```diff
@@ -321,16 +321,19 @@ export function getNuxtConfig(_options) {
...
-        escapeRegExp(
-          path.normalize(
-            module.replace(/\\/g, '/')
-          )
-        )
-      )
-    )
+  const include = []
+  const whitelist = []
+  for (const module of options.build.transpile) {
```
pi0: Not related to this PR, but module is a CJS reserved keyword :D

packages/config/src/options.js
```diff
@@ -321,16 +321,19 @@ export function getNuxtConfig(_options) {
...
+    if (module instanceof RegExp) {
+      include.push(module)
+      whitelist.push(module)
+    } else {
+      const posixModule = module.replace(/\\/g, '/')
+      include.push(new RegExp(escapeRegExp(path.normalize(posixModule))))
+      whitelist.push(new RegExp(escapeRegExp(posixModule)))
+    }
+  }
+  Object.assign(options.build.transpile, { include, whitelist })
```
pi0: Is the type of `build.transpile` an object? Also can't we allow users manually providing their `include`/`whitelists`? I feel changing type of an option is little confusing.

packages/config/src/options.js
aldarund: Shouldnt eslint catch this?

packages/config/src/options.js
clarkdo: include/whitelists should be invisible to users, they only config `transpile`.
We adjust them for users.

packages/config/src/options.js
clarkdo: It also happened somewhere else in project, let me change this for now.

pi0: @aldarund ESLint is not always smart! Try`import path from 'path'` and then modify `path` variable 😆

packages/config/src/options.js
pi0: @clarkdo So if it is an internal resolve mechanism, why don't do it inside `Builder` constructor? Less code execution for `nuxt start` :)

clarkdo: It could be, let me think

aldarund (approved):

PR #4365, opened 2018-11-19
fix: fall back to default value when `publicPath` is falsy

### LINKED ISSUE #4364: Validating publicPath value in Nuxt Config [v2.3.1]
### Version

[v2.3.1](https://github.com/nuxt.js/releases/tag/v2.3.1)

### Reproduction link

[https://codesandbox.io/s/pxxv9r660](https://codesandbox.io/s/pxxv9r660)

### Steps to reproduce

In the `build` section of the nuxt.config.js file, set the `publicPath` value to an empty string, or false. 

Then re-run `npm run dev`

### What is expected ?

Nuxt should default to `/_nuxt/` in the event of an invalid value.

If you set it to an empty string, the app starts up as expected, but every page returns a 404 with no client or server side errors messages. 

If you set it to `false`, you receive this error:
`url.startsWith is not a function`

### Additional comments?

In my 2.2.0 build, I had this value set to an empty string for my local environment and an S3 bucket when live. 

This worked fine for the last few months. Upon updating, every page of my app 404'd with no explanation. 

I tore down the config line-by-line and discovered that the publicPath was causing the issue. 

It's fairly clearly stated in the docs that the default value should be `/_nuxt/`, but since an empty string worked in previous versions, other people who are as foolish as I am might run into this issue.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8175">#c8175</a>)</em></sub></div>

TheAlexLichter:
## Types of changes

## Description

Resolves #4364 

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -5,7 +5,7 @@ import defaults from 'lodash/defaults'
 import pick from 'lodash/pick'
 import isObject from 'lodash/isObject'
 import consola from 'consola'
-import { isPureObject, isUrl, guardDir, isNonEmptyString } from '@nuxt/common'
+import { guardDir, isNonEmptyString, isPureObject, isUrl } from '@nuxt/common'
 import { getDefaultNuxtConfig } from './config'
 
 export function getNuxtConfig(_options) {
@@ -53,7 +53,14 @@ export function getNuxtConfig(_options) {
 
   // Apply defaults
   const nuxtConfig = getDefaultNuxtConfig()
+
   nuxtConfig.build._publicPath = nuxtConfig.build.publicPath
+
+  // Fall back to default if publicPath is falsy
+  if (options.build && !options.build.publicPath) {
+    options.build.publicPath = undefined
+  }
+
   defaultsDeep(options, nuxtConfig)
 
   // Check srcDir and generate.dir excistence
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -118,10 +118,9 @@ export function getNuxtConfig(_options) {
     options.appTemplatePath = path.resolve(options.srcDir, options.appTemplatePath)
   }
 
-  // Add trailing slash to path if not already present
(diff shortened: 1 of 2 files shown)


manniL:
Would suggest to release a patch (details in the linked issue above) after merging.

pi0 (dismissed):

pi0:
@manniL I have an alternative way to fix this: We can force add a trailing slash to the publicPath. This not only resolves this bug but also auto-fixes the situations that user forgot to add it.

manniL:
@pi0 I'd combine both solutions 👍

packages/config/src/options.js
```diff
@@ -111,6 +118,10 @@ export function getNuxtConfig(_options) {
     options.appTemplatePath = path.resolve(options.srcDir, options.appTemplatePath)
   }
 
+  // Add trailing slash to path if not already present
+  if (!/\/$/.test(options.build.publicPath)) {
```
clarkdo: Same for `_publicPath`?

clarkdo:
```suggetsion
options.build.publicPath = options.build.publicPath.replace(/([^/])$/, '$1/')
options.build._publicPath= options.build._publicPath.replace(/([^/])$/, '$1/')
```

clarkdo:
```suggestion
  options.build.publicPath = options.build.publicPath.replace(/([^/])$/, '$1/')
  options.build._publicPath= options.build._publicPath.replace(/([^/])$/, '$1/')
```

clarkdo:
```suggestion
  options.build.publicPath = options.build.publicPath.replace(/([^/])$/, '$1/')
  options.build._publicPath= options.build._publicPath.replace(/([^/])$/, '$1/')
```

manniL: Good catch!

clarkdo (approved):

PR #4378, opened 2018-11-21
fix: modern=true or false not work as expected

clarkdo:
## Types of changes

## Description

`modern` is `string` in minimist, so we need to convert `true`, `false` and '' to boolean or corresponding value.

## Checklist:

### DIFF
--- a/packages/cli/src/utils/index.js
+++ b/packages/cli/src/utils/index.js
@@ -112,19 +112,19 @@ export function showBanner(nuxt) {
 }
 
 /**
- * Normalize string option in command
+ * Normalize string argument in command
  *
  * @export
- * @param {String} option
+ * @param {String} argument
  * @param {*} defaultValue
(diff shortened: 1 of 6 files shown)


packages/cli/src/utils/index.js
```diff
@@ -110,3 +110,21 @@ export function showBanner(nuxt) {
...
+
+/**
+ * Normalize string option in command
+ *
+ * @export
+ * @param {String} option
+ * @param {*} defaultValue
+ * @returns formatted option
+ */
+export function normalizeOption(option, defaultValue) {
```
pi0: I would suggest renaming to `normalizeArg`

pi0 (dismissed):

packages/cli/src/utils/index.js
clarkdo: ok

PR #4398, opened 2018-11-24
chore(examples): update scroll-behavior example

TheAlexLichter:


examples/scroll-behavior/layouts/default2.vue
```diff
@@ -0,0 +1,3 @@
+<template>
```
pi0: @manniL It is intentional to use `default2` as the layout name?

examples/scroll-behavior/pages/another-long.vue
```diff
@@ -0,0 +1,14 @@
...
+  <div class="container">
+    <h1>Another long page</h1>
+    <div class="spacer" style="width: 100%; height: 3000px; background: darkgreen" />
+    <nuxt-link to="/">Home page</nuxt-link>
+  </div>
+</template>
+
+<script>
+export default {
+  layout: 'default2',
```
pi0: It seems yes 💃

examples/scroll-behavior/layouts/default2.vue
manniL: @pi0 it is :relaxed:

manniL: related to https://github.com/nuxt/nuxt.js/issues/2738#issuecomment-379489775

pi0 (approved):

PR #4406, opened 2018-11-25
feat: add typescript support + an example

### LINKED ISSUE #4361: feat: honor .ts/.tsx pages

This will allow us to have better DX when developing application with Nuxt.js and be able to use these file extensions.

Currently Nuxt pages only pick up .js/.vue files.

Since Nuxt.js `2.0` is already using Babel 7. I currently manage to use Typescript by just adding `@babel/preset-typescript` and a few other babel plugins.

**But** Nuxt only pick up pages with .js/.vue files.

```vue
<script lang="ts">
// this works by adding those typescript preset & plugins
<script>
```

Nuxt.js needs to allow pages with .ts/.tsx also.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8168">#c8168</a>)</em></sub></div>

chanlito:
## Types of changes

## Description

With this PR, nuxt.js users should be able to use typescript out of the box without any manual configuration.

Leveraging babel 7, this PR adds `@babel/typescript` to handle `.ts` and `.tsx` files.

Also with this PR, Nuxt.js should be able to pick up `.ts` and `.tsx` pages and layouts.

Resolves: https://github.com/nuxt/nuxt.js/issues/4361


chanlito:
Hey @manniL I've add the test fixture, could you take a look to see if it's correct? I was following other test fixtures. And BTW there's one unit test that's failing which I'm not sure what's wrong with it.

packages/webpack/src/config/base.js
```diff
@@ -189,10 +189,13 @@ export default class WebpackBaseConfig {
...
-        use: perfLoader.js().concat({
-          loader: require.resolve('babel-loader'),
-          options: this.getBabelOptions()
-        })
+        use: perfLoader.js().concat(
+          {
+            loader: require.resolve('babel-loader'),
+            options: this.getBabelOptions()
+          },
+          'vue-jsx-hot-loader'
```
manniL: Is that loader necessary? Nuxt is supporting JSX already (ref: https://github.com/nuxt/nuxt.js/blob/52a92fbc48510c7fbdaf3506345c0b686ee07b1f/examples/jsx/pages/about.js#L17)

chanlito: That one is for HMR when using jsx or tsx.

manniL:
The test fixture looks good so far! Now please add some unit tests to test that the pages are showing the correct content as well :relaxed:

chanlito:
@manniL I just added the unit test for typescript.

pi0:
@kevinmarrec As you initially worked on vue-app typings, would you please have a review?

examples/with-typescript-pages/components/CounterCard.tsx
```diff
@@ -0,0 +1,29 @@
...
+    return (
+      <div>
+        <div>Count: {this.$store.state.count}</div>
+        <button onClick={e => this.$store.commit('increment')}>+</button>
+        <button onClick={e => this.$store.commit('decrement')}>-</button>
+        <input
+          value={this.password}
+          onInput={e => (this.password = e.target.value)}
+        />
+        <nuxt-link
```
pi0: I suggest using  `<NuxtLink`

examples/with-typescript-pages/shims-tsx.d.ts
```diff
@@ -0,0 +1,13 @@
+import Vue, { VNode } from 'vue';
+
+declare global {
+  namespace JSX {
```
pi0: Shouldn't we move this typings into `vue-app` package?

examples/with-typescript-pages/components/CounterCard.tsx
chanlito: Where can I import `NuxtLink` component from? 
Typescript does type-check differently between `NuxtLink` & `nuxt-link`, you can check it here https://www.typescriptlang.org/docs/handbook/jsx.html

kevinmarrec:
@pi0 Well it seems to have a lot of JSX/TSX stuff around here, and it's using Babel.

As I'm not using Babel and JSX/TSX I'm afraid I can't really help to say if something's wrong or not.

I think @chanlito is using the second/other way to do TypeScript around a Vue project.
You can deal with Babel as @chanlito does OR uses https://github.com/TypeStrong/ts-loader.

Using one way or the other will totally change a lot of setup around the project.

I would recommend using the `ts-loader` as it's the way recommended in official Webpack docs https://webpack.js.org/guides/typescript/.

IMO this PR must be a new template named like "TypeScript Babel JSX", cause for me it's not really the standard TypeScript setup 😄

chanlito:
@kevinmarrec what do you mean by a new template? I have also used `ts-loader` pre Babel 7, now that Babel 7 is released I prefer using `@babel/preset-typescript` approach which make life easier, and cleaner TBH.

pi0:
@chanlito @kevinmarrec I see both sides. For this decision:

## `ts-loader`

- [install size](https://packagephobia.now.sh/result?p=ts-loader)
- Is more official way and probably more tested by TypeStrong
- It is more webpack friendly and potentially more hackable

## `@babel/preset-typescript` (`@babel/plugin-transform-typescript`) 

- [install size](https://packagephobia.now.sh/result?p=@babel/preset-typescript)
- Some people say it is faster (WO type checking enabled)
- We can use the same setup for Nuxt internals with rollup
- Used by `@zeit/next-typescript` after Babel7 upgrade (zeit/next-plugins#162, zeit/next-plugins#171)

manniL:
@chanlito I'd move certain parts of that PR in the Nuxt core as you proposed (ie. the support of .ts/.tsx pages). But for the parser- and build-related parts (`ts-loader`/babel-preset) I'd suggest to create an official module.

Pros:

* No affects (in terms of build time) for non-TS-users (no matter which way we choose)
* Decoupled from the core
* Easier to maintain
* 1 line to actually activate TS support

Cons:
* No "instant" support


What do you think?

kevinmarrec:
@pi0 @chanlito Alright, I'll then try the Babel approach to see if I'm facing any issues. And @chanlito By new template I was talking about the example name which I think should contain "jsx" keyword, as you can do TypeScript without jsx/tsx, it depends on people (or maybe needs). 😁

chanlito:
@kevinmarrec I see. 
@manniL your proposal sounds good to me.

I guess this PR should be closed?

MacZel:
@chanlito with this PR tsx is going to be supported in sfc?

kevinmarrec:
@pi0 @manniL @chanlito I tried the `babel-typescript` way, I wasn't able to make working Component Decorators with `vue-class-component`, and I tried many many things to get it work.

So I'll stick with `ts-loader`.

`ts-loader` can be used with `thread-loader` and `cache-loader` (both already used Nuxt.js) to have faster builds, but I wasn't able to use them for theses reasons :
1) `thread-loader` through nuxt `parallel`, cause `ts-loader` needs to be setup with `happyPackMode: true` and I end out in an issue related to https://github.com/amireh/happypack/issues/242
2) `cache-loader` through nuxt `cache` option makes one of my font unvailable (only in dev mode)

examples/with-typescript-pages/shims-tsx.d.ts
Atinux: I agree with @pi0, is it necessary to have this file in the example @chanlito?

chanlito:
@Atinux I totally agree with @manniL suggestion and a `@nuxt/typescript` official package would be nice.

kevinmarrec:
@Atinux Yeah that sounds perfect to have an official Nuxt TypeScript Module package.

kevinmarrec:
@Atinux @manniL @pi0 @chanlito 

Well, after more thoughts and some tests, I now I've a new judgement about `@babel/preset-typescript`:
1) ts-loader doesn't work well with `parallel`/`cache` Nuxt build options, meanwhile with `@babel/preset-typescript` everything work as expected.
2) I couldn't be able to make .tsx file work with ts-loader, meanwhile it just work with `@babel/preset-typescript`
3) With `babel/preset-typescript`, we are sure that everything will pass through babel and output correct syntax for old browsers, meanwhile with `ts-loader`, if you doesn't setup your `tsconfig.json` correctly, you can have issues with your final bundles !

So 3 BIG advantages for `@babel/preset-typescript`, added to the one @pi0 described : smaller package size, seems to be faster, ZEIT did the move from ts-loader to babel for Next Typescript https://github.com/zeit/next-plugins/pull/171

Then about this PR, i really think it's worth to not having TypeScript module to maintain, it will be easier to just maintain the babel presets as it's already done. And for non-TS users, this PR will not affect them on performance.

Let's make TS users happy and include TS support directly within Nuxt itself.

I plan to go even further this PR, providing support for a `nuxt.config.ts`, with autocompletion on configuration. There will be also typedefs for Middlewares & Plugins. This way it will become a really nice TS support for Nuxt.

=> Final thought : To have TypeScript working out of the box, I finally highly recommend to resolve conflicts, review and merge this PR.

pi0:
@kevinmarrec Thanks for your inputs. Yes, I think `babel/preset-typescript` would be the best way to go.

Regarding this PR, resolved conflicts but I think these concerns would be worth to think about:

- `vue-jsx-hot-loader`: [npm](https://www.npmjs.com/package/babel-plugin-vue-jsx-hot-reload) [github](https://github.com/skyrpex/vue-jsx-hot-loader) - Last NPM publish is **2 years ago** and seems no one already uses it. There are compatibility, stability, security and maintenance concerns about it.
- The JSX tags are not in PascalCase.  
- Typedefs added in the example, would be better to move into `vue-app`, if possible

My overall comment is that **JSX** is not a TS requirement. It would be better to discussed and added in a separate PR.

Regarding more TS support:

I highly agree and love this approach. But some considerations:

- We have to compare tradeoffs of using [node-ts](https://www.npmjs.com/package/ts-node) vs babel for Node.js runtime related areas including server-middleware and nuxt config. This is also important because currently, we use a similar tech for esm support.
- Regarding auto-complete for `nuxt.config.js` it worth to also consider requirements of nuxt/rfcs#16. Other than typing for auto-complete, we also need a validation system for nuxt config. These two parts have shared requirements. We may auto-generate types based on a good validation system.

kevinmarrec:
@pi0 Yeah, agreed for the JSX thoughts and we'll need further discussion about more TS support.

@chanlito Can I take the lead on this and submit a new PR regarding this TS support ? And let you discuss and think about a new different PR for the JSX/TSX support when mine will be merged ?

I also planned to add some more uses cases regarding tests using `vue-property-decorator`

kevinmarrec:
I also get a new idea :

What about a `typescript` Boolean option in Nuxt config file and only do the following things if it's enabled :

Config (@nuxt/config) :
- Add support for TS plugins and middleware

Builder (@nuxt/builder) :
- Watch & Use `.ts` files to generate routes (layouts & pages)

Webpack (@nuxt/webpack):
- Resolve `.ts` extension through Babel while passing the `typescript` option value to the Nuxt babel preset

Babel (@nuxt/babel-preset-app):
- Retrieve the `typescript` option from the preset options to use `@babel/preset-typescript` preset OR `@babel/plugin-transform-typescript` plugin (idk what's the best, the latter seems updated and bumped to last Babel version)

CLI (@nuxt/cli):
- Warning or Error if tsconfig.json not found in `options.srcDir`

MacZel:
Last week I've managed to build a working seed that supports TS in all project folders, including pages and layouts, also supporting TSX in .vue SFC. If it's relevant I can submit a PR.

chanlito:
Hey guys I think it seems that there are some confusion here.

@kevinmarrec this PR allows you `.ts` or `.tsx` if you don't want `.tsx` (which is just `jsx` with typescript) that's fine.

@pi0 surrounding the typings definition & https://github.com/skyrpex/vue-jsx-hot-loader  it's not necessary to be in nuxt.js - end-users can it themselves.

kevinmarrec:
https://github.com/nuxt/rfcs/issues/18#issuecomment-447048733

chanlito:
@kevinmarrec I don't see why we need that typescript option in nuxt.config.js

kevinmarrec:
@chanlito Cause we don't want this support to be a breaking change 😇

Without implementing as I suggested with a tyescript option it will :
- Break all Nuxt TS projects, as they are using a custom Nuxt module that already handle the TypeScript support. It will do some Babel + TsLoader random behavior
- Suscribe the TypeScript Babel preset even for non-TS projects
- Watch for TS extensions for non-TS projects

My best argument for this new option is that having it false (default) will not alter Nuxt behavior. It needs to be an optional support and not a "feature" to handle every project needs.

galvez:
Closing this in favor of the upcoming #4557 -- cheers @chanlito and @kevinmarrec

PR #4407, opened 2018-11-25
feat: use individual lodash package to reduce size of nuxt-start

aldarund:
## Description
Import used lodash function from individual lodash npm packages to reduce size of nuxt-start
Improves https://github.com/nuxt/nuxt.js/issues/4377

## Checklist:


manniL:
We still need the full lodash for backwards compat (https://github.com/nuxt/nuxt.js/pull/4368). So the earliest time to merge this is 3.0.

Also, as @pi0 told me once, we'd add additional space if people are using Lodash in their libs/nuxt.config.js anyway)

aldarund:
@manniL if they use it and concerned about size they could also easily just use single npm package functions :)

manniL:
@aldarund Sometimes they can't (eg. when a dependency depends on the whole lodash package). :man_shrugging:

TheAlexLichter (approved):

pi0:
@aldarund Loadash is inlined using roll-up. Example: https://unpkg.com/@nuxt/common-edge@2.4.0-25718324.4eff50f/dist/common.js.

The code that instructs inlining: https://github.com/nuxt/nuxt.js/blob/dev/scripts/rollup.config.js#L51

So for `nuxt-start` dependencies, we are safe to simply remove `lodash` dependency and add it only to the top-level `package.json` :)

aldarund:
@pi0 ye, true :) new PR https://github.com/nuxt/nuxt.js/pull/4411

PR #4408, opened 2018-11-25
chore: add license to top level package json 

TheAlexLichter:
might fix
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "license": "MIT",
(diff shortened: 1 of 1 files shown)


pi0:
Private/Workspace packages are not required to have essential fields `name`, `version` and `license`. @manniL How do can I reproduce this warning?

manniL:
You can't :see_no_evil: 
It was my bad :sweat_smile: Somehow a `package.json` came in to my / dir.

Issue #4377, opened 2018-11-21
Nuxt node_modules exceeds 100mb in production and could be a lot smaller


rlindskog:
### What problem does this feature solve?

Cold starts on serverless platforms like Now 2.0 would be more efficient.

### What does the proposed changes look like?

When I have a `package.json` file that looks like this ([working example](https://github.com/rlindskog/nuxt-server-size/tree/master))

```
dependencies: {
  "nuxt": "^2.3.2"
}
```
Run with `nuxt-start`

`npm ci —production`

`nu -sch ./node_modules`: > **100mb**

When I have a `package.json` file that looks like this ([working example](https://github.com/rlindskog/nuxt-server-size/tree/standard))

```
"dependencies": {
  "@nuxt/core": "^2.3.2",
  "core-js": "^2.5.7"
},
devDependencies: {
  "nuxt": "^2.3.2"
}
```

and programatically run with `nuxt.render`

`npm ci`
`npm run build`
`npm ci —production`
`nu -sch ./node_modules` is **26mb**

Which is still pretty big.

I noticed the `nuxt-start` package, but it doesn’t work unless `nuxt` is installed as well, which brings us back to where we started at > 100mb.

In comparison `next` with `next-server` has a ***2.8mb*** `node_modules` footprint ([example](https://github.com/rlindskog/nuxt-server-size/tree/next)).

[next-server](https://github.com/zeit/now-builders/blob/master/packages/now-next/utils.js#L143) is used in the [@next/builder](https://github.com/zeit/now-builders/blob/master/packages/now-next/utils.js#L143).

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8183">#c8183</a>)</em></sub></div>

aldarund:
>I noticed the nuxt-start package, but it doesn’t work unless nuxt is installed as well, which brings us back to where we started at > 100mb.

That shouldnt happen. It work without nuxt and im using it for aws lambda serverless.

rlindskog:
@aldarund I was able to get `nuxt-start` with the recent updates, but also had to run `npm install --production core-js` to get it to work. Even with `nuxt-start`, the `node_modules` output is still **26mb** (it's basically the same as the [@nuxt/core implementation](https://github.com/rlindskog/nuxt-server-size/tree/standard)), which seems relatively large compared to what it could be.

[...]

PR #4419, opened 2018-11-26
refactor: remove caniuse-lite dependency

rlindskog:
## Types of changes

## Description
- This change lowers the size of `@node/server` from **12mb** to **8.1mb** (measured with `du -sh`, will be different when measuring with [package-phobia](https://packagephobia.now.sh))
- Why? Removes the large [`caniuse-lite` dependency - **3.1mb**](https://gist.github.com/pi0/dcf528794d29fa455c9ca745fb85fe04#file-nxut-start-txt-L136), which helps with #4377 
- Why not [`browserslist-useragent`](https://github.com/browserslist/browserslist-useragent)? `browserslist-useragent` checks compares the user agent against some conditionals, _and_ [checks if the browser name exists in `browser-list`](https://github.com/browserslist/browserslist-useragent/blob/master/index.js#L204) (which depends on `caniuse-lite`). All we need to do is check if the user-agent browser exists in [`modern-browsers.json`](https://github.com/nuxt/nuxt.js/blob/dev/packages/common/data/modern-browsers.json), and if the version is `gte` to the corresponding version.
- Why change  [`modern-browsers.json`](https://github.com/nuxt/nuxt.js/blob/dev/packages/common/data/modern-browsers.json)? To add the other browsers that `browserlist-useragent` bundled into "families", and changing the naming to match [`ua-parser-js` possible browser names](https://github.com/faisalman/ua-parser-js#methods)

**Notes**
- It might be a good idea to use [`useragent`](https://github.com/3rd-Eden/useragent#readme) instead, as `ua-parser-js` isn't meant for browser detection.

## Checklist:


pi0 (dismissed):

clarkdo (dismissed):

packages/common/data/modern-browsers.json
```diff
@@ -1,9 +1,11 @@
...
-  "edge": "16",
-  "firefox": "60",
-  "chrome": "61",
-  "safari": "10.1",
-  "opera": "48",
-  "ios_saf": "10.3",
-  "and_ff": "60"
+  "Edge": "16",
+  "Firefox": "60",
+  "Firefox Mobile": "60",
```
clarkdo: Does ua-parser-js support `Firefox Mobile` and `Chrome Headless` ? Because I don't see them in [`Possible 'browser.name'`](https://github.com/faisalman/ua-parser-js#methods)

aldarund: @clarkdo maybe we should use engine instead of browsers? There a lot of webkit based browsers, e.g. yandex, vivaldi etc.  While old browserslist-useragent was working for yandex, this new code will require adding Yandex to this list, as well as Vivaldi ( wasnt working before too )

rlindskog: @clarkdo Good catch!
- ["Firefox" takes care of mobile as well](https://github.com/faisalman/ua-parser-js/blob/master/src/ua-parser.js#L350) so "Firefox Mobile" can be removed for now.  There is a [PR up for that](https://github.com/faisalman/ua-parser-js/pull/251).
- ["Chrome Headless" is taken care of](https://github.com/faisalman/ua-parser-js/blob/master/src/ua-parser.js#L325)

@aldarund Should we change it to engine in this PR or address it in another?

aldarund: Actually it wont work, user agent report layout version and as i see from https://en.wikipedia.org/wiki/Google_Chrome_version_history
and it dont change from chrome 27 till now and we need javascript engine version and its tied to browser version itself

clarkdo (approved):

aldarund (approved):

pi0 (approved):

PR #4422, opened 2018-11-26
feat(modern): auto detect modern mode

clarkdo:
## Types of changes

## Description

If user doesn't specify `modern` in `command` and `nuxt.config.js`, a middleware will detect by if there is modern built asset in dist:
1. ssr: server
1. spa: client
1. no modern assets: false

## Checklist:

### DIFF
--- a/packages/server/src/middleware/modern.js
+++ b/packages/server/src/middleware/modern.js
@@ -1,3 +1,5 @@
+import chalk from 'chalk'
+import consola from 'consola'
 import { ModernBrowsers } from '@nuxt/common'
 import { matchesUA } from 'browserslist-useragent'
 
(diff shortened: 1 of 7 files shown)


packages/server/src/middleware/modern.js
```diff
@@ -11,12 +13,33 @@ const isModernBrowser = (ua) => {
...
 
-export default function (req, res, next) {
-  const { socket = {}, headers } = req
-  if (socket.modernMode === undefined) {
-    const ua = headers && headers['user-agent']
-    socket.modernMode = isModernBrowser(ua)
+let detected = false
+
+const detectModernBuild = ({ options, resources }) => {
+  if (detected === false && ![false, 'client', 'server'].includes(options.modern)) {
```
pi0: Can't we relax condition to `typeof options.modern !== `undefined`?

packages/server/src/middleware/modern.js
```diff
@@ -11,12 +13,33 @@ const isModernBrowser = (ua) => {
...
+    }
+  }
+}
+
+const detectModernBrowser = (req, options) => {
+  if (options.modern === 'server') {
+    const { socket = {}, headers } = req
+    if (socket.modernMode === undefined) {
+      const ua = headers && headers['user-agent']
+      socket.modernMode = isModernBrowser(ua)
```
pi0: Why it is directly set on `socket` object?

pi0 (approved):

packages/server/src/middleware/modern.js
clarkdo: I think what you mean is `modern === undefined`?

If user specified `modern:true`, we also need to detect the mode 😄

packages/server/src/middleware/modern.js
clarkdo: For reducing the matching times, we only match one time in a connection.

pi0: Damn good idea 👍 It worth mentioning in the comments :D

clarkdo:
Resolve #4422

PR #4428, opened 2018-11-26
fix: offer a new port and listen if already used, use consola on server error

### LINKED ISSUE #4421: Nuxt could check if localhost is free

If you run nuxt in dev mode and on the localhost is already running another nuxt instance, than you get very strange error. Maybe it will be nice to have cleaner error. Nice to have ;)

Human readable error

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8207">#c8207</a>)</em></sub></div>

ricardogobbosouza:
## Types of changes

## Description

- If `port` is already in use, offer a new and listen
- Use `consola` on server error

Resolves: #4421

## Checklist:


manniL:
#Before
image

# Current (after that PR)
image

`consola.fatal` will actually terminate the process because of https://github.com/nuxt/nuxt.js/blob/7c4e77ffb9d18279bd9d0c5f4d39eb55569c6a81/packages/cli/src/setup.js#L25-L42. So we might should roll with `consola.error`

galvez (dismissed):
Need to change to `consola.error()` like @manniL suggested but otherwise LGTM!

galvez (dismissed):

Atinux:
I think we can improve more the DX on this.

Why not checking the ECODE and display a better error instead when we know it?

manniL:
@Atinux Had the same idea. 

@ricardogobbosouza Do you want to take that or should we?

ricardogobbosouza:
@manniL, @Atinux,  I thought of something like this:

```js
this._server.on('error', (e) => {
  if (e.code === 'EADDRINUSE') {
    return consola.error(`Address already in use \`${this.host}:${this.port}\``)
  }

  consola.error(e)
}
```

What do you think?

manniL:
@ricardogobbosouza  Maybe like

> Address \\\`${this.host}:${this.port}\\\` is already in use. Do you run another service on the same port?

We could do that for *more common codes* as well (list of all: http://man7.org/linux/man-pages/man3/errno.3.html)

ricardogobbosouza:
What will we customize? Everyone would be unnecessary, do not you think?

manniL:
@ricardogobbosouza Not every single one :relaxed: 
Maybe `EDQUOT` and `EACCES` as well?
What do you think @pi0? :relaxed:

ricardogobbosouza:
@pi0 I fully agree that the error handling is done in the consola and that a new port should be suggested to the user.
I'm making this improvement...

manniL:
A new port suggestion is a nice idea! 👍

Atinux:
Looks good to me, great work @ricardogobbosouza 

The only thing missing seems to be the yarn.lock not updated

manniL:
@ricardogobbosouza I'd probably try to avoid that dependency by using a no-so-well-known trick 😏 

trick

Source: https://twitter.com/TheAlexLichter/status/1050862540035633152

ricardogobbosouza:
Hi @manniL, nice trick...
I confess I did not know :smiley:

pi0:
@ricardogobbosouza For zero random port, you may have to add a line after [here](https://github.com/nuxt/nuxt.js/blob/678605aa79822190391025e9453297f58aa5a0c7/packages/server/src/listener.js#L34) to update `this.port` value with `address.port`

HugoHeneault:
@manniL Nice idea. But it will be annoying to change ports at each nuxt restart. We need to restart it when we change nuxt.config.js and so we should update our tabs port? Would be nicer to take 3001 if 3000 is already taken, then 3002, etc.

manniL:
@HugoHeneault Well, `get-port` would also choose a random port 🤷‍♂️ 
But I see your point there 🤔

ricardogobbosouza:
@HugoHeneault I think a random port is better than creating a counter for it.
Because if the defined port is already in use, the `consola` will warn.

ricardogobbosouza:
@pi0 @manniL @Atinux 
What do you think? Better to put on an counter?
For me a random port resolves well

manniL:
I'm fine with a random port if the initial one is blocked 👍

packages/server/src/listener.js
```diff
@@ -72,4 +70,30 @@ export default class Listener {
     // Set this.listening to true
     this.listening = true
   }
+
+  async serverErrorHandler(e) {
+    if (e.code === 'EADDRINUSE') {
+      consola.warn(`Address \`${this.host}:${this.port}\` is already in use.`)
```
pi0: For production, we have to throw a FATAL error because the server is not usable and we silently continue listening on a random port. This makes production debug hard

packages/server/src/listener.js
```diff
@@ -72,4 +77,36 @@ export default class Listener {
...
+    if (addressInUse && this.dev && this.port !== '0') {
+      consola.warn(error.message)
+      consola.info('Trying a random port...')
+      this._server.close()
+      this.port = '0'
+      return this.serverListen()
+    }
+
+    // Throw error
+    throw error
```
pi0: Tip: This error is handled and should be handled by consola. The reason why not using `consola` here is that programatic usage needs to capture the listen error instead of silently skipping

ricardogobbosouza:
@pi0 nice!! :+1:

manniL:
:fire: :fire: :fire:

TheAlexLichter (approved):

TheAlexLichter (approved):

Issue #4225, opened 2018-10-29
CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory


ntvuong1709it:
### Version

[2](https://github.com/nuxt.js/releases/tag/2)

### Reproduction link

[https://github.com/ntvuong1709it/crawler-nuxt](https://github.com/ntvuong1709it/crawler-nuxt)

### Steps to reproduce

yarn run build

### What is expected ?

Build complete and be able to run using command yarn run start

### What is actually happening?

I received this error as following:
<--- Last few GCs --->

[9428:0000017514B27C20]    70741 ms: Mark-sweep 1377.3 (1467.6) -> 1377.3 (1468.1) MB, 1134.8 / 0.0 ms  allocation failure GC in old space requested
[9428:0000017514B27C20]    71815 ms: Mark-sweep 1377.3 (1468.1) -> 1377.2 (1434.1) MB, 1073.2 / 0.0 ms  last resort GC in old space requested
[9428:0000017514B27C20]    72862 ms: Mark-sweep 1377.2 (1434.1) -> 1377.2 (1434.1) MB, 1046.5 / 0.0 ms  last resort GC in old space requested

<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 000002521B2257C1 <JSObject>
    0: builtin exit frame: defineProperty(this=000001A2E8319A51 <Object map = 000003413FAF15F1>,0000038B0E00C759 <Object map = 000001FC8BD43E91>,0000033D19816A
F9 <String[5]: 10923>,0000004D3F016009 <Object map = 00000067614D3C41>)

    1: addGetter [My_Directory\node_modules\esm\esm.js:~1] [pc=0000000AC9FB3D64](this=0000004D3F015E59 <bc map = 000003B38C4262D1>,e=0000033D19816
AF9 <...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node_module_register
 2: v8::internal::FatalProcessOutOfMemory
 3: v8::internal::FatalProcessOutOfMemory
 4: v8::internal::Factory::NewFixedArray
 5: v8::internal::HashTable<v8::internal::SeededNumberDictionary,v8::internal::SeededNumberDictionaryShape>::EnsureCapacity
 6: v8::internal::Dictionary<v8::internal::SeededNumberDictionary,v8::internal::SeededNumberDictionaryShape>::Add
[...]

Issue #1728, opened 2017-09-25
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory


STnetwork:
Hi,

I use the nuxtent module, when I `yarn generate` I have this message : 

```
<--- Last few GCs --->

[8503:0x2a784d0]   124609 ms: Mark-sweep 1402.9 (1492.6) -> 1402.9 (1488.1) MB, 1949.4 / 0.0 ms  (+ 0.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 1949 ms) last resort 
[8503:0x2a784d0]   126604 ms: Mark-sweep 1402.9 (1488.1) -> 1402.9 (1488.1) MB, 1993.6 / 0.0 ms  last resort 


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x3949dd29cef1 <JSObject>
    0: builtin exit frame: assign(this=0x3949dd29b919 <JSFunction Object (sfi = 0x3949dd29b869)>,0x1400f1560559 <Object map = 0x2313e04a25f9>,0x7863dde1671 <Object map = 0x3ecc755372b9>,0x19704383cf91 <Object map = 0x2313e0484829>)

    1: renderAttrs [/home/nginx/domains/stnetwork.fr/public/test/nuxtent/src/node_modules/vue-server-renderer/build.js:~335] [pc=0xe90ecc5a480](this=0xea43946c941 <J...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x13740dc [node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewTransitionArray(int) [node]
 6: v8::internal::TransitionArray::Insert(v8::internal::Handle<v8::internal::Map>, v8::internal::Handle<v8::internal::Name>, v8::internal::Handle<v8::internal::Map>, v8::internal::SimpleTransitionFlag) [node]
[...]

Issue #1601, opened 2017-09-07
possible memory leak dev


uptownhr:
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

Started happening on the latest RC8. Happens on hmr rebuild. Sorry but can't provide more detail... I will try as I find more but I wanted to report this. 

Have not pushed to production yet, so not sure if it occurs but I assume it should not.

uptownhr:
https://ipfs.io/ipfs/QmQ91tdMog8yPJ7AN2zrreEza6hyYnb7x7EMS9YVg2pS35

uptownhr:
@alexchopin any tips on what I can due to isolate this? It does happen pretty frequently and I can confirm a memory leak. If I top, I can see the memory increase with every rebuild.

Atinux:
@uptownhr  Can you give us more informations on your system? RAM, node version, nuxt version, npm version, OS, etc.

choiRyan:
I've been having similar issues, having started with the express template with nuxt v0.10.5 on a generic t2.medium and t2.large Ubuntu14 instances on AWS. I've recently upgraded to 1.0.0-rc11 to see if it would fix it, but it hasn't, or another issue has come up.

Would the size of `server-bundle.json` make this issue worse? It's at 9.8MB when building with 1.0.0 rc-11, but "only" 3.58MB on my 0.10.5 builds.

Node versions: 6.11.1, 6.11.2, 6.11.3
NPM version 3.10.10

t2.medium | 2vCPUs | 4.0 GB RAM
t2.large | 2vCPUs | 8.0 GB RAM

The error I got when the process died:
Nuxt v0.10.5, in production:
```
<--- Last few GCs --->

 8720921 ms: Mark-sweep 2120.8 (1404.4) -> 2120.8 (1404.4) MB, 2169.9 / 0.0 ms (+ 0.7 ms in 1 steps since start of marking, biggest step 0.7 ms) [allocation failure] [GC in old space requested].
 8722890 ms: Mark-sweep 2120.8 (1404.4) -> 2139.1 (1403.4) MB, 1968.5 / 0.0 ms (+ 0.6 ms in 1 steps since start of marking, biggest step 0.6 ms) [last resort gc].
 8724856 ms: Mark-sweep 2139.1 (1403.4) -> 2157.4 (1403.4) MB, 1966.1 / 0.0 ms [last resort gc].

<--- JS stacktrace --->

[...]

Issue #1481, opened 2017-08-25
 allocation failure GC in old space requested


SzHeJason:
version : nuxt@rc-6

![image](https://user-images.githubusercontent.com/20609396/29712722-d2114a82-89cd-11e7-9f97-a25aaf2eac35.png)

pi0:
`nuxt.js:2066` in the picture above maps to [renderer.js:87](https://github.com/nuxt/nuxt.js/blob/dev/lib/core/renderer.js#L87) also trace shows error going into a JSON parse. We have two json parses one for client manifest and [one](https://github.com/nuxt/nuxt.js/blob/dev/lib/core/renderer.js#L562) for `server.bundle.json`. So i think there are two possible cases for your error:

- Limited resources available on your server.
  - We can try with node flag: `--max_old_space_size=8048`
- A REAL big server bundle file generated (Usually some compatible libraries)

Would you please run `nuxt build --analyze` and providing console output and bundle anylize results?

SzHeJason:
anylize results : 
![image](https://user-images.githubusercontent.com/20609396/29757014-4483d9d6-8bdb-11e7-8592-083aca16e169.png)
![image](https://user-images.githubusercontent.com/20609396/29757052-83d2d09c-8bdb-11e7-9b43-7e1f6af789b0.png)

SzHeJason:
Error JSON waiting for next time

SzHeJason:
Does not exist manifest and server.bundle.json in develop environment
![image](https://user-images.githubusercontent.com/20609396/29757852-9faaf406-8be1-11e7-8fc1-3c128c34e292.png)

alexchopin:
Should be fixed with the last release

yoyo837:
I just got some this：

91% additional chunk assets processing
<--- Last few GCs --->

[12584:000001F7C848A9E0] 13966592 ms: Mark-sweep 1268.4 (1463.1) -> 1268.4 (1463.1) MB, 338.8 / 0.1 ms  allocation failure GC in old space requested
[12584:000001F7C848A9E0] 13966862 ms: Mark-sweep 1268.4 (1463.1) -> 1252.9 (1424.6) MB, 270.6 / 0.1 ms  last resort GC in old space requested
[12584:000001F7C848A9E0] 13967108 ms: Mark-sweep 1252.9 (1424.6) -> 1252.9 (1414.6) MB, 245.3 / 0.1 ms  last resort GC in old space requested


[...]

PR #4439, opened 2018-11-27
feat: improve SSR bundle

pi0:
## Types of changes

## Results

Tested 0f104aa against a mid-size project:

 _         | Before                         | After                             | Diff
-------|----------------------|-----------------------|-------------------------
Dist     | 8.2M                 | 7.2M                  | -1M / 12%
Dev     | 281 MB (RSS: 439 MB) | 237 MB (RSS: 354 MB)  | -44M / 16% (RSS: -85M / 19%)
Start   | 106 MB (RSS: 169 MB) | 71.7 MB (RSS: 137 MB) | -34M / 32% (RSS: -32M / 19%)

## Objectives

- Better insights and inspection for server bundle

- Remove ~~all~~ vue related dependencies from `vue-renderer` package as much as possible to reduce install size of `nuxt-start`

- Support for single file distributions (serverless)

- Remove `server-bundle.json` and use the standard `.js` files for `dist/server`

- Mitigate **CALL_AND_RETRY_LAST Allocation failed** errors. Most of the cases happen on `JSON.parse()` the part when loading bundle. (#4225, #3465, #1728, #1601, #1481)

- Reduce server dist size by removing escape characters caused by JSON serialize

- Faster dev reloads and production start by removing extra `JSON.serialize`/`JSON.parse` time

- Less memory usage

- General performance improvements and refactors

- ~~Correctly ensure resources are available when using `nuxt start` or programmatically starting in production.~~ (Changes applied but to prevent breaking changes delayed to Nuxt3)

### TODO

- [x] Minify server files for production 
- [x] Cleanup source-map handling for reading from serverBundle
- [x] Remove fs-extra dependency from vue-renderer
- [x] Decouple loadResources
- [x] ~~Programmatic build warn when no dist exists (vue-renderer)~~ 

## Checklist:


TheAlexLichter (dismissed):

packages/vue-renderer/src/renderer.js
```diff
@@ -1,6 +1,6 @@
 import path from 'path'
 import crypto from 'crypto'
-import fs from 'fs-extra'
+import fs, { readFile, exists } from 'fs-extra'
```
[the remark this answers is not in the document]
pi0: @manniL I'll try to entirely remove the dependency of `fs-extra` and coupling

packages/vue-renderer/src/renderer.js
```diff
@@ -348,39 +363,55 @@ export default class VueRenderer {
...
+            maps
+          }
+        }
       },
-      {
-        key: 'ssrTemplate',
+      ssrTemplate: {
         fileName: 'index.ssr.html',
-        transform: this.parseTemplate
+        transform: src => this.parseTemplate(src)
```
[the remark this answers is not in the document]
pi0: Reasons that I changed:

- Consistency with other transforms
- Make it more explicit which args are being passed
- Make it explicitly context-free (no bind)

If you still think the previous style is more suitable I'll revert it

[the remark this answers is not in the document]
manniL: Somehow I actually missed how it was before 😂 

Good reasoning 👍

clarkdo (commented):
For server building, do we need code-splitting ?
We may build a pure single file bundle like ncc is doing.

packages/webpack/src/builder.js
```diff
@@ -27,9 +27,11 @@ export class WebpackBundler {
     this.devMiddleware = {}
     this.hotMiddleware = {}
 
-    // Initialize shared FS and Cache
+    // Initialize shared MFS for dev
     if (this.context.options.dev) {
       this.mfs = new MFS()
+      this.mfs.exists = function (...args) { return Promise.resolve(this.existsSync(...args)) }
```
clarkdo: It seemed that the rewrited `exists` and  `readFile` are not used in `loadResources`.

BTW: `memory-fs` already provided `exists` with `callback` version, so maybe we just need a pify ?

```suggestion
this.mfs.exists = pify(this.mfs.exists)
```

packages/webpack/src/plugins/vue/server.js
```diff
@@ -44,20 +44,17 @@ export default class VueSSRServerPlugin {
 
       stats.assets.forEach((asset) => {
         if (isJS(asset.name)) {
-          bundle.files[asset.name] = compilation.assets[asset.name].source()
+          bundle.files[asset.name] = asset.name
         } else if (asset.name.match(/\.js\.map$/)) {
-          bundle.maps[asset.name.replace(/\.map$/, '')] = JSON.parse(compilation.assets[asset.name].source())
+          bundle.maps[asset.name.replace(/\.map$/, '')] = asset.name
```
clarkdo: Love this part neat and smart change!

packages/webpack/src/builder.js
pi0: I've tried pify. It is not a standard callback (error first) and even with passing proper options it was still buggy. So i ended up with this solution that was the safest one!

packages/webpack/src/config/server.js
```diff
@@ -61,6 +61,27 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
     return plugins
   }
 
+  extendConfig() {
+    const config = super.extendConfig(...arguments)
+
+    // Add minimizer plugins
+    if (config.optimization.minimize && config.optimization.minimizer === undefined) {
```
clarkdo: Since both client and server need minimize, can we move it into base ?

packages/vue-renderer/src/renderer.js
```diff
@@ -348,39 +378,57 @@ export default class VueRenderer {
...
+        transform: src => JSON.parse(src)
       },
-      {
-        key: 'serverBundle',
-        fileName: 'server-bundle.json',
-        transform: JSON.parse
+      serverManifest: {
+        fileName: 'server.manifest.json',
+        // BundleRenderer needs resolved contents
+        transform: (src, { readResource, oldValue = { files: {}, maps: {} } }) => {
```
clarkdo: Can we use `this.context.resources. serverManifest ` instead of oldValue ?

packages/webpack/src/config/server.js
pi0: There may be more differences such as the way of comment extraction. What do you think to add `minimizer()` function into base and override it in server?

packages/vue-renderer/src/renderer.js
pi0: Exactly. I just wanted to make transforms contextless :)

packages/webpack/src/config/server.js
clarkdo: Yes, `minimizer` is a good idea

clarkdo (dismissed):
LGTM

packages/vue-app/package.json
```diff
@@ -10,6 +10,14 @@
   ],
   "main": "dist/vue-app.js",
   "typings": "types/index.d.ts",
+  "dependencies": {
+    "vue": "^2.5.17",
+    "vue-meta": "^1.5.6",
+    "vue-no-ssr": "^1.1.0",
+    "vue-router": "^3.0.2",
+    "vue-template-compiler": "^2.5.17",
+    "vuex": "^3.0.1"
```
aldarund: 3.0.2

PR #4467, opened 2018-12-02
feat: check modern build file in modern mode

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -103,10 +103,16 @@ export default class VueRenderer {
       await this.loadResources(fs)
 
       // Verify
-      if (!this.isReady && this.context.options._start) {
-        throw new Error(
-          'No build files found. Use either `nuxt build` or `builder.build()` or start nuxt in development mode.'
-        )
+      if (this.context.options._start) {
+        if (!this.isReady) {
+          throw new Error(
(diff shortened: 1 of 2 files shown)


packages/vue-renderer/src/renderer.js
```diff
@@ -103,10 +103,16 @@ export default class VueRenderer {
...
-      if (!this.isReady && this.context.options._start) {
-        throw new Error(
-          'No build files found. Use either `nuxt build` or `builder.build()` or start nuxt in development mode.'
-        )
+      if (this.context.options._start) {
+        if (!this.isReady) {
+          throw new Error(
+            'No build files found. Use either `nuxt build` or `builder.build()` or start nuxt in development mode.'
+          )
+        } else if (this.context.options.modern && !this.context.resources.modernManifest) {
```
pi0: Shouldn't we also include this check for isReady?

Atinux: I think so too

clarkdo: @pi0 @Atinux Since modern needs a different error message, so I didn't add it into isReady, do you mean change isReady to not only return boolean ?

pi0: > do you mean change isReady to not only return boolean

I like this idea

pi0 (approved):
Since this PR is a hotfix, we can improve internals and do refactor for `isReady` later

PR #4472, opened 2018-12-02
feat: add build.crossorgin

clarkdo:
## Types of changes

## Checklist:


TheAlexLichter (approved):

aldarund (approved):

packages/server/src/middleware/nuxt.js
```diff
@@ -52,7 +52,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
       const links = pushAssets
         ? pushAssets(req, res, publicPath, preloadFiles)
-        : defaultPushAssets(preloadFiles, shouldPush, publicPath, options.dev)
+        : defaultPushAssets(preloadFiles, shouldPush, publicPath, options)
```
pi0: This change would be a breaking change for `2.x` releases (changing function signature). If the options than `dev` we can add it as another param. For `3.x` i suggest changing signature to accept an object with named args.

clarkdo: `defaultPushAssets` is not the one exposed to users, it's only for internal use.
`pushAssets` is for user config, we can add options to it in 3.x.

packages/webpack/src/config/client.js
```diff
@@ -124,6 +125,12 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       }))
     }
 
+    if (this.options.build.crossorigin) {
+      plugins.push(new CorsPlugin({
```
manniL: Could be require the plugin lazily here (Would that make any difference)?

clarkdo: This part is for building, so it won't be bundled into server.js, so I thinks it's fine.

atinux (approved):

pi0 (approved):

PR #4485, opened 2018-12-05
Add components nuxtServerInit hooks on components matched by the route

Pavel-Mishalov:
This hook will be executed when the server is loading the application
and will not be executed when routing in the browser.

## Types of changes

## Description

This hook is required to perform actions in the component when the application is first loaded and does not require a call when routing on the client side

## Checklist:

### DIFF
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -103,6 +103,18 @@ export default async (ssrContext) => {
   <% } %>
 
   /*
+  ** Call components nuxtServerInit hooks on components matched by the route.
+  */
+  const pageInitStatus = await Promise.all(
+    Components.map(Component => {
+      if (Component.options.nuxtServerInit && typeof Component.options.nuxtServerInit === 'function') {
+        return Component.options.nuxtServerInit(app.context)
+      }
+      return null
+    })
+  )
+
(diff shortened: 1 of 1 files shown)


manniL:
Hey @Pavel-Mishalov 👋 

Thanks for the PR. I'd probably wait for https://github.com/vuejs/vue/pull/9017 to get merged into Vue so we have that functionality out of the box. 

For now I'd go with:

1. `nuxtServerInit` in the store
2. `asyncData` + `process.server` for page components

Pavel-Mishalov:
Thank @manniL for considering my proposal, I didn’t immediately check option 2 for the idea it should work out accordingly

galvez:
@Atinux @pi0 also agree with @manniL here. Should we close?

@Pavel-Mishalov the problem is that we're likely introducing similar enhancements based on the upcoming `ssrPrefetch` option, so introducing an isolated, stored-specific API for this right now might create a conflict in the near future.

pi0:
@manniL @galvez Thanks for your insights.

@Pavel-Mishalov You contribution is more than appreciated by us. To prevent doing the same thing in a different way I suggest waiting for Vue.js ssrPrefetch.

PR #4515, opened 2018-12-10
fix(webpack): generate client source map if devtool set in extend of …

aldarund:
…config

## Types of changes

## Description
make source map generated based on devtool value that could be set in extend

## Checklist:

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -399,7 +399,7 @@ export default class WebpackBaseConfig {
     // Clone deep avoid leaking config between Client and Server
     const extendedConfig = this.extendConfig(cloneDeep(config))
     const optimization = extendedConfig.optimization
-    if (optimization.minimizer && extendedConfig.devtool) {
+    if (optimization && optimization.minimizer && extendedConfig.devtool) {
       const terser = optimization.minimizer.find(p => p instanceof TerserWebpackPlugin)
       if (terser) {
         terser.options.sourceMap = extendedConfig.devtool && /source-?map/.test(extendedConfig.devtool)
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -397,6 +397,15 @@ export default class WebpackBaseConfig {
     }
 
     // Clone deep avoid leaking config between Client and Server
-    return this.extendConfig(cloneDeep(config))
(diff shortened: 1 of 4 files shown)


packages/webpack/src/config/base.js
```diff
@@ -397,6 +397,15 @@ export default class WebpackBaseConfig {
     }
 
     // Clone deep avoid leaking config between Client and Server
-    return this.extendConfig(cloneDeep(config))
+    const extendedConfig = this.extendConfig(cloneDeep(config))
+    const optimization = extendedConfig.optimization
```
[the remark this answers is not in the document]
aldarund: it can, the thing is that in our docs we specifically provide example where its set in extend
https://nuxtjs.org/api/configuration-build#extend
And breaking it somehow a breaking change for some users. Thats first point. Second point is that it might be set in extend only for client like in example from link, while nuxt.config dont allow such option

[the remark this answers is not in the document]
clarkdo: Oh, ok.
BTW, optimization can be false

pi0: This can be simplified into:
```js
const { optimization } = extendedConfig
```

packages/webpack/src/config/base.js
```diff
@@ -397,6 +397,15 @@ export default class WebpackBaseConfig {
...
+    if (optimization && optimization.minimizer && extendedConfig.devtool) {
+      const terser = optimization.minimizer.find(p => p instanceof TerserWebpackPlugin)
```
pi0: General tip: Terser is not the only thing that is affected. All loaders for css also need an update. I suggest adding a comment in above of this condition to remove this in Nuxt 3.x.

pi0 (dismissed):
Other than comments, LGTM

packages/webpack/src/config/base.js
aldarund: Not quite following :) for css we have cssSourceMap property.  How do u want to handle it in nuxt 3 ? Something like devtool: { server: 'server-sourcemap', client: 'client-source-map' }  ?

pi0: Yes. I think it would be better to explicitly set devTool options and not using `extend`. Or better [SourceMapDevToolPlugin](https://webpack.js.org/plugins/source-map-dev-tool-plugin) for more control.

pi0 (approved):

PR #4525, opened 2018-12-11
feat(vue-app): Add n-link and n-child components shortcuts

atinux:
I do want the shortcuts :)

We will still support `<nuxt-link>` and `<nuxt-child/>` but I do recommend to document `<n-link>` and `<n-child/>` instead.

screenshot 2018-12-11 at 17 40 18

### DIFF
--- a/examples/hello-world/pages/about.vue
+++ b/examples/hello-world/pages/about.vue
@@ -1,9 +1,7 @@
 <template>
   <div>
     <p>Hi from {{ name }}</p>
-    <NuxtLink to="/">
-      Home page
-    </NuxtLink>
+    <n-link to="/">Home page</n-link>
   </div>
 </template>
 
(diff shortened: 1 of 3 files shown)


pi0 (approved):
Nice idea

manniL:
Things to do:

* [ ] Change all examples
* [ ] Change documentation (base branch: 2.4)

manniL:
Though it's already merged (but not released), I'd love to hear the opinion of the whole team on that :relaxed: 

cc @aldarund @clarkdo @galvez

clarkdo:
I think add one test for asserting origin and shortcut component is necessary. 
About expamples, hello word shoud be enough, because nuxt-* are still official ones, they won’t be replaced by shortcuts.
In doc, we can add alias or shortcut field after component name.

manniL:
So basically these are the two key questions:


1. Should we add the aliases?
2. Should we use them primarily in docs and examples?

If I got it right, @Atinux proposed to do both, 1 and 2.

PR #4548, opened 2018-12-13
fix(resolver) allow directories with index.[ext] files to be resolved as modules

mrkstwrt:
## Types of changes

## Description

Currently if I have modules in my project like this
```
~
| - modules
  | - my-module
    | - index.js
```
I can have this in my `nuxt.config.js`
```
{
  modules: [
    '~/modules/my-module',
  ]
}
```

and it works fine. However if you use any other extension (including extensions listed in `extensions` in config) it will not work. This effects me because I'm using TS, it will work if you use `~/modules/my-module/index.ts` or `~/modules/my-module/index` however since it works for .js files it feels like this is a bug.

## Checklist:

### DIFF
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -69,15 +69,19 @@ export default class Resolver {
       resolvedPath = path
     }
 
-    // Check if resolvedPath exits
-    if (fs.existsSync(resolvedPath)) {
+    // Check if resolvedPath exits and is not a directory
+    if (fs.existsSync(resolvedPath) && !fs.lstatSync(resolvedPath).isDirectory()) {
       return resolvedPath
     }
 
-    // Check if any resolvedPath.[ext] exists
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Nice improvements ☺️

kevinmarrec:
@pi0 I updated my project to use last version of `nuxt-edge` and after investigation it seems that this PR broke it.

It seems we can't use `modules: ['@@']` anymore in `nuxt.config.js` (this behavior is used when testings modules (e.g. https://github.com/Developmint/nuxt-webfontloader/blob/master/test/fixture/configs/default.js)

kevinmarrec:
I don't know what does the `'@@'` btw, I just saw that @manniL was using this behavior to test his module so I copied the same test setup in my project, and it was working, until I upgraded `nuxt-edge` :P

PR #4557, opened 2018-12-14
feat(builder): optional typescript support

kevinmarrec:
## Types of changes

## ⚠ Warning ⚠ (EDIT AFTER MERGE)

➡ Please be aware of #4563 , which will likely make this merged PR outdated. ⬅

## Description

What if Typescript Support what just about setting a boolean variable to `true` ? 😎

### Prequel

This PR is highly inspired by #4406 made by @chanlito, so special thanks to him.

His current PR state alters Nuxt packages even for non-TS projects, and will also break projects using their own TypeScript module, that's why I implemented a new way to provide the TS support.

Existing TS projects will be able to either keep their own implementation of TypeScript support *OR* use this built-in **optional** one.

### I. Usage 👨‍💻👩‍💻

```js
// nuxt.config.js
export default {
  build: {
    typescript: true // 🚀
  }
}
```

### II. How it works behind the scenes ? ⚙

When the `typescript` option is set to true, and only if is set true, it alters the following Nuxt internal packages :

✏ Config (**@nuxt/config**)  
*Accept `.ts(x)` files for plugins and middleware*

✏ Builder (**@nuxt/builder**) & Common(**@nuxt/common**)  
*Watch & Use `.ts(x)` files to generate routes (layouts & pages)*

✏ Webpack (**@nuxt/webpack**)  
*Resolve `.ts(x)` files through Babel, passing the typescript option value to the Nuxt babel preset*

✏ Babel (**@nuxt/babel-preset-app**)  
*Retrieve the typescript option from the preset options to use the [@babel/preset-typescript](https://www.npmjs.com/package/@babel/preset-typescript) Babel TypeScript preset.*

✏ CLI (**@nuxt/cli**) [*Suggestion not implemented yet*]  
*Warning or Error if `tsconfig.json` not found in the source folder, using the `srcDir` config option*

### III. What about the documentation ? 📖

I suggest a **Typescript** section in the documentation *OR* a **How to use Typescript ?** in the FAQ.

I will (or at least propose to) take the lead on it if my proposal is accepted and submit a PR regarding this documentation. This latter will talk about type definitions as well (#4164).

### IV. What about a Nuxt TS project example ? 📁

An example highly similar to the one in #4406 will be proposed in a different PR if this proposal is accepted.

### V. What about tests ? 🛠

A test fixture and an unit test are linked in the PR.

**NOTE :** Nuxt TS community is growing, we already have a bunch of developers who will be ready to test this support using `nuxt-edge` if this proposal gets accepted and merged 😇

@chanlito @hartmut-co-uk @husayt @ksnyde @NickBolles

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -277,7 +277,7 @@ export default class Builder {
       layoutsFiles.forEach((file) => {
         const name = file
           .replace(new RegExp(`^${this.options.dir.layouts}/`), '')
-          .replace(new RegExp(`\\.(${this.supportedExtensions.join('|')})$`, ''))
+          .replace(new RegExp(`\\.(${this.supportedExtensions.join('|')})$`), '')
         if (name === 'error') {
           if (!templateVars.components.ErrorPage) {
             templateVars.components.ErrorPage = this.relativeToBuild(
@@ -317,7 +317,7 @@ export default class Builder {
         cwd: this.options.srcDir,
         ignore: this.options.ignore
       })).forEach((f) => {
-        const key = f.replace(new RegExp(`\\.(${this.supportedExtensions.join('|')})$`, ''))
+        const key = f.replace(new RegExp(`\\.(${this.supportedExtensions.join('|')})$`), '')
         if (/\.vue$/.test(f) || !files[key]) {
           files[key] = f.replace(/(['"])/g, '\\$1')
         }
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -46,7 +46,7 @@ export default class Builder {
 
     this.supportedExtensions = ['vue', 'js']
     if (this.options.build.typescript) {
-      this.supportedExtensions = this.supportedExtensions.concat(['ts', 'tsx'])
+      this.supportedExtensions.push('ts', 'tsx')
     }
 
     // Helper to resolve build paths
@@ -326,7 +326,7 @@ export default class Builder {
         Object.values(files),
         this.options.srcDir,
         this.options.dir.pages,
-        this.options.build.typescript
+        this.supportedExtensions
       )
     } else { // If user defined a custom method to create routes
       templateVars.router.routes = this.options.build.createRoutes(
@@ -508,20 +508,19 @@ export default class Builder {
 
   watchClient() {
     const src = this.options.srcDir
+    let rGlob = (dir) => ['*', '**/*'].map(glob => r(src, `${dir}/${glob}.{${this.supportedExtensions.join(',')}}`))
 
     let patterns = [
       r(src, this.options.dir.layouts),
       r(src, this.options.dir.store),
       r(src, this.options.dir.middleware),
-      r(src, `${this.options.dir.layouts}/*.{${this.supportedExtensions.join(',')}}`),
-      r(src, `${this.options.dir.layouts}/**/*.{${this.supportedExtensions.join(',')}}`)
+      ...rGlob(this.options.dir.layouts)
     ]
 
     if (this._nuxtPages) {
       patterns.push(
         r(src, this.options.dir.pages),
-        r(src, `${this.options.dir.pages}/*.{${this.supportedExtensions.join(',')}}`),
-        r(src, `${this.options.dir.pages}/**/*.{${this.supportedExtensions.join(',')}}`)
+        ...rGlob(this.options.dir.pages)
       )
     }
 
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -508,7 +508,7 @@ export default class Builder {
 
   watchClient() {
     const src = this.options.srcDir
-    let rGlob = (dir) => ['*', '**/*'].map(glob => r(src, `${dir}/${glob}.{${this.supportedExtensions.join(',')}}`))
+    const rGlob = dir => ['*', '**/*'].map(glob => r(src, `${dir}/${glob}.{${this.supportedExtensions.join(',')}}`))
 
     let patterns = [
       r(src, this.options.dir.layouts),
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -44,6 +44,11 @@ export default class Builder {
       restart: null
     }
 
+    this.supportedExtensions = ['vue', 'js']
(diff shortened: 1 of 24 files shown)


kevinmarrec:
It seems that I break something around the Regex's regarding layouts & pages. I will fix it soon 💪.

NickBolles:
@kevinmarrec Not sure if this is how this works, or if it would be an issue, but I think that [this line](https://github.com/nuxt/nuxt.js/compare/dev...kevinmarrec:typescript-support#diff-f61e149585ca91093ef9329346d5e992R96
) will evaluate to true for typescript options for [`nuxt-ts-module`](https://github.com/chanlito/nuxt-ts-module)

maybe do `=== true`? 

I can try to test if using nuxt from this PR will break anything in my app, but probably wont get to it for a few days.

kevinmarrec:
@NickBolles Yeah it will cause issues indeed, your suggestion or finding a different name for the option are 2 valid solutions.

NickBolles:
Are we ever going to want to pass options to Typescript? We might want it to be an object at some point, for example to pass a location for the tsconfig or something.

Could we simply move it to the build object? it kinda is a build-related option.

packages/builder/src/builder.js
```diff
@@ -44,6 +44,11 @@ export default class Builder {
       restart: null
     }
 
+    this.supportedExtensions = ['vue', 'js']
+    if (this.options.build.typescript) {
+      this.supportedExtensions = this.supportedExtensions.concat(['ts', 'tsx'])
```
pi0: A simple push (with multi args) should be enough

packages/builder/src/builder.js
```diff
@@ -44,6 +44,11 @@ export default class Builder {
       restart: null
     }
 
+    this.supportedExtensions = ['vue', 'js']
```
pi0: There are common use cases with joined version of this array. We may somehow refactor it

pi0 (dismissed):
Generally changes (till know) LGTM. Nice works 😊

pi0:
Also good decision to move it into build.typescript. I was going to tell that :))

kevinmarrec:
@pi0 Thanks for your fast feedback ! I will take into accounts your code comments soon and I need to fix the unit test :).

packages/common/src/utils.js
```diff
@@ -315,12 +315,16 @@ const sortRoutes = function sortRoutes(routes) {
   return routes
 }
 
-export const createRoutes = function createRoutes(files, srcDir, pagesDir) {
+export const createRoutes = function createRoutes(files, srcDir, pagesDir, typescript = false) {
```
galvez: @pi0 @kevinmarrec I wonder if we should add `supportedExtension` somewhere under `nuxt/config`.

galvez: Perhaps under the `router` property in `nuxt.config.js`.

kevinmarrec: @galvez This `createRoutes` is used in `test/unit/utils.test.js` which only test utils, so i don't know if we can pass supportedExtensions here :/.

And i don't know about the `nuxt/config`

packages/builder/src/builder.js
```diff
@@ -502,20 +508,19 @@ export default class Builder {
 
   watchClient() {
     const src = this.options.srcDir
+    const rGlob = dir => ['*', '**/*'].map(glob => r(src, `${dir}/${glob}.{${this.supportedExtensions.join(',')}}`))
```
kevinmarrec: @pi0 Thoughts about this thing are welcome. This refactored well what was before in this `watchClient` method. For the rest of the code using `this.supportedExtensions`, I'm not sure how we can refactor it, any ideas or current state is ok to you ?

kevinmarrec:
Tests are green 🚥✔

galvez (dismissed):
You're right @kevinmarrec -- LGTM!

packages/builder/src/builder.js
pi0: Let's keep it as is for now. Nothing blocking :)

pi0:
Merging to get this enhancenment sooner with `nuxt-edge`. According to PR notes these items are delegated to other PRs:

- Warning or Error if tsconfig.json not found in the source folder, using the srcDir config option
- A Typescript section in the documentation OR a How to use Typescript ? in the FAQ.
- Nuxt TS project example

pi0 (approved):

manniL:
We need docs as well!

kevinmarrec:
@husayt There are issues with Vue SFC's using the Babel approach, i'm not sure you'll be able to make your project work ^^

hartmut-co-uk:
@husayt see https://github.com/nuxt/nuxt.js/pull/4563

kevinmarrec:
I updated the description of this PR with a warning refering to #4563

chanlito:
@kevinmarrec What is wrong with this PR?

kevinmarrec:
Check #4563 first commit, babel doesn't transpile well Vue SFC.

vlajos:
@kevinmarrec 
Should this 
```
// nuxt.config.js
export default {
  build: {
    typescript: true // 🚀
  }
}
```
still work with 2.6.x ?

kevinmarrec:
@vlajos No, there has been multiple refactors/breaking changes since this behavior, you can check https://nuxtjs.org/guide/typescript to understand properly how you enable TypeScript support in your Nuxt project :)

vlajos:
@kevinmarrec I see. Thanks. I wanted to use it with the nuxtbuilder, but could not really make it working.
I wanted to migrate a nuxt+socket.io project. The basics are roughly like this: https://github.com/nuxt/nuxt.js/tree/dev/examples/with-sockets
Is this possible at all?

kevinmarrec:
@vlajos Hm yes but partially, cause we don't have Nuxt Builder types defined, cause they are depending on too many things in core which also need their types to be defined, which would mean rewrite Nuxt in TypeScript.

If you need further help, join our Discord and I'll try to take some time to give you more guidelines for your migration.

vlajos:
@kevinmarrec Thank you,but practically I gave it up. I wanted to use it mainly to be able to work with typeorm easier, but it seems that it works without typescript as well. Anyway when it becomes a little more supported I may give it another go.
Thanks again!

PR #4558, opened 2018-12-15
fix: properly serialize head functions

galvez:
Picking up on #4079 -- updated tests.

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -429,11 +429,22 @@ export default class Builder {
       resolve: r
     })
 
+    const serializeHead = (obj) => {
+      obj = Object.assign({}, obj)
+      for (const member in obj) {
+        if (typeof obj[member] === 'function') {
+          obj[member] = (...args) => obj[member](...args)
+        }
+      }
+      return serialize(obj)
+    }
+
     // Prepare template options
     let lodash = null
     const templateOptions = {
       imports: {
         serialize,
+        serializeHead,
         devalue,
         hash,
         r,
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -60,7 +60,7 @@ export default class Builder {
           serialized[member] = eval(`(function${body})`)
         }
       })
-      return serialize(obj)
+      return serialize(obj).replace('head(', 'function(')
     }
 
     this._buildStatus = STATUS.INITIAL
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -48,6 +48,21 @@ export default class Builder {
     this.relativeToBuild = (...args) =>
       relativeTo(this.options.buildDir, ...args)
 
+    this.serializeHead = (obj) => {
+      const serialized = Object.assign({}, obj)
+      let body
+      Object.keys(obj).forEach((member) => {
+        if (typeof obj[member] === 'function') {
+          body = obj[member].toString()
+          body = body.slice(body.indexOf('('))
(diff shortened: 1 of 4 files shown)


galvez:
So I have no idea why the test is failing since

```js
.replace(/:\w+\(/gm, ':function(').replace('head(', 'function(')
```

returns the same result as the new `serializeHead()` does.

Still looking into it -- but appreciate any insight @pi0 @manniL @aldarund @clarkdo :)

galvez:
Thnk I got it now :)

btkostner:
I would still recommend not serializing the head to avoid issues

galvez:
@btkostner I'm afraid not serializing is impossible, even in your original PR, you kept the `serialize()` call. That is because using a _raw head_ would require loading and parsing the head from `nux.config.js` as a string, in isolation. Using `Function.toString()` by itself still yields broken code for method functions. So manipulation is required here -- the problem was that it was previously doing an uncontrolled, global replace on a overly eager pattern. This PR's approach does the same basic operation, but to more strictly matched function declarations.

packages/builder/src/builder.js
```diff
@@ -48,6 +48,23 @@ export default class Builder {
     this.relativeToBuild = (...args) =>
       relativeTo(this.options.buildDir, ...args)
 
+    this.serializeHead = (obj) => {
```
pi0: This function can be potentially moved to `@nuxt/common` utils, with the enhancenment of getting `"head"` as the second param or generalizing regex. There are more use cases than would benefit from a consistent utility.

pi0 (dismissed):

packages/builder/src/builder.js
galvez: @pi0 sure thing -- please wait until I do this before merging then :)

galvez: @pi0 done!

packages/common/src/utils.js
```diff
@@ -457,3 +458,20 @@ export function defineAlias(src, target, prop, opts = {}) {
     }
   })
 }
+
+export function serializeWithMethods(obj, name) {
```
pi0: A final change request 🙈 Why not calling it `serializeFunction`? (And also ensuring first arg `fn` is an instance of function)

pi0: PS: We may infer the function name using `fn.name` :)  (Maybe an extra replace to ensure the name is stripped off)

galvez: done! :)

packages/common/src/utils.js
```diff
@@ -457,3 +458,20 @@ export function defineAlias(src, target, prop, opts = {}) {
...
+  return serialize(func)
+    .replace(/^(\s*):(\w+)\(/gm, (_, spaces) => {
+      return `${spaces}:function(`
+    })
+    .replace(/^(\s*)(\w+)\s*\((.*?)\)\s*\{/gm, (_, spaces, name, args) => {
+      if (open) {
+        return `${spaces}${name}:function(${args}) {`
+      } else {
+        open = true
+        return _
```
pi0: General tip: _ as function argument is usually for ignoring that arg. (IDEs and Eslint can imply this and don't warn about unused arg)

pi0 (approved):
Looking good 🤘

packages/common/src/utils.js
galvez: @pi0 yeah, I initially used it with this purpose, but the algorithm changed -- since it's meant to be discarded after the 1st iteration I thought it was OK to leave as-is.

PR #4564, opened 2018-12-16
feat: allow scrollToTop to be explicitly disabled

galvez:
Picking up #4170 -- debugged and added test.


pi0 (approved):

manniL:
Is a change in the docs necessary?

galvez:
@manniL yep, will do!

homerjam:
Thanks @galvez ! Holidays got in the way for me I'm afraid 😊

PR #4565, opened 2018-12-16
fix: add serialize-javascript to common package

galvez:
As reported by @kevinmarrec -- apparently this should be here, please shed a light @pi0 @clarkdo if I'm mistaken :)

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -8,7 +8,8 @@
   ],
   "main": "dist/common.js",
   "dependencies": {
-    "consola": "^2.3.0"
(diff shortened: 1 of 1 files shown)


pi0:
@galvez Seems reasonable. Without this, it will still work because common is not meant to be used standalone. But in the future (long or short term I mean) it would be better splitting common helpers that are only used in one core package into their own. BTW OverHead ([install size](https://packagephobia.now.sh/result?p=serialize-javascript)) of this package is not even noticable for nuxt-start :)

pi0 (approved):

PR #4573, opened 2018-12-17
feat(store): nuxtClientInit action

galvez:
As requested in https://github.com/nuxt/nuxt.js/issues/3938 -- I think it makes sense, for consistency.

If approved I'll update docs right away :)


packages/vue-app/template/client.js
```diff
@@ -490,6 +490,14 @@ function fixPrepatch(to, ___) {
 }
 
 function nuxtReady(_app) {
+  <% if (store) { %>
+  if (app.store._actions && app.store._actions.nuxtClientInit) {
+    app.store.dispatch('nuxtClientInit', app.context)
```
manniL: We should await the dispatch here, shouldn't we?

TheAlexLichter (changes requested):

packages/vue-app/template/client.js
galvez: @manniL you're too fast :) check again!

TheAlexLichter (dismissed):

clarkdo (commented):
And some points we may discuss.

1. nuxtClientInit is still executed in the page rendered by ssr
1. nuxtClientInit is executed after `middlewares` which is different from `nuxtServerInit`

packages/vue-app/template/client.js
```diff
@@ -629,7 +640,11 @@ async function mountApp(__app) {
     // Listen for first Vue update
     Vue.nextTick(() => {
       // Call window.{{globals.readyCallback}} callbacks
+      <% if (store) { %>
+      nuxtClientInit(_app).then(() => nuxtReady(_app))
```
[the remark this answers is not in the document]
galvez: @clarkdo I'm not entirely sure how to go about this, can you exemplify? Or do you mean returning to the first implementation (with no await)?

packages/vue-app/template/client.js
```diff
@@ -629,7 +640,11 @@ async function mountApp(__app) {
...
+      <% } else { %>
       nuxtReady(_app)
+      <% } %>
       <% if (isDev) { %>
       // Enable hot reloading
       hotReloadAPI(_app)
```
manniL: I think @clarkdo refers to that line here @galvez

packages/vue-app/template/client.js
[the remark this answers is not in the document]
clarkdo: I mean sth like:

```js
  Vue.nextTick(asunc () => {
    // Call window.{{globals.readyCallback}} callbacks
    await nuxtClientInit(_app)
    nuxtReady(_app)
    <% if (isDev) { %>
    // Enable hot reloading
    hotReloadAPI(_app)
    <% } %>
  })
```

galvez:
@clarkdo Added a check for `process.client` -- I am unsure how to make it run before middleware though, client.render() (which calls middleware) doesn't seem to get called in the first render at least. Perhaps this is because the app I'm test with is set to `universal`?

clarkdo:
@galvez Have you tried SPA?

galvez:
@clarkdo Trying that now, slapped a WIP label on this PR for now.

packages/vue-app/template/client.js
```diff
@@ -627,7 +644,10 @@ async function mountApp(__app) {
     _app.$mount('#<%= globals.id %>')
 
     // Listen for first Vue update
-    Vue.nextTick(() => {
+    Vue.nextTick(async () => {
+      <% if (store) { %>
+      await nuxtClientInit(_app)
```
clarkdo: Since it's already in `render`, maybe here is unnecessary.

clarkdo:
Sorry, maybe I misunderstood the usage, is `nuxtClientInit` supposed executed in `universal` 
since `nuxtServerInit` is not executed  in `SSR`?

galvez:
@clarkdo I think both are supposed to run. I added that check there because in SPA/generate mode the call placed in `render()` suffices.

clarkdo:
@galvez Then the change makes sense.

Let's see what's @pi0 's opinion.

Atinux:
I suggest we wait for store RFC before implementing this since we may change the behaviour of it.

husayt (commented):
This was really missing. I will gladly remove my custom nuxtClientInit code.
Thanks @galvez

rchl:
Do I understand correctly that this PR adds `nuxtInit` action that will be called on client-side requests only and the old `nuxtServerInit` will still be called for ssr requests?

If yes, then I have some questions/concerns:
 - Why is there no symmetry in naming? Client vs. Server? I see the name was changed at one point from `nuxtClientInit` to `nuxtInit` but there seems to be no reasoning.
 - Why not have one action only (`nuxtInit`?) and call it both on server and client request, just with different arguments to identify calling context? Or is it the case with this PR? I don't see nuxtServerInit being removed though.
 - If this is for next major version, maybe it's a good time to rethink naming of this API? I think `nuxtServerInit` is a pretty bad name as it can easily be understood as something that is called only once, on nuxt server initializing. Possible alternative would be `onRequest` (if there is one function for both server and client).

manniL:
@rchl Good points! All of this should/will be discussed in an RFC, which is why the PR is on hold.

mrleblanc101:
How long before the RCF.
Can't it be started even if the PR is not ready ?

galvez:
Upon further consideration, I've decided to officially withdraw this PR.

The functionality it introduces is very easily achievable with a simple plugin now.

You just have to use the `.client.js` suffix.

```js
export default (ctx) => {
  window.onNuxtReady(() => {
    // ...
  })
}
```

You have access to `store` in ctx or `$state` if using [`vue-stator`](https://github.com/galvez/vue-stator).

PR #4574, opened 2018-12-17
feat(nuxt-link): Smart prefetching and $nuxt.isOffline

atinux:
The idea is to provide the same features as [quicklink](https://github.com/GoogleChromeLabs/quicklink) **by default** in Nuxt.js.

We use the `IntersectionObserver` if available, otherwise we document it to explain how to use Polyfill (see https://github.com/nuxt/docs/pull/1068). When a `<nuxt-link>` is detected within the viewport, we prefetch the linked page to it (thanks to Webpack code-splitting <3).

A `no-prefetch` prop is also introduced to specify a `<nuxt-link>` to not automatically prefetch the page (it won't be observed as well).

The usage of `destroyed` hook let us free the observer as well as when the page is already prefetched, so the performances are really good. 🚀 

Like quicklink, we check is the user internet connection is fast enough (and online) to request the splitted-page file, we also keep observing the links so when she keeps navigating and her internet connection change, it could preload the link 😄 

It also introduce the way of splitting components for both client & server to optimize bundle size (`nuxt-link.client.js` and `nuxt-link.server.js`), this way, the server component does not rely on the IntersectionObserver. When a use disable this smart prefetching feature globally `router.prefetchLinks: false`, we directly use the `nuxt-link.server.js` component on the client-bundle to reduce the bundle size 🔥 

This PR also introduce `$nuxt.isOnline` and `$nuxt.isOffline` so users can directly rely on it for their Vue components (these props are reactive), a little Christmas gift 🎁.

Demo video:
[Inuxt-prefetch-demo](https://www.youtube.com/watch?v=J6airiY8e84)


clarkdo:
Also a good plugin from @Atinux https://github.com/nuxt-community/guess-module/blob/master/lib/plugin.js#L7

manniL:
We shouldn't forget that preloading happens for scripts and styles by default. https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/config/render.js#L6

packages/vue-app/template/components/nuxt-link.js
```diff
@@ -1,8 +1,34 @@
...
-  render(h, { data, children }) {
-    return h('router-link', data, children)
+  props: {
+    noPrefetch: {
+      type: Boolean,
+      default: false
+    }
+  },
+  mounted() {
+    if (!this.noPrefetch) this.$nextTick(this.prefetch)
```
manniL: I'd suggest to better stick to a curly if here, especially bc we don't have a fully working autocompletion in the template file.
```suggestion
    if (!this.noPrefetch){
      this.$nextTick(this.prefetch)
    }
```

Atinux:
Agree, here we are working only for `prefetch` (which is disabled by default and should be deprecated for) :)

manniL:
@Atinux But isn't that the "wrong way around"? 🤔 

> When should you \<link rel=”preload”> vs \<link rel=”prefetch”>?
>
> Tip: Preload resources you have high-confidence will be used in the current page. Prefetch resources likely to be used for future navigations across multiple navigation boundaries.

(from https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf#89a4)

manniL:
Ideally we would `preload` all linked pages and `prefetch` the rest (if that's wanted). Or am I mistaken there?

Atinux:
That's exactly the behaviour of this :)

This is the `examples/hello-world/` source code with it:

```html
<!DOCTYPE html>
<html data-n-head-ssr data-n-head="">
  <head data-n-head="">
    <title data-n-head="true">Home page</title>
    <link rel="preload" href="/_nuxt/runtime.js" as="script" />
    <link rel="preload" href="/_nuxt/commons.app.js" as="script" />
    <link rel="preload" href="/_nuxt/vendors.app.js" as="script" />
    <link rel="preload" href="/_nuxt/app.js" as="script" />
    <link rel="preload" href="/_nuxt/pages/index.js" as="script" />
    <style data-vue-ssr-id="d71a54be:0">/* ... */</style>
  </head>
  <body data-n-head="">
    <div data-server-rendered="true" id="__nuxt">
      <div id="__layout">
        <div>
          <h1>Welcome!</h1>
          <a href="/about"> About Page </a>
        </div>
      </div>
    </div>
    <script>
      window.__NUXT__ = {
        layout: 'default',
        data: [{}],
        error: null,
        serverRendered: true
      }
    </script>
    <script src="/_nuxt/runtime.js" defer></script>
    <script src="/_nuxt/pages/index.js" defer></script>
    <script src="/_nuxt/commons.app.js" defer></script>
    <script src="/_nuxt/vendors.app.js" defer></script>
    <script src="/_nuxt/app.js" defer></script>
  </body>
</html>
```

As you can see, only the preloaded files are the bundle and current page that is rendered.
Then, when the `nuxt-link` are mounted (after hydration), then we will prefetch the linked pages automatically :)

manniL:
Makes total sense 🙈 Sorry! 😄

aldarund:
We shouldnt prefetch links when when data saving enabled like in quicklinks.
https://github.com/GoogleChromeLabs/quicklink/blob/c5ed3437ddfd5c28aba5816e57a2e3634f09213a/src/prefetch.mjs#L106

aldarund:
Another thing is that not everyone will want this and the only way to disable it is to set noPrefetch to every link. Maybe add option  in config for default value for noPrefetch

pi0:
Benefits of the way of this PR over directly prefetching next SSR rendered page:

- Non-initial page renders are handled by client-side (or at least hydrated). Prefetching SSR content for subsequent pages is likely to introduce more network overheads especially when extractCSS is turned off. If smooth responsiveness during navigation is important, we need to prefetch only needed chunks for related router links. This is done by calling next route matched page chunks. 
- This works for SPA mode too :)

packages/vue-app/template/client.js
```diff
@@ -14,8 +14,13 @@ import {
   compile,
   getQueryDiff,
   globalHandleError
-} from './utils'
-import { createApp, NuxtError } from './index'
+} from './utils.js'
+import { createApp, NuxtError } from './index.js'
```
manniL: Is that change necessary?

packages/vue-app/template/client.js
```diff
@@ -14,8 +14,13 @@ import {
...
+import NuxtLink from './components/nuxt-link.client.js' // should be included after ./index.js
```
manniL: Consistency: Use `.js` extension everywhere or nowhere for import

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+  methods: {
+    observe() {
+      // If no IntersectionObserver, prefetch directly
+      if (!observer) {
+        return this.prefetch()
+      }
+      // Add to observer
+      if (this.shouldPrefetch()) {
+        this.$el.__prefetch = () => this.prefetch()
```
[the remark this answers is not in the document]
pi0: Correcting my comment: According to ECMA, the `bind` method still creates a wrapper **bound function (BF)**. So no perf of non transpiled code. But as babel tries to transpile arrow functions regardless of their usage, this still saves us few bytes: ([repl](https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABMOcAUBKRBvAUIxXAX11wgQGcpEAjAQwCcBGRAXkUzYD5lVNTyYKrUYAmNrzgA6GjDAATTEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=))

image

packages/vue-app/template/client.js
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-import middleware from './middleware'
+import middleware from './middleware.js'
```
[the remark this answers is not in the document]
aldarund: @pi0 I was wondering too, but then checked other imports and they use .js. So it will make it more consistent with other imports. Or change other to be wihtout .js..

packages/vue-app/template/client.js
aldarund: In other files in vue-app it import with .js.

packages/vue-app/template/client.js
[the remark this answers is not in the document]
pi0: @aldarund For valid esm outputs and consistently seems better idea having .js extension in the end of template imports 😊

packages/vue-app/template/components/nuxt-link.server.js
```diff
@@ -0,0 +1,13 @@
+<%= isTest ? '// @vue/component' : '' %>
+import Vue from 'vue'
+
+export default {
+  extends: Vue.component('RouterLink'),
```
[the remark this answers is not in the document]
Atinux: Actually, I don't know how to do better for this. Assigning to a variable would be useless

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,85 @@
+<%= isTest ? '// @vue/component' : '' %>
+import Vue from 'vue'
+
+const requestIdleCallback = window.requestIdleCallback ||
```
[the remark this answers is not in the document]
Atinux: Do you have an example?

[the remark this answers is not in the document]
Atinux: Because it's just a fake polyfill of `window.requestIdleCallback`

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,85 @@
...
+  function (cb) {
```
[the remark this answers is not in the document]
Atinux: Actually it throws an error:
screenshot 2018-12-24 at 15 38 51

Also, since we don't need `this` here it would be useless to have an arrow function :)

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
Atinux: You are right, I thought about it and I remembered that IE < 9 does not support `.bind`, but I think we can drop support for IE < 9 right?

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+  beforeDestroy() {
+    if (this.__observed) {
+      observer.unobserve(this.$el)
+      delete this.$el.__prefetch
+    }
+  },
+  methods: {
+    observe() {
+      // If no IntersectionObserver, prefetch directly
```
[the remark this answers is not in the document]
Atinux: Well, it will make XHR requests that will be handled by the browser, I just need to add a function to handle poor connection to not do such requests :)

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,85 @@
...
+    getPrefetchComponents() {
+      var ref = this.$router.resolve(this.to, this.$route, this.append)
+      const Components = ref.resolved.matched.map((r) => r.components.default)
+
+      return Components.filter((Component) => typeof Component === 'function' && !Component.options && !Component.__prefetched)
+    },
+    prefetch() {
+      const Components = this.getPrefetchComponents()
+
+      if (!Components.length) {
```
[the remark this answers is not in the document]
Atinux: You are right :)

packages/vue-app/template/client.js
Atinux: I do believe we will benefit by giving extension name for faster path resolution for the loader.

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -53,31 +50,41 @@ export default {
...
       }
     },
     shouldPrefetch() {
       return this.getPrefetchComponents().length > 0
     },
+    canPrefetch() {
+      const conn = navigator.connection
+
+      // Don't prefetch if the user is on 2G. or if Save-Data is enabled..
+      if (conn && ((conn.effectiveType || '').includes('2g') || conn.saveData)) {
```
manniL: You can directly us the expression as return value

```js
const hasBadConnection = conn && ((conn.effectiveType || '').includes('2g') || conn.saveData)

return !hasBadConntection
```

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
pi0: Yes. PS if babel targetted to IE < 9, which not supported by default nuxt preset or observer, we have [babel-plugin-transform-function-bind](https://babeljs.io/docs/en/babel-plugin-transform-function-bind) for this transform :)

packages/vue-app/template/client.js
pi0: If someday we want to directly serve templates in the browser (script with module type), `.js` is required. PS more explicit is always better and already has consistency benefit for us :)

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
pi0: No just suggesting to replace it with something like `window.NuxtrequestIdleCallback` indicating it is a nuxt related callback :D

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
pi0: Still, they are normal XHR requests with the same priority as normal API requests. We can't also at least with current webpack API prioritize its XHR requests. So at least we can try to reduce them.

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
Atinux: Since it's a pretty poor polyfill and it's used only in this particular area, I suggest to not add this function to the globals and keep it local (at the moment, and could be moved to the utils)

packages/vue-app/template/components/nuxt-link.client.js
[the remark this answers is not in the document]
Atinux: I removed the call finally, the will use a polyfill (added to the docs) if they want to support this feature :)

Atinux:
Ready for final review :)

Docs PR: https://github.com/nuxt/docs/pull/1068

pi0 (dismissed):

galvez (dismissed):

galvez:
Man, this is a seriously badass PR -- many kudos @Atinux

packages/vue-app/template/App.js
```diff
@@ -65,8 +66,12 @@ export default {
     // Add this.$nuxt in child instances
     Vue.prototype.<%= globals.nuxt %> = this
     // add to window so we can listen when ready
-    if (typeof window !== 'undefined') {
+    if (process.client && typeof window !== 'undefined') {
```
aldarund: Why double check here?

packages/vue-app/template/App.js
```diff
@@ -79,7 +84,24 @@ export default {
     'nuxt.err': 'errorChanged'
   },
   <% } %>
+  computed: {
+    isOffline() {
+      return !this.isOnline
+    }
```
kevinmarrec: If they can be used from everywhere through `this.$nuxt.isOffline` and `this.$nuxt.isOnline`, I suggest adding these new type definitions of the `$nuxt` object :
```ts
isOffline: boolean;
isOnline: boolean;
```
here https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/types/vue.d.ts#L29

As this PR is related to the `vue-app` package, you can just add them in a new commit I think :)

I'm not sure if they should be placed before or after `$loading` though

kevinmarrec (changes requested):

packages/vue-app/template/App.js
galvez: @Atinux look here ^

packages/vue-app/template/App.js
Atinux: Fixed, thanks @aldarund

packages/vue-app/template/App.js
Atinux: Added, thanks @kevinmarrec 

I added them before since $loading is added on `mounted()` hook and also is not available if `loading: false` is defined in `nuxt.config.js`

benoitemile:
Hello there, appreciate that work on integration of quicklink directly in nuxt.
I have a question though, what if one uses Guessjs-nuxt module already ?

packages/vue-app/types/vue.d.ts
```diff
@@ -25,6 +25,8 @@ declare module "vue/types/options" {
 declare module "vue/types/vue" {
   interface Vue {
     $nuxt: {
+      $isOnline: boolean;
```
kevinmarrec: `$` prefix is not needed
It's `$nuxt.$loading`, `$nuxt.isOffline` and `$nuxt.isOnline` unless you rename them with the prefix in your changes 😜

And I suggest puting offline before online to keep alphabetical order ;)

kevinmarrec (changes requested):

kevinmarrec:
@Atinux Seems that with your work, people can easily know the online status of their app through `this.$nuxt.isOnline`, right ? What about referencing that in the documentation ?

packages/vue-app/types/vue.d.ts
Atinux: Agree for alaphabetical order.
About $loading, it’s because it’s a ref of the Nuxt-loading instance exposing methods :)

kevinmarrec: You can update the typedefs with that : https://github.com/nuxt/nuxt.js/compare/dev...kevinmarrec:nuxt-typedef-example

=> This way people will have the right autocompletion on `$nuxt` including all Vue instance properties/methods

Atinux: You are awesome.

packages/vue-app/types/vue.d.ts
```diff
@@ -5,7 +5,7 @@
 import Vue, { ComponentOptions } from "vue";
 import { Route } from "vue-router";
 import { MetaInfo } from "vue-meta";
-import { Context, Middleware, Transition, LoadingObject } from "./index";
+import { Context, Middleware, Transition, LoadingObject, NuxtApp } from "./index";
```
kevinmarrec: `LoadingObject` import no longer needed 🙃

packages/vue-app/types/index.d.ts
```diff
@@ -63,3 +63,12 @@ export interface LoadingObject {
   start(): void;
```
kevinmarrec: These typedefs are no longer needed, they have been moved directly in `NuxtApp` interface 😉

kevinmarrec (changes requested):
`LoadingObject` related stuff can be safely removed, type definition of `$loading` is directly declared in the `NuxtApp` interface.

😇

packages/vue-app/types/index.d.ts
kevinmarrec: https://github.com/nuxt/nuxt.js/blob/78580a6fa73f065c775f0dbb09a27c7d6e2ec201/packages/vue-app/types/index.d.ts#L62

packages/vue-app/types/vue.d.ts
```diff
@@ -24,10 +24,6 @@ declare module "vue/types/options" {
 
 declare module "vue/types/vue" {
   interface Vue {
-    $nuxt: {
-      $isOnline: boolean;
-      $isOffline: boolean;
-      $loading: LoadingObject;
-    };
+    $nuxt: NuxtApp
```
kevinmarrec: And sorry here, I forgot a semicolon in the first commit of my proposal changes, so we need to add it 😁

PS : I may propose to remove semicolon and double quotes everywhere in typedefs later, but lets keep the code style already used for now

kevinmarrec (changes requested):

packages/vue-app/types/vue.d.ts
kevinmarrec: https://github.com/kevinmarrec/nuxt.js/commit/caa3efd1763706977c98090c71c9d4994827547d 🙈

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -0,0 +1,89 @@
...
+        didTimeout: false,
+        timeRemaining: function () {
+          return Math.max(0, 50 - (Date.now() - start))
+        },
+      })
+    }, 1)
+  }
+const observer = window.IntersectionObserver && new window.IntersectionObserver(entries => {
+  entries.forEach(({ isIntersecting, target: link }) => {
+    if (!isIntersecting) {
```
maoberlehner: This could be changed to use `intersectionRatio` instead, for maximum browser support (https://github.com/w3c/IntersectionObserver/issues/211).

```js
if (intersectionRatio <= 0) {
```

Atinux:
@benoitemile exactly, it will need to globally disable this feature if he only want the guess.js logic :)

addyosmani:
Excited to see quicklink's providing some inspiration here :) 

I think this discussion may have already happened, but in general you want to preload resources needed for the current navigation/page (e.g what static resources needs to be fetched with high priority for first render or for the page to be interactive). 

You prefetch (at low priority) pages for future navigations. `rel=prefetch` performs low priority prefetches which is why it's well suited for that behaviour. 

Is that how we've approached this PR? :)

packages/vue-app/types/vue.d.ts
hartmut-co-uk: 👻

Atinux:
Thank you for your comment @addyosmani 

You can watch a demo video of this feature:

[Inuxt-prefetch-demo](https://www.youtube.com/watch?v=J6airiY8e84)

Actually, Nuxt already add the `<link rel="preload">` on SSR for critical resources.
When the application is mounted on client-side, we wait for `requestIdleCallback`, then we check if the linked page is not already prefetched before starting to observe the link.

When the link is intersecting, we check if the user has a decent internet connection (online + `4g` for  `effectiveType` + not in data saving mode), then we call Webpack lazy import feature.

Sadly, at the moment, Webpack does not support low priority request for lazy import (cc @sokra @TheLarkInn).
We will work on a solution later on to prefetch the links in the background the service workers are available.

packages/vue-app/template/components/nuxt-link.client.js
Atinux: Updated, thanks @maoberlehner :)

galvez (dismissed):

kevinmarrec (dismissed):

benoitemile (dismissed):

kevinmarrec (approved):

tmorehouse:
Just wondering about action links... i.e. ones that may make state changes on the back end server (ie. disabling a user, deleting an item from a database, etc).  Could this inadvertently do things that it shouldn't?

tmorehouse:
@manniL yeah, I know, but there might be a few users who blindly will use `<nuxt-link>` for all links (as they may want to render a page with a result of the action, and don't do any confirmation checks beforehand, or may have an `@click` handler and an 'alert' prompt).

manniL:
@tmorehouse Even then it shouldn't execute the action as just the js file for the page is loaded and not the page itself

tmorehouse:
@manniL Ah yeah, that is true. :smile:

PR #4582, opened 2018-12-18
feat: add store module HMR

### LINKED ISSUE #325: HMR for vuex store
I thought using modules mode will automatically add in the HMR code but found out that it is not. 

https://vuex.vuejs.org/en/hot-reload.html

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt/nuxt.js/issues/c276">#c276</a>)</em></sub></div>

TheAlexLichter:
## Types of changes

## Description

Feedback and Refactoring highly appreciated! It's just a rudimentary version but it works. Similar to normal Vuex module HMR, the state changes are not hot reloaded but getters, actions and mutations are.

I had to export the store *somehow* to trigger the HMR (similar to https://vuex.vuejs.org/guide/hot-reload.html or https://github.com/chrisvfritz/vue-enterprise-boilerplate/blob/master/src/state/modules/index.js) as we export a function so the store can be generated from a new SSR context all the time.

Resolves #325 

## Checklist:


packages/vue-app/template/store.js
```diff
@@ -3,73 +3,86 @@ import Vuex from 'vuex'
...
       }
-    }
 
-    module[name] = Object.assign({}, module[name], fileModule, appendedMods)
-    module[name].namespaced = true
+      module[name] = Object.assign({}, module[name], fileModule, appendedMods)
+      module[name].namespaced = true
+    }
+    // If the environment supports hot reloading...
+    if (module.hot) {
```
manniL: We could guard that code with a lodash template and check for prod to remove the extra bytes I guess 🤔

Atinux: Agree, removing dead code is smart, I don't know if webpack will automatically remove it, but we can use lodash template to directly remove this part :)

Same for the server, we could also use `if (process.client && module.hot)` around a `<% if (isDev) {%>`

atinux (changes requested):
Great work <3

galvez (dismissed):
So awesome @manniL! -- so is this mimicking that bit from vue-enterprise you linked to?

manniL:
@galvez Yup, basically I implemented the same approach that @chrisvfritz took but with our current functions ☺️

chrisvfritz:
Glad to have helped indirectly! I think I originally stole the nested module registration strategy from Nuxt. 😄

manniL:
Hehe, awesome! 🙊 That's how we all benefit from open source ☺️

manniL:
Okay, looks like all initial problems are solved ☺️

packages/vue-app/template/store.js
```diff
@@ -3,73 +3,87 @@ import Vuex from 'vuex'
...
 let storeData = {}
 
-// Check if {dir.store}/index.js exists
-const indexFilename = filenames.find(filename => filename.includes('./index.'))
+let files;
 
-if (indexFilename) {
-  storeData = getModule(indexFilename)
-}
+(function updateModules() {
```
galvez: @manniL quick tip, it's possible to write:

```js
void function() {

}()
```

manniL: Thanks, implemented ☺️

atinux (approved):

PR #4585, opened 2018-12-19
fix: prevent matching native statements in serializeFunction

galvez:
As reported in https://github.com/nuxt/nuxt.js/commit/25d26ba42f97fa738592df4f85e692a557ffa1bf#r31725441, `serializeFunction()` introduce a bug where it's matching native JS statements.


TheAlexLichter (approved):

pi0 (commented):
As a better solution, I suggest just doing replace on the first line of the serialized function. I think the purpose of function is this right?

pi0:
Merging as a hotfix. Improvement can be addressed later.

PR #4592, opened 2018-12-19
feat: mode for plugin

clarkdo:
## Types of changes

## Description

The idea is from @Atinux , use `.server.js` or `.client.js` to automatically detect server/client side plugin.

1. introduce `mode: 'client|server'` in plugin options
2. deprecate `ssr` in plugin options and use `mode: 'client|server'` instead, for now auto adapt `ssr: false` to `mode: 'client'` and show deprecating log.
3. apply `server/client` plugin on file ext like: `server.js` or `client.js` (**specified options has higher priority**)

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -114,7 +114,6 @@ export default class Builder {
         // TODO: remove deprecated check in Nuxt 3
         if (p.ssr === false) {
           p.mode = 'client'
-          consola.warn(`ssr in plugin options has been deprecated, please use mode instead`)
         } else if (!['client', 'server'].includes(p.mode)) {
           p.mode = 'all'
         }
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -116,7 +116,7 @@ export default class Builder {
         } else if (p.mode === undefined) {
           p.mode = 'all'
         } else if (!['client', 'server'].includes(p.mode)) {
-          consola.warn(`Invalid plugin mode (server/client/all): '${p.mode}'. Falling back to 'all'.`)
+          consola.warn(`Invalid plugin mode (server/client/all): '${p.mode}'. Falling back to 'all'`)
           p.mode = 'all'
         }
 
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -111,13 +111,12 @@ export default class Builder {
           ''
         )
 
-        // TODO: remove deprecated check in Nuxt 3
         if (p.ssr === false) {
           p.mode = 'client'
         } else if (p.mode === undefined) {
           p.mode = 'all'
         } else if (!['client', 'server'].includes(p.mode)) {
-          consola.warn(`Invalid plugin mode (server/client): ${p.mode}. Falling back to default`)
+          consola.warn(`Invalid plugin mode (server/client/all): '${p.mode}'. Falling back to 'all'.`)
           p.mode = 'all'
         }
 
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -110,9 +110,18 @@ export default class Builder {
           /[^a-zA-Z?\d\s:]/g,
           ''
         )
+
+        // TODO: remove deprecated check in Nuxt 3
+        if (p.ssr === false) {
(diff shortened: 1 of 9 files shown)


packages/builder/src/builder.js
```diff
@@ -110,9 +110,18 @@ export default class Builder {
           /[^a-zA-Z?\d\s:]/g,
           ''
         )
+
+        // TODO: remove deprecated check in Nuxt 3
+        if (p.ssr === false) {
+          p.mode = 'client'
+          consola.warn(`ssr in plugin options has been deprecated, please use mode instead`)
```
pi0: This warning could be really annoying as modules should still prefer using it for 2.x compat. I suggest doing warn on N3 upwards.

packages/builder/src/builder.js
```diff
@@ -224,6 +240,8 @@ export default class Builder {
     // Plugins
     this.plugins = Array.from(this.normalizePlugins())
 
+    await this.resolvePlugins()
```
pi0: Why we should resolve plugins after build?

packages/builder/src/builder.js
```diff
@@ -110,9 +110,18 @@ export default class Builder {
           /[^a-zA-Z?\d\s:]/g,
           ''
         )
+
+        // TODO: remove deprecated check in Nuxt 3
+        if (p.ssr === false) {
+          p.mode = 'client'
+          consola.warn(`ssr in plugin options has been deprecated, please use mode instead`)
+        } else if (!['client', 'server'].includes(p.mode)) {
+          p.mode = 'all'
```
pi0: A warning about invalid mode can be useful here. Users may accidentally set things like `clien`, ... that we silently translate it to `all`.

clarkdo: `all` is not a options that user need to use, it's just a default condition for both client and server.

`client/server` are the only valid value that should be configured in `mode`, if we want to add warning here, we need to another `else if` for `p.mode === undefined`, if it's ok for you, I can add.

pi0: True, `all` is not what people should use by default. But invalid values (typos) are silently ignored.

clarkdo: OK, I'll add the message.

BTW, this can be also added in eslint-plguin

manniL: :+1: for that idea

pi0: Yes. Supporting in ESLint plugin would be awesome.) But we also need to have proper CLI warns too.

clarkdo: Warn msg added

pi0 (approved):

clarkdo:
Will add a pr for doc

Atinux:
This is awesome @clarkdo

CurtisBelt:
@clarkdo Hello, I am using nuxt-edge while project is in development, and after updating to a more recent commit, none of my plugins load when using `nuxt start`, but they work fine when using `nuxt dev`.

I did some trial and error, moving backwards through available `nuxt-edge` versions, and determined that the problem starts after this feature was merged. The issue still exists as of `2.4.0-25768669.8786ff73` (8786ff731767425bb0c1c72af6d8a8895155acf5).

I did not create a reduced test since I am curious if you will simply reproduce bug right away when using `nuxt start`. If you are unable to reproduce, I am happy to do more research. Thank you!

clarkdo:
@CurtisBelt I’ll look into it, thank you for the reporting 🙏

clarkdo:
Hi @CurtisBelt Is your plugin specified ssr:false? If yes, it won’t be loaded in server side

manniL:
@clarkdo I experience the same on latest nuxt edge for https://github.com/Developmint/developmint.de

manniL:
@clarkdo You need `modern: 'server'` set to trigger the error.

nuxt.config.js
```js
  GNU nano 2.9.3                    nuxt.config.js                              

module.exports = {
  modern: 'server',
  plugins: [
{ src:    '~/plugins/test', ssr: false}
  ]
}
```

plugins/test.js
```js
window.test = true
```

manniL:
Update: fix @ #4659

PR #4600, opened 2018-12-20
feat: better stack traces for SSR error, show error with correct URL and use eventsource-polyfill

atinux:
These are fixes to improve DX when using Nuxt:
- When an error occured during navigation (error in asyncData, middleware , ...), we won't stop the navigation with `next(false)`, this way, the error is on the correct url, and we don't even display the next page since `<nuxt>` component does not render it if an error is present.
- Added eventsource-polyfill in dev mode to handle IE and old browsers that don't support it
- Overwrite errorFull.stack if `stack` is present in original error to at least display better stack traces on `youch`

Please review :)


TheAlexLichter (dismissed):

aldarund (dismissed):

pi0 (dismissed):

pi0:
Azure is randomly failing. Triggered rebuild.

aldarund:
Locally on win machine it dont fail

pi0:
Damn still failing! :))

pi0 (approved):

PR #4609, opened 2018-12-21
refactor: @nuxt/utils

clarkdo:
- rename @nuxt/common to @nuxt/utils
- move `hookable` and `modern-browsers` to corresponding module
- split utils.js into small and single responsibility files


Atinux:
Why not having @nuxt/utils @clarkdo ?

So we could have utils for end-users and us and @nuxt/common for Nuxt internals mostly?

For instance, `waitFor`, `isString`, 'sequence', 'parallel' could be in `@nuxt/utils` directly.

Also, to give more context about `@nuxt/utils`: https://github.com/mono-js/mono-utils

pi0 (approved):

PR #4527, opened 2018-12-11
feat: add HardSourceWebpackPlugin

TheAlexLichter:
Adds experimental support for https://github.com/mzgoddard/hard-source-webpack-plugin (disabled by default)

## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Resolves https://github.com/nuxt/nuxt.js/issues/4107

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

PR #4616, opened 2018-12-22
feat: upgrade to webpack5

pi0:
## Changelog and Links

-  https://github.com/webpack/changelog-v5/blob/master/README.md
- https://github.com/webpack/webpack/issues/8537

## Missing / Delegated

- Arrow syntax is not transpiled for all modes
- ExtractCSS disabled
- Deprecate warnings (for 3rd party WP plugins)
- Font preload test 
- Module Federation investigation
- `@nuxt/components`

## Checklist

- [x] Pin `webpack` version to v5-
- [x] Pin `html-webpack-plugin` to v4-
  - [x] replace `dependency` chunk sort mode by `auto`
- [x] `optimization.splitChunks name: true` removed: Automatic names are no longer supported
- [x] Prefer using `chunkIds: "named"` and `idHint` (default value in development mode is `named`)
- [x] `compilation.hooks.normalModuleLoader` was moved to `NormalModule.getCompilationHooks(compilation).loader`
  - [x] `vue-loader/lib/plugin.js` PR: https://github.com/vuejs/vue-loader/pull/1469
  - [x] `extract-css-chunks-webpack-plugin` (faceyspacey/extract-css-chunks-webpack-plugin#135)
- [x] `[hash]` is now `[fullhash]`
   - [x] Consider using `[chunkhash]` or `[contenthash]`
   - [x] nuxt webpack
   - [x] loader-utils/lib/interpolateName.js
   - [x] webpack/lib/EvalSourceMapDevToolModuleTemplatePlugin.js
- [ ] `mini-css-extract-plugin` and `extract-css-chunks-webpack-plugin` are incompatible (RuntimeModules change) (MainTemplate.hooks.requireEnsure removed)
- [x] `fork-ts-checker-webpack-plugin` is incompatible (frozen hooks)
- [x] `cache-loader` is no longer needed
- [x] `hard-source-webpack-plugin` is obsolete
- [ ] `webpack-node-externals` PR: https://github.com/liady/webpack-node-externals/pull/59 Mirgrated to local for now
- [x] `webpack-dev-middleware` PR: https://github.com/webpack/webpack-dev-middleware/pull/459
- [x] `last-call-webpack-plugin`: dependency of `optimize-css-assets-webpack-plugin`
- [ ] ensure devtool is always matching `/^(inline-|hidden-|eval-)?(nosources-)?(cheap-(module-)?)?source-map$/`
- [x] Update AsyncMFS to use `memfs`


pi0:
Tip: Trace deprecation warnings via `node -r esm --trace-warnings --trace-deprecation ./node_modules/.bin/nuxt examples/hello-world`

aldarund:
too early :)

manniL:
@pi0 I guess `HardSourceWebpackPlugin` is obsolete then as well? (PR #4527)

ScriptedAlchemy:
Appreciate the work you are doing on Extract CSS Chunks!

ScriptedAlchemy (commented):
How’s the stability? And its in house css loading. Does css or sass hot reload?

packages/webpack/src/config/base.js
```diff
@@ -127,6 +126,18 @@ export default class WebpackBaseConfig {
     }
   }
 
+  cache() {
+    if (this.options.build.cache) {
```
manniL: Maybe it's worth inverting this condition here and use an early return? :)

clarkdo: Yes, that will be better

ScriptedAlchemy:
For the removal of requireEnsure - this looks like what it became. https://github.com/webpack/webpack/blob/master/lib/async-modules/AwaitDependenciesInitFragment.js

MainTemplate will be removed in v6

ScriptedAlchemy:
Also, I wrote this feature into Webpack 5 called Module Federation @manniL - I’ve rebuilt next.js in order to implement this, since this project is the Vue equivalent, and likely face the “how do we do micro-frontends, with shared resources and no page reloads, along with how do we to the same in SSR as well?” 

This might be of use: https://link.medium.com/i7cpPsXJ64

clarkdo (dismissed):

packages/webpack/src/plugins/wp5-fix.js
```diff
@@ -0,0 +1,12 @@
+module.exports = class WP5Fix {
```
clarkdo, line 1: Can be removed?

pi0: Yes actually it was to hide deprecated warnings but we should indeed try to update any problematic 3rd party plugin instead of hiding :) Will remove

packages/webpack/src/config/client.js
```diff
@@ -74,7 +74,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
         test: /node_modules[\\/](vue|vue-loader|vue-router|vuex|vue-meta|core-js|@babel\/runtime|axios|webpack|setimmediate|timers-browserify|process|regenerator-runtime|cookie|js-cookie|is-buffer|dotprop|nuxt\.js)[\\/]/,
         chunks: 'all',
         priority: 10,
-        name: true
+        name: 'commons'
```
clarkdo, line 77: Will this affect  hash name in prod?

pi0: >  optimization.splitChunks name: true removed: Automatic names are no longer supported

I think was related to this

clarkdo: WP5 should give name by default, what’s the name on dev/prod if we remove name: commons

pi0: image

@clarkdo seems ok. It is like a hint

clarkdo: Great 👍

pi0: For reference: Removing explicit name, make it hard for bundle analyzer:

image

packages/webpack/src/plugins/wp5-fix.js
pi0: Done

ScriptedAlchemy:
I’m thrilled to see you’ve added module Federation investigation!
Get hold of me when you start looking into it. I’m supporting angular’s core team as they upgrade to module federation. 

Happy to help or provide information

clarkdo:
@ScriptedAlchemy Thanks, it would be great that we can get your help when we start.

ScriptedAlchemy:
Get me on twitter and we can set up a zoom call. We should be cutting beta.17 which triples the capabilities of module federation 

If you’ve got time to waste. Here’s a intro to advance api now in master. This only covers shared and sharing. There’s at least 2 more videos that need to be made for the other things we did. https://youtu.be/9F7GoDMrzeQ

PR #4645, opened 2018-12-28
fix: nuxt-start cli error

clarkdo:
## Types of changes

## Description

Issue is reported from @aldarund 
Bundle `vue-router`, `vue-no-ssr` and `vuex` into server bundle for fixing global nuxt-start usage.

## Checklist:

### DIFF
--- a/packages/webpack/src/config/server.js
+++ b/packages/webpack/src/config/server.js
@@ -17,7 +17,8 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
   normalizeWhitelist() {
     const whitelist = [
       /\.css$/,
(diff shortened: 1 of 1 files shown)


clarkdo:
The current problem is that if we partially bundle libs into `server.js`, it may lead to duplicate require cache in server side.

For exmaple:
```js
// app/index.js
import Router from 'vue-router' // this vue-router is from server.js

// node_modules/some-3rd-lib
import Router from 'vue-router' // this vue-router is from node_modules/vue-router
```

So the properties registered on `vue-router` from `server.js` are not on `vue-router` from `node_modules`

pi0 (approved):

PR #4658, opened 2018-12-30
feat: nuxt-ts

pi0:
## Types of changes

## Description
This PR adds a new distribution package `nuxt-ts`. 

The basic idea is first class support for typescript with just a single package (Aka, the Nuxt way 🔥 )

(Original credits for the idea is for @Atinux and @kevinmarrec)

- [x] Add `nuxt-ts`  command -- a wrapper around `nuxt` and (currently) globally registers `node-ts` for `.ts` files support
- [x] Add all dependencies for typescript development. Using this package sets up everything for us
- [x] `nuxt-ts` npm package is taken! Find another name.
- [ ] Combine and bundle all type declarations with the distribution package.
- [x] `nuxt.config.ts` 
- [x] `serverMiddleware`
- [x] `.ts` modules (only for testing purposes. not recommended/support plan for node_modules. As they should be compiled before publish for perfomance)
- [x] Fix tests by adding ts support to jest configuration

## Checklist:


manniL:
Regarding the pkg name, we could also ask @hmsk (https://github.com/hmsk/nuxt-ts)

kevinmarrec:
@manniL True, if we tell him that official TypeScript support is coming he might let us the pkg name

Atinux:
The best would be to supports also what the actual `nuxt-ts` supports by adding @nuxt/typescript module with this distribution (as dependency and module inclusion by default)?

So Nuxt TS is full TS support for server & the vue app 🚀

pi0:
Hey @hmsk. Thanks for your response. 

Actually, we tried to include most of the TS functionality right into the core. (Supporting *.ts, fork-ts-checker, etc). So everything magically works if users of nuxt 2.4+ : 1) install all required extra dependencies 2) register node-ts. We can use `nuxt-ts` as a shortcut to do these steps automatically too. 

We can bump `nuxt-ts` version to `2.4.0` after the changes, so it does not affect its current users (0.x) :)

PS: I've sent you an [invitation](https://github.com/nuxt-community) for nuxt-community, so you can help to maintain ts template if interested :)

hmsk:
Cool, okay let me try with the edge (maybe this weekend). Then, I'm going to pass the name if you show about my original repo for the time being 🗿 

And thanks for the invitation. It'll be great if you give some intro about the organization (process, do/dont...etc) 😉

hmsk:
Finally, I added @pi0 and @Atinux to the owners of `nuxt-ts`.

Before that, I tried `nuxt-ts-edge` in some of my projects and worked well. Really great work all!
I can't wait to see 2.4 is released ✨

PR #4661, opened 2018-12-30
feat(webpack,cli): standalone build mode

pi0:
## Types of changes

## Description

This PR implements @clarkdo idea (https://github.com/nuxt/nuxt.js/pull/4645#issuecomment-450365285) for standalone build mode and resolves #4645.

The `build.standalone` option is turned off by default to prevent any behaviour/breaking changes.

## Benchmarks

- Tested on bare-bone hello-world
- Cold start time is measured using the production build of nuxt. From reading config until `nuxt.ready()`
- Built on windows (Yeah I'm using windows again :P)

Normal build: 
- Size of `server.js`: `22 Kib`
- Cost of dependencies (`vue`, `vue-router`,  `lodash` (/omit), `vue-no-ssr`, `debug`): `7.4M`
- Cold start: `306.385ms`
- Memory usage: `29.9` MB (RSS: `106 MB`) 

Standalone build:
- Size of `server.js`:  `145 Kib`
- cold start: `306.550ms`
- Memory usage: `28.6 MB` (RSS: `106 MB`)

Results prove that there should be no big runtime diffs with the standalone mode. Still, we need to benchmark nuxt against real-world projects to decide on the default value.

## Remarks

- This is strange what dependency adds `lodash/omit` to the server bundle. Standalone build tree shakes only needed parts but should be still addressed
- We need to announce users to use `nuxt build --standalone` when they need to use it with `nuxt-start`. 
  - Alternative: For now we can add dependencies again to the `nuxt-start` distribution but it makes package bigger again

## Checklist:

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -33,7 +33,7 @@ export default {
     generate: {
       type: 'boolean',
       default: true,
-      description: 'Don\'t generate static version for SPA mode (useful for nuxt start)'
+      description: 'Don\'t generate static version for SPA mode (useful for nuxt-start)'
     },
     quiet: {
       alias: 'q',
@@ -44,6 +44,16 @@ export default {
         options.build = options.build || {}
         if (argv.quiet) {
           options.build.quiet = !!argv.quiet
+        }
+      }
+    },
+    standalone: {
+      type: 'boolean',
+      default: false,
+      description: 'Bundle all server dependencies (useful for nuxt-start)',
+      prepare(cmd, options, argv) {
+        if (argv.standalone) {
+          options.build.standalone = true
         }
       }
     }
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -33,7 +33,7 @@ export default {
     generate: {
       type: 'boolean',
       default: true,
-      description: 'Don\'t generate static version for SPA mode (useful for nuxt-start)'
+      description: 'Don\'t generate static version for SPA mode (useful for nuxt start)'
     },
     quiet: {
       alias: 'q',
--- a/packages/config/src/config/build.js
+++ b/packages/config/src/config/build.js
@@ -10,6 +10,7 @@ export default () => ({
   ssr: undefined,
   parallel: false,
   cache: false,
+  standalone: false,
   publicPath: '/_nuxt/',
   filenames: {
     // { isDev, isClient, isServer }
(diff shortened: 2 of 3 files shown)


clarkdo:
In standalone mode, webpack will bundle all deps, but if 3rd party lib is using dynamic require, it may throw module not found error.

For example: [`consola` requires `winston`](https://github.com/nuxt/consola/blob/master/src/reporters/winston.js#L9), but `winston` is `devDependency`, so webpack will throw error when bundle `winston` into server bundle

pi0:
@clarkdo Yes. That's a known problem of webpack with Node.js libraries like consola. We can whitelist some in standalone mode. But i don't recommend because it is dangerous in terms of a standalone build.

aldarund:
What about when someone dont want ( or just updated nuxt from 2.3 to 2.4 and it will break) to use standalone mode, but rather just use nuxt start as solo production dependency. Then there still will be problem from #4645.

pi0:
@aldarund I just mentioned this in PR body as an alternative option:

> Alternative: For now we can add dependencies again to the nuxt-start distribution but it makes package bigger again

But I assume it is a soft breaking change. We can make `nuxt-start` smarter to detect non-standalone builds and warn users.

aldarund (dismissed):

galvez (dismissed):
This is absolutely great.

atinux (dismissed):

pi0:
Updates:

- Dependencies for alternative approach added right into `nuxt-start` package (c664e3df7b2175d23e88e8c2df95c85ce005fcb1, https://github.com/nuxt/nuxt.js/commit/b5a0f788dc01ce60403835ffb59fa73978c21179)

- The mentioned extra import for lodash removed from server.js. (https://github.com/nuxt/nuxt.js/commit/6178c470ba81367e0c9f856b38781d270d65edfe). Reducing hello world's server size to `124 KiB` (Saving 21Kib)

manniL:
We need docs for that, right? 🤔

pi0:
@manniL please add todo but it is experimental for now :)

Norserium:
@pi0, could you describe the current status of this future? It looks undocumented now.

pi0:
Hi @Norserium it is indeed stable and being used for vercel serverless builder for a long time. I think there should be an open issue for documenting /cc @debs-obrien

Norserium:
@pi0, yes i[t's](https://github.com/nuxt/nuxtjs.org/issues/499). But it's defined as an experimental mode.

pi0:
That's why issue reopened ;)

mercs600:
@pi0 how should I work with standalone option when modules defined in nuxt.config.js are required on demand ? Do we have some "whitelist" do tell the builder we want to use these specific module in standalone mode ?

pi0:
@mercs600 You can instead move build-only modules to `buildModules` and list runtime modules in `modules` :)

mercs600:
right! I forgot about buildModules, thanks ✋

PR #4668, opened 2019-01-02
fix: Vue.component('RouterLink') is undefined in vue-router 3.0.0

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/vue-app/template/router.js
+++ b/packages/vue-app/template/router.js
@@ -81,6 +81,12 @@ const _routes = recursiveRoutes(router.routes, '  ', _components, 2)
 }).join('\n')%>
 
 Vue.use(Router)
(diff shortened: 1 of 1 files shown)


manniL:
But they are installed as `PascalCase` versions :thinking: 
(see https://github.com/vuejs/vue-router/blob/dev/src/install.js#L46-L47)

clarkdo:
@manniL It's not in 3.0.0 https://github.com/vuejs/vue-router/blob/v3.0.0/src/install.js 😄

manniL:
@clarkdo Omg :see_no_evil: 
Thanks, you are right :sweat_smile:

aldarund (approved):

pi0:
If this problem is fixed with `3.0.2`, should we still add this patch?

Issue #2406, opened 2017-12-16
How to update vue-template-compiler


feedgit:
I installed Nuxt and this current version of **vue-template-compiler** package is `vue-template-compiler@2.4.4`. But in this file: https://github.com/nuxt/nuxt.js/blob/dev/package.json is version 2.5.11. When I update using npm install it came to error mismatch with `vue@2.4.4` too. I cant update two packages. Please help me!!

clarkdo:
Nuxt with `Vue 2.5.11` still don't release yet (only available in `next` which is an unstable version), if you are using `1.0.0-rc11` and there are other libs using `2.5.11`, you should specify to use `vue@2.4.4`, `vue-template-compiler@2.4.4` and `vue-server-renderer@2.4.4` manually.

Issue #2079, opened 2017-11-10
Vue plugins with generic vue versions causing issues


robyedlin:
Some Vue plugins are causing version mismatch issues with vue, vue-template-compiler, and vue-server-renderer

e.g. https://github.com/Inndy/vue-clipboard2/blob/master/package.json#L28

Is there a recommended way to guarantee the Nuxt version of these packages without having to manually specify the vue versions in `package.json`?

clarkdo:
you may add `vue` into dependencies manually.
```
dependencies: {
  "vue": "~2.4.4"
}
```

Nuxt.js will bump `vue` to `2.5.x`, this issue will be resolved.

manico:
> Nuxt.js will bump vue to 2.5.x, this issue will be resolved.

Do we know when will this happen?

clarkdo:
@manico 
We are working on it, about the details, pls refer: https://github.com/nuxt/nuxt.js/issues/2142

Issue #1851, opened 2017-10-13
Update to Vue v2.5.0+


hirokiosame:
Vue 2.5.1 was released last night. Can the dependency in NUXT be updated?

hojas:
And Vue 2.5.2 was released.

ooade:
Can we just add @greenkeeper to keep our packages updated

Atinux:
Actually a new version of Vue can bring some breaking changes, that’s why we want to make sure it’s fully compatible before releasing a new version.

Next release coming will have Vue 2.5!

hirokiosame:
@Atinux 
Yes, I fully expect thorough testing to come with it.

We were waiting on the "default" feature in provide/inject to be released in 2.5.0+, but currently blocked by NUXT from upgrading; don't mean to rush any work but do you have a ballpark date?

JamieLottering:
If you can't wait you can manually update all the dependencies using the "resolutions" field.
In your package.json:

```
"resolutions": {
  "nuxt/vue": '^2.5.2"
}
```

You'll most likely get a warning that vue has a version mismatch so you'll have to add all of those updated dependencies into the resolutions field until it works. This is also assuming you're using `yarn` over `npm install`

hirokiosame:
@JamieLottering Are you sure that's a real property? It's [not documented](https://docs.npmjs.com/files/package.json) and it didnt work

JamieLottering:
@hirokiosame That's because it's not an NPM feature, it's a yarn feature

https://yarnpkg.com/en/docs/selective-version-resolutions

hirokiosame:
@JamieLottering Oh sorry I overlooked that. Thank you!

dohomi:
@Atinux any chance to see `rc-12` sometimes soon to make use of the `vue 2.5` features? Thanks!

cliffrowley:
Just FYI if you want to use this now with `npm` rather than `yarn` you can simply use the git URL and commit hash.

    "dependencies": {
        ...
        "nuxt": "https://github.com/nuxt/nuxt.js#<hash>",
        ...
    }

Use this in your dependencies and then `npm i` and you should be set.  As of this writing, the current hash is `de2b0a6d5619c5f5334a289ac3287e1ee838b9bc` but it'll obviously change as more commits are added.

Note that the commit hash is important because without it you won't get the dev branch.

hirokiosame:
@cliffrowley 
[...]

Issue #1414, opened 2017-08-18
"Vue packages version mismatch" error


galvez:
I get an error when using nuxt@latest:

Vue packages version mismatch:

- vue@2.2.6
- vue-server-renderer@2.4.2

pi0:
This may be with one your dependencies requiring 2.2.6. Please explicitly upgrade vue version too.

PR #4669, opened 2019-01-02
feat(builder): validate vue-app dependencies and suggest fix

pi0:
## Types of changes

## Description

**Background:** Nuxt requires all peer like dependencies as a dependency for easier usage. Sometimes this causes unwanted behaviors when users explicitly add a specific version in their `package.json` which is incompatible with nuxt.

This could also help to resolve popular `Vue packages version mismatch` error (#198, #669, #1084, #1414, #1851, #2079, #2406, #3454) and also regressions like #4668 that `vue-router@^3.0.2` is required by Nuxt and has fixes but users may encounter problems if anyhow they are still using `vue-router@3.0.0`.

Example error:
image

## Future improvements

- Automate dependency installs

## Checklist:

### DIFF
--- a/packages/vue-app/src/index.js
+++ b/packages/vue-app/src/index.js
@@ -1,26 +1,26 @@
 import path from 'path'
-import pkg from '../package.json'
+import { dependencies } from '../package.json'
 
-export const meta = pkg
-
-export const templatesDir = path.join(__dirname, '..', 'template')
-
-export const templatesFiles = [
-  'App.js',
-  'client.js',
-  'index.js',
-  'middleware.js',
-  'router.js',
-  'server.js',
-  'utils.js',
-  'empty.js',
-  'components/nuxt-error.vue',
-  'components/nuxt-loading.vue',
-  'components/nuxt-child.js',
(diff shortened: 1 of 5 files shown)


clarkdo (approved):
LGTM

packages/vue-app/src/index.js
```diff
@@ -1,26 +1,26 @@
...
-  'components/nuxt-loading.vue',
-  'components/nuxt-child.js',
-  'components/nuxt-link.server.js',
-  'components/nuxt-link.client.js',
-  'components/nuxt.js',
-  'components/no-ssr.js',
-  'views/app.template.html',
-  'views/error.html'
-]
+export const template = {
```
clarkdo: Should this export be default ?

pi0: Yes. But it breaks `yarn dev` script as we do `export * from '@nuxt/vue-app'`. Also reserving namespace for other top-level keys like `version` and `manifest`

clarkdo: gotcha

PR #4671, opened 2019-01-02
feat: add styleExtensions

TheAlexLichter:
In your `css` array, you have to write the corresponding extension when providing your stylesheets. While transitioning from SCSS to PostCSS in several projects, I thought: "Why not completely omitting the extension as we do it for JS resolves :thinking:?"

There we go! 

## Types of changes

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -43,6 +43,21 @@ export function getNuxtConfig(_options) {
     options.extensions = [options.extensions]
   }
 
+  const styleExtensionDefaults = ['css', 'pcss', 'postcss', 'styl', 'stylus', 'scss', 'sass', 'less']
+
+  if (!options.styleExtensions) {
+    options.styleExtensions = styleExtensionDefaults
+  }
+
+  if (typeof options.styleExtensions === 'function') {
+    // Execute functions with defaults passed in as first argument
+    options.styleExtensions = options.styleExtensions(styleExtensionDefaults)
+  }
+
+  if (typeof options.styleExtensions === 'string') {
+    options.styleExtensions = [options.styleExtensions]
+  }
+
   options.globalName = (isNonEmptyString(options.globalName) && /^[a-zA-Z]+$/.test(options.globalName))
     ? options.globalName.toLowerCase()
     : `nuxt`
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -43,21 +43,6 @@ export function getNuxtConfig(_options) {
     options.extensions = [options.extensions]
   }
 
-  const styleExtensionDefaults = ['css', 'pcss', 'postcss', 'styl', 'stylus', 'scss', 'sass', 'less']
-
-  if (!options.styleExtensions) {
-    options.styleExtensions = styleExtensionDefaults
-  }
-
-  if (typeof options.styleExtensions === 'function') {
-    // Execute functions with defaults passed in as first argument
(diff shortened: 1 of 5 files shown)


packages/core/src/resolver.js
```diff
@@ -46,7 +46,7 @@ export default class Resolver {
     return resolve(this.options.srcDir, path)
   }
 
-  resolvePath(path, { alias, module } = {}) {
+  resolvePath(path, { alias, module, isStyle } = {}) {
```
manniL: Maybe we should move to `isXYZ` syntax generally for booleans here? :thinking:

galvez (dismissed):
LGTM!

packages/core/src/resolver.js
pi0: I like this convention :+1:

packages/config/src/options.js
```diff
@@ -43,6 +43,21 @@ export function getNuxtConfig(_options) {
...
+  if (!options.styleExtensions) {
+    options.styleExtensions = styleExtensionDefaults
+  }
+
+  if (typeof options.styleExtensions === 'function') {
+    // Execute functions with defaults passed in as first argument
+    options.styleExtensions = options.styleExtensions(styleExtensionDefaults)
+  }
+
+  if (typeof options.styleExtensions === 'string') {
```
pi0: Why supporting such rare types and adding complexity?

packages/config/src/options.js
```diff
@@ -43,6 +43,21 @@ export function getNuxtConfig(_options) {
...
 
+  const styleExtensionDefaults = ['css', 'pcss', 'postcss', 'styl', 'stylus', 'scss', 'sass', 'less']
+
+  if (!options.styleExtensions) {
+    options.styleExtensions = styleExtensionDefaults
+  }
+
+  if (typeof options.styleExtensions === 'function') {
+    // Execute functions with defaults passed in as first argument
+    options.styleExtensions = options.styleExtensions(styleExtensionDefaults)
```
pi0: If the function returns nothing (undefined) we will get into trouble

packages/config/src/options.js
```diff
@@ -43,6 +43,21 @@ export function getNuxtConfig(_options) {
     options.extensions = [options.extensions]
   }
 
+  const styleExtensionDefaults = ['css', 'pcss', 'postcss', 'styl', 'stylus', 'scss', 'sass', 'less']
```
pi0: If defaults ordering would be important I suggest mentioning it in the docs too :)

pi0 (commented):
Nice enhancement. But I don't see any reason adding complexity for handling option as a function or string unless a specific use case exists. replace or append merge strategy would be enough and more consistent. Modules has the opprotounity to modify options in any manner they need :)

packages/config/src/options.js
manniL: TBH: I mimicked that behavior from the `extensions` property. :see_no_evil: 

If it's not needed, I can remove it for sure :)

packages/config/src/options.js
manniL: Agreed. Maybe even remove the function check (and possible function usage) completely? :thinking:

packages/config/src/options.js
manniL: Yeah, I'll create a doc PR afterwards anyway :relaxed: 
Thanks for the heads up!

pi0 (dismissed):

pi0 (approved):

PR #4674, opened 2019-01-03
refactor: use is prefix for booleans

TheAlexLichter:
See the small remark in #4671 

## Types of changes

- [x] Refactor

## Checklist:


packages/core/src/resolver.js
```diff
@@ -1,10 +1,13 @@
 import Module from 'module'
 import { resolve, join } from 'path'
 import fs from 'fs-extra'
+import consola from 'consola'
 import esm from 'esm'
 
 import { startsWithRootAlias, startsWithSrcAlias } from '@nuxt/utils'
 
+const logger = consola.withTag('@nuxt/resolver')
```
pi0: We don't use scopped loggers anywhere else. If we are going to use them i suggest continue it in another PR and adding a unified helper in core or utils for access (second part is just an idea)

packages/core/src/resolver.js
```diff
@@ -102,15 +113,27 @@ export default class Resolver {
     throw new Error(`Cannot resolve "${path}" from "${resolvedPath}"`)
   }
 
-  requireModule(path, { esm, alias, intropDefault } = {}) {
+  requireModule(path, { esm, isEsm = esm, alias, isAlias = alias, intropDefault, shouldIntropDefault = intropDefault } = {}) {
```
pi0: isESm -> useESM? 

Also shouldIntropDefault looks little verbose to me for usage in code :)

pi0 (dismissed):
LGTM. Just left some personal toughts

galvez (dismissed):

packages/core/src/resolver.js
galvez: @pi0 maybe something like https://github.com/nuxt/fabula/blob/master/src/logging.js?

manniL: Agee. Transforming it back to normal `consola` but add it to the feature/idea backlog

packages/core/src/resolver.js
```diff
@@ -55,12 +66,12 @@ export default class Resolver {
...
 
     // Try to resolve it as a regular module
-    if (module !== false) {
+    if (isModule !== false) {
       resolvedPath = this.resolveModule(path)
     }
 
     // Try to resolve alias
-    if (!resolvedPath && alias !== false) {
+    if (!resolvedPath && isAlias !== false) {
```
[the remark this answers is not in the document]
manniL: Maybe `asAlias` instead (we could align it with `asStyle` (instead of `isStyle`) then too)?

packages/core/src/resolver.js
pi0: @galvez Yes. Exactly. Consola 3 uses lightweight tagged loggers using Proxy.

pi0: @manniL Thanks :)

packages/core/src/resolver.js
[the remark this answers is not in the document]
pi0: isAlias seems better :)

PR #4697, opened 2019-01-06
fix: convert nuxt-error template to render fn

pimlie:
## Types of changes

## Description

As discussed on discord

## Checklist:


packages/vue-app/template/components/nuxt-error.vue
```diff
@@ -1,25 +1,3 @@
...
-      <% } %>
-
-      <div class="logo">
-        <a href="https://nuxtjs.org" target="_blank" rel="noopener"><%= messages.nuxtjs %></a>
-      </div>
-    </div>
-  </div>
-</template>
-
 <script>
```
pi0: This part is still passed through vue-loader. Maybe we can rename this file to `.js`?

pimlie: Unfortunately yes, when I implemented the changes for nuxt-loading.vue @Atinux mentioned he preferred to keep the js code together with the css in one file (due to maintainability?) so I assumed that'd be the case here as well.

pi0: Yes. Makes sense for maintainability. So if we still pass it into vue-loader, there are no perf benefits (at least notable) and makes maintenance of error template harder.

clarkdo: I agree with @pi0 , since vue-loader will pre-compile the `template` so there'is no harm in runtime and Vue compiler may even have some extra performance enhancements in pre-compiling.

PR #4641, opened 2018-12-27
fix(builder): layouts condition

ricardogobbosouza:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Minor bug if layout already exists in listing

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #4701, opened 2019-01-07
fix(builder): js layout

clarkdo:
## Types of changes

## Description

Relate to (and reverts) #4641

## Checklist:


TheAlexLichter (dismissed):

pi0:
Thanks @clarkdo, @manniL for addressing this so fast :heart:

pi0 (dismissed):

packages/builder/src/builder.js
```diff
@@ -349,7 +349,7 @@ export default class Builder {
           }
           return
         }
-        if (!templateVars.layouts[name] && /\.vue$/.test(file)) {
+        if (!templateVars.layouts[name] || /\.vue$/.test(file)) {
```
pi0: Can we sort [`layoutFiles`](https://github.com/nuxt/nuxt.js/blob/fc6fb7a75719fdb6b9c5f4fb375d9db340f26ff4/packages/builder/src/builder.js#L339) here to give `.vue` a higher priority or at least add a comment above this line to explain the reason of check? It seems a hidden workaround like this.

clarkdo: I think comment is better, since sorting may introduce extra overhead

manniL: As long as we make it *understandable* to ppl who read the code, why not ☺️

TheAlexLichter (approved):

pi0 (approved):

PR #4702, opened 2019-01-07
fix(builder): Module.addLayout/nuxt.config take precedence over auto-scanned layouts

clarkdo:
## Types of changes

## Description

Fix issue reported by #4641 and https://github.com/nuxt/nuxt.js/pull/4647#discussion_r245661636

## Checklist:


ricardogobbosouza:
Nice, we should be able to write a layout without the need for a warning, do not you think?
From what I've looked at, this condition is not necessary `|| /\.vue$/.test(file)`, since `js, vue` files are already filtered and this would solve all problems

clarkdo:
> Nice, we should be able to overwrite a layout without the need for a warning, do not you think?
In this case, layout has been defined in `addLayout`, I suppose it has higher priority than auto-scanned layout, how do you think ?

> From what I've looked at, this condition is not necessary `|| /\.vue$/.test(file)`, since `js, vue` files are already filtered and this would solve all problems

The condition is for making `layouts/demo.vue` layouts take precedence over `layouts/demo.js`.

manniL:
@ricardogobbosouza I think a warning is fine there. Are there use-cases where you want to override the layout and *not* delete the overridden file? 🤔

atinux (approved):
Nice job

ricardogobbosouza:
@clarkdo I think we can remove the builder warning and just put in `addLayout` and add a third parameter `override = false` and change condition `if (layout && !override) {`

ricardogobbosouza:
@manniL Yes, I believe that the module should be able to replace a layout without the need for a warning

manniL:
@ricardogobbosouza I got that. But can you explain me in which scenario(s) that'd be helpful when the user simply could delete or rename the file being overriden?

clarkdo:
@ricardogobbosouza If we add the parameter, the control is still in lib author side instead of user side and the underlying override behaviour is still invisible to users 😄

manniL:
@ricardogobbosouza I wanted to suggest that first too but most of the time the module creator is not the actual "module consumer" (and the latter should govern the enabled warnings).

Still, I want to understand in which situations a disabled warning would make sense (from your point of view). Could you **please** explain?

ricardogobbosouza:
What if I add a setting in the options? This would be in control of the user in `nuxt.config.js`
In my case I have a module that always replaces a layout and in all my projects I'm going to receive this notice

clarkdo:
@ricardogobbosouza In your case, I think you can replace the layout file in layouts dir instead of using `addLayout` 😃

TheAlexLichter (approved):

Atinux:
Actually, we should not forget about module authors since it is what makes Nuxt so cool to use.

Actually, the warning popup only when we use `addLayout` (for example a module) AND if a file with the same name is used right @clarkdo ?

I do believe that is a module add a layout, it should be prefixed anyway to avoid conflicts, or, exposes a wrapper component that can be used in user layouts.

WDYT?

manniL:
Then the question is: "What is the module name"? How can we apply that consistently and still "easy to discover"? 🤷‍♂️

I think that'd delegate the problem to a new problem.

ricardogobbosouza:
In my opinion the name of the layout should not be prefixed automatically, this would not allow a module overwriting a layout at the root of the project.

pi0:
Merging this to prevent stale. Feel free opening an Issue/PR/RFC for behavior improvements. PS with module improvements (nuxt/rfcs#10) module container is being deprecated.

PR #4717, opened 2019-01-08
fix: add option to rewatch on path after raw fs event

pimlie:
## Types of changes

## Description

This pr improves watching for file changes on linux in dev mode. Node on linux watches on inodes rather than paths, which means if you move a file you are watching an inode with a different name then intended (or no name at all when the file is removed).

Chokidar has a ~bug~ feature that if you pass a path to a non-existing fie it will listen on the folder instead. Due to this the implementation in the first commit was fine at first, but when you started moving files the number of events triggered for the same file grew enormously. The implementation in the second commit prevents this behaviour. There didnt seem to be a real performance difference between the two commits, the time it took for both implementations was somewhere between 0.1 to 0.4ms on my system (both not really consistent).

Due to that ~bug~ feature in chokidar it also has the side effect that every change in the folder (of the missing file) triggers a refresh. Unfortunately this is also not mitigable by comparing to `customPatterns`. But once another file exists with the name we are looking for that stops (with the second implementation).

Currently only implemented for `watcher.custom`, although the same issue applies to `watcher.files`

Not sure if and where I should add tests for this

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -467,21 +467,30 @@ export default class Builder {
 
     // -- Loading indicator --
     if (this.options.loadingIndicator.name) {
-      const indicatorPath1 = path.resolve(
+      let indicatorPath = path.resolve(
         this.template.dir,
         'views/loading',
         this.options.loadingIndicator.name + '.html'
       )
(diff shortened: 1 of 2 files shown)


pi0 (approved):
LGTM. For experiment :)

PR #4722, opened 2019-01-09
feat(vue-renderer): improvements

pi0:
## Types of changes

## Description
- [x] refactor renderRoute
- [x] new hooks

### New Hooks

Shared:
- `vue-renderer:context`: Final chance to extend context. (Setting modern, spa, url, etc)

SSR:
- `vue-renderer:ssr:prepareContext`: Inject to the context before `renderToString`
- `vue-renderer:ssr:context`: Extend rendered context. Ultimate possibilities for SSR hacks!
- `vue-renderer:ssr:csp`: Extend generated CSP string hashes
- `vue-renderer:ssr:templateParams`: Customize HTML template params

SPA:
- `vue-renderer:spa:templateParams`: Cutomize HTML template params for SPA

## Checklist:


packages/vue-renderer/src/renderer.js
```diff
@@ -291,7 +291,26 @@ export default class VueRenderer {
...
+    const content = await this.renderer.spa.render(context)
+
+    const APP =
+      `<div id="${this.context.globals.id}">${this.context.resources.loadingHTML}</div>` +
+      content.BODY_SCRIPTS
+
+    const html = this.renderTemplate(false, {
+      ...content,
+      APP,
+      ENV
```
Atinux: Why not giving `this.context.options.env`?

pi0: Indeed better idea 👍  Funny fact: It was in my last commit which I didn't push before leaving home 😆

packages/vue-renderer/src/renderer.js
```diff
@@ -298,64 +298,47 @@ export default class VueRenderer {
...
-    if (req && req.modernMode) {
-      APP = await this.renderer.modern.renderToString(context)
-    } else {
-      APP = await this.renderer.ssr.renderToString(context)
-    }
+    const renderer = context.modren ? this.renderer.modern : this.renderer.ssr
+    let APP = await renderer.renderToString(context)
+
+    // Call ssr:context hook
+    this.context.nuxt.callHook('vue-renderer:ssr:context', context)
```
Atinux: This hook is not asynchronous anymore?

pi0: Typo. Thanks :)

packages/vue-renderer/src/renderer.js
```diff
@@ -298,64 +298,47 @@ export default class VueRenderer {
...
+    await this.context.nuxt.callHook('render:routeContext', context.nuxt) // Legacy
```
Atinux: Why not having a mapping inside `callHook`that when `vue-renderer:ssr:context` is called, also call `render:routeContext`? Just exploring some idea to move legacy code somewhere else?

pi0: Indeed we have this mapping:

https://github.com/nuxt/nuxt.js/blob/324135df5128957041131e96deb6b8122a88f4c7/packages/core/src/nuxt.js#L28

But params are different.

packages/vue-renderer/src/renderer.js
```diff
@@ -298,64 +298,47 @@ export default class VueRenderer {
...
-      APP = await this.renderer.modern.renderToString(context)
-    } else {
-      APP = await this.renderer.ssr.renderToString(context)
-    }
+    const renderer = context.modren ? this.renderer.modern : this.renderer.ssr
+    let APP = await renderer.renderToString(context)
+
+    // Call ssr:context hook
+    await this.context.nuxt.callHook('vue-renderer:ssr:context', context)
+    await this.context.nuxt.callHook('render:routeContext', context.nuxt) // Legacy
```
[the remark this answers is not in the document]
pi0: Good idea. Will use `TODO:`

pimlie:
Could we extract all those hook identifiers into some object? That way IDE's can possibly autocomplete them and the chance of errors (by nuxt users) due to small typos is much smaller. Also there is no overview now anywhere with a full list of all available hooks.

I would image something like:
```js
export const hookIds = {
  build: {
    done: 'build:done',
  },
  vue-renderer: { // <- dash might be an issue
    ssr: {
      context: 'vue-rendere:ssr:content'
  }
 ...
}
```

Then if someone wants to use a hook they can either use the text itself (but that'd be mainly for backwards compatiblity) but the preferred way would be to import the nuxtHooks object and just do `nuxt.hook(hookIds.build.done, () => {})`

packages/vue-app/template/utils.js
```diff
@@ -134,6 +134,7 @@ export async function setContext(app, context) {
     // Only set once
     if (context.req) app.context.req = context.req
     if (context.res) app.context.res = context.res
+    if (context.ssrContext) app.context.ssrContext  = context.ssrContext
```
manniL: We have a bunch of inline if's here! :see_no_evil:

pi0: Yeah. These were from before. PS i think sometimes one line ifs are more readable :see_no_evil:

packages/vue-renderer/src/renderer.js
```diff
@@ -407,6 +413,46 @@ export default class VueRenderer {
     }
   }
 
+  async renderRoute(url, context = {}, retries = 5) {
+    /* istanbul ignore if */
+    if (!this.isReady) {
+      if (this.context.options.dev && retries > 0) {
+        consola.info('Waiting for server resources...')
```
manniL: We should show that message only once every X (as briefly discussed before).

packages/vue-renderer/src/renderer.js
```diff
@@ -407,6 +413,46 @@ export default class VueRenderer {
...
+        await waitFor(1000)
+        return this.renderRoute(url, context, retries - 1)
+      } else {
```
manniL: Could we invert that condition or remove the `else` :thinking:

packages/vue-renderer/src/renderer.js
```diff
@@ -407,6 +413,46 @@ export default class VueRenderer {
...
+
+    // Log rendered url
+    consola.debug(`Rendering url ${url}`)
+
+    // Add url to the context
+    context.url = url
+
+    // context.spa
+    if (context.spa === undefined) {
+      // TODO: Remove reading from context.res in Nuxt3
```
manniL: Why :thinking:

packages/vue-renderer/src/renderer.js
```diff
@@ -407,6 +413,46 @@ export default class VueRenderer {
...
+      context.modern = context.req ? (context.req.modernMode || context.req.modern) : false
+    }
+
+    // Call context hook
+    await this.context.nuxt.callHook('vue-renderer:context', context)
+
+    // Render SPA or SSR
+    if (context.spa) {
+      return this.renderSPA(context)
+    } else {
```
manniL: I'd suggest either a ternary or no `else` here

packages/vue-renderer/src/renderer.js
pi0: Because module authors can still use `context.req` :)

pi0: And better using the directly passed options + hooks

packages/vue-app/template/utils.js
manniL: @pi0 They sometimes are *but* can lead to errors quite frequently as well :see_no_evil: 

We could use lodash's `pick` +`Object.assign` in this specific case, couldn't we? :thinking: 

`Object.assign(context, objectWithPickedResReqAndSsrContextHere)`

packages/vue-renderer/src/renderer.js
manniL: Makes totally sense :speak_no_evil:

packages/vue-app/template/utils.js
pi0: Using lodash has both bundling and perf overhead. We can use a for loop :)

packages/core/src/nuxt.js
```diff
@@ -26,7 +26,8 @@ export default class Nuxt extends Hookable {
 
     // Deprecated hooks
```
manniL: Not related to that PR but while you are in here

A `TODO` to remove the hooks in Nuxt 3 would be :+1: 
Do we have warnings for them in place (when using the deprecated ones)?

packages/vue-app/template/utils.js
manniL: That's fine as well :+1:

packages/core/src/nuxt.js
pi0: Yeah. Hookable has built-in support for deprecated hooks and throws warn.

packages/vue-app/template/index.js
```diff
@@ -123,7 +123,8 @@ async function createApp(ssrContext) {
     payload: ssrContext ? ssrContext.payload : undefined,
     req: ssrContext ? ssrContext.req : undefined,
     res: ssrContext ? ssrContext.res : undefined,
-    beforeRenderFns: ssrContext ? ssrContext.beforeRenderFns : undefined
+    beforeRenderFns: ssrContext ? ssrContext.beforeRenderFns : undefined,
```
manniL: For the 5 lines above, would something like `...ssrContext` work or could that copy some variables over that were not intended to be there?

packages/vue-app/template/utils.js
pi0: image

@manniL I proposed this. But still, think it just introduces more complexity and perf overhead. :thinking:

packages/core/src/nuxt.js
pi0: Will add TODO

packages/vue-app/template/index.js
pi0: I think the top-level ones were kept for backward-compatibility only. Spreading all of the `ssrContext` would introduce unwanted keys and perf overhead. I suggest to someday remove top level ones at all.

packages/vue-app/template/utils.js
manniL: Yeah, you are right. Let's stick with the `if`s. Would you leave them inlined or use "full" blocks?

pi0: Will do full blocks to comply with lint rules. We can talk more about it later :)

TheAlexLichter (approved):

PR #4737, opened 2019-01-11
fix: require postcss module via resolver

TheAlexLichter:
Can be useful when trying to run a nuxt app in another folder, eg. via

`yarn nuxt ../my-app` or `yarn nuxt -c ../my-app/nuxt.config.js`

Otherwise, PostCSS plugins will be taken from the current working dir which probably doesn't have the plugins (in the above case it's the root dir of the nuxtjs repo)

## Types of changes


packages/webpack/src/utils/postcss.js
```diff
@@ -6,6 +6,7 @@ import cloneDeep from 'lodash/cloneDeep'
 import createResolver from 'postcss-import-resolver'
 
 import { isPureObject } from '@nuxt/utils'
+import { requireModule } from '@nuxt/cli/src/utils'
```
pi0: By design, No non-distro package should depend on `@nuxt/cli`

packages/webpack/src/utils/postcss.js
```diff
@@ -113,7 +114,7 @@ export default class PostcssConfig {
       // Map postcss plugins into instances on object mode once
       config.plugins = this.sortPlugins(config)
         .map((p) => {
-          const plugin = require(p)
+          const plugin = requireModule(this.nuxt.resolver.resolveModule(p))
```
pi0: Can we use `this.nuxt.resolver.requreModule(p)` ? :)

packages/webpack/src/utils/postcss.js
manniL: Whoops. That was not intended 🙈

pi0 (dismissed):

packages/webpack/src/utils/postcss.js
```diff
@@ -113,7 +113,7 @@ export default class PostcssConfig {
       // Map postcss plugins into instances on object mode once
       config.plugins = this.sortPlugins(config)
         .map((p) => {
-          const plugin = require(p)
+          const plugin = this.nuxt.resolver.requireModule(this.nuxt.resolver.resolveModule(p))
```
clarkdo: What's the purpose of require twice ?

manniL: It's not required twice.

`resolveModule` will get you the correct path (for the project), eg. `/path/to/project/node_modules/tailwindcss/lib/index.js`

`requireModule` will then require the resolved path.

If you mean that we could switch out `requireModule` for `require`, yes, that'd be possible

clarkdo: I mean that it seemed `requireModule` will resolve the path as well 😃

pi0: I agree. requireModule internally does it itself

manniL: Ohh! Good point 👍

pi0 (approved):

PR #4741, opened 2019-01-11
refactor: add module not found fatal message in cli/imports

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/cli/src/imports.js
+++ b/packages/cli/src/imports.js
@@ -1,4 +1,5 @@
 import path from 'path'
+import consola from 'consola'
 
 const localNodeModules = path.resolve(process.cwd(), 'node_modules')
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/imports.js
```diff
@@ -8,7 +9,19 @@ async function _import(modulePath) {
   try {
     m = await import(path.resolve(localNodeModules, modulePath))
   } catch (e) {
-    m = await import(modulePath)
+    try {
+      m = await import(modulePath)
+    } catch (e) {
+      consola.fatal(
```
pi0: This error maybe false positive? For example when importing itself cause a runtime error. We may use error.code

clarkdo: You're right

packages/cli/src/imports.js
```diff
@@ -8,7 +9,23 @@ async function _import(modulePath) {
   try {
     m = await import(path.resolve(localNodeModules, modulePath))
   } catch (e) {
-    m = await import(modulePath)
+    try {
+      m = await import(modulePath)
```
clarkdo: @pi0 I think the first `import` catch also need to exclude error other than `MODULE_NOT_FOUND`

pi0 (approved):

PR #4746, opened 2019-01-12
feat: set html attr `dir` to ltr by default

TheAlexLichter:
## Types of changes

Related: https://github.com/nuxt/nuxt.js/issues/4555

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -297,6 +297,13 @@ export function getNuxtConfig(_options) {
     options.build.optimizeCSS = options.build.extractCSS ? {} : false
   }
 
+  // Set dir to `ltr` if htmlAttrs not set otherwise
+  if (!options.head || !options.head.htmlAttrs) {
+    options.head.htmlAttrs = {
+      dir: 'ltr'
+    }
+  }
+
   const loaders = options.build.loaders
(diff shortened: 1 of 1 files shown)


pi0 (changes requested):
I really disagree with changing default browser behavior by implicitly adding top level dir prop. If it is a tooling problem with postcss/autoprefixer, we should document it.

Also this is a breaking change for RTL projects.

manniL:
> I really disagree with changing default browser behavior by implicitly adding top level dir prop. If it is a tooling problem with postcss/autoprefixer, we should document it.

Agree to . Where should we document that? (As it's a very niche problem and probably should be in the postcss docs)

image

> Also this is a breaking change for RTL projects.

That might be true (if they did not set the `dir` attribute own their own), right.

Atinux:
This is why I love having a core team from all part of the globe <3

I like the idea of helping the user setting the reading direction but I don't think the default should be set to `ltr`.

I think documentation is the key to it. And **if it's linked to multiple entries**, html attrs, postcss, why not giving an alias for it and updating the tools to respect it?

manniL:
rel: https://github.com/nuxt/docs/issues/1108

PR #4754, opened 2019-01-13
feat: add exclude regex array for generated pages

TheAlexLichter:
This PR will add a property called `generate.exclude`. It accepts an array of regular expressions and will prevent generation of routes matching them. The routes will still be accessible when `generate.fallback` is used.

## Use cases

* Having routes like `/admin` in the Nuxt app that don't need to be generated upfront but still need to be accessible
* Modules like [sitemap-module](https://github.com/nuxt-community/sitemap-module/blob/dev/src/index.js#L17) could use it as default for excluding routes in `generate` mode.

## Usage

```js
// nuxt.config.js
export default {
  // ...
  generate: {
    //...
    exclude: [/^\/admin/] // exclude every URL starting with "/admin"
  }
}
```

## Testing

Testing this feature is a bit tricky because we never test that the `generated` fallback actually loads the routes in SPA mode. We only test that it has been generated correctly [here](https://github.com/nuxt/nuxt.js/blob/fe0516978a650166bf178c0339dd7657c29e2504/test/unit/fallback.generate.test.js).

## Types of changes

## Checklist:

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -4,7 +4,7 @@ import consola from 'consola'
 import fsExtra from 'fs-extra'
 import htmlMinifier from 'html-minifier'
 
-import { flatRoutes, isUrl, promisifyRoute, waitFor, isString } from '@nuxt/utils'
+import { flatRoutes, isString, isUrl, promisifyRoute, waitFor } from '@nuxt/utils'
 
 export default class Generator {
   constructor(nuxt, builder) {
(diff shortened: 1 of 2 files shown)


manniL:
image

packages/generator/src/generator.js
```diff
@@ -83,6 +83,11 @@ export default class Generator {
       this.options.router.mode === 'hash'
         ? ['/']
         : flatRoutes(this.options.router.routes)
+
+    this.options.generate.exclude.forEach((regex) => {
+      routes = routes.filter(route => !regex.test(route))
```
pi0: each .filter` creates a clone of the array. So can we use something like `filter(route =>  this.options.generate.exclud.some()` ?

pi0 (approved):
LGTM. BTW do you mind adding a small test for it? @manniL

manniL:
@pi0 That might be a *bit* problematic as I can only test that the file is *not* generated. I can't test if it can be correctly accessed through the `fallback` file. (see "Testing" in the PR text).

Would that suffice?

pi0:
@manniL Seems OK 👍😊

PR #4755, opened 2019-01-13
Add new hook call 'render:headersSet'

### LINKED ISSUE #4614: Hook into Nuxt before the headers are completely sent

Hi guys,

I was looking for a way to remove 'unsafe-inline' from my CSP style-src configuration but didn't find anything about it.
I considered working on a module on my own to be able to do that. Basically, what this module is trying to achieve is just locatiing all `<style ...>xxxx</style>` tags, computing a digest for the content and rewriting style-src CSPs on the fly.

```
import crypto from 'crypto'

export default function CspStyleSrc (moduleOptions) {
    const cspStyleSrcHashes = new Set();
    const _this = this;

  
    this.nuxt.hook('render:route', (url, result, context) => {
        if (!_this.options.render.csp) {
            return;
        }

        const { hashAlgorithm } = _this.options.render.csp; 

        const styleTagRegex = /<style.*>([\s\S]*?)<\/style>/gm;

        let match = null;

        while ((match = styleTagRegex.exec(result.html)) != null) {
            const hash = crypto.createHash(hashAlgorithm);
            hash.update(match[0]);
            cspStyleSrcHashes.add(`'${hashAlgorithm}-${hash.digest('base64')}'`);
        }
    });

    this.nuxt.hook('render:routeDone', (url, result, { res }) => {
        if (!_this.options.render.csp) {
            return;
        }

        const cspHeader = _this.options.render.csp.reportOnly ? 'Content-Security-Policy-Report-Only' : 'Content-Security-Policy';

        const styleCspRegex = /(style-src .*?;)/gm;

        const actualCspString = res.getHeader(cspHeader);

        if (!actualCspString) {
            return;
        }

        const match = styleCspRegex.exec(actualCspString);

        if (!match) {
            return;
        }
        
        const joinedHashes = Array.from(cspStyleSrcHashes).join(' ');

        let newCspString = actualCspString.slice(0, match.index);
        newCspString += `${actualCspString.slice(match.index + match[0].length, actualCspString.length)}; `;
        newCspString += `${match[0].slice(0, -1)} ${joinedHashes};`;

        res.setHeader(cspHeader, newCspString);
    });
}
```

Technically, it works. However, when I try to replace the CSP header, I have an error `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client` which makes perfectly sense as `res.end()` has been called just before calling `render:routeDone` hook.

Having a hook like `render:headersSet`  would help achieve this kind of modifications.

To solve, the problem I would expose a `render:headersSet` hook just before calling `res.end()` in `https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/middleware/nuxt.js`

This way, we would be able to rewrite the headers easily.

Another use case, I can see, still in the context of CSP would be to be able to rewrite `connect-src` directive to allow the application to execute requests on the backend host.

For example, if you follow DevOps principle `Build once, deploy many`, you would end having some routine to dynamically read the target host to use at runtime. As you can't know at build which host will be used, you can't specify it in the CSP policies in `nuxt.config.js` and you'll need to rely on something of your own to add it to the `connect-src` allowed resources.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8346">#c8346</a>)</em></sub></div>

BboyKeen:
This PR adds a new hook call 'render:headersSet' in nuxt server just before calling res.end().

## Types of changes

## Description

This change will allow framework's users overriding the headers before they are sent to the browser
Resolves #4614 

## Checklist:


packages/server/src/middleware/nuxt.js
```diff
@@ -22,6 +22,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
     } = result
 
     if (redirected) {
+      nuxt.callHook('render:headersSet', url, result, context)
```
Atinux: Actually, this one could not work since the response is ended here: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L40

BboyKeen:
Interesting.. 
Does it mean I should add it in the file you mentionned ? 
Or just remove it ?

⁣Kévin Liagre - CTO

https://www.yogosha.com

Mobile : +33(0)7 62 74 90 70

E-Mail | Twitter | LinkedIn | PGP : 0x6EE858D6 ​

Le 14 janv. 2019 16:41, à 16:41, "Sébastien Chopin" <notifications@github.com> a écrit:

packages/server/src/middleware/nuxt.js
BboyKeen: Interesting.. 
Does it mean I should add it in the file you mentionned ? 
Or just remove it ?

clarkdo:
@BboyKeen You can use https://github.com/jshttp/on-headers and add servermiddleware for listening headers

Atinux:
I believe @clarkdo solution is perfect for what you are trying to achieve @BboyKeen

BboyKeen:
Thanks @clarkdo 
Didn't know this kind of solution exists. It's perfect for my case ;)

martinmcd1:
Do you have an example of how you implemented this? I am trying to achieve a similar thing by removing 'unsafe-inline' from the csp?

BboyKeen:
What are you trying to achieve concretely ?

martinmcd1:
I have configured the csp in the nuxt.config.js and I would like to remove 'unsafe-inline' from style-src. 
I seen your previous issue https://github.com/nuxt/nuxt.js/issues/4614 and was hoping to try and implement something similar by intercepting the request and updating the csp.

manniL:
@martinmcd1 	please follow #5627 and the linked PRs

williamchong007:
@martinmcd1 I have an example for adding extra rules to CSP header here, should work similar for removing.
https://github.com/likecoin/likecoin-button/blob/082f791fa2f5d40bb911343a577f4169dbec6378/src/server_middleware/header-listener.js

PR #4756, opened 2019-01-13
chore: don't show stable version number when running from git/source

pimlie:
### _Hey vSource, Michael here_

Currently if you run nuxtcli from source it shows the latest stable version number in the banner. This can sometimes be confusing, this pr changes that behaviour to show _vSource_ instead of eg _v2.3.4_

We could also skip the global variable by adding `return __dirname.endsWidth('src') ? 'Source' : version` instead, let me know if you like that better (i slightly preferred the less complexity of the global variable)

## Types of changes

## Checklist:


scripts/dev
```diff
@@ -22,6 +22,8 @@ const _require = esm(module, {
   }
 })
 
+global.NUXT_VERSION = 'Source'
```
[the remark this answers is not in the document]
pi0: Something like `<dev-08a3375>` (status ~> dirty status)

pimlie:
Now it shows: 
```
Nuxt.js vgit<feat-nuxtcli-dev~5a64819> (+1 -0)
```

Would it be ok to move the `v` from showBanner to nuxt get version? I tried to include a backspace char with git but that messes up the green banner border.

pi0:
Nice work @pimlie. 👍 💯 

> Would it be ok to move the v from showBanner to nuxt get version? I

Yes please.

scripts/dev
```diff
@@ -22,6 +22,37 @@ const _require = esm(module, {
   }
 })
 
+const execa = require('execa')
+
+global = new Proxy(global, {
```
pi0: Do we really need to proxy entire global for lazy evaluating `__NUXT_VERSION_`? I mean maybe there will be easier ways for that.

pimlie: It wasnt actually about lazy evaluation but to make sure that whenever the version is used the dirtyness is correct. It would probably be fine I guess to remove the global Proxy as currently the version is only used/displayed on initial startup. But if the version would be displayed after say every dev run then the version string should be composed every time to check for correct dirtyness.

pimlie: But actually you have a good point. Please see the new commit, now we only proxy global.__NUXT

pi0 (approved):

PR #4766, opened 2019-01-15
refactor: simplify return value in addModule

clarkdo:
## Types of changes

## Description

If return value of handler is promise, then just return the promise is fine.
Otherwise if return value of handler is synchronized, it's also fine, so it's unnecessary to have a promise wrapper

## Checklist:


TheAlexLichter (dismissed):

packages/core/src/module.js
```diff
@@ -157,17 +157,6 @@ export default class ModuleContainer {
...
-
-      // If module send back a promise
-      if (result && result.then) {
-        return resolve(result)
-      }
-
-      // synchronous
-      return resolve()
-    })
+    return handler.call(this, options)
```
pi0: However there is no reason supporting modules that don't return a promise, this is a breaking change in two conditions:
1. This function should always return a promise. We can ensure it by wrapping into `Promise.resolve` or making `addModule` and `async` function. 
2. Non-async module throws an error it is not chained to `promise.catch`.

Because of these, I suggest making function `async` and `await` on `handler.call`

pi0 (changes requested):

packages/core/src/module.js
clarkdo: I suppose that in most cases, `addModule` is not for users since it will be called in `ready`, but if it's used by user, user should already know if the handler is async or sync, then use await/promise or sync way to call `addModule`.

And in our doc, `module` can be either synchronous or async, so it's possible that handler doesn't return promise.

But it's OK for me to add `async` for keeping the potential consistency, I'll add it.

Another thing is: In the `doc`, we don't have any info for `addModule` must be async, so maybe we should add the description.

pi0 (approved):

PR #4777, opened 2019-01-16
fix(examples): use testURL and module names

posva:
## Types of changes

- [X] Other: updated example

## Checklist:

---

I updated the jest example by removing unnecessary <rootDir> that I have found to create problem when having multiple version of babel. I also added the testURL property (https://github.com/jsdom/jsdom/issues/2304)

### DIFF
--- a/examples/jest-vtu-example/package.json
+++ b/examples/jest-vtu-example/package.json
@@ -5,7 +5,7 @@
     "build": "nuxt build",
     "start": "nuxt start",
     "generate": "nuxt generate",
-    "test": "jest ."
+    "test": "jest"
   },
   "dependencies": {
     "nuxt-edge": "latest"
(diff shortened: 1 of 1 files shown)


TheAlexLichter (dismissed):

pi0 (approved):
Thank you @posva :+1: :100:

examples/jest-vtu-example/package.json
```diff
@@ -31,16 +31,17 @@
...
-      "<rootDir>/node_modules/jest-serializer-vue"
+      "jest-serializer-vue"
     ],
     "collectCoverage": true,
     "collectCoverageFrom": [
       "<rootDir>/components/**/*.vue",
       "<rootDir>/pages/*.vue"
-    ]
+    ],
+    "testURL": "http://localhost/"
```
pi0: Shouldn't  we specify default listen port here? (http://localhost:3000)

posva: apparently not, having just the hostname is enough to remove the `SecurityError: localStorage is not available for opaque origins` error

pi0: Thanks for the clarification.

PR #4781, opened 2019-01-17
fix: Allow listening on number 0 port

amoshydra:
## Types of changes

### Why is this change required?
This PR allow the server to listen on port number `0`. In some machine, port `0` let the host machine dynamically generate a unused port for an application.

https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports
> In programming APIs (not in communication between hosts), requests a system-allocated (dynamic) port[5]
>  ["Linux/net/ipv4/inet_connection_sock.c". LXR. Retrieved 2015-01-17.](https://elixir.bootlin.com/linux/v3.18/source/net/ipv4/inet_connection_sock.c#L89)

### What problem does it solve?
When a number `0` is supplied to `nuxt.server.listen` as port, it will be treated as FALSEY. This caused the default server port (3000) to be used instead.

```js
const listener = new Listener({
      port: port || this.options.server.port,
```

## Checklist:

```
 PASS  test/unit/server.listen.test.js
  server listen
    √ should throw error when listening on same port (prod) (179ms)
    √ should assign a random port when listening on same port (dev) (12ms)
    √ should skip the use of default port when listening on port 0 (5ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        4.558s, estimated 5s
```

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -225,7 +225,7 @@ export default class Server {
   async listen(port, host, socket) {
     // Create a new listener
     const listener = new Listener({
-      port: port || this.options.server.port,
+      port: isNaN(parseInt(port)) ? this.options.server.port : port,
       host: host || this.options.server.host,
       socket: socket || this.options.server.socket,
       https: this.options.server.https,
--- a/test/unit/server.listen.test.js
+++ b/test/unit/server.listen.test.js
@@ -41,4 +41,33 @@ describe('server listen', () => {
 
     await nuxt.close()
   })
+
+  test('should skip the use of default port when listening on port 0', async () => {
+    // Stub process.env.PORT
+    const stubDetails = {
+      originalValue: process.env.PORT,
+      hasProperty: 'PORT' in process.env
+    }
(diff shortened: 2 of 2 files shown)


manniL:
Thanks for the PR! That'll be convenient ☺️ 

PS: This is already possible via `port: '0'`.

TheAlexLichter (approved):

amoshydra:
@manniL Yeah, didn't know `0` was causing my tests to fail. I have change my code to use `'0'` at the moment.

pi0 (approved):
Fix looks good to me.

pi0:
only regarding unit tests, using `dev: true` and bootstrap entire nuxt, seems costly. @clarkdo is already working on test rewrites. So waiting for his approval for this change.

clarkdo:
LGTM, I think we can just remove the test since this will be replaced by new unit tests

BTW, how about using `Number.isInteger`

PR #4791, opened 2019-01-18
fix(vue-app): Fix Vuex HMR & refactor for better modules usage

atinux:
## Types of changes

## Description

Fix Vuex HMR with modules & refacto some code.
This way, we can have `todos.js`, `todos/index.js` and `todos/mutations.js` with all having mutations, they will all be merged together (only `state` is overwritten).

Related: #4589 #4582 

## Checklist:

### DIFF
--- a/packages/vue-app/template/store.js
+++ b/packages/vue-app/template/store.js
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
 
 Vue.use(Vuex)
 
-let storeData = {}
-
-let files
+const VUEX_PROPERTIES = ['state', 'getters', 'actions', 'mutations']
+let store = {}
+let fileResolver
 
 void (function updateModules() {
-  files = require.context('@/<%= dir.store %>', true, /^\.\/(?!<%= ignorePrefix %>)[^.]+\.(<%= extensions %>)$/)
-  const filenames = files.keys()
+  fileResolver = require.context('@/<%= dir.store %>', true, /^\.\/(?!<%= ignorePrefix %>)[^.]+\.(<%= extensions %>)$/)
+
+  // Paths are sorted from low to high priority (for overwriting properties)
+  const paths = fileResolver.keys().sort((p1, p2) => {
+    let res = p1.split('/').length - p2.split('/').length
+
+    if (res === 0) {
+      if (p1.includes('/index.')) res = -1
+      if (p2.includes('/index.')) res = 1
+    }
+    return res
+  })
 
   // Check if {dir.store}/index.js exists
-  const indexFilename = filenames.find(filename => filename.includes('./index.'))
+  const indexPath = paths.find(path => path.includes('./index.'))
 
-  if (indexFilename) {
-    storeData = getModule(indexFilename)
+  if (indexPath) {
+    store = requireModule(indexPath, { isRoot: true })
   }
 
-  // If store is not an exported method = modules store
-  if (typeof storeData !== 'function') {
-    // Store modules
-    if (!storeData.modules) {
-      storeData.modules = {}
-    }
-
-    for (const filename of filenames) {
-      let name = filename.replace(/^\.\//, '').replace(/\.(<%= extensions %>)$/, '')
-      if (name === 'index') continue
-
-      const namePath = name.split(/\//)
-
-      name = namePath[namePath.length - 1]
-      if (['state', 'getters', 'actions', 'mutations'].includes(name)) {
-        const module = getModuleNamespace(storeData, namePath, true)
-        appendModule(module, filename, name)
-        continue
-      }
-
-      // If file is foo/index.js, it should be saved as foo
-      const isIndex = (name === 'index')
-      if (isIndex) {
-        namePath.pop()
-      }
-
-      const module = getModuleNamespace(storeData, namePath)
-      const fileModule = getModule(filename)
-
-      name = namePath.pop()
-      module[name] = module[name] || {}
-
-      // if file is foo.js, existing properties take priority
-      // because it's the least specific case
-      if (!isIndex) {
-        module[name] = Object.assign({}, fileModule, module[name])
-        module[name].namespaced = true
-        continue
-      }
-
-      // if file is foo/index.js we want to overwrite properties from foo.js
-      // but not from appended mods like foo/actions.js
-      const appendedMods = {}
-      if (module[name].appends) {
-        appendedMods.appends = module[name].appends
-        for (const append of module[name].appends) {
-          appendedMods[append] = module[name][append]
-        }
-      }
-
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
-  const filenames = files.keys()
+  fileResolver = require.context('@/<%= dir.store %>', true, /^\.\/(?!<%= ignorePrefix %>)[^.]+\.(<%= extensions %>)$/)
+
+  // Paths are sorted from low to high priority (for overwriting properties)
+  const paths = fileResolver.keys().sort((p1, p2) => {
+    let res = p1.split('/').length - p2.split('/').length
+
+    if (res === 0) {
+      if (p1.includes('/index.')) res = -1
+      if (p2.includes('/index.')) res = 1
```
galvez: @Atinux no single-line ifs! :P

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
-        window.<%= globals.nuxt %>.$store.hotUpdate(storeData)
-      })
-    }<% } %>
-  } else {
-    const log = (process.server ? require('consola') : console)
-    log.warn('Classic mode for store/ is deprecated and will be removed in Nuxt 3.')
+
+    const storeModule = getStoreModule(store, namespaces)
+
+    VUEX_PROPERTIES.forEach((property) => mergeProperty(storeModule, moduleData[property], property))
```
galvez:
```js
for (const property of VUEX_PROPERTIES) {
  mergeProperty(storeModule, moduleData[property], property))
}
```

Friendly reminder of our preference for breaking across multiple lines for readability :)

galvez (dismissed):
Sweet changes @Atinux

galvez:
Need to address some failing tests though.

Atinux:
Still a WIP but wanted to open the PR to gather feedback, that was fast :)

packages/vue-app/template/store.js
manniL: From my POV this line is fine/readable enough :man_shrugging: :relaxed:

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
 
 Vue.use(Vuex)
 
-let storeData = {}
-
-let files
+const VUEX_PROPERTIES = ['state', 'getters', 'actions', 'mutations']
+let store = {}
+let fileResolver
```
manniL: Is there any advantage to have this as global variable? We could pass in `fileResolver(path)` to `requireModule` to avoid that

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
+  const indexPath = paths.find(path => path.includes('./index.'))
 
-  if (indexFilename) {
-    storeData = getModule(indexFilename)
+  if (indexPath) {
+    store = requireModule(indexPath, { isRoot: true })
+  }
+
+  // If store is an exported method = classic mode (deprecated)
+  if (typeof store === 'function') {
```
manniL: :+1: for early return :raised_hands:

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
-      storeData.modules = {}
+  // Enforce store modules
+  store.modules = store.modules || {}
+
+  for (const path of paths) {
+    // Remove store path + extension (./foo/index.js -> foo/index)
+    const namespace = path.replace(/^\.\//, '').replace(/\.(<%= extensions %>)$/, '')
+
+    // Ignore indexFile, handled before
+    if (namespace === 'index') continue
```
manniL: Same here as @galvez mentioned before. 

We should prefer `if (cond){ doSthHere }` instead of `if (cond) doSth`. :relaxed:

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
+
+    const namespaces = namespace.split(/\//)
```
manniL: Inconsistent with line 15. Either as string (`split('/')`) or as Regexp everywhere.

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
-  if (module.state && typeof module.state !== 'function') {
-    throw new Error('[nuxt] state should be a function in <%= dir.store %>/' + filename.replace('./', ''))
+  if (isRoot && moduleData.commit) {
+    throw new Error('[nuxt] <%= dir.store %>/' + path.replace('./', '') + ' should export a method which returns a Vuex instance.')
   }
-  return module
+
+  if (isRoot) {
+    // Avoid TypeError: setting a property that has only a getter when overwriting top level keys
+    const state = moduleData.state && moduleData.state !== 'function' ? (() => state) : moduleData.state
```
manniL:
```suggestion
    const state = moduleData.state && typeof moduleData.state !== 'function' ? (() => state) : moduleData.state
```

packages/vue-app/template/store.js
```diff
@@ -3,130 +3,136 @@ import Vuex from 'vuex'
...
-  module.appends = module.appends || []
-  module.appends.push(name)
-  module[name] = file.default || file
+function mergeProperty(storeModule, moduleData, property) {
+  if (!moduleData) return
+
+  if (property === 'state') {
+    storeModule.state = moduleData || storeModule.state
+  } else {
+    storeModule[property] = Object.assign({}, storeModule[property], moduleData)
```
manniL:
```suggestion
    Object.assign(storeModule[property], moduleData)
```

Maybe? :relaxed:

packages/vue-app/template/store.js
Atinux: Woups Nice found, did not notice it

packages/vue-app/template/store.js
Atinux: I don’t think it will work since it will replace the whole content of the property

packages/vue-app/template/store.js
Atinux: I don't see your point here, it's just to avoid re-creating the array in memory with HMR

Atinux:
Next step is to abstract some of the logics right from the server part with `chokidar` to use `.nuxtignore`

packages/vue-app/template/store.js
manniL: But won't this happen anyway because of https://github.com/nuxt/nuxt.js/blob/4e0ab55ed8ea57b107fdf2157fb781ba22f790b3/packages/vue-app/template/store.js#L9-L12

PR #4796, opened 2019-01-18
feat(webpack): support `build.hotMiddleware.client`

wangxingkai:
## Types of changes

## Description
The `build.hotMiddleware` in the `nuxt.config.js` is not working.
According to the config section in the document of [webpack-hot-middleware](https://github.com/webpack-contrib/webpack-hot-middleware#client), the config for client should be set to the path in the webpack config. Currnently we don't have them in the path.

## Checklist:


pi0 (changes requested):
HotMiddlware options are already set here:

https://github.com/nuxt/nuxt.js/blob/40fbe5ba47573341d17823219033fb05b567bab3/packages/webpack/src/builder.js#L212

pi0:
@wangxingkai I still understand the use case for setting client options. But we can't stringify entire `hotMiddleware` option into client query string. Maybe using `build.hotMiddleware.client` ?

wangxingkai:
> @wangxingkai I still understand the use case for setting client options. But we can't stringify entire `hotMiddleware` option into client query string. Maybe using `build.hotMiddleware.client` ?

Yep, sure. We still keep { log, path, heartbeat } in the `build.hotMiddleware`, and put all the client config inside `build.hotMiddleware.client`.

pi0:
@wangxingkai Now what do you think of something like this:

```js
const clientOptions = {
  reload: true,
  timeout: 30000,
  name: this.name,
  path: `${this.options.router.base}/__webpack_hmr/${this.name}`,
  ...this.options.build.hotMiddleware.client
}

const clientOptionsStr = querystring.stringify(clientOptions).replace(/\/\//g, '/')

`webpack-hot-middleware/client?${clientOptionsStr}`
```

wangxingkai:
@pi0 Looks simpler. but two things:
1. `path` and `name` may not want to be overridden.
2. `this.options.build.hotMiddleware` may be undefined.

I have updated this PR.

pi0 (dismissed):
Seems great!

pi0:
@wangxingkai would you please take a look at unit-tests?

pi0 (approved):

PR #4814, opened 2019-01-19
fix(config): define once default nuxt config filename

kevinmarrec:
## Types of changes

## Description
No more multiple `nuxt.config.js` literal strings in core, it's now defined once in `@nuxt/config` and take the value `nuxt.config.js` or `nuxt.config.ts` depending on if we're running in a TypeScript environment or not (`process.env.NUXT_TS`).

It also fixes 1 or 2 issues where Nuxt was checking a condition against `nuxt.config.js` instead of `nuxt.config.ts` when in TypeScript environment.

=> Example of issue before this refactor : Run `nuxt-ts` without `nuxt.config.ts` ends up in 
screenshot_19
Default configuration file was compared to `nuxt.config.js`, now it's fixed by : https://github.com/nuxt/nuxt.js/pull/4814/files#diff-ee3bc4b396043c3228e1a978a05819a4R57


TheAlexLichter (dismissed):

husayt (approved):

atinux (approved):

Atinux:
LGTM

TheAlexLichter (approved):

kevinmarrec:
I remember that @pi0 had some improvements to propose regarding this PR around having just `nuxt.config` without extensions

pi0:
Agree. Can be delagated to later

PR #4823, opened 2019-01-21
fix: watch custom patterns only when it exists

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -624,34 +624,37 @@ export default class Builder {
       .on('add', refreshFiles)
       .on('unlink', refreshFiles)
 
+    this.watchCustom(refreshFiles)
+  }
+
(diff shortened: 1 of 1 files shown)


packages/builder/src/builder.js
```diff
@@ -630,27 +630,29 @@ export default class Builder {
...
-      if (rewatchOnRawEvents && Array.isArray(rewatchOnRawEvents)) {
-        this.watchers.custom.on('raw', (_event, _path, opts) => {
-          if (rewatchOnRawEvents.includes(_event)) {
-            this.watchers.custom.close()
-            this.watchers.custom = null
-
-            watchCustom(true)
-          }
-        })
+    if (customPatterns.length > 0) {
```
manniL: Could we use an early return here?

TheAlexLichter (approved):

PR #4850, opened 2019-01-24
fix: match subdir under node_module  in transpile

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -225,8 +225,10 @@ export default class WebpackBaseConfig {
       {
         test: /\.jsx?$/i,
         exclude: (file) => {
+          file = file.split('node_modules', 2)[1]
+
           // not exclude files outside node_modules
-          if (!/node_modules/.test(file)) {
+          if (!file) {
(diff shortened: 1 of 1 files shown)


packages/webpack/src/config/base.js
```diff
@@ -225,8 +225,10 @@ export default class WebpackBaseConfig {
       {
         test: /\.jsx?$/i,
         exclude: (file) => {
+          file = file.split('node_modules', 2)[1]
```
manniL: What if `node_modules` is not included in that path? 🤔

(Is that even possible?)

clarkdo: If it doesn't include `node_modules`, then it won't be included in loader, same as before

atinux (approved):

Haroenv:
Thanks @clarkdo, @manniL and of course also @Atinux for finding this! I was staring at the wrong things altogether :)

clarkdo:
@Haroenv You're welcome, thank you so much for your interest in Nuxt.js

PR #4855, opened 2019-01-25
chore(examples): Add tsx example

andoshin11:
## Types of changes

## Description
Added new example application written with TSX using [vue-tsx-support](https://github.com/wonderful-panda/vue-tsx-support) and CSS Modules

Type checking feature working as expected :)

## Checklist:


kevinmarrec:
Hi @andoshin11, Thanks for your PR ! 
But I'm wondering why do you need `vue-tsx-support`, as `nuxt-ts-edge` already ships this feature.
I know there is no detailed documentation yet but could you try removing the dependency and see how it goes ? :)

andoshin11:
@kevinmarrec Thanks for your review!
I used vue-tsx-support mainly because of its HTMLAttributes definition.(which was originally copied from @types/react and still is being maintained by @wonderful-panda)
https://github.com/wonderful-panda/vue-tsx-support/blob/master/types/dom.d.ts

However you made me think that it would be more beneficial to publish @types/vue first by extracting HTML Attributes types from vue-tsx-support, then use them on Nuxt globally so that our tsx support becomes more complete.

andoshin11:
Or I should make a PR to create 'vue/types/dom.d.ts'

andoshin11:
@kevinmarrec Just opened a new Pull Request on official Vue repository.

https://github.com/vuejs/vue/pull/9369

Hope this could push discussions forward.

kevinmarrec:
@andoshin11 I think it should indeed be shipped in Vue typedefs.

I suggest you update your example by doing this:
- Replace `import * as tsx from 'vue-tsx-support'` by `import Vue from 'vue'`
- Replace `tsx.Component` by `Vue.extend`
- Remove `vue-tsx-support` dependency
- Copy the typedefs needed inside this example under a `dom.d.ts` or in a folder (`types/dom.d.ts`)
- Reference these typedefs in `tsconfig.json`

andoshin11:
Hmm...
I spent some times without vue-tsx-support and keep getting errors on children component's prop def.



May still need some tweaks on JSX Element definition. I'll keep investigating on it.

kevinmarrec:
@andoshin11 Alright, thanks for your time !
Anyway there is better TSX and overall better TypeScript support announced for Vue 3.0.

andoshin11:
FYI: working on the occurring problem with Vue core team ↓↓

https://github.com/vuejs/vue/pull/9379

But it'll take some time to completely solve the problem, so I updated this example into much simpler version without using props and other complicating features.

Maybe this will do for now as a very first example code.
I'm definitely going to update this example when the Vue type definition is updated :)
@kevinmarrec

examples/tsx/pages/index.tsx
```diff
@@ -0,0 +1,13 @@
+import * as tsx from 'vue-tsx-support'
```
kevinmarrec: Please replace that by `import Vue from 'vue'` then use `Vue.extends({})` instead of `tsx.compoent({})` (which is not anymore in dependencies and should throw an error with current code).

Then we're good !

kevinmarrec (changes requested):

examples/tsx/pages/index.tsx
andoshin11: Oh...that is totally embarrassing 😑 

I pushed the fix!

andoshin11: Thanks for your time!

kevinmarrec (dismissed):
Maybe should we use `typescript-tsx` instead of `tsx` example name, but anyway everything is ok for me.

manniL:
Yeah, `typescript-tsx` would be more descriptive, I agree.

examples/tsx/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "build": "nuxt-ts build",
+    "start": "nuxt-ts start",
+    "generate": "nuxt-ts generate",
+    "lint": "tslint --project .",
+    "lint:fix": "tslint --project . --fix",
+    "post-update": "yarn upgrade --latest",
+    "watch:css": "tcm components -w"
+  },
+  "dependencies": {
+    "nuxt-ts-edge": "latest"
```
manniL:
```suggestion
    "nuxt-ts": "latest"
```

manniL: (see #4874)

andoshin11: 👍 
changed `tsconfig.json`, too

andoshin11:
`typescript-tsx` would be better, so I renamed it 👍

kevinmarrec (approved):

PR #4856, opened 2019-01-25
fix(ts): better `tsconfig.json` handling & improve tests

kevinmarrec:
## Types of changes

## Description
I refactored code and fixed `ts-node` registration by telling him where is the right `tsconfig.json`.

## Checklist:
- [x] I have updated tests to cover my changes (if not applicable, please state why)
- [x] All updated and existing tests are passing.


distributions/nuxt-ts/bin/nuxt-ts.js
```diff
@@ -5,10 +5,11 @@ process.env.NUXT_TS = 'true'
 
 const nuxtCommands = ['dev', 'build', 'generate', 'start']
 const rootDir = (process.argv[2] && !nuxtCommands.includes(process.argv[2])) ? process.argv[2] : process.cwd()
+const tsConfigPath = require('path').resolve(rootDir, 'tsconfig.json')
```
pi0: Prefer moving require/import to top of the script, when lazy loading is not happening

packages/typescript/src/index.js
```diff
@@ -1,7 +1,42 @@
...
+        '@nuxt/vue-app'
+      ]
+    }
+  }, { spaces: 2 })
+  consola.info(`Extending ${chalk.bold.blue(`node_modules/${configToExtend}/tsconfig.json`)}`)
+  consola.success(`Generated successfully at ${chalk.bold.green(tsConfigPath)}`)
+}
+
+export async function setup(tsConfigPath) {
+  if (!existsSync(tsConfigPath)) {
```
pi0: Can we add a guard to prevent calling function multiple times?

pi0 (dismissed):

distributions/nuxt-ts/bin/nuxt-ts.js
kevinmarrec: I don't think `import { resolve } from 'path'`  will work as it's the binary and not compiled.

So just `const { resolve } = require('path')` at the top should be ok ?

pi0: @kevinmarrec YES :)

packages/typescript/src/index.js
kevinmarrec: What could be the case when it's called multiple times ?

How do we implement such `guard` ? :)

pi0: Like this:

https://github.com/nuxt/nuxt.js/blob/dev/packages/cli/src/setup.js#L13

> What could be the case when it's called multiple times?

TS Node will be registered twice :)

kevinmarrec: I was meaning "In which case `setup` will be called twice ?", but I'll put a guard anyway ;)

kevinmarrec: Thanks for the code ref

pi0: Programmatic usage or when forking workers. But generally my reason is ensurance :)

kevinmarrec: If I define a `_setup` variable, can it conflicts with `_setup` of CLI ? Or is `let` make it defined at the file scope ?

kevinmarrec:
@pi0 Gonna edit tests, I think I'm doing something bad as i'm not awaiting `setup` in `beforeAll`, and will also add a short test to test the guard :)

packages/typescript/src/index.js
pi0: @kevinmarrec rollup takes care of renaming even if they are bundled. No worries about same variable name or scope.

kevinmarrec: Perfect then, thanks for the info 💪

clarkdo:
@kevinmarrec The pr looks great.

Could you move all typescript package related tests into `packages/typescript/test` dir.

kevinmarrec:
Will do, thanks @clarkdo

kevinmarrec:
@clarkdo Should we also move `typescript` fixtures under `packages/typescript/test/fixtures` ?
Could be cleaner to put everything TS related things around this package now we have it.

kevinmarrec:
I would also want to remove `typescript-custom`, I consider it obsolete as we've built-in TS support.

clarkdo:
I think we only need unit tests, fixtures will be removed in future

clarkdo:
So for now it would be great if we write unit tests which can cover 100% packages/typescript

kevinmarrec:
@clarkdo 100% coverage right now, but not 100% branches covered, I probably need to mock `existsSync` and enquirer `prompt` to get 100% branches covered

pi0 (approved):

PR #4875, opened 2019-01-29
fix: add missing dependencies

aparajita:
`@nuxt/vue-renderer` and `jsdom` are imported by `server.js`, but are not included here.

**TIP:** Always try installing with `pnpm` as a sanity check. Because it does not flatten the dependency tree, you don't get a free pass if a sub-dependency provides a dependency you miss.

## Types of changes

## Description

If `nuxt` is installed with `pnpm`, there are missing dependencies. `npm` (and presumably `yarn`) hide this because they flatten the dependency tree. This change adds the missing dependencies.

## Checklist:


ricardogobbosouza:
Hi @aparajita
I could take advantage of and remove the `@nuxt/config` dependency that is not used

aparajita:
Not sure how that solves anything. The point is there are imports in the source that are not listed as dependencies. Should that be the case?

ricardogobbosouza:
The title of your PR also does not agree, it should be something like `fix(server): necessary dependencies`
And you can remove `@nuxt/config` of dependencies, it is not being used in this package

clarkdo:
@aparajita 

Thank you for the pr 😄 

`jsdom` is not a required dependency for most users since it's used by `renderAndGetWindow` which is a method for test purpose, so `jsdom` is dynamic imported in nuxt for reducing the package size, if you need it, I suggest adding it into your `devDependencies`

BTW, could you please remove `nuxt/vue-renderer` in `nuxt/core` as well ?

pi0 (changes requested):
`@nuxt/vue-renderer` is also a dynamic dependency of `@nuxt/server`. They are only required by the distribution packages. We can't break architecture because of PNPM compatibility.

aparajita:
> @nuxt/vue-renderer is also a dynamic dependency of @nuxt/server.

Dynamic or not, shouldn't all dependencies be declared?

> We can't break architecture because of PNPM compatibility

How does **adding** a dependency break anything?

pi0:
> Dynamic or not, shouldn't all dependencies be declared?

It is a **peer dependency** of `server` package which both are defined in [core/package.json](https://github.com/nuxt/nuxt.js/blob/dev/packages/core/package.json#L15)

> How does adding a dependency break anything?

Increased install size in case of using an alternative package instead of `vue-renderer.
-> Nuxt migrated to mono-repo to make it happen. But we can't do breaking changes so had to define dependencies like this which works but commonly supported package managers (NPM/Yarn)

aparajita:
Does the current build work with yarn workspaces?

pi0:
> Does the current build work with yarn workspaces?

Yes :) Even with different dependency versions. Yarn "hoist"s common packages to the root but it doesn't make any problems because `@nuxt` dependency versions are pinned.

image

aparajita:
FYI, at least in my testing yarn pnp fails. Is it working for you?

pi0:
@aparajita Not tested PNP but if you can create an issue for PNP support I will try to help make it work.

aparajita:
Will do.

At some point *all* of the package managers are going to go the yarn-pnp/tink/pnpm route and stop using flattened package hierarchies, because they're enormously inefficient. At that point a lot of packages (including nuxt) are going to break because they are deliberately or unwittingly depending on the package manager's low-level implementation.

pi0:
@aparajita Agree and fully understand your point. But even in that case solution is not adding optional dependencies to upper packages. Would be probably explicitly listing in `package.json` or another method. There should be a way to define an alternative to `@nuxt/vue-renderer` in `@nuxt/server` or we miss the entire point of modularity. (They should be on the same package)

aparajita:
@pi0 I'll let you figure out how to do it. I only know that it eventually has to be done. 😄

pi0:
@aparajita Yes, it should. And sorry for my bad answers. I did not mean any offense or something like that. At nuxt, we insist to do everything in the best possible way.

aparajita:
@pi0 No problem, I have been an open source maintainer before, I know what it's like. 👍

aparajita:
@pi0 See https://cmty.app/nuxt/nuxt.js/issues/c8642 for the pnp issue.

PR #4883, opened 2019-01-29
refactor: generate routes and template files in builder

clarkdo:
## Types of changes

## Checklist:


packages/builder/src/builder.js
```diff
@@ -155,7 +155,7 @@ export default class Builder {
       )
     }
     await Promise.all(buildDirs.map(dir => fsExtra.mkdirp(dir)))
-
+\\\\\\\\\\\\
```
manniL: Leftovers?

clarkdo: Oh shit

packages/builder/src/context/template.js
```diff
@@ -0,0 +1,76 @@
+import hash from 'hash-sum'
+import consola from 'consola'
+import uniqBy from 'lodash/uniqBy'
+import serialize from 'serialize-javascript'
+
+import devalue from '@nuxt/devalue'
+import { r, wp, wChunk, serializeFunction } from '@nuxt/utils'
+
+export default class TemplateContext {
```
pi0: Nice refactor :+1:

packages/builder/src/builder.js
```diff
@@ -541,47 +488,26 @@ export default class Builder {
...
-            if (!lodash) {
-              consola.warn('Avoid using _ inside templates')
-              lodash = require('lodash')
-            }
-            return lodash[prop]
-          }
-        })
-      },
-      interpolate: /<%=([\s\S]+?)%>/g
+    const templateOptions = templateContext.templateOptions
```
pi0: We can use destructure syntax here :)

packages/builder/src/builder.js
```diff
@@ -297,100 +220,118 @@ export default class Builder {
...
-      transition: this.options.transition,
-      layoutTransition: this.options.layoutTransition,
-      dir: this.options.dir,
-      components: {
-        ErrorPage: this.options.ErrorPage
-          ? this.relativeToBuild(this.options.ErrorPage)
-          : null
-      }
-    }
+    const templateContext = new TemplateContext(this, this.options)
```
pi0: Can we refactor rest of this function to a `TemplateBuilder` class?

pi0 (dismissed):
LSexyTM at this point :D

packages/builder/src/builder.js
```diff
@@ -297,100 +220,118 @@ export default class Builder {
...
+
+    await this.resolveLayouts(templateContext)
+    await this.resolveRoutes(templateContext)
+    await this.resolveStore(templateContext)
+    await this.resolveMiddleware(templateContext)
+    await this.resolveCustomTemplates(templateContext)
+    await this.resolveLoadingIndicator(templateContext)
```
manniL: Would `Promise.all` be useful here?

packages/builder/src/builder.js
```diff
@@ -297,100 +220,118 @@ export default class Builder {
...
+      this.options.plugins.map((p) => {
+        if (typeof p === 'string') {
+          p = { src: p }
+        }
+        const pluginBaseName = path.basename(p.src, path.extname(p.src)).replace(
+          /[^a-zA-Z?\d\s:]/g,
+          ''
+        )
+
+        if (p.ssr === false) {
```
manniL: We could refactor that block into a dedicated function

packages/builder/src/builder.js
```diff
@@ -459,51 +401,57 @@ export default class Builder {
...
-    templatesFiles = templatesFiles
-      .map((file) => {
-        // Skip if custom file was already provided in build.templates[]
-        if (customTemplateFiles.includes(file)) {
-          return
-        }
-        // Allow override templates using a file with same name in ${srcDir}/app
-        const customPath = r(this.options.srcDir, 'app', file)
-        const customFileExists = fsExtra.existsSync(customPath)
+    let templateFiles = templateContext.templateFiles
```
manniL: Destructuring would work here as well

packages/builder/src/builder.js
```diff
@@ -525,14 +473,13 @@ export default class Builder {
...
+        templateFiles.push({
           src: indicatorPath,
           dst: 'loading.html',
           custom: customIndicator,
           options: this.options.loadingIndicator
         })
       } else {
-        /* istanbul ignore next */
         // eslint-disable-next-line no-console
         console.error(
```
ricardogobbosouza: Use `consola`?

ricardogobbosouza (changes requested):

packages/builder/src/builder.js
pi0: A fact: All console logs with `nuxt` cli will be redirected to consola's corresponding logger. But would be good idea for consistency and cleaner code.

packages/builder/src/builder.js
clarkdo: I wrote it like at first, but it feels less readable and will trigger `getter` twice

```js
const templateOptions = {
 ...templateContext.templateOptions,
 imports: {
    ...templateContext.templateOptions.imports
    resolvePath: this.nuxt.resolver.resolvePath,
 }
}
```

packages/builder/src/builder.js
clarkdo: I tried, but the methods are all couple with builder's members and methods like `resolveRelativeToBuild`, `supportedExtensions`, `template`, `routes`, `watch`....

So I reverted my changed for avoiding leak `builder` into other class.

packages/builder/src/builder.js
clarkdo: Yes it's a good idea, but the order of the resolve method may have dependencies between each other.

Some methods may change the `options` or member variables which are used in next methods, we may need to  go through all the methods and refactor them

packages/builder/src/builder.js
clarkdo: Do you mean marshalling `mode` in `normalizePlugins` ?
I think current method is not so complex and the marshalling is not shareable as well, if it can be shared, we can move it into some common util class then.

packages/builder/src/builder.js
clarkdo: For destructure, good idea, will change it:

```js
const { templateOptions } = templateContext
```

packages/builder/src/builder.js
clarkdo: I'll do it now.

packages/builder/src/builder.js
clarkdo: @ricardogobbosouza I may work on log migration for all nuxt code in another pr.

clarkdo: You're write, will change it

packages/builder/src/builder.js
```diff
@@ -525,63 +475,43 @@ export default class Builder {
...
-        templatesFiles.push({
+        templateFiles.push({
           src: indicatorPath,
           dst: 'loading.html',
           custom: customIndicator,
           options: this.options.loadingIndicator
         })
       } else {
-        /* istanbul ignore next */
         // eslint-disable-next-line no-console
```
ricardogobbosouza: Now do not need to deactivate anymore. :smile::

ricardogobbosouza (changes requested):

packages/builder/src/builder.js
clarkdo: Oh, thank you, sorry for I didn't notice it

TheAlexLichter (approved):

Issue #3828, opened 2018-09-01
safari: infinite loop when hot reloading


ydnar:
### Version

[v1.4.2](https://github.com/nuxt.js/releases/tag/v1.4.2) (or latest 2.0.0 pre code)

### Steps to reproduce

1. Launch dev server via `yarn dev` (`HOST=0.0.0.0 nuxt`)
2. Load http://localhost:3000/
3. Edit code, any code. Watch the nuxt dev server recompile everything.
4. Reload
5. HMR fails, triggering rapid reloading of the entire page, forever, until Safari kills the tab for excessive CPU usage.
6. The only way to test new code is to close the tab and open a new tab to localhost (and re-open dev console).

### What is expected ?

Hot reload to work without forcing me to close the tab and open a new tab every time I change a line of code.

### What is actually happening?

Browser crushing infinite reload.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7666">#c7666</a>)</em></sub></div>

manniL:
@ydnar Have you tried to remove all page information about localhost in safari? I had similar issues when a service worker stuck.

aldarund:
I have same\similar issue on chrome all time. I have use of nuxt/pwa. And reloading just  hangup on some of hot reload and after it page dont load at all no matter if u close tab, open new tab or anything it will just load forever. Only anonymous window helps. Pretty terrible dev experience

manniL:
@aldarund See https://github.com/nuxt-community/pwa-module/pull/60#issuecomment-407008485. This should be an issue of `pwa-module`, not nuxt-core itself.

@ydnar Please ping me if you **do not** use workbox/pwa-module

ydnar:
@manniL we’re not using `pwa-module`, just vanilla Nuxt (`dev` branch).

Aric-sky:
> I have same\similar issue on chrome all time. I have use of nuxt/pwa. And reloading just hangup on some of hot reload and after it page dont load at all no matter if u close tab, open new tab or anything it will just load forever. Only anonymous window helps. Pretty terrible dev experience

Have you solved this problem?

Norbz:
We have the same issue here, some things I tried without any success : 
- Removing the .nuxt folder
- Completely killing the server and restarting
- clearing all cookie, localStorage and stuff...
- reinstalling all node_modules

[...]

PR #4895, opened 2019-01-30
feat(server): ip and rate-limit middleware

pi0:
## Types of changes

## Description

Credits: @manniL helps and @yoyo837 useful comment about reproduction for safari bug.

Add a new option `server.rateLimit` to enable rate limiting support. (Based on [express-rate-limit](https://github.com/nfriedly/express-rate-limit) -- a rewrite with modern syntax and improvements)

A rate limit is enabled by default on **dev** mode that limits requests to **individual resources** in a period of time(5/sec)

A console warning makes the developer aware of strange behavior:

image

This PR also resolves loop reported in #3828.

## Objectives

Production:
- A built-in solution to reduce/prevent DDOS attacks

Development:
- Warn users about the request loops
- Prevent browser/server crash

## Checklist:


packages/config/src/options.js
```diff
@@ -324,5 +324,17 @@ export function getNuxtConfig(_options) {
     bundleRenderer.runInNewContext = options.dev
   }
 
+  // Enable rate limit for dev mode
+  if (options.dev) {
+    if (typeof options.server.rateLimit === 'undefined') {
+      // It is unsual to request same resource more than 5 times in a 1sec!
```
pi0: Is it rational !?

manniL:
@yoyo837 Not really. It's strange because Safari will load the `hot-reload` JSON files again and again from disk, even if you exactly specify not to cache it. Really strange.

Rate limiting will provide a reliable workaround for it. Of course, finding the root cause would be better.

atinux (dismissed):

packages/config/src/options.js
```diff
@@ -324,5 +324,20 @@ export function getNuxtConfig(_options) {
...
+      options.server.rateLimit = {
+        windowMs: 1000,
+        max: 5,
+        keyGenerator: req => req.url,
+        onLimitReached: (req) => {
+          consola.warn('Too fast requests detected on path: ' + req.url)
```
```suggestion
          consola.warn('Too many requests in a short time detected for path: ' + req.url)
```

Maybe?

TheAlexLichter (dismissed):

packages/config/src/options.js
Atinux: Agree with the typo change

clarkdo:
We'd better have at least unit tests for the pr.

packages/config/src/options.js
```diff
@@ -324,5 +324,20 @@ export function getNuxtConfig(_options) {
     bundleRenderer.runInNewContext = options.dev
   }
 
+  // Enable rate limit for dev mode
+  if (options.dev) {
```
clarkdo: Any reason for enable it in dev mode ?
I think there is no big value to limit rate in development except testing

Atinux: It's to fix Safari issue, see #3828.

packages/config/src/options.js
clarkdo: I suggest moving it to middleware defaults

packages/config/src/options.js
manniL: @clarkdo the safari bug mentioned in the original text ☺️  (#3828)

packages/config/test/__snapshots__/options.test.js.snap
```diff
@@ -305,6 +305,7 @@ Object {
     "host": "localhost",
     "https": false,
     "port": 3000,
+    "rateLimit": undefined,
```
clarkdo: We should have a full configuration for ratelimt in snapshot

packages/config/src/options.js
Atinux: But I agree that it could create unwanted side effects on development (getting a `429` status code when requesting static files)

Atinux:
I also think we should add a way to populate `req.ip` with Nuxt server, default to: `req.headers['x-forwarded-for'] || req.connection.remoteAddress || req.socket.remoteAddress || (req.connection.socket ? req.connection.socket.remoteAddress : null)`

packages/server/src/middleware/rate-limit.js
```diff
@@ -0,0 +1,163 @@
...
+
+export default function RateLimit(_options) {
+  const options = {
+    ...defaults,
+    ..._options
+  }
+
+  if (typeof options.handler === 'undefined') {
+    options.handler = (req, res) => {
+      const secondsLeft = Math.ceil((+req.rateLimit.resetTime - Date.now()) / 1000)
```
clarkdo: How about using hrtime ?

packages/config/src/options.js
```diff
@@ -324,5 +324,20 @@ export function getNuxtConfig(_options) {
...
+    if (typeof options.server.rateLimit === 'undefined') {
+      // It is unsual to request same resource more than 5 times in a 1sec!
+      options.server.rateLimit = {
+        windowMs: 1000,
+        max: 5,
+        keyGenerator: req => req.url,
```
clarkdo: Is this in purpose that make rateLimit work for all clients connections instead of specific client ?

packages/server/src/middleware/rate-limit.js
Atinux: Never heard of it, why not :)

Also, we could support [ms](https://www.npmjs.com/package/ms) package for `windowMs`

packages/config/src/options.js
Atinux: Exactly, this is to avoid Safari loop requests on the same resource issue.

packages/server/src/middleware/rate-limit.js
clarkdo: Actually we already have the timer in utils, maybe we can leverage it.

packages/config/src/options.js
pi0: I'll make cache key `ip + url`. Actually it is an odd behaviour from browser to request same resource more than once.

packages/server/src/middleware/rate-limit.js
```diff
@@ -0,0 +1,163 @@
...
+    if (this.hits[key]) {
+      this.hits[key]++
+    } else {
+      this.hits[key] = 1
+    }
+
+    cb(null, this.hits[key], this.resetTime)
+  }
+
+  decrement(key) {
```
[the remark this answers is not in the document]
manniL: Either that or make `incr` -> `increment` 👍

packages/config/src/options.js
clarkdo: For dev mode, I think ip with a bigger max is better

packages/config/test/__snapshots__/options.test.js.snap
pi0: It is initially set `undefined` by default :)

packages/config/src/options.js
pi0: This may cause bad DX. Like display rate limit error page if an internal ajax was repeated.

packages/config/test/__snapshots__/options.test.js.snap
clarkdo: I mean we may need to have all configurable keys tested, maybe in middleware unit test

packages/server/src/middleware/rate-limit.js
pi0: hrTime is the legacy version of bigInt. We don't need that amount of accuracy here.

packages/config/test/__snapshots__/options.test.js.snap
pi0: Agree @clarkdo. Will include them for unit test.

packages/config/src/options.js
pimlie: I agree with @clarkdo, but maybe just only enable this middleware by default for safari. As in _don't fix it if it aint broken_ in Linux, Chrome or other browsers. 
The modern middleware already uses ua-parser, but if you think its to cumbersome to check rate-limiting based on the browser then please include a config option that reflects this behaviour is a workaround for an issue with safari and use `os.type() === 'Darwin'` to enable that by default

pi0: @pimlie This is not just a workaround for Safari bug. Please see PR description.

pi0:
@Atinux IP middleware added

packages/config/src/options.js
clarkdo: My concern is that if check it with ip and url, it may be not that useful.
If only works for a lot of frequent single resource request.

I suppose the rate-limit may have `per user`/`per service` which has longer time window and bigger maximum count, and `per url` which time window is shorter。

For dev mode, it's not that strict, it's fine for me to use current config for detect some infinite loop or duplicate renderering.

pimlie:
For what its worth, but I am strongly against including rate-limiting as default functionality in Nuxt. If I look at how this issue came to be then again the most important reason seems to be to fix that safari issue even though this should really be a separate nuxt package and not part of the core. Or in other words I dont understand why `html-minifier` should be a separate package (while it is already included for generate) and `rate-limiting` should be included in the core.

I just really dont care about rate-limiting, not in production (cause then I will just use nginx) and not in dev mode (when I develop I have the js console open 99% of the time so you dont have to tell me some request was sent twice because I can already see that). So please dont make it more difficult for me as I just know I will have some issue with this in the future because (most likely) the default key is not deterministic enough but by that time I have forgotten about this rate limiting and I will spent hours debugging a non-existing issue of why some stuff isnt loading.

Also good rate-limiting is quite hard to do and requires more work then just a date per ip, you want bursts etc. See eg this nginx blog page: https://www.nginx.com/blog/rate-limiting-nginx/

pi0:
@pimlie Again, Safari is not the only reason of this feature. Development mode of nuxt is weak in case of every type of loops. Misconfigurations like (axios-module#189) may easily happen and the user cannot easily detect them. This type of limit is only enabled for dev mode and can be easily changed or disabled per project. I think we should provide the best defaults that are suitable for most of the users. 

I honor your ideas and if there are real concerns about bad DX by enabling path based limiting will agree simply disabling it by default. But are there any real reasons/concerns?

pi0:
@pimlie Talked with the core team. We are probably going to disable this behavior by default and allow CSB and Safari users optionally enable rate limit middleware.

lbogdan:
@pi0 And also enable it on the nuxt main sandbox, I hope? 🙂 Although finding the root cause of that Safari bug would indeed be much better!

pi0:
I'm closing this PR because it is probably not needed for everyone.

For CSB: https://github.com/nuxt/codesandbox-nuxt/pull/11

Also as a result of this PR and motivation, we made a full connect compatible middleware called [reqlim](https://github.com/jsless/reqlim) which is usable in every nuxt project using `serverMiddleware`.

PR #4906, opened 2019-01-30
perf(ts): transpileOnly when using `nuxt-ts start`

kevinmarrec:
## Types of changes

- [x] Performance (a non-breaking change which improves performance)

## Description
Knowing Runtime TypeScript are already type checked during `nuxt-ts build`, type checking is not needed during `nuxt-ts start`.

By using `ts-node` `transpileOnly` option for the `nuxt-ts start` command, it results in a faster startup & lower base RAM usage.

### Some stats around my own production server (VPS)
| transpileOnly | Startup time | Memory usage | 📊
|---------|---|---|---|
| false     | ~10s | 120MB | 👎 |
|   **true**      | **~3s** | **70MB** | 👍 |

## Checklist
- [x] I have updated tests to cover my changes
- [x] All updated and existing tests are passing.

### DIFF
--- a/packages/typescript/src/index.js
+++ b/packages/typescript/src/index.js
@@ -47,6 +47,7 @@ export async function setup(tsConfigPath) {
     project: tsConfigPath,
     compilerOptions: {
       module: 'commonjs'
-    }
+    },
+    transpileOnly: process.argv[2] === 'start'
   })
 }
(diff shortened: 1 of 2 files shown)


dschewchenko:
can you try `typeCheck: false` option for server?

kevinmarrec:
@dschewchenko `typeCheck` is false by default but type-checking is enabled only because `transpileOnly` is set to false by default : https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L194

It's kind of weird what they're doing right here, but as far I know, just `transpileOnly: true` should be fine.

pi0 (approved):

galvez (approved):

pi0 (approved):

PR #4907, opened 2019-01-30
chore(examples): zero-downtime pm2 typescript example

### LINKED ISSUE #3078: Zero downtime deploy
Hey all!

I haven't found a good solution for Nuxt.js zero downtime deployments (I tried using `pm2` but without success in terms of zero downtime deployments).

Has anyone found a good solution and is willing to share it? :thinking: 

<div align="right"><sub><em>This feature request is available on <a href="https://nuxtjs.cmty.io">Nuxt.js</a> community (<a href="https://nuxtjs.cmty.io/nuxt/nuxt.js/issues/c2662">#c2662</a>)</em></sub></div>

dschewchenko:
## Types of changes

New example

## Description

~Add `process.send('ready')` when nuxt server is listening~
Resolves [#3078](https://github.com/nuxt/nuxt.js/issues/3078)
I added simple example with pm2 configuration and use listen hook in nuxt.config for sending `ready` event to pm2

## Checklist:


examples/pm2-typescript/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "nuxt-ts-edge": "latest",
```
kevinmarrec: We're going to make examples use `stable` version instead of `edge` builds. 
See https://github.com/nuxt/nuxt.js/pull/4874
So I suggest switching to `nuxt-ts`

examples/pm2-typescript/package.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "scripts": {
+    "dev": "nuxt-ts",
+    "build": "nuxt-ts build",
+    "start": "nuxt-ts start",
+    "generate": "nuxt-ts generate",
+    "lint": "tslint --project ."
+  },
+  "devDependencies": {
+    "@types/node": "^10.12.18",
```
kevinmarrec: The `@types/node` is shipped by `nuxt-ts` through `@nuxt/typescript`, so you can securely remove it here.

examples/pm2-typescript/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
kevinmarrec: To follow Nuxt TS best practices, I suggest doing like https://github.com/nuxt/nuxt.js/blob/dev/examples/typescript/tsconfig.json
You can copy-paste and replace `@nuxt/typescript-edge` by `@nuxt/typescript` & `@nuxt/vue-app-edge` by `@nuxt/vue-app`.

packages/cli/src/commands/start.js
```diff
@@ -15,6 +15,11 @@ export default {
 
     // Listen and show ready banner
     await nuxt.server.listen()
+
+    if (typeof process.send === 'function') {
```
kevinmarrec: This `if` statement might be not needed. I don't know which versions of Node support it but it might be since a while ?

examples/pm2-typescript/pages/index.vue
```diff
@@ -0,0 +1,14 @@
...
+  <h1>
+    Hello world !
+  </h1>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({
+  components: {}
```
kevinmarrec: Just a minimal remark but you can simply stick with `@Component({})` or `@Component` if there is no options to set.

kevinmarrec (changes requested):
Thank you for the PR ! 

I suggested some changes 🙂

Can't wait to be able to have a true zero-downtime PM2 setup 😃

pimlie:
Would it be possible to extend the example to include documentation and/or a command to have zero-downtime while also having to (re)build Nuxt? It seems to me that currently this example is only about zero-downtime while restarting nuxt server but (again it seems, maybe nuxt-ts works differently on start) to skip the build phase?

packages/cli/src/commands/start.js
pimlie: Do we really want a pm2 specific feature to be part of the Nuxt core? I think this would be much better placed in a nuxt ready hook:
```js
// nuxt.config.js
export default {
  hooks: {
    ready() {
      process.send('ready')
    }
  }
}
```

kevinmarrec: (I don't think it's pm2 specific but it helps pm2 anyway)
I will check where the hook is trigerred but yeah, it seems a better solution for sure !

dschewchenko:
> Would it be possible to extend the example to include documentation and/or a command to have zero-downtime while also having to (re)build Nuxt? It seems to me that currently this example is only about zero-downtime while restarting nuxt server but (again it seems, maybe nuxt-ts works differently on start) to skip the build phase?

All depends on your need.
Of course, before starting server, you need to build app. It's described in documentation.
Before restarting service you may to deploy all files to temp dir, build there, then move files to your project root, then restart `pm2 startOrReload ecosystem.config.js`. All described in docs. It's simple.
Recommended steps I will describe in example Readme

packages/cli/src/commands/start.js
dschewchenko: From node docs
> Child Node.js processes will have a process.send() method of their own that allows the child to send messages back to the parent.

dschewchenko: Due to this small clarification, we need `if` statement

kevinmarrec: @dschewchenko Thanks for clarifications. I think we should end up with @pimlie proposal, so we'll not have to change Nuxt core. Can you confirm it works ? And the `listen` `hook` might be better than `ready` one, IDK 🙂

dschewchenko: Ok. I will update example to use hook, and delete changes in core

examples/pm2-typescript/nuxt.config.ts
```diff
@@ -0,0 +1,7 @@
+export default {
+  hooks: {
+    listen: () => {
+      process.send('ready')
```
kevinmarrec: screenshot_9

It seems that we need to at least check `if (process.send)` to prevent TypeScript error.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L848

kevinmarrec (changes requested):

examples/pm2-typescript/nuxt.config.ts
kevinmarrec: You can also define the hook without async function
```js
export default {
  listen () {
    if (process.send) {
      process.send('ready')
    }
  }
}
```

kevinmarrec (approved):

galvez (approved):

PR #4909, opened 2019-01-31
fix typescript examples tilde path module resolution in tsconfig

acrylix:
updating `tsconfig.json` in typescript examples that properly sets tilde (~) module resolution

## Types of changes

## Description

this fixes the following ts error:

screen shot 2019-01-30 at 9 49 04 pm

## Checklist:

### DIFF
--- a/examples/typescript-tsx/pages/index.tsx
+++ b/examples/typescript-tsx/pages/index.tsx
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-import HelloWorld from '../components/HelloWorld'
+import HelloWorld from '~/components/HelloWorld'
 
 export default Vue.extend({
   name: 'Home',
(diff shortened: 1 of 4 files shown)


kevinmarrec:
@acrylix It's already provided by `https://github.com/nuxt/nuxt.js/blob/dev/packages/typescript/tsconfig.json`

So TypeScript compiles and resolves well. The `Cannot find module` is only an error that happens in editor cause it seems it can't read `paths` from the extended Nuxt `tsconfig.json`.

There is the same issue with `experimentalDecorators` option which need to be put in your local `tsconfig.json` aswell to let VSCode knows (are you using VSCode ?)

kevinmarrec:
I think `baseUrl` & `paths` of `compilerOptions` should always in final `tsconfig.json` (not in the extended one), as it's mandatory for VSCode to have it defined in your project.

We would also need to change Nuxt TS `tsconfig.json` generation (for projects missing the file) to take in account this behavior.

acrylix:
That makes sense, I guess this is just with how the examples folders are set up, and not being able to read the `tsconfig.json` from package level. 

dropping this PR

Issue #4884, opened 2019-01-29
scrollBehaviour can't be connected as external module in Nuxt@2.4.0


DreaMinder:
### Version

[v2.4.0](https://github.com/nuxt.js/releases/tag/v2.4.0)

### Reproduction link

[https://github.com/DreaMinder/nuxt-scrollbehaviour-reproduction](https://github.com/DreaMinder/nuxt-scrollbehaviour-reproduction)

### Steps to reproduce

1. npm i
2. npm run dev

### What is expected ?

To work as expected 

### What is actually happening?

error:in ./.nuxt/router.js                                                                                                            

```

Vue.use(Router)

> const scrollBehavior = functionfunction (to, from){
                                                 ^
let position = 0

```

### Additional comments?

2.3.4 was working ok with
```
scrollBehavior: require('./scroll')
```

2.4.0 works only with inlined function:
```
 scrollBehavior: function(to, from){
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8563">#c8563</a>)</em></sub></div>

manniL:
Maybe related to https://github.com/nuxt/nuxt.js/commit/7831e57943d8a8a5fe2a576f42f562347ab5d1f8

cc @galvez

galvez:
@DreaMinder @manniL thanks, indeed, it's related to that change. I'll push a fix.

aaronransley:
I have updated to nuxt 2.4.3 and continue to receive errors when attempting to import my `scrollBehavior` function from an external module. 

I tried a `yarn upgrade nuxt`, tried manually incrementing the version in `package.json`, verified the newest Nuxt is installed, and even deleted my node_modules folder & reinstalled. No dice.

Errors and snippets below:

**nuxt.config.js**
```js
import { scrollBehavior, extendRoutes } from './lib/route-utils'

module.exports = {
  // ...
  router: {
    extendRoutes,
    scrollBehavior
  }
  // ...
}
```

**lib/route-utils.js**
```js
export function scrollBehavior(to, from, savedPos) {
  // ...
}
```

**When running `nuxt dev`**

```js
ERROR  Failed to compile with 1 errors
ERROR  in ./.nuxt/router.js  friendly-errors 22:51:54
Syntax Error: Unexpected token, expected "(" (34:32)
  32 | Vue.use(Router)
  33 | 
[...]

PR #4916, opened 2019-01-31
fix: serializeFunction for external imports

galvez:
Fixes minor issue involving external imports in `serializeFunction` #4884

### DIFF
--- a/packages/utils/src/serialize.js
+++ b/packages/utils/src/serialize.js
@@ -1,4 +1,3 @@
-
 import serialize from 'serialize-javascript'
 
 export function serializeFunction(func) {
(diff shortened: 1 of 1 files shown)


TheAlexLichter (dismissed):

packages/utils/src/serialize.js
```diff
@@ -14,7 +14,7 @@ export function serializeFunction(func) {
         return _
       }
     })
-    .replace(`${func.name}(`, 'function (')
+    .replace(`${func.name || 'function'}(`, 'function(')
```
pi0: @galvez Unit test is failing because of whitespace change here.

manniL:
I think the fix should be based against `2.x` branch 🤔

PR #4925, opened 2019-02-01
fix: extra properties in templateFiles

clarkdo:
## Types of changes

## Description

If `t` is string like 'app/template.js', the returned object will be:

```js
{
	0: "a"
	1: "p"
	2: "p"
	3: "/"
	4: "t"
	5: "e"
	6: "m"
	7: "p"
	8: "l"
	9: "a"
	10: "t"
	11: "e"
	12: "."
	13: "j"
	14: "s",
	src: '...',
	dst: '...',
	custom: '...'
}
```

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -447,7 +447,7 @@ export default class Builder {
               dst: t.dst || path.basename(t.src || t),
               custom: true
             },
-            t
+            typeof t === 'object' ? t : undefined
           )
         })
       )


packages/builder/src/builder.js
```diff
@@ -447,7 +447,7 @@ export default class Builder {
               dst: t.dst || path.basename(t.src || t),
               custom: true
             },
-            t
+            typeof t === 'object' ? t : undefined
```
pi0: Isn't better-throwing errors instead of silent pass?

clarkdo: I think it's a correct behaviour, t can be String here, like 

```js
// nuxt.config.js

build: {
  templates: [
    'foo.js'
  ]
}
```

pi0: NVM. Didn't noticed it is passed as second arg.

pi0 (approved):

PR #4930, opened 2019-02-02
feat(cli): option to open the project in the browser

ricardogobbosouza:
## Types of changes

## Description

Opens the server listeners url in the default browser

```bash
nuxt -o
//or
nuxt --open
````

## Checklist:

### DIFF
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -1,5 +1,6 @@
 import consola from 'consola'
 import chalk from 'chalk'
+import opn from 'opn'
 import { common, server } from '../options'
 import { showBanner, eventsMapping, formatPath } from '../utils'
 
@@ -9,17 +10,29 @@ export default {
   usage: 'dev <dir>',
   options: {
     ...common,
-    ...server
+    ...server,
+    open: {
+      alias: 'o',
+      type: 'boolean',
+      description: 'Opens the server listeners url in the default browser'
+    }
   },
 
   async run(cmd) {
     const { argv } = cmd
-    await this.startDev(cmd, argv)
+    const nuxt = await this.startDev(cmd, argv)
+
+    // Opens the server listeners url in the default browser
+    if (argv.open) {
+      for (const listener of nuxt.server.listeners) {
+        await opn(listener.url)
+      }
+    }
   },
 
   async startDev(cmd, argv) {
     try {
-      await this._startDev(cmd, argv)
+      return await this._startDev(cmd, argv)
     } catch (error) {
       consola.error(error)
     }
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -32,7 +32,9 @@ export default {
 
   async startDev(cmd, argv) {
     try {
-      return await this._startDev(cmd, argv)
+      const nuxt = await this._startDev(cmd, argv)
+
+      return nuxt
     } catch (error) {
       consola.error(error)
     }
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -1,6 +1,6 @@
 import consola from 'consola'
 import chalk from 'chalk'
-import opn from 'opn'
+import opener from 'opener'
 import { common, server } from '../options'
 import { showBanner, eventsMapping, formatPath } from '../utils'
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by pi0** [comment]
`open` package size seems good: [install size](https://packagephobia.now.sh/result?p=open)

File: packages/cli/src/commands/dev.js [hunk 1]
```diff
@@ -9,17 +10,29 @@ export default {
...
+      for (const listener of nuxt.server.listeners) {
+        await opn(listener.url)
+      }
+    }
   },
 
   async startDev(cmd, argv) {
     try {
-      await this._startDev(cmd, argv)
+      return await this._startDev(cmd, argv)
```
pi0: Please use named return like `const nuxt = await` + `return { nuxt }`

ricardogobbosouza: @pi0 renamed

pi0 (dismissed):

ricardogobbosouza:
@pio `opn` package was the best and most used I found

pi0:
@ricardogobbosouza Yeah seems light-weight. The only problem I can predict is that we have the plan to try out zero dependency builds of nuxt packages. Packages that have binary files inside can be little problematic as it can't be bundled. However, my comment was general. No worries until then.

kevinmarrec (dismissed):

kevinmarrec:
Nice ! I'm wondering why [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) (that we use for the `--analyze` option of the `build`command) is using [opener](https://www.npmjs.com/package/opener) instead of [opn](https://www.npmjs.com/package/opn)

pi0:
[opener code](https://unpkg.com/opener@1.5.1/lib/opener.js) seems has an explicit workaround for WSL and also has no binary inside it. So maybe worth comparing them?

ricardogobbosouza:
In my opinion I prefer the `opener` because it is smaller and does not contain a binary inside it.

galvez (dismissed):

TheAlexLichter (changes requested):
Evaluate `opn` vs `opener` first.

ricardogobbosouza:
Any tips on what you guys use to compare?

manniL:
@ricardogobbosouza 

Some criterias
* Bundle Size
* System compatibility (OSX, Win, *NIX) / Interoperability
* Stability
* "Burdens" (like extra binaries)

ricardogobbosouza:
@manniL a quick comparison

https://bundlephobia.com/result?p=opn@5.4.0
https://bundlephobia.com/result?p=opener@1.5.1

`opener` is almost half the size, with no dependencies and no binaries

ricardogobbosouza:
I will change to the `opener`

pimlie:
One of the reasons opener is smaller is because opn bundles the xdg-open sh script itself. I guess thats to not to rely on the local system to have (an updated version of) xdg-open installed.

Looking at the wsl detection opn (or more specifically is-wsl) has an extra check. Unfortunately the extra dependency cannot really be used as an argument, because sindresorhus always splits his code in tiny packages and have them as deps. He claims that that increases his productiveness.

Although I am not sure what the usefulness is for us, the opn package seems to have a bit more features with the trade-off that its a bigger install size.

pi0 (approved):

pi0:
We want something small that just works. I suggest let's try `opener`. If we found any unusual problems with it fork and improve or change to another package to prevent reinventing.

Atinux:
Actually, I believe this open should be by default on `nuxt dev` for Nuxt 3 when we will have:
- Module install with the command line = Nani?
- HMR for server parts (server middleware, etc) = Nuxt workers
- Nuxt UI that show the build progress (instead of a pending page waiting for server resources)

I am open to have it as an option at the moment, thank you @ricardogobbosouza for this PR <3

ricardogobbosouza:
@Atinux We can add as default in Nuxt 3.
Most frameworks come as standard, eg: Ionic

pimlie:
Please also keep in mind that not everybody develops on localhost. We could check for e.g. `SSH_TTY` env variable to check if somebody is logged in from remote and then not open the url by default (Because otherwise you have to close Links or lynx everytime you run nuxt dev and nobody develops their website for text-mode anyway)

Atinux:
Really good remark @pimlie

TheAlexLichter (approved):

PR #4933, opened 2019-02-03
chore: bump Otechie fee

galvez:
Figured this slipped :)


manniL:
I would wait until Monday's meeting with that.

PR #4936, opened 2019-02-03
fix: warn when using array postcss configuration

TheAlexLichter:
Warn users when using an array as postcss configuration (legacy). Docs are already up to date for that

## Types of changes

## Checklist:


pi0 (approved):
Please propose an RFC for such decisions.

PR #4937, opened 2019-02-03
fix: improve warning for untranspiled modules

TheAlexLichter:
## Types of changes

## Description

The improved error message should give users a better idea what to do :thinking: 

image

### DIFF
--- a/packages/server/src/middleware/error.js
+++ b/packages/server/src/middleware/error.js
@@ -5,8 +5,11 @@ import consola from 'consola'
 import Youch from '@nuxtjs/youch'
 
 export default ({ resources, options }) => function errorMiddleware(err, req, res, next) {
-  // ensure statusCode, message and name fields
+  if (/unexpected token/gi.test(err.message)) {
+    err.message += ' - You might have to transpile third party packages!'
+  }
 
+  // ensure statusCode, message and name fields
   const error = {
     statusCode: err.statusCode || 500,
     message: err.message || 'Nuxt Server Error',


packages/server/src/middleware/error.js
```diff
@@ -5,8 +5,11 @@ import consola from 'consola'
 import Youch from '@nuxtjs/youch'
 
 export default ({ resources, options }) => function errorMiddleware(err, req, res, next) {
-  // ensure statusCode, message and name fields
+  if (/unexpected token/gi.test(err.message)) {
```
pi0: This error is _originating_ from webpack. So would be better to handle in friendly-errors plugin.

Atinux: Here? https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/plugins/warnfix.js

Atinux: I will also recommend to add an url to the documentation and avoid `!` that feels frustrating.

pi0: I mean add a similar transformer like this for `unexpectedToken.js`:

https://github.com/nuxt/friendly-errors-webpack-plugin/blob/master/src/transformers/moduleNotFound.js

And add it to [`defaultTransformers`](https://github.com/nuxt/friendly-errors-webpack-plugin/blob/master/src/friendly-errors-plugin.js#L12)

manniL: Will do ☺️

clarkdo: @manniL Can we close this now ?

manniL: @clarkdo Feel free to close it, open an issue on https://github.com/nuxt/friendly-errors-webpack-plugin/ and assign me ☺️

PR #4955, opened 2019-02-04
fix: await buildDone hook

clarkdo:
## Types of changes

## Checklist:


packages/server/src/middleware/nuxt.js
```diff
@@ -74,7 +74,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
     res.setHeader('Accept-Ranges', 'none') // #3870
     res.setHeader('Content-Length', Buffer.byteLength(html))
     res.end(html, 'utf8')
-    nuxt.callHook('render:routeDone', url, result, context)
+    await nuxt.callHook('render:routeDone', url, result, context)
```
pi0: This creates extra promise object on each request. Is it necessary for other than tests?

clarkdo: `callHook` will return promise anyway, if we don't wait here, there will have unhandled promise.

pi0 (approved):

PR #4956, opened 2019-02-05
feat(module): support src as a function in addModule

ricardogobbosouza:
## Types of changes

## Description

While developing a module I came across a problem in its coverage.

If you are analyzing the modules such as [pwa](https://github.com/nuxt-community/pwa-module), [moment-module](https://github.com/nuxt-community/moment-module), [nuxt-i18n](https://github.com/nuxt-community/nuxt-i18n) the coverage is made on top of the `test` folder and not the `lib` or `src`, this is why nuxt solves the module using `esm` by default, so the module is not is covered.

Today to test the modules we created a `nuxt.config.js` like this:
```js
modules: [
  '@@'
]
```
and this has no coverage.

With this correction we can create like this:
```js
modules: [
  require('../../')
]
```
and this has coverage

If we want to create the module using ESM and with coverage we have to add new configurations in `jest` and `babel`:

```json
  "babel": {
    "presets": [
      [
        "env",
        {
          "modules": false
        }
      ]
    ],
    "env": {
      "test": {
        "presets": [
          [
            "env",
            {
              "targets": {
                "node": "current"
              }
            }
          ]
        ]
      }
    }
  },
  "jest": {
    "testEnvironment": "node",
    "collectCoverage": true,
    "collectCoverageFrom": [
      "lib/**/*.js"
    ],
    "transform": {
      "^.+\\.js$": "babel-jest"
    }
  }
```
and add the dependencies `babel-jest` and `babel-preset-env`

## Checklist:

### DIFF
--- a/packages/core/src/module.js
+++ b/packages/core/src/module.js
@@ -115,8 +115,8 @@ export default class ModuleContainer {
     let options
     let handler
 
-    // Type 1: String
-    if (typeof moduleOpts === 'string') {
+    // Type 1: String or Function
+    if (typeof moduleOpts === 'string' || typeof moduleOpts === 'function') {
       src = moduleOpts
     } else if (Array.isArray(moduleOpts)) {
       // Type 2: Babel style array
@@ -124,6 +124,11 @@ export default class ModuleContainer {
     } else if (typeof moduleOpts === 'object') {
       // Type 3: Pure object
       ({ src, options, handler } = moduleOpts)
+    }
+
+    // Define handler if src is a function
+    if (typeof src === 'function') {
+      handler = src
     }
 
     // Resolve handler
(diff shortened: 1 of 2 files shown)


pi0 (approved):

pi0:
In description, should it be like this ?

```js
modules: [ () => require('../../') ]
```

ricardogobbosouza:
@pi0 I will suggest improvements in `module-template`

ricardogobbosouza:
> Fantastic investigation and improvement for module coverage. For ecosystem consistency, I suggest not documenting this to prevent module incompatibility for 2.x users prior to this version.
> 
> @ricardogobbosouza Would you please also open a PR on module template to reflect changes to the default test suit?

https://github.com/nuxt-community/module-template/pull/9

PR #4958, opened 2019-02-05
feat: show warning on forced exit

### LINKED ISSUE #4944: Nuxt 2.4 will keep the build hanging with intervals and/or open connections
### Version

[v2.4.1](https://github.com/nuxt.js/releases/tag/v2.4.1)

### Reproduction link

[https://codesandbox.io/s/lxwr482vwl](https://codesandbox.io/s/lxwr482vwl)

### Steps to reproduce
Add a module which sets a interval or opens a connection.
Run `npm run build`

### What is expected ?

That Nuxt.js will finish building (so CI will continue)

Build keeps hanging

### Additional comments?

Would be nice if Nuxt will close the process after building, as it always has done.

Although intervals are cleared easily, modules will be broken. Connections should be opened and tested on build (in dev mode at least), otherwise it would be quite annoying for module users to wait for finishing building in order to receive errors about incorrect credentials.

It won't be possible to add server middleware at the `listen` hook. So a module like [nuxt-session](https://github.com/Timkor/nuxt-session) will not be possible. Because it needs a open connection (to Redis for example) before the middleware is added.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8603">#c8603</a>)</em></sub></div>

pimlie:
feat: warn when nuxt doesnt exit automatically
fix: add default value when no charsPerLine specified to foldLines fn
feat: make maxCharsPerLine a fn as the console width can be changed while running a command
fix: move format settings to separate module so we can mock them
feat: add different box types
fix: move forceExitTimeout to settings

## Types of changes

## Description

In previous versions Nuxt always called process.exit when eg the build command was finished. Fortunately that has been removed as calling process.exit hides badly written code. But to improve DX I think its better to (try to) detect when Nuxt doesnt exit automatically and show a warning when that happens. This resolves #4944 

## Checklist:


packages/cli/src/command.js
```diff
@@ -43,6 +49,19 @@ export default class NuxtCommand {
     }
 
     return Promise.resolve(this.cmd.run(this))
+      .then(() => {
+        if (this.forceExit) {
```
pi0: What's your idea to move this delayed util to the utils?

packages/cli/src/command.js
```diff
@@ -43,6 +49,19 @@ export default class NuxtCommand {
...
+          const exitTimeout = setTimeout(() => {
+            let msg = `The command 'nuxt ${this.cmd.name}' finished but Nuxt.js did not exit after ${forceExitTimeout}s\n`
+            msg += 'This is most likely not caused by a bug in Nuxt\n'
+            msg += 'Make sure to cleanup all timers and listeners you or your plugins/modules start.\n'
+            msg += 'Nuxt.js will now force exit'
+            process.stderr.write(warningBox(msg))
+            process.exit(0)
```
pi0: Is it a good idea to silently exit with code 0?

packages/cli/src/utils/formatting.js
```diff
@@ -36,3 +32,32 @@ export function colorize(text) {
     .replace(/ (-[-\w,]+)/g, m => chalk.bold(m))
     .replace(/`(.+)`/g, (_, m) => chalk.bold.cyan(m))
 }
+
+export function box(message, title, options) {
```
pi0: Good refactor :+1:

packages/cli/src/utils/formatting.js
```diff
@@ -36,3 +32,32 @@ export function colorize(text) {
...
+  ].join('\n'), Object.assign({
+    borderColor: 'white',
+    borderStyle: 'round',
+    padding: 1,
+    margin: 1
+  }, options)) + '\n'
+}
+
+export function warningBox(message, title) {
+  return box(message, title || chalk.yellow('❗ Nuxt Warning'), {
```
pi0: What about ⚠️ ? :D

packages/cli/src/utils/settings.js
```diff
@@ -0,0 +1,8 @@
+export const forceExitTimeout = 5
```
pi0: I vote to call this file `constants`.

packages/cli/src/command.js
pimlie: My main reason for not doing that was because you dont always want to forceExit so you'd still need that if-statement. And I didnt see much re-usability for the setTimeout part, but can change it anyway. Will then also change the order, so the .then is only added when we want to set the forceExit timeout.

packages/cli/src/command.js
pimlie: I did this on purpose because of the complaints in the issue I mentioned above, exiting with non-zero exit code would still be a breaking change for them compared to previous nuxt behaviour. I would propose to leave it at 0 for now and change it to non-zero in v3.0.

packages/cli/src/utils/formatting.js
pimlie: Yeah, I should've known to use that one as consola also uses it. Actually now I searched for the ✖ char, noticed it was a 'heavy multiplication sign' unicode char and then searched for 'heavy exclamation mark' ;)

packages/cli/src/command.js
pi0: @pimlie Seems reasonable. Let's do this for v3.0. Feel free adding a `TODO` statement above.

packages/cli/src/command.js
pi0: Can be reused by either programmatic users or workers even if kept in cli package but refactored into a util :) Anyway just a comment.

packages/cli/src/utils/formatting.js
pi0: Yes. Would be a better idea to keep consistency (Emoji or ascii anywhere) Also windows support should be considered.

pimlie:
@pi0 What do you think about adding a `--force-exit` cli option which is default true for now and default false as of v3. Similar to https://jestjs.io/docs/en/cli#forceexit ?

pi0:
Good idea @pimlie

manniL:
Hmm. Shouldn't we classify that PR as a `fix`? 🤔

pimlie:
FYI this is a WIP, still future-proofing the force-exit option. Will update when ready

packages/cli/src/utils/index.js
```diff
@@ -144,3 +137,22 @@ export function normalizeArg(arg, defaultValue) {
...
+      msg += 'Nuxt.js will now force exit\n\n'
+      msg += chalk.bold('DeprecationWarning: Starting with Nuxt version 3 this will be a fatal error')
+
+      // TODO: Change this to a fatal error in v3
+      process.stderr.write(warningBox(msg))
+      process.exit(0)
+    }, timeout * 1000)
+    exitTimeout.unref()
+  } else {
+    process.exit(0)
```
pimlie: When force-exit is explicitly used (in v3) this should be 0, unless we can let the command's run method returns a status.

packages/cli/src/utils/index.js
```diff
@@ -144,3 +137,22 @@ export function normalizeArg(arg, defaultValue) {
...
+    const exitTimeout = setTimeout(() => {
+      let msg = `The command 'nuxt ${cmdName}' finished but did not exit after ${timeout}s\n`
+      msg += 'This is most likely not caused by a bug in Nuxt.js\n'
+      msg += 'Make sure to cleanup all timers and listeners you or your plugins/modules start.\n'
+      msg += 'Nuxt.js will now force exit\n\n'
+      msg += chalk.bold('DeprecationWarning: Starting with Nuxt version 3 this will be a fatal error')
+
+      // TODO: Change this to a fatal error in v3
+      process.stderr.write(warningBox(msg))
+      process.exit(0)
```
pimlie: Maybe use https://github.com/cowboy/node-exit instead? (jest does)

packages/cli/src/utils/index.js
```diff
@@ -139,16 +139,20 @@ export function normalizeArg(arg, defaultValue) {
...
-
-    // TODO: Change this to a fatal error in v3
-    process.stderr.write(warningBox(msg))
+  if (timeout) {
+    const exitTimeout = setTimeout(() => {
+      let msg = `The command 'nuxt ${cmdName}' finished but did not exit after ${timeout}s\n`
+      msg += 'This is most likely not caused by a bug in Nuxt.js\n'
+      msg += 'Make sure to cleanup all timers and listeners you or your plugins/modules start.\n'
+      msg += 'Nuxt.js will now force exit\n\n'
+      msg += chalk.bold('DeprecationWarning: Starting with Nuxt version 3 this will be a fatal error')
```
manniL: Could we utilize a template string here?

pimlie: Yes and done

pimlie:
ready again

packages/cli/src/setup.js
```diff
@@ -26,17 +26,8 @@ export default function setup({ dev }) {
...
-          '',
-          chalk.white(String(logObj.args[0]))
-        ].join('\n'), {
-          borderColor: 'red',
-          borderStyle: 'round',
-          padding: 1,
-          margin: 1
-        }) + '\n')
-        process.exit(1)
+        process.stderr.write(fatalBox(String(logObj.args[0])))
```
manniL: Nit: Could we extract `logObj.args[0]` to a descriptive variable?

atinux (approved):
This is an awesome PR @pimlie 👏

pi0 (approved):

PR #4973, opened 2019-02-07
fix: dont force exit when it was explicitly disabled

pimlie:
fix: remove slash from warning text
fix: dont force-exit when explicitly disabled
chore: add force-exit tests for command

Follow-up of: #4958 

## Types of changes

## Description

As discussed on discord, some fixes and stability improvements for my pr from yesterday

## Checklist:

### DIFF
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -48,9 +48,12 @@ export default class NuxtCommand {
 
     const runResolve = Promise.resolve(this.cmd.run(this))
 
-    // TODO: For v3 set timeout to 0 when force-exit === true
-    if (!this.isServer || this.argv['force-exit']) {
-      runResolve.then(() => forceExit(this.cmd.name, forceExitTimeout))
+    if (
+      this.argv['force-exit'] ||
+      // dont force exit when it was explicitly disabled by the user
+      (!this.isServer && !this.isExplicitArgument('force-exit'))
+    ) {
+      runResolve.then(() => forceExit(this.cmd.name, this.argv['force-exit'] ? 0 : forceExitTimeout))
     }
 
     return runResolve
@@ -100,6 +103,10 @@ export default class NuxtCommand {
     const { Generator } = await imports.generator()
     const builder = await this.getBuilder(nuxt)
     return new Generator(nuxt, builder)
+  }
+
+  isExplicitArgument(option) {
+    return this._argv.includes(`--${option}`) || this._argv.includes(`--no-${option}`)
   }
 
   _getMinimistOptions() {
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -13,9 +13,6 @@ export default class NuxtCommand {
     }
     this.cmd = cmd
 
-    // If the cmd is a server then dont forcibly exit when the cmd has finished
-    this.isServer = cmd.isServer !== undefined ? cmd.isServer : Boolean(this.cmd.options.hostname)
-
     this._argv = Array.from(argv)
     this._parsedArgv = null // Lazy evaluate
   }
@@ -48,12 +45,9 @@ export default class NuxtCommand {
 
     const runResolve = Promise.resolve(this.cmd.run(this))
 
-    if (
-      this.argv['force-exit'] ||
-      // dont force exit when it was explicitly disabled by the user
-      (!this.isServer && !this.isExplicitArgument('force-exit'))
-    ) {
-      runResolve.then(() => forceExit(this.cmd.name, this.argv['force-exit'] ? 0 : forceExitTimeout))
+    if (this.argv['force-exit']) {
+      const forceExitByUser = this.isUserSuppliedArg('force-exit')
+      runResolve.then(() => forceExit(this.cmd.name, forceExitByUser ? false : forceExitTimeout))
     }
 
     return runResolve
@@ -105,8 +99,12 @@ export default class NuxtCommand {
     return new Generator(nuxt, builder)
   }
 
-  isExplicitArgument(option) {
+  isUserSuppliedArg(option) {
     return this._argv.includes(`--${option}`) || this._argv.includes(`--no-${option}`)
+  }
+
+  _getDefaultOptionValue(option) {
+    return typeof option.default === 'function' ? option.default(this.cmd) : option.default
   }
 
   _getMinimistOptions() {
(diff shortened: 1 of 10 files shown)


packages/cli/src/command.js
```diff
@@ -48,9 +48,12 @@ export default class NuxtCommand {
 
     const runResolve = Promise.resolve(this.cmd.run(this))
 
-    // TODO: For v3 set timeout to 0 when force-exit === true
-    if (!this.isServer || this.argv['force-exit']) {
-      runResolve.then(() => forceExit(this.cmd.name, forceExitTimeout))
+    if (
+      this.argv['force-exit'] ||
+      // dont force exit when it was explicitly disabled by the user
+      (!this.isServer && !this.isExplicitArgument('force-exit'))
```
manniL: Why `!this.isServer`? 🤔

packages/cli/src/options/common.js
```diff
@@ -28,7 +28,6 @@ export default {
       }
     }
   },
-  // TODO: Change this to default: true in Nuxt 3
```
manniL: Why removing that todo? 🤔

packages/cli/src/command.js
pimlie: Its to differentiate between build/generate and dev/start without having the need to change the default value of force-exit

pimlie:
As there appear to be some confusion (not in the least by myself), here is to clarify the force-exit behaviour:

- We want nuxt-cli to be able to forcibly exit after `NuxtCommand.run()` has finished. This behaviour can be toggled on the cli by using `--force-exit` and is _independent_ of the actual command that is run (its an abstract feature).
That means you can also run `nuxt start --force-exit` and then nuxt will immediately exit after it started listening. Maybe not really usefull normally but it could be used to quickly or automatically test whether the Nuxt server wa setup correctly and didnt threw any errors on server middleware etc.

- If you use `--force-exit` then process.exit will _always_ be called with exitCode `0`. This is because this way you have a way to ignore bad code and force eg your ci pipeline to continue. Besides, when there are fatal errors then nuxt will already exit with a non-zero exit code in the consola reporter so we dont have to check that here <sub>(but we might want to anyway in the future for non-fatal errors)</sub>.

- For some commands we want an additional check, `nuxt build` and `nuxt generate` are commands of which we know they should immediately exit after `NuxtCommand.run()` has finished. For this we added a (reversed) additional check on `this.isServer`. isServer means we _assume_ the command is starting a server and its normal the command keeps running after `NuxtCommand.run()` finishes. So the reverse is that when `!isServer` it shouldnt keep running so we only force-exit when `bin/nuxt-cli.js` doesnt exit by itself.

- For this additional check we _dont_ always want to call process.exit as it would hide bad code. So instead of calling process.exit immediately we set an unref timeout that calls process.exit. By unreferencing the timeout node will just normally exit when the event loop is empty, but if there are unreleased listeners still using the event loop the timeout will trigger after 5 seconds and still exit nuxt.

The difference between nuxt v2 and v3 will be that the additional check for `nuxt build/generate` is a warning and exits with code 0 in v2 while in v3 it will be a (fatal?) error and will exit with non-zero code.

packages/cli/src/command.js
```diff
@@ -102,6 +105,10 @@ export default class NuxtCommand {
     return new Generator(nuxt, builder)
   }
 
+  isExplicitArgument(option) {
```
pi0: Can be better to name it `hasExplicitArg` ?

pi0: or `argDisabled` or something more meaningful :)

pi0:
> * That means you can also run `nuxt start --force-exit` and then nuxt will immediately exit after it started listening. Maybe not really usefull normally but it could be used to quickly or automatically test whether the Nuxt server wa setup correctly and didnt threw any errors on server middleware etc.

This throws inappreciate warning/error message.

> * For some commands we want an additional check, `nuxt build` and `nuxt generate` are commands of which we know they should immediately exit after `NuxtCommand.run()` has finished. For this we added a (reversed) additional check on `this.isServer`

Two remarks:
- `isServer` check is not accurate. We rely on an implicit option to determine it.
- I think it could be better to explicitly enable `force-exit` by default in their cli options. And probably allow `--no-force-exit` as well.

pimlie:
> This throws inappreciate warning/error message.

I dont think it does? As mentioned above when you explicitly set `--force-exit` I dont set the timeout and you dont get the warning message. See here: https://github.com/nuxt/nuxt.js/blob/e84af4b79d289c9de8ab0ba196e02727d27fc5ef/packages/cli/src/utils/index.js#L143

> * `isServer` check is not accurate. We rely on an implicit option to determine it.
> * I think it could be better to explicitly enable `force-exit` by default in their cli options. And probably allow `--no-force-exit` as well.

By using `isServer` I wanted to differentiate between the two options, because using `--force-exit` should be _always call process.exit when cmd.run finishes_ and the additional check is _only force exit when nuxt-cli doesnt exit by itself_. It doesnt feel correct to me to use `--force-exit` for both situations.

Maybe we should otherwise add a cmd property?
```js
export default {
name: 'dev',
isServer: true,
run() ...
}
```

pi0:
@pimlie Exactly. I suggest allowing setting forceExit in `export default` of `build` and `generate` command that explicitly needs them. Less config and best behavior by default. Use can disable them if rare conditions by `--no-force-exit` them. And no more relies on `isServer`.

pimlie:
I added the possibility of making the default option value a fn. Good or not? Description is now a bit less descriptive because its the same description for both `--force-exit` as `--no-force-exit`.

packages/cli/src/options/common.js
```diff
@@ -30,8 +30,10 @@ export default {
   },
   'force-exit': {
     type: 'boolean',
-    default: false,
-    description: 'Force Nuxt.js to exit after the command has finished (this option has no effect on commands which start a server)'
+    default(cmd) {
+      return ['build', 'generate'].includes(cmd.name)
```
pi0: Good trick :+1:

pimlie:
Another day and another night thinking about this in bed. Or should we just leave the default option value and its description as is/was and do this:

```js
const cmdShouldFinish = ['build', 'generate'].includes(this.cmd.name)
const forceExitByUser = this.isUserSuppliedArg('force-exit')
if (this.argv['force-exit'] || (cmdShouldFinish && !forceExitByUser)) {
  runResolve.then(() => forceExit(this.cmd.name, forceExitByUser ? false : forceExitTimeout))
}
```

So that would actually be still some sort of a `isServer` variable but instead of looking at options.hostname it just checks the cmd.name. I like this better I think

pi0 (approved):

PR #4979, opened 2019-02-07
refactor: some small stuff

TheAlexLichter:


galvez (dismissed):
Love this style :)

packages/webpack/src/builder.js
```diff
@@ -226,9 +224,7 @@ export class WebpackBundler {
   }
 
   async unwatch() {
-    for (const watching of this.compilersWatching) {
-      await watching.close()
-    }
+    await Promise.all(this.compilersWatching.map(watching => watching.close()))
```
manniL: Not sure if order is important for unwatch?

pi0: no it isn't.

packages/webpack/src/utils/style-loader.js
```diff
@@ -37,39 +37,50 @@ export default class StyleLoader {
...
   }
 
   css(options) {
     options.exportOnlyLocals = this.exportOnlyLocals
-    return [
-      ...options.exportOnlyLocals ? [] : [this.styleLoader()],
-      { loader: 'css-loader', options }
-    ]
+
+    const cssLoader = { loader: 'css-loader', options }
```
clarkdo:
```suggestion
const cssLoader = { loader: 'css-loader', options }

if (options.exportOnlyLocals) {
  return [ cssLoader ]
}

return [ this.styleLoader(), cssLoader ]
```

manniL: Neat idea!

husayt (dismissed):
Really neat

packages/builder/src/builder.js
```diff
@@ -61,7 +60,7 @@ export default class Builder {
       })
 
       // Close hook
-      this.nuxt.hook('close', () => this.close())
+      this.nuxt.hook('close', this.close)
```
clarkdo: This may lose the context for `close`

manniL: Damn, right! I wasn't sure about that change and remembered that we talked about that before.

atinux (dismissed):
😍 
I love to see these refactors!

manniL:
I hope CI passes. If that's the case, I'm done 🙌

PR #4985, opened 2019-02-08
feat: lock project during build or generate 

pimlie:
## Types of changes

## Description

This PR adds support for locking as proposed in https://github.com/nuxt/rfcs/issues/23

I was not sure about the usefulness of locking when running `nuxt dev`, but will add it if you think it should also support it.

## Checklist:


packages/cli/src/utils/index.js
```diff
@@ -158,3 +159,7 @@ ${chalk.bold('DeprecationWarning: Starting with Nuxt version 3 this will be a fa
     exit(0)
   }
 }
+
+export function createLock(...args) {
```
pimlie: This might seem redundant but I had a problem with mocking this function when I immeditately exported it. Jest threw: `TypeError: Cannot set property createLock of #<Object> which has only a getter`

packages/cli/src/commands/generate.js
```diff
@@ -44,6 +45,25 @@ export default {
     config.build.analyze = false
 
     const nuxt = await cmd.getNuxt(config)
+
+    if (cmd.argv.lock) {
+      cmd.setLock(await createLock({
```
pimlie: Is it useful at all to create separate locks for the build and generate phases?

packages/utils/src/locking.js
```diff
@@ -0,0 +1,65 @@
...
+export const defaultLockOptions = {
+  stale: 30000,
+  onCompromised: err => consola.fatal(err)
+}
+
+export function getLockOptions(options) {
+  return Object.assign({}, defaultLockOptions, options)
+}
+
+export function createLockPath({ id = 'nuxt', dir, root }) {
```
pimlie: I added the `id` due to a remark of pi0 in the rfc about having separate locks for separate commands. I agree that that wouldnt be very useful, but having the possibility to have different locks on the same dir could be useful. Also the `id` is printed in the error when a lock already exists and could be useful to have a clearer understanding of why the lock exists in the first place

packages/cli/src/commands/generate.js
manniL: @pimlie As "build" is also included in the generate step, we *could* remove the lock after build for `nuxt build` and after generate for `nuxt generate` (if we can distinguish)

packages/utils/src/locking.js
```diff
@@ -0,0 +1,65 @@
...
+import consola from 'consola'
+import hash from 'hash-sum'
+import fs from 'fs-extra'
+import properlock from 'proper-lockfile'
+import onExit from 'signal-exit'
+
+export const lockPaths = new Set()
+
+export const defaultLockOptions = {
+  stale: 30000,
```
pimlie: There is no direct need for checking the validity of a lockfile when a lock already exists, `proper-lockfile` sets a timer which updates the lockfile regularly and if a lockfile exists it already checks for staleness. See here for more info: https://github.com/moxystudio/node-proper-lockfile#comparison

I have set a default value for staleness of 30s, but maybe thats a bit much and we can just use the default value.

packages/cli/src/commands/generate.js
pimlie: We could distinguish by either using different id's or using different lock dir's. Besides as a proof-of-concept I thought having separate locks might be useful in an edge-case where the generate phase takes quiet a while to run but in the mean time you want to already rebuild the project because you also want to run the nuxt server but with a slightly different configuration.

pimlie: If you want to distinguish between (similar) locks started by separate commands we could remove the lock `id` from the actual lockPath but then we should add a mapping between lockPaths and id's.

packages/cli/src/command.js
```diff
@@ -99,6 +103,18 @@ export default class NuxtCommand {
     return new Generator(nuxt, builder)
   }
 
+  setLock(lockRelease) {
+    if (lockRelease) {
```
pi0: Can we ensure former lock to be released here? (And possibly throwing an error)

packages/cli/src/utils/index.js
pi0: Please add a comment above explaining about the reason.

pi0 (commented):
Overall LGTM

packages/cli/src/command.js
```diff
@@ -1,4 +1,4 @@
-
```
pimlie: Maybe we could add an eslint rule for this?

packages/cli/src/command.js
pimlie: Dont but releasing the lock is async so now setLock is also async

pi0 (dismissed):

atinux (approved):

pi0 (approved):

husayt:
This is causing an error when building on netlify

```
12:26:56 PM: webpackbar 12:26:56 ℹ Compiling Client
12:27:18 PM: /opt/build/repo/node_modules/proper-lockfile/lib/lockfile.js:181
12:27:18 PM:         onCompromised: (err) => { throw err; },
12:27:18 PM:                                   ^
12:27:18 PM: Error: Unable to update lock within the stale threshold
12:27:18 PM:     at options.fs.utimes (/opt/build/repo/node_modules/proper-lockfile/lib/lockfile.js:109:21)
12:27:18 PM:     at FSReqWrap.oncomplete (fs.js:135:15)
```

pimlie:
You can run nuxt with `--no-lock` to disable locking.

Any idea if this is netlify specific or ci generic? Maybe we should turn locking of by default then when running on ci?

pimlie:
@husayt Could you maybe create a small reproduction that triggers the error? I tried adding timers in a build:before hook and asyncData (see here: https://github.com/pimlie/nuxt-issues/) but netlify still builds my site

husayt:
@pimlie seems like it is working now, but not sure if it is latest version or some of my workarounds. I will keep my eye on it and will let you know, if anything

pimlie:
@husayt Thanks for the feedback. It seems it was not just you, @aldarund also had issues with it on wercker. There appears to be an issue with proper-lockfile due to the staleness timer not updating the lock under heavy cpu load, but there might be another issue as well. Still investigating, so please let me know here or on discord if it happens again (and if you can help with a small reproduction that always triggers on netlify that would be very helpful)

husayt:
I have it again, this time with `nuxt-ts` and  `--no-lock` doesn't help either

PR #5004, opened 2019-02-11
fix: disable parallel build when extractCSS is enabled

### LINKED ISSUE #4965: Cannot read property 'outputOptions' of undefined
### Version

[2.4.3](https://github.com/nuxt.js/releases/tag/2.4.3)

### Reproduction link

[https://codesandbox.io/s/30k541yq35](https://codesandbox.io/s/30k541yq35)

### Steps to reproduce

* Create new Nuxt@^2.4.0
* Add `SCSS` support `npm install --save-dev node-sass sass-loader bootstrap bootstrap-vue`
* `Add `.scss` files in `assets/scss/`
* Set `extractCSS` and `parallel` to `true` in the `build` property of `nuxt.config.js`
* Add `~assets/scss/app.scss` in `css` array, for example, in `nuxt.config.js`

### What is expected ?

To work as expected.

Error in console: 
ERROR  in ./assets/scss/app.scss                    friendly-errors 19:20:49

Module build failed (from ./node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 9)
Cannot read property 'outputOptions' of undefined

    at Object.pitch (/sandbox/node_modules/extract-css-chunks-webpack-plugin/dist/loader.js:67:98)
                                                     friendly-errors 19:20:49
 @ ./.nuxt/App.js 3:0-33
 @ ./.nuxt/index.js
 @ ./.nuxt/client.js
 @ multi eventsource-polyfill webpack-hot-middleware/client?reload=true&timeout=30000&ansiColors=&overlayStyles=&name=modern&path=/__webpack_hmr/modern ./.nuxt/client.js}

And set `parallel` to `false`, CSS file didn't loaded.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8616">#c8616</a>)</em></sub></div>

stevefan1999-personal:
## Types of changes

- [ ] Breaking change (perhaps, it is a breaking change for the people who are broke with this before)

## Description
Fixes https://github.com/nuxt/nuxt.js/issues/4965#issuecomment-462375082, details can be seen in the comments

## Checklist:


clarkdo:
@stevefan1999-personal Thank you for the pr.

About the `thread-loader` context issue, I actually met the similar one which is solved by adding properties to `thread-loader` context, how do you think fix it in same way ?

https://github.com/webpack-contrib/thread-loader/pull/25

manniL:
Please fix the lint issues ☺️

stevefan1999-personal:
@clarkdo Nope, you still cannot access the rest of the webpack context:
https://github.com/faceyspacey/extract-css-chunks-webpack-plugin/blob/43e1672cd231301581df929a4159c7b716e64065/src/loader.js#L43-L46

So, unless `thread-loader` did an architectural change which allows master-worker communication, this will be philosophically impossible to fix.

pi0 (dismissed):

clarkdo:
@stevefan1999-personal Could you fix the broken code shown in failed test ?

packages/config/src/options.js
```diff
@@ -281,6 +281,11 @@ export function getNuxtConfig(_options) {
     consola.warn('vendor has been deprecated due to webpack4 optimization')
   }
 
+  if (options.build.parallel && typeof options.build.extractCSS !== 'undefined') {
```
```suggestion
  if (options.build && options.build.parallel && typeof options.build.extractCSS !== 'undefined') {
```

Maybe? ☺️

stevefan1999-personal: Although it is rare to not have `build` option, but still a potential read hazard, kudos.

clarkdo:
Actually error is here https://github.com/nuxt/nuxt.js/blob/master/packages/config/src/options.js#L285 since `extractCSS` is deleted

stevefan1999-personal:
@clarkdo what are the reasons those `typeof ... === 'undefined'` are needed instead of boolean coerce?

clarkdo:
@stevefan1999-personal There is no much difference in this case, you can use both. 

Performance is only slightly different https://jsperf.com/type-of-undefined-vs-undefined/30

clarkdo:
Now the issue is that this pr changed a little bit allchunks warn message in test/fixtures/with-config since it's using parallel build and extactCSS, maybe we should move the allchunks warn message test to extractcss fixtures.

packages/config/src/options.js
```diff
@@ -281,11 +281,17 @@ export function getNuxtConfig(_options) {
     consola.warn('vendor has been deprecated due to webpack4 optimization')
   }
 
+  if (options.build && options.build.parallel && typeof options.build.extractCSS !== 'undefined') {
+    delete options.build.extractCSS
```
clarkdo: This may be simple and no need to other checks
```suggestion
options.build.extractCSS = false
```

packages/config/src/options.js
```diff
@@ -281,14 +281,14 @@ export function getNuxtConfig(_options) {
...
+  // Disable CSS extraction due to incompatibility with thread-loader
+  if (options.build && options.build.parallel && options.build.extractCSS) {
     options.build.extractCSS = false
     consola.warn('extractCSS cannot work with parallel loaders due to missing webpack context')
   }
 
   // build.extractCSS.allChunks has no effect
-  if (typeof options.build.extractCSS !== 'undefined' &&
-      typeof options.build.extractCSS.allChunks !== 'undefined') {
+  if (options.build.extractCSS && options.build.extractCSS.allChunks) {
```
clarkdo: what is options.build.extractCSS.allChunks = false ?

packages/config/src/options.js
```diff
@@ -281,8 +281,14 @@ export function getNuxtConfig(_options) {
...
 
+  // Disable CSS extraction due to incompatibility with thread-loader
+  if (options.build && options.build.parallel && options.build.extractCSS) {
+    options.build.extractCSS = false
+    consola.warn('extractCSS cannot work with parallel loaders due to missing webpack context')
+  }
+
   // build.extractCSS.allChunks has no effect
-  if (typeof options.build.extractCSS.allChunks !== 'undefined') {
+  if (typeof options.build.extractCSS === 'object' && options.build.extractCSS.allChunks) {
```
clarkdo: I think you can just revert this part since `extractCSS` is not undefined by `false` after your previous change

clarkdo:
Sorry for the so many comments 🤖

stevefan1999-personal:
Okay. So I squashed those commits in hope to restore the CI test and it fails me again and again. I don't know why are there 503s and 500s. I'm completely speechless now.

clarkdo:
The test error is because extractcss is turned off in parallel mode.

But in with-config, it’s built with parallel mode and extractcss, so the expected warning message disappeared with your pr change, I’ve suggested that move this part test to extractcss fixture😺

clarkdo:
No worries, I’ll do the test chages for you, the current change LGTM

williamchong007:
Just my 2 cents
Shouldn't `extractCSS` be kept and `parallel` be disabled instead?
css affects the end product, and parallel just affects build time performance
Should we prefer to keep what user expect to get as a result, instead of a performance setting?

clarkdo:
@williamchong007 Good idea, I agree with you

PR #5027, opened 2019-02-13
perf: await routeData promises in parallel

TheAlexLichter:
Not sure if it makes a huge difference but running promises without dependencies in parallel here (and early return if the given arg is falsy) should work.
## Types of changes
- [x] Refactor

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -191,14 +191,8 @@ export async function setContext(app, context) {
       app.context.nuxtState = window.<%= globals.context %>
     }
   }
-  // Dynamic keys
-  app.context.next = context.next
-  app.context._redirected = false
-  app.context._errored = false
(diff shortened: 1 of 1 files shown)


pi0 (dismissed):

packages/vue-app/template/utils.js
```diff
@@ -193,13 +196,20 @@ export async function setContext(app, context) {
   app.context._redirected = false
   app.context._errored = false
   app.context.isHMR = !!context.isHMR
-  if (context.route) {
-    app.context.route = await getRouteData(context.route)
-  }
   app.context.params = app.context.route.params || {}
```
clarkdo: params and query need to be set after route data fetched

manniL: Oh, sure 😅

PR #5037, opened 2019-02-14
refactor: improve modern middleware and spa modern rendering

clarkdo:
## Types of changes

## Checklist:


packages/vue-renderer/src/renderer.js
```diff
@@ -438,15 +438,17 @@ export default class VueRenderer {
...
     if (context.spa === undefined) {
       // TODO: Remove reading from context.res in Nuxt3
-      context.spa = !this.SSR || context.spa || (context.req && context.req.spa) || (context.res && context.res.spa)
+      context.spa = !this.SSR || context.spa || req.spa || (context.res && context.res.spa)
     }
 
-    // context.modern
-    if (context.modern === undefined) {
-      context.modern = context.req ? (context.req.modernMode || context.req.modern) : false
+    // context.useModernRenderer
```
pi0: I think `context.modern` is more consistent with other context keys. Also allows passing an object instead of Boolean for behavior changing in future

clarkdo: It makes sense, I was just thinking it may be kind of confusing with options.modern.

pi0: Yes. That's also a note. But maybe this change worth for consistency.

awronski:
Off Topic: The modern browser list is now [hardcoded](https://github.com/nuxt/nuxt.js/blob/f0b00acbb0711568a5782c162c9d67617a937083/packages/server/src/middleware/modern-browsers.js).

Is there a risk, that some modern future is supported for example in Chrome 61 but not in Edge 16?
Or do you use the same list when transpiling the modern version (can't fint it in the source code)?

clarkdo:
@awronski The list is from babel-preset-env, the modern build is equal to babel-preset-env with `esMdoules: true`, so there is not conflict between them.

packages/vue-renderer/src/renderer.js
```diff
@@ -438,15 +435,17 @@ export default class VueRenderer {
...
     if (context.spa === undefined) {
       // TODO: Remove reading from context.res in Nuxt3
-      context.spa = !this.SSR || context.spa || (context.req && context.req.spa) || (context.res && context.res.spa)
+      context.spa = !this.SSR || context.spa || req.spa || (context.res && context.res.spa)
     }
 
     // context.modern
     if (context.modern === undefined) {
-      context.modern = context.req ? (context.req.modernMode || context.req.modern) : false
+      context.modern = req.modernMode && this.context.options.modern === 'server'
```
manniL: Do we always have `req` here?

TheAlexLichter (approved):

pi0 (approved):

Issue #4939, opened 2019-02-03
Fastify Custom Server


nahtnam:
### What problem does this feature solve?

It would be nice to have a custom server for https://www.fastify.io. It's really fast and pretty popular.

### What does the proposed changes look like?

There would be one more option in the nuxt cli to choose fastify as the server option.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8600">#c8600</a>)</em></sub></div>

pi0:
Nuxt.js uses the most minimal and low-level possible implementation, [connect](https://www.npmjs.com/package/connect). You can always compose the middleware stack with your own favorite server. BTW we may consider doing this for Nuxt3 because of native async/await support. Thanks for the issue.

emansayed913:
@ 1s

pi0:
(Comment by @sh  )

> We're running different services in our stack that all can share common middlewares (session, csrf) thanks to nuxt's connect compatibility.

> We're also using nuxt as middleware in our express server. This allows us to have automatic request tracing through [GCP Stackdriver Trace Agent for Node.js](https://github.com/googleapis/cloud-trace-nodejs#stackdriver-trace-agent-for-nodejs) which is a massive help for debugging and measuring issues in production. As it stands there is no fastify tracing plugin available, migrating to fastify (without the option of still using connect) would be a massive drawback for us.

> I'm not sure how much of a gain fastify (besides having a bit more req/s in benchmarks compared to connect/connect-router) really would be as it breaks the compatibility with the proven connect/express ecosystem.

> I'd vote for support of both frameworks (connect+fastify) or stay with connect if this change introduces too much complexity.

sh:
Thanks for copying this over @pi0 🎉 

I'd also like to add the fact that the [benchmarks](https://github.com/fastify/benchmarks#benchmarks) are running against a simple middleware responding with "hello world". 

I simply can't imagine a scenario where `connect` would become a performance bottleneck for nuxt. Running a separate app without nuxt to serve api / graphql requests is the way to go.

Most of the time SSR requests for simple pages are served < 100ms by nuxt. 

The slowest requests are pages with several components that require data to be prefetched via asyncData / vue-apollo. Our graphql endpoint responds < 200ms however ~600ms are spent populating data/ rendering and sending the response to the client.

nahtnam:
[...]

PR #5066, opened 2019-02-19
[experiment] feat: migrate to fastify

clarkdo:
Please use #4939 issue for discussion.

This is a proposal pull request for migrating connect to fastify, it's still WIP.

- [ ] If it's necessary and feasible to support `connect` and `fastify` at same time by dynamic import different server module like `server-connect` or `server-fastify`
- [ ] The future of custom server and programmatic usage
- [ ] serverMiddleware supports [fastify route](https://www.fastify.io/docs/latest/Routes/)


packages/server/src/listener.js
```diff
@@ -17,29 +13,25 @@ export default class Listener {
...
     this.address = null
     this.url = null
   }
 
   async close() {
     // Destroy server by forcing every connection to be closed
-    if (this.server && this.server.listening) {
-      await this.server.destroy()
+    if (this.app.server && this.app.server.listening) {
+      await this.close()
```
manniL: wait for `this.close` inside `this.close`? :thinking:

galvez:
It would be nice if we could have built-in JSON reading / writing. I'm currently using this:

```
import express from 'express'

const jsonMiddleware = express.json()

export default function(req, res, next) {
  res.json = (obj) => {
    res.write(JSON.stringify(obj))
  }
  return jsonMiddleware(req, res, next)
}
```

added to my `serverMiddleware`.

sh:
>  If it's necessary and feasible to support connect and fastify at same time by dynamic import different server module like server-connect or server-fastify

We're running different services in our stack that all can share common middlewares (session, csrf) thanks to nuxt's connect compatibility.

We're also using nuxt as middleware in our express server. This allows us to have automatic request tracing through [GCP Stackdriver Trace Agent for Node.js](https://github.com/googleapis/cloud-trace-nodejs#stackdriver-trace-agent-for-nodejs) which is a massive help for debugging and measuring issues in production. As it stands there is no fastify tracing plugin available, migrating to fastify (without the option of still using connect) would be a massive drawback for us.

I'm not sure how much of a gain fastify (besides having a bit more req/s in benchmarks compared to connect/connect-router) really would be as it breaks the compatibility with the proven connect/express ecosystem.

I'd vote for support of both frameworks (connect+fastify) or stay with connect if this change introduces too much complexity.

Atinux:
After talking with the team about it, I think we can close this one directly.

Thank you for your great feedback @sh

PR #5069, opened 2019-02-19
refactor: use webpack-dev-middleware options fs

clarkdo:
https://github.com/nuxt/nuxt.js/pull/5057#issuecomment-465193390


clarkdo:
Also need to set `fs` for server compiler separately, so it's not as clear as current code, close the pr then.

PR #5081, opened 2019-02-21
feat: stateful component injection via plugins

galvez:
This PR introduces a backwards-compatible extension to plugins in the form of a `componentInject()` helper, preceded by `inject()` in the plugin function signature. The difference between `inject()` and `componentInject()` is that `componentInject()` may only take a function as value (a _setter_), which in turn receives a reference to the *_current_* component.

```js
export default function(ctx, inject, componentInject) {
  componentInject('api', (component) => {
    return new Proxy({}, {
      get(_, apiMethod) {
        return async (...args) {
          // Look ma, altering current component state only!
          component.loading = true
          const result = await component.$axios.post(`/call/${apiMethod}`, { args })
          component.loading = false
          return result
        }
      }
    })
  })
}
```

TODO:

- [ ] Update tests
- [ ] Update docs

Calling for early feedback on the idea before I devote work on tests/docs.

### DIFF
--- a/packages/vue-app/template/index.js
+++ b/packages/vue-app/template/index.js
@@ -128,7 +128,13 @@ async function createApp(ssrContext) {
   })
 
   <% if (plugins.length) { %>
-  const _componentInject = function(key, setter) {
+  const componentInject = function(key, setter) {
+    if (!key) {
+      throw new Error('componentInject(key, setter) has no key provided')
+    }
+    if (typeof setteer !== 'function') {
+      throw new Error('componentInject(key, setter) has no setter provided')
+    }
     const installKey = '__<%= globals.pluginPrefix %>_' + key + '_installed__'
     // Keeping this check here so it never clashes with the global inject()
     if (Vue[installKey]) {
@@ -150,15 +156,12 @@ async function createApp(ssrContext) {
       }
     })
   }
-  const inject = function (key, value) {
+  const inject = function (key, value, setter) {
     if (!key) {
       throw new Error('inject(key, value) has no key provided')
     }
     if (typeof value === 'undefined') {
       throw new Error('inject(key, value) has no value provided')
-    }
-    if (typeof value === 'function') {
-      return _componentInject(key, value)
     }
     key = `$${key}`
     // Add into app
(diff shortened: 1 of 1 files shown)


nagipogu:
import { Injectable } from '@angular/core';
import { ConfigService } from './config.service';
import { Observable } from 'rxjs/Rx';
import { DataService } from './data.service';
import { CacheService } from './cache.service';

@Injectable({
  providedIn: 'root'
})
export class GetListService {

  constructor(private configService: ConfigService,
    public dataService: DataService,
    public cacheService: CacheService,) { }

    getOccupationList(obj) {
      return this.dataService.postData(this.configService.OCCUPATION_LIST, obj)
        .flatMap(rData => {
          console.log('--Occupation List----------->')
          console.log(rData)
          if (this.cacheService.CashedOccupationList) {
            this.cacheService.CashedOccupationList = _.concat(this.cacheService.CashedOccupationList, rData.Occupationlist)
          } else {
            this.cacheService.CashedOccupationList = rData.Occupationlist;
          }
          return Observable.of(rData)
        });
    } 
}

nagipogu:
this is code i have been trying
 this.cacheService.CashedOccupationList = _.concat(this.cacheService.CashedOccupationList, rData.Occupationlist)

here i am getting error.

nagipogu:
error is

ERROR in src/app/services/get-list.service.ts(22,54): error TS2304: Cannot find name '_'.                   src/app/services/get-list.service.ts(36,51): error TS2304: Cannot find name '_'.                            src/app/services/get-list.service.ts(50,48): error TS2304: Cannot find name '_'.                            src/app/services/get-list.service.ts(64,55): error TS2304: Cannot find name '_'.
--

galvez:
@nagipogu Please open an issue or try our [Discord channels][https://discord.nuxtjs.org/], commenting on a PR is not the appropriate way to seek help. I've deleted your comments here.

manniL:
@galvez Could you elaborate some use cases (besides changing a component's loading state)? ☺️

galvez:
@Atinux @manniL The use-case is the same as mixins, but lighter. Say you want to make sure a method is available for every component but Vue.mixin would be too costly: `componentInject()` to the rescue. It will only do anything if you try and use it (since it uses `defineProperty`).

galvez:
After discussing this with @Atinux I've decided to offer it as an external utility.

galvez:
@Atinux @pi0 @manniL just found another use for this:

https://twitter.com/anothergalvez/status/1110622070881353729

Do you think there's a more elegant way to achieve the same without `componentInject()`?

pi0:
@galvez Sweet use case! But really can't be a component mixin? A nuxt module/plugin can be then able to register that mixin.

galvez:
@pi0 wouldn't this perform better than Vue.mixin()? I mean, wouldn't `Vue.mixin()` cause every Vue instance to have an individual copy of the method? With `componentInject()`, we ensure it's only defined once?

pi0:
I think it's already handled by Vue itself for global mixins. The _copy_ is happening for function reference not entirely copying.

galvez:
Right, still one could argue _safely_ injecting into `Vue.prototype` performs better than duplicating function references around? (I'm really curious about this) :)

PR #5092, opened 2019-02-22
fix(ts): improve tsconfig.json compatibility

ykzts:
## Types of changes

## Description

It became possible to resolve a Node.js package for the `"extends"` field since TypeScript v3.2 (`tsconfig.json` inheritance via Node.js packages](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#tsconfigjson-inheritance-via-nodejs-packages)). But vetur 's langserver is running under TypeScript v3.2 ([package.json#L40](https://github.com/vuejs/vetur/blob/v0.16.2/server/package.json#L40). Therefore, when using vetur it will not be resolved correctly.

## Checklist:

### DIFF
--- a/examples/pm2-typescript/tsconfig.json
+++ b/examples/pm2-typescript/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "@nuxt/typescript",
+  "extends": "./node_modules/@nuxt/typescript/tsconfig",
   "compilerOptions": {
     "baseUrl": ".",
     "types": [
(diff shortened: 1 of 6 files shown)


kevinmarrec:
Hmm, as there was lof of issues using `extends`, I chose to replace it by a `defaultsDeep` strategy in https://github.com/nuxt/nuxt.js/pull/5079.

I think that `tsconfig.json` should be directly reflect what options are used and not hide them through `extends`, at least in a Nuxt project.

galvez:
@kevinmarrec Should we close this PR then? :)

ykzts:
I think it will be fixed at # 5079.

kevinmarrec:
@ykzts Yes ! Nevertheless, thanks for your time !

PR #5102, opened 2019-02-25
feat: support `devModules` option

galvez:
As per discussion in #core -- @pi0 does this look right to you?


packages/core/src/module.js
```diff
@@ -16,6 +16,11 @@ export default class ModuleContainer {
     // Call before hook
     await this.nuxt.callHook('modules:before', this, this.options.modules)
 
+    if (this.options._start) {
```
pi0: I think this condition should be like `if (!...)`

pi0: PS: Seems good idea to run devModules **before** modules :+1:

galvez: Ops, updated the check -- will try and add a test next :)

manniL:
I really like the idea but I'm not sure if it should be implemented that way, especially because if we want to change it later, we'd have to wait for the next major.

Looking at https://github.com/nuxt/rfcs/issues/10, how will this fit in the concept there? 🤔 Shouldn't we discuss this idea in the RFC / as part of the RFC first?

Wouldn't it be nice to give the module itself an option to automatically register only in dev/prod/both?

Not sure if an extra array in the config this the right approach for it. 
Maybe an option instead?

pi0:
@manniL I taught about module improvement RFC as well. But this is a different matter. A module to be able to decide to register or not have to be _installed_ first which means have to be always in "dependencies" section of package.json.

manniL:
@pi0 I know what you mean, yes. I solved this in the past via checking NODE_ENV manually (rel: https://twitter.com/TheAlexLichter/status/1085179308606349314/photo/1).

Anyway, having `modules` and `devModules` feels a bit redundant. Might be just me though 🤷‍♂️

pi0:
@manniL NODE_ENV of `nuxt build` and `nuxt start` is both `production` :)

manniL:
> @manniL NODE_ENV of `nuxt build` and `nuxt start` is both `production` :)


Right, but not for `nuxt dev` (which I thought what that feat it was for).

So, I hope I get this right now:

`devModules` -> modules that are need for `build` (no matter if  `nuxt dev` or `nuxt generate` or `nuxt build`) but not needed for `nuxt start`?

pi0:
Yeah. Basically `modules`/`devModules` is mapped to `dependencies`/`devDependencies`. Modules like `@nuxt/pwa` which are only needed for build/dev are inside `devDependencies` and `devModules` :) BTW I see your point.

manniL:
@pi0 Ahh, got it now! That makes sense but is confusing when it comes to "running in dev/prod mode"

pi0:
I think would be better to add it as a non documented feature for now. Modules that support this mode (pwa for example breaks for spa) can advice users to use correct section. Still I'm open to better wording for devModules :)

pi0 (dismissed):

manniL:
I thought about using a nested object but that doesn't seem sexy as well :D

```js
{
  modules: {
    dev: [],
    prod: [],
    all: [] // or so?
  }
}
```

pi0:
@manniL is really more confusing. Also somehow breaking change without patch on `options.js`

manniL:
@pi0 Yeah, that was my thought too.

Atinux:
Side note: We should also be able to handle it here when importing modules: https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/module.js#L113

Maybe with a check like:
```
if (`handler.devModule` && `this.options._start`) {
  return
}
```

pi0:
@atinux a non-dev module should never call `this.addModule(someDevModule)` 🙈

galvez:
@Atinux good call, I'll update and add tests tonight.

galvez:
> ```js
> if (handler.devModule && this.options._start) {
>   return
> }
> ```

@Atinux @pi0 For now I've added this bit:

```
    if (this.options.devModules.includes(handler) && this.options._start) {
      return
    }
```

Since we can't know if an external module is flagged with `devModule` without first requiring it. Advice welcome, but perhaps we should add the flag as part of the module improvements RFC?

-- Jonas

pi0 (dismissed):
Good

Atinux:
Failing tests:
screenshot 2019-02-28 at 11 30 07

You may want to add `devModules: []` before this line https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/config/_app.js#L25

😄

galvez:
@Atinux weird, if you see the PR changelog, you'll see I added it already. I'll revisit now.

pi0 (approved):

Issue #127, opened 2017-01-13
Async data on components


danieloprado:
Loading data with axios only work in a page not in a component:
See this example: https://github.com/danieloprado/nuxt-async-data

![image](https://cloud.githubusercontent.com/assets/7683575/21931277/cbb3b5aa-d981-11e6-87fd-c86ca69fa54a.png)

```js
[Vue warn]: data functions should return an object:
https://vuejs.org/v2/guide/components.html#data-Must-Be-a-Function 
(found in component <category-header-component> at /home/daniel/projects/tegra-ecommerce-site/app/components/CategoryLinks.vue)
warn @ vue.runtime.common.js:521
initData @ vue.runtime.common.js:1901
initState @ vue.runtime.common.js:1848
Vue._init @ vue.runtime.common.js:3365
VueComponent @ vue.runtime.common.js:3485
createComponentInstanceForVnode @ vue.runtime.common.js:2761
init @ vue.runtime.common.js:2771
createComponent @ vue.runtime.common.js:4122
createElm @ vue.runtime.common.js:4065
updateChildren @ vue.runtime.common.js:4335
patchVnode @ vue.runtime.common.js:4398
updateChildren @ vue.runtime.common.js:4314
patchVnode @ vue.runtime.common.js:4398
patch @ vue.runtime.common.js:4523
Vue._update @ vue.runtime.common.js:2497
(anonymous) @ vue.runtime.common.js:2464
get @ vue.runtime.common.js:1661
run @ vue.runtime.common.js:1730
flushSchedulerQueue @ vue.runtime.common.js:1544
(anonymous) @ vue.runtime.common.js:473
nextTickHandler @ vue.runtime.common.js:422
vue.runtime.common.js:521 

[Vue warn]: Property or method "title" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option. 
(found in component <category-header-component> at /home/daniel/projects/tegra-ecommerce-site/app/components/CategoryLinks.vue)
```

Atinux:
Hi @danieloprado 

As I mentioned in https://github.com/nuxt/nuxt.js/issues/32#issuecomment-264509187, nuxt.js only supercharges the `data` method in the pages components.

I tried different ways to make the sub components asynchronous but sadly I did not find any way of doing it.

laosb:
@Atinux How about supporting SSR layout async data? That's would be helpful for fetching some config from backend.

Atinux:
[...]

Issue #32, opened 2016-12-02
Async data in nested component


Xowap:
Hi,

It seems that loading async data is pretty easy from pages, yet I don't find any way to load data asynchronously from nested components.

To explain what I try to do, I made some example project: https://github.com/Xowap/nuxttest

You'll see that data loaded from `index.vue` is correctly rendered server-side yet data in `HttpBin.vue` isn't displayed at all and triggers an error.

Is that an anti-pattern? A feature to come?

Thanks,
Rémy

Atinux:
Actually it is not possible because it's not linked to a route, Nuxt.js surcharges the component data() associated to a route to allow async data.

For sub components, there are 2 ways of achieving it:
1. Making the API call in the mounted() hook and setting the data afterwards, downside: no server rendering
2. Making the API call in the data() of the page component and giving the data as a prop to the subComponent: server rendering OK. But the data() of the page might be less readable because it's loading the async data of the sub components

It all depends if you want the sub components to be server-rendered or not.

I hope it solves your issue :)

Xowap:
This is kind of what I feared.

I've been reading Vue 2's documentation, it looks like Vuex becomes basically mandatory. Which also solves this problem. I guess the real clean solution is to use Vuex, right?

Atinux:
Yes of course I forgot about vuex, you can integrate the store easily with Nuxt.js, just add a store/index.js (look at examples/vuex-store)

danieloprado:
@Xowap Can you provide an example how implement it?

visualcookie:
An example would really help.

Etheryte:
@Atinux Could you elaborate why this is not possible? All the components should be known beforehand statically and it should be possible to make the appropriate calls with the given context. I'm looking into how to solve the issue myself, too, as it would tremendously improve modularization.

Atinux:
Hi @Etheryte 

Sadly, there is no way to know the component three inside `lib/app/server.js`, I can only get the routes components and update them by calling the `asyncData` & `fetch` methods. I tried to find a way for it but it seems impossible right now.

But it's also a good thing to not call asyncData for "normal" components since it let's everyone create Vue component compatible with the same behaviour under nuxt.js and a basic vue app.

pea3nut:
Use this function. That will test all child component `.fetch` and call it.

```javascript
function fetchDeep(component){
    var originFn =component.asyncData;
    component.asyncData =async function (ctx) {
        if(component.components){
[...]

PR #5118, opened 2019-02-26
feat(vue-app): new fetch

### LINKED ISSUE #3776: Allow layouts or the error layout to use fetch/asyncData

My error page lives in a `layout` that needs to pull data for menus/footer, since layouts can't have fetch/asyncData blocks I normally add the fetch logic to the pages that use a specific layout. However this workaround doesn't work for the error page since it is not a page, but a layout.

This can't be done on NuxtServerInit because:

a) requests dispatched in nuxtServerInit are blocking, i.e. nuxt waits until the promise returned from nuxtServerInit is resolved before continuing, which makes things x2 slower when a page has other requests to make (all requests can and should be run in parallel)
b) different layouts have different fetch needs
c) wouldn't work on SPA mode

Do you guys have any workaround for this? Currently, my 404 error pages render with an empty footer and menu.

Add `fetch` and/or `asyncData` support to layouts, or at least the error layout.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7606">#c7606</a>)</em></sub></div>

atinux:
## RFC

Please carefully read our RFC for it: https://github.com/nuxt/rfcs/issues/27

## Description

> To see an example in action, please take a look at the [examples/v3/fetch/](https://github.com/nuxt/nuxt.js/blob/feat/async-data/examples/v3/fetch/README.md) with a detailed explanation.

This PR will allow to use `fetch` in every Vue components with access to `this` inside the method. I've been waiting since the creation of Nuxt.js to have this feature, and I am really happy that we worked on a workaround with the core team (thanks Evan & Guillaume Chau) and now shipped with Vue 2.6: `serverPrefetch`.

`serverPrefetch` is great for the SSR part. The complicated part is when navigating on client-side since Nuxt 1 & 2 **wait** for the `asyncData/fetch` calls to be done before rendering the next route (by using `router.beforeEach`).

It introduces at the component level:
-  `fetch` hook in the page/component to be used while asyncData is called, it should be a `string` for a global component or directly a component object
-  `this.$isFetching` in the component to display a placeholder while `fetch` is being called (on client-side or when `fetchOnServer` is  `false` when rendering the component on the server).
-  `fetchOnServer` (default to `true`) to call or not `fetch` hook when server rendering the component
- `this.$fetch()` in the component to re-call `fetch` hook (`$isFetching` will become `true` again)
- `this._lastFetchAt` (exists if `fetch` hook is present): timestamp of the last fetch call made (ssr included), useful for caching combined with `keep-alive` on `<nuxt>` and `<nuxt-child>` components
- `this._hydrated` to know if this component has been hydrated and `fetch` has not been called on client-side.

It also introduces at the app level:
- `this.$nuxt.state` (before: `nuxtState` given in `beforeNuxtRender`) that you can overwrite and will be rendered as `window.__NUXT__`. The value is also setup on client side to hydrate your plugins.
- `this.$nuxt.isFetching` to know if any component is actually calling `fetch` (useful for a global loading page)
- `this.$nuxt.nbFetchin` to know how many `fetch` are pending

It removes:
- `asyncData`: Since you can mutate the components data right inside `fetch`, you don't need to have `asyncData` anymore. Also, Nuxt smartly detects what properties has been mutated in order to returned the minimum JSON with the HTML.

### What about waiting `fetch` calls before navigating to the next page?

I found a way to wait  before rendering the next route (like actual Nuxt UX), but it was too tricky:
Using `<transition>` listeners `beforeEnter + enter` and `leave` to **wait before transitioning the component** while initializing the virtual DOM and calling fetch.
-> `keep-alive` and page transitions were broken (since I was hacking <transition>).

I also tried to overwrite the `render` method of the component by giving a component to a  `placeholder` property. I also worked on a `<placeholder>` block next to the `<template>` block (I worked with @znck on it).
-> It looked more messy that simply deal with `$isFetching` and it did not fix the way to "wait" for the components to be fetched before showing the next route.

Some people may say that it's the same UX of React Suspense, I did not know, I had this feature in mind since a long time (I was calling it "Reactive Loading" 😄).

Resolves #3776, #32 and #127

TODO:
- [ ] `$emit('hook:fetch')`, see https://vuedose.tips/tips/10/
- [ ] `preFetch(ctx)` to be able to add some cache in `this.$options`?

## Checklist:


galvez:
giphy

jagreehal:
Just so I understand correctly does this resolve https://github.com/nuxt/nuxt.js/issues/32?

manniL:
@jagreehal Yes.

lilianaziolek:
@Atinux These changes in fetch which seem great, I always missed the ability to put fetch of stuff in components not just pages. 
One perhaps related question is how fetch is supposed to work in regards to route params (i.e. products/_category) - mind, I don't mind route query (products?category=XXX). It seems to me that there is currently no way to make fetch be called on route params changes (it can be done on route query changes with the extra flag). I've been using watch $route for this purpose but it feels like fetch should react to changes in params too, as they are actually very likely to require change data in the component/page. Otherwise I end up with some weird workarounds to avoid duplicating code between fetch and watch.
Can you please 
1) Confirm that this is the case and I'm not missing something, and it is correct that fetch cannot react on params changes right now (if not correct - what have I missed?)
2) Was there any reason for this design decision? Will this reason still stand with the changes described above, or perhaps this could be improved as well?

Atinux:
Hi @lilianaziolek 

Actually, when a component is re-used and a route params change, Vue will call `beforeMount` hook that will automatically call `fetch` 😄 
-> YES, `fetch` will react to **params changes**

Also, for sub components, you can directly `watch` for `$route` or directly give props to the components, example: `<Post :post-id="$route.params.id" />`

lilianaziolek:
> Hi @lilianaziolek
> 
> Actually, when a component is re-used and a route params change, Vue will call `beforeMount` hook that will automatically call `fetch` 😄
> -> YES, `fetch` will react to **params changes**

@Atinux
Hmmm, you mean this already should work this way in Nuxt 2? In this case something really odd is happening in our project. We have the (I think) standard structure with pages/products/_category/_subcategory folders, and I placed console.logs in fetch and watch - when I first hit a category, I can see the fetch log statement with no issues. It appears on server side for first hit, or on client side if I navigate from another page (e.g. if I go from product/id1 to products/category1, it works fine). However, if I directly switch between categories (i.e. if I go from products/category1 to products/category2), it does not trigger that log statement (i.e. fetch is not called) - only watch is called. If that's not expected then maybe I can see if I can reproduce this in a smaller project...

Unless you have a hunch of a config somewhere that could be causing such behaviour? (for example a flag we might have accidentally changed...)

Atinux:
Well, this is because you are using the "old" fetch (current implementation with Nuxt 2).

If you want to try the new fetch, the only way right now is to:
1. Clone nuxt.js repo
2. Run `yarn`
3. `yarn examples/v3/fetch/`
4. And play with `examples/v3/fetch/` files :)

lilianaziolek:
> Well, this is because you are using the "old" fetch (current implementation with Nuxt 2).
> 
> If you want to try the new fetch, the only way right now is to:
> 
> 1. Clone nuxt.js repo
> 2. Run `yarn`
> 3. `yarn examples/v3/fetch/`
> 4. And play with `examples/v3/fetch/` files :)

:) Haha, I think we had a misunderstanding (maybe due to late edits) :) My question was "can this be achieved in Nuxt 2 as well ?" and you replied "-> YES, fetch will react to params changes". - but seems you were referring to new nuxt? So, to clarify
* in Nuxt 2, fetch is NOT called where *only* route params change, it is called on route changes (and route query, if flag is set).
* In Nuxt 3, fetch will can be called in all 3 situations.
If that's correct, then that's great news, we're looking forward to Nuxt 3 (this, amongst many other reasons!)

hartmut-co-uk:
> What about waiting fetch calls before navigating to the next page?

So there are no plans to support await fetch before route change?

Question: If there's a route change to e.g. `/nuxt/nuxt.js/pulls` -> `/nuxt/nuxt.js/pull/5118` - with the PR #5118 to be requested via API, async, via `fetch`, ..
How to deal with errors, e.g. if the post doesn't exist (API response 404). Within `fetch` would you then call `error` which would redirect (delayed) to -> error page, 404 not found?
What about the intermediate loading state? 
And what about the browser history? (first a seemingly valid route update to `/nuxt/nuxt.js/pull/5118`, then a redirect to `/404`)

I think I'd have scenarios where I'd prefer to await at least the main 'route specific' API response before the actual route change should take place. Does that make sense?

henriqemalheiros:
> So there are no plans to support await fetch before route change?

@hartmut-co-uk there is [this example](https://github.com/nuxt/nuxt.js/tree/feat/async-data/examples/v3/async-data) of `asyncData` being supported via a separate Nuxt module. I have some projects which where it would be better to keep the current behaviour, so this looks promising.

@Atinux I like @hartmut-co-uk's idea of waiting for the "route specific" API response. Maybe we could await `fetch` only on page components? Maybe even put it behind a flag, like `loading`, set as `false` by default.

Atinux:
Thank you for your feedbacks @hartmut-co-uk and @henriqemalheiros 

I just created an RFC for it so we can discuss about this situation: https://github.com/nuxt/rfcs/issues/27

tsongas:
I'm just starting to learn Nuxt so I have feedback from that perspective. I'm stoked on Vue for bringing more simplicity and fun to front end development. I was stoked on Nuxt for similar reasons, until I got to Vuex which is when it became less fun. I think React developers are figuring out Redux was an important step in the development of state management, but want less boilerplate and are moving to alternatives like MobX-State-Tree. So, I would hate to see Nuxt continue to follow what has turned out to be not such a developer friendly solution to state management in Redux/Vuex if there are better alternatives. See https://hackernoon.com/the-react-state-museum-a278c726315 for a rundown. To me the most exciting is https://github.com/jamiebuilds/unstated, I wonder if it could help inspire a way to simplify state management and fetching data with Nuxt, especially universal which gets confusing.

radiantstatic:
I think this is a great idea. Been using nuxt for past year or so and loving it. One concern I've had is the inability to fetch on an error page which causing headaches galore when attempting to navigate away from the page. It is the last thing in my mind thats keeping us from pushing to production. Are there any reasonable work arounds until this gets dropped? Thanks for all of your efforts @Atinux

Atinux:
Thank you for your feedback @tsongas 

Simplicity is part of Nuxt vision, this is why we have the `store/` directory that simplify Vuex usage (see https://nuxtjs.org/guide/vuex-store).

Actually, `fetch` is not related to Vuex and works without it.
The next step is indeed to simplify the way of calling asynchronous data while being as simple (and powerful) as possible :)

husayt:
@tsongas it might be slightly bumpy to get to understand vuex at first, but it is not even closely as complex as Redux. In fact, with Vuex they managed to simplify what normally is a very complex concept - *State management*. Nuxt takes it to next level by providing it hassle and setup free. 

What I could suggest is to use it with [vuex-class](https://github.com/ktsn/vuex-class) (via [nuxt-property-decorator](https://github.com/nuxt-community/nuxt-property-decorator) or directly). 

Rather than moving away from Vuex, which has become a standard in Vue community, we should seek to make it richer and simpler to use. I am really looking forward to Vuex 4 and wish it could be prioritised, as it covers many of your concerns:

Here is some of the coming features from [Vue.js roadmap](https://github.com/vuejs/roadmap)

### Vuex 4.x

- Simplify usage
  - Getting rid of mapXXX helpers via scoped-slot based store consumer component
  - Getting rid of the need for separating actions and mutations
- Better TypeScript inference (may require API redesign)
- Accompanying DevTools update
- Actionable: kickoff public design thread

MLDMoritz:
This would allow keep-alive to not wait for the asyncData but loading it 'on-the-fly' - am I right? So getting back to the page before won't trigger fetch?

Atinux:
I am closing this PR in favour of https://github.com/nuxt/nuxt.js/pull/6880 since it does not remove the actual behaviour 😄

PR #5157, opened 2019-03-04
refactor: remove builder coupling from server

pi0:
Changes:
- Move middleware for dev (devMiddleware and hotMiddleware) into `@nuxt/builder:builder.middleware()` and just keep a placeholder inside `@nuxt/server`
- Don't call `nuxt.ready()` in constructor anymore
- Don't create a server during `nuxt build` or when `options.server` is `false`


Atinux:
😍😍😍

galvez:
Good stuff, I can see where you're going with this :)

galvez (dismissed):

packages/webpack/src/builder.js
```diff
@@ -216,11 +216,22 @@ export class WebpackBundler {
...
+
+  async middleware(req, res, next) {
+    const name = req.modernMode ? 'modern' : 'client'
+
+    if (this.devMiddleware && this.devMiddleware[name]) {
+      await this.devMiddleware[name](req, res)
+    }
+
+    if (this.hotMiddleware && this.hotMiddleware[name]) {
+      await this.hotMiddleware[name](req, res)
```
manniL: Is the order meaningful here? If not, could we get a perf bonus when using `Promise.all` here?

pi0: Middleware are sequential by design. Even if it works we are doing an anti pattern. Also only one of them matches a request so the difference. Combining this with the fact that node is single threaded, there should be no difference.

clarkdo:
Did I see Chinese ? @pi0 😄

PR #5163, opened 2019-03-05
fix(builder, vue-app): order of plugin execution based on order in array

aldarund:
## Types of changes

## Description
Order of plugin execution based on order in plugins array independent of mode

## Checklist:


packages/builder/src/builder.js
```diff
@@ -235,6 +235,8 @@ export default class Builder {
     // Plugins
     this.plugins = Array.from(this.normalizePlugins())
 
+    this.resolvePluginsMode()
```
clarkdo: We can move the call to https://github.com/nuxt/nuxt.js/blob/17259d5f617214cac140e51809958bdcc41322ac/packages/builder/src/builder.js#L275

packages/builder/src/builder.js
```diff
@@ -542,6 +544,20 @@ export default class Builder {
     )
   }
 
+  resolvePluginsMode() {
```
clarkdo: The same code in `resolvePlugins` should be removed.
And this method could be async and wrapped with Promise.all.

clarkdo (commented):
And a new test case in `with-config.test.js` that verify the `plugin.client.js` will not be executed in server side would be better 😄

packages/builder/src/builder.js
aldarund: Oh, yes, it should be removed, it was in fact, but then i debugged and forgot to remove again :)
@clarkdo why async? It just iterate over plugins with map

clarkdo:
Actually even `plugin.client.js` is executed in server side, but it's already replaced as an empty file via webpack alias in https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/builder.js#L65-L77

packages/builder/src/builder.js
clarkdo: All the resolving can be in parallel for better performance

aldarund:
@clarkdo the problem is on client. ANd the problem is about order of execution, not that server code executed on client or vice versa.
Basically u can open generated .nuxt/index.js and see something like this for .client.js plugins at top

`import nuxt_plugin_test2client_4591c188 from 'nuxt_plugin_test2client_4591c188' // Source: ..\\src\\plugins\\test2.client.js (mode: 'all')
`

clarkdo:
Oh, I got you, but if the execution order should be same as config array order, I think this change still cannot fix all problems since ‘all’ plugins are always executed before client and server in template/index, maybe you should change that part instead of bringing up resolving

packages/vue-app/template/index.js
```diff
@@ -166,20 +166,21 @@ async function createApp(ssrContext) {
...
-
-  <% if (plugins.filter(p => p.mode === 'server').length) { %>
-  if (process.server) {
-    <% plugins.filter(p => p.mode === 'server').forEach((plugin) => { %>
-    if (typeof <%= plugin.name %> === 'function') await <%= plugin.name %>(app.context, inject)<% }) %>
-  }<% } %>
+  <% if (plugins.length) { %>
+    <% plugins.forEach((plugin) => { %>
+      <% if (plugin.mode == 'client') { %>
+        if (process.client) {
```
clarkdo: This part would be better to be follow eslint rules in compiled template.

pi0: Can be also combined with next condition

aldarund: @pi0 but if we combine wont it lead to not treeshaking?
@clarkdo ye, i though its run automatically during tests but seems not :)

pi0: `typeof` guard is to prevent executing plugins when not exporting a function.

Webpack would compile condition to something like `if (false && typeof ...)` which should be eliminated by minifier.

Tree shaking cannot happen because we don't set side-effects field for above imports BTW it is zero overhead as server plugins as aliased to an empty module for client-side.

clarkdo: @pi0 I think import of `empty.js` can be tree-shaked after dead code removal, @aldarund you may test it that if empty.js is removed with you changes.

pi0: True. The server code is not included in the client bundle regardless of using one line or two separate `if`s.

clarkdo: And the reason that client code is not tree-shaked is `await` is transpiled into:

```js
if (true) {
  _context2.next = 33;
  break;
}

return plugin(app.context, inject);
```

So the plugin call isn't removed.

aldarund: changed into one statemement and styled for compiled template

packages/vue-app/template/index.js
```diff
@@ -166,20 +166,23 @@ async function createApp(ssrContext) {
...
-    <% plugins.filter(p => p.mode === 'client').forEach((plugin) => { %>
-    if (typeof <%= plugin.name %> === 'function') await <%= plugin.name %>(app.context, inject)<% }) %>
-  }<% } %>
-
-  <% if (plugins.filter(p => p.mode === 'server').length) { %>
-  if (process.server) {
-    <% plugins.filter(p => p.mode === 'server').forEach((plugin) => { %>
-    if (typeof <%= plugin.name %> === 'function') await <%= plugin.name %>(app.context, inject)<% }) %>
-  }<% } %>
+  <% if (plugins.length) { %>
```
pi0: Is this check required? Calling for each on a zero length array should add no new lines

aldarund: ye, true story :) meaningless optimization

aldarund: changed

pi0: NP. Thanks :)

pi0 (approved):

clarkdo:
An update on release and doc would be great 🤗

PR #5186, opened 2019-03-08
feat(vue-renderer): use async fs

pi0:
## Types of changes

## Description

Refactors to use async FS operations for vue-renderer. Some over-optimizations for early cleanup of MFS removed to reduce complexity, more consistency (not `isMFS`) and less dependency on webpack specific MFS implementation.

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -183,8 +183,12 @@ export default class VueRenderer {
     // Load templates
     await this.loadTemplates()
 
-    // Call createRenderer if any resource changed
+    // Detect if any resource updated
     if (updated.length > 0) {
+      // Invalidate assetsMapping cache
+      delete this._assetsMapping
+
+      // Create new renderer
       this.createRenderer()
     }
 
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -467,15 +467,18 @@ export default class VueRenderer {
         transform: async (src, { readResource }) => {
           const serverManifest = JSON.parse(src)
 
-          const resolveAssets = async (obj) => {
-            for (const name in obj) {
-              obj[name] = await readResource(obj[name])
-            }
-            return obj
+          const readResources = async (obj) => {
+            const _obj = {}
+            await Promise.all(Object.keys(obj).map(async (key) => {
+              _obj[key] = await readResource(obj[key])
+            }))
+            return _obj
           }
 
-          const files = await resolveAssets(serverManifest.files)
-          const maps = await resolveAssets(serverManifest.maps)
+          const [files, maps] = await Promise.all([
+            readResources(serverManifest.files),
+            readResources(serverManifest.maps)
+          ])
 
           // Try to parse sourcemaps
           for (const map in maps) {
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -1,6 +1,6 @@
 import path from 'path'
 import crypto from 'crypto'
-import fs from 'fs'
+import fs from 'fs-extra'
 import consola from 'consola'
 import devalue from '@nuxt/devalue'
 import invert from 'lodash/invert'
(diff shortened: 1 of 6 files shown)


packages/vue-renderer/src/renderer.js
```diff
@@ -141,34 +141,28 @@ export default class VueRenderer {
...
 
-    const readResource = (fileName, encoding) => {
+    const readResource = async (fileName, encoding) => {
       try {
         const fullPath = path.resolve(distPath, fileName)
-        if (!_fs.existsSync(fullPath)) {
+        if (!await _fs.exists(fullPath)) {
           return
         }
-        const contents = _fs.readFileSync(fullPath, encoding)
```
clarkdo: No need cleaning ?

packages/vue-renderer/src/renderer.js
```diff
@@ -470,22 +464,18 @@ export default class VueRenderer {
...
+            for (const name in obj) {
+              obj[name] = await readResource(obj[name])
+            }
             return obj
           }
 
-          const files = resolveAssets(serverManifest.files, oldValue.files)
-          const maps = resolveAssets(serverManifest.maps, oldValue.maps)
+          const files = await resolveAssets(serverManifest.files)
+          const maps = await resolveAssets(serverManifest.maps)
```
manniL: Maybe a `Promise.all` here?

packages/vue-renderer/src/renderer.js
pi0: Cleaning only (potentially) saves a few KBs (Bigger projects, few MBs) of RAM. But adds the complexity of oldAssets. Also if we want to support independent server/webpack restarts we need correct MFS snapshot.

I thought the cleaner code with fewer hacks worth it. What do you think?

clarkdo: I agree for most resources.

Only `_assetsMapping`, if we don't clear the old object, it may have wrong mappings after rebuilding.

packages/vue-renderer/src/renderer.js
pi0: Done.

packages/vue-renderer/src/renderer.js
pi0: Oh you are right. We have to invalidate that mapping when loading assets. Does it look good?

https://github.com/nuxt/nuxt.js/pull/5186/files#diff-015dff3d45b52153ef5f24ef104e954bR189

clarkdo: Great, I always love your code, bro, neat refactoring!

clarkdo:
We should `await` all `build:resources` hook call as well 😃

pi0:
@clarkdo Promises are chained for hooks (Arrow function return shorthand)

packages/webpack/src/utils/async-mfs.js
```diff
@@ -0,0 +1,24 @@
...
+
+const syncRegex = /Sync$/
+
+const propsToPromisify = Object.getOwnPropertyNames(MFS.prototype).filter(n => syncRegex.test(n))
+
+for (const prop of propsToPromisify) {
+  const asyncProp = prop.replace(syncRegex, '')
+  const origAsync = AsyncMFS.prototype[asyncProp]
+
+  AsyncMFS.prototype[asyncProp] = function (...args) {
```
clarkdo: If it would be better that use `async` for this function ?

pi0: Good idea :) Makes code cleaner.

pi0: Oops no we can't. Because of webpack callback support with promisifed version :( MFS async functions  are internally called by webpack too.

clarkdo:
@pi0 I mean the call of hook, like [here](https://github.com/nuxt/nuxt.js/blob/ce6e2cdc7c544ef87605490441ff59dd0c1eb295/packages/webpack/src/builder.js#L134), but it seemed all calls are `await`, so only removing isMFS is enough.

PR #5188, opened 2019-03-08
fix: keep-alive component data should not be updated

### LINKED ISSUE #5166: keep-alive not working, resets data after change tab
### Version

[v2.4.5](https://github.com/nuxt.js/releases/tag/v2.4.5)

### Reproduction link

[https://codesandbox.io/s/9z8p69427r](https://codesandbox.io/s/9z8p69427r)

### Steps to reproduce

1. Click on ROEL USER button to go to Profile tab, which is an url (/roelmagdaleno/profile)
2. Go to Posts or Payments tab, when the component renders for the first time, it will show you the data.
3. Go back to Profile and then go to posts or payments again.
4. The current data that should kept alive has been reset to the initial values, an empty array.

### What is expected ?

Keep alive the current tab data, so it won't make another API call.

The data is reset when user change the tab and enter again.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8776">#c8776</a>)</em></sub></div>

clarkdo:
## Types of changes

## Description

Fix #5166

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -491,6 +491,7 @@ function fixPrepatch(to, ___) {
       if (
         instance.constructor._dataRefresh &&
         Components[i] === instance.constructor &&
+        instance.$vnode.data.keepAlive !== true &&
         typeof instance.constructor.options.data === 'function'
       ) {
         const newData = instance.constructor.options.data.call(instance)


packages/vue-app/template/client.js
```diff
@@ -491,6 +491,7 @@ function fixPrepatch(to, ___) {
       if (
         instance.constructor._dataRefresh &&
         Components[i] === instance.constructor &&
+        instance.$vnode.data.keepAlive !== true &&
```
pi0: Just to be sure, is the default value of `keepAlive`, `false` by default? If it is true, isn't this a minor breaking change?

clarkdo: The default value is `undefined`, it's just a flag in Vue for `keepAlive` cached component.

I don't think it's a breaking change, it should be a bug because `component` under `keep-alive` should be cached and never reload data as Vue is doing.

pi0 (approved):

clarkdo:
Thanks, bro.

packages/vue-app/template/client.js
ppsshh: Just wondering, shouldn't it be `data.props.keepAlive` instead of `data.keepAlive`? Because in bunch of files (`packages/vue-app/template/components/nuxt.js`, etc), `keepAlive` property is being wrapped into `props` object.

I ran into an issue when component is still trying to update its data() even though it has `keep-alived` enabled.
Changing this line to `data.props.keepAlive` solved problem for me.

PS: Nuxt v2.15.2 (seems like nothing have been changed with keep-alived behaviour ever since this commit)

clarkdo: keepAlive here is checking vue keep-alive component which will add keepAlive to vnode.data https://github.com/vuejs/vue/blob/b51430f598b354ed60851bb62885539bd25de3d8/src/core/components/keep-alive.js#L120

ppsshh: Thanks for your reply.

Sorry, but what should be the keep-alive component in here? Because debugger says that `instance` is a page component. Also, `instance.$vnode.data` doesn't have `keepAlive` property, but it does have `props.keepAlive`.

Screenshot:

clarkdo: I think the component in your screenshot is NuxtChild which has keep-alive prop for enabling keepAlive on it.

https://nuxtjs.org/docs/2.x/features/nuxt-components#keep-alive

ppsshh: You right, it is NuxtChild component.
I thought it should work the same way for any components, but seems like it is not the case.
I'll try to investigate it a little bit further and maybe come back with new issue and/or PR.
Thanks for your comments 👍

PR #5195, opened 2019-03-10
feat(generate) return non-zero code or page error (#4991)

### LINKED ISSUE #4991: `nuxt generate` should exit with non-zero code on errors
### Version

### Reproduction link

[https://github.com/njam/nuxt-generate-error](https://github.com/njam/nuxt-generate-error)

### Steps to reproduce

If an exception is thrown during rendering of a page in `nuxt generate`, this error is printed, but still the script exits with status 0. For both handled and unhandled errors. This is especially problematic in CI, as unhandled errors go unnoticed.

The repo to reproduce the error just throws an error in `created()`:
```js
  created() {
    throw new Error("My error");
  }
```

Clone repo to reproduce:
```console
git clone git@github.com:njam/nuxt-generate-error.git
cd nuxt-generate-error
yarn install
yarn run generate
echo $?
```

### What is expected ?

Should exit with a code higher than 0.

Exits with code 0.

### Additional comments?

As a workaround I'm currently registering a modules like this:
```js
export default function() {
  this.nuxt.hook('generate:routeFailed', ({ route, errors }) => {
    let error = errors[0].error
    throw new Error(`Error generating ${route}:\n${error}`)
  })
}
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8629">#c8629</a>)</em></sub></div>

loomchild:
Introduce --fail-on-page-error option that will cause nuxt generate to return non-zero status code when it encounters a page generation error. This is useful to let CI know that something went wrong.

## Types of changes

## Description

Resolves: #4991 
This PR is a result of the discussion under this ticket and the fact that nuxt-generate-cluster implements similar feature: see https://github.com/nuxt-community/nuxt-generate-cluster/pull/17.

This feature allows `nuxt generate` to return non-zero status code in case of page generation error. This is useful to let CI know that something went wrong, otherwise such issues might get unnoticed for a long time. 

In order to avoid changing existing behavior and introducing a breaking change I am proposing to add a parameter `--fail-on-page-error`. Only if this parameter is present the error status code will be returned.

## Checklist:

Note - I am not sure if it requires documentation update and where I should update it, but I am happy to do it if needed. For the moment I have added a option description which will be available when running `nuxt generate --help`.

### DIFF
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -71,12 +71,12 @@ export default {
 
     const generator = await cmd.getGenerator(nuxt)
 
-    const errors = await generator.generate({
+    const generateResult = await generator.generate({
       init: true,
       build: cmd.argv.build
     })
 
-    if (cmd.argv['fail-on-error'] && errors.errors.length > 0) {
+    if (cmd.argv['fail-on-error'] && generateResult.errors.length > 0) {
       throw new Error('Error generating pages, exiting with non-zero code')
     }
   }
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -34,7 +34,7 @@ export default {
         }
       }
     },
-    'fail-on-page-error': {
+    'fail-on-error': {
       type: 'boolean',
       default: false,
       description: 'Exit with non-zero status code if there are errors when generating pages'
@@ -76,7 +76,7 @@ export default {
       build: cmd.argv.build
     })
 
-    if (cmd.argv['fail-on-page-error'] && errors.errors.length > 0) {
+    if (cmd.argv['fail-on-error'] && errors.errors.length > 0) {
       throw new Error('Error generating pages, exiting with non-zero code')
     }
   }
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -33,6 +33,11 @@ export default {
           options.modern = 'client'
         }
       }
+    },
+    'fail-on-page-error': {
+      type: 'boolean',
+      default: false,
+      description: 'Exit with non-zero status code if there are errors when generating pages'
     }
   },
   async run(cmd) {
@@ -66,9 +71,13 @@ export default {
 
     const generator = await cmd.getGenerator(nuxt)
 
-    await generator.generate({
+    const errors = await generator.generate({
       init: true,
       build: cmd.argv.build
     })
+
+    if (cmd.argv['fail-on-page-error'] && errors && errors.errors && errors.errors.length > 0) {
+      throw new Error('Error generating pages, exiting with non-zero code')
+    }
   }
 }
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -71,12 +71,12 @@ export default {
 
     const generator = await cmd.getGenerator(nuxt)
 
-    const generateResult = await generator.generate({
+    const { errors } = await generator.generate({
       init: true,
       build: cmd.argv.build
     })
 
-    if (cmd.argv['fail-on-error'] && generateResult.errors.length > 0) {
+    if (cmd.argv['fail-on-error'] && errors.length > 0) {
       throw new Error('Error generating pages, exiting with non-zero code')
     }
   }
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -76,7 +76,7 @@ export default {
       build: cmd.argv.build
     })
 
-    if (cmd.argv['fail-on-page-error'] && errors && errors.errors && errors.errors.length > 0) {
+    if (cmd.argv['fail-on-page-error'] && errors.errors.length > 0) {
       throw new Error('Error generating pages, exiting with non-zero code')
     }
   }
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/generate.js
```diff
@@ -33,6 +33,11 @@ export default {
           options.modern = 'client'
         }
       }
+    },
+    'fail-on-page-error': {
+      type: 'boolean',
+      default: false,
+      description: 'Exit with non-zero status code if there are errors when generating pages'
```
loomchild: I am open for naming suggestions for this option.

packages/cli/src/commands/generate.js
```diff
@@ -66,9 +71,13 @@ export default {
...
 
-    await generator.generate({
+    const errors = await generator.generate({
       init: true,
       build: cmd.argv.build
     })
+
+    if (cmd.argv['fail-on-page-error'] && errors.errors.length > 0) {
+      throw new Error('Error generating pages, exiting with non-zero code')
+    }
```
loomchild: I was considering to do `throw String` as it is done in run.js - please let me know what's better.

I could also check if `errors` and `errors.errors` is not null, but for the moment from what I see it should always be present. Please let me know if you prefer to add this extra validation.

packages/cli/test/unit/generate.test.js
```diff
@@ -134,4 +134,20 @@ describe('generate', () => {
...
+    await expect(cmd.run()).rejects
+  })
+
+  test('do not throw an error when fail-on-page-error disabled and page errors', async () => {
+    mockGetNuxt()
+    mockGetGenerator(() => ({ errors: [{ type: 'dummy' }] }))
+
+    const cmd = NuxtCommand.from(generate, ['generate', '.'])
+    await cmd.run()
+  })
```
loomchild: I decided to test on unit level since it's a simple feature for the moment - please let me know if other e2e tests would be interesting to add.

packages/cli/src/commands/generate.js
pi0: I think this can be simply `fail-on-error` to fail on any kind of error. It may happen on middleware, store or plugins too.

packages/cli/src/commands/generate.js
pi0: Thanks for the review. Can you please update `generate.generate` to return `{ errors: [] }` ?

packages/cli/test/unit/generate.test.js
pi0: I think that would be more than enough. E2E test would be somehow costly.

packages/cli/src/commands/generate.js
loomchild: Sorry, I don;t think I understand. You want to modify `generator.generate` function (it already returns `{errors: []}` if there are no errors)? Or do you want me to return `errors` from `generate.run` instead of throwing an error?

pi0: I wanted to ensure that generate() function always returns the same data structure like generateResult with `errors` array. So we don't need to check it here.

loomchild: Ah. In fact it does, and the check is not ncecessary. However, I would need to update the mock to behave the same manner. I will do it.

loomchild: Done.

packages/cli/src/commands/generate.js
loomchild: Sorry, I missed this comment. Renamed.

packages/cli/src/commands/generate.js
```diff
@@ -66,9 +71,13 @@ export default {
 
     const generator = await cmd.getGenerator(nuxt)
 
-    await generator.generate({
+    const errors = await generator.generate({
```
pi0: @loomchild Last request! :see_no_evil:  Can we rename this variable from `errors` to `generateResult` or something similar?

loomchild: Sure, you are right.

loomchild: Done.

pi0 (dismissed):

packages/cli/src/commands/generate.js
pimlie: I think we should strive to always halt on errors, but as a non-zero exit code for page errors would be a breaking change on eg CI deployments we cant do that now until v3. But as said before, then even with v3 you still want to differentiate between page errors and other errors. Unless its ok that `--no-fail-on-error` only doesnt fail page errors. But might be confusing.

packages/cli/src/commands/generate.js
```diff
@@ -66,9 +71,13 @@ export default {
 
     const generator = await cmd.getGenerator(nuxt)
 
-    await generator.generate({
+    const generateResult = await generator.generate({
```
pimlie: Can we destruct this into `const { errors }` ?

loomchild: Sure, done. 
It looks like I forgot one test where `mockGetGenerator` was used, but it worked since fail-on-error was disabled. Fixed.

pi0 (approved):

loomchild:
Yay, my first PR to Nuxt. Thanks guys!

pimlie:
@loomchild Great work, that many more may follow! :)

PR #5214, opened 2019-03-12
fix(vua-app): clone mount error to prevent mutating read-only error object

### LINKED ISSUE #5056: Attempted to assign to readonly property.
### Version

### Reproduction link

[https://github.com/nuxt/nuxt.js/blob/f791d786e0996e4cad2b1ddbe244a747e7e700aa/packages/vue-app/template/client.js#L81](https://github.com/nuxt/nuxt.js/blob/f791d786e0996e4cad2b1ddbe244a747e7e700aa/packages/vue-app/template/client.js#L81)

### Steps to reproduce

Throw a error while mounting Nuxt (client)

### What is expected ?

To adjust the error message

A new error is thrown saying: `Attempted to assign to readonly property.`

### Additional comments?

https://github.com/nuxt/nuxt.js/blob/f791d786e0996e4cad2b1ddbe244a747e7e700aa/packages/vue-app/template/client.js#L81

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8681">#c8681</a>)</em></sub></div>

TheAlexLichter:
## Types of changes

Resolves #5056 

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -81,14 +81,9 @@ const errorHandler = Vue.config.errorHandler || console.error
 createApp()
   .then(mountApp)
   .catch((err) => {
-    const improvedErr = new Error('[nuxt] Error while mounting app: ' + err.message)
-    improvedErr.original = err
-
-    if(improvedErr.hasOwnProperty('stack')) {
-      improvedErr.stack = improvedErr.stack.split('\n').slice(0, 2).join('\n') + '\n' + err.stack
-    }
-
-    errorHandler(improvedErr)
+    const wrapperError = new Error(err)
+    wrapperError.message = '[nuxt] Error while mounting app: ' + wrapperError.message
+    errorHandler(wrapperError)
   })
 
 function componentOption(component, key, ...args) {
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -81,8 +81,10 @@ const errorHandler = Vue.config.errorHandler || console.error
 createApp()
   .then(mountApp)
   .catch((err) => {
-    err.message = '[nuxt] Error while mounting app: ' + err.message
-    errorHandler(err)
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/client.js
```diff
@@ -81,8 +81,10 @@ const errorHandler = Vue.config.errorHandler || console.error
 createApp()
   .then(mountApp)
   .catch((err) => {
-    err.message = '[nuxt] Error while mounting app: ' + err.message
-    errorHandler(err)
+    const improvedErr = new Error('[nuxt] Error while mounting app: ' + err.message)
+    improvedErr.original = err
+    improvedErr.stack = improvedErr.stack.split('\n').slice(0,2).join('\n') + '\n' + err.stack
```
aldarund: I think stack might be undefined.
```
x = new DOMException()
DOMException
x.stack
undefined
```

aldarund: When err.stack undefined will it look good?

pi0: This is a very low-level hack which may make error handler even more unstable for different browsers / errors

manniL: Good catch!

manniL: Could be, yeah. But What would be another solution? Checking for a DOMException (and other errors with read-only values)? :thinking:

pi0:
@manniL Isn't `new Error(error)` enough to workaround readonly object?

packages/vue-app/template/client.js
aldarund: Another thing is that in IE for example improvedErr.stack would be undefined and crash
```
new Error().stack
undefined
```

manniL:
Welp, that was too simple :joy:

You are right @pi0 :relaxed:

aldarund (approved):

pi0 (approved):

PR #5215, opened 2019-03-12
fix(postcss): default to preset-env and cssnano last

TheAlexLichter:
## Types of changes

## Description

Currently, `cssnano` is moved at the end of the Postcss plugin chain. This totally makes sense. However, as Nuxt includes `postcss-preset-env` by default which contains `autoprefixer`, it should come second last to apply prefixing to all previous plugin outputs (e.g. `tailwindcss`).

## Checklist:


pi0 (approved):

pi0:
Nice find @manniL

PR #5221, opened 2019-03-13
chore: faster ci

pi0:
Currently, CircleCI uses run-in-band to prevent previous inconsistencies. This makes build step too long (~5min) let's try if it is better now :)

### results 

- `-i` => `~5min`
- Without worker limit => ENNOMEM


pi0:
-w2 failed too :(

PR #5236, opened 2019-03-14
fix(webpack): clone `config.entry`

### LINKED ISSUE #4849: Specify custom webpack entry points for the client and server

Currently specifying `config.entry` in `build.extend(config)` is ignored.

The **client** entry is assigned without merging the value of `config.entry` that might have been set in `build.extend`:

https://github.com/nuxt/nuxt.js/blob/dcdbaba710c88c2f2b4c848aaa5f881074764556/packages/webpack/src/config/client.js#L178

The **server** `config.entry` is the result of merging a default config—overriding any `config.entry` value passed in `build.extend`:

https://github.com/nuxt/nuxt.js/blob/89f88669c10ce9100b109749aeab445638b851a8/packages/webpack/src/config/server.js#L71

It would be great if custom entry points could be specified in `build.extend` so that custom chunks could be generated independently of the Nuxt app.

Specify a custom entry name of `app` could throw an error notifying the user that this is reserved by Nuxt.

```js
// nuxt.config.js
import { resolve } from "path"

export default {
  build: {
    extend(config, { isClient }) {
      if (isClient) {
        console.log(config.entry) // { app: [ "path/to/client.js", "eventsource-polyfill", "etc..." ] } (currently undefined)
      } else {
        console.log(config.entry) // { app: [ "path/to/server.js" ] } (currently undefined)
      }
      
       // This is ok
      config.entry.custom = resolve("custom-entry.js")
      
      // Throws error "entry.app is reserved by Nuxt"
      // Could use Object.defineProperty for the
      // config object passed to build.extend to
      // throw an error when attempting to set entry.app
      config.entry.app = resolve("app.js") 
    }
  }
}
```

Running `dev|build|generate` would create another chunk called `custom`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8543">#c8543</a>)</em></sub></div>

juliendargelos:
## Types of changes

## Description
Assign to webpack config entry instead of overwriting.
Allow custom entries to be specified in `build.extend`.
Resolves: #4849

## Checklist:
- [ ] I have added tests to cover my changes.

### DIFF
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -182,9 +182,9 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       `${querystring.stringify(hotMiddlewareClientOptions)}&path=${clientPath}`.replace(/\/\//g, '/')
 
     // Entry points
-    config.entry = {
+    config.entry = Object.assign({}, config.entry, {
       app: [path.resolve(buildDir, 'client.js')]
-    }
+    })
 
     // Add HMR support
     if (this.dev) {
--- a/packages/webpack/src/config/server.js
+++ b/packages/webpack/src/config/server.js
@@ -69,9 +69,9 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
     Object.assign(config, {
       target: 'node',
       node: false,
-      entry: {
+      entry: Object.assign({}, config.entry, {
         app: [path.resolve(this.buildContext.options.buildDir, 'server.js')]
-      },
+      }),
       output: Object.assign({}, config.output, {
         filename: 'server.js',
         libraryTarget: 'commonjs2'


packages/webpack/src/config/client.js
```diff
@@ -182,9 +182,9 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       `${querystring.stringify(hotMiddlewareClientOptions)}&path=${clientPath}`.replace(/\/\//g, '/')
 
     // Entry points
-    config.entry = {
+    config.entry = Object.assign({}, config.entry, {
```
pi0: Would you please use rest operator here? `{ ...config.entry, `

juliendargelos: I wanted to be consistent with existing style:

https://github.com/nuxt/nuxt.js/blob/89f88669c10ce9100b109749aeab445638b851a8/packages/webpack/src/config/server.js#L74-L77

pi0: That's ok. will refactor later :)

pi0 (approved):

PR #5243, opened 2019-03-14
fix(server): handle decodeURI errors

### LINKED ISSUE #4977: URI malformed
### Version

### Reproduction link

[https://nuxtjs.org/%c1%81](https://nuxtjs.org/%c1%81)

### Steps to reproduce

1) Load https://nuxtjs.org/%c1%81
2) See console error

### What is expected ?

Trigger 404/500 errors

Execution stops

### Additional comments?

(node:87717) UnhandledPromiseRejectionWarning: URIError: URI malformed
    at decodeURI (<anonymous>)
    at nuxtMiddleware (node_modules/@nuxt/server/dist/server.js:120:15)

Perhaps the decodeURI should be wrapped in the try {} block?
https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/middleware/nuxt.js#L10

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8619">#c8619</a>)</em></sub></div>

phof:
## Types of changes

## Description

Resolves #4977

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -11,21 +11,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
   try {
     url = decodeURI(req.url)
-  } catch (err) {
-    /* istanbul ignore if */
-    if (context && context.redirected) {
-      consola.error(err)
-      return err
-    }
-
-    const e = new Error('URL malformed')
-    e.name = 'Bad Request'
-    e.statusCode = 400
-    return next(e)
-  }
-
-  res.statusCode = 200
-  try {
+    res.statusCode = 200
     const result = await renderRoute(url, context)
     await nuxt.callHook('render:route', url, result, context)
     const {
@@ -97,6 +83,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
       return err
     }
 
+    if(err.name === 'URIError') err.statusCode = 400
     next(err)
   }
 }
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -83,7 +83,9 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
       return err
     }
 
-    if (err.name === 'URIError') err.statusCode = 400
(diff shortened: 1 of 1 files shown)


phof:
@manniL I've included a try/catch for the `decodeURI` function.

I'm not immediately clear if there's a way to pass the error's name/message to the server-side error template, as this should result in a "Client Error" (400 response status), not a "Server Error".

packages/server/src/middleware/nuxt.js
```diff
@@ -7,7 +7,22 @@ import { getContext } from '@nuxt/utils'
...
   // Get context
   const context = getContext(req, res)
-  const url = decodeURI(req.url)
+  let url
+
+  try {
+    url = decodeURI(req.url)
+  } catch (err) {
+    /* istanbul ignore if */
+    if (context && context.redirected) {
```
manniL: You probably don't need that block here as it'll be caught further down the road.

packages/server/src/middleware/nuxt.js
```diff
@@ -7,7 +7,22 @@ import { getContext } from '@nuxt/utils'
...
+      consola.error(err)
+      return err
+    }
+
+    const e = new Error('URL malformed')
```
manniL: I'd love to see a more descriptive variable name here :relaxed:

manniL: Maybe you could use the error coming from the failed decoding directly anyway? :thinking: 

(see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#Catching_errors)

TheAlexLichter (changes requested):

packages/server/src/middleware/nuxt.js
phof: Moved everything in the same `try` block and setting `statusCode = 400` in the `catch`.

pi0 (approved):

japboy:
this change prevents to control application level error handling for malformed urls and forces me to show error pages to users...

PR #5251, opened 2019-03-15
feat: loading screen

atinux:
## Types of changes

## Description

When running `nuxt`, you won't be able to access the url until the resources are available. **This is really frustating** and can also display this kind of error:
53035161-b50ab180-346c-11e9-8705-3d0942bf766d

After working with @pi0, we introduce a loading screen for Nuxt, a GIF is way better to understand 😄 
**Note:** I am using the `-o` option at the end to automatically open the browser window

nuxt-loading-screen

## Checklist:


packages/server/src/middleware/nuxt.js
```diff
@@ -12,6 +12,10 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
   res.statusCode = 200
   try {
     const result = await renderRoute(url, context)
+
+    // If result === false, rendered a loading screen
+    if (!result) return
```
manniL:
```suggestion
    if (!result) {
      return
    }
```

No inline-ifs! :stuck_out_tongue_winking_eye:

packages/vue-renderer/src/loading.html
```diff
@@ -0,0 +1,280 @@
...
+          modern: {
+            progress: 0,
+            status: 'Bundling...'
+          }
+        }
+      }
+    },
+    created () {
+      this.ws = null
+      this.autoReconnectInterval = 1000 // 1 second
```
manniL:
```suggestion
      this.autoReconnectInterval = 1000 // in milliseconds
```

In case this value will be changed in the future, the comment can stay "as is"

packages/vue-renderer/src/loading.html
```diff
@@ -0,0 +1,280 @@
...
+        console.log('WebSocket(client): connecting...')
+        this.ws = new WebSocket(this.url)
+        this.ws.onopen = () => {
+          console.log('WebSocket(client): connected')
+          this.connected = true
+        }
+        this.ws.onclose = (e) => {
+          this.connected = false
+          switch (e.code){
+            case 1000: // CLOSE_NORMAL
```
manniL: Suggestion: To avoid magic numbers, use constants named by the code names (like `CLOSE_NORMAL`? :thinking:

packages/vue-renderer/src/loading.html
```diff
@@ -0,0 +1,280 @@
...
+            case 1005:
+              console.log('WebSocket(client): closed')
+              break;
+            default: // Abnormal closure
```
manniL:
```suggestion
            default: // Unkown error
```

Does that fit better? :thinking:

packages/vue-renderer/src/loading.html
```diff
@@ -0,0 +1,280 @@
...
+        document.write(html)
+        document.close()
+        if (typeof console.clear === 'function') {
+          console.clear()
+        }
+      }
+    },
+    filters: {
+      capitalize (value) {
+        if (!value) return ''
```
manniL:
```suggestion
        if (!value) {
          return ''
        }
```

TheAlexLichter (dismissed):
Some feedback :relaxed:

packages/vue-renderer/src/loading.html
```diff
@@ -0,0 +1,280 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0, minimum-scale=1.0">
+  <title>Nuxt.js: Loading app...</title>
+  <link href="https://fonts.googleapis.com/css?family=Quicksand:400,500" rel="stylesheet">
+  <script src="https://cdn.jsdelivr.net/npm/vue"></script>
```
pimlie: Would it be possible to just include external dependencies? It doesnt feel correct to me that in `dev` mode you are required to have an internet connection to make something like a loading screen appear as intended. I know that for 98+% of the people this would probably be ok anyway, but I am a strong believer that you should be able to develop offline without strange artifacts. Would be a perfect case for https://github.com/nuxt/rfcs/issues/22 and embedded resources ;)

packages/server/src/middleware/nuxt.js
pi0: Done

packages/vue-renderer/src/loading.html
pi0: Interval is in most of the cases _considered_ to be type of ms and `1 second` makes it more readable.

packages/vue-renderer/src/loading.html
pi0: These are [standard](https://tools.ietf.org/html/rfc6455#section-11.7) numbers. I've added a comment above of this for reference.

packages/vue-renderer/src/loading.html
pi0: done

packages/vue-renderer/src/loading.html
pi0: done

packages/vue-renderer/src/loading.html
pi0: @pimlie You are right. We thought about it as well with @Atinux. The problem is that serving from `node_modules` requires extra complexities. Let's see if we can find a way for this PR or not. But finally, it should be served from local FS.

pi0:
LoaderScreen moved to its own dedicated repository (because of packaging requirements)

https://github.com/nuxt/loading-screen

pi0 (dismissed):

clarkdo (dismissed):

galvez (dismissed):

TheAlexLichter (dismissed):

Atinux:
I also updated the `nuxt` loading indicator for SPA:
nuxt-loading-indicator

It supports a `theme` key to `'dark'` for the dark background :)
carbon (3)

pi0 (approved):

PR #5279, opened 2019-03-19
[v1] chore(vue-renderer): move title above meta

NightDiRaven:
## Types of changes

## Checklist:

### DIFF
--- a/lib/core/meta.js
+++ b/lib/core/meta.js
@@ -51,8 +51,8 @@ module.exports = class MetaRenderer {
     meta.BODY_ATTRS = m.bodyAttrs.text()
     // HEAD tags
     meta.HEAD =
-      m.meta.text() +
       m.title.text() +
+      m.meta.text() +
       m.link.text() +
       m.style.text() +
       m.script.text() +
--- a/lib/core/renderer.js
+++ b/lib/core/renderer.js
@@ -348,8 +348,8 @@ module.exports = class Renderer {
     }
     const m = context.meta.inject()
     let HEAD =
-      m.meta.text() +
       m.title.text() +
+      m.meta.text() +
       m.link.text() +
       m.style.text() +
       m.script.text() +


manniL:
@NightDiRaven 	Please explain your proposed change properly :relaxed:

NightDiRaven:
@manniL Its really only for some dumb CEO, they request proper order (i dunno why). In ver 2 of nuxt all in title -> meta order. But in 1.x its meta -> title
before
image 
after
image

pi0:
Is this PR targeting nuxt 1.x? 🤔

Atinux:
Shouldn't the title be after `<meta charset="utf-8">`?

visualfanatic:
@Atinux exactly!
https://code.google.com/archive/p/doctype-mirror/wikis/MetaCharsetAttribute.wiki

> In order for all browsers to recognize a `<meta charset>` declaration, it must be
> 
> - Within the `<head>` element,
> - **Before any elements that contain text, such as the `<title>` element,** AND
> - Within the first 512 bytes of your document, including DOCTYPE and whitespace

Atinux:
But actually we cannot filter the meta just for the charset, is it a real issue that title is after all the metas then?

pimlie:
The official W3C HTML5 Recommendation doesn't seem to mention a need for ordering at all: https://www.w3.org/TR/html5/document-metadata.html#specifying-the-documents-character-encoding

As the document @visualfanatic links to seems to be from 2011 (the site was taken down in 2011) and first release of html5 was in 2014, I think its safe to ignore the ordering

pi0:
@pimlie Yes. But I think even with the lack of spec reasoning about `encoding` ordering is valid. A text parser needs to know how many bytes is a character. Parsing first 512 bytes for encoding detection and then upgrading would be possible but not the opposite.

pi0:
@NightDiRaven Thanks for this PR, but as v1 is in out of maintenance state, i would suggest to continue this topic in #5433.

PR #5309, opened 2019-03-21
fix(ts): wrong export used in render.d.ts

RomainLanz:
Hey all 👋 

In the current version of `@nuxt/config`, [we try to get `NuxtConfigurationServerMiddleware` from `./index.d.ts`](https://github.com/nuxt/nuxt.js/blob/dev/packages/config/types/render.d.ts#L14) in `render.d.ts`. This export doesn't exist.

[It has been renamed to `ServerMiddleware`](https://github.com/nuxt/nuxt.js/blob/dev/packages/config/types/index.d.ts#L89).

**NOTE:** It's currently impossible to build a Nuxt application using TS because the compiler throw an error.

## Types of changes

### DIFF
--- a/packages/config/types/render.d.ts
+++ b/packages/config/types/render.d.ts
@@ -11,11 +11,11 @@ import { CompressionOptions } from 'compression'
 import { Options as EtagOptions } from 'etag'
 import { ServeStaticOptions } from 'serve-static'
 import { BundleRendererOptions } from 'vue-server-renderer'
-import { NuxtConfigurationServerMiddleware } from './index'
+import { ServerMiddleware } from './index'
 
 export interface NuxtConfigurationRender {
   bundleRenderer?: BundleRendererOptions
-  compressor?: CompressionOptions | NuxtConfigurationServerMiddleware
+  compressor?: CompressionOptions | ServerMiddleware
   csp?: any // TBD
   dist?: ServeStaticOptions
   etag?: EtagOptions | false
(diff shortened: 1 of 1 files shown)


packages/config/types/render.d.ts
```diff
@@ -11,11 +11,11 @@ import { CompressionOptions } from 'compression'
 import { Options as EtagOptions } from 'etag'
 import { ServeStaticOptions } from 'serve-static'
 import { BundleRendererOptions } from 'vue-server-renderer'
-import { NuxtConfigurationServerMiddleware } from './index'
+import { ServerMiddleware } from './index'
 
 export interface NuxtConfigurationRender {
   bundleRenderer?: BundleRendererOptions
-  compressor?: CompressionOptions | NuxtConfigurationServerMiddleware
+  compressor?: CompressionOptions | ServerMiddleware
```
kevinmarrec:
```suggestion
  compressor?: CompressionOptions | NuxtConfigurationServerMiddleware
```

packages/config/types/render.d.ts
```diff
@@ -11,11 +11,11 @@ import { CompressionOptions } from 'compression'
 import { Options as EtagOptions } from 'etag'
 import { ServeStaticOptions } from 'serve-static'
 import { BundleRendererOptions } from 'vue-server-renderer'
-import { NuxtConfigurationServerMiddleware } from './index'
+import { ServerMiddleware } from './index'
```
kevinmarrec:
```suggestion
import { NuxtConfigurationServerMiddleware } from './server-middleware'
```

kevinmarrec (changes requested):
Thanks @RomainLanz for pointing this, what about just importing it from `./server-middleware` ? (instead of going through `index`)

RomainLanz:
I merged your suggestions.

kevinmarrec (approved):

pi0:
Thanks @RomainLanz @kevinmarrec

Issue #5313, opened 2019-03-21
core-js problems with 2.5.0


andrewgorpenko:
### Version

[v2.5.0](https://github.com/nuxt.js/releases/tag/v2.5.0)

### Reproduction link

[http://not-needed.com](http://not-needed.com)

### Steps to reproduce

Do the fresh install of the Nuxt(via cli) with the following options:

- Project name project
- Project description Cool project
- Use a custom server framework none
- Choose features to install Axios
- Use a custom UI framework element-ui
- Use a custom test framework none
- Choose rendering mode Universal
- Author name Andrew Gorpenko
- Choose a package manager npm

After running the npm run dev I am getting the error:

```bash
 ERROR  Failed to compile with 3 errors

These dependencies were not found:
* core-js/library/fn/object/assign in ./node_modules/babel-runtime/core-js/object/assign.js
* core-js/library/fn/symbol in ./node_modules/babel-runtime/core-js/symbol.js
* core-js/library/fn/symbol/iterator in ./node_modules/babel-runtime/core-js/symbol/iterator.js
```

### What is expected ?

Clean install without errors

### What is actually happening?

Fails to compile because of missing dependencies

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8883">#c8883</a>)</em></sub></div>

tmorehouse:
We are having similar issues with upgrading from 2.4.5 to 2.5.0 when building the BootstrapVue docs:

```
SyntaxError: Unexpected token export
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:617:28)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at r (/root/bootstrap-vue/node_modules/vue-server-renderer/build.dev.js:9295:16)
    at Object.<anonymous> (webpack:/external "@babel/runtime/helpers/esm/asyncToGenerator":1:0)
    at o (webpack/bootstrap:36:0)
    at Module.<anonymous> (.nuxt/middleware.js:1:19)
    at o (webpack/bootstrap:36:0)
[...]

PR #5317, opened 2019-03-21
hotfix: revert to corejs@2

pi0:
Revert core-js to v2 as it is the default for current babel and forcing to v3 causes issues like #5313. Upgrading is also a breaking change for projects that have a babel.config file. So it makes sense to do it for nuxt3 or when ecosystem adopted more to v3.

### DIFF
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -1,19 +1,19 @@
 const defaultPolyfills = [
   // Promise polyfill alone doesn't work in IE,
   // Needs this as well. see: #1642
-  'es.array.iterator',
+  'es6.array.iterator',
   // This is required for webpack code splitting, vuex etc.
-  'es.promise',
+  'es6.promise',
   // this is needed for object rest spread support in templates
   // as vue-template-es2015-compiler 1.8+ compiles it to Object.assign() calls.
-  'es.object.assign',
-  // #2012 es.promise replaces native Promise in FF and causes missing finally
-  'es.promise.finally'
+  'es6.object.assign',
+  // #2012 es7.promise replaces native Promise in FF and causes missing finally
+  'es7.promise.finally'
 ]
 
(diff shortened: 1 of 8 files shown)


packages/babel-preset-app/src/index.js
```diff
@@ -1,19 +1,19 @@
...
-  'es.promise.finally'
+  'es6.object.assign',
+  // #2012 es7.promise replaces native Promise in FF and causes missing finally
+  'es7.promise.finally'
 ]
 
 function getPolyfills(targets, includes, { ignoreBrowserslistConfig, configPath }) {
   const { isPluginRequired } = require('@babel/preset-env')
-  const builtInsList = require('core-js-compat/data')
+  const builtInsList = require('@babel/preset-env/data/built-ins.json.js')
```
clarkdo: Shoule be json?

clarkdo:
LGTM

clarkdo:
One thing need to be noticed, if any lib has already updated to core-js 3, this is still an issue.

simllll:
just ran into this issue, we adopted already everything to core js 3.. actually I was just waiting for 2.5 to come because of core js 3. Any other idea how we could solve this?

pi0:
@simllll sorry to hear that :( Actually we've fased lots of problems because of libs that _implicitly_ depend on corejs 2 and we can't support both at a same time. How did you get prepared for 3.x?

simllll:
Yeah we had similar troubles, but switched or patched some libraries to get over it. Wasn't that much effort in the end for us, but still not so cool if we would revert this now.

I just read clarkdos comment that he is working on a option/flag to set the core is version, this would solve our problem! So I just fixed our version to 2.5.0 for now.

pi0:
@simllll You requested and it will be possible with next version (2.6.0) via #5411. All thanks to @clarkdo :heart:

simllll:
Very nice :) thanks guys for this awesome project!

PR #5326, opened 2019-03-21
fix(types): add type definition for typescript.typeCheck

danielroe:
## Types of changes

## Description

Added type definition for NuxtConfiguration for new `build.typescript` option.

TypeScript validation was failing with new configuration model.

## Checklist:


danielroe:
I can also provide more precise types from [fork-ts-checker](https://github.com/Realytics/fork-ts-checker-webpack-plugin) docs if that would be helpful, but we'd then need to keep them in sync. I don't think they currently export the Config type, though it is present in the built npm module.

packages/config/types/build.d.ts
```diff
@@ -58,6 +58,9 @@ export interface NuxtConfigurationBuild {
   templates?: any
   terser?: TerserPluginOptions | boolean
   transpile?: (string | RegExp)[]
+  typescript?: {
+    typeCheck?: { [key: string]: string } | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
```
kevinmarrec:
```suggestion
    typeCheck?: { [key: string]: any} | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
```

kevinmarrec (changes requested):

kevinmarrec:
Thanks for  your PR @danielroe !

I suggested a little change. And for the type of the options of ForkTsChecker, there is IMO too many properties to stay in sync, I prefer see if we can import it from somewhere, or see with the maintainer how we could have them.

packages/config/types/build.d.ts
```diff
@@ -58,6 +58,9 @@ export interface NuxtConfigurationBuild {
...
+  }
   useForkTsChecker?: { [key: string]: string } | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
```
kevinmarrec:
```suggestion
```

`useForkTsChecker` has been removed so it needs to be removed from types.

kevinmarrec (changes requested):

packages/config/types/build.d.ts
```diff
@@ -58,6 +58,8 @@ export interface NuxtConfigurationBuild {
   templates?: any
   terser?: TerserPluginOptions | boolean
   transpile?: (string | RegExp)[]
-  useForkTsChecker?: { [key: string]: string } | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
+  typescript?: {
+    typeCheck?: { [key: string]: any} | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
```
kevinmarrec:
```suggestion
    typeCheck?: { [key: string]: any } | boolean // TBD - Couldn't find typedefs for the forkTsCheckerWebpackPlugin options
```

kevinmarrec (changes requested):

packages/config/types/build.d.ts
kevinmarrec: Oops @danielroe , could you accept this last suggestion so I can aproove the PR ? Thanks.

kevinmarrec (approved):

pi0:
Thanks @danielroe @kevinmarrec :heart:

PR #5327, opened 2019-03-22
update css nuxt config type

bichikim:
## Types of changes

## Description

Css in nuxtConfig can be {src: string, lang: string}

## Checklist:

- [x] I have updated the documentation accordingly. (PR: #) https://nuxtjs.org/api/configuration-css

### DIFF
--- a/packages/config/types/css.d.ts
+++ b/packages/config/types/css.d.ts
@@ -1,0 +1,11 @@
+/**
+ * NuxtConfigurationEnv
+ * Documentation: https://nuxtjs.org/api/configuration-css
+ */
+
+interface NuxtConfigurationCssObject {
+  src: string
+  lang: string | 'stylus' | 'sass' | 'scss' | 'less' | 'css'
+}
+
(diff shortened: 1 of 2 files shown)


kevinmarrec:
@bichikim Thanks for your PR, but i'm not sure about the relevance of such way to use `css`, I mean it's handled by Nuxt but `file.styl` instead of  `{ src: 'file.styl', lang: 'stylus'}` will work and is cleaner / easier to read. Is there any use case `src` + `lang` would be needed ?

@manniL WDYT ? I wasn't aware Nuxt was handling such behavior.

pi0:
`lang` property is no longer used. Previously nuxt was using `<style lang=>` tags but now it is using `import`:

https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/App.js#L4

So the type detection is handled by extensions and webpack.

manniL:
I don't think we need to handle the explicit language value anymore. Maybe that should be deprecated as Nuxt should recognize it by default.

cc @pi0

PR #5337, opened 2019-03-22
fix(vue-app): decode uri in `getlocation`

### LINKED ISSUE #4948: router.base value with Extended unicode (åäö) fails
### Version

[v2.4.1](https://github.com/nuxt.js/releases/tag/v2.4.1)

### Steps to reproduce

`yarn create nuxt-app test_nuxt && cd test_nuxt` (I selected express and not much else)

Drop ```router: {
    base: "/%C3%B6l/",
  }``` 
intro you nuxt config. start a dev server. go to http://localhost:3000/öl. 

### What is expected ?

I expect the site to load properly from the /öl url. 

you may see the page blink and disappear with a DOM tree mismatch hydration error in the console. mode: 'spa' doesn't work at all (it can't find the page). base addresses without weird characters work fine. 

my live nuxt site running at https://litteraturbanken.se/översättarlexikon/ exhibits another symptom: using the back button at any time throws up the 404 page. cannot reproduce this with a minimal example, however. 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8606">#c8606</a>)</em></sub></div>

jroxendal:
## Types of changes

- [ x] Bug fix (a non-breaking change which fixes an issue)

## Description
fixes bug with extended unicode not working in router.base urls. resolves #4948.

## Checklist:

A test was added to the previous PR for this issue but that test fails. the page renders, though, so it looks like some detail is off. see test/unit/unicode-base.test.js

test output:
<details>
FAIL  test/unit/unicode-base.test.js
  ● unicode-base › /ö/ (router base)

    expect(received).toContain(expected) // indexOf

    Expected substring: "<h1 class=\"page-enter page-enter-active\">Unicode base works!</h1>"
    Received string:    "
        <div id=\"__nuxt\"><div id=\"__layout\"><h1>Unicode base works!</h1></div></div><script>window.__NUXT__={layout:\"default\",data:[{}],error:null,serverRendered:true};</script><script src=\"/%C3%B6/_nuxt/7995e158150a2be3d182.js\" defer=\"\"></script><script src=\"/%C3%B6/_nuxt/c50c5c4ec9a00b965824.js\" defer=\"\"></script><script src=\"/%C3%B6/_nuxt/7f8c828bc677f8267d14.js\" defer=\"\"></script><script src=\"/%C3%B6/_nuxt/b27d25b79711b6dc7435.js\" defer=\"\"></script>

    "

      21 |     const html = window.document.body.innerHTML
      22 |     // important to have the actual page transition classes here -> page works, no hydration error
    > 23 |     expect(html).toContain('<h1 class="page-enter page-enter-active">Unicode base works!</h1>')
         |                  ^
      24 |   })
      25 |
      26 |   // Close server and ask nuxt to stop listening to file changes

      at Object.toContain (test/unit/unicode-base.test.js:23:18)
</details>

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -266,7 +266,7 @@ export function getLocation(base, mode) {
   if (base && path.indexOf(base) === 0) {
     path = path.slice(base.length)
   }
-  return decodeURI(path || '/') + window.location.search + window.location.hash
+  return (path || '/') + window.location.search + window.location.hash
 }
 
 export function urlJoin() {
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -259,7 +259,7 @@ export function promisify(fn, context) {
 
 // Imported from vue-router
 export function getLocation(base, mode) {
-  let path = window.location.pathname
+  let path = decodeURI(window.location.pathname)
   if (mode === 'hash') {
     return window.location.hash.replace(/^#\//, '')
   }
--- a/test/unit/unicode-base.test.js
+++ b/test/unit/unicode-base.test.js
@@ -19,8 +19,7 @@ describe('unicode-base', () => {
     const window = await nuxt.server.renderAndGetWindow(url('/ö/'))
 
     const html = window.document.body.innerHTML
-    // important to have the actual page transition classes here -> page works, no hydration error
-    expect(html).toContain('<h1 class="page-enter page-enter-active">Unicode base works!</h1>')
+    expect(html).toContain('<h1>Unicode base works!</h1>')
   })
 
   // Close server and ask nuxt to stop listening to file changes


packages/vue-app/template/utils.js
```diff
@@ -259,7 +259,7 @@ export function promisify(fn, context) {
 
 // Imported from vue-router
 export function getLocation(base, mode) {
-  let path = window.location.pathname
+  let path = decodeURI(window.location.pathname)
```
aldarund: How about catching and processing error from decodeUri?

jroxendal: normally I'd be all for it but in this case we're decoding directly from window.location.pathname. since that's been encoded by the browser we're know it's been done without error and so decoding won't throw an error either. I'd be more worried about the decoding of user-generated strings from the config such as this one:
https://github.com/nuxt/nuxt.js/blob/5331a91223c7febd2ed612e2fe7264dd9c3c5928/packages/vue-app/template/router.js#L157

pi0:
@jroxendal Can you please update the corresponding test as well?

> test/unit/unicode-base.test.js

manniL:
As this code is imported from Vue-router we should think about submitting a PR there too

jroxendal:
I did try this with vue-cli a while back and found that 
```
export default new Router({
  mode: 'history',
  base: "/öl/"
```
worked without a hitch. so they're doing something differently that doesn't require decoding in the getLocation function.

pi0 (approved):

manniL:
Alright!

One last question: does the PR work? without the changes from https://github.com/nuxt/nuxt.js/issues/5297

jroxendal:
Nope, i tried without, with a few variants. No dice 🤷‍♂️

manniL:
Good. Happy it works now ☺️

Issue #5324, opened 2019-03-21
Lock dir not cleaned on compromised lock & fatal error lock is already released


pimlie:
### Version

[v2.5.1](https://github.com/nuxt.js/releases/tag/v2.5.1)

### Reproduction link

[https://github.com/nuxt/nuxt.js](https://github.com/nuxt/nuxt.js)

### Steps to reproduce

- Start a build that takes some time to finish
- run: `touch -t 200012311800 ./node_modules/.cache/nuxt/build-lock-*.lock`
- wait for the `Unable to update lock` warning to be displayed
  - this appears to only ever show between webpack builds (eg when a Client/Modern/Server finishes)

![image](https://user-images.githubusercontent.com/1067403/54789491-ec6aaa80-4c32-11e9-8345-b9756788b6bc.png)
- wait for the build to finish

![image](https://user-images.githubusercontent.com/1067403/54789501-f42a4f00-4c32-11e9-87f9-422c8ac84ac7.png)

### What is expected ?

Only the warning should be displayed. Not sure why the trace is showing, probably consola feature?

### What is actually happening?

- A Nuxt fatal error is printed about 'Error: Lock is already released'
- the `./node_modules/.cache/nuxt` folder still contains the `.lock` folder (the other folder is removed)

### Additional comments?

Will have a look at this, creating this issue for reference

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8888">#c8888</a>)</em></sub></div>

pi0:
> run: touch -t 200012311800 ./node_modules/.cache/nuxt/build-lock-*.lock

How this is happening in the real world?

> Only the warning should be displayed. Not sure why the trace is showing, probably consola feature?

CLI reports all thrown exceptions to `consola.error` and it shows the stack trace. I think this is expected behavior to correctly show a trace of an error. We can throw non-standard `new String("Error")` to prevent having stack but I'm not sure if this is a good practice.

highalps:
it occurred to me when `nuxt build` in version 2.5.1

pimlie:
@highalps At what point in your build does it happen and what is the cpu but especially the io usage at that time? (a playback of nuxt build with top & iotop would be preferred) I am having difficulties to reproduce the compromised lock behaviour consistently in the real world.

[...]

PR #5347, opened 2019-03-22
fix(cli): catch all possible errors thrown from proper-filelock

pimlie:
image

## Types of changes

## Description

Ref: #5324

Although I could have expected it maybe, its not really documented that proper-lockfile also throws errors when it cant release a lock. This pr should fix that behaviour by wrapping any code which originates from proper-lockfile in a `try/catch` and warn when an error was thrown.

The _only_ time a lock should result in a `Nuxt Fatal Error` is when a lock with the same id already exists.

## Checklist:

### DIFF
--- a/packages/utils/src/locking.js
+++ b/packages/utils/src/locking.js
@@ -41,7 +41,7 @@ export async function lock({ id, dir, root, options }) {
       consola.fatal(`A lock with id '${id}' already exists on ${dir}`)
     }
   } catch (e) {
-    consola.warn(`Check for existing lock failed`, e)
+    consola.debug(`Check for an existing lock with id '${id}' on ${dir} failed`, e)
   }
 
   let lockWasCompromised = false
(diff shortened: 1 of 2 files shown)


packages/utils/src/locking.js
```diff
@@ -59,8 +76,27 @@ export async function lock({ id, dir, root, options }) {
...
+        // as well, but in our case its much more likely the lock was
+        // compromised due to mtime update timeouts
+        const lockDir = `${lockPath}.lock`
+        if (await fs.exists(lockDir)) {
+          await fs.remove(lockDir)
+        }
+        return
+      }
+
+      consola.warn(e)
```
pi0: Can we early return by checking opposite of if condition?

packages/utils/src/locking.js
```diff
@@ -35,16 +35,33 @@ export async function getLockPath(config) {
...
-  const locked = await properlock.check(lockPath)
-  if (locked) {
-    consola.fatal(`A lock with id '${id}' already exists on ${dir}`)
+  try {
+    const locked = await properlock.check(lockPath)
+    if (locked) {
+      consola.fatal(`A lock with id '${id}' already exists on ${dir}`)
+    }
+  } catch (e) {
+    consola.warn(`Check for existing lock failed`, e)
```
pi0: Is this warn useful for normal users? What if we use `debug` level?

packages/utils/src/locking.js
pi0: Also what about using debug level?

pi0 (commented):
Nice attempt @pimlie :+1:

packages/utils/src/locking.js
pimlie: Hypothetically speaking you could argue that when you were unable to check if a lock exists, you should assume the lock exists. But as in Nuxt' case the lock is more just a helper then a full guard, you could just warn or ignore the fact. I prefer the warning, but probably more from the hypothetical standpoint then a practical one. But I am happy to change it if you prefer

packages/utils/src/locking.js
pimlie: Thats probably a good idea indeed here

packages/utils/src/locking.js
pi0: If it is a safe thing to pass probably hiding for normal users would be better DX. Anyway choose the log level that you prefer :)

pi0 (approved):

aldarund (approved):

PR #5354, opened 2019-03-23
feat(vue-renderer): add csp meta tags

### LINKED ISSUE #2874: Allow to enable CSP with a meta element
Hello! 

I'm just about to implement a proper CSP setup for my site and came across https://github.com/nuxt/nuxt.js/pull/2549. This is exactly what I need except for the fact that I'm using `generate` and can not set headers.

Would you folks be open for a config option to decide between headers or a similar meta element in the head?

Thanks a bunch! :)

jamesgeorge007:
Enable CSP with a meta element

## Types of changes

## Description
Closes #2874 
> Allow to enable CSP with a meta element


pi0:
CSP header is already added by nuxt middleware: https://github.com/nuxt/nuxt.js/blob/ef41e205e6787b784ab5907d99652b9d74891c6c/packages/server/src/middleware/nuxt.js#L76

packages/vue-renderer/src/renderer.js
```diff
@@ -413,6 +413,7 @@ export default class VueRenderer {
       const hash = crypto.createHash(hashAlgorithm)
       hash.update(serializedSession)
       cspScriptSrcHashes.push(`'${hashAlgorithm}-${hash.digest('base64')}'`)
+      HEAD += cspScriptSrcHashes.join()
```
pi0: I think this would be better to added after `vue-renderer:ssr:csp` hook. Also shouldn't we wrap it around any tag?

jamesgeorge007:
@pi0 Wrapped the csp header information within `meta` tag as suggested. Hope this is fine :+1:

packages/vue-renderer/src/renderer.js
```diff
@@ -418,6 +418,9 @@ export default class VueRenderer {
     // Call ssr:csp hook
     await this.context.nuxt.callHook('vue-renderer:ssr:csp', cspScriptSrcHashes)
 
+    // Add csp meta tags
+    HEAD += `<meta http-equiv="Content-Security-Policy" content="script-src ${cspScriptSrcHashes.join()}">`
```
manniL: I'd guard that and checka whether CSP is enabled or not.

Also, I might would hide that behind a feature flag 🤔 	
And maybe only enable it "by default" (if CSP is enabled/set) on nuxt generate
(cc @pi0)

pi0: Nice point @manniL. I agree

jamesgeorge007: @manniL @pi0 I didn't get you :confused: 
Can you just have a mention no how should it be changed :thinking:

jamesgeorge007: @manniL @pi0 I didn't get you :confused: 
Can you just have a mention on how should it be changed :thinking:

manniL: @jamesgeorge007 	No worries ☺️ 	

1. Check whether the user actually set sth. in the config's [render.csp](https://nuxtjs.org/api/configuration-render#csp)
2. We should use another config value to only enable that feature if the user *explicitly* enables it through the config. It could be also inside the render.csp object

jamesgeorge007: I guess wrapping  it within ```js if (this.context.options.csp) ``` checks whether `csp` is enabled and acts appropriately. Does this makes sense to you?
 I'm not sure on what is to be done further :thinking:

jamesgeorge007: I guess wrapping  it within ``if (this.context.options.csp)` checks whether `csp` is enabled and acts appropriately. Does this makes sense to you?
 I'm not sure on what is to be done further :thinking:

jamesgeorge007: @manniL I guess wrapping  it within ``if (this.context.options.csp)` checks whether `csp` is enabled and acts appropriately. Does this make sense to you?
 I'm not sure on what is to be done further :thinking:

jamesgeorge007: I guess wrapping  it within `if (this.context.options.csp)` checks whether `csp` is enabled and acts appropriately. Does this make sense to you?
 I'm not sure on what is to be done further :thinking:

packages/vue-renderer/src/renderer.js
```diff
@@ -418,6 +418,11 @@ export default class VueRenderer {
     // Call ssr:csp hook
     await this.context.nuxt.callHook('vue-renderer:ssr:csp', cspScriptSrcHashes)
 
+    // Add csp meta tags
+    if (this.context.options.render.csp) {
```
manniL: # Only render if CSP is used

You did it right 👍 

# Feature flag
As I mentioned, a feature flag would be nice. A proper name could be `asMeta` as the CSP string is rendered **as meta** element.

Sth. like:

```suggestion
    if (this.context.options.render.csp && this.context.options.render.csp.asMeta) {
```

# Bonus

Try to enable that one by default in case we use `nuxt generate`. A starting point would be this part:
https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L190-L192

# Bonus Bonus

We need tests for that as well ☺️

jamesgeorge007: @manniL Currently the `options.render` object lacks the `asMeta` property right?
> Try to enable that one by default in case we use nuxt generate

I'm not pretty sure on what you intend here :confused: 

Pardon me for all the pain here. I'm not that familiar with the code base and that requires me to ask such questions :smiley:

manniL: @jamesgeorge007 No worries! Happy that you ask and keep on going ☺️ 	I know that a new large code base can be quite intimidating sometimes 🙈 

> Currently the options.render object lacks the asMeta property right?

That's right. I should be part of the `render.csp` object (if `render.csp` isn't set to false). That's why I made the check above. First check if `render.csp` is truthy (`{} == true`), then check if the object has an `asMeta` entry. In fact, there is not more needed for that change then the change I applied in the suggestion ☺️ 	

> I'm not pretty sure on what you intend here 😕

When using `nuxt generate` you can't use the "header" CSP approach that Nuxt uses by default as no server is running. That's why I suggested to enable that meta element with the CSP string by default when generating the page (and when CSP is enabled ofc).
Hope that makes sense ☺️

jamesgeorge007: @manniL As per what I infer what you intend is to place these `if` conditionals as given below just after these lines:-

https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L190-L192

```js
if (render.csp) {
  if (render.csp.hasOwnProperty('asMeta')) {
   // Not exactly sure what should go here 
  }
}
```
> Still I'm not able to connect the dots

jamesgeorge007: Am I required to make these workarounds within the `config/src/options.js` file :thinking:

manniL: @jamesgeorge007 The current state is good!

- [x] Check whether CSP is enabled
- [x] Feature flag (`render.csp.asMeta`)
- [ ] Enable by default (when CSP truthy) on `nuxt generate`
- [ ] Tests
- [ ] Docs

manniL: For the last "logic" part (enable on generate by default), you have to use `options.js`, yes ☺️

packages/config/src/options.js
```diff
@@ -191,6 +191,13 @@ export function getNuxtConfig(_options) {
     options.render.csp = defaults(isObject(csp) ? csp : {}, cspDefaults)
   }
 
+  // Check if options.render is truthy
+  if (options.render.csp) {
+    if (options.render.csp.hasOwnProperty('asMeta')) {
+      // ToDo
```
jamesgeorge007: Is this the right implementation :thinking:

manniL: I would go back to

` if (this.context.options.render.csp && this.context.options.render.csp.asMeta) {`

jamesgeorge007: > For the last "logic" part (enable on generate by default), you have to use `options.js`, yes

- What you suggested above had already been brought about within `renderer.js`
- Now, in order to enable it on generate by default I had made an initial workaround as given above :arrow_up: 
- It goes within `options.js` (it checks whether `options.render.csp` is truthy and if is the case proceed with the check for existence of the `asMeta` property within the `render.csp` object.)
- What should go within the nested if block :thinking: 

Hope you got the sense of where am I stuck at :smiley:

jamesgeorge007: @manniL 

> For the last "logic" part (enable on generate by default), you have to use `options.js`, yes

- What you suggested above had already been brought about within `renderer.js`
- Now, in order to enable it by default on `nuxt generate` I had made an initial workaround as given above :arrow_up: 
- It goes within `options.js` (checks whether `options.render.csp` is truthy and if that's the case proceed with the check for existence of the `asMeta` property within the `render.csp` object.)
- What should go within the nested if block :thinking: 

Hope you got the sense of where I'm stuck at :smiley:

manniL: Whoops, I mixed up the files, sorry! 🙈

manniL: What we want for `options` is:

1. Check if CSP is truthy
2. Check if in "generation mode"
3. enable `asMeta` then.

jamesgeorge007: How are we gonna check whether if it is in the `generate` mode :thinking: 
Currently I can see an `options.modes` array, but not sure if it is the required one to be updated as per context.

> 3. Check if asMeta is strictly set to false, if so, return (explicitly disabled)
> 4. enable asMeta otherwise.
Is this the required workaround :-
```js
if (options.render.csp.asMeta === false) {
  return
} else {
     // Enable asMeta 
}
```
I'm not sure on what do you mean by enabling `asMeta`
Initially the `csp` object lacks that property right?

pi0: The only possibility to detect generate mode is `process.static` which is only filled during generate. We may use this value to force enable `asMeta` inside rendered. /cc @manniL @jamesgeorge007

pi0: OR: Set `options._generate` from CLI/commands/generate

manniL: > I'm not sure on what do you mean by enabling asMeta
 Initially the csp object lacks that property right?

If `render.csp` is truthy, is *should* be an object. Everything else doesn't make sense. If so, we can set the `asMeta` key easily.

Regarding your first question, @pi0 	already presented two options which you can consider.

> `if (options.render.csp.asMeta === false) {
  return`

That wasn't meant literally, sorry :see_no_evil: 	I mean that we should skip the force activation if it's set to false explicitly. Returning early would lead to way more skipping as it's a huge fn (which we may should split up :thinking: )

pi0:
@manniL @jamesgeorge007 I've pushed some refactors. Please review :)

TheAlexLichter (approved):

manniL:
@jamesgeorge007 	Thanks a lot for the contribution :raised_hands:

jamesgeorge007:
@manniL Thank you so much for the wonderful guidance :clap:

Issue #5338, opened 2019-03-22
Using @nuxt/typescript overrides any existing tsconfig.json file


sambowler:
### Version

[v2.5.1](https://github.com/nuxt.js/releases/tag/v2.5.1)

### Reproduction link

[https://codesandbox.io/s/0qmykr7wq0](https://codesandbox.io/s/0qmykr7wq0)

### Steps to reproduce

Using the official Nuxt with TypeScript example on CodeSandbox:
- Adjust the `tsconfig.json` in any way
- Restart the sandbox

TS support will be detected again and any modifications to `tsconfig.json` will be overridden.

### What is expected ?

The existing `tsconfig.json` is unmodified.

### What is actually happening?

The `tsconfig.json` content is overridden.

### Additional comments?

I've also noticed that the generation of the `tsconfig.json` doesn't seem to respect the `srcDir` setting in the Nuxt config.

For some background: we have a setup where we use TS for our server-side code and our Nuxt code in the same repo. At present, we have a `tsconfig.json` in the root for our server code and then a separate `tsconfig.json` for our Nuxt setup in a subdirectory (which is ignored by the first config file).

Since 2.5.0 our root `tsconfig.json` is being overridden by the automatically generated one. I don't think this would cause any issues if it respected the `srcDir` config option (i.e. generated it in the subdirectory) although it seems feasible to me that people may want to modify their config and overriding it every time may not be the best option.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8900">#c8900</a>)</em></sub></div>

kevinmarrec:
Hi @sambowler, we're gonna switch back to the old behavior which was extending a `tsconfig.json`.

Overriding tsconfig.json was not a good Idea (FYI it doesn't exactly overrided but add missing values in it).

The main issue we have is that we need to take root `tsconfig.json` for runTime support (nuxt.config.ts / serverMiddlewares), and `tsconfig.json` from `srcDir` for webpack related TS files (vue components, plugins).

Thanks for your feedback, it will be closely taken into account. I will refer this issue when I'll open the PR which brings better TS integration.

husayt:
@kevinmarrec 
Since latest it doesn't   like `"extends": "@nuxt/typescript-edge",` line in tsconfig.json. What is the right value for that?

kevinmarrec:
[...]

PR #5356, opened 2019-03-23
fix(packages/typescript): Fix not to overwrite tsconfig

sawa-zen:
## Types of changes

## Description

The `packages/typescript` setup process is breaking the existing tsconfig.
For example, there is the following destruction.
```
{
  ...
  "compilerOptions": {
    "lib": [
      "dom",
      "es2015"
    ]
  },
  ...
}
↓
{
  "compilerOptions": {
    "lib": [
      "dom",
      "es2015",
      "dom"
    ]
  },
}
```
This is a problem due to using `loash.defaultsDeep`.
https://github.com/lodash/lodash/issues/2801

But, I thought that overriding the default tsconfig settings to the existing tsconfig was a problem. So I changed to create a default tsconfig only when tsconfig did not exist.

Fixes: #5338

## Checklist:


kevinmarrec:
@sawa-zen Thanks for your time, we already planned to change this behavior, we will switch back to the `extends` way. So there won't be defaultDeeps anymore.

danielroe:
This should also fix deploying Nuxt 2.5.x with TypeScript on Now, which currently produces the following fatal error:

```
FATAL  EACCES: permission denied, open '/home/nowuser/src/tsconfig.json'
```

pi0:
Merged in #5367

PR #5363, opened 2019-03-24
chore(readme): add nuxt.ready() for middleware

HapLifeMan:
Starting on 2.5.0, it is now required to call `nuxt.ready()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ let config = require('./nuxt.config.js')
 config.dev = (process.env.NODE_ENV !== 'production')
 
 let nuxt = new Nuxt(config)
+await nuxt.ready()
 
 // Start build process (only in development)
 if (config.dev) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,6 +104,7 @@ let config = require('./nuxt.config.js')
 config.dev = (process.env.NODE_ENV !== 'production')
 
 let nuxt = new Nuxt(config)
+await nuxt.ready()
```
pi0: As this is a top-level code and await is not possible, can you please remove `await` part?

HapLifeMan: Yes I'll do it! Maybe it might be interesting to say in comment that it should be asynchronous if it's not top-level?

pi0: Yeah. We can also catch errors to prevent Unhandled Promises. For simplicity of readme, i suggest something like `nuxt.ready().catch(console.error)`.

pi0: Or... Just commenting that ready() returns a promise :)

HapLifeMan: `.catch(console.error)` is self-explanatory, nice 👍🏻

pi0 (approved):
Thanks!

PR #5367, opened 2019-03-24
refactor(ts): only generate tsconfig.json if missing

### LINKED ISSUE #5338: Using @nuxt/typescript overrides any existing tsconfig.json file

kevinmarrec:
## Types of changes

## Description
Fixes #5338 
Closes #5356 

Makes Nuxt not overriding `tsconfig.json` anymore, and only generates it with defaults if its missing.

Furthermore, TypeScript support is now only initialized if `@nuxt/typescript` is installed, so no more issues for projects which have a `tsconfig.json` in their root folder but doesn't want Nuxt to enable its TypeScript support.

The initializing state of TypeScript Support (includes `tsconfig.json` autogeneration step) is also more clear now :

Screenshot from 2019-03-26 10-42-18 

## Migration Guide

If you're using `2.4.x` :
1. Remove `nuxt-ts` dependency
2. Install `nuxt` and `@nuxt/typescript` dependencies
3. Remove your `tsconfig.json`, let Nuxt generate it with defaults at first run, then you can edit it to fit your needs.

If you're using `2.5.x` and have a working project, no changes are needed.

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1269)


packages/cli/src/utils/typescript.js
```diff
@@ -1,34 +1,16 @@
...
-    return false
-  }
-
-  consola.info(`${chalk.bold.blue('tsconfig.json')} found, enabling TypeScript runtime support`)
-
   try {
     const { setup } = require('@nuxt/typescript')
-    await setup(tsConfigPath, options)
+
+    consola.info('Initializing TypeScript Support')
```
pi0: Can we show a message like "Creating /full/path/to/tsconfig.json" ?

```js
const tsConfigPath = path.resolve()
consola.info
await setup
```

pi0:
@kevinmarrec I think we would also need a proper migration guide alongside with this PR changes

packages/cli/src/utils/typescript.js
kevinmarrec: What about just replace `tsconfig.json` by full path here ? : https://github.com/nuxt/nuxt.js/pull/5367/commits/fe3ecd63588390fdea30ba231c03a8a3eed4f78a#diff-8404199e6ca4961076ae35d7afeb980aR48

kevinmarrec:
@pi0 yes sure, will only make the PR out of DRAFT when everything will be good

packages/cli/src/utils/typescript.js
pi0: Yeah. Cool.

kevinmarrec:
Btw, I don't know what's that test unit error output, looks weird, gonna try to figure out why it's happening and why it's failing

kevinmarrec:
Well I tried to find what's wrong with tests, but as my changes are not relevant to the error and that it works on `dev` branch, I'm really confused...

kevinmarrec:
@pi0 **IMPORTANT** : I'd like to have a "Co-authored-by" #5356 owner in the merge commit message when it gets merged in `dev`

pi0 (approved):

pi0:
@kevinmarrec 9a3fc8a44e3cf3a8f551fa93ba60eb88a8c39663

kevinmarrec:
Thanks @pi0

PR #5366, opened 2019-03-24
feat(webpack): export webpack's config builder classes

damianstasik:
## Types of changes
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
Currently there is no way to extract a webpack config from a Nuxt project, because there is no access to `WebpackClientConfig`, `WebpackServerConfig` and `WebpackModernConfig`.

Here is what I wanted to do:
```js
const { Builder } = require('@nuxt/builder');
const { Nuxt } = require('@nuxt/core');
const { WebpackClientConfig } = require('@nuxt/webpack'); // currently not exported

const nuxt = new Nuxt();
const builder = new Builder(nuxt);
console.log(new WebpackClientConfig(builder.getBundleBuilder()).config()); // webpack config
```

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #5378, opened 2019-03-25
refactor(webpack): add ability to get webpack config class

damianstasik:
## Types of changes

## Description
This PR is an improvement of #5366 and adds a new method to `WebpackBundler` that returns a webpack config class (`WebpackClientConfig`, `WebpackServerConfig` or `WebpackModernConfig`).

I also cleaned up the `build` method a bit and moved the plugin aliases into `WebpackClientConfig` and `WebpackServerConfig` so that each class has all of it's webpack aliases in one place.

## Checklist:


galvez:
@visualfanatic Please ensure the test suite passes first.

visualfanatic:
@galvez @pi0 fixed, could you take a look?

packages/webpack/src/builder.js
```diff
@@ -37,44 +37,32 @@ export class WebpackBundler {
     }
   }
 
+  getWebpackConfig(name) {
```
clarkdo: How about returning `config` object here, then we don't need an extract `bind`

clarkdo:
Changed title to `refactor` for being eligible for v2.5.2

packages/webpack/src/builder.js
pi0: @clarkdo Can you please elaborate more about bind?

This function is not async to leverage `Promise.all`. The runner will be specified based on build type [here](https://github.com/nuxt/nuxt.js/blob/82abc2632f8da8343d32128377b654838b963f51/packages/webpack/src/builder.js#L106)

pi0: @clarkdo Refactor applied. Please review again.

visualfanatic: @pi0 your change was a surprise, but it was appreciated, thank you.

juliendargelos:
As I proposed [here](https://github.com/nuxt/nuxt.js/pull/5425), I think it could be useful for some advanced modules to be able to extend the webpack base config class for independent webpack build based on Nuxt base config. It's just about exposing [WebpackBaseConfig](https://github.com/visualfanatic/nuxt.js/blob/5c1e540747067c6a938acbe812dd6624767e265b/packages/webpack/src/config/base.js) (and possibly `WebpackClientConfig`, `WebpackServerConfig` and `WebpackModernConfig`) in [index.js](https://github.com/visualfanatic/nuxt.js/blob/5c1e540747067c6a938acbe812dd6624767e265b/packages/webpack/src/index.js):

```javascript
export { WebpackBundler as BundleBuilder } from './builder'
export { default as WebpackBaseConfig } from './config/base'
```

pi0:
There is no guarantee about structure of webpack config classes to be the same across releases. So one can use final result and use/extend it for advanced use cases.

PR #5387, opened 2019-03-26
fix(vue-renderer): add the csp hash if `unsafe-inline` hasn't been specified

sambowler:
Only adds the CSP hash if `'unsafe-inline'` hasn't been specified.

## Types of changes

## Description
If both a hash and `'unsafe-inline'` are present, browsers ignore the rule (see screenshot). This change will allow users to specify the option effectively.

Screen Shot 2019-03-26 at 10 11 18

## Checklist:


packages/vue-renderer/src/renderer.js
```diff
@@ -408,7 +408,11 @@ export default class VueRenderer {
 
     // Calculate CSP hashes
     const cspScriptSrcHashes = []
-    if (this.context.options.render.csp) {
+    const csp = this.context.options.render.csp
+    const containsUnsafeInlineScriptSrc = csp && csp.policies['script-src'] && csp.policies['script-src'].indexOf('\'unsafe-inline\'') !== -1
```
manniL:
```suggestion
    const containsUnsafeInlineScriptSrc = csp && csp.policies['script-src'] && csp.policies['script-src'].includes('\'unsafe-inline\'')
```

packages/vue-renderer/src/renderer.js
```diff
@@ -408,7 +408,11 @@ export default class VueRenderer {
...
+
+    // Only add the hash if 'unsafe-inline' rule isn't present to avoid conflicts
```
manniL:
```suggestion
    // Only add the hash if 'unsafe-inline' rule isn't present to avoid conflicts (#5387)
```

TheAlexLichter (changes requested):
Hey! Good catch. I left you a few suggestions.

packages/vue-renderer/src/renderer.js
```diff
@@ -408,7 +408,11 @@ export default class VueRenderer {
 
     // Calculate CSP hashes
     const cspScriptSrcHashes = []
-    if (this.context.options.render.csp) {
+    const csp = this.context.options.render.csp
+    const containsUnsafeInlineScriptSrc = csp && csp.policies && csp.policies['script-src'] && csp.policies['script-src'].indexOf('\'unsafe-inline\'') !== -1
```
manniL:
```suggestion
    const containsUnsafeInlineScriptSrc = csp && csp.policies && csp.policies['script-src'] && csp.policies['script-src'].includes(`'unsafe-inline'`)
```

sambowler:
Thanks @manniL, I've applied those suggestions. 🙂

packages/vue-renderer/src/renderer.js
```diff
@@ -408,7 +408,11 @@ export default class VueRenderer {
 
     // Calculate CSP hashes
     const cspScriptSrcHashes = []
-    if (this.context.options.render.csp) {
+    const csp = this.context.options.render.csp
+    const containsUnsafeInlineScriptSrc = csp && csp.policies && csp.policies['script-src'] && csp.policies['script-src'].includes(`'unsafe-inline'`)
```
manniL: Maybe even 

```suggestion
    const containsUnsafeInlineScriptSrc = csp.policies && csp.policies['script-src'] && csp.policies['script-src'].includes(`'unsafe-inline'`)
```

as the csp truth check is already done before

sambowler: I think this is still necessary -- if `csp` were undefined but we tried to access `policies` we'd get a "Cannot read property..." error

manniL: Whoops, you are right!

pi0:
@manniL Can you please review this again?

TheAlexLichter (approved):

manniL:
@pi0 fine from my POV :relaxed:

williamchong007:
Defining BOTH `unsafe-inline` and hash in CSP is actually a valid use case, it is for backward compatibility of CSPv1. That's why browsers with CSPv2+ support ignore it when hash is present at the same time.

[MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src):

    See unsafe inline script for an example. Specifying nonce makes a modern browser ignore 'unsafe-inline' which could still be set for older browsers without nonce support.

It is also a recommended practice suggested by google
image

With this PR in, how can I specify BOTH `unsafe-inline` and SHA hash  for my CSP rule?

@manniL @sambowler

manniL:
@williamchong007 Interesting. Please raise an extra issue so we can discuss it further there!

PR #5411, opened 2019-03-29
feat(babel-preset-app): support specifying core-js version

clarkdo:
## Types of changes

## Description

With current change, user can user functional `presets` to set core-js version:

```sh
yarn add --dev core-js@3 @babel/runtime-corejs2 --save
```
```js
export default {
  build: {
    babel: {
      presets({ isServer }) {
        const targets = isServer ? { node: '10' } : { ie: '11' }
        return [
          [
            require.resolve('@nuxt/babel-preset-app'),
            {
              targets,
              corejs: { version: 3 }
            }
          ]
        ]
      }
    }
  }
}

```

But it may be not very simple, so we can discuss to see if you have any idea @nuxt/core-team . 

I have a rough idea which may be not very elegant as well, current build.babel may be kind of complex and unhelpful, but for not breaking anything, how do you think add `build.babel.nuxtPreset` or sth else:

```js
export default {
  build: {
    babel: {
      nuxtPreset: {
        corejs: { version: 2 },
        // or any option of @nuxt/babel-preset-app
        // or client/server specific
        serverTargets: {},
        clientTargets: {}
      }
    }
  }
}
```


packages/babel-preset-app/src/index.js
```diff
@@ -48,6 +68,12 @@ module.exports = (context, options = {}) => {
     absoluteRuntime
   } = options
 
+  let { corejs = { version: 2 } } = options
+
+  if (typeof corejs === 'number') {
```
pi0: I think `version` type can be also a `string` for minor versions no?

clarkdo: Unfortunately it only supports number https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/index.js#L57

pi0:
@clarkdo What if we extract nuxtPreset option to something like `build.preset` and leaving `build.babel` to directly being passed to babel-loader?

packages/babel-preset-app/src/index.js
pi0: `Number('2.3.1')` is `NaN` but i think still makes sense to allow passing values like `'2'`.

clarkdo:
> @clarkdo What if we extract nuxtPreset option to something like `build.preset` and leaving `build.babel` to directly being passed to babel-loader?

Yes, it's also an idea what I'm thinking.

packages/babel-preset-app/src/index.js
clarkdo: Right, actually babel only supports `2, 3, '2', '3', false`

pi0:
@clarkdo As default is 2, should not we add `@babel/runtime-corejs2` to `@nuxt/babel-preset-app` dependencies?

clarkdo:
> @clarkdo As default is 2, should not we add `@babel/runtime-corejs2` to `@nuxt/babel-preset-app` dependencies?

Currently default config is `@babel/runtime` since `useBuiltIns` is always `usage`

pi0:
But docs says:

```sh
yarn add --dev core-js@3 @babel/runtime-corejs3
 # or
 yarn add --dev core-js@2 @babel/runtime-corejs2
 ```

Should we also change to something like this?

```sh
yarn add --dev core-js@3 @core-js-compat
 # or
 yarn add --dev core-js@2 @babel/preset-env
 ```

clarkdo:
`core-js-compat` is directly dependent by `@babel/preset-env`, so I think we don't need to let user install it.

For `@babel/runtime-corejs3` and `@babel/runtime-corejs3`, it's for the users who are not using `useBuiltIns: 'usage'`

pi0 (dismissed):

PR #5412, opened 2019-03-29
feat(typescript): detect typescript based on tsconfig.json

pi0:
- Refactor `registerTSNode` from `@nuxt/typescript` into `@nuxt-cli`. This way users can keep `@nuxt-typescript` (**~55.2 MB**) into `devDependencies` and `ts-node` (**1.6 MB**) in `dependencies`. This means much less production install size and also have **runtime support**.
- Move external imports into `cli/imports.js` for better maintenance.
- Only enable typescript for a project if `tsconfig.json` exists in the `rootDir`.
- Only If `tsconfig.json` is empty JSON, defaults will be written into it by `@nuxt/typescript`.


manniL:
We have to highlight that change in the release notes (as done before with changes on experimental features) :relaxed:

pi0:
@manniL Sure. Actually it is a complement of #5367. New migration guide would be:

1. Remove `nuxt-ts` dependency
2. Install `nuxt` and `@nuxt/typescript` dependencies
3. Create an empty `tsconfig.json` file. Nuxt will automatically generate it with defaults at first run, then you can edit it to fit your needs.

kevinmarrec (dismissed):
Wow, nice work @pi0 !!!

So I guess we can close https://github.com/nuxt/docs/pull/1269 (which was linked to #5367 ) and start from current docs https://nuxtjs.org/guide/typescript to just add the `ts-node` thing ?

kevinmarrec:
Migration guide for 2.5.x projects should be just installing `ts-node` as `dependency` if they need runtime support.

kevinmarrec:
@pi0 `ts-node` shouldn't be removed from `package.json` then ?

pi0:
@kevinmarrec Migration guide updated. Does it looks good now? (Feel free editing my comment). 

> @pi0 ts-node shouldn't be removed from package.json then ?

For lazy people who only install `@nuxt/typescript` that would be harmless. But can be removed as well.

kevinmarrec:
@pi0 perfect

Yeah I see, but if it's not used anymore within this package, we probably should remove it ?
It would indeed mean we need to warn users in release notes they need to install `ts-node` back when migrating to 2.6

pi0:
@kevinmarrec Done. Dependency removed and guide updated. I had to add `ts-node` to nuxt mono-repo root package.json in order to CI passing

kevinmarrec:
Alright perfect @pi0, worth mentionning that the guide is applicable for `2.4.x` -> `2.6.x`.
For `2.5.x` -> `2.6.x` it's only the `ts-node` step.

Don't know how exactly we're gonna wrap it up in release notes but yeah 🙈

PR #5413, opened 2019-03-29
revert(core): call ready to prevent breaking changes

### LINKED ISSUE #5408: Serverless error when upgrading from v2.4 to v2.5
### Version

### Reproduction link

[http://donthaveone.com](http://donthaveone.com)

### Steps to reproduce

Deploy Nuxt application using v2.4 to lambda function with express.

Based on https://github.com/serverless/examples/tree/master/aws-node-vue-nuxt-ssr

### What is expected ?

Application should function normally.

After upgrading to v2.5.1 express always returns a "cannot GET /" message.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8955">#c8955</a>)</em></sub></div>

pi0:
fixes #5408.

This fixes the condition when nuxt is starting programmatically in production mode with a custom server like express that does not calls `ready()`. We shouldn't change it until 3.0 because it is a breaking change for this condition. PS: No action is required from the side of users. Calling and awaiting on `ready()` after the constructor is _still_ recommended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,11 +100,10 @@ Or you can start by using one of our starter templates:
 const { Nuxt, Builder } = require('nuxt')
 
 // Import and set nuxt.js options
-let config = require('./nuxt.config.js')
+const config = require('./nuxt.config.js')
 config.dev = (process.env.NODE_ENV !== 'production')
 
-let nuxt = new Nuxt(config)
-nuxt.ready().catch(console.error)
+const nuxt = new Nuxt(config)
 
 // Start build process (only in development)
 if (config.dev) {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -100,11 +100,10 @@ Or you can start by using one of our starter templates:
 const { Nuxt, Builder } = require('nuxt')
 
 // Import and set nuxt.js options
-let config = require('./nuxt.config.js')
+const config = require('./nuxt.config.js')
 config.dev = (process.env.NODE_ENV !== 'production')
 
-let nuxt = new Nuxt(config)
-nuxt.ready().catch(console.error)
```
manniL: Shouldn't we still advise to wait for nuxt ready (ideally)?

pi0: Yes but it also caused complains. Maybe for 3.x we can show a more complete example with await. Currently nuxt exactly does this and also shows error on console.

TheAlexLichter (approved):

PR #5417, opened 2019-03-29
fix(renderer): retry render if renderer is in loading or created state

pi0:
## Types of changes

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -457,7 +457,7 @@ export default class VueRenderer {
     if (!this.isReady) {
       // Production
       if (!this.context.options.dev) {
-        if (!_retried && (['loading', 'created'].includes(this._state)) {
+        if (!_retried && ['loading', 'created'].includes(this._state)) {
           await this.ready()
           return this.renderRoute(url, context, true)
         }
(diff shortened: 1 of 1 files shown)


packages/vue-renderer/src/renderer.js
```diff
@@ -452,11 +452,15 @@ export default class VueRenderer {
     }
   }
 
-  async renderRoute(url, context = {}) {
+  async renderRoute(url, context = {}, _retried) {
     /* istanbul ignore if */
     if (!this.isReady) {
       // Production
       if (!this.context.options.dev) {
+        if (!_retried && (this._state === 'loading' || this._state === 'created')) {
```
manniL: Maybe add a comment to change that behavior in v3?

manniL:
```suggestion
        if (!_retried && (['loading', 'created'].includes(this._state)) {
```

TheAlexLichter (dismissed):

packages/vue-renderer/src/renderer.js
pi0: Generally uaing array causes triggering iterator (for includes) call. But as this is an error handler it is fine 😊

TheAlexLichter (approved):

PR #5419, opened 2019-03-29
chore(postcss): deprecate config support

TheAlexLichter:
## Types of changes

- [x] Feature deprecation

## Description

Using a `postcss.config.js` or a similar file will not apply any of the default plugins coming with Nuxt which could lead to severe issues (like not resolving aliases).

## Checklist:

### DIFF
--- a/packages/webpack/src/utils/postcss.js
+++ b/packages/webpack/src/utils/postcss.js
@@ -34,7 +34,7 @@ const postcssConfigFileWarning = (() => {
     if (executed) {
       return
     }
-    consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3.`)
+    consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3 as they remove all defaults set by Nuxt and can cause severe problems with features like alias resolving inside your CSS.`)
     executed = true
   }
 })()
--- a/packages/webpack/src/utils/postcss.js
+++ b/packages/webpack/src/utils/postcss.js
@@ -28,6 +28,17 @@ export const orderPresets = {
   }
 }
 
+const postcssConfigFileWarning = (() => {
+  let executed
+  return () => {
+    if (executed) {
+      return
+    }
+    consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3.`)
+    executed = true
+  }
+})()
+
 export default class PostcssConfig {
   constructor(buildContext) {
     this.buildContext = buildContext
@@ -82,7 +93,7 @@ export default class PostcssConfig {
       ]) {
         const configFile = path.resolve(dir, file)
         if (fs.existsSync(configFile)) {
-          consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of a${file} file. Support for such files will be removed in Nuxt 3.`)
+          postcssConfigFileWarning()
           return configFile
         }
       }
--- a/packages/webpack/src/utils/postcss.js
+++ b/packages/webpack/src/utils/postcss.js
@@ -28,16 +28,13 @@ export const orderPresets = {
   }
 }
 
-const postcssConfigFileWarning = (() => {
-  let executed
-  return () => {
-    if (executed) {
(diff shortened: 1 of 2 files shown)


packages/webpack/src/utils/postcss.js
```diff
@@ -28,6 +28,17 @@ export const orderPresets = {
   }
 }
 
+const postcssConfigFileWarning = (() => {
+  let executed
+  return () => {
+    if (executed) {
+      return
+    }
+    consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3.`)
```
pi0: Probably also adding a small description of why? Like because it overrides all nuxt postcss config. OR linking do the docs

manniL:
```suggestion
    consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3 as they remove all defaults set by Nuxt and can cause severe problems with features like alias resolving inside your CSS.`)
```

Like that? :relaxed:

pi0: Yep :

test/fixtures/with-config/with-config.test.js
```diff
@@ -14,26 +14,18 @@ const hooks = [
...
+        'message': 'Found 2 plugins that match the configuration, suggest to specify extension:'
       }],
-      [
-        'Using styleResources without the nuxt-style-resources-module is not suggested and can lead to severe performance issues.',
-        'Please use https://github.com/nuxt-community/style-resources-module'
-      ],
-      [
-        'Notice: Please do not deploy bundles built with analyze mode, it\'s only for analyzing purpose.'
-      ]
+      ['Please use `build.postcss` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3.'],
```
manniL:
```suggestion
      ['Please use `build.postcss` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3 as they remove all defaults set by Nuxt and can cause severe problems with features like alias resolving inside your CSS.'],
```

packages/webpack/src/utils/postcss.js
```diff
@@ -28,6 +28,17 @@ export const orderPresets = {
   }
 }
 
+const postcssConfigFileWarning = (() => {
```
pi0: We can in the future probably assign a unique code to the warnings so it can be both used to report them once and also link to the docs. Like https://nuxtjs.org/KB/107

pi0 (dismissed):

packages/webpack/src/utils/postcss.js
pi0: @manniL What about storing state in function object? (And a normal function)

```js
function postcssConfigFileWarn() {
  if (postcssConfigFileWarn.executed) {
    return
  }
  consola.warn(`Please use \`build.postcss\` in your nuxt.config.js instead of an external config file. Support for such files will be removed in Nuxt 3 as they remove all defaults set by Nuxt and can cause severe problems with features like alias resolving inside your CSS.`)
  postcssConfigFileWarn.executed = true
}
```

manniL: > We can in the future probably assign a unique code to the warnings so it can be both used to report them once and also link to the docs. Like https://nuxtjs.org/KB/107

Like the idea for the future!

manniL: > @manniL What about storing state in function object? (both normal and arrow can do this)

Didn't know about that TBH :thinking: 	
Do you think that approach would make it more readable?

pi0: Actually, it took me a few seconds to understand what that IIFE does. I think yes, it can probably increase readability and scalability (for multiple warning handling). 

Perf is not a valid concern but it also reduces one function creation :)

manniL: Would you mind adding the change then as you suggested it? :relaxed:

pi0: Sure :)

pi0: @manniL Does it looks good to you now?

manniL: Yup, LGTM :relaxed:

PR #5422, opened 2019-03-30
chore(types): add interface for readyCallback

orblazer:
## Types of changes

## Description

Add interface for easily add types for `onNuxtReady`.

Now everybody can declare `onNuxtReady` like that :
```typescript
declare global {
  interface Window {
    onNuxtReady: ReadyCallback;
  }
}
```


kevinmarrec:
Hi @orblazer, thanks for your time but I don't think this feature is needed for the following reasons:
- The `window.onNuxtReady` is a hidden undocumented feature
- Even if this type would live here, people wouldn't even look for it, as it's undocumented
- This type can directly implemented in your project, and is not dependent on other Nuxt types (it only depends on Vue) like

```ts
declare global {
  interface Window {
    onNuxtReady(cb: (app: Vue) => void): void
  }
}
```

orblazer:
OK ok, thanks @kevinmarrec.

PR #5430, opened 2019-03-31
examples: add docker example

UtkarshGupta-CS:
## Types of changes

- [x] Example for running nuxt app with docker

## Description

This PR adds example for running nuxt app with docker. This could highly useful for nuxt community as there no proper resource available to build nuxt app with docker.

Screenshot 2019-04-01 at 1 37 00 AM

## Checklist:


examples/docker-build/dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM node:10-stretch-slim as builder
```
pi0: It would be a better idea to use a consistent runtime under the hood for all steps. Either Alpine or a Debian based one. There may be modules in `node_modules` with native bindings.

examples/docker-build/dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+WORKDIR /src
+
+# Remove unnecessary files from node_modules (.md, .ts, ...)
+RUN cd /tmp && \
+  curl -#L https://github.com/tj/node-prune/releases/download/v1.0.1/node-prune_1.0.1_linux_amd64.tar.gz | tar -xvzf- && \
```
pi0: node-prune, has known (tj/node-prune#50) issues that may introduce problems for users. `.yarnclean` is more adviced.

examples/docker-build/dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  mv -v node-prune /usr/local/bin && rm -rvf * && \
+  echo "node-prune" > /usr/local/bin/node-clean && chmod +x /usr/local/bin/node-clean
+
+COPY package*.json ./
+RUN npm install --no-progress && node-clean
```
pi0: Internally all of the nuxt projects are using yarn and it is included by default in official docker images. I suggest to use yarn as example + appreciate flags to prevent lock file modification.

examples/docker-build/dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Build app
+COPY . .
+RUN npm run build
+
```
pi0: After building, it would better idea to strip all `devDependencies` and only keep `dependencies` (prod) for less image size and reducing the attack surface.

examples/docker-build/dockerfile
```diff
@@ -0,0 +1,31 @@
...
+FROM node:10-alpine
+
+ENV APP_ROOT /src
+
+RUN mkdir -p ${APP_ROOT}
```
pi0: I think docker should create this directory in `COPY` step.

examples/docker-build/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "private": true,
+  "scripts": {
+    "dev": "nuxt",
+    "build": "nuxt build",
+    "start": "nuxt start",
+    "generate": "nuxt generate"
+  },
+  "dependencies": {
+    "cross-env": "^5.2.0",
+    "nuxt": "^2.4.0"
```
pi0: It would be a better idea to advice using `nuxt-start` in `dependencies` and `nuxt` in `devDependencies`

examples/docker-build/.gitignore
```diff
@@ -0,0 +1,84 @@
+# Created by .ignore support plugin (hsz.mobi)
```
pi0: To keep example small, I suggest removing extra boilerplate like `.gitignore` from the example.

examples/docker-build/components/Logo.vue
```diff
@@ -0,0 +1,79 @@
+<template>
```
pi0: To keep example small, I suggest removing extra boilerplate like this from the example.

pi0 (changes requested):

examples/docker-build/dockerfile
pi0:
```
yarn install \
 --prefer-offline \
 --frozen-lockfile \
 --non-interactive \
 --production=false
```

examples/docker-build/dockerfile
pi0: `.yarnclean`:

```
*.yaml
*.md
*.lock
LICENSE
CHANGELOG
```

examples/docker-build/dockerfile
pi0: Something like this:

```
RUN rm -rf node_modules && \
 NODE_ENV=production yarn install \
  --prefer-offline \
  --pure-lockfile \
  --non-interactive \
  --production=true
 ```

**Comment by UtkarshGupta-CS** [comment]
@pi0 Made the requested changes.

File: examples/docker-build/dockerfile [hunk 9]
```diff
@@ -0,0 +1,31 @@
+FROM node:latest as builder
```
pi0: I think `node:latest` is not alpine based. Can we use `node:alpine` here as well?

examples/docker-build/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "build": "nuxt build",
+    "start": "nuxt-start",
+    "generate": "nuxt generate"
+  },
+  "dependencies": {
+    "cross-env": "latest",
+    "nuxt-start": "latest"
+  },
+  "devDependencies": {
+    "nodemon": "latest",
```
pi0: Is `nodemon` required?

pi0 (commented):
Thanks, @UtkarshGupta-CS. I've just left two minor comments.

pi0:
@UtkarshGupta-CS I didn't locally test. Can you please also confirm all changes are working fine with docker?

examples/docker-build/dockerfile
UtkarshGupta-CS: Since this dockerfile is multistage, `node:latest` won't affect any kind of build. Final build will always be small because of `node:latest-alpine`. Moreover I would suggest using `node:latest` image as a builder image like how we are using here because modules like `image-webpack-loader` have native dependencies, won't compile directly in `node:latest-alpine` image(requires other packages).

pi0: My reasoning is more because of that there may be native modules that are incompatible with alpine. We install production `node_modules` in a Debian based container and then run it in an alpine based one. Maybe that would be better to move second `yarn install` step to the second stage and also remove yarn cache but it makes our example more complex. So we can ignore it for now and later use better documentation to explain such in-depth details.

examples/docker-build/package.json
UtkarshGupta-CS: removed `nodemon` as it is not required.

UtkarshGupta-CS:
@pi0 Tested locally, works fine.
Screenshot 2019-04-01 at 1 37 00 AM

Screenshot 2019-04-01 at 2 39 30 AM

pi0 (approved):

tbanys:
How to run nuxt in development mode with docker?

pi0:
@tbanys This PR is related to production hosting nuxt so it could be better to discuss in Discord or an Issue.

Running a docker based development environment for nuxt, could be tricky and vary depending on OS. You can try by  creating a docker-compose file like this:

```yaml
version: '3'

services:
    nuxt:
        image: node:alpine
        command: yarn dev
        volumes:
            - .:/app
        environment:
            - HOST=0.0.0.0
        working_dir: /app
        ports:
            - 3000:3000
```

Initialize project: (on/off command)

```
docker-compose run --rm nuxt yarn create nuxt-app .
```

Run development server:

```
docker-compose up nuxt
```

You can now access it via http://localhost:3000

```
 docker-compose up nuxt
Creating foo_nuxt_1 ... done
Attaching to foo_nuxt_1
nuxt_1  | yarn run v1.13.0
nuxt_1  | $ nuxt
nuxt_1  | 12:44:02 ℹ Listening on: http://localhost:3000
nuxt_1  | 12:44:03 ℹ Preparing project for development
nuxt_1  | 12:44:03 ℹ Initial build may take a while
nuxt_1  | 12:44:03 ✔ Builder initialized
nuxt_1  | 12:44:03 ✔ Nuxt files generated
nuxt_1  | webpackbar 12:44:04 ℹ Compiling Client
nuxt_1  | webpackbar 12:44:04 ℹ Compiling Server
nuxt_1  | webpackbar 12:44:08 ✔ Server: Compiled successfully in 3.38s
nuxt_1  | webpackbar 12:44:08 ✔ Client: Compiled successfully in 3.66s
nuxt_1  | 12:44:08 ℹ Waiting for file changes
```

jjelev:
@UtkarshGupta-CS 
Can you explain why do you run 2 different `yarn install` commands and `yarn build` is not with production settings ?
Also why don't you use `node:alpine` for everything ?

[Dockerfile:](https://github.com/nuxt/nuxt.js/blob/dev/examples/docker-build/dockerfile)
```
FROM node:latest as builder

WORKDIR /src

COPY . .

RUN yarn install \
  --prefer-offline \
  --frozen-lockfile \
  --non-interactive \
  --production=false

RUN yarn build

RUN rm -rf node_modules && \
  NODE_ENV=production yarn install \
  --prefer-offline \
  --pure-lockfile \
  --non-interactive \
  --production=true

FROM node:alpine

WORKDIR /src

COPY --from=builder /src  .

ENV HOST 0.0.0.0
EXPOSE 3000

CMD [ "yarn", "start" ]
```

UtkarshGupta-CS:
@jjelev We used 1st `yarn install` for getting dependencies and devDependencies for `yarn build`. Then we removed all the dependencies because after the build we have all the static files we need to serve. Then we only installed dependencies (In production mode, does not install devDependencies) to serve using `yarn start`. The 2nd `yarn install` becomes quite faster as yarn get dependency from its cache.

`nuxt` dependency is actually quite big in size but it is required for building the app. But after the build, we can use `nuxt-start` (package made by nuxt team, which is immensely small compare to `nuxt`) to serve the static files.
> package.json file
```
{
  "name": "docker-build",
  "scripts": {
    "dev": "nuxt",
    "build": "nuxt build",
    "start": "nuxt-start",
    "generate": "nuxt generate"
  },
  "dependencies": {
    "cross-env": "latest",
    "nuxt-start": "latest"
  },
  "devDependencies": {
    "nuxt": "latest"
  }
}
```

Why we didn't use `node:alpine` for everything because alpine OS doesn't support some native dependency like `libpng` and many others out of the box. User need to install it manually so it's better to use `node:latest`. Here we used multi staged build in which in 1st stage we build the app and generated the static files and in 2nd stage we don't need any other dependencies other than `nuxt-start` which works fine in `node:alpine` image. 

You can change this dockerfile as per your need, for any standard nuxt application setup this dockerfile works just fine. Any other suggestion will be appreciated.

PR #5444, opened 2019-04-02
fix(webpack): preserve `externals` set by `build.extend`

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/webpack/src/config/server.js
+++ b/packages/webpack/src/config/server.js
@@ -95,7 +95,7 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
         maxEntrypointSize: Infinity,
         maxAssetSize: Infinity
       },
-      externals: []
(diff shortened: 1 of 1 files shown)


packages/webpack/src/config/server.js
```diff
@@ -95,7 +95,7 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
         maxEntrypointSize: Infinity,
         maxAssetSize: Infinity
       },
-      externals: []
+      externals: [].concat(config.externals || [])
```
pi0: Can be simplified to `Array.from(config.externals)` ?

clarkdo: Array.from('node_modules') is `["n", "o", "d", "e", "_", "m", "o", "d", "u", "l", "e", "s"]`

PR #5474, opened 2019-04-05
fix: set default for missing cspScriptSrcHashes

zoellner:
## Types of changes

## Description

Minor change setting a default to cspScriptSrcHashes if it is not set. Prevents bug when using CSP in SPA mode in local development

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -122,7 +122,7 @@ const defaultPushAssets = (preloadFiles, shouldPush, publicPath, options) => {
   return links
 }
 
-const getCspString = ({ cspScriptSrcHashes, allowedSources, policies, isDev }) => {
+const getCspString = ({ cspScriptSrcHashes = [], allowedSources, policies, isDev }) => {
   const joinedHashes = cspScriptSrcHashes.join(' ')
   const baseCspStr = `script-src 'self'${isDev ? ` 'unsafe-eval'` : ''} ${joinedHashes}`
 


pi0 (approved):

TheAlexLichter (approved):

packages/server/src/middleware/nuxt.js
```diff
@@ -122,7 +122,7 @@ const defaultPushAssets = (preloadFiles, shouldPush, publicPath, options) => {
   return links
 }
 
-const getCspString = ({ cspScriptSrcHashes, allowedSources, policies, isDev }) => {
+const getCspString = ({ cspScriptSrcHashes = [], allowedSources, policies, isDev }) => {
```
pi0: This function should not be called in this case at all. I'll add some change

pi0:
Fixed in 2ca08a6bf36bde92d8ab6c3084ea489d642576c3

packages/server/src/middleware/nuxt.js
williamchong007: Wouldn't there be a place where hashes were not defined, but other CSP policies is defined?
especially when https://github.com/nuxt/nuxt.js/pull/5387 is in, a simple `unsafe-inline` would cause `cspScriptSrcHashes` to be empty

pi0: @williamchong007 we are checking `cspScriptSrcHashes` to be defined before calling `getCspString`. The default value of `cspScriptSrcHashes` is an empty array from `renderer.js`. (So still leads to calling this function)

PR #5475, opened 2019-04-06
examples: improve storybook and add with-lokijs

husayt:
## Types of changes

## Checklist:


examples/with-lokijs/nuxt.config.js
```diff
@@ -0,0 +1,80 @@
...
+  loading: { color: "green" },
+
+  /*
+   ** Add server middleware
+   ** Nuxt.js uses `connect` module as server
+   ** So most of express middleware works with nuxt.js server middleware
+   */
+  serverMiddleware: [
+    // body-parser middleware
+    bodyParser.json(),
```
pi0: Using bodyParser for all requests slows down SSR and makes problem with Proxy. So i suggest moving it to express middleware stack.

examples/with-lokijs/renovate.json
```diff
@@ -0,0 +1,6 @@
+{
```
pi0: I think this would not be required for example

examples/with-lokijs/store/index.js
```diff
@@ -0,0 +1,93 @@
...
+          },
+          autosave: false
+          // autosaveInterval: 4000
+        })
+      })
+      console.log("async loaded,", res.data)
+      await dispatch("LOAD_DB")
+    } else {
+      console.log(
+        "HEYYYYYYYYYYYYWHOOOOOOISCALLLINGGGGGGGGGGGGMEEEEEEEEEEEEE???????"
```
pi0: :D

examples/with-lokijs/store/index.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+export const actions = {
+  async nuxtServerInit({ dispatch }) {
+    console.log("STARTED SERVER")
+
+    await dispatch("INIT_DB")
+  },
+  async INIT_DB({ dispatch }) {
```
pi0: The preferred way would be creating a plugin like `loki.server.js`

examples/with-lokijs/.codesandbox/workspace.json
```diff
@@ -0,0 +1,22 @@
+{
```
pi0: I think either this file is not required too.

examples/with-lokijs/api/index.js
```diff
@@ -0,0 +1,112 @@
+console.log("API INITIAL ================================================== ")
```
pi0: No function calls should happen before imports.

examples/with-lokijs/store/index.js
```diff
@@ -0,0 +1,93 @@
...
+    if (process.server) {
+      const res = await new Promise(resolve => {
+        const Loki = require("lokijs")
+        db = new Loki("quickstart.db", {
```
pi0: Doing this inside nuxtServerInit or plugin function makes Loki to be loaded per each SSR request. It has to be done in a global scope.

pi0: Plugin can inject `$loki` into context so other store action would benefit from it.

pi0 (changes requested):

pi0:
Thanks @husayt for all your efforts on this. If you can split this PR into separate ones for each example that would be GREAT. There are also some lint problems to be resolved.

husayt:
Thanks. I am currently away.
Will look into it when I get back.

manniL:
cc @husayt

examples/with-lokijs/nuxt.config.js
husayt: very good point. Thanks

husayt:
@pi0 I have taken into account all your feedback and rewritten the whole thing. [Here](https://github.com/nuxt/nuxt.js/pull/7034) is a new pull request for this.

PR #5478, opened 2019-04-07
fix(cli): throw error with proper code for failed imports

pi0:
## Types of changes

## Description

The usage of `_import` was originally expanding search path. Using `consola.fatal` disallows catching errors.

This PR fixes problem with typescript package detection by not forcing the user to install `@nuxt/typescript` when `tsconfig.json` exists in a project. ([source](https://github.com/nuxt/nuxt.js/blob/v2.6.1/packages/cli/src/utils/typescript.js#L22)) -- Thanks to @rchl for reporting this issue.

## Checklist:

### DIFF
--- a/packages/cli/src/imports.js
+++ b/packages/cli/src/imports.js
@@ -1,5 +1,4 @@
 import path from 'path'
-import consola from 'consola'
 
 const localNodeModules = path.resolve(process.cwd(), 'node_modules')
 
--- a/packages/cli/src/imports.js
+++ b/packages/cli/src/imports.js
@@ -5,24 +5,22 @@ const localNodeModules = path.resolve(process.cwd(), 'node_modules')
 
 // Prefer importing modules from local node_modules (for NPX and global bin)
 async function _import(modulePath) {
-  let m
-  for (const mp of [ path.resolve(localNodeModules, modulePath), modulePath ]) {
+  for (const mp of [
+    path.resolve(localNodeModules, modulePath),
+    modulePath
+  ]) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/imports.js
```diff
@@ -5,24 +5,22 @@ const localNodeModules = path.resolve(process.cwd(), 'node_modules')
 
 // Prefer importing modules from local node_modules (for NPX and global bin)
 async function _import(modulePath) {
-  let m
-  for (const mp of [ path.resolve(localNodeModules, modulePath), modulePath ]) {
+  for (const mp of [
+    path.resolve(localNodeModules, modulePath),
+    modulePath
+  ]) {
     try {
```
clarkdo: Is await for more readable?

pi0: Without await, (and returning promise) we cannot catch error here. One way would be separating it like this:

```js
const m = await import(mp)
return m
} catch(e) [
```

But I taught one-liner would be more readable.

clarkdo: Will throw inside catch instead of end be better?

pi0: We have to only throw an error if all possible locations (for loop) has been checked. Throwing inside the catch of for loop requires one additional check (`m === mp`) to see if we are reaching to the end or not. This is same as refactoring out of for loop and using early return

clarkdo: Gotcha

clarkdo (dismissed):

PR #5489, opened 2019-04-08
fix(vue-renderer): use `modulepreload` for modern mode spa generate

### LINKED ISSUE #5486: Modern build still loads "nomodule" scripts
### Version

[2.6.1](https://github.com/nuxt.js/releases/tag/2.6.1)

### Reproduction link

[http://example.com/sorry_it_is_impossible](http://example.com/sorry_it_is_impossible)

### Steps to reproduce

Have a simple nuxt.js SPA.

`nuxt.config.js`

```js
module.exports = {
    mode: "spa",

    modern: "client",

...
```

Build it:
```sh
NODE_ENV=production ./node_modules/.bin/nuxt generate
```

### What is expected ?

See the picture. Only the green scripts are expected to be loaded.

The "nomodule" scripts are still loaded.

### Additional comments?

How would I completely drop support of the non-modern browsers?

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9013">#c9013</a>)</em></sub></div>

clarkdo:
## Types of changes

## Description

Fix #5486

In generate, use `modulepreload` modern resources instead of `preload` legacy resources for avoiding downloading both `modern` and `legacy` resources.

@pi0 I'm using `_generate` in context, not sure if it's a stable flag, please review 😄  

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/spa-meta.js
+++ b/packages/vue-renderer/src/spa-meta.js
@@ -30,8 +30,9 @@ export default class SPAMetaRenderer {
     return vm.$meta().inject()
   }
 
-  async render({ url = '/', req = {} }) {
-    const cacheKey = `${req.modernMode ? 'modern:' : 'legacy:'}${url}`
+  async render({ url = '/', req = {}, _generate }) {
+    const modern = req.modernMode || _generate
+    const cacheKey = `${modern ? 'modern:' : 'legacy:'}${url}`
     let meta = this.cache.get(cacheKey)
 
     if (meta) {
@@ -75,7 +76,7 @@ export default class SPAMetaRenderer {
     meta.resourceHints = ''
 
     const { resources: { modernManifest, clientManifest } } = this.renderer.context
-    const manifest = req.modernMode ? modernManifest : clientManifest
+    const manifest = modern ? modernManifest : clientManifest
 
     const { shouldPreload, shouldPrefetch } = this.options.render.bundleRenderer
 
@@ -90,7 +91,7 @@ export default class SPAMetaRenderer {
         meta.preloadFiles = manifest.initial
           .map(SPAMetaRenderer.normalizeFile)
           .filter(({ fileWithoutQuery, asType }) => shouldPreload(fileWithoutQuery, asType))
-          .map(file => ({ ...file, modern: req.modernMode }))
+          .map(file => ({ ...file, modern }))
 
         meta.resourceHints += meta.preloadFiles
           .map(({ file, extension, fileWithoutQuery, asType, modern }) => {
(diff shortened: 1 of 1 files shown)


packages/vue-renderer/src/spa-meta.js
```diff
@@ -30,8 +30,9 @@ export default class SPAMetaRenderer {
     return vm.$meta().inject()
   }
 
-  async render({ url = '/', req = {} }) {
-    const cacheKey = `${req.modernMode ? 'modern:' : 'legacy:'}${url}`
+  async render({ url = '/', req = {}, _generate }) {
+    const modern = req.modernMode || _generate
```
pi0: How `nuxt generate` implies modern mode is enabled?

clarkdo: `req.modernMode` is set through middleware which is not called in `generate`.

With this fix, the behaviour is same as `vue-cli` doing

`modulepreload`: all modern js, only in modern browsers
`<script type='module'>`: all modern js, only in modern browsers
`<script nomodule>`: all legacy js, only in legacy browsers

Atinux: I agree on this, we want to make sure he built with the `modern` mode too no?

clarkdo: Oh, sorry, I missed the `modern` is not enabled case

clarkdo: @pi0 @Atinux I added `options.modern` checking, only generate `modern` resources in modern mode.

packages/vue-renderer/src/spa-meta.js
```diff
@@ -30,8 +30,9 @@ export default class SPAMetaRenderer {
     return vm.$meta().inject()
   }
 
-  async render({ url = '/', req = {} }) {
-    const cacheKey = `${req.modernMode ? 'modern:' : 'legacy:'}${url}`
+  async render({ url = '/', req = {}, _generate }) {
```
pi0: This is extracting `renderRoute` `context`. Are you sure this context also includes `_generate`? I think it should be accessed with `this.options._generate`

clarkdo: I found this flag from https://github.com/nuxt/nuxt.js/blob/dev/packages/generator/src/generator.js#L204

The reason why I did't use `this.options._generate` is: `generate` in `build` or programmatic `generate` don't have `_generate`

pi0: Nice find 😊 I think that's an older flag but you are right

pi0 (approved):

pi0 (approved):

Atinux:
One last question, will it work with `nuxt build --spa --modern`?

clarkdo:
> One last question, will it work with `nuxt build --spa --modern`?

Yes, it works

PR #5500, opened 2019-04-09
fix(cli): config cache invalidation + refactors

### LINKED ISSUE #5476: Nuxt v2.6.0 ignores nuxt.config.js
### Version

[v2.6.0](https://github.com/nuxt.js/releases/tag/v2.6.0)

### Reproduction link

[https://cmty.app/nuxt/issues/new?type=bug-report&repo=nuxt.js](https://cmty.app/nuxt/issues/new?type=bug-report&repo=nuxt.js)

### Steps to reproduce

I did an npm install that pulled down v2.6.0 and my app broke. After investigation if appears that the config file is being completely ignored. Rolling back to v2.5.1 fixed it.

### What is expected ?

Read config file

I did an npm install that pulled down v2.6.0 and my app broke. After investigation if appears that the config file is being completely ignored. Rolling back to v2.5.1 fixed it.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9004">#c9004</a>)</em></sub></div>

pi0:
This is PR #5500 :dancing_men: 

## Types of changes

## Description

Currently changing `nuxt.config.js` needs a full server to reload as cache invalidation is broken since 2.5.0 after typescript additions. (#5476)

- Refactor some utils into their own file (`utils/config` and `utils/banner`)
- Add CJS cache invalidation

This PR also introduces fixes and enhancements for users that split config file into multiple files.

- Correctly invalidate cache for all config related files
- Watch all dependencies of `nuxt.config.js` (outside `node_modules`) and reload `nuxt dev` if they have changed

Resolves #5476 

## Checklist:


packages/cli/src/utils/config.js
```diff
@@ -0,0 +1,97 @@
...
+    host: argv.hostname || undefined,
+    socket: argv['unix-socket'] || undefined
+  }, options.server || {}, getDefaultNuxtConfig().server)
+
+  return options
+}
+
+function clearRequireCache(id) {
+  const entry = require.cache[id]
+  if (!entry || /node_modules/.test(id)) {
```
pimlie: Maybe use `id.includes('node_modules')` instead of a regex? If not, it could maybe be a small improvement to compile the regex once instead of on every fn call

packages/cli/src/utils/config.js
```diff
@@ -0,0 +1,97 @@
...
+  for (const child of entry.children) {
+    clearRequireCache(child.id)
+  }
+
+  delete require.cache[id]
+}
+
+function scanRequireTree(id, files = new Set()) {
+  const entry = require.cache[id]
+  if (!entry || /node_modules/.test(id) || files.has(id)) {
```
pimlie: same

packages/cli/src/utils/config.js
pi0: I think V8 precompiles and caches in case of using in such a place. BTW this is a matter of 1-2ms for an occasional reload or startup of app. But yes. String.includes can also work :)

packages/cli/src/utils/config.js
```diff
@@ -0,0 +1,97 @@
...
+  const entry = require.cache[id]
+  if (!entry || id.includes('node_modules')) {
+    return
+  }
+
+  for (const child of entry.children) {
+    clearRequireCache(child.id)
+  }
+
+  delete require.cache[id]
```
clarkdo: @pi0 
How do you think delete `nuxt.config.js` module from its parent ?

For example:
```js
const parent = require.cache[nuxtConfigFile].parent
delete require.cache[nuxtConfigFile]
console.log(parent.children) // `nuxt.config.js` module still exists in `children`
```

pi0: It is an unsafe operation to modify parent. Clearing from global cache just works fine :)

clarkdo: I'm just worrying about the old config module is still in memory and maybe can not be collected.

pi0: You are right. Despite [docs](https://nodejs.org/api/modules.html#modules_require_cache) parent keeps a reference of all old child. It is fixed now.

clarkdo (dismissed):

pimlie (dismissed):

kevinmarrec (dismissed):

PR #5509, opened 2019-04-11
Remove inaccurate and misleading "memory usage" in dev CLI output

koresar:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

After introducing the new DX feature - the "building project" page - the memory usage is printed **before** the app start. Whereas previously the memory usage was printed **after** the app start.

image

On the picture above the 58MB is quite inaccurate. Actual memory usage is around 300-500MB.

**Suggestion - do not print memory usage at all.**

Hence the PR.

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #5514, opened 2019-04-11
fix(cli): show memory usage after build for `nuxt dev`

pimlie:
## Types of changes

## Description

As reported by @koresar in #5509, the memory usage is printed too early for the dev command. This pr makes it optional to print memory usage in the banner and adds a simple show mem usage method for printing with consola.info

## Checklist:


koresar:
Thank you. I appreciate!

packages/cli/src/utils/banner.js
```diff
@@ -1,10 +1,10 @@
 import consola from 'consola'
-import prettyBytes from 'pretty-bytes'
 import env from 'std-env'
 import chalk from 'chalk'
 import { successBox } from './formatting'
+import { getMemoryUsage } from './memory'
 
-export function showBanner(nuxt) {
+export function showBanner(nuxt, showMemoryUsage = true) {
```
pi0: Banner will be always shown by CLI before any build happening, so it may be more consistent to show it only as a info message not in banner.

packages/cli/src/commands/dev.js
```diff
@@ -42,6 +43,7 @@ export default {
     // Setup hooks
     nuxt.hook('watch:restart', payload => this.onWatchRestart(payload, { nuxt, builder, cmd, argv }))
     nuxt.hook('bundler:change', changedFileName => this.onBundlerChange(changedFileName))
+    nuxt.hook('build:done', showMemoryUsage)
```
pi0: This hook is not exact as the bundle will be loaded after it. Maybe `render:resourcesLoaded` hook would fit better. or a conditional new hook before it to check `if ((this.isReady) || (this.context.options.modern && this.isModernReady))`

pi0: Also, a timer with a low amount of timeout to ensure GC freed up build time unreferenced resources.

packages/cli/src/utils/banner.js
pimlie: showBanner seems only to be used in `dev` and `start`, for start I think it would look strange to print an additional line

pi0: OK. Makes sense

packages/cli/src/commands/dev.js
pimlie: Fyi, the reason I added the hook is so the memory usage text is printed before the line `Waiting for file changes`. But I indeed already noticed that when you print memory usage after that line is its 1M bigger, will have a look at this

pimlie: Ah wait, [the waiting message is also done from `build:done` hook](https://github.com/nuxt/nuxt.js/blob/e5369db2d539cf580e9ce609e895c67bc01c9b3c/packages/builder/src/builder.js#L56)...

So currently it is:
```js
ℹ Memory usage: 256 MB (RSS: 335 MB)                                                                                     09:10:16
ℹ Waiting for file changes                                                                                               09:10:16
```
but if we delay printing the memory usage it will become
```js
ℹ Waiting for file changes                                                                                               09:10:16
ℹ Memory usage: 257 MB (RSS: 335 MB)                                                                                     09:10:16
```
Allthough technically the second is more correct, personally I like the first one better. Wdyt?

pi0: I think we should either show it in the most exact manner or not at all. Is second one delay+renderer hook or just delay?

pimlie: The second one could also be just in `commands/dev` just after `await builder.build()`

pi0: For production it is correct but for dev, renderer will load resources in the background and final memory usage depends on its behavior (if any leak happens or big amount of data loaded)

pimlie: I checked your suggestion btw about gc but just setting a timeout doesnt seem to work.

What you see below is the result of:
```js
// packages/cli/src/commands/dev.js
    // Start Build
    await builder.build()

    // Print memory usage
    showMemoryUsage()
    for (let i = 0;i < 20; i++) {
      await new Promise(resolve => setTimeout(resolve, 100))
      showMemoryUsage()
    }

    global.gc()
    showMemoryUsage()
```
Only calling global.gc seems to have an effect

```js
ℹ Waiting for file changes                                                                                               09:21:54
ℹ Memory usage: 243 MB (RSS: 329 MB)                                                                                     09:21:54
ℹ Memory usage: 245 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:54
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 250 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 251 MB (RSS: 334 MB)                                                                                     09:21:55
ℹ Memory usage: 251 MB (RSS: 334 MB)                                                                                     09:21:56
ℹ Memory usage: 251 MB (RSS: 334 MB)                                                                                     09:21:56
ℹ Memory usage: 213 MB (RSS: 333 MB)                                                                                     09:21:56
```

pi0: global.gc() is not exposed without a flag. We can imediately show it after renderer is ready.

pi0 (changes requested):
Please change it to show usage after renderer loaded resources

pimlie:
`render:resourcesLoaded` is called twice and `render:done` is never called, will have a look at your other suggestion of adding another hook

pimlie:
This is getting way too complicated for a simple and not precise feature anyway, but I cant get a 'render:resourcesDone' hook as proposed by @pi0 to work because `isModernReady` already returns true on the first call to `build:resources` from webpack/builder which causes the new `resourcesDone` hook still to be called twice.

I can fix it probably but then I need to change the isReady / isModernReady behaviour in renderer so they only return true when their respective `build:resources` hook has been called from webpack/builder. Or do you have a better idea?

packages/cli/src/utils/memory.js
```diff
@@ -0,0 +1,13 @@
+import chalk from 'chalk'
+import consola from 'consola'
+import prettyBytes from 'pretty-bytes'
+
+export function getMemoryUsage() {
+  // https://nodejs.org/api/process.html#process_process_memoryusage
+  const { heapUsed, rss } = process.memoryUsage()
+  return `Memory usage: ${chalk.bold(prettyBytes(heapUsed))} (RSS: ${prettyBytes(rss)})`
```
Atinux: About `getMemoryUsage`, I would return an object instead of a string directly:

```suggestion
  return { heap: heapUsed, rss }
```

packages/cli/src/utils/memory.js
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export function showMemoryUsage() {
+  consola.info(getMemoryUsage())
```
Atinux: And here I will add the text :)

```suggestion
  const { heap, rss } = getMemoryUsage()
  consola.info(`Memory usage: ${chalk.bold(prettyBytes(heap))} (RSS: ${prettyBytes(rss)})`)
```

atinux (changes requested):
I like the idea of showing the correct memory usage in dev, thanks for the PR <3

Actually we need to wait for the `renderer` to be ready (since the bundleRenderer will take some memory usage).

We need indeed to add a new hook for this, I believe together we will find a nice way to do so :)

pi0 (dismissed):

packages/cli/src/utils/memory.js
```diff
@@ -5,7 +5,7 @@ import prettyBytes from 'pretty-bytes'
 export function getMemoryUsage() {
   // https://nodejs.org/api/process.html#process_process_memoryusage
   const { heapUsed, rss } = process.memoryUsage()
-  return `Memory usage: ${chalk.bold(prettyBytes(heapUsed))} (RSS: ${prettyBytes(rss)})`
+  return { heap: heapUsed, rss }
```
pi0: This is breaking banner.js

PR #5516, opened 2019-04-12
fix(vue-app): consider `watchQuery` option in `routerViewKey`

hatashiro:
## Types of changes

## Description

According to the Nuxt documentation, the default `key` provided to `<nuxt-child>` should be `$route.fullPath`. Please check the reference of `nuxtChildKey` in the following link:

- https://nuxtjs.org/api/components-nuxt#the-lt-nuxt-gt-component

However, the current code seems like using `$route.path`. Because of that, transition is not fired correctly when moving between the same page component, by changing query while keeping path the same.

I guess the documentation is describing the correct behavior (`fullPath`) and the current code should be changed.

It's my first contribution on Nuxt, so please feel free to guide me if I did something wrong :)

### References

- https://github.com/vuejs/vue-router/issues/802
- https://github.com/nuxt/nuxt.js/pull/1022

### Commits

This p-r has 4 commits.

1. Before I work on the fix, I investigated test cases. I found the current scroll-to-top test is only checking `scrollToTop: false`, so I added a test case to check `scrollToTop: true` case.
2. Before I add a failing test case, I found that there is no helper method on browser util to go back. I added `go()` helper method to browser test util.
3. The last 2 commits are to add a failing test case and actually fix the issue.

## Checklist:


Atinux:
Hi @utatti and thank you for your PR.

I remember having something like this by default for a while, you can see some history here: https://github.com/nuxt/nuxt.js/pull/1022

You can see in this issue why I had to change the behaviour: https://github.com/nuxt/nuxt.js/issues/1255#issuecomment-343143312

This is not as simple as it looks like when doing such changes :/

But I do like what you did on the tests, and I think we should be able to:
- Use a computed property to use `$route.fullPath`
- Test this behaviour
- Update the docs to explain what is the actual default and how to customize it

utatti:
So if I understand correctly, we keep the current behaviour and update the doc? I also thought that using the `key` option should be mentioned in the doc too.

Also, I still think that triggerScroll should at least work with query changes specified in `watchQuery`. What so you think?

utatti:
If using `fullPath` is too radical, I'd like to suggest using `watchQuery` like below:

```js
let key = this.$route.path
if (Component && Component.options && Component.options.watchQuery) {
  const watchQuery = Component.options.watchQuery
  if (watchQuery === true) {
    key = this.$route.fullPath
  } else if (Array.isArray(watchQuery)) {
    key += `?${watchQuery.map(key => `${key}=${this.$route.query[key]}`).join('&')}`
  }
}
return key
```

The reason why I'd like to add a way to consider query in building key is that saving/restoring scroll position doesn't work currently when only query is changed. It's basically because query change is not considered as a vue-router transition (thus no `beforeEnter` hook), because they use the same `key`.

I understand reusing component is important, but I guess considering `watchQuery` is still an option?

Let me know your thought about the change above. I'll add follow-up fixes and test cases.

Cheers

utatti:
I've updated the p-r and now it's to consider `watchQuery` on building `routerViewKey`.

Atinux:
Hi @utatti 

Thank you so much for investigating into this, I like your changes actually!

Could you also help to update the documentation to explain a bit more how it works, then I am ready to merge this PR :)

(https://github.com/nuxt/docs/blob/master/en/api/components-nuxt.md)

utatti:
@Atinux 

Thanks! I've just uploaded a p-r on docs: https://github.com/nuxt/docs/pull/1331

manniL:
As @clarkdo correctly mentioned in nuxt/docs#1331, this is a breaking change.

Atinux:
Actually it's not really a breaking change, but mostly an improvement in the DX about transitions and scrollToTop from my POV

atinux (dismissed):

pi0:
@clarkdo Can you please re-review this PR?

packages/vue-app/template/components/nuxt.js
```diff
@@ -40,10 +40,33 @@ export default {
       if (typeof this.nuxtChildKey !== 'undefined' || this.$route.matched.length > 1) {
         return this.nuxtChildKey || compile(this.$route.matched[0].path)(this.$route.params)
       }
+
       const Component = this.$route.matched[0] && this.$route.matched[0].components.default
-      if (Component && Component.options && Component.options.key) {
-        return (typeof Component.options.key === 'function' ? Component.options.key(this.$route) : Component.options.key)
+      if (Component && Component.options) {
```
clarkdo:
```suggestion
      if (Component && Component.options) {
        const { key, watchQuery } = Component.options

        if (key) {
          return (typeof key === 'function' ? key(this.$route) : key)
        } else if (watchQuery === true) {
          return this.$route.fullPath
        } else if (Array.isArray(watchQuery)) {
          const pickedQuery = {}
          for (const key of watchQuery) {
              pickedQuery[key] = this.$route.query[key]
          }

          return this.$router.resolve({
              path: this.$route.path,
              query: pickedQuery
          }).href
        }
      }
```

packages/vue-app/template/components/nuxt.js
```diff
@@ -40,10 +40,33 @@ export default {
...
+
+        if (options.key) {
+          return (typeof options.key === 'function' ? options.key(this.$route) : options.key)
+        }
+
+        if (options.watchQuery) {
+          const watchQuery = options.watchQuery
+          if (watchQuery === true) {
+            return this.$route.fullPath
+          } else if (Array.isArray(watchQuery)) {
```
clarkdo: If `watchQuery` is `[]`, is `router.resolve` or `fullPath` supposed to be the expected value ?

clarkdo (commented):
Not a breaking change, new option `watchQuery` as great enhancement.

All looks good, expect tiny change in my opinion.

utatti:
Thanks for your comment. I've updated the code.

packages/vue-app/template/components/nuxt.js
```diff
@@ -40,10 +40,30 @@ export default {
...
       }
+
       const Component = this.$route.matched[0] && this.$route.matched[0].components.default
-      if (Component && Component.options && Component.options.key) {
-        return (typeof Component.options.key === 'function' ? Component.options.key(this.$route) : Component.options.key)
+      if (Component && Component.options) {
+        const { key, watchQuery } = Component.options
+        if (key) {
+          return (typeof key === 'function' ? key(this.$route) : key)
+        } else if (watchQuery === true) {
```
clarkdo: How do you think ?
```suggestion
        } else if (watchQuery) {
          if (watchQuery.length) {
            const pickedQuery = {}
            for (const key of watchQuery) {
                pickedQuery[key] = this.$route.query[key]
            }
            return this.$router.resolve({
                path: this.$route.path,
                query: pickedQuery
            }).href
          } else {
            return this.$route.fullPath
          }
        }
      }
    }
```

utatti: Fixed!

clarkdo (dismissed):
LGTM

TheAlexLichter (dismissed):
Almost LGMT from my POV ☺️ . Would go with `if` instead of `else if` and remove the last `else` then (keeping the return though).

https://github.com/nuxt/nuxt.js/blob/ded2eae1f019ea5220799742a414d2ef8e0965d9/packages/vue-app/template/components/nuxt.js#L47-L61

packages/vue-app/template/components/nuxt.js
```diff
@@ -44,9 +44,12 @@ export default {
...
+
         if (key) {
           return (typeof key === 'function' ? key(this.$route) : key)
-        } else if (watchQuery) {
+        }
+
+        if (watchQuery) {
           if (watchQuery.length) {
             const pickedQuery = {}
             for (const key of watchQuery) {
```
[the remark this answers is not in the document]
utatti: Fixed!

atinux (dismissed):
Good for me now

packages/vue-app/template/components/nuxt.js
```diff
@@ -40,10 +40,30 @@ export default {
...
+
+        if (key) {
+          return (typeof key === 'function' ? key(this.$route) : key)
+        }
+
+        if (watchQuery) {
+          if (watchQuery.length) {
+            const pickedQuery = {}
+            for (const queryKey of watchQuery) {
+                pickedQuery[queryKey] = this.$route.query[queryKey]
```
galvez: Please fix indentation to 2 spaces here and other places I noticed it's off.

galvez (dismissed):
LGTM aside from formatting issues which I've commented on.

packages/vue-app/template/components/nuxt.js
utatti: I've fixed wrong indents.

pi0 (approved):

clarkdo (approved):

utatti:
Thanks all!

clarkdo:
Hi @utatti , as #5516 reported, this pr is a breaking change in some case.

How do you think change the `fullPath` option more compatible ? 

Actually your requirement can be implemented by:

```js
export default {
  key(route) {
    return route.fullPath
  }
}
```

utatti:
@clarkdo Yes, this is a breaking change in that manner.

I understand we can use the `key` option, but I still think that it's natural to consider `watchQuery` on creating `routerViewKey`.

According to the `watchQuery` documentation:

> If the defined strings change, all component methods (asyncData, fetch, validate, layout, ...) will be called.

However, only the scroll restoration and transition hooks are not working with `watchQuery`.

I understand the final decision is up to maintainers, but I personally think the current behaviour is natural and the case like #5738 (where `watchQuery` is used but the component should be reused) should be handled as an exception.

dhritzkiv:
@utatti Hmm. I would argue that re-using the component as much as possible should not be the exception, but is instead the Vue way.

As it stands, there is no way to achieve the old behaviour in `2.7.x`, but under `2.6.x`, you should have been able to achieve a Nuxt-child key override with `key`. If `bar` was the query string in `watchQuery` that you wanted to invalidate the component (thus forcing a complete re-creation), then this should work:

```ts
// this PR's behaviour in v2.6.3
export default {
  key(route) {
    return route.path + route.query.bar
  }
}
```

utatti:
I understand the we can override the route key using `key(route)` in `2.6.x`. The issue is not if we can achieve it or not, but how we achieve it.

I agree with re-using the component is essential, but the key discussion here is what `watchQuery` results in:

1. `2.6.x`: `asyncData`, `fetch`, `validate`, `layout`, ..., most of data and methods are re-calculated and re-called, but no transition hook and scroll restoration.
2. `2.7.x`: Everything is re-created, works as usual page transition between different page components.

What I meant was 2 seemed natural. I didn't mean that re-using component is exceptional.

I'm actually good to revert the change if scroll restoration works at least with `watchQuery`. Queried transition in the same page component is actually a super common case: imagine building a paginated page with a `page=n` query.

I'd like to suggest the following solution, avoiding breaking change:

```js
{
  watchQuery: true, // work as 2.6.x
  watchQuery: ['somequery'], // work as 2.6.x

  // introduce watchQuery as object 
  watchQuery: { 
    somequery: true, // work as 2.6.x
    somequery: 'reload', // work as 2.6.x
    somequery: anyTruthyValue, // work as 2.6.x
    
    somequery: 'transition', // work as 2.7.x
  }
}
```

utatti:
Today I'll investigate if there is a better way that we reuse page components and make scroll restoration work at the same time.

PR #5517, opened 2019-04-12
feat(generator): option to merge generate output log into single line

### LINKED ISSUE #5510: Setting to disable console.logs per generated page

* When generating thousands of pages the build log becomes enormous.
* I'm using Netlify for hosting my static website. They're streaming their build logs to their UI and have a build time limit of 30 minutes. When generating thousands of pages, a delay starts to exist between the log output and the log preview in their UI, causing the build to exceed the build time limit.

Add an option to the generate options to make logging a line per generated page conditional. Just log something like `Generated 9000 pages`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9030">#c9030</a>)</em></sub></div>

Frankwarnaar:
Configuration option to merge the logs of all the generate routes, to just one line displaying `Generated x routes`.

## Types of changes

## Description

Resolves #5510

## Checklist:


Atinux:
Hi @Frankwarnaar 

Thank you for this PR (and feature request) 💚 

I actually like this idea, avoiding to bloat the console when having a lot of route to generate.

Could you fix the failing tests? See https://circleci.com/gh/nuxt/nuxt.js/38598?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Also, the PR for the docs should be on these files:
- https://github.com/nuxt/docs/blob/master/en/api/configuration-generate.md
- https://github.com/nuxt/docs/blob/master/en/api/menu.json#L87-L97

To @nuxt/core-team: I will be in favour of having this behaviour by default for Nuxt 3

manniL:
@Atinux agree to have that as default, at least for NODE_ENV !=== development
For dev it might be useful to still have the "full log".

pimlie:
This sounds mostly useful for CI environments, so I would like to propose to only make it default on CI environments.

But although quite a bit more work, I think a log output aggregator that behaves similar like eg webpackbar would be much better. Especially when the generate command also can use multiple nuxt workers (and replaces nuxt generate cluster). Then you could have something like a `top` command to see all statistics while generating your pages. I looked briefly into this for ncg some time ago but decided to not continue working on it because of the expected changes due to nuxt worker adaptation.

pimlie:
Also, wouldnt it be more useful in the long run to control log verbosity by the log level? Eg if you set the log level to `2` the `consola.success` messages are already ignored

clarkdo (commented):
@Atinux @pi0 @pimlie @Frankwarnaar 

Instead of introducing a new option, I have another idea.

We can add `Generated ${count} routes` as a summary info for generation.

And also keep `Generated ${route}` log, but may be with a lower log level `debug` or a scope which is enabled by default, user can switch it off by having higher log level or ignoring scope in `consola`.

Because we may have other similar requirements in future, so introducing more new options may increase complexity. Use an unified logging system is better in my POV.

Atinux:
100% agree with @clarkdo comment, adding another option mean more complexity and this is related to logging only.

manniL:
Closing here in favor of using logLevels.
Thanks anyway @Frankwarnaar! If you want, you can implement the suggested solution in another PR :relaxed:

Issue #4008, opened 2018-09-29
Asyncdata and fetch called twice in nested route


husayt:
### Version

[v2.0.0](https://github.com/nuxt.js/releases/tag/v2.0.0)

### Reproduction

https://codesandbox.io/s/k0rj8844nv

### Steps to reproduce

let's say I have the following structure:

```
pages\user\_id.vue
pages\user\_id\_name.vue
```

Note , in one _id is a file, in the other is folder

### What is expected ?

only `pages\user\_id\_name.vue` should be rendered.

### What is actually happening?

when I go to page /user/12/john
then I see asyncdata being called twice.

```
pages_user__id.js:xxx
pages_user__id__name.js:xxx
```

 seems like both pages being rendered. Asyncdata is called on both, although created and mounted called once

### Additional comments?

This might explain many other asyncdata multiple call issues

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7867">#c7867</a>)</em></sub></div>

manniL:
@husayt Could you add a [CodeSandBox](https://codesandbox.io/s/github/nuxt/codesandbox-nuxt/tree/master/?autoresize=1&hidenavigation=1&view=preview) for reproduction?

Fubinator:
I was just playing around with that:
https://codesandbox.io/s/xl312qo8ow

If you browse to `/user/12/john` it shows the page for the  `_id.vue-file and asyncData gets called for both pages.

manniL:
Thanks @Fubinator! We will look into it :relaxed:

manniL:
Same goes for fetch.

husayt:
Thanks @Fubinator , you did codesandbox, before I could come in

GeneralworksInc:
I found same case.(Maybe)
I avoided this problem that I moved 'computed' variable in component to 'method'.

msacar:
is also called multiple times in this structure

pages/
--| _category/
-----| _subCategory/
--------| _id.vue
--------| index.vue
-----| _subCategory.vue
-----| index.vue
--| _category.vue
--| index.vue

MR-zhou-xx:
hello, What can be expected to solve this problem?

msacar:
Hello you can use custom routes to solve this problem.

in nuxt.config.js file 


```
 router: {
    extendRoutes(routes, resolve) {
[...]

PR #5590, opened 2019-04-23
feat: router.leafRoutes for excluding from nested pages

clarkdo:
## Types of changes

## Description

A potential fix for #4008 and also many similar issues.

In current router.js, if we have pages like:

```
+-- nuxt.config.js
+-- pages
|   +-- _id.vue
|   +-- _id
|       +-- _name.vue
|       +-- _name
|           +-- _sub.vue
```

The generated `routes` will be like:

```js
[{
  path: "/:id",
  component: _1808b1d3,
  name: "id",
  children: [{
    path: ":name?",
    component: _43247de7,
    name: "id-name",
    children: [{
      path: ":sub?",
      component: _19715728,
      name: "id-name-sub"
    }]
  }]
}]
```

In this case, if `_id.vue` **does't contain `NuxtChild`** and we access `http://localhost:3000/foo/123`, only `_id.vue` will be rendered and both `asyncData` will be executed which are all inappropriate.

The expected behaviour should be only `pages/_id/_name.vue` and it's `asyncData` is rendered/executed.

But it's also kind of complex to know that if `page` contains `NuxtChild`, so I proposed this pr which introduces `leafRoutes` (temporary name) for marking the page as a `leaf route`.

The `leaf route` means that the children of this page are not nested, so the generated routes will be like:

```js
// nuxt.config.js
export default {
  router: {
    leafRoutes: ['_id', '_id/_name']
  }
}
// router.js
[{
  path: "/:id",
  component: _1808b1d3,
  name: "id"
}, {
  path: "/:id/:name?",
  component: _43247de7,
  name: "id-name"
}, {
  path: "/:id/:name?/:sub?",
  component: _19715728,
  name: "id-name-sub"
}]
```

## Checklist:


Atinux:
Thanks Clark, it looks good to me at first sight.

But instead of defining again in nuxt.config.js, I would like that we start using router-extras to instead set this « leaf » mode or « not nested » right inside the page. So when removing the file, we don’t have to change anything else.

Cc @pi0

pi0:
Agree with @Atinux about advising to use route-extras for this mode. Can be both an option in `nuxt.config` or using `router-extras`. 

---

What if we suggest such structure to enable leaf mode:

```
+-- nuxt.config.js
+-- pages
|   +-- _slug.vue
|   +-- _id
|       +-- _name.vue
|       +-- _name
|           +-- _sub.vue
```

The point is that if dynamic param names are not equal, we assume they have no relation together. (So no `<nuxt-child>` in `_slug.vue`

PR #5605, opened 2019-04-26
feat(server): configurable app option

galvez:
@pi0 @Atinux @clarkdo @manniL @kevinmarrec @aldarund @alexchopin

Please see http://hire.jonasgalvez.com.br/2019/apr/26/revamping-nuxts-http-server

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -30,8 +30,13 @@ export default class Server {
     // Will be set after listen
     this.listeners = []
 
-    // Create new connect instance
-    this.app = connect()
+    // Use different server instance if provided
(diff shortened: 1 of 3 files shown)


clarkdo:
I think we need type definition for `app` to constraint its `API` like `use`, `req` and `res` or at least doc/comment to describe it, so that user won't make any mistake to use a customized server middleware.

galvez:
@clarkdo as per discussion in the #core channel, I'm withdrawing this in favor of pushing to a full replacement of `connect` soon.

douira:
What does "soon" mean? When is this replacement of connect coming? Is it going to have to wait until nuxt 3?

PR #5620, opened 2019-04-29
fix(builder): call `watch:restart` after `watch:fileChanged` hook

clarkdo:
## Types of changes

## Description

@pi0 Should this be deprecacted ? I don't see any doc is refering it.

If there is anyone using it, this may be a slightly breaking change since there is no `builder` as param anymore.

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -663,12 +663,12 @@ export default class Builder {
     this.createFileWatcher(
       nuxtRestartWatch,
       ['all'],
-      (event, fileName) => {
+      async (event, fileName) => {
         if (['add', 'change', 'unlink'].includes(event) === false) {
           return
         }
-        this.nuxt.callHook('watch:fileChanged', this, fileName) // Legacy
-        this.nuxt.callHook('watch:restart', { event, path: fileName })
+        await this.nuxt.callHook('watch:fileChanged', this, fileName) // Legacy
+        await this.nuxt.callHook('watch:restart', { event, path: fileName })
       },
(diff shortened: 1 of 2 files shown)


pi0 (dismissed):

pi0:
@clarkdo Yes. Although almost all hooks should be deprecated in flavor on new namespaces with nuxt 3. This may just be annoying for some module users that are using this hook. Not something emergency.

Still if you feel is good to deprecate sooner feel free to merge. But i vote for deprecating all once with a single PR in flavor of better namespaces.

clarkdo:
@pi0 Yes, a separated pr make sense.

clarkdo:
@pi0 I reverted the deprecation change and make hooks run in-order

manniL:
@clarkdo I tried to fix the PR titled ☺️

pi0:
If we allow async watch handlers, promise rejection should be somehow handled. Probably via a console.fatal or error.

clarkdo:
The handler is async now since callHook returns promise, the changing in this pr is just keeping order of the execution.

About the `rejection`, will `restart` hook process it now ?

pi0:
@clarkdo Both previously and now errors was unhandled. It was an enhancenment idea to catch and pipe errors to core/nuxt

clarkdo:
> @clarkdo Both previously and now errors was unhandled. It was an enhancenment idea to catch and pipe errors to core/nuxt

Cool, let me add it

pi0:
@clark right. I didn't remembered that.

pi0 (approved):

PR #5635, opened 2019-05-02
feat(webpack): suppress not found typescript warnings

Yama-Tomo:
## Types of changes

## Description

Hi all ✋ 

This PR is suppress `export ... was not found ...` warnings with typescript.

relate: https://github.com/TypeStrong/ts-loader/issues/653

I usually suppress by adding a custom plugin to `nuxt.config.ts` but, I thought it would be useful to support in the Nuxt.

This is a suggestion. please close if you don't needs this  🙇 
Thank you

## Checklist:

- [x] I have updated the documentation accordingly. (PR: #)
  - **TODO**: If you need this PR I will update the document.


packages/webpack/src/config/base.js
```diff
@@ -372,7 +372,18 @@ export default class WebpackBaseConfig {
...
+    plugins.push(new WarnFixPlugin(warn =>
+      !(warn.name === 'ModuleDependencyWarning' &&
+        warn.message.includes(`export 'default'`) &&
+        warn.message.includes('nuxt_plugin_'))
+    ))
+
+    if (_typescript.build && buildOptions.typescript && buildOptions.typescript.ignoreNotFoundWarnings) {
+      // NOTE: this needs to be added to the plugin stack before the `WebpackBar`
+      plugins.push(new WarnFixPlugin(warn =>
+        !(warn.name === 'ModuleDependencyWarning' && /export .* was not found in /.test(warn.message))
```
manniL: You are pushing a 2nd plugin instance here. Wouldn't it be better to compose/alter the filter function instead of a second instance? 🤔 ☺️

Yama-Tomo: exactly. creating multiple instances is redundant...
I will change so that only one instance is enough 👍

kevinmarrec:
Hi @Yama-Tomo, thanks for your work !

I'm not really a fan of such plugins suppressing warnings, so ideally I'd prefer an use case where the warnings are not fixable and would showcase the need of this. Could you maybe provide some reproduction where these warnings are persisting in a Nuxt project ?

Yama-Tomo:
@kevinmarrec  Thanks for your comment!

Here is code for my boilerplate for nuxt project.
https://github.com/Yama-Tomo/vue-vuex-typescript-sample

How to reproduction

- 1: checkout the `nuxt` branch after git clone
　
- 2.a: comment out the L.59 of `extend.webpack.config.ts`
- 3.a: execute `yarn && yarn dev`

or 

- 2.b: comment out the L.32-34 of `nuxt.config.ts`
- 3.b: execute `yarn && yarn build`

kevinmarrec:
@Yama-Tomo Hm alright, seems to be really specific case.

IMO I don't think that such "suppress warning" feature should be included in Nuxt core, I mean for specific cases it could be useful, but I think it could also suppress warnings that people would still want to be shown.

So even if it would be a togglable feature, I'm kind of against that for now but I would like @pi0 opinion about this stuff :)

/cc @pi0

atinux (approved):

kevinmarrec (approved):
I've discussed further with @Atinux on Discord and we decided that's all good :). 

Thanks for your great work @Yama-Tomo !

Atinux:
Thank you for your work @Yama-Tomo 

Could you now open a PR for the docs, the documents to update are:
- https://github.com/nuxt/docs/blob/master/en/api/configuration-build.md#typescript

Please comment that `ignoreNotFoundWarnings` has been introduced in `v2.7.0` :)

pi0 (approved):

Yama-Tomo:
@Atinux  I've created PR of docs 👍 https://github.com/nuxt/docs/pull/1374

PR #5637, opened 2019-05-02
feat(webpack): extendable functional babel.presets and babel envName

clarkdo:
<details>
<summary>Hide Comment</summary>

## What's current situation

Current babel config in Nuxt.js is kind of complex and limited:

1. `build.babel.presets` supports `Function`, `Array` which may be confusing
2. Hard to apply customized config
3. Change options for `@nuxt/babel-preset-app` needs more useless code (via functional babel.presets)

## What's in this pr

This pr suggests using `babel.config.js` in `rootDir` which aligns to [`babel default behaviour`](https://babeljs.io/docs/en/configuration#babelconfigjs), user can easily config babel like:

```js
// babel.config.js
module.exports = function (api) {
  return {
    presets: [
      require.resolve('@nuxt/babel-preset-app'),
      {
        buildTarget: api.env('client') ? 'client' : 'server'
      }
    ],
	plugins: [...]
  }
}
```

For more complex scenario, it's also easier which is not possible in current Nuxt.js:

```js
// babel.config.js
module.exports = function (api) {
  const env = api.env()
  api.cache.using(() => env)

  const presets = []
  const plugins = []

  if (env === 'server') {
    // set server presets/plugins
  } else if (env === 'client') {
    // set client presets/plugins
  } else if (env === 'modern') {
    // set modern presets/plugins
  }

  return { presets, plugins }
}

```

## What's next (Nuxt.js v3)

I suggest that we remove `build.babel` and use native babel config way, actually `babel-loader` has already been using [`babel options`](https://babeljs.io/docs/en/options) as options of loader instead of self-defined options.

The current default preset config can be kept as default value when no babel config file found.

</details>

After discussion with @pi0 , we think keeping good DX and avoiding conflicts are very important.

So I have changed the pr:

1. Support api like `build.extend` for functional `presets`:

> `preset` is @nuxt/babel-preset-app
> `options` are [here](https://github.com/nuxt/nuxt.js/tree/dev/packages/babel-preset-app#options)

```js
// nuxt.config.js
export default {
  build: {
    babel: {
      presets({ isServer }, [preset, options]) {
        // change options directly
        options.targets = ...
        options.corejs = ...

        // or return whole preset list

        return [
          [
            preset, {
              buildTarget: ...,
              corejs: ...,
              ...,
              ...options
          }],
          [
            ...
          ]
        ]
      }
    }
  }
}
```

2. Add envName for users are using `configFile`

⚠️ **NOTE**⚠️ : `configFile` is not suggested to be used only if you know what you're doing

```js
// nuxt.config.js
export default {
  build: {
    babel: {
      configFile: path.resolve(rootDir, 'babel.config.js')
    }
  }
}

// babel.config.js
module.exports = function (api) {
  const env = api.env()
  api.cache.using(() => env)

  const presets = []
  const plugins = []

  if (env === 'server') {
    // set server presets/plugins
  } else if (env === 'client') {
    // set client presets/plugins
  } else if (env === 'modern') {
    // set modern presets/plugins
  }

  return { presets, plugins }
}

```

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -1,7 +1,6 @@
 import path from 'path'
 import consola from 'consola'
 import TimeFixPlugin from 'time-fix-plugin'
-import clone from 'lodash/clone'
 import cloneDeep from 'lodash/cloneDeep'
 import escapeRegExp from 'lodash/escapeRegExp'
 import VueLoader from 'vue-loader'
@@ -70,21 +69,28 @@ export default class WebpackBaseConfig {
   }
 
   getBabelOptions() {
-    const options = clone(this.buildContext.buildOptions.babel)
+    const options = {
+      ...this.buildContext.buildOptions.babel,
+      envName: this.name
+    }
+
+    if (options.configFile !== false) {
+      return options
+    }
+
+    const defaultPreset = [
+      require.resolve('@nuxt/babel-preset-app'),
+      {
+        buildTarget: this.isServer ? 'server' : 'client'
+      }
+    ]
 
     if (typeof options.presets === 'function') {
-      options.presets = options.presets({ isServer: this.isServer })
+      options.presets = options.presets({ isServer: this.isServer }, defaultPreset)
     }
 
     if (!options.babelrc && !options.presets) {
-      options.presets = [
-        [
-          require.resolve('@nuxt/babel-preset-app'),
-          {
-            buildTarget: this.isServer ? 'server' : 'client'
-          }
-        ]
-      ]
+      options.presets = defaultPreset
     }
 
     return options
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -90,7 +90,7 @@ export default class WebpackBaseConfig {
     }
 
     if (!options.babelrc && !options.presets) {
-      options.presets = defaultPreset
+      options.presets = [ defaultPreset ]
     }
 
     return options
--- a/packages/webpack/src/config/modern.js
+++ b/packages/webpack/src/config/modern.js
@@ -1,4 +1,3 @@
-import clone from 'lodash/clone'
 import WebpackClientConfig from './client'
 
 export default class WebpackModernConfig extends WebpackClientConfig {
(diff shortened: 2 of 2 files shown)


pi0:
I think this change is in contrast to other areas and nuxt configuration. We already hide webpack config and for postcss suggest not using a seperate config file. Also depending on external babel config means we will get into a conflict when another babel dependent system and harder to diagnostic user issues because we are not sure what babel config is being applied and how it changes nuxt behaviour.

clarkdo:
@pi0 

> I think this change is in contrast to other areas and nuxt configuration. We already hide webpack config and for postcss suggest not using a separate config file.

I agree, hiding the complex config for users is one of our design rule, but I think this rule should and has been done in `@nuxt/babel-preset-app` which is trying to provide best babel config for users who are not babel experts or don't want to look too much into it.

And `postcss` may be different since it's kind of static, and babel is dynamic because of `server`, `client` and event `modern` all have different settings.

> Also depending on external babel config means we will get into a conflict when another babel dependent project like jest test is on the same project. 

I have a different idea on this, I think one project would be better to unify the syntax, but if it happens, according to babel's suggestion, a `.babelrc` inside `test` folder works or they can set special config by:

```js
// babel.config.js
module.exports = function (api) {
  if (api.env('test')) {
    // set test presets/plugins
  }
  return { ... }
}
```

> And also harder to diagnostic user issues because we are not sure what babel config is being applied and how it changes nuxt behavior.

Yes, this problem is hard, actually it happens in current Nuxt.js as well.
For the users who want to config babel, they will user `build.babel: { presets: [...], plugins: [...]}` to make same trouble.

In my perspective, after this pr, most users are still same as it is now, they don't need `babel.config.js` and we provide the best default solution. This pr solves the problem for the users who are experts and know exactly what they want, like different 'target', `polyfill`, `plugins` or more aggressive `modern` level.

What this proposal is trying to reduce the entries and solve restrictions which are brought from internal-only config

pi0 (approved):

pi0:
A note that could be probably useful for documentation. It is also supported that user directly modify second argument of `presets` and return nothing :)

clarkdo:
@pi0 You’re right, bro, it’s already in the code of pr message, maybe not well organized,I’ll refine it and also create doc or

atinux (approved):

Atinux:
Nice work on this PR @clarkdo 

It keeps simplicity while giving advanced configuration for experts.

It could be nice to add a PR to update https://nuxtjs.org/api/configuration-build#babel

manniL:
Sorry that I missed the conversation here:

What's the current stand regarding applying a .babelrc or babel.config.js? Isn't Nuxt ignoring them by default right now?

clarkdo:
@manniL The `build.babel` in `nuxt.config.js` is still preferred.

Although we support `babel.congfig.js` (disabled by default), but it's not suggested unless if user knows exactly what is the influence.

PR #5673, opened 2019-05-08
feat(vue-renderer/vue-app): report SSR console logs to the browser with consola

atinux:
## Types of changes

## Description

This feature is for **development** mode to improve the developer experience while developing.

### Problem

Switching and looking at the terminal when you want to debug your Nuxt app by doing some `console.log` can be annoying. This can disturb newcomers to universal rendering.

### Solution

In dev only, it copies the SSR logs to the browser and display them before mounting the Vue app.

This is how it look like when adding some logs to `examples/hello-world/pages/index.vue`:
```js
asyncData() {
  console.log('Log')
  console.warn('Warning')
  console.error('Error')
  console.error(new Error('Aie'))
}
```

Result:
Screenshot 2019-05-08 at 17 10 11

## Checklist:

- [x] https://github.com/nuxt/consola/pull/49

@clarkdo I would like your insight regarding the tests and the best place to test this feature.


galvez (dismissed):
YES! So simple, very nicely done @Atinux :)

galvez:
@Atinux perhaps this could be an opt-in feature?

kevinmarrec (dismissed):
Great stuff @Atinux, well done !

clarkdo (commented):
Do you think if this feature needs an extra option ? Maybe sth like: `server.console: true|false|'info'|'warn'|'error'`

If there is any user who is using dev mode in production or integration testing environment, this may be a risk for sensitive information leak.

About the test, I think it could be in fixture like `base`, `ssr` or `with-config`, and verify `NUXT.logs` in `jsdom` or `html string`

packages/vue-app/template/client.js
```diff
@@ -36,6 +37,11 @@ const NUXT = window.<%= globals.context %> || {}
 
 Object.assign(Vue.config, <%= serialize(vue.config) %>)<%= isTest ? '// eslint-disable-line' : '' %>
 
+<% if (isDev) { %>
+const logger = consola.withScope('nuxt:ssr')
+const logs = NUXT.logs || []
+logs.forEach(logObj => logger[logObj.type](logObj))
```
clarkdo: `logs` can be deleted after logging to console.

Atinux:
I added a PR to consola to support `console.warn` on the browser (https://github.com/nuxt/consola/pull/49).

@clarkdo I am not up to add another option for this, it's like for the loading-screen in dev.
They should not run `nuxt dev` in production in any way.

packages/vue-app/template/client.js
Atinux: Added, thanks :)

clarkdo:
@Atinux Another thing comes into my mind is: Do we need to support messages from `console` as well ?

packages/vue-app/template/server.js
```diff
@@ -219,7 +243,7 @@ export default async (ssrContext) => {
     return Promise.all(promises)
   }))
 
-  <% if (isDev) { %>if (asyncDatas.length) debug('Data fetching ' + ssrContext.url + ': ' + (Date.now() - s) + 'ms')<% } %>
+  <% if (debug) { %>if (asyncDatas.length) debug('Data fetching ' + ssrContext.url + ': ' + (Date.now() - s) + 'ms')<% } %>
```
pi0: Do we need `debug` package anymore? We can use `consola.debug` here.

packages/vue-app/template/server.js
```diff
@@ -47,13 +48,33 @@ const createNext = ssrContext => (opts) => {
...
+  // console.log = function (message, ...args) {
+  //   logs.push({ message, stack: this.stack})
+  //   _consoleLog(message, ...args)
+  // }
+  <% if (isDev) { %>
+  const logs = []
+  const devReporter = {
+    log(logObj) {
+      if (logObj.args[0] instanceof Error) {
+        logObj.args[0] = logObj.args[0].stack
```
pi0: Nice check. Devalue will be broken without this.

packages/vue-app/template/server.js
```diff
@@ -47,13 +48,33 @@ const createNext = ssrContext => (opts) => {
 // Since data fetching is async, this function is expected to
 // return a Promise that resolves to the app instance.
 export default async (ssrContext) => {
+  // const _consoleLog = console.log
```
pi0: Are these comments required?

packages/vue-app/template/server.js
Atinux: Yes I’ve got the warning when I tried without :)

packages/vue-app/template/server.js
Atinux: Agree

Atinux: Will it be a breaking change?

pi0:
consola updated to 2.6.1 (#5674)

packages/vue-app/template/server.js
pi0: I don't think :) Usage is the same as before.

pi0 (dismissed):

Atinux:
I could not find a quick way to check for logs since they are mocked in testing and catched by Jest.

pi0 (approved):

PR #5681, opened 2019-05-09
fix(builder): watch store dir and `serverMiddleware` paths

atinux:
## Types of changes

### Problems

Right now, when we use `serverMiddleware` and change them (the files linked to `options.serverMiddleware`), a restart happen but the file is cached, leading to a useless restart.

The other problem is when we start a nuxt app without a `store/` directory, and while running `nuxt dev`, when we add the directory with a file in it (`store/index.js`), Nuxt does nothing and we have to manually restart the project (really annoying).

### Solution

- Watch paths of `serverMiddleware` + clear cache of the file

- Watch `store/` directory for full reload when `this.options.store === false`
nuxt-store-fix

## Checklist:


packages/cli/src/commands/dev.js
```diff
@@ -41,7 +41,7 @@ export default {
     const nuxt = await cmd.getNuxt(config)
 
     // Setup hooks
-    nuxt.hook('watch:restart', payload => this.onWatchRestart(payload, { nuxt, builder, cmd, argv }))
```
Atinux: `builder` was not used here

packages/builder/src/builder.js
```diff
@@ -667,6 +688,11 @@ export default class Builder {
         if (['add', 'change', 'unlink'].includes(event) === false) {
           return
         }
+        /* istanbul ignore if */
+        if (this.serverMiddlewarePaths.includes(fileName)) {
```
pi0: This getter is being computed after every file change. Maybe we can somehow cache it...

clarkdo: Agree, I think `this.serverMiddlewarePaths` can be assigned to variable at beginning of `watchRestart`, then it will be shared by watcher and also re-calculated after every building

Atinux: Good catch, fixed!

pi0 (approved):

PR #5682, opened 2019-05-09
fix(server): Setup Nuxt middleware only in dev or start

atinux:
## Types of changes

## Description

This should fixes issues like https://github.com/nuxt/nuxt.js/issues/5669
It should as well avoid useless operations when not needed.

**Warning:** The breaking change will be related to programmatically usage (since they don't provide `_start`), so it may not the best solution. But what I see from https://nuxtjs.org/api/nuxt, it seems it's not up to date.

## Checklist:

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -59,8 +59,10 @@ export default class Server {
     this.renderer = new VueRenderer(this.serverContext)
     await this.renderer.ready()
 
-    // Setup nuxt middleware
-    await this.setupMiddleware()
+    // Setup nuxt middleware (only with nuxt dev && nuxt start)
+    if (this.options.dev || this.options._start) {
+      await this.setupMiddleware()
+    }
 
     // Call done hook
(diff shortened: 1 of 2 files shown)


clarkdo:
If this is not an urgent pr to be released, I think we should consider more refactoring on this.

Actually in programmatically usage, we should not create any server related stuff (all middlewares, listener, socket) except a renderer or single nuxt middleware for rendering

So it would be great that we separate and provide a pure renderer/middleware to programmatical user.

clarkdo (commented):
If this is not an urgent pr to be released, I think we should consider more refactoring on this.

Actually in programmatically usage, we should not create any server related stuff (all middlewares, listener, socket) except a renderer or single nuxt middleware for rendering

So it would be great that we separate and provide a pure renderer/middleware to programmatical user.

Atinux:
Agree, we will work on this for Nuxt 3 instead. Closing it now.

PR #5695, opened 2019-05-10
fix(builder): corretly detect mode of hashed plugins

### LINKED ISSUE #5694: Plugin Filename drives mode via addPlugin

With the latest version of nuxt, plugins go to the right build via their filename when they are referenced in nuxt.config.js
  
plugin.server.js > goes to server bundle
plugin.client.js > goes to client bundle
plugin.js > goes to both

This functionality doesn't seem to work when adding plugins via modules using .addPlugin

This forces people to do .addPlugin( src:'...', mode:'something') when I'd expect the 'default' mode to be based off the source filename

https://nuxtjs.org/api/internals-module-container#addplugin-template-

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9184">#c9184</a>)</em></sub></div>

pi0:
## Types of changes

## Description

`moduleContainer.addPlugin` generates a hash suffixed name if no explicit `fileName` is provided to prevent name collisions inside `.nuxt`. This PR ensures mode of file names like `mode-test.plugin.client.530b6c6a.js` is also correctly detected.

fixes #5694

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -256,7 +256,7 @@ export default class Builder {
 
   normalizePlugins() {
     const modes = ['client', 'server']
-    const modePattern = new RegExp(`\\.(${modes.join('|')})(\\.\\w+)?$`)
+    const modePattern = new RegExp(`\\.(${modes.join('|')})(\\.\\w+)*$`)
     return uniqBy(
       this.options.plugins.map((p) => {
         if (typeof p === 'string') {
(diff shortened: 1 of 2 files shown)


clarkdo (commented):
Any file’s name includes client or server will be considered as mode plugin, like foo.client.bar.hash.js, is this behavior appropriate?

How about limting 0-2 times on \.\w or explicitly setting mode in addPlugin before hash generation

pi0:
@clarkdo As the pattern is like `.(client|server|all).` I think it is safe to just allow extending suffix. Resolving mode in `addPlugin` introduces duplicate code. Any concerns for not allowing more than 2 suffixes?

clarkdo:
Maybe an inappropriate example, foo.client.server.js

clarkdo:
In normal usage, it’s fine, so feel free to merge.

pi0:
The interesting fact is that with current replace strategy both previous and current result in something wrong :D

image

PR #5729, opened 2019-05-14
fix(webpack): resolve consola to latest

clarkdo:
Fix #5721 
Fix nuxt-community/auth-module#348

PR #5733, opened 2019-05-14
v2.7.1

pi0:
## Fixes

- **builder**: use warn only for mismatched dependencies (#5723)
- **webpack**: correctly resolve consola for the client bundle (#5729)

## Dependency Upgrades

- **postcss-import-resolver:** prefer `style` property when resolving (jsless/postcss-import-resolver#2)

## Thanks to

- Blake Kostner (@btkostner)

### DIFF
--- a/distributions/nuxt-start/package.json
+++ b/distributions/nuxt-start/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nuxt-start",
-  "version": "2.7.0",
+  "version": "2.7.1",
   "description": "Starts Nuxt.js Application in production mode",
   "keywords": [
     "nuxt",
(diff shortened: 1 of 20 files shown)


clarkdo:
The postcss-import-resolver change will align for [postcss-import behaviour
](https://github.com/postcss/postcss-import/blob/master/lib/resolve-id.js#L23), so it should be OK to include in release.

pi0:
@clarkdo  A fresh install will get this update anyway and seems not a breaking change. We may do a patch if got breaking reports.

atinux (approved):

aldarund (approved):

kevinmarrec (approved):

PR #5745, opened 2019-05-15
feat(vue-renderer): `render.injectScripts` option

### LINKED ISSUE #5664: inject: false in bundleRenderer config not work
### Version

[v2.6.3](https://github.com/nuxt.js/releases/tag/v2.6.3)

### Reproduction link

[https://github.com/maunier/nuxt-learning](https://github.com/maunier/nuxt-learning)

### Steps to reproduce

1. npm run dev 
2. visit http://localhost:3000/

### What is expected ?

html shouldn't load any script because I have configed "inject: false" for bundleRenderer in nuxt.config.js

it loads "commons.app.js" "vendors.app.js" and so on.

### Additional comments?

i need a template that automaticly inject all the script at IE10+ otherwise  don't inject any script 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9162">#c9162</a>)</em></sub></div>

maunier:
## Types of changes

## Description

Resolves: #5664

## Checklist:


pi0 (dismissed):

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -122,7 +122,9 @@ export default class SSRRenderer extends BaseRenderer {
     }
 
     // Prepend scripts
-    APP += this.renderScripts(renderContext)
+    if (this.options.render.bundleRenderer.inject) {
```
manniL: Shouldn't we provide a default value for `inject` (`true`) to avoid a breaking change? :thinking:

TheAlexLichter (dismissed):

packages/vue-renderer/src/renderers/ssr.js
pi0: @manniL Nice catch. Yes should be added here:

https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/config/render.js#L4

clarkdo: I think it should be `this.options.render.bundleRenderer.inject !== false`

manniL: Even better :+1:

maunier:
sorry, it's  my first time to pull request. so what should I do now？

packages/vue-renderer/src/renderers/ssr.js
maunier: sorry, it's my first time to pull request. so what should I do now？

maunier: I have a new commit and I have pushed to the 'dev' branch.

pi0:
This PR is not making the same effect of `bundleRenderer.inject` and just disabling javascript injection. Maybe move it to `options.render.injestScripts`?

maunier:
PTAL

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -82,14 +82,16 @@ export default class SSRRenderer extends BaseRenderer {
       m.style.text() +
       m.script.text() +
       m.noscript.text()
+    const needInjectScripts = this.options.render.injectScripts !== false
+    const needInjectState = this.options.render.injectState !== false
```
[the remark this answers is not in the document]
maunier: In my requirements, I only need the data fetched at the server side and no scripts related to the vue. And the two options correspond to the two functions in custom template —— renderState and renderScripts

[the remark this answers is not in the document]
maunier: > I think no extra option would be needed. Both are scripts and not usable standalone

But in my requirements, I only need the data fetched at the server side and no scripts related to the vue. And the two options correspond to the two functions in custom template —— renderState and renderScripts

pi0 (dismissed):

packages/vue-renderer/src/renderers/ssr.js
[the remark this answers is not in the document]
clarkdo: Agree with @pi0 , we'd better keep same API as vue-ssr-renderer which only covers `inject`.

For `needInjectState`, actually you can do some process in your store dir or asyncData, exclude data fetching in lower version browsers, this would be more efficient since it won't trigger data fetching request at all.

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -83,14 +83,18 @@ export default class SSRRenderer extends BaseRenderer {
       m.style.text() +
       m.script.text() +
       m.noscript.text()
+    
+    // Check if we need to inject scripts and state
+    const needInjectScripts = this.options.render.injectScripts !== false
+    const needInjectState = this.options.render.injectState !== false
```
Atinux:
```suggestion
    const shouldInjectState = this.options.render.injectState !== false
```

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -83,14 +83,18 @@ export default class SSRRenderer extends BaseRenderer {
...
 
     // Add <base href=""> meta if router base specified
     if (this.options._routerBaseSpecified) {
       HEAD += `<base href="${this.options.router.base}">`
     }
 
     // Inject resource hints
-    if (this.options.render.resourceHints) {
+    if (this.options.render.resourceHints && needInjectScripts) {
```
Atinux:
```suggestion
    if (this.options.render.resourceHints && shouldInjectScripts) {
```

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -123,7 +129,9 @@ export default class SSRRenderer extends BaseRenderer {
     }
 
     // Prepend scripts
-    APP += this.renderScripts(renderContext)
+    if (needInjectScripts) {
```
Atinux:
```suggestion
    if (shouldInjectScripts) {
```

atinux (changes requested):
Small changes required but overall good PR 👍

pi0 (approved):

atinux (approved):

PR #5748, opened 2019-05-15
refactor: small readability improvements

TheAlexLichter:
# Changes

* Refactor `!!sth` to `Boolean(sth)`
* Refactor ` { abc: abc }` to `{ abc }`
* Refactor ` const abc = someArray[0]` to ` const [abc] = someArray`

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -128,7 +128,7 @@ function mapTransitions(Components, to, from) {
 
 async function loadAsyncComponents(to, from, next) {
   // Check if route path changed (this._pathChanged), only if the page is not an error (for validate())
-  this._pathChanged = !!app.nuxt.err || from.path !== to.path
+  this._pathChanged = Boolean(app.nuxt.err) || from.path !== to.path
   this._queryChanged = JSON.stringify(to.query) !== JSON.stringify(from.query)
   this._diffQuery = (this._queryChanged ? getQueryDiff(to.query, from.query) : [])
 
@@ -267,7 +267,7 @@ async function render(to, from, next) {
     next: _next.bind(this)
   })
   this._dateLastError = app.nuxt.dateErr
-  this._hadError = !!app.nuxt.err
+  this._hadError = Boolean(app.nuxt.err)
 
   // Get route's matched components
   const matches = []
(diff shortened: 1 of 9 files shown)


packages/vue-app/template/client.js
```diff
@@ -128,7 +128,7 @@ function mapTransitions(Components, to, from) {
 
 async function loadAsyncComponents(to, from, next) {
   // Check if route path changed (this._pathChanged), only if the page is not an error (for validate())
-  this._pathChanged = !!app.nuxt.err || from.path !== to.path
+  this._pathChanged = Boolean(app.nuxt.err) || from.path !== to.path
```
pi0: neat: If we swipe conditions, we can prevent unnecessary Boolean cast where  `from.path !== to.path`

```suggestion
  this._pathChanged = from.path !== to.path || Boolean(app.nuxt.err)
```

manniL: Do you mean this? :relaxed:
 
```suggestion
  this._pathChanged = from.path !== to.path || app.nuxt.err
```

pi0: Where actually it depends. `|| app.nuxt.err` can cause `_pathChanged` to be an error instead of instance of `err`. This is why we use `!!` or `Boolean`. If such type safety is not important we can omit `Boolean` everywhere :)

clarkdo (commented):
@manniL Good changes.

For `Boolean`, I agree with using it in property assignment `this.foo = Boolean(bar) || ...`.

For local variable or if statement, `!!` may have some benefits.

```js
if( !!this.app.err && a !== 'abc'){
    console.log(1)
}

// after terser compression, !! will be removed, Boolean(..) won't

this.app.err&&"abc"!==a&&console.log(1)
```

The `!!` is removed, the code size is smaller and there is no reference leak risk in these cases, so it should be fine.

manniL:
@clarkdo Didn't know about the difference between `!!` and `Boolean` for terser minification/compression :astonished: 

The question is if it's worth the few bytes, but the trade-off is always present :see_no_evil:

pi0:
Terser is only effective on `vue-app` files. BTW agree to keep consistency and byte-saving where we can. The benefit of `Boolean` is type safety and readability.

manniL:
I'd keep that PR as is and create new ones for larger changes :relaxed:

manniL:
@pi0 IMO no, I also use `const { color } = myFavorites` instead of `const color = myFavorites.color` when it comes to objects :D 

But if you disagree with that I'm happy to revert :relaxed:

pi0:
@manniL No both array and obj destruction are nice things. I just meant about single value arrays. We can do it for consistency or keep `[0]` for more readability. I'm OK with both.

pi0 (approved):

clarkdo (approved):

kevinmarrec (approved):
Good changes @manniL !

Regarding single value arrays access, I've exact same POV as @pi0 and I'm also OK with both :+1:

PR #5753, opened 2019-05-16
fix(builder): watch for `pages/` creation when default page displayed

atinux:
## Types of changes

## Description

Same as `store/` but applied for `pages/` creation.

## Checklist:


packages/builder/src/builder.js
```diff
@@ -659,6 +659,10 @@ export default class Builder {
     if (this.ignore.ignoreFile) {
       nuxtRestartWatch.push(this.ignore.ignoreFile)
     }
+    // If default page displayed, watch for first page creation
+    if (this._nuxtPages && this._defaultPage) {
```
clarkdo: Is `_defaultPage` is enough here ?

If `_defaultPage` is true, `_nuxtPages` must be `true`.

packages/cli/src/commands/dev.js
```diff
@@ -48,7 +49,11 @@ export default {
     await nuxt.ready()
 
     // Start listening
-    await nuxt.server.listen()
+    await nuxt.server.listen(this._usedPort)
```
clarkdo: What if `server.port` in `nuxt.config` is changed at runtime ?

pi0: The `server.port` may be in use when we used the random port. checking for that value means we again fall back trying a used port and then choosing another random one. 

Alternatively, we can preserve a random chosen PORT in a global variable inside `listener` module and reuse it.

(PS: I think this fix can be moved to another PR for better history)

Atinux: Will create another PR for this.

packages/builder/src/builder.js
Atinux: Yes but readability is better :)

pi0 (approved):

packages/cli/src/commands/dev.js
Atinux: Let's move to #5793 5793

PR #5757, opened 2019-05-17
fix(vue-app): reuse page component on watchQuery

### LINKED ISSUE #5738: watchQuery on page re-creates page component
### Version

[v2.7.0](https://github.com/nuxt.js/releases/tag/v2.7.0)

### Reproduction link

[https://codesandbox.io/s/pj3jj4884j](https://codesandbox.io/s/pj3jj4884j)

### Steps to reproduce

1. Press the "Bump foo" button
2. Note that this button sets the browser's location bar to the current page with an updated query string

### What is expected ?

Setting/changing the `foo` query string parameter does not recreate the page component. Instead, the page component is re-used

The page component is recreated each time the `foo` query string parameter is changed (as evidenced by the console logging 'created' on each query string parameter changed).

This has performance/usability implications for more complex pages

### Additional comments?

Repro link is for v2.7.x.
Here is a link showing the old, preferred behaviour (v2.6.x): https://codesandbox.io/s/mzl7n86jv9

I believe this to be a breaking change, even if this new behaviour is intended. And if it is intended, what is a work around to achieve the old behaviour (as I'm relying on the page component to be reused)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9223">#c9223</a>)</em></sub></div>

hatashiro:
## Types of changes

## Description

Related issue and pull request:
- Resolves: #5738  #5759 
- Related: #5516 

I've submitted the pull request #5516 and it's about making `watchQuery` considered on creating a route key. However, by the change, page components are no more reused but recreated for every query route specified in `watchQuery` (#5738). The pull request was actually a breaking change.

The main goal of the original pull request was to make scroll restoration work with query route. Disabling component reuse is unexpected side-effect. Thus I'd like to revert the original change, with still enabling scroll restoration in a different way. So, here is this pull request.

This pull request restores the original `routerViewKey` behaviour (`$route.path`) so that page components are reused. Instead, I added `$emit('triggerScroll')` in `fixPrepatch`, which is called only when `navigating on a different route but the same component is used` according to the function's comment. I tested the change out on example pages and it worked well. [The original test cases on scroll behaviour](https://github.com/nuxt/nuxt.js/blob/dev/test/e2e/basic.browser.test.js#L144-L181) pass too.

Because of this change the documentation about `routerViewKey` should be updated (`watchQuery` no more affects the key). I'll update it and upload a pull request on docs after this pull request is reviewed.

Thanks in advance!

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1395)
  + There is already tests for the behaviour: https://github.com/nuxt/nuxt.js/blob/dev/test/e2e/basic.browser.test.js#L144-L181

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -506,6 +506,11 @@ function fixPrepatch(to, ___) {
         for (const key in newData) {
           Vue.set(instance.$data, key, newData[key])
         }
+
+        // Ensure to trigger scroll event after calling scrollBehavior
+        window.<%= globals.nuxt %>.$nextTick(() => {
+          window.<%= globals.nuxt %>.$emit('triggerScroll')
+        })
       }
     })
     showNextPage.call(this, to)
--- a/packages/vue-app/template/components/nuxt.js
+++ b/packages/vue-app/template/components/nuxt.js
@@ -43,11 +43,12 @@ export default {
 
       const [matchedRoute] = this.$route.matched
       const Component = matchedRoute && matchedRoute.components.default
-      if (Component && Component.options) {
-        const { key } = Component.options
 
-        if (key) {
-          return (typeof key === 'function' ? key(this.$route) : key)
+      if (Component && Component.options) {
+        const { options } = Component
+
+        if (options.key) {
+          return (typeof options.key === 'function' ? options.key(this.$route) : options.key)
         }
       }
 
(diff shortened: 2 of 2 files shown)


clarkdo (dismissed):

clarkdo:
Documentation also needs to be reverted 😄

utatti:
I've submitted a pull request on docs too.

clarkdo (approved):

PR #5784, opened 2019-05-22
feat(vue-renderer): support `render.ssrLog` for controlling SSR logs

### LINKED ISSUE #5728: Add `forwardLog` option to NuxtConfiguration

in [Nuxt v2.7.0](https://github.com/nuxt/nuxt.js/releases/tag/v2.7.0) or later, `nuxt dev` forwards SSR's console.log to the front end.
However, if you have a large number of logs, you do not necessarily want to forward them to the front end.

To that end, it would be nice to be able to turn it off from nuxt.config.js.

1. Add a flag (Boolan) whether to transfer to NuxtConfiguration
2. Make a change to make it work properly (I would like to send a Pull Request for it)

thanks.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9214">#c9214</a>)</em></sub></div>

clarkdo:
## Types of changes

## Description

Resolve #5728 and issue @pimlie mentioned in Discord.

Introduce `render.ssrLog` for configuring SSR logs:

- `true` enable SSR log
image
- `'collapsed'` enable SSR log, but collapse log group.
image
- `false` disable SSR log

Default Value: `true` (**NOTE**: `ssrLog` only works in `dev` mode)

## Checklist:


packages/vue-app/template/client.js
```diff
@@ -37,10 +37,10 @@ const NUXT = window.<%= globals.context %> || {}
 
 Object.assign(Vue.config, <%= serialize(vue.config) %>)<%= isTest ? '// eslint-disable-line' : '' %>
 
-<% if (isDev) { %>
+<% if (options.render.ssrLog) { %>
```
pi0: Displaying SSR logs for production is dangerous anyway. Can lead to potential data leak.

pi0: PS: With request concurrency in prod logs can be shared between users :)

clarkdo: It's disabled by default

pi0: I see. But for the technical reason above, it is non-sense at all to have this functionality for production

pimlie:
Really like being able to show the SSR logs collapsed by default :rocket:

Wdyt about logging a message like `Nuxt.js SSR logs are disabled in the config` with `ssrLog: false`? It might help to remind people the feature exists and they will also know why no SSR logs are showing (eg that its not a bug)

clarkdo:
> Really like being able to show the SSR logs collapsed by default 🚀
> 
> Wdyt about logging a message like `Nuxt.js SSR logs are disabled in the config` with `ssrLog: false`? It might help to remind people the feature exists and they will also know why no SSR logs are showing (eg that its not a bug)

Since `SSR Log` can be disabled only by specifying `ssrLogs: false` manually, do you think it's necessary to remind users ?

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -39,7 +39,10 @@ export default class SSRRenderer extends BaseRenderer {
     )
   }
 
-  async devRenderToString(renderContext) {
+  creatSsrLog() {
+    if (!this.options.render.ssrLog) {
+      return { finish() {} }
```
pi0: What if we just `return` here and later check for logger existence to finish it? This can prevent creating an extra closure for each SSR request in prod..

pi0: Anyway, just a side suggestion. Probably this overhead is almost 0 ;)

clarkdo: Make sense, what about now ?

pi0: Seems good :)

pimlie:
> do you think it's necessary to remind users ?

Its not necessary in this case I guess, but in general reminders like that could be helpful I think when you eg start working on an old project again and you either forgot you set that option or even that the option exists at all because you got so used at it. But I dont mind either way, so feel free to do as you see fit ;)

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -39,7 +39,10 @@ export default class SSRRenderer extends BaseRenderer {
     )
   }
 
-  async devRenderToString(renderContext) {
+  useSsrLog() {
```
pi0: neat: isn't better to name it `useSSRLog` ?

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -50,19 +53,25 @@ export default class SSRRenderer extends BaseRenderer {
...
+      consola.removeReporter(devReporter)
+      return logs
+    }
   }
 
   async render(renderContext) {
     // Call ssr:context hook to extend context from modules
     await this.serverContext.nuxt.callHook('vue-renderer:ssr:prepareContext', renderContext)
 
+    const getSsrLog = this.useSsrLog()
```
pi0: neat: what if we name it `getSSRLogs`?

pi0 (dismissed):

pi0 (dismissed):

packages/vue-app/template/client.js
manniL:
```suggestion
<% if (options.render.ssrLog && isDev) { %>
```

Agree with @pi0 here.

packages/config/types/render.d.ts
```diff
@@ -23,5 +23,6 @@ export interface NuxtConfigurationRender {
   http2?: any // TBD
   resourceHints?: boolean
   ssr?: boolean
+  ssrLog?: true | false | 'collapsed'
```
kevinmarrec:
```suggestion
  ssrLog?: boolean | 'collapsed'
```

Dunno if this suggestion is enough relevant, in any case it's typed correctly 😁

kevinmarrec (dismissed):

kevinmarrec (dismissed):

galvez (dismissed):

atinux (dismissed):

pi0 (dismissed):

pi0 (approved):

PR #5785, opened 2019-05-22
fix(ts): add type definition for functional babel.presets

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/config/types/build.d.ts
+++ b/packages/config/types/build.d.ts
@@ -23,7 +23,7 @@ interface NuxtBabelPresetEnv {
 }
 
 interface NuxtBabelOptions extends Pick<TransformOptions, Exclude<keyof TransformOptions, 'presets'>> {
-  presets?: ((env: NuxtBabelPresetEnv, defaultPreset: [string, object | null]) => PluginItem[] | void) | PluginItem[] | null
+  presets?: ((env: NuxtBabelPresetEnv, defaultPreset: [string, object]) => PluginItem[] | void) | PluginItem[] | null
 }
 
 export interface NuxtConfigurationBuild {
--- a/packages/config/types/build.d.ts
+++ b/packages/config/types/build.d.ts
@@ -8,7 +8,7 @@ import {
   Options as WebpackOptions,
   Plugin as WebpackPlugin
 } from 'webpack'
-import { TransformOptions } from '@babel/core'
+import { TransformOptions, PluginItem } from '@babel/core'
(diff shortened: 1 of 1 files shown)


packages/config/types/build.d.ts
```diff
@@ -18,9 +18,17 @@ import { TerserPluginOptions } from 'terser-webpack-plugin'
 
 type NuxtConfigurationLoaders = any // TBD
 
+interface NuxtBabelPresetEnv {
+  isServer: boolean;
+}
+
+interface NuxtBabelOptions extends Pick<TransformOptions, Exclude<keyof TransformOptions, 'presets'>> {
+  presets?: ((env: NuxtBabelPresetEnv, presets: [string, object | null]) => PluginItem[] | void) | PluginItem[] | null
```
kevinmarrec: Is there really a case that we accept functional `presets` returning `void` ? Should it not be always an array ?

Assuming https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/config/base.js#L81 & https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/config/base.js#L89, 
`presets` function parameter should be named `defaultPreset` instead and seems to fit `PluginItem` type.

```suggestion
  presets?: ((env: NuxtBabelPresetEnv, defaultPreset: PluginItem) => PluginItem[]) | PluginItem[] | null
```

kevinmarrec (dismissed):

packages/config/types/build.d.ts
clarkdo: `presets` can modify the second argument and return nothing.

`presets: [string, object | null]` is fixed strcture, so it shouldn't 100% follow the structure of PluginItem

kevinmarrec: Alright all good !

clarkdo: Arg name is changed, thanks

packages/config/types/build.d.ts
```diff
@@ -18,9 +18,17 @@ import { TerserPluginOptions } from 'terser-webpack-plugin'
 
 type NuxtConfigurationLoaders = any // TBD
 
+interface NuxtBabelPresetEnv {
+  isServer: boolean;
```
kevinmarrec:
```suggestion
  isServer: boolean
```

kevinmarrec (dismissed):

kevinmarrec (approved):

PR #1136, opened 2017-07-21
Resolve non-js package importing

shirohana:
Belongs to issue https://github.com/nuxt/nuxt.js/issues/375#issuecomment-317099812

PR #5796, opened 2019-05-23
refactor(core): use `require.resolve` instead of `Module` internals

Globegitter:
## Types of changes

## Description
Is there any reason to not use require.resolve but rather a private module function? We use [Bazel](https://bazel.build/) as our build tool to call nuxt and it implements a custom module resolution. Using this private method currently does not work, but using `require.resolve` does. Would be great to get this in if there is no specific reason to use this private method.

## Checklist:

### DIFF
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -1,4 +1,3 @@
-import Module from 'module'
 import { resolve, join } from 'path'
 import fs from 'fs-extra'
 import consola from 'consola'
@@ -23,7 +22,7 @@ export default class Resolver {
 
   resolveModule(path) {
     try {
-      return Module._resolveFilename(path, {
+      return require.resolve(path, {
         paths: this.options.modulesDir
       })
     } catch (error) {
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -28,13 +28,10 @@ export default class Resolver {
         paths: this.options.modulesDir
       })
     } catch (error) {
-      if (error.code === 'MODULE_NOT_FOUND') {
-        return undefined
-      } else {
+      if (error.code !== 'MODULE_NOT_FOUND') {
         // TODO: remove after https://github.com/facebook/jest/pull/8487 released
-        if (process.env.NODE_ENV === 'test' &&
-          error.message.startsWith('Cannot resolve module')) {
-          return undefined
+        if (process.env.NODE_ENV === 'test' && error.message.startsWith('Cannot resolve module')) {
+          return
         }
         throw error
       }
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -6,7 +6,7 @@ import esm from 'esm'
 import { startsWithRootAlias, startsWithSrcAlias } from '@nuxt/utils'
 
 export default class Resolver {
-  constructor(nuxt) {
+  constructor(nuxt, moduleResolver) {
     this.nuxt = nuxt
     this.options = this.nuxt.options
 
@@ -18,11 +18,14 @@ export default class Resolver {
 
     // ESM Loader
     this.esm = esm(module)
+
+    // module resolver
+    this.moduleResolver = moduleResolver || require.resolve
   }
 
   resolveModule(path) {
     try {
-      return require.resolve(path, {
+      return this.moduleResolver(path, {
         paths: this.options.modulesDir
       })
     } catch (error) {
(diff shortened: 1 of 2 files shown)


pi0 (dismissed):

pi0:
Historically coming from #1136. It seems `require.resolve` also accepts second `paths` param so looks a valid fix.

Globegitter:
@pi0 I did some local testing and it seems the returned errors are different from `require.resolve`, so this should hopefully fix it.

clarkdo:
@pi0 It looks like that `paths` in `_resolveFilename` and `resolve` are different.

`paths` in `_resolveFilename` is parent, in `resolve` is module location.

https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/helpers.js#L21

Globegitter:
Good point @clarkdo, but I guess the main question is, does it still achieve the same thing, or is it maybe even the more correct thing to do?

clarkdo:
@Globegitter You're right, `options.paths` in `resolve` will be converted to fake parent in `_resolveFilename`, but I'm not sure why the error thrown is not `MODULE_NOT_FOUND`, I need more investigation for avoiding unexpected behaviour.

Globegitter:
@clarkdo Yeah that is strange indeed - but at least all the tests are passing now, so that is something. In my local testing `error.code` was not set in most of the cases, not sure why that is the case. Anyway, I can understand that you need more time to investiage.

packages/core/src/resolver.js
```diff
@@ -19,15 +18,18 @@ export default class Resolver {
 
     // ESM Loader
     this.esm = esm(module)
+
+    // module resolver
+    this.moduleResolver = moduleResolver || require.resolve
```
clarkdo: I think we may not need `moduleResolver` for consistency if `require.resolve` works fine 😄

clarkdo:
@Globegitter Error is from https://github.com/facebook/jest/blob/master/packages/jest-runtime/src/index.ts#L650-L654 which doesn't set code, so it only happens in jest.

packages/core/src/resolver.js
```diff
@@ -19,15 +18,18 @@ export default class Resolver {
...
 
   resolveModule(path) {
     try {
-      return Module._resolveFilename(path, {
+      return this.moduleResolver(path, {
         paths: this.options.modulesDir
       })
     } catch (error) {
-      if (error.code === 'MODULE_NOT_FOUND') {
+      if (error.code === 'MODULE_NOT_FOUND' || error.message.includes('Cannot resolve module')) {
```
clarkdo: @Globegitter Could you please add a link here https://github.com/nuxt/nuxt.js/pull/5796

@pi0 What do you think adding `process.env.NODE_ENV === 'test'` ?

pi0: Probably we can open a repo in jest for adding correct code and reference it here.

pi0: https://github.com/facebook/jest/pull/8487 :+1:  :D

clarkdo: @pi0 See my comment, I've opened pr in jest, but I suppose that they may not approve and release in short term.

packages/core/src/resolver.js
Globegitter: Not sure I follow - I needed to do this for testing to be able to mock `require.resolve`: https://stackoverflow.com/questions/47907454/mocking-and-testing-require-resolve-calls-in-jest

But maybe there is abetter way?

clarkdo: You can do it by sth like:

```js
jest.spyOn(require, 'resolve').mockImplementation(() => 'test')
console.log(require.resolve());
require.resolve.mockRestore()

File: packages/core/src/resolver.js [hunk 3]
```diff
@@ -19,19 +19,23 @@ export default class Resolver {
...
       })
     } catch (error) {
-      if (error.code === 'MODULE_NOT_FOUND' || error.message.includes('Cannot resolve module')) {
+      if (error.code === 'MODULE_NOT_FOUND') {
         return undefined
       } else {
+        // TODO: remove after https://github.com/facebook/jest/pull/8487 released
+        if (process.env.NODE_ENV === 'test' &&
+          error.message.startsWith('Cannot resolve module')) {
+          return undefined
```
pi0: `return` is same as `return undefined` (also can we please put it in a single line? :D)

pi0 (approved):

PR #5812, opened 2019-05-26
fix(builder): only listen for file changes for supported extensions

pimlie:
## Types of changes

## Description

- the patterns used for listening to file changes contained some duplicates, removed those
- the workaround for recreating the watcher on raw events on linux is not needed anymore due to fixed in chokidar (left the code in for now, it just isnt used by default anymore as the config is empty)
- made all patterns less greedy by always only listening for supported file extensions

Eg listening on a directory without the supported file extensions could be prone to unnecessary reloads due to file changes which are not used by Nuxt. Eg `*.swp`, `*~` or `*.bak` files used by editors

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -209,6 +209,10 @@ export default class Builder {
     }
   }
 
+  globPathWithExtensions(path) {
+    return `${path}/**/*.{${this.supportedExtensions.join(',')}}`
+  }
+
   async generateRoutesAndFiles() {
     consola.debug('Generating nuxt files')
 
@@ -229,7 +233,7 @@ export default class Builder {
     await this.resolveLoadingIndicator(templateContext)
 
     // Add vue-app template dir to watchers
-    this.options.build.watch.push(this.template.dir)
+    this.options.build.watch.push(this.globPathWithExtensions(this.template.dir))
 
     await this.compileTemplates(templateContext)
 
@@ -274,7 +278,7 @@ export default class Builder {
   }
 
   async resolveFiles(dir, cwd = this.options.srcDir) {
-    return this.ignore.filter(await glob(`${dir}/**/*.{${this.supportedExtensions.join(',')}}`, {
+    return this.ignore.filter(await glob(this.globPathWithExtensions(dir), {
       cwd,
       ignore: this.options.ignore
     }))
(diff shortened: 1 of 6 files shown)


packages/builder/src/builder.js
```diff
@@ -209,6 +209,10 @@ export default class Builder {
     }
   }
 
+  globPathWithExtensions(path) {
+    return `${path}/**/*.{${this.supportedExtensions.join(',')}}`
```
pi0: Elegant refactor!

packages/builder/src/builder.js
```diff
@@ -568,6 +572,7 @@ export default class Builder {
       watcher.on(event, listener)
     }
 
+    // TODO: due to fixes in chokidar this isnt used anymore and could be removed in Nuxt v3
```
pi0: As this was internal/undocumented is it necessary to keep it?

packages/builder/src/builder.js
```diff
@@ -592,26 +597,20 @@ export default class Builder {
...
     if (this._nuxtPages && !this._defaultPage) {
-      patterns.push(
-        r(src, this.options.dir.pages),
-        ...rGlob(this.options.dir.pages)
-      )
+      patterns.push(r(this.options.srcDir, this.options.dir.pages))
     }
 
-    patterns = patterns.map(upath.normalizeSafe)
+    patterns = patterns.map((path, ...args) => upath.normalizeSafe(this.globPathWithExtensions(path), ...args))
```
pi0: As we are mapping to glob does it still covers directory creation? (`store`, `pages`)

packages/builder/src/builder.js
pimlie: I would leave it in for now until we are sure the chokidar fix works for everybody (and not just me). Ofc we could always do a hotfix in that case, but having a config backup option is nice too I guess. I will keep track of this TODO and submit a PR to remove this code in a next release if nobody complains :)

pi0: Looks like a good plan :) (However we also have the possibility to do a fast patch release in case of complain but makes sense keeping it for a while)

packages/builder/src/builder.js
pimlie: Yes, if they dont exists yet they are added in `watchRestart`: https://github.com/nuxt/nuxt.js/blob/4cefc7bb893c74829a466c9f0ea9edcf61afec8e/packages/builder/src/builder.js#L660-L666

pi0 (approved):
Locally tested. Works fine!

kevinmarrec (approved):
Nice ! 

*Regarding Builder `supportedExtensions` and incoming TypeScript refactors, I planned to make the array editable/extendable from `nuxt.config` options. So that we can remove `ts` and `tsx` extensions from core and only extend with these 2 TS extensions through new refactored `typescript` module*.

PR #5824, opened 2019-05-28
fix(vue-app): fixPrepatch in-out transition fix (#5797)

### LINKED ISSUE #5797: data function is being called after mounted when using transition in-out mode
### Version

[2.7.1](https://github.com/nuxt.js/releases/tag/2.7.1)

### Reproduction link

[https://codesandbox.io/embed/codesandboxnuxt-i8l14](https://codesandbox.io/embed/codesandboxnuxt-i8l14)

### Steps to reproduce

set nuxt-child page transition mode to "in-out" and set transition css and set something new value in mounted lifecycle

please check Reproduction link

### What is expected ?

call data function once

data function is being called after mounted and  $data is being overwritten

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9259">#c9259</a>)</em></sub></div>

Elevista:
## Types of changes

## Description

Resolves: #5797 

Problem: fixPrepatch + $data updates after component mounted twice. (in-out transition mode)

Please review and check if has side effect

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -488,14 +488,14 @@ function showNextPage(to) {
 // Will not update the instance data, so we have to update $data ourselves
 function fixPrepatch(to, ___) {
   if (this._pathChanged === false && this._queryChanged === false) return
+  
+  const matches = []
+  const instances = getMatchedComponentsInstances(to, matches)
+  const Components = getMatchedComponents(to, matches)
 
   Vue.nextTick(() => {
-    const matches = []
-    const instances = getMatchedComponentsInstances(to, matches)
-    const Components = getMatchedComponents(to, matches)
-
     instances.forEach((instance, i) => {
-      if (!instance) return
+      if (!instance || instance._isDestroyed) return
       // if (
       //   !this._queryChanged &&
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/client.js
```diff
@@ -488,14 +488,14 @@ function showNextPage(to) {
 // Will not update the instance data, so we have to update $data ourselves
 function fixPrepatch(to, ___) {
   if (this._pathChanged === false && this._queryChanged === false) return
+  
+  const matches = []
+  const instances = getMatchedComponentsInstances(to, matches)
```
Atinux: Why moving this outside of `Vue.nextTick`?

Elevista: Unlike transition out-in mode, in in-out mode,
the "transition In" component instance has already been created in this nextTick,
so "transition In" component instance is included in array returned from the getMatchedComponentsInstances function.

Assuming the purpose of this function,
it is assumed that the newly created component should not be listed in this array.
So to get the same result as out-in mode, the code should be outside the nextTick.

Atinux: Alright sounds good 😄

PR #5825, opened 2019-05-28
refactor(webpack): bring mini-css-extract-plugin back

clarkdo:
## Types of changes

## Description

Revert to `mini-css-extract-plugin` since it has supported HMR 

## Checklist:


pi0 (approved):

clarkdo:
Hi @ScriptedAlchemy , thanks for the wonderful HMR support pr in `mini-css-extract-plugin`

I found that you've made a lot great changes in both `mini-css-extract-plugin` and `extract-css-chunks-webpack-plugin`, so I think you're the right expert that I should ask.

What do you think that we revert to `mini-css-extract-plugin` ?

ScriptedAlchemy:
Damn. Tough one. But I appreciate you seeking my advise. Appreciate your interest in anything I have to say on the subject ❤️

It really depends on what you need out of the plugin. In general, extract-css-chunks is going to have newer features - at least from my contributions. I roll most work to my own plugin first, then roll it to webpack. 

I do keep them in sync most of the time, so I’ll take contributions from mini-css and other authors and roll them back and forward. So either way the same features of mini-css should be in my plugin  (extract-css-chunks) 

As I work on both and maintain them, I can say that mini-css has a way larger user base, it’s much harder to release code or add features at will. We have to be careful and there’s not a lot of contributors, so if we screw something up... it’s a huge wave of GitHub issues and not a lot of coders who can resolve them. 

Mini-css is generally more stable due to this. I tend to open many “v4.3.6-beta.0” beta style releases to npm for extract-css-chunks. My @ next version is usually more advanced because I have found that my users are pretty happy to upgrade and help me test new code. If you like updating to the latest often, mini-css might be a slightly safer choice.  However, I also have complete control over extract-css-chunks so I can add features or change the code to help a project. It would be harder to get core webpack projects to cater to the needs of a handful of projects. 


For example, right now HMR doesn’t work in mini-css if the css file is served off another port or is a url (regex issue) But I was able to easily fix that issue on extract-css and push it out :) the fix for mini-css which I myself will use as well, takes longer because we need to deal with more use cases. 

There’s also just the attitude to how the project is treated. Webpack has a very enterprise feel to it in many ways. Where as on my project, I have ultimate control and make changes that can be drastically but bring a major improvement. I’m one person and I’ll do as I wish with the project - many see that as a risk - but I respect my users hugely and often seek their input before making a decision. 

At the end of the day, you will likely get the same experience from either plugin. Mine will usually move faster or at least stay in sync with mini-css. But I’m also one person, so webpacks mini-css might be better and have more capacity when it comes to support. 

**what do I do** I use both projects, big complex ones get my plugin, smaller ones get mini-css. Again I do this because I can offer big projects better control.

ScriptedAlchemy:
There’s also the server-side rendering aspect. We are the original authors of code-split SSR. when new innovations happen or drastic technological changes were needed - we own the whole stack. We push to standardize technology so there’s not much ecosystem, but if webpack 5 caused major SSR challenges, it’s my group who’d be first responders and extract-css-chunks would be the first in line to become operational in SSR environments if it were to ever stop. 

There’s alternate SSR systems out there now, many based on our original work. We are friends with others in the industry so there’s no competitive aggression towards each other. 

For me, use whatever works best and you feels fit your system best :) be that extract-css or mini-css. 

Hopefully I’ve provided enough info

clarkdo:
@ScriptedAlchemy Thank you so so so much for your quick and detailed reply, it will definitely help use a lot.

If I understood correctly, `extract-css` normally have more brilliant features and `mini-css` has a larger user group and more stable functionality

@Atinux @pi0 What do you think about this? Is it necessary to provide the possibility to switch the plugin is used since they have almost exact same options ?

ScriptedAlchemy:
More than welcome! Appreciate your support and use of my work! 

More or less. In no official capacity of course. I just happen to maintain both plugins and release all my contributions to my users ahead of time. I sync changes from others on mini css so that my plugin generally carries all the fixes plus a little extra. 

Your understanding is correct.

Atinux:
Thank you so much for your input @ScriptedAlchemy and the take taken to guide us.

@clarkdo I believe we could update the [`build.extractCSS`](https://nuxtjs.org/api/configuration-build#extractcss) option to accept a string equals to `'mini-css-extract-plugin'` so our userbase can start testing it and give us feedback on it?

pi0:
As alternative we can create a community module that replaces loader. (If mini-css has advantages at a point of time)

jasonc08:
Is it able to fix this issue?
https://github.com/nuxt/nuxt.js/issues/4219

ScriptedAlchemy:
I know there was an ordering issue which was fixed @jasonc08 - I'm pretty sure i rolled the patch over, ill diff the projects again and re-check.

manniL:
@clarkdo 	I guess this PR can be closed in favor of #5906 ?

clarkdo:
This pr is not for fixing extract css bug, it’s for consideration of which lib is more proper, personally I prefer mini-css since it has larger user amount and stable functionality

PR #5826, opened 2019-05-28
refactor(vue-app): simplify mount error log 

TheAlexLichter:
In case the mounting of the app fails, the error should be logged to the console in dev mode so it's easier to trace

## Types of changes
- [x] DX Improvements

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -93,6 +93,7 @@ const errorHandler = Vue.config.errorHandler || console.error
 createApp()
   .then(mountApp)
   .catch((err) => {
+    <% if (isDev) { %>console.error(err)<% } %>
     const wrapperError = new Error(err)
     wrapperError.message = '[nuxt] Error while mounting app: ' + wrapperError.message
     errorHandler(wrapperError)
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -489,14 +489,14 @@ function showNextPage(to) {
 // Will not update the instance data, so we have to update $data ourselves
 function fixPrepatch(to, ___) {
   if (this._pathChanged === false && this._queryChanged === false) return
+  
+  const matches = []
+  const instances = getMatchedComponentsInstances(to, matches)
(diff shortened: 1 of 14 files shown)


pi0 (dismissed):

pi0:
@manniL For confirmation, dosen't it cause duplicate errors in console?

manniL:
@pi0 	Stacktrace in FF w/o

image

With the error logged
image


It's easier to trace then :relaxed:

atinux (dismissed):

clarkdo (commented):
I'm a little confused, because we already log error in default errorHandler, so is it only for less stacktrace? If so, I think we'd better optimize original log format. 

The only difference is a error `wrapper`, may be we can change the wrapper to prepended message.

pi0:
Maybe we can just pass the original error to errorHandler. `[error while mountin]` is not so useful for tracing.

packages/vue-app/template/client.js
```diff
@@ -93,6 +93,7 @@ const errorHandler = Vue.config.errorHandler || console.error
 createApp()
   .then(mountApp)
   .catch((err) => {
```
clarkdo: @pi0 Then `.catch(errorHandler)` is enough

pi0: Yes. Agree. Cleaner code :)

packages/vue-app/template/client.js
```diff
@@ -93,6 +93,7 @@ const errorHandler = Vue.config.errorHandler || console.error
...
+    <% if (isDev) { %>console.error(err)<% } %>
```
clarkdo:
```js
createApp().then(mountApp).catch(errorHandler)
```

packages/vue-app/template/client.js
clarkdo:
```js
createApp().then(mountApp).catch(errorHandler)
```

manniL: Sure, makes sense :)

PR #5841, opened 2019-05-31
fix(config): fix `extendRoutes` method type

antfu:
Add ts declaration missing `component` and `children` fields for Route in `route.extendRoutes` method.

## Types of changes

## Description
Fix typescript declaration

## Checklist:

### DIFF
--- a/packages/config/types/router.d.ts
+++ b/packages/config/types/router.d.ts
@@ -12,7 +12,7 @@ type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent;
 
 export interface NuxtRouteConfig extends RouteConfig {
   chunkNames: { [key: string]: string }
-  component: Component | string,
+  component: RouteConfig['component'] | string
   children?: NuxtRouteConfig[]
 }
 
--- a/packages/config/types/router.d.ts
+++ b/packages/config/types/router.d.ts
@@ -10,7 +10,7 @@ import { RouterOptions, RouteConfig } from 'vue-router'
 export interface NuxtRouteConfig extends RouteConfig {
   chunkNames: { [key: string]: string },
   component: string,
-  children?: NuxtRouteConfig[],
+  children?: NuxtRouteConfig[]
 }
 
 export interface NuxtConfigurationRouter extends RouterOptions {
--- a/packages/config/types/router.d.ts
+++ b/packages/config/types/router.d.ts
@@ -9,7 +9,7 @@ import { RouterOptions, RouteConfig } from 'vue-router'
 
 export interface NuxtRouteConfig extends RouteConfig {
   chunkNames: { [key: string]: string }
-  component: RouteConfig['component'] | string
+  component?: RouteConfig['component'] | string
   children?: NuxtRouteConfig[]
 }
 
--- a/packages/config/types/router.d.ts
+++ b/packages/config/types/router.d.ts
@@ -6,10 +6,13 @@
  */
 
 import { RouterOptions, RouteConfig } from 'vue-router'
+import Vue, { ComponentOptions, AsyncComponent } from 'vue'
+
+type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent;
 
 export interface NuxtRouteConfig extends RouteConfig {
   chunkNames: { [key: string]: string },
-  component: string,
+  component: Component | string,
   children?: NuxtRouteConfig[]
 }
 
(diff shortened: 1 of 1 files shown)


kevinmarrec (changes requested):
@antfu You can omit commas

And how did you find these types were missing ?
Is there in Nuxt docs maybe ?

antfu:
@kevinmarrec They are from
https://github.com/nuxt/nuxt.js/blob/111132df23d7fbaa115e694289ea639e7e1ae24f/packages/utils/src/route.js#L134-L172

Also there are some examples
https://github.com/nuxt/nuxt.js/blob/01acf66c675e0e6f08b9bb24a1cbe98c0939256e/test/fixtures/with-config/nuxt.config.js#L17-L35

https://github.com/nuxt/nuxt.js/blob/b1b9e0bcbc2eb328452c164af4777c860095a6d4/examples/named-views/nuxt.config.js#L3-L28

kevinmarrec:
@antfu `component` & `children` type definitions are defined in `RouteConfig` from `vue-router` https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L74, and as we're extending this interface, there is no need to define again the types.

antfu:
@kevinmarrec Sorry I missed that part. But there are still two problem:

image

1. `string` is not a valid type of `component`, which you can found in may example using string for the value
https://github.com/nuxt/nuxt.js/blob/01acf66c675e0e6f08b9bb24a1cbe98c0939256e/test/fixtures/with-config/nuxt.config.js#L23

2. Although the `children` is defined in `vue-router`, the type the child is the original `RouteConfig`, which the `chunkNames` field definition is missing


The code in screenshot:
```ts
    extendRoutes(routes, resolve) {
      const firstLevel = routes[0] // the type is NuxtRouteConfig
      firstLevel.component = '~/pages/about.vue' // string is not allow there
      firstLevel.chunkNames = { top: 'chunk_name' }

      if (firstLevel && firstLevel.children) {
        const secondLevel = firstLevel.children[0] // the type is RouteConfig
        secondLevel.component = resolve(__dirname, 'pages/about.vue') // resolve also returns string
        secondLevel.chunkNames = { top: 'chunk_name' }
      }

      return routes
    }
```

packages/config/types/router.d.ts
```diff
@@ -6,9 +6,14 @@
  */
 
 import { RouterOptions, RouteConfig } from 'vue-router'
+import Vue, { ComponentOptions, AsyncComponent } from 'vue'
+
+type Component = ComponentOptions<Vue> | typeof Vue | AsyncComponent;
 
 export interface NuxtRouteConfig extends RouteConfig {
-  chunkNames: { [key: string]: string }
+  chunkNames: { [key: string]: string },
```
kevinmarrec:
```suggestion
  chunkNames: { [key: string]: string }
```

packages/config/types/router.d.ts
```diff
@@ -6,9 +6,14 @@
...
+  component: Component | string,
```
kevinmarrec:
```suggestion
  component: RouteConfig['component'] | string
```

It will directly infer RouteConfig.component type (which is Component)

packages/config/types/router.d.ts
```diff
@@ -6,9 +6,14 @@
  */
 
 import { RouterOptions, RouteConfig } from 'vue-router'
+import Vue, { ComponentOptions, AsyncComponent } from 'vue'
```
kevinmarrec: You can remove these 3 following added lines (see how in other requested change)

kevinmarrec (changes requested):
Thanks for all the clarifications.
Your changes are totally relevant then.

Please take into account my following requested changes.

antfu:
@kevinmarrec Done

packages/config/types/router.d.ts
antfu: Good to know that, thank you

packages/config/types/router.d.ts
```diff
@@ -9,6 +9,8 @@ import { RouterOptions, RouteConfig } from 'vue-router'
 
 export interface NuxtRouteConfig extends RouteConfig {
   chunkNames: { [key: string]: string }
+  component: RouteConfig['component'] | string
```
kevinmarrec:
```suggestion
  component?: RouteConfig['component'] | string
```

I just checked https://github.com/vuejs/vue-router/blob/dev/types/router.d.ts#L77, the property should be optional.

kevinmarrec (changes requested):

packages/config/types/router.d.ts
antfu: Sorry for that 🤦‍♂️

kevinmarrec (approved):

PR #5854, opened 2019-06-02
refactor: prepare for external typescript support

kevinmarrec:
## Types of changes

## Description
Refactor some internals so it can fits to the outsourcing/externalization of the typescript support.

1) Ability to make modules extend Builder supportedExtensions + refactor of `createRoutes` function signature to fit better and pass these supported extensions.

2) Ability to make modules customize Webpack WarningFix plugin filters.

3) Ability to make modules grab the Nuxt babelLoader (with our predefined babel options) to use it when extending webpack config. Useful to use it with in typescript module with `ts-loader` to handle `tsx` files.

Related : https://github.com/nuxt/nuxt.js/pull/5858

## Checklist:


packages/builder/src/builder.js
```diff
@@ -44,7 +44,7 @@ export default class Builder {
       restart: null
     }
 
-    this.supportedExtensions = ['vue', 'js', 'ts', 'tsx']
+    this.supportedExtensions = ['vue', 'js', 'ts', 'tsx', ...(this.options.build.customSupportedExtensions || [])]
```
pi0: Can we simplify the option to something like `build.additionalExtensions`?

packages/webpack/src/config/base.js
```diff
@@ -259,12 +262,12 @@ export default class WebpackBaseConfig {
           // item in transpile can be string or regex object
           return !this.modulesToTranspile.some(module => module.test(file))
         },
-        use: perfLoader.js().concat(babelLoader)
+        use: perfLoader.js().concat(this.getBabelLoader())
```
pi0: Why we have to call `getBabelLoader` multiple times?

packages/webpack/src/config/base.js
```diff
@@ -422,12 +425,7 @@ export default class WebpackBaseConfig {
 
   warningFixFilter() {
     const { buildOptions, options: { _typescript = {} } } = this.buildContext
-    const filters = [
-      // Hide warnings about plugins without a default export (#1179)
-      warn => warn.name === 'ModuleDependencyWarning' &&
-        warn.message.includes(`export 'default'`) &&
-        warn.message.includes('nuxt_plugin_')
-    ]
+    const filters = buildOptions.warningFixFilters || []
```
pi0: Actually, I've used a really internal name when creating this plugin. For exposing I suggest a better name like `warningIgnoreFilters` (for consistency we can also rename internal plugin name)

PS: `buildOptions` is same as `options` (https://github.com/nuxt/nuxt.js/blob/v2.8.0/packages/builder/src/context/build.js#L10) -- we can try to make usage more consistent in config refactor.

packages/webpack/src/config/base.js
```diff
@@ -443,7 +441,7 @@ export default class WebpackBaseConfig {
     const { extend } = this.buildContext.buildOptions
     if (typeof extend === 'function') {
       const extendedConfig = extend.call(
-        this.builder, config, { loaders: this.loaders, ...this.nuxtEnv }
+        this.builder, config, { babelLoader: this.getBabelLoader(), loaders: this.loaders, ...this.nuxtEnv }
```
pi0: What's the point of creating a new babel loader object and passing it to extend here?

packages/builder/src/builder.js
```diff
@@ -327,12 +327,11 @@ export default class Builder {
...
 
     if (this._defaultPage) {
-      templateVars.router.routes = createRoutes(
-        ['index.vue'],
-        this.template.dir + '/pages',
-        '',
-        this.options.router.routeNameSplitter
-      )
+      templateVars.router.routes = createRoutes({
+        files: ['index.vue'],
```
pi0: As of usage change can you please double check that `createRoutes` is not exposed anywhere? (ensuring no one will break by this)

pi0: BTW nice change :+1:

packages/config/src/config/build.js
```diff
@@ -120,5 +120,11 @@ export default () => ({
       /vue-ssr-(client|modern)-manifest.json/
     ]
   },
-  friendlyErrors: true
+  friendlyErrors: true,
+  warningFixFilters: [
+    // Hide warnings about plugins without a default export (#1179)
```
pi0: Maybe we can move to add this filter to `warningFixFilter()` and always append? (as it doesn't make sense to be configurable)

packages/webpack/src/config/base.js
pi0: OK. I saw the intention in the comments but still, this is making an unnecessary object when not needed. Can't we extract babel loader config from `config`?

packages/builder/src/builder.js
kevinmarrec: @pi0 Sure, I had difficulties to find a proper name but your proposal is perfect :)
I'll add it in default config `packages/config/src/config/build.js` with empty array default value and will update snapshots.

packages/webpack/src/config/base.js
kevinmarrec: Well in fact it in this method it's called 3 times, 2 being TypeScript related and will be removed soon.
So there will be one call here + one call later for the ability to have this babelLoader.

I resolve this as you suggested extracting babel loader config from config in another comment, let's talk there.

packages/webpack/src/config/base.js
kevinmarrec: Alright, will update that.

packages/builder/src/builder.js
kevinmarrec: @pi0 I checked, updated tests & snapshots

packages/config/src/config/build.js
kevinmarrec: @pi0 So make it as empty array as default value, move back the current filter in `warningFixFilters` and append here ?

Like 

```js
warningIgnoreFilter() {
  const filters = [
    warn => warn.name === 'ModuleDependencyWarning' &&
        warn.message.includes(`export 'default'`) &&
        warn.message.includes('nuxt_plugin_'),
    ...this.buildContext.buildOptions.warningIgnoreFilters
  ]

  return warn => !filters.some(ignoreFilter => ignoreFilter(warn))
}
``` ?

File: packages/webpack/src/config/base.js (hunk 4 above)
Comment (kevinmarrec) [reply packages/webpack/src/config/base.js]: @pi0 The thing is that the babel loader config depends on `this.getBabelOptions` which are dynamic depending on if it's `modern` mode or not (with class inheritance), and they also depend on config build options.

What should we do then ? I don't know how we can extract it.

All I know is that the final result need to make `babelLoader` config available in Nuxt modules.

File: packages/config/src/config/build.js (hunk 6 above)
Comment (pi0) [reply packages/config/src/config/build.js]: The proposed change looks good to me. About wording, yes please that would be great!

File: packages/webpack/src/config/base.js (hunk 4 above)
Comment (pi0) [reply packages/webpack/src/config/base.js]: The module can use this in `extend` to extract specific babel config:

```js
const babelLoader = config.module.rules.find(r => r.test.toString() === '/\\.jsx?$/i').use[0]
```

Seems little hacky but prevents calling `getBabelLoader()` multiple times...

----

Actually, my concern is just that duplicate call... What do you think about these possible options:

1. Compute `babelLoader` inside the base constructor and pass it to the context like `{ babelLoader: this.babelLoader }`
2. Pass `{ webpackConfig: this, loaders: ... }`
3. The `find` method proposed above
4. Keep it as is. Ignoring little `getBabelLoader()` overhead

/cc @clarkdo

kevinmarrec: Even if hacky I like the 3rd one.
As the statement is `use: perfLoader.js().concat(babelLoader)` I guess it should be `use[use.length - 1]` ?

kevinmarrec: And honestly has the babelLoader is need for a `tsx` context, it sounds not so hacky to grab babelLoader with `jsx` regex :)

kevinmarrec: Reverted babelLoader changes

kevinmarrec:
@pi0 Should we rename https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/plugins/warnfix.js#L12 too ?
`warignore.js` / `warning-ignore.js` ?

packages/webpack/src/config/base.js
kevinmarrec: https://github.com/nuxt/typescript/blob/master/packages/typescript/lib/module.js#L33

pi0:
@kevinmarrec yes, please. seems a more reasonable name now

kevinmarrec:
PR should be fine now, i'm not sure if it's relevant adding documentation for `build.warningIgnoreFilters` & `build.additionalExtensions` as it's things that users won't customize at all.

pi0 (approved):
LGTM

atinux (approved):
LGTM

PR #5865, opened 2019-06-04
fix(cli): don't mutate options export

pi0:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Don't directly mutate the exported object from `nuxt.config.js`.

fixes #5840 and a workaround for standard-things/esm#813

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #5864, opened 2019-06-04
fix(vue-app): reduce consola direct access

pi0:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

- Consola is a fancy wrapper. Directly using for server bundle adds extra dependency which is unnecessary for serverless environments
- Don't override console.log for browser. fixes #5842

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #5866, opened 2019-06-04
v2.8.1

pi0:
### 🐛 Bug Fixes

- `vue-renderer`
  - #5867 Fix exception on property access of undefined object
  - #5863 Prevent "can't set headers after they are sent" error in modern server mode
- `vue-app`
  - #5864 Reduce consola direct access and don't override browser `console.log`
- `cli`
  - #5865 Don't mutate options export to prevent ESM regression with nuxt.config in mixed (cjs + esm) mode (**TIP:** never mix them as is non-standard and may be unsupported in any future version)

### 💅 Refactors

- `general`
  - #5854 Prepare for external typescript support

### 💖 Thanks to

- Rafał Chłodnicki (@rchl)


atinux (approved):

kevinmarrec:
Dunno if we can ship https://github.com/nuxt/nuxt.js/pull/5854 as it's ready, but it's also fine if it only lands in next minors or next major.

pi0:
@kevinmarrec Refactor is safe to be in a patch. But do you think releasing just these changes would be useful for users? (Without also moving leftovers to the module)

kevinmarrec:
@pi0 No but you can look my draft PR which remove TypeScript support from core and it's kind of lot of changes, it will make easier to review if we can merge this first part before.

aldarund (approved):

PR #5746, opened 2019-05-15
fix(vue-app): remove trailing slash in vue-router non-strict mode

clarkdo:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Fix #5593 

According to https://github.com/vuejs/vue-router/blob/dev/src/create-route-map.js#L167 , `vue-router` will remove the trailing slash in non strict mode.

Imagine we have routes like and there is `NuxtChild` and `NuxtLink to='/test/post/1'` in `pages/test.vue`:

In non-strict mode, after we click link, the `route.matched[0]` is always `{ path: '/test', ... }`

```js
{
  path: "/test",
  name: "test",
  children: [{
    path: "post/:slug"
  }]
}
``` 

When we access `http://localhost:3000/test`, `routerViewKey` is always `/test` before and after clicking link.

But if we access `http://localhost:3000/test/`, the `routerViewKey` is `/test/` from [`this.route.path`](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt.js#L67). After clicking link, the `routerViewKey` is `/test` from [`this.$route.matched[0].path`](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt.js#L41), so `test.vue` will be destroyed and recreated which is incorrect.

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

Issue #5593, opened 2019-04-24
Modal post doesn't open inside index route but it does in different routes


roelmagdaleno:
### Version

[v2.6.2](https://github.com/nuxt.js/releases/tag/v2.6.2)

### Reproduction link

[https://codesandbox.io/s/2v4vyr850n](https://codesandbox.io/s/2v4vyr850n)

### Steps to reproduce

You have to go to these routes (or click the links in the index route):

`/timewall` - works
`/testing` - works
`/` - not working

After that, you need to click on the post card and it will open the modal. This won't happen inside the index route and will show you the error.

### What is expected ?

Open the post modal inside `/timewall` and `/`

### What is actually happening?

Trying to open the modal in index route, will throw this error:

`_vm.$refs.postModal is undefined`

### Additional comments?

I configured my nuxt.config.js file to add to the previous routes a children so it can access the `post/_slug.vue` template inside the routes. 

![2019-04-24 00_47_10-web-app [F__Desarrollos_Doki_web-app] - ..._nuxt.config.js [web-app] - PhpStorm.png](https://cmty.s3.eu-west-1.amazonaws.com/71995937/966fdbf0-6654-11e9-bc99-8390039d252a.png)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9096">#c9096</a>)</em></sub></div>

clarkdo:
@roelmagdaleno 

In you case, `/post/_slug` is child of `/testing`, but `/post/_slug` is a root level path, it's not a suggested structure.

I suggest that you can remove `/` of `path: "/post/:slug"` in `extendRoutes` of `nuxt.config.js` and leverage lifecycle method to control the modal.

```js
// nuxt.config.js
path: "post/:slug",

// timewall.vue

// remove click event
-- @click.prevent="$refs.postModal.show()"

// control modal in mounted
  mounted() {
    if (this.$route.name.startWith('modal-post-from-')) {
      this.$refs.postModal.show();
    }
  },
```

roelmagdaleno:
Hey @clarkdo, thanks for your solution.

I tested it in *v2.6.0* and works perfectly, but in more updated versions (*v2.8.0*) doesn't work, it doesn't open the modal.

So what I did to fix it was to add a watcher inside my `Timewall.vue` component:

````
watch: {
        '$route': function(to) {
[...]

PR #5867, opened 2019-06-04
fix(vue-app): fix exception on property access of undefined object

rchl:
## Types of changes

## Description
After PR #5746, which has added code to remove trailing slash
in non-strict mode, I've gotten exception reported through Sentry due to
trying to get proper 'path' of undefined 'matchedRoute' object.

I don't know how it happens but it's clear error as code above checks
for null `matchedRoute` but not the code that checks for trailing slash.

Made the logic so that non-replaced path is returned when no route
matches. Same as before original fix.

Follow-up fix to #5593

## Checklist:

### DIFF
--- a/packages/vue-app/template/components/nuxt.js
+++ b/packages/vue-app/template/components/nuxt.js
@@ -42,7 +42,12 @@ export default {
       }
 
       const [matchedRoute] = this.$route.matched
-      const Component = matchedRoute && matchedRoute.components.default
+
+      if (!matchedRoute) {
+        return this.$route.path
+      }
+
+      const Component = matchedRoute.components.default
 
       if (Component && Component.options) {
         const { options } = Component


rchl:
@clarkdo please review

clarkdo (commented):
@rchl Good catch.

The checking is necessary, but I suggest returning first if `matchedRoute` is undefined.

```js
const [matchedRoute] = this.$route.matched

if (!matchedRoute) {
  return this.$route.path
}

const Component = matchedRoute && matchedRoute.components.default
// ...
```

clarkdo (approved):

clarkdo:
@rchl Thanks

pimlie:
@rchl You mentioned you didnt know how the error happened, are you using nested routes in your app by any chance? If so then this issue might be related: https://github.com/nuxt/nuxt.js/issues/5857

rchl:
@pimlie No, I'm not using nested routes (nuxtChild I presume). It only happened once though and it is nothing given number of requests. I get strangest errors reported by Sentry browser client anyway...

PR #5869, opened 2019-06-04
chore: rename backers -> nuxters

znarf:
It needs to match the tier that was renamed on Open Collective

see: https://opencollective.com/nuxtjs

Looks like "partners" and "sponsors" also needs to be updated, what to do?


atinux (approved):

Atinux:
Thanks @znarf 

About the sponsors, I will let @alexchopin  see how he wants to change it :)

znarf:
@Atinux @alexchopin I see 2 possibilities:

- Combine all sponsors together (regardless their tiers)
- Keep one entry per sponsor tier (but this will be verbose, moving from 2 tiers to 4 tiers).

alexchopin:
Done 👍

PR #5891, opened 2019-06-07
fix(vue-app): always finish transition.leave asynchronously

### LINKED ISSUE #5703: Mounted twice when layout change
### Version

[v2.6.2](https://github.com/nuxt.js/releases/tag/v2.6.2)

### Reproduction link

[https://codesandbox.io/s/wn7j570y85](https://codesandbox.io/s/wn7j570y85)

### Steps to reproduce

Simply click on button named "Mounted triggered twice, see console" and see in console.
(reproducable only with css:false for transition)

### What is expected ?

mounted once

mounted twice

### Additional comments?

The bug is still present on last release 2.7.1
It can be fixed quickly ? it's really a brake on client-side performance.
(Related : https://github.com/nuxt/nuxt.js/issues/5616)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9188">#c9188</a>)</em></sub></div>

pimlie:
Fix: #5703

## Types of changes

## Description

When `transition.css = false` and no leave listener with a callback/done argument is provided, the leave transition callback is called [synchronously](https://github.com/vuejs/vue/blob/4de4649d9637262a9b007720b59f80ac72a5620c/src/platforms/web/runtime/modules/transition.js#L277).

This causes a problem when switching layouts in Nuxt. Eg given page A with layout A and page B with layout B the following roughly happens (thus with `transition.css = false`):
1. when page A is loaded the component tree is: Layout A -> Nuxt -> NuxtChild -> Transition -> RouterView -> Page A
2. next user clicks link to page B
3. Client app loads page component B and determines it needs layout B
4. layout B is preloaded (App.loadLayout)
5. router.next() is called
6. leave transition for Page A is called
7. leave transition immediately finishes (synchronously)
8. Page B is loaded within Layout A
9. router.afterEach calls Client.fixPrepatch
10. Client.showNextPage is called
11. Layout B is loaded (App.setLayout)
12. Page B is loaded

By calling the leave callback in step 6/7 asynchronous (as would happen when eg transition.css is not false), all subsequent steps are called before the leave callback is called which means that Layout B will already be loaded and Page B is never loaded in Layout A.

## Checklist:

### DIFF
--- a/packages/vue-app/template/components/nuxt-child.js
+++ b/packages/vue-app/template/components/nuxt-child.js
@@ -60,7 +60,8 @@ export default {
         if (leave) {
           leave.call(_parent, el)
         }
-        return _parent.$nextTick(done)
+
+        _parent.$nextTick(done)
       }
     }
 
--- a/packages/vue-app/template/components/nuxt-child.js
+++ b/packages/vue-app/template/components/nuxt-child.js
@@ -51,18 +51,25 @@ export default {
       if (beforeEnter) return beforeEnter.call(_parent, el)
     }
 
-    let routerView = [
-      h('router-view', data)
-    ]
-    if (props.keepAlive) {
-      routerView = [
-        h('keep-alive', { props: props.keepAliveProps }, routerView)
-      ]
+    // make sure that leave is called asynchronous (fix #5703)
+    if (transition.css === false) {
+      const leave = listeners.leave
+      listeners.leave = (el, done) => {
+        if (leave) leave.call(_parent, el)
+        return _parent.$nextTick(done)
+      }
     }
+
+    let routerView = h('routerView', data)
+
+    if (props.keepAlive) {
+      routerView = h('keep-alive', { props: props.keepAliveProps }, [routerView])
+    }
+
     return h('transition', {
       props: transitionProps,
       on: listeners
-    }, routerView)
+    }, [routerView])
   }
 }
 
--- a/packages/vue-app/template/components/nuxt-child.js
+++ b/packages/vue-app/template/components/nuxt-child.js
@@ -48,14 +48,18 @@ export default {
       window.<%= globals.nuxt %>.$nextTick(() => {
         window.<%= globals.nuxt %>.$emit('triggerScroll')
       })
-      if (beforeEnter) return beforeEnter.call(_parent, el)
+      if (beforeEnter) {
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/components/nuxt-child.js
```diff
@@ -51,18 +51,25 @@ export default {
...
     }
 
-    let routerView = [
-      h('router-view', data)
-    ]
+    // make sure that leave is called asynchronous (fix #5703)
+    if (transition.css === false) {
+      const leave = listeners.leave
+      listeners.leave = (el, done) => {
+        if (leave) leave.call(_parent, el)
```
manniL: Nit: no inline if's

packages/vue-app/template/components/nuxt-child.js
```diff
@@ -51,18 +51,25 @@ export default {
...
-        h('keep-alive', { props: props.keepAliveProps }, routerView)
-      ]
+      routerView = h('keep-alive', { props: props.keepAliveProps }, [routerView])
     }
+
     return h('transition', {
       props: transitionProps,
       on: listeners
-    }, routerView)
+    }, [routerView])
```
pi0: Why do we need to pass children as an array? Is this a workaround or something?

pimlie: Not sure, but previously the routerView variable itself was an array. I changed that to this for better readability

pimlie: Fyi, this was without an array :)
```
Test Suites: 25 failed, 1 skipped, 74 passed, 99 of 100 total
Tests:       97 failed, 15 skipped, 655 passed, 767 total
```

pi0 (approved):

pimlie:
@pi0 @Atinux One consideration that we might take away from this issue is that Nuxt.js fully relies on the asynchronous nature of transition resolving to prevent page components to be loaded twice when switching layouts. Not sure if this was an explicitly known thing or not, it took me quite a while at least to discover that ;)

atinux (approved):

Atinux:
Good catch @pimlie, these transitions makes it harder to work with the layouts actually, as well as the scrollBehaviour.

PR #5896, opened 2019-06-08
examples(typescript): use Vue.extend

isy:
## Types of changes

## Description
The syntax of typescript example has been changed. Because the proposal of Class API has been dropped.
https://github.com/vuejs/rfcs/pull/17#issuecomment-494242121

## Checklist:

### DIFF
--- a/examples/typescript-eslint/package.json
+++ b/examples/typescript-eslint/package.json
@@ -3,8 +3,7 @@
   "private": true,
   "dependencies": {
     "nuxt": "latest",
-    "ts-node": "^8.1.0",
-    "vue-property-decorator": "^7.3.0"
+    "ts-node": "^8.1.0"
   },
   "scripts": {
     "dev": "nuxt",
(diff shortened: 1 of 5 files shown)


Atinux:
I actually do like these changes (more deletion than addition) but I am not a TS expert, will let @kevinmarrec approve/reject.

Thanks for the PR @isy

kevinmarrec:
@isy @Atinux I think there is a misunderstanding.

Vue Class API RFC **IS NOT** `vue-class-component`, it was something new that was supposed to replace it.

So there is no need for update yet, I suggest wait for their composition functions https://github.com/vuejs/rfcs/blob/function-apis/active-rfcs/0000-function-api.md to see how it handle better TS support.

I still advice `vue-class-component` / `vue-property-decorator` over `Vue.extend` for typed component interfaces & overall better DX.

isy:
I was misunderstood.

Thx for your response. @kevinmarrec  🙏

PR #5911, opened 2019-06-11
WIP: feat: new routing system to allow multiple route params

atinux:
## Types of changes

The breaking change only affects users that have a `$` in the files inside `pages/` (should not exists but we never know).

## Description

The idea is to replace the `pages/_dynamic.vue` to `pages/$dynamic.vue`.

Main advantage, usage of params:
- In the middleware of the route (`pages/author-$username.vue`)
- Multiple route params (`pages/product-$sku-$variant.vue`)

I also added a warning for easier migration.

This change is planned for **Nuxt 3**.

## Checklist:

### DIFF
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -210,7 +210,7 @@ export const guardDir = function guardDir(options, key1, key2) {
   }
 }
 
-const isWildcardRoute = (key) => (key === '_' || key === '$')
+const isWildcardRoute = key => (key === '_' || key === '$')
 
 const getRoutePathExtension = (key) => {
   if (isWildcardRoute(key)) {
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -141,6 +141,7 @@ export const createRoutes = function createRoutes({
   const routes = []
   files.forEach((file) => {
     const keys = file
+      .toLowerCase()
       .replace(new RegExp(`^${pagesDir}`), '')
       .replace(new RegExp(`\\.(${supportedExtensions.join('|')})$`), '')
       .replace(/\/{2,}/g, '/')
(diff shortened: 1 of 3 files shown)


Atinux:
We may need to clean the `next` branch so we can start working on it again :)

ricardogobbosouza:
@Atinux 
I think the `$` key could be configurable, what do you think?

begueradj:
But this is a breaking change. I mean application done with previous Nuxt.js versions will need to be re-written.

ricardogobbosouza:
Being configurable and defined the default with `_` can be released as an improvement and in nuxt 3 it would change to `$`

Atinux:
@begueradj actually it will still work with `_dynamic.vue` with Nuxt 3 but will be deprecated and removed with Nuxt 4.

I don't think it's a good idea to make it configurable since it starts becoming a standard.

The *only breaking change* is that if someone has a `$` in the pages files 😄

evillt:
Awesome!

Did you guys ever considered support [Passing Props to Route Components](https://router.vuejs.org/guide/essentials/passing-props.html#passing-props-to-route-components)?

I don't know how many guys using this feature, but support this feature not a bad idea to nuxt, huh?🤔

Atinux:
You can add this feature by using `extendRoutes` @evillt :)

We are going to move to `[variable]` instead since `$` is used by the shell, see https://github.com/zeit/next.js/issues/7607#issuecomment-503767041

evillt:
@Atinux 

Yes!

But I want to mean is:

Create a file `{id}.vue` under the `pages` folder, will auto transform to

```js
{
  path: '/:id',
  props: true,
  ...
}
```

by nuxt routing system.

Of course advanced route config will still using `extendRoutes`.

Atinux:
@evillt for this I recommend to use the [router-extras-module](https://github.com/nuxt-community/router-extras-module) to active such options :)

evillt:
@Atinux Sure!

Atinux:
Finally, since `[]` will also break in next OS X terminal, the `_` is the better alternative from my POV so we will stick to it :)

PR #5929, opened 2019-06-13
feat: upgrade css-loader to v3

clarkdo:
### Breaking change

- @nuxt/webpack requires Node.js >= 8.9
- css-loader v3 breaking changes in options


packages/config/src/config/build.js
```diff
@@ -41,7 +41,9 @@ export default () => ({
     },
     css: {},
     cssModules: {
-      localIdentName: '[local]_[hash:base64:5]'
+      modules: {
+        localIdentName: '[local]_[hash:base64:5]'
```
pi0: Do you think is it good to normalize in `options.js` or make a warning about config breaking change?

clarkdo: I think it would be better to keep consistency with css-loader, so maybe warning is better

clarkdo: css-loader is a very basic loader, v3 breaking changes must have its consideration, so I think it’s better to align its behavior.
And it only affects users who have customized cssModues config, they should have a clear idea about what need to be changed

clarkdo: BTW, css-loader has options schema validation 😺

pi0:
@clarkdo is this PR ready to merge?

pi0 (approved):

PR #5858, opened 2019-06-03
feat: externalize typescript support

kevinmarrec:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
External repo : https://github.com/nuxt/typescript

## TODO before ready
- [x] Refactor core to be able to externalize TypeScript Support implementation (#5854)
- [x] Merge the refactor in `dev` & merge `dev` into the branch related to this PR
- [x] Implement TypeScript support in external repo
- [x] Write new tests in external repo
- [x] Move and update examples in external repo
- [x] Publish new `@nuxt/typescript-build`, `@nuxt/typescript-runtime` & `@nuxt/types` packages
- [ ] Prepare docs for migration (https://github.com/nuxt/typescript/docs)

## <span id="migration">Notes for Nuxt 2.9 migration</span>

TypeScript support has been externalized in dedicated packages and have been removed from core starting Nuxt 2.9.  Here are the guidelines to migrate your existing Nuxt TypeScript project to 2.9 without any trouble.

The following migration guide works for either `nuxt` or `nuxt-edge`.

#### 1) Migrate dependencies

```
yarn remove @nuxt/typescript
yarn add -D @nuxt/typescript-build @nuxt/types
```
> Use `npm uninstall` and `npm install -D` instead if you're using **npm**

#### 2) Replace `@nuxt/vue-app` and `@nuxt/config` by `@nuxt/types` in your `tsconfig.json`

```js
// tsconfig.json

"compilerOptions": {
  "types": [
     "@nuxt/types"
  ]
}
```

If you were importing types from `@nuxt/config` you need to instead import them from `@nuxt/types`. Types imports might have changed a little, you can get familiar with them either by triggering intellisense when importing or watch them live [here](https://github.com/nuxt/typescript/tree/master/packages/types) now.

#### 3) Add `@nuxt/typescript-build` module to your `nuxt.config.js`

```js
// nuxt.config.js

export default {
  modules: ['@nuxt/typescript-build']
}
```

#### 4) Move the customized options from `build.typescript` to module options

```js
// nuxt.config.js

export default {
  typescript: {
    typeCheck: true,
    ignoreNotFoundWarnings: true
  }
}
```

You can also do it this way :
```js
// nuxt.config.js

export default {
  modules: [
    ['@nuxt/typescript-build', {
      typeCheck: true,
[...]

PR #5946, opened 2019-06-17
feat(types): add vuex definition

orblazer:
## Types of changes

## Description

This add types definitions for `vuex`. With that we now have types in Actions/Mutations on :
- [`this.app`](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L105)
- [`this.$router`](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L52)

## Checklist:

### DIFF
--- a/packages/vue-app/types/vuex.d.ts
+++ b/packages/vue-app/types/vuex.d.ts
@@ -4,7 +4,7 @@ import VueRouter from 'vue-router'
 
 declare module 'vuex/types/index' {
   interface Store<S> {
-    app: Vue;
-    $router: VueRouter;
+    app: Vue
+    $router: VueRouter
   }
 }
--- a/packages/vue-app/types/vuex.d.ts
+++ b/packages/vue-app/types/vuex.d.ts
@@ -1,0 +1,10 @@
+import Store from 'vuex'
+import Vue from 'vue'
+import { Route } from 'vue-router'
+
+declare module 'vuex/types/index' {
+  interface Store<S> {
+    app: Vue;
+    route: Route;
+  }
+}
--- a/packages/vue-app/types/vuex.d.ts
+++ b/packages/vue-app/types/vuex.d.ts
@@ -1,10 +1,10 @@
 import Store from 'vuex'
 import Vue from 'vue'
-import { Route } from 'vue-router'
+import VueRouter from 'vue-router'
 
 declare module 'vuex/types/index' {
   interface Store<S> {
     app: Vue;
-    route: Route;
+    $router: VueRouter;
   }
 }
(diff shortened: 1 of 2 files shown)


kevinmarrec:
@orblazer Thanks for the PR

I've found https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L105 for `store.app` & https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L52 for `store.$router` but I couldn't find for `store.route`, maybe you found it somewhere in documentation ?

Maybe another `store`props added by Nuxt would need type definitions ? I would want to be sure we cover properties that are properly existing.

Also, for consistency with other typedefs, you can remove semicolons from your type definition (I may review all types code style later as i'm working on a dedicated `@nuxt/types` in association with externalization of TypeScript support #5858)

packages/vue-app/types/vuex.d.ts
```diff
@@ -0,0 +1,10 @@
+import Store from 'vuex'
+import Vue from 'vue'
+import VueRouter from 'vue-router'
+
+declare module 'vuex/types/index' {
+  interface Store<S> {
+    app: Vue;
```
kevinmarrec:
```suggestion
    app: Vue
```

packages/vue-app/types/vuex.d.ts
```diff
@@ -0,0 +1,10 @@
...
+    $router: VueRouter;
```
kevinmarrec:
```suggestion
    $router: VueRouter
```

kevinmarrec (changes requested):

kevinmarrec (changes requested):
You also need to update `types/index.d.ts` and add

```js
// augment typings of Vuex
import './vuex'
```
after
```js
import './vue'
```

😜

kevinmarrec:
Thanks for quick updates @orblazer , do you know if there's any other prop we may miss ?
Either way, looks good to me

orblazer:
For Vuex types its goof for me too @kevinmarrec

kevinmarrec (changes requested):
Hum, after some thoughts, `app` should be of type `NuxtAppOptions`, see https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/types/index.d.ts#L17

can be imported like that : `import { NuxtAppOptions } from '.'` or `import { NuxtAppOptions } from './index'`

orblazer:
Yes i have just realized that ;)

kevinmarrec (approved):

PR #5947, opened 2019-06-17
fix(vue-app): prevent calling `asyncData` in keep-alive pages (#5945)

### LINKED ISSUE #5945: calling asyncData multiple times when keep-alive page activated
### Version

### Reproduction link

[https://codesandbox.io/embed/codesandbox-nuxt-7ostd](https://codesandbox.io/embed/codesandbox-nuxt-7ostd)

### Steps to reproduce

0. open browser tab to index page(ssr)
1. click link to route to /about
2. click button to go back to keep-alive index page

### What is expected ?

do not call asyncData in client when SSR rendered

calling asyncData multiple times when keep-alive page activated

### Additional comments?

related code
https://github.com/nuxt/nuxt.js/blob/f3d29e378087b6c6c686a03292c2bce409161110/packages/vue-app/template/client.js#L386-L397
https://github.com/nuxt/nuxt.js/blob/f3d29e378087b6c6c686a03292c2bce409161110/packages/vue-app/template/utils.js#L24-L31

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9372">#c9372</a>)</em></sub></div>

Elevista:
## Types of changes

## Description

Resolves: #5945
Reproduction link: https://codesandbox.io/embed/codesandbox-nuxt-7ostd

related code
https://github.com/nuxt/nuxt.js/blob/f3d29e378087b6c6c686a03292c2bce409161110/packages/vue-app/template/client.js#L386-L397
https://github.com/nuxt/nuxt.js/blob/f3d29e378087b6c6c686a03292c2bce409161110/packages/vue-app/template/utils.js#L24-L31
`Component.options.__hasNuxtData` value of SSR rendered keep-alive page is `true` 

## Checklist:


Atinux:
Thanks for the PR.

A test is failing and actually it should not since it actually does not use `keep-alive` (https://github.com/nuxt/nuxt.js/blob/dev/test/fixtures/children/layouts/patch.vue#L3).

In order to check it, I believe you may want to keep the instance of the page, if it exists, get its VNode and checks if it uses `keepAlive` (see https://github.com/vuejs/vue-router/blob/dev/src/components/view.js#L34)

Elevista:
@Atinux oh yeah... I forgot fixtures test 😨
I'll figure this out

Atinux:
Nice @Elevista 

Would be nice to complete with another test (I guess adding a new layout with `<keep-alive>`, something like `patch-keep-alive.vue` there: https://github.com/nuxt/nuxt.js/blob/dev/test/fixtures/children/layouts/)

leebyungjun-nbt:
@Atinux 
Ok, I'll try

Elevista:
little late because I'm busy with work these days. 😅

ericmuyser:
Why shouldn't asyncData be called again with keep-alive? `data` is, and `asyncData` is often used as a replacement simply to make it SSR compatible. I use the page `:id` param to fetch different model data in asyncData. Right now, the function is called, but it's not updating the UI, because it's missing the same workaround that `data` has..
https://github.com/nuxt/nuxt.js/blob/055e99a21b0c2f681025c5adc90fef19bed76b96/lib/app/client.js#L455-L457

Elevista:
@ericmuyser 
`data` function will be called only once in keep-alive page.
but `asyncData` is being called every time it active and makes nothing change
because it will be ignored here https://github.com/nuxt/nuxt.js/blob/f3d29e378087b6c6c686a03292c2bce409161110/packages/vue-app/template/utils.js#L24-L31
If there's something heavy ajax job in `asnycData`, it's really a problem.

Key of keep-alive page is route url, so if `params.id` is in url, 
It will create new instance for each `params.id`. This is not going to be your problem.

541xxx:
It's been a long time, could this merge request be reviewed or merged?

Elevista:
This was wrong approach..

PR #5950, opened 2019-06-17
feat(types): add nuxt error type

orblazer:
## Types of changes

## Description

This add type for Nuxt error with interface `NuxtError`.
With that on error layout we can now use :
```ts
import { Component, Vue, Prop } from 'nuxt-property-decorator'
import { NuxtError } from '@nuxt/vue-app'

@Component({})
export default class ErrorPage extends Vue {
  @Prop({ required: true }) readonly error!: NuxtError;
}
```

## Checklist:

### DIFF
--- a/packages/vue-app/types/index.d.ts
+++ b/packages/vue-app/types/index.d.ts
@@ -70,8 +70,8 @@ export interface Transition {
 
 export interface NuxtError {
   message?: string
-  statusCode?: number
   path?: string
+  statusCode?: number
 }
 
 export interface NuxtLoading extends Vue {
--- a/packages/vue-app/types/index.d.ts
+++ b/packages/vue-app/types/index.d.ts
@@ -73,6 +73,12 @@ export interface ErrorParams {
   message?: string
 }
 
+export interface NuxtError {
+  message?: string
+  statusCode?: string
+  path?: string
+}
+
 export interface NuxtLoading extends Vue {
   fail?(): NuxtLoading
   finish(): NuxtLoading
--- a/packages/vue-app/types/index.d.ts
+++ b/packages/vue-app/types/index.d.ts
@@ -39,7 +39,7 @@ export interface Context {
   res: ServerResponse
   redirect(status: number, path: string, query?: Route['query']): void
   redirect(path: string, query?: Route['query']): void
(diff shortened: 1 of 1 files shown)


kevinmarrec:
@orblazer Isn't it related to https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/types/index.d.ts#L71 ?

orblazer:
Yes @kevinmarrec we probably can merge the 2 interface et keep the name `NuxtError`

kevinmarrec:
@orblazer Alright, agreed. I would just want to know how you found the `path` property if you can link me code or documentation. Thanks in advance.

orblazer:
@kevinmarrec I have found `path` here : https://github.com/nuxt/nuxt.js/blob/31f5729828e1009a6db28fe90296c178451f9efd/packages/vue-app/template/server.js#L83

kevinmarrec:
@orblazer Alright, then I'd say we can remove `ErrorParams` in favor of  `NuxtError`. Then do
```ts
export interface NuxtError {
  message?: string
  path?: string
  statusCode?: number
}
```
> Comparing to your changes, notice `statusCode` as `number` and the alphabetical order of properties.

And also change `Context.error` signature to
```ts
error(error: NuxtError): void
```

packages/vue-app/types/index.d.ts
```diff
@@ -68,9 +68,10 @@ export interface Transition {
   leaveCancelled?(el: HTMLElement): void
 }
 
-export interface ErrorParams {
-  statusCode?: number
+export interface NuxtError {
   message?: string
+  statusCode?: number
+  path?: string
```
kevinmarrec: alphebtical order
```
message
path
statusCode
```

kevinmarrec (changes requested):

packages/vue-app/types/index.d.ts
orblazer: Its good, i have miss re order property ;)

kevinmarrec: 😁

kevinmarrec (approved):

kevinmarrec:
LGTM, thanks for the types improvement @orblazer

PR #5964, opened 2019-06-21
fix(vue-renderer): clone spa meta to prevent cache modification

### LINKED ISSUE #5962: Allow post-processing of the HTML response in SPA mode (for injecting initial data from server per request or runtime env variables)

On SPA projects, some sort of request specific data needs to be injected into the HTML. 

For example, by injecting currently logged in user's data to the html, we prevent an unnecessary api call when the page is rendered. 

The `render:route` hook could almost be used for this purpose, but it doesn't work properly.

Observe the following hook,
```
  hooks: {
    render: {
      route(url, result, context) {
        result.html = result.html + " transformed by render:route hook"
      }
    }
  },
```

This seems to work on first request, however due to LRU caching of the `result` object in SPA renderer (renderer/spa.js) which is mutable and hence can be modified afterwards, the string " transformed by render:route hook" is appended multiple times as the html property in the cached context is modified as well for subsequent reqs.

From what I can see, the current code does not allow for another a hook, and what we want is only possible via hijacking the response with a middleware (and res object's end method), which is not ideal.

Running in SSR mode is overkill for most apps. This change will make it easier to inject some server data in SPA apps.

Several options: 

- Use object.freeze to make sure that the cached object cannot be modified (but this could break other code that depends on this?) The PR at https://github.com/nuxt/nuxt.js/pull/5964 fixes this.
- Add ability to return a context from the hook, which can be destructured to get the latest html.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9391">#c9391</a>)</em></sub></div>

ustun:
This ensures that post-processing of the result object does not cause cache corruption.

When transforming the response in SPA mode via `render:route` hook, setting `result.html` changes the data in cache, which is not ideal. 

Fixes https://github.com/nuxt/nuxt.js/issues/5962

## Types of changes

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderers/spa.js
+++ b/packages/vue-renderer/src/renderers/spa.js
@@ -1,4 +1,5 @@
 import { extname } from 'path'
+import cloneDeep from 'lodash/cloneDeep'
 import Vue from 'vue'
 import VueMeta from 'vue-meta'
 import { createRenderer } from 'vue-server-renderer'
@@ -43,7 +44,9 @@ export default class SPARenderer extends BaseRenderer {
     let meta = this.cache.get(cacheKey)
 
     if (meta) {
-      return meta
+      // Return a copy of the content, so that future
+      // modifications do not effect the data in cache
+      return cloneDeep(meta)
     }
 
     meta = {
@@ -149,7 +152,9 @@ export default class SPARenderer extends BaseRenderer {
     // Set meta tags inside cache
     this.cache.set(cacheKey, content)
 
-    return content
+    // Return a copy of the content, so that future
+    // modifications do not effect the data in cache
+    return cloneDeep(content)
   }
 
   static normalizeFile(file) {
--- a/test/unit/spa.test.js
+++ b/test/unit/spa.test.js
@@ -138,6 +138,22 @@ describe('spa', () => {
     consola.log.mockClear()
   })
 
+  test('render:route hook does not corrupt the cache', async () => {
+    const window1 = await nuxt.server.renderAndGetWindow(url('/'))
+    const html1 = window1.document.body.innerHTML
+    expect(html1).toContain('extra html from render:route hook')
+    expect(html1.match(/render:route/g).length).toBe(1)
+
+    window1.close()
+
+    const window2 = await nuxt.server.renderAndGetWindow(url('/'))
+    const html2 = window2.document.body.innerHTML
+    expect(html2).toContain('extra html from render:route hook')
+    expect(html2.match(/render:route/g).length).toBe(1)
+
+    window2.close()
+  })
+
   // Close server and ask nuxt to stop listening to file changes
   afterAll(async () => {
     await nuxt.close()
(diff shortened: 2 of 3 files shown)


ustun:
Is this stuck because the commit starts with `Fix` instead of `fix`? I'm not very familiar with semantic pull request bot.

test/unit/spa.test.js
```diff
@@ -138,6 +138,22 @@ describe('spa', () => {
...
+    const html1 = window1.document.body.innerHTML
+    expect(html1).toContain('extra html from render:route hook')
+    expect(html1.match(/render:route/g).length).toBe(1)
+
+    window1.close()
+
+    const window2 = await nuxt.server.renderAndGetWindow(url('/'))
+    const html2 = window2.document.body.innerHTML
+    expect(html2).toContain('extra html from render:route hook')
+    expect(html2.match(/render:route/g).length).toBe(1)
```
pi0: How does this test ensure modification? Shouldn't we ensure html1 and html2 to not be the same?

Extra point: jest has [toMatch](https://jest-bot.github.io/jest/docs/expect.html#tomatchregexporstring) helper.

packages/vue-renderer/src/renderers/spa.js
```diff
@@ -149,7 +152,9 @@ export default class SPARenderer extends BaseRenderer {
     // Set meta tags inside cache
     this.cache.set(cacheKey, content)
 
-    return content
+    // Return a copy of the content, so that future
+    // modifications do not effect the data in cache
+    return cloneDeep(content)
```
pi0: content is not cached (and meta is already cloned) so do we need to clone it again?

pi0:
@ustun no PR is ok and thanks for that ❤️ Sorry for late review.

test/unit/spa.test.js
ustun: no, actually we want them to be different. without the patch, they are different, but the extra test is added multiple times.

on first request, we calculate the html and put it into the cache.
the hook modifies the data in the cache.

on second request, we get the html from the cache (with hook's modification) and hook runs again and adds the data another time.

What I'm trying to achieve here is the following:

- make sure the data is cached, but then modify it per request (per user etc). the modified response should not be put into the cache.

packages/vue-renderer/src/renderers/spa.js
ustun: content is cached on line 153, when we return this to be used later, any changes to content is reflected inside the cache.

test/unit/spa.test.js
ustun: The reason I didn't use toMatch is that I want to ensure there is a single match. Without the match, there is a match, but twice.

Let me give another example:

- First request, html is calculated and cached. Let's say html = "first request".
- Let's say we have a hook that adds: 'hook was here". Now the html returned (and html in cache - due to mutable js objects is) " first request hook was here"
- On second request, we get the html from the cache, append "hook was here", the result becomes "" first request hook was here hook was here"

And it keeps appending the same string on subsequent requests.

pi0: Ok. I was confused by `new Date()`

pi0 (approved):

PR #5975, opened 2019-06-24
feat(vue-renderer): add csp option for csp v1 compatibility

### LINKED ISSUE #5627: Cannot define both SHA hash and `unsafe-inline` in CSP for best compatibility
### Version

[v2.6.3](https://github.com/nuxt.js/releases/tag/v2.6.3)

### Reproduction link

[https://codesandbox.io/s/p91l8w1zzq?module=%2Fnuxt.config.js](https://codesandbox.io/s/p91l8w1zzq?module=%2Fnuxt.config.js)

### Steps to reproduce

Define both `sha256` as hashAlgorithm and  `'unsafe-inline' ` in CSP script-src policy, in nuxt.config.js

### What is expected ?

Both dynamic page SHA hash and  `'unsafe-inline' ` should be present in CSP header

Only  `'unsafe-inline' ` is present in CSP header

### Additional comments?

Follow up on https://github.com/nuxt/nuxt.js/pull/5387#issuecomment-486996379

original comment:

==========

Defining BOTH `unsafe-inline` and hash in CSP is actually a valid use case, it is for backward compatibility of CSPv1. That's why browsers with CSPv2+ support ignore it when hash is present at the same time.

[MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src):

    See unsafe inline script for an example. Specifying nonce makes a modern browser ignore 'unsafe-inline' which could still be set for older browsers without nonce support.

It is also a recommended practice suggested by google
image

With this PR in, how can I specify BOTH `unsafe-inline` and SHA hash  for my CSP rule?

==========

Would reverting the PR, and allow a special disable value for `hashAlgorithm` to solve @sambowler 's use case a better idea?

@manniL 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9134">#c9134</a>)</em></sub></div>

williamchong:
Fixes CSP compatibility mentioned in https://github.com/nuxt/nuxt.js/issues/5627

## Types of changes

## Description
Add a csp option `unsafeInlineCompatiblity` which allows `'unsafe-inline'` and hash based rules to appear together.
Not sure if the option name can be shorter.
Resolves: #5627 

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -220,6 +220,7 @@ export function getNuxtConfig(_options) {
       allowedSources: undefined,
       policies: undefined,
       addMeta: Boolean(options._generate),
+      unsafeInlineCompatiblity: false,
       reportOnly: options.debug
     })
   }
--- a/packages/config/test/options.test.js
+++ b/packages/config/test/options.test.js
@@ -92,6 +92,7 @@ describe('config: options', () => {
     expect(csp).toEqual({
       hashAlgorithm: 'sha256',
       addMeta: false,
+      unsafeInlineCompatiblity: false,
       allowedSources: true,
       policies: undefined,
       reportOnly: false,
--- a/packages/vue-renderer/src/renderers/ssr.js
+++ b/packages/vue-renderer/src/renderers/ssr.js
@@ -121,7 +121,7 @@ export default class SSRRenderer extends BaseRenderer {
     if (csp) {
       // Only add the hash if 'unsafe-inline' rule isn't present to avoid conflicts (#5387)
       const containsUnsafeInlineScriptSrc = csp.policies && csp.policies['script-src'] && csp.policies['script-src'].includes(`'unsafe-inline'`)
-      if (!containsUnsafeInlineScriptSrc) {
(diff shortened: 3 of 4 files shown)


TheAlexLichter (approved):
LGTM ☺️

clarkdo (approved):

clarkdo:
@williamchong007 A pr for updating doc will be great.

koehn:
This would have been better had you spelled `compatability` correctly…

pi0:
@koehn It is already fixed with https://github.com/nuxt/nuxt.js/pull/6583 can you spot typo in docs?

koehn:
Not sure if this is what you mean; AFAICT it's still in use as of 2.12.2. Am I looking in the wrong place?

https://nuxtjs.org/api/configuration-render/

pi0:
Apologies for wrong docs :see_no_evil: Patch is available since [2.10.2](https://github.com/nuxt/nuxt.js/blob/v2.10.2/packages/config/src/options.js#L244). You can use `unsafeInlineCompatibility`. Docs need to be updated. PR welcome :)

PR #5977, opened 2019-06-24
fix: downgrade fs-extra to v7 for fixing jest crashing

ricardogobbosouza:
## Types of changes

## Description

We can't upgrade to the latest version of `fs-extra` due to: https://github.com/jprichardson/node-fs-extra/issues/687

## Checklist:


clarkdo:
From the issue, I think issue is from `graceful-js`, should we wait ?

pi0:
[vue-cli](https://github.com/vuejs/vue-cli/commit/43bdc333f4c714786bc8b5c67b34d74f465d8151) also downgraded because of this. Seems like a good workaround but the root cause is unlikely to be resolved anytime soon which means keep using an outdated dependency. I'm +1 to merge this fix but suggest using a simpler alternative for promise based fs.

ricardogobbosouza:
@clarkdo Yeah, but it is causing problems in dev modules, for example sitemap-module

clarkdo:
Cool, then please update yarn.lock and add it into ignore list in renovate config.

clarkdo (dismissed):

packages/webpack/package.json
```diff
@@ -22,7 +22,7 @@
     "eventsource-polyfill": "^0.9.6",
     "extract-css-chunks-webpack-plugin": "^4.5.2",
     "file-loader": "^4.0.0",
-    "fs-extra": "^8.0.1",
+    "fs-extra": "^7.0.1",
```
clarkdo: It seemed that webpack no longer depends on fs-extra,  this can be removed

ricardogobbosouza: Yeah

renovate.json
```diff
@@ -6,7 +6,8 @@
     "dev"
   ],
   "ignoreDeps": [
-    "core-js"
+    "core-js",
+    "fs-extra"
```
clarkdo: @ricardogobbosouza Could you please add a todo comment here for tracking the real root issue ?

ricardogobbosouza: Sure!

clarkdo:
@ricardogobbosouza Thanks 😄

pi0:
Seems merged: https://github.com/isaacs/node-graceful-fs/pull/162

clarkdo:
@pi0 @ricardogobbosouza PR: https://github.com/nuxt/nuxt.js/pull/5994, please verify and review

PR #6026, opened 2019-07-03
feat: allow disabling loading screen

### LINKED ISSUE #6025: Add option to disable the loading screen

This solves the issue where some e2e test runners depend on the site being fully loaded when starting tests. Currently the site shows as loaded to the test framework which causes testing to prematurely start and fail.

I propose to add a CLI flag to disable the loading screen for testing purposes.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9450">#c9450</a>)</em></sub></div>

rinu:
## Types of changes

## Description

Adds a cli option to disable the loading screen.

Resolves #6025

## Checklist:


packages/cli/src/utils/config.js
```diff
@@ -74,5 +74,7 @@ export async function loadNuxtConfig(argv) {
     socket: argv['unix-socket'] || undefined
   }, options.server || {}, getDefaultNuxtConfig().server)
 
+  options.noLoadingScreen = !!argv['no-loading-screen']
```
Atinux: I will instead call is `loadingScreen` which is `true` by default :)

packages/config/src/options.js
```diff
@@ -369,7 +369,7 @@ export function getNuxtConfig(_options) {
   }
 
   // Add loading screen
-  if (options.dev) {
+  if (options.dev && !options.noLoadingScreen) {
```
Atinux: Same here, using `options.loadingScreen`

atinux (changes requested):

packages/cli/src/utils/config.js
rinu: I tried to do that but it didn't work. Help will still display `--no-loading-screen`

Atinux: What about `options.loadingScreen = !argv['no-loading-screen']`

Atinux:
We also need to set next to https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/config/_app.js#L43

`loadingScreen: true`

Also, a test need to be added as well as docs.

packages/cli/src/utils/config.js
rinu: That might work.

rinu:
Looking at the current documentation for cli I don't see any reasonable way to add it. Perhaps a change to the documentation is not required after all.

pimlie (commented):
I am not sure I agree that this should be a cli option. The linked issue talks about testing, but the loading screen is only enabled in dev mode. So why are they testing in development mode?

I understand that it would be nice to add a config option to enable/disable the loading screen, but we should limit the cli options as much as possible and I am not convinced adding one for loading screen is worthwhile.

rinu:
@pimlie Indeed there are reasons to test using development mode. Building and running takes more time and enables some unwanted features. If we must we will migrate to building and running but it's easier to just disable the loading screen. Doing it with a cli option seemed naturual. It would also be possible to create a special env variable for the tests and set the loadingScreen flag in nuxt config according to that. I'd like to see another opinion on that.

Atinux:
Since it's mostly for CI, could an environment variable do the trick?

packages/config/src/options.js
```diff
@@ -369,7 +369,7 @@ export function getNuxtConfig(_options) {
   }
 
   // Add loading screen
-  if (options.dev) {
+  if (options.dev && options.loadingScreen) {
```
pi0: We already have `options.build.indicator` which can be checked here to not be `false`. Adding another option makes no sense.

packages/cli/src/options/common.js
```diff
@@ -44,5 +44,10 @@ export default {
     alias: 'h',
     type: 'boolean',
     description: 'Display this message'
+  },
+  'no-loading-screen': {
```
pi0: Adding a CLI flag for a special requirement is unnecessary. With supporting `build.indicator` option, one can disable it based on environment variables from `nuxt.config`.

pi0 (changes requested):

packages/config/src/options.js
rinu: It seems the build indicator and loading screen are related. With the loading screen disabled I also don't have the build indicator anymore. I was expecting these to be independent features. Can you explain how these are related?

Anyway, turning off the build indicator but leaving the loading screen does currently work. I wouldn't want to make a breaking change.

rinu: Turns out the build indicator uses the websocket server that loading screen sets up.

packages/cli/src/commands/dev.js
```diff
@@ -66,13 +55,32 @@ export default {
...
     // Return instance
     return nuxt
   },
 
+  async listen(nuxt, argv) {
+    // Start listening
+    await nuxt.server.listen()
+
+    // Show banner when listening
+    showBanner(nuxt, false)
```
pimlie: Was the move of showBanner on purpose? I believe `Show banner when listening` is meant to be read as `showBanner after nuxt has started listening` and not `only showBanner when nuxt is listening`. Probably the showBanner call should be placed just after the call to this listen fn in _startDev

packages/config/test/options.test.js
```diff
@@ -237,6 +237,13 @@ describe('config: options', () => {
       expect(consola.warn).toHaveBeenCalledWith('build.extractCSS.allChunks has no effect from v2.0.0. Please use build.optimization.splitChunks settings instead.')
     })
   })
+
+  describe('config: loading screen', () => {
+    test('should add loading screen', () => {
+      const config = getNuxtConfig({ dev: true })
+      expect(config.devModules).toStrictEqual(['@nuxt/loading-screen'])
```
pimlie: Would it be possible to use [expect.arrayContaining](https://jestjs.io/docs/en/expect#expectarraycontainingarray) here? Then if we add a second devModule this test will still succeed.

pimlie (changes requested):
Thanks for working on this!

packages/cli/src/commands/dev.js
rinu: Yes, `showBanner` should happen after `nuxt.server.listen`. These lines should move together. If I moved it into `_startDev` I would have to call `showBanner` in two places after calling `this.listen`.
I struggled to find a good name for `this.listen`. As you pointed out it doesn't just start listening, it also displays the banner and opens the page in the browser. These should all be in the same method not to duplicate code.

pimlie: Hmm, moving the listen call after the build is a second feature change and I actually missed that yesterday. Although I understand your use-case, I think it could be confusing that a seemingly unrelated toggle of loadingScreen also changes the point where Nuxt starts listening, @pi0 wdyt?

Also, when we would approve this we could just have showBanner print the memory usage on !loadingScreen instead of printing a separate line.

rinu: Calling it `loadingScreen` at this stage can indeed be confusing. I wouldn't really want to call it `buildBeforeListen` or something since it wouldn't make sense to nuxt users. As a nuxt user I want an option to disable the loading screen which also happens to require starting to listen after building has completed.

pi0:
This PR is in the wrong state. Adding an option to disable `loading-screen` module, makes sense but `nuxt dev` CLI is **NOT** meant to be used in CI/Testing environments. If performance is a real concern increase CI resources or disables optimizations like minify.

I think adding an option would be enough. You can create a custom build script that programmatically import `Builder` from `@nuxt/builder` and initiate build (with `dev: true`?) and start listening at any moment. This is not an official solution nor recommended or guaranteed to always work.

packages/cli/src/commands/dev.js
pi0: Please see my latest comment

Atinux:
Should we then close this PR?

rinu:
I still want it implemented. The new option is not yet documented, everything else has been changed as requested. Perhaps the option needs a better name but I can't think of one.

pi0:
Still changes to cli are unnecessary as stated above. For building with special use cases, a custom script should work.

rinu:
Changes to cli have been removed. The only change is a new option that changes when the dev server starts listening.

pi0:
PR is conflicting (so commands/dev.js changes are remaining). I'll try to resolve.

pimlie (commented):
As mentioned before I dont think the move of listen after the build has finished is a favorable solution. Nuxt has started listening _as soon as possible_ for a very long time now. Also I think its strange that when you disable the loading screen its still being added to `devModules` and the only reason its not showing is because you start listening _after_ the build has finished.

Imo a cleaner solution would be to just remove the loading screen from devModules completely and to bring back the `Waiting for server resources...` message which was removed in the initial loading screen pr: https://github.com/nuxt/nuxt.js/pull/5251/files#diff-015dff3d45b52153ef5f24ef104e954bL436

But I will leave it to @Atinux or @pi0 to make a decision about this.

rinu:
@pimlie That was exactly what I did initially. Unfortunately there are issues with that approach. The loading screen is a module that can't be removed, you will run into serveral issues if you try doing that. I was really upset when I discovered that.

For my use case the loading screen itself is not even the issue. I thoght it was which is why I named this PR the way I did and tried removing the loading screen module etc. But what I really need is the server to start listening later, when the actual site is ready to be displayed. If it starts listening too soon requests to the site will just result in a timeout or reading the loading screen, both of these are not desireable. Even if I do start listening later the loading screen module can't be removed because the build indicatior still relies on it.

The timeout issue might be resolvable, haven't looked into it. Without the loading screen module the server started listening and requests made before build finished would hang for an infinite time. My pipeline has a 2 hour time limit which was exceeded. Requests get a response only after build has finished.

Atinux:
Actually, you can remove the loading screen with using a hook (that you can merge with your app config in the tests):
```js
export default {
  hooks: {
    'modules:before': function (nuxt) {
      nuxt.options.devModules = nuxt.options.devModules.filter((name) => name !== '@nuxt/loading-screen')
    }
  }
}
```

You can also use the `build:done` hook to know when the build is done and you can start calling your application for e2e testing.

Also, you can also use `nuxt.hook('build:done', () => ...)` syntax.

rinu:
I think I should just make a fresh new PR and close this. We are just not on the same page here.

Atinux:
Sure :)

pi0:
Followed by #6272.

pi0:
[https://github.com/nuxt/nuxt.js/pull/6272](https://trello.com/c/YyN0diFz/32-https-githubcom-nuxt-nuxtjs-pull-6272)

PR #6051, opened 2019-07-09
fix error when use alias for nuxt-edge

clarkdo:
## Types of changes

## Description

Fix CLI failure when use yarn alias `nuxt: "npm:nuxt-edge"`

## Checklist:

### DIFF
--- a/packages/cli/src/run.js
+++ b/packages/cli/src/run.js
@@ -1,20 +1,25 @@
 import fs from 'fs'
-import path from 'path'
 import execa from 'execa'
 import { name as pkgName } from '../package.json'
 import NuxtCommand from './command'
 import setup from './setup'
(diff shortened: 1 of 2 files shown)


packages/cli/src/run.js
```diff
@@ -7,7 +7,7 @@ import setup from './setup'
 import getCommand from './commands'
 
 function checkDuplicateNuxt() {
-  const dupPkg = pkgName === '@nuxt/cli' ? 'cli-edge' : 'cli'
```
clarkdo: `@nuxt/cli` will be replaced to `@nuxt/cli-edge` by rollup, so `pkgName === '@nuxt/cli'` is always true

pi0 (dismissed):

PR #6055, opened 2019-07-10
feat(vue-app): support `app/router.scrollBehavior.js` and deprecate `scrollBehavior`

atinux:
This PR allows to overwrite application and modules templates with the `app/` directory directly.

## Types of changes

## Description

Not really a feature, but a refactor to move the router `scrollBehavior` to `router.scrollBehavior.js`.

**Why?**

In order to stop serializing function and start opening another way of changing the scrollBehavior with a file.
It also avoid restarting the whole Nuxt application since it's actually a config related to the Vue app.

**Before:**

```js
// nuxt.config.js
export default {
  router: {
    scrollBehavior: function (to, from, savedPosition) {
       return { x: 0, y: 0 }
    }
  }
}
```

**After:**

```js
// app/router.scrollBehavior.js
export default function(to, from, savedPosition) {
  return { x: 0, y: 0 }
}
```

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1472)


galvez:
YES, thank you -- I was getting close to going down the `acorn` rabbit hole just to write us a working serializeFunction. It's definitely much better to have these imported as live JavaScript instead.

Atinux:
I plan to do the same logic to most of the configuration option that have to be serialised (head, transition, etc) 😄

packages/config/src/options.js
```diff
@@ -35,6 +35,12 @@ export function getNuxtConfig (_options) {
     options._routerBaseSpecified = true
   }
 
+  // TODO: Remove for Nuxt 3
+  // router.scrollBehavior -> app/router.scrollBehavior.js
+  if (options.router && typeof options.router !== 'undefined') {
+    consola.warn('`router.scrollBehavior` property is deprecated in favor of using `~/app/router.scrollBehavior.js` file')
```
pi0: Can we probably guide users to the new docs for quick migration?

pi0 (dismissed):
Neat :+1:

packages/config/src/options.js
Atinux: Yes,  I will make a PR to the docs + link in the warn for it :)

Atinux:
I need to add support for watching `app/**/*.(js|vue)` when creating or removing file there.

Atinux:
Watching `app/` is done, I also added `dir.app` key in config

Atinux:
We can now overwrites custom templates given in `build.templates` with `app/${dst}`.

Example of a module adding `press/common/pages/source.vue` with `this.addTemplate()`, the user could overwrite it by adding `app/common/pages/source.vue`.

Nuxt will automatically watch the `app/` dir and handle file added/updated/removed for it so the module author won't have to deal with watchers.

galvez (dismissed):

pimlie (commented):
Is `app` discriminatory enough to be used for overwriting the nuxt-app / templates? Its such a common name that I can imagine people are already using it in their projects.

Atinux:
@pimlie actually it was already doing this since a long time (can't remember the version). Actually i found the name perfect to overwrite our default `vue-app` template (and also modules like NuxtPress: `app/press/common/pages/source.vue`)

pimlie (dismissed):
Ok, thanks :)

packages/config/src/options.js
```diff
@@ -35,6 +35,12 @@ export function getNuxtConfig (_options) {
     options._routerBaseSpecified = true
   }
 
+  // TODO: Remove for Nuxt 3
+  // router.scrollBehavior -> app/router.scrollBehavior.js
+  if (options.router && typeof options.router !== 'undefined') {
```
kevinmarrec: Seems the warning would display whenever `router` is defined, I think it should check if `scrollBehavior` is defined too.

kevinmarrec (changes requested):
Great work ! 

See my comment for the little requested change.

Aside that, I think we may need a TypeScript typing change around https://github.com/nuxt/typescript/blob/master/packages/types/config/router.d.ts that may exclude `scrollBehavior` from `RouterOptions` so that people doesn't have autocompletion on obsolete option.

It also made me notice that we don't have configuration types for `dir` 🤦🏼😁

packages/config/src/options.js
Atinux: Fixed, good catch @kevinmarrec

Atinux:
Good catch for typings Kevin, I think it's fine if they have autocomplete and then Nuxt warning 🤔

pi0 (approved):

packages/builder/src/builder.js
```diff
@@ -405,42 +405,47 @@ export default class Builder {
...
+        typeof t === 'object' ? t : undefined
+      )
+    })
     const customTemplateFiles = this.options.build.templates.map(
       t => t.dst || path.basename(t.src || t)
     )
+    const templatePaths = uniq([
+      // Modules & user provided templates
+      // first custom to keep their index
+      ...customTemplateFiles,
```
pi0: Shouldn't they come second? As they should override built-ins? :thinking:

kevinmarrec (approved):

pi0:
2.9.0 notes updated to mention

manniL:
I really like the fact that we avoid serialization with that step, however (as briefly discussed with @Atinux), this means we can't provide data from the `nuxt.config.js` inside such a external file in `app` (at the moment at least). For `scrollBevahior`, this might not have a huge impact, however for further split ups of things like transitions or `head` it could create a few problems if we don't handle it. :relaxed:

PR #6078, opened 2019-07-16
fix(config): add standalone option to NuxtConfigurationBuild

aldarund:


kevinmarrec (approved):
👍 but `standalone` seems to be missing in docs

PR #6087, opened 2019-07-18
refactor(babel-preset-app): simplify babel preset config

clarkdo:
~~This is a breaking change for manually babel config user.~~

Changes:

- remove `buildTarget` and `modern`, the value if from babel api now. https://babeljs.io/docs/en/config-files#apienv which is bound to webpack build (client, server, modern)
- change param of `build.babel.presets ` from `isServer` to `envName` which supports client, server, modern


manniL:
if it's a breaking change, should we move it to Nuxt 3 then?

clarkdo:
@manniL That's why this is a draft pr 😄

clarkdo:
I can make this pr be compatible with 2.x, will change it later

packages/babel-preset-app/README.md
```diff
@@ -69,14 +67,16 @@ There are [detailed docs](https://babeljs.io/docs/en/babel-preset-env#options) f
 export default {
   build: {
     babel: {
-      presets({ isServer }) {
+      presets({ envName }) {
+        const envTargets = {
```
dbachko: @clarkdo should `client` and `server` be swapped?

clarkdo: Correct, thanks.

pi0 (dismissed):

pi0:
@clarkdo please merge if you wish :)

clarkdo:
@pi0 Thanks, I’ll do more tests

packages/babel-preset-app/src/index.js
```diff
@@ -76,23 +73,34 @@ module.exports = (context, options = {}) => {
     corejs = { version: Number(corejs) }
   }
 
-  let { targets } = options
-  if (modern === true) {
-    targets = { esmodules: true }
-  } else if (targets === undefined && typeof buildTarget === 'string') {
-    targets = buildTarget === 'server' ? { node: 'current' } : { ie: 9 }
+  const defaultTargets = {
+    'server': { node: 'current' },
```
pi0: I think keys do not need `'` :)

packages/babel-preset-app/README.md
```diff
@@ -98,12 +98,12 @@ yarn add --dev core-js@2 @babel/runtime-corejs2
 export default {
   build: {
     babel: {
-      presets({ isServer }) {
+      // envName: server, client, modern
+      presets({ envName }) {
         return [
           [
-            require.resolve('@nuxt/babel-preset-app'),
+            '@nuxt/babel-preset-app',
```
pi0: Not sure why it was required as before. I hope it is not breaking NPX.

packages/webpack/src/config/base.js
```diff
@@ -80,24 +80,22 @@ export default class WebpackBaseConfig {
...
-      require.resolve('@nuxt/babel-preset-app'),
-      {
-        buildTarget: this.isServer ? 'server' : 'client'
-      }
-    ]
+    const defaultPreset = [ require.resolve('@nuxt/babel-preset-app'), {} ]
 
     if (typeof options.presets === 'function') {
-      options.presets = options.presets({ isServer: this.isServer }, defaultPreset)
+      // TODO: remove isServer in Nuxt 3
```
pi0: Some may prefer this convention (as it is also in other ideas) so maybe worth having `isServer`/`isClient`/`isModern` with nuxt 3 or deprecate them everywhere?

packages/babel-preset-app/README.md
clarkdo: The difference is the resolving happens in `node_modules/@babel/..` instead of `nuxt.config.js`, I think they should have same resolving result in NPM, how do you think ?

pi0: Me too. As long as explicit dependencies are there (and webpack has explicit dependenc on babel-preset-app) so at least _seems_ to be safe :)

packages/webpack/src/config/base.js
clarkdo: My thought was that `isServer` in current babel is for user to differ babel config in different builds(server/client), with this pr, user doesn't need to check it in most cases since `preset-app` has done it automatically via `envName`.

You comment also makes sense, we can add `isClient, isModern`, but do we have envName as arg as well ?

pi0: Don't know. That's totally something optional and personally even i like the new `envName` more! Tought if it does not add much complexity or perf concerns, we can have both :)

clarkdo: Cool, I'll add them all

clarkdo: Added, I'll update readme and doc tonight.

pi0 (approved):

Issue #2421, opened 2017-12-19
Support redirect to route name in middleware


pSchaub:
In a middleware we can do a redirect with e.g. `redirect('/login')`.

But how can we redirect based on a route name?

`redirect({name: 'login'})` is currently not possible.

I think it's a bad idea to do redirects on absolute paths instead of route names.

PR #6095, opened 2019-07-19
fix(types): add type definition for redirect

ysfada:
## Types of changes

## Description
adds missing type definition for redirect by route name

#2421
#2432


kevinmarrec (changes requested):
Thanks for the PR @ysfada , here is my quick review :

`RawLocation` is of type `string | Location`

It may conflicts with `redirect(path: string, query?: Route['query']): void`
I would suggest to change `RawLocation` by just `Location`, so that the new added signature for `redirect` only deal with a `Location` object.

kevinmarrec (approved):

atinux (approved):

pi0:
Thanks for this PR @ysfada . It is fixed in 2.9 upcoming [typescript repo](https://github.com/nuxt/typescript/blob/master/packages/types/app/index.d.ts)

PR #6120, opened 2019-07-24
feat(webpack): allow function entries for `build.transpile`

atinux:
## Types of changes

## Description

It's mostly a feature for module, thinking of `@nuxt/http` for example that need to add `ky` in `build.transpile` but only for legacy mode.

### Before:

```js
{
  build: {
    transpile: ['ky']
  }
}
```

### After:

```js
{
  build: {
    transpile: [
      ({ isModern }) => isModern ? null : 'ky'
    ]
  }
}
```

The main issue I see is that module author will have to check Nuxt version to be able to push a function in `build.transpile` otherwise it justs breaks 😢 

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1472)


pi0 (dismissed):

pi0:
> The main issue I see is that module author will have to check Nuxt version to be able to push a function in build.transpile otherwise it justs breaks cry

This is a feat/fix by next release. Nothing will be breaking.

---

Some other enhancements we could do is adding an option like `transpileLegacy` or allow object form like `{ name: 'ky', legacy: true }` or `{ name: 'ky', modern: false }`

galvez (dismissed):

Atinux:
I also thought of keeping a string directly and do something like `legacy:ky`, WDYT?

packages/webpack/src/config/base.js
```diff
@@ -57,10 +58,13 @@ export default class WebpackBaseConfig {
...
     const items = [/\.vue\.js/i]
-    for (const pattern of this.buildContext.buildOptions.transpile) {
+    for (let pattern of this.buildContext.buildOptions.transpile) {
+      if (typeof pattern === 'function') {
+        pattern = pattern(this.nuxtEnv)
+      }
       if (pattern instanceof RegExp) {
         items.push(pattern)
-      } else {
+      } else if (pattern) {
```
clarkdo: Should this be string ?

clarkdo (commented):
If function is one of array items, then it can only return string or regex pattern.

Do you think is there any use case for multiple functions, maybe modules ?

How about just making `transpile` as function, like:

```js
transpile: () => {
  return [
    'vue-test',
    '@vue/test'
  ]
}
```

packages/webpack/src/config/base.js
Atinux: Yes, Will update to check `typeof pattern === 'string'`

Atinux:
@clarkdo Making `build.transpile` a function will be harder for modules author to extend it I guess, this is why I thought about handling a push as a method (see https://github.com/nuxt/http/commit/2ce08de2ecceda27d362982eebdf0f15470f7fee#diff-d1234a869b3d648ebfcdce5a76747d71R100)

Atinux:
Ready to review

packages/webpack/src/config/server.js
```diff
@@ -20,10 +20,13 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
     const whitelist = [
       /\.(?!js(x|on)?$)/i
     ]
-    for (const pattern of this.buildContext.buildOptions.transpile) {
+    for (let pattern of this.buildContext.buildOptions.transpile) {
```
manniL: Looks like we have a duplicate code block here (as in `src/config/base.js`, L61-70). Could we extract it somehow?

packages/webpack/src/config/base.js
```diff
@@ -57,10 +58,13 @@ export default class WebpackBaseConfig {
...
     const items = [/\.vue\.js/i]
-    for (const pattern of this.buildContext.buildOptions.transpile) {
+    for (let pattern of this.buildContext.buildOptions.transpile) {
+      if (typeof pattern === 'function') {
+        pattern = pattern(this.nuxtEnv)
+      }
       if (pattern instanceof RegExp) {
         items.push(pattern)
-      } else {
+      } else if (typeof pattern === 'string') {
```
clarkdo: How about giving a warning message if pattern is not string or regexp ?

packages/webpack/src/config/server.js
clarkdo: Could be, but please note the difference in `path.normalize`

clarkdo (approved):

packages/webpack/src/config/server.js
clarkdo: I will do it in another pr since this has been a long time

PR #6134, opened 2019-07-26
feat(vue-renderer): support prepend/append body tags during ssr for all allowed tag types

### LINKED ISSUE #6097: Appending scripts and styles to the body not work on initial page load
### Version

### Reproduction link

[https://codesandbox.io/embed/codesandbox-nuxt-rodbc](https://codesandbox.io/embed/codesandbox-nuxt-rodbc)

### Steps to reproduce

Click on the button and then go back. You will see that component inline style and script tags are loaded only on hot reload.

### What is expected ?

Styles and scripts that are defined in head() method should be appended to the body on initial page load & generate.

Styles and scripts that are defined in head() method are appended to the DOM only during SPA navigation.

### Additional comments?

Thank you for resolving the issue.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9515">#c9515</a>)</em></sub></div>

pimlie:
~~This is a draft until #6121 is merged~~

Resolves: #6097 

## Types of changes

## Description

According to the html5 specs both `meta`, `link`, `style`, `script` as `noscripts` are allowed in the body. This was already support by VueMeta (thus Nuxt) on the client, but during SSR it wasnt yet.

Note, some tags are only allowed with certain attributes. Eg link _must_ have a `itemprop` attribute, a link with a rel attribute is not allowed in the body. Its up to the user to use this features according to the spec.

## Checklist:


packages/vue-renderer/src/renderers/spa.js
```diff
@@ -78,11 +79,23 @@ export default class SPARenderer extends BaseRenderer {
       m.script.text() +
       m.noscript.text()
```
clarkdo: text has been called here, so does .text() without arg only return pbody: false data?

pimlie: [Indeed](https://github.com/nuxt/vue-meta/blob/master/src/server/generators/tag.js#L33)

pimlie:
@clarkdo @pi0 Do you have any idea why the SPARenderer is being loaded on `mode: 'ssr'` ?

The basic.test fails at the moment due to my latest commit. The basic fixture is _not_ a SPA build but the SPARendered is still being initiailzed. Because the SPARenderer is initialized before the SSR/ModernRenderer, VueMeta is loaded on the server but with the wrong config. Subsequent loads of VueMeta within the SSR App are ofc ignored because VueMeta is already loaded.

clarkdo:
It makes switch SPA and SSR at runtime possible.

pimlie:
Yes, it uses the plugin options throughout its lifetime.

I have reverted the last commit as that wont ever work probably until v3. I either need to add support to VueMeta to make it possible to override options on calling vm.$meta.inject or Nuxt should do a search and replace in the SPA renderer. Both are not ideal ofc but probably the first is a bit cleaner.

Or maybe you have other/better suggestions?

clarkdo:
I think you can check the mode config when passing ssrAppId and only do it when the value is spa, is that ok?

pimlie:
That would only be a partial fix I believe. Eg if somebody would switch during runtime from SSR to SPA then their SPA pages would still be generated with the wrong app id.

clarkdo:
@pimlie I think swiching at runtime will re-run vue.use(VueMeata)

pimlie:
I have added support to VueMeta for directly generating tags from a metaInfo object. This means this PR is depending on #6153 for VueMeta v2.2

Creating the tags renderer (the object containing all the meta.text() stuff) is now _at least_ two times faster. Before it took ~8-12ms, now it takes ~4-5ms

Atinux:
> Before it took ~8-12ms, now it takes ~4-5ms

🔥

packages/vue-renderer/src/renderers/spa.js
```diff
@@ -13,27 +12,21 @@ export default class SPARenderer extends BaseRenderer {
...
     this.cache = new LRU()
 
-    // Add VueMeta to Vue (this is only for SPA mode)
-    // See app/index.js
-    Vue.use(VueMeta, {
+    this.vueMetaConfig = {
       keyName: 'head',
       attribute: 'data-n-head',
       ssrAttribute: 'data-n-head-ssr',
+      ssrAppId: '1',
```
pi0: Number (`1`) implies multiple SSR apps are possible. Is that a case?

packages/vue-renderer/src/renderers/spa.js
```diff
@@ -13,27 +12,21 @@ export default class SPARenderer extends BaseRenderer {
...
 
-  async getMeta () {
-    const vm = new Vue({
-      render: h => h(), // Render empty html tag
-      head: this.options.head || {}
-    })
-    await this.vueRenderer.renderToString(vm)
-    return vm.$meta().inject()
+  getMeta () {
+    return VueMeta.generate(this.options.head || {}, this.vueMetaConfig)
```
pi0: What do you think renaming `getMeta` to `generateMeta` for consistency?

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -140,12 +151,17 @@ export default class SSRRenderer extends BaseRenderer {
...
+    APP += m.meta.text({ body: true })
+    APP += m.link.text({ body: true })
+    APP += m.style.text({ body: true })
     APP += m.script.text({ body: true })
     APP += m.noscript.text({ body: true })
 
     // Template params
     const templateParams = {
-      HTML_ATTRS: 'data-n-head-ssr ' + m.htmlAttrs.text(),
+      HTML_ATTRS: m.htmlAttrs.text(true),
```
pi0: Can we annotate param like `true /* prepend */`

packages/vue-renderer/src/renderers/spa.js
pimlie: Actually the `'1'` implies that we expect that the Nuxt SPA app is always the first APP on the page (I made it a string on purpose as a small hint).

The term `ssrAppId` might be a bit confusing here, because in SPA mode there is no ssr app but we still need to make sure the head config from nuxt.config is recognised once the page is loaded on the client.

pi0: Hmm why not simply naming it `nuxt` or `main` (like `__nuxt`)

packages/vue-renderer/src/renderers/spa.js
pimlie: I thought about just removing that function but wasnt sure if that would be a breaking change. Not sure if its programmatically possible to extend SPARenderer and register your own. If not them I am happy to change it into whatever you like :)

pi0: Unless if we had docs or typescript which I don't think we have. That's an internal breaking change. /cc @clarkdo are you up to this change?

packages/vue-renderer/src/renderers/spa.js
pimlie: That would be fine if I would know there would only ever be one app using VueMeta on the page, but unfortunately I dont know that.

As mentioned in my edit, this boils down to a hydration issue as I cant reliably pass an appId from ssr to the client. We could maybe support it for Nuxt if we just add an additional attribute in the app template.

Fyi, [this](https://github.com/vuejs/vue/blob/e90cc60c4718a69e2c919275a999b7370141f3bf/packages/vue-server-renderer/basic.js#L8757) is were Vue sets the data-server-rendered attribute and afaik I cant do something similar (through Vue)

packages/vue-renderer/src/renderers/ssr.js
manniL: I'd favor a variable like `shouldPrepend`

pimlie: The `true` here doesnt depict prepend / append but whether to automatically add the [`ssrAttribute`](https://vue-meta.nuxtjs.org/api/#ssrattribute) or not

packages/vue-renderer/src/renderers/spa.js
pimlie: Have removed the getMeta function (doesnt seem to be possible to override the SPArenderer as the renderers are hardcoded loaded in renderer.js)

packages/vue-renderer/src/renderers/spa.js
pimlie: Let me know if you want me to take the route of adding a vue-meta specific attribute on the APP div. Thinking about it, it should be possible to just remove that ssrAttribute.

pi0 (approved):

PR #6139, opened 2019-07-26
feat: add optional chaining and null coalescing operator proposals

TheAlexLichter:
## Types of changes

## Description

Add the two new stage 3 proposals to Nuxt by default via Nuxt's babel preset.
* https://babeljs.io/docs/en/babel-plugin-proposal-nullish-coalescing-operator
* https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining.html

## Checklist:

### DIFF
--- a/packages/babel-preset-app/package.json
+++ b/packages/babel-preset-app/package.json
@@ -13,6 +13,8 @@
     "@babel/core": "^7.5.5",
     "@babel/plugin-proposal-class-properties": "^7.5.5",
     "@babel/plugin-proposal-decorators": "^7.4.4",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
+    "@babel/plugin-proposal-optional-chaining": "^7.2.0",
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@babel/plugin-transform-runtime": "^7.5.5",
     "@babel/preset-env": "^7.5.5",
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -121,6 +121,8 @@ module.exports = (context, options = {}) => {
 
   plugins.push(
     require('@babel/plugin-syntax-dynamic-import'),
+    require('@babel/plugin-proposal-optional-chaining'),
+    require('@babel/plugin-proposal-nullish-coalescing-operator'),
     [require('@babel/plugin-proposal-decorators'), {
       decoratorsBeforeExport,
       legacy: decoratorsLegacy !== false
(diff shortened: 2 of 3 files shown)


manniL:
Open question: Should be set `loose` to true for both by default? :thinking:

See https://babeljs.io/docs/en/babel-plugin-proposal-nullish-coalescing-operator#loose

clarkdo:
I’m not sure if they should be introduced as we never introduce stage 3(candidate) proposals into core

manniL:
@clarkdo 	we never did by now, right. But what reason would speak against it?

clarkdo:
I think it’s the same reason as babel-preset-env doesn’t support any stage-x proposal since proposals can continue to change before landing in browsers.

clarkdo:
And we should find a good balance between development experience  and performance (size and time),  so following standard may be appropriate.

clarkdo:
A good point from Babel and TC39 https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets

manniL:
The blog post from the Babel team was kinda enlightening :+1: 	

Maybe we should provide an easy way to add it then, e.g. through a nuxt module? :thinking: 	(So ppl who need it can do it easily and w/o actually copy and pasting definitions in all projects?)

clarkdo:
Plugins can be easily added by build.babel,plugins

manniL:
True, but you still have to add the dependencies (of "all the proposals" you wanna add). However, this seems like it's not a huge effort in the end :relaxed: 	

Closing here :+1:

clarkdo:
Update: optional chaining and null coalescing have been in stage-4 and @babel/preset-env will include them in short future https://github.com/babel/babel/pull/10811

mathieutu:
Hi @clarkdo.

I've a brand new nuxt app, with @babel/preset-env:7.9.6, and I've a syntax error when using a optional chainin: 
```
Syntax Error: Unexpected token (1:112)                                                                                                                                 friendly-errors 20:27:39

                                                                                                                                                                       friendly-errors 20:27:40
 @ ./components/form/SubmitButton.vue?vue&type=template&id=f475626c& 1:0-222 1:0-222
 @ ./components/form/SubmitButton.vue
 @ /Users/mathieutu/Projects/dsecurite/node_modules/babel-loader/lib??ref--2-0!/Users/mathieutu/Projects/dsecurite/node_modules/vue-loader/lib??vue-loader-options!./pages/login.vue?vue&type=script&lang=js&
 @ ./pages/login.vue?vue&type=script&lang=js&
```

clarkdo:
@mathieutu Can you provide more details like a reproducible repo?

mathieutu:
I'll do a reproducible repo. For now I've just quickly replaced it with the standard `foo && foo.bar`. Actually, I think the problem is only in template. I remember using optional and coallescing in computed.

clarkdo:
Ah, do you mean in template tag?

mathieutu:
Ok, so nevermind, and thanks for the quick answer!

PR #6145, opened 2019-07-27
chore: turn on no-console and no-debugger

clarkdo:
## Types of changes

## Description

Turn on `no-cosnole` and `no-debugger` as they've been turned off in exlint-config-nuxt which may make sense to nuxt user, but they should be enabled for nuxt itself.

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,10 @@ module.exports = {
   "globals": {
     "BigInt": true
   },
+  rules: {
+    'no-console': 'error',
+    'no-debugger': 'error'
+  },
   overrides: [{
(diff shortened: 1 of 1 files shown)


manniL:
They should depend on the env though through eslint-config-nuxt :thinking:

clarkdo:
This config is only for nuxt contributors, nuxt is a framework instead web app (heavy usage in dev debugging) and we are using consola, so I think the rules can be more strict.

Actually it has been turned on long time ago.

TheAlexLichter (approved):

manniL:
Ah yeah, it doesn't cover consola at all, you are right :+1:

aldarund (approved):

PR #6149, opened 2019-07-28
feat: optionally strip non-client or non-server code

TheAlexLichter:
Types of changes

## Description

We already have `process.client` and `process.server` that take care of separating "client" and "server" code (if needed). However, it can sometimes be hard to control that in 3rd party dependencies.

With this change, the webpack define plugin will set the `typeof window` check to  `object` on client and `undefined` on server, so the code can (and will be) correctly removed by webpack.

## Checklist:


clarkdo:
What will happen to jsdom case？

manniL:
Made it optional but couldn't find a better name for it :see_no_evil:

packages/config/src/config/build.js
```diff
@@ -57,6 +57,7 @@ export default () => ({
   plugins: [],
   terser: {},
   hardSource: false,
+  aggressiveClientServerCodeRemoval: false,
```
clarkdo: How about aggressiveCodeRemoval ?

manniL: I thought it'd be too unspecific, but I'm fine with that :relaxed:

packages/webpack/src/config/client.js
```diff
@@ -30,13 +30,17 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
...
+    return Object.assign(
+      super.env(),
+      {
+        'process.env.VUE_ENV': JSON.stringify('client'),
+        'process.browser': true,
+        'process.client': true,
+        'process.server': false,
+        'process.modern': false
+      },
+      ...(this.buildContext.buildOptions.aggressiveClientServerCodeRemoval && { 'typeof window': JSON.stringify('object') })
```
clarkdo: I think this can be moved to `base.js::env`:

```js
if(this.buildContext.buildOptions...) {
	env['typeof window'] = this.isServer
}
```

clarkdo (dismissed):

packages/config/src/config/build.js
clarkdo: A doc would cover this, I just thought the shorter name will be better for usage.

packages/webpack/src/config/client.js
manniL: Couldn't `	env['typeof process'] = this.isServer` possibly remove valid code if it runs before webpack define plugin?

clarkdo: I didn’t get you, can you give me example?

manniL: Nvm, I thought twice about the use case I had in mind and it'll work nevertheless :relaxed:

pi0 (approved):

galvez (approved):

PR #6150, opened 2019-07-28
feat: add time-based build and generation hash for every nuxt build

TheAlexLichter:
## TODO

- [ ] Dev mode
- [x] Verify it works with `nuxt generate --no-build`

## Types of changes

## Description

This PR adds two new hash variables (`process.BUILD_HASH` and `process.EXPORT_HASH`). 

The build hash differs per `nuxt build` (and therefore also per `nuxt generate` *with rebuild*) and indicates changes to the code.

The export hash will change only when re-generating the Nuxt project via `nuxt generate` and **also works with** `nuxt generate --no-build`. However, it is only available from the generation process on and *might not be available* (?) in the final app itself.

The idea originated a few months ago when @hootlex suggested a build token feature. :relaxed: 	

These hashes could be used for various reasons, but especially for `export-based` caching of resources that aren't touched by webpack and have a common name otherwise. This will also be the case for the upcoming full static mode ( https://github.com/nuxt/rfcs/issues/22 ), where payloads can't be processed by webpack but caching of them would be a great benefit. 

## Checklist:


packages/webpack/src/config/base.js
```diff
@@ -18,6 +19,8 @@ import WarningIgnorePlugin from '../plugins/warning-ignore'
 
 import { reservedVueTags } from '../utils/reserved-tags'
 
+const BUILD_HASH = crypto.createHash('md5').update(new Date().toISOString()).digest('hex').substr(0, 20)
```
clarkdo: Different builds will be same hash in dev reloading.

clarkdo (commented):
Do we consider that two builds with nothing changed?

packages/webpack/src/config/base.js
manniL: That's right, haven't thought about dev at all here. Do you think it'd be clever to refresh the build hash on dev?

manniL:
That'd mean that e.g. the API content has changed (for example when re-generating a page with `--no-build'). Seems like a valid use case to me :relaxed:

packages/webpack/src/config/base.js
clarkdo: If user have any code differing from builds, it should work for both dev and production, or they can’t verify the functionality in dev

clarkdo:
It depends on what is this hash used for.

In current or change, every build will be a different build since they have different hash

manniL:
@clarkdo 	I see. Would it make sense to split it up in `BUILD_HASH` and `GENERATION_HASH` or so?

clarkdo:
For now we can leave it as your current change and see what the future feedback will be like, then apply more changes.

Thanks for the great idea!

manniL:
@clarkdo I had to split it up anyway, as the build hash doesn't change on `nuxt generate --no-build` and overriding the build hash could lead to logic errors where on would assume another hash (as WP define plugin fully replaces occurrences as you know)

packages/webpack/src/config/base.js
```diff
@@ -10,14 +10,16 @@ import TerserWebpackPlugin from 'terser-webpack-plugin'
...
-import { isUrl, urlJoin } from '@nuxt/utils'
+import { createMd5HashFromString, isUrl, urlJoin } from '@nuxt/utils'
 
 import PerfLoader from '../utils/perf-loader'
 import StyleLoader from '../utils/style-loader'
 import WarningIgnorePlugin from '../plugins/warning-ignore'
 
 import { reservedVueTags } from '../utils/reserved-tags'
 
+const BUILD_HASH = createMd5HashFromString(new Date().toISOString())
```
clarkdo: Can be moved to util as createTimeHash

packages/utils/src/crypto.js
```diff
@@ -0,0 +1,7 @@
+import crypto from 'crypto'
+
+export const createMd5HashFromString = (string, length = 20) => {
+  return crypto.createHash('md5').update(string).digest('hex').substr(0, length)
+}
+
+export const createTimeHash = () => createMd5HashFromString(new Date().toISOString())
```
pi0: Why hashing date in an ireversable way? By using a timestamp we can compare/sort builds by date.

packages/webpack/src/config/base.js
```diff
@@ -10,14 +10,16 @@ import TerserWebpackPlugin from 'terser-webpack-plugin'
...
-import { isUrl, urlJoin } from '@nuxt/utils'
+import { createTimeHash, isUrl, urlJoin } from '@nuxt/utils'
 
 import PerfLoader from '../utils/perf-loader'
 import StyleLoader from '../utils/style-loader'
 import WarningIgnorePlugin from '../plugins/warning-ignore'
 
 import { reservedVueTags } from '../utils/reserved-tags'
 
+const BUILD_HASH = createTimeHash()
```
pi0: I think this would be better to at least set inside constructor.

packages/generator/src/generator.js
```diff
@@ -47,6 +54,8 @@ export default class Generator {
     // Wait for nuxt be ready
     await this.nuxt.ready()
 
+    process.EXPORT_HASH = createTimeHash(new Date().toISOString())
```
pimlie: no arg needed

packages/webpack/src/config/base.js
clarkdo: Move into env ?

packages/generator/src/generator.js
pi0: This util does not accept an argument (time). Also what do you think to first attempt reading from builder instance to have a consistent hash?

packages/webpack/src/config/base.js
manniL: won't the hash be different for server and client then?

packages/utils/src/crypto.js
pimlie: from a crypto point of view it would probably be better to prefix the time hash with a nuxt-unique seed. That way the hash is really unique

pimlie (changes requested):

packages/generator/src/generator.js
manniL: You mean setting the export hash equal to the build hash and only have it differently on `nuxt generate --no-build`?

packages/webpack/src/config/base.js
clarkdo: Right, but we need a way to generate new hash for different builds

packages/utils/src/crypto.js
manniL: Right, that would be possible. But that would also mean that every Nuxt application will have the timestamp of the last build/generate automatically inserted. Not sure if that's a good idea

packages/webpack/src/config/base.js
manniL: won't the hash be different for server and client then?

packages/utils/src/crypto.js
manniL: Totally right. Wasn't sure if that's necessary or if we even should have an "app-specific" seed to.

packages/webpack/src/config/base.js
manniL: Oh, so you would have different hashes for server and client *intentionally*?

clarkdo: No, I mean your concern is right, we need to find some way else

Atinux:
I think it should not be done through Webpack finally but from the generating part of the HTML, and then inserting a script with the data (like window.__NUXT__) that Nuxt will use in static mode to know the last hash when generated for the payloads.

packages/webpack/src/config/base.js
pi0: Builder instance is shared.

manniL: Ah, so you want to add it in the Builder itself instead of the webpack config?

packages/utils/src/crypto.js
pi0: Hmm what's the concern of not exposing timestamp of last build? Hash is purely based on time. Except that we obfuscate it for an unknown reason. But can sometimes be useful.

packages/webpack/src/config/base.js
pi0: Yep :) Inside build class constructor. So we can also use it by generator.

packages/utils/src/crypto.js
manniL: I'm a huge fan privacy by design and I'd consider the last build time of a website as a partially sensible data point. You could infer a lot from that, not only because most of the builds happen automatically after a commit.

But we could give the user the option override the "hash" (or build "token") creation method so every user can decide on their own

packages/webpack/src/config/base.js
manniL: @pi0 Sure, we can do that. But this still wouldn't solve the `nuxt generate --no-build` use case, right?

packages/utils/src/crypto.js
pi0: File creation time can be also exposed by the other ways (like last modified header of static generated HTML files)

While seeing the point, I don't think that is really a (at least security) concern.

packages/webpack/src/config/base.js
pi0: Yeah. Unless storing in a manifest file after build and restore on generate --no-build

manniL: That'd work in terms of sharing, but then the hashes wouldn't be different  (and we'd have to keep relying on the export hash)

packages/utils/src/crypto.js
pimlie: Agree with manni, privacy by design means only disclosing information when you really need to. What might not be interesting today could be that one essential bit of information needed tomorrow. File created/modified times could already be masked by eg a recursive touch, so its also all about having a choice.

packages/webpack/src/config/base.js
pi0: Yep. --no-build is not relative to this change request. Maybe we should use a fresh hash on each generate.

packages/utils/src/crypto.js
pi0: In this case, privacy/security is overseen. By default, build dates are already exposed. If that is a privacy concern could be discussed separately. 

image

I'm OK with adding option but not hashing by _default_ makes builds easier to understand and maintain without exposing (at least a new) concern.

packages/generator/src/generator.js
```diff
@@ -47,6 +54,8 @@ export default class Generator {
     // Wait for nuxt be ready
     await this.nuxt.ready()
 
+    process.EXPORT_HASH = createTimeHash()
```
pi0: This is making side-effects. IE: Two generator instances in the same process will conflict.

manniL: Could you elaborate on that?

pi0: We are using a shred reference inside class to store state. This is not a good practice. Please review other comments regarding storing hash in builder instance (and inheriting by generator / generate one if no buolder provided)

manniL:
Stale

PR #6159, opened 2019-07-30
feat: options.target and full-static export

atinux:
## Types of changes

## Description

Introducing a `target` option in `nuxt.config.js`, available values:
- `'server'` (default)
- `'static'`

I also add the information in the Nuxt.js banner:
Screenshot 2019-07-31 at 17 31 24

Add `process.target` in the Vue app + update `process.static` to use `target === 'static'` so it works in dev mode.

- With CLI: `nuxt <command = dev> <dir = '.'> [--target='server']`

This option will allow to:
- **static**: Enhance the developer experience when running `nuxt dev`:
  - Remove `req` & `res` from context
  - Fallback to client-side on 404 / error / redirect
  - [x] Dissociate `nuxt build` & `nuxt generate` -> creating `nuxt export` to only pre-render routes
  - [x] Generate a payload for not-excluded routes + EXPORT_TIME (if `generate.static = true`)
  - [x] Crawl every dynamic pages (if `generate.crawler = true`)
  - [x] Handle `Page.static = false` or `$nuxt.isPreview`
  - [x] Force `$route.query` to be empty on SSR to mimic `nuxt export`

- [ ] Check side effects of `created` vs `beforeCreate` for `fetch`

## Checklist:


packages/vue-renderer/src/renderer.js
```diff
@@ -269,13 +269,15 @@ export default class VueRenderer {
 
     // Add url to the renderContext
     renderContext.url = url
+    // Add target to the renderContext
+    renderContext.target = this.serverContext.nuxt.options.target
 
-    const { req = {} } = renderContext
+    const { req = {}, res = {} } = renderContext
```
manniL: Clever :relaxed:

packages/vue-renderer/src/renderer.js
```diff
@@ -284,6 +286,11 @@ export default class VueRenderer {
       renderContext.modern = modernMode === 'client' || isModernRequest(req, modernMode)
     }
 
+    // renderContext.static (old _generate)
```
manniL: Maybe we should add a TODO here?
```suggestion
    // renderContext.static (old _generate) TODO: Remove in Nuxt 3
```

packages/webpack/src/config/base.js
```diff
@@ -118,7 +122,9 @@ export default class WebpackBaseConfig {
     const env = {
       'process.env.NODE_ENV': JSON.stringify(this.mode),
       'process.mode': JSON.stringify(this.mode),
-      'process.static': this.buildContext.isStatic
+      'process.dev': this.dev,
+      'process.static': this.target === 'static',
```
manniL: I guess it'd make sense to have a utility object or function that covers all possible targets :thinking:

packages/vue-renderer/src/renderer.js
Atinux: 😄

packages/webpack/src/config/base.js
Atinux: I thought about it too, how would you implement it?

manniL: If we don't have sth. like `utils/constants.js`, I'd introduce it and then use a named export for the targets.

Atinux: Will update thanks!
The targets are `server`, `serverless` and `static` :)

husayt:
This looks really interesting. One question. What is the difference between static and serverless?

As I understand new modes are more or less similar to the following:
* generate spa= serverless
* generate universal = static
* build spa/universal =  server

Atinux:
Hey @husayt 

Target is directly linked to the platform you want to host your Nuxt.js App, by default, it's `server` and is a normal node.js server with SSR.

- `server`: Normal node.js hosting (Heroku, Amazon EC2, etc)
- `serverless`: Serverless hosting (Now V2, AWS Lambda, Google Cloud Functions, etc)
- `static`:  Static/CDN hosting (Netlify, GitHub Pages, Surge.sh, Amazon S3, etc)

packages/config/src/options.js
```diff
@@ -88,6 +88,26 @@ export function getNuxtConfig (_options) {
 
   defaultsDeep(options, nuxtConfig)
 
+  // Target
+  options.target = options.target || 'server'
+  if (!Object.values(TARGETS).includes(options.target)) {
+    consola.warn(`Unknown target: ${options.target}. Falling back to server`)
```
pi0:
```suggestion
    consola.warn(`Unknown target: ${options.target}. Falling back to server`)
```

packages/config/src/options.js
```diff
@@ -88,6 +88,26 @@ export function getNuxtConfig (_options) {
...
+    options.target = 'server'
+  }
+
+  // User server middleware warning if target=static
+  if (options.target === TARGETS.static && options.serverMiddleware.length) {
+    consola.warn(`serverMiddleware are not recommended in static target and can lead to side effects`)
```
pi0: Suggestion for better wording: serverMiddleware does _not_ work with a static target.

packages/core/src/module.js
```diff
@@ -45,11 +45,9 @@ export default class ModuleContainer {
       throw new Error('Template src not found: ' + src)
     }
 
-    // Generate unique and human readable dst filename
-    const dst =
-      template.fileName ||
-      path.basename(srcPath.dir) + `.${srcPath.name}.${hash(src)}` + srcPath.ext
-
+    const fileName = template.fileName || template.filename
```
pi0: Neat: Adding an extra line before next comment makes it more readable. Also is it for backward support of both casings or simply DX? Maybe could be good idea to also document it if it is a new feture. A module may use `filename` which is unsupported for < 2.9

packages/vue-app/template/router/routes.json
```diff
@@ -0,0 +1 @@
+<%= JSON.stringify(router.routes, null, 2) %>
```
pi0: HAHA Nice :+1:

packages/vue-app/template/server.js
```diff
@@ -15,9 +15,9 @@ Vue.component('NLink', NuxtLink)
 const noopApp = () => new Vue({ render: h => h('div') })
 
 const createNext = ssrContext => (opts) => {
+  // If static target, render on client-side
   ssrContext.redirected = opts
-  // If nuxt generate
-  if (!ssrContext.res) {
+  if (ssrContext.target === 'static' || !ssrContext.res) {
```
pi0: Can we use `TARGETS` constant here as well?

packages/generator/src/generator.js
```diff
@@ -94,6 +101,10 @@ export default class Generator {
     return routes
   }
 
+  getAppRoutes () {
+    return require(path.join(this.options.buildDir, 'router/routes.json'))
```
pi0: Isn't a `fs.exists` guard necessary?

packages/config/src/options.js
pi0: Also what if we automatically empty this array or disable all serverMiddleware at `@nuxt/server` level to prevent users going wrong?

packages/vue-app/template/server.js
Atinux: Actually it's inside Webpack bundle, I don't know if we should import the constants there? 🤔

packages/core/src/module.js
Atinux: It's mostly for DX since I see myself always writing `filename` instead of `fileName`.

What about documenting `filename` only for Nuxt `v3.X`?

pi0: Hmm more DX taste it is. Other places we always use camelCase :) Maybe we can just add a comment above this that we _do_ support both

packages/vue-app/template/server.js
pi0: It is server bundle and webpack should treeshake esm imports like this. Anyway just a minor comment :)

packages/config/src/options.js
Atinux: Well the point is to discourage user `serverMiddleware`,  but don't remove the modules `serverMiddleware` since they can work in devMode (pushing the `serverMiddleware` only in dev)

pi0: What's the use case of server middleware for full-static mode? If the case is during production generate then IMO they should not be warned/discouraged. If not, why only supporting in dev mode?

pi0: I thought the reason for introducing a new `static` target is giving near to production experience during development so developers don't do something wrong/unsupported.

packages/core/src/module.js
Atinux: Comment added :)

packages/config/src/options.js
Atinux: Indeed, it could be used for production generate as well, thinking of NuxtPress for instance.

Atinux: I will remove the warning and we will just have to add a note in the documentation of the serverMiddleware :)

packages/generator/src/generator.js
Atinux: Well actually the guard is done in `initiate` directly :)

pimlie (dismissed):
If you approve the description changes we should probably also review the docs

packages/generator/src/generator.js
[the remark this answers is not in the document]
pi0: Final dist will always use require. Dybamic import by rollup just wraps this into a Promise.resolve :)

packages/cli/src/commands/export.js
```diff
@@ -0,0 +1,46 @@
+import { TARGETS } from '@nuxt/utils'
+import { common, locking } from '../options'
+import { createLock } from '../utils'
+
+export default {
+  name: 'export',
+  description: 'Export a static generated web application',
```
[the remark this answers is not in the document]
pi0: I believe new command is to avoid breaking changes and something better for name. Being fullStatic or not can be an option for nuxt3

packages/generator/src/generator.js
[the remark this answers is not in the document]
pimlie: Both rollup as webpack seem to discourage it:
https://github.com/rollup/rollup/issues/1058#issuecomment-254187433
https://github.com/rollup/rollup-plugin-commonjs/issues/273#issuecomment-504758287
https://webpack.js.org/api/module-methods/ (top yellow block)

But as it seems there are more places in Nuxt where these two are mixed I guess we can leave it as is.

packages/cli/src/commands/export.js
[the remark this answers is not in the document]
Atinux: Indeed, `nuxt export` is planned to be the new command to export (and will activate by default the full static option), `nuxt generate` will be deprecated since it also run the build which `nuxt export` does not.

Kronhyx (dismissed):

packages/generator/src/generator.js
```diff
@@ -24,15 +25,20 @@ export default class Generator {
 
   async generate ({ build = true, init = true } = {}) {
     consola.debug('Initializing generator...')
-
     await this.initiate({ build, init })
 
-    consola.debug('Preparing routes for generate...')
+    // Paylods for full static
```
indirectlylit:
```suggestion
    // Payloads for full static
```

husayt:
This is the most awaited feature in Nuxt currently. Seems like this is currently on ice waiting for some dependent work to finish to pave the way. 

@Atinux is this dependent on new fetch? That means once fetch is out this could be the next thing. I guess this is not going to make it in v.12. This is all exciting stuff and thank you for all the amazing work, just wanted to understand the roadmap so we can plan accordingly.

packages/cli/src/options/common.js
```diff
@@ -28,6 +28,14 @@ export default {
       }
     }
   },
+  target: {
+    alias: 't',
+    type: 'string',
+    description: 'Build/Start app for different target, e.g. server, serverless and static',
```
manniL:
```suggestion
    description: 'Build/start app for a different target, e.g. server, serverless and static',
```

packages/generator/src/generator.js
```diff
@@ -24,15 +25,20 @@ export default class Generator {
...
     consola.debug('Initializing generator...')
-
     await this.initiate({ build, init })
 
-    consola.debug('Preparing routes for generate...')
+    // Payloads for full static
+    if (this.isFullStatic) {
+      consola.info('Full static mode activated')
+      const exportTime = String(Date.now())
+      this.payloadDir = path.join(this.distNuxtPath, 'payloads', exportTime)
```
manniL: Is there a particular reason why we don't nest it inside `dist/assets/` instead of "just" `/dist/`? ☺️

packages/generator/src/generator.js
```diff
@@ -273,6 +320,19 @@ export default class Generator {
...
 
+  async addPayloadScript (route, html) {
+    const windowNamespace = this.options.globals.context(this.options.globalName)
+    const chunks = html.split(`<script>window.${windowNamespace}=`)
+    const [pre] = chunks
+    const payload = chunks[1].split('</script>').shift()
+    const post = chunks[1].split('</script>').slice(1).join('</script>')
+
+    // Write payload.js file
+    await fsExtra.writeFile(path.join(this.payloadDir, route, 'payload.js'), `window.${windowNamespace}=${payload}`, 'utf-8')
```
manniL: Have you thought about a caching solution (e.g. my proposed build hash option or sth. more customizable) so we can ensure payloads can be fully cached? 😋

packages/generator/src/generator.js
Atinux: I prefer to separate them from the `assets` directory actually for better logic separation, also it open the possibility to upload only the `payloads` when running `nuxt export` again

packages/generator/src/generator.js
Atinux: Well actually I want to separate the **build** and the **generation** so if only the content changes, only the payloads will change and no webpack build is necessary.

Also, since they live inside `dist/` they will be cached on any static hosting.

manniL: > Also, since they live inside dist/ they will be cached on any static hosting.

Without fingerprinting/hashes, this could easily cause caching issues when you treat them as immutable (as most of the generated JS files should be treated) ☺️

Atinux: Actually the payloadPath includes an exportTime so the path will be unique for each « nuxt export » call 😁

manniL: Ohh, must've missed that. Clever 👍🏻

TheAlexLichter (dismissed):

husayt:
As it is getting shape this is getting even cooler than anticipated.

ion1:
Nice work!

It would be useful for me if it was possible to merge the payloads of every page corresponding to a certain dynamic route into a single file. The total data in my use case would be about 100 kB uncompressed (about 10 kB concatenated and compressed) and cover about 100 pages. Users are likely to navigate to many of the pages in a session, and each navigation involving the fetch of a <1 kB payload.json seems less preferrable than fetching  the whole database once when any one of the pages is accessed.

Maybe payload.json could be a key-value map with keys corresponding to the dynamic part of the route.

Additionally, would it be possible to generate a `link rel="preload"` corresponding to the payload file?

Thanks!

Atinux:
Hi @ion1

I understand this improvement but I don’t believe that it will scale nicely, some people have 10k dynamic pages and that would create a really big payload.

Regarding the rel=preload, it’s a good idea and I should add it for the payload.js which is added at the end of the body indeed :)

ion1:
I completely agree merging payloads should not be the default, but it would be very nice if it was configurable for a dynamic route.

manniL:
One thing that just came to my mind: Do we already handle the edge case that some pages are rendered through the SPA fallback (and that there shouldn't be payloads loaded but the actual asyncData function executed there instead)? ☺️

Atinux:
I need to add a fallback when the payload is not found to not mock asyncData/fetch indeed, good one @manniL !

manniL:
> I need to add a fallback when the payload is not found to not mock asyncData/fetch indeed, good one @manniL !

Maybe we can restore the original asyncData function when fetching fails? ☺️

husayt:
Hi @Atinux , can you please provide an update on the status of this long awaited feature. It has been announced few times, but seems some problem is holding it up. Perhaps there is an issue we as a community can help with.  

Alos, what will you suggest to use meanwhile as a replacement? There are some solutions which are partially doing this, like:
https://github.com/stursby/nuxt-static/
https://github.com/DreaMinder/nuxt-payload-extractor
https://github.com/pimlie/nuxt-lambda
and others

Thanks.

edimitchel:
Something comes.. https://twitter.com/Atinux/status/1249385748874309633

packages/generator/src/generator.js
```diff
@@ -273,6 +350,19 @@ export default class Generator {
     return true
   }
 
+  async addPayloadScript (route, html) {
```
pi0, line 353: Side note: later we may need CSP handling

packages/server/src/server.js
```diff
@@ -320,13 +320,11 @@ export default class Server {
   renderAndGetWindow (url, opts = {}, {
     loadingTimeout = 2000,
     loadedCallback = this.globals.loadedCallback,
-    ssr = this.options.render.ssr,
```
pi0, line 323: Why?

packages/vue-app/src/index.js
```diff
@@ -5,11 +5,13 @@ export const template = {
   dependencies,
   dir: path.join(__dirname, '..', 'template'),
   files: [
+    'nuxt/config.json',
     'App.js',
     'client.js',
     'index.js',
     'router.js',
     'router.scrollBehavior.js',
+    'router/routes.json',
```
pi0, line 14: While I agree it is a cleaner structure, but we already have router.js and router.scrollBehaviour.js. We end up a directory with one file :D

packages/vue-app/template/App.js
```diff
@@ -106,12 +106,17 @@ export default {
   <% } %>
   beforeCreate () {
     Vue.util.defineReactive(this, 'nuxt', this.$options.nuxt)
+    <% if ((features.asyncData || features.fetch) && isFullStatic) { %>
+    if (process.client) {
+      Vue.util.defineReactive(this, 'payloadPath', window.__PAYLOAD_PATH__)
```
pi0, line 111: Side note: It means globalName is not supported

packages/vue-app/template/App.js
```diff
@@ -277,9 +282,54 @@ export default {
...
+    hasPayload(href) {
+      return !!(this._payloadCache && this._payloadCache[this.getPayloadPath(href)])
+    },
+    async fetchPayload (href) {
+      this._payloadCache = this._payloadCache || {}
+      const payloadPath = this.getPayloadPath(href)
+      let payload = this._payloadCache[payloadPath]
+
+      // If payload is a promise, returns it
+      if (payload && payload instanceof Promise && typeof payload.then === 'function') {
```
pi0, line 305: `then` is promise proto. I think `instance of` a check is extra :D

packages/vue-app/template/client.js
```diff
@@ -501,6 +517,11 @@ async function render (to, from, next) {
       this.$loading.manual = Component.options.loading === false
 
       <% if (features.fetch) { %>
+        <% if (isFullStatic) { %>
+        if (!this.$isPreview) {
+          promises.push(this.fetchPayload(to.path).catch(err => null))
```
pi0, line 522: Why are we hiding error?

packages/server/src/server.js
Atinux: This option was not useful and leading to errors in tests: https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/jsdom.js#L53

packages/vue-app/src/index.js
Atinux: It could be moved to `routes.json` indeed

packages/vue-app/template/App.js
Atinux: Indeed but not needed for this particular feature :)

packages/vue-app/template/App.js
Atinux: You can update, I was using old code for this check :D

packages/vue-app/template/client.js
Atinux: If an error occurs, it is because we did not generated the payload and should go back to normal fetch behaviour (SPA fallback case)

packages/builder/src/builder.js
```diff
@@ -122,6 +124,10 @@ export default class Builder {
       consola.info('Initial build may take a while')
     } else {
       consola.info('Production build')
+      const rendering = this.options.render.ssr ? 'server-side' : 'client-side'
+      consola.info(`Rendering: ${chalk.bold.yellow(rendering)}`)
+      const target = this.options.target === TARGETS.static && this.options.generate.static && this.options.render.ssr ? 'full static' : this.options.target
```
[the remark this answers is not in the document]
Atinux: Godd idea

packages/vue-app/template/client.js
pi0: But it might be also an internal error with `fetchPayload` or network request. Maybe we can add reason as a comment above and mention in docs about this fallback behaviour on any network issue? SPA fallback means api call?

packages/builder/src/builder.js
```diff
@@ -122,6 +124,10 @@ export default class Builder {
       consola.info('Initial build may take a while')
     } else {
       consola.info('Production build')
+      const rendering = this.options.render.ssr ? 'server-side' : 'client-side'
+      consola.info(`Rendering: ${chalk.bold.yellow(rendering)}`)
```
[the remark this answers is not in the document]
Atinux: This is an info message to say what "rendering" mode we are using: server-side or client-side
This is simply for DX

packages/vue-app/template/client.js
Atinux: Adding a comment

packages/builder/src/context/build.js
```diff
@@ -3,7 +3,7 @@ export default class BuildContext {
     this._builder = builder
     this.nuxt = builder.nuxt
     this.options = builder.nuxt.options
-    this.isStatic = false
+    this.target = builder.nuxt.options.target || 'server'
```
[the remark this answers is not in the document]
Atinux: I think it was because of unit testing 🤔

packages/cli/src/commands/generate.js
```diff
@@ -82,12 +90,14 @@ export default {
     }
 
     const generator = await cmd.getGenerator(nuxt)
+    await nuxt.server.listen()
```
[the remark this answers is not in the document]
Atinux: It's actually an open issue regarding the usage of serverMiddleware when using `nuxt generate` (example: content module)

packages/generator/src/generator.js
```diff
@@ -1,16 +1,18 @@
...
 
-import { flatRoutes, isString, isUrl, promisifyRoute, waitFor } from '@nuxt/utils'
+import { flatRoutes, isString, isUrl, promisifyRoute, waitFor, urlJoin, TARGETS, MODES } from '@nuxt/utils'
 
 export default class Generator {
   constructor (nuxt, builder) {
     this.nuxt = nuxt
     this.options = nuxt.options
     this.builder = builder
+    this.isFullStatic = false
```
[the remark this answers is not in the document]
Atinux: Actually since it can change when running `builder.forGenerate()` as well as saved in `nuxt/config.json`

packages/generator/src/generator.js
Atinux: cc @clarkdo on this

packages/builder/src/builder.js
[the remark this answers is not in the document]
pi0: Is see DX point but this is wrong information when using `nuxt build` we don't actually render. Maybe we can use "Building with/without server-side rendering support?"

image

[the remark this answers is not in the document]
Atinux: Indeed, what about saying: `Bundling for server and client side` or `Bundling for client side only`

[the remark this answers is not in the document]
pi0: Actually (currently) we always build for client-side too. It is optional/default server-side rendering that important. We one day supporting ssr-only makes sense :p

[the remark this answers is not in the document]
Atinux: This is just for DX, explaining what we build, one day we could have `Building for server side only` 😄 

Will update this code

pi0 (approved):

pi0:
Confirming with @Atinux and after team meeting:

- PR status is fully ready now to be tested on edge
- We need to add tests covering new functionalities later on before next release
- We need to release docs before next release
- `created` hook change for new fetch should be tested if introducing a breaking change or not (unwanted behaviour changed)
- Some performance parts (including initial chunk) need to be tested against real projects to tune up

J05HI:
@Atinux @pi0 
I found a bug in full static export:
If I switch routes it doesn't load the state for the new route.

Atinux:
Indeed @J05HI, this is not a bug but a specific behavior. We don’t want to save the state in each payload for performances but also the fact that replacing the whole state each time is not good.

J05HI:
@Atinux
Okay makes sense.
Is there a workaround for this? Or shouldn't I just use the store if I need full static?

Atinux:
You need to avoid using the store. Could you share and example of how and why you are using it?

Atinux:
@J05HI Actually you can use an anonymous middleware in order to dispatch actions to the store: https://nuxtjs.org/api/pages-middleware#anonymous-middleware

Keep in mind that they won’t be cached as full static though

J05HI:
@Atinux 
https://github.com/J05HI/nuxtdoc
Take a look at this demo. I changed it to nuxt-edge and enabled static export.
At this line it tries to change the store through a mutation which doesn't work on static: https://github.com/J05HI/nuxtdoc/blob/a52bf35b17b5120100e7d41237e9351e246bed03/pages/_version/_category/_doc.vue#L41

J05HI:
@Atinux @pi0
I have a question/addition for the payload extraction.
Is it possible to extract and save only used data to payload.js? Would save a lot of data because you never need the whole api response.

Atinux:
@J05HI Well you only need to return the data you will use in asyncData :)

J05HI:
@Atinux
The solution is to filter out not needed data before returning.
Thank you very much for your help! Keep the good work!

J05HI:
@Atinux
I have another question.
Is there a way to rebuild only specific pages? If the api side content only changes for a specific page I would like to only export it, not all the others which stayed the same.

gridsystem:
@pi0 @Atinux  Awesome! Really excited to use first party code for this.

What does this functionality cover in terms of asyncdata, fetch and nuxtserverinit? Will all data from those methods be included?

DreaMinder:
Woah, awesome. I'm really glad to deprecate my module in favor of native implementation. And since I've seen some edge cases, I'll be able to check them this week...
I have one concern so far: versioning. The native implementation uses timestamps to invalidate browser cache on every build. That's what I'm using for my module. 
I noticed once that netlify builds are faster without payload-extractor because when enabled, netlify has to reupload all payloads to CDN from scratch. It would be ideal to upload only files that actually changed.
Do you see any way to use hashes on payload filenames instead of timestamps? I know, it may make builds slower and more CPU-intensive. And you wouldn't like to have additional hashing dependency, especially in nuxt-start. But still, maybe you have a better idea on how to solve it?

And answering @J05HI question:
You can implement it yourself using nuxt programmatically. Just override routes array before running `generator.generate()`, Or you could use one of the [hooks](https://nuxtjs.org/api/internals-generator#hooks) to do the same.

lihbr:
> I have one concern so far: versioning. The native implementation uses timestamps to invalidate browser cache on every build. That's what I'm using for my module.

About that what's the preferred workaround for this scenario? 🤔 (it existed before with a similar module as nuxt-payload-extractor, and from what I understand it still exists(?))
1. User is on the website
2. Website is regenerated & redeployed
3. User try to get to another existing page
4. Because site has been updated previous timestamped file is not available anymore and lead to 404 where there isn't a 404

J05HI:
@DreaMinder 
Thank you very much! I'll try that.

DreaMinder:
@lihbr I think the best way to solve this issue is to use own deployment system that stores old static files along with new ones. Proper payloads versioning could help a little bit, but not in the case when user tries to open a page that has been updated.

Atinux:
@J05HI actually @pi0 implemented the store mutations when done in `asyncData / fetch`: https://github.com/nuxt/nuxt.js/pull/7350

This should fix your issue :)

@DreaMinder thank you again on your work for paylaod-extractor, it helped a lot shaping this feature!

Regarding versionning, I thought about it a lot, but the biggest issue is the ability to guess the filename when navigating on client-side, and if each payload file has different hash, this will be impossible without keeping a manifest file, but this one can become really big depending of how many routes you have in your application. We could consider such feature in the future as enhancements :)

DreaMinder:
@Atinux oh, right, I remember now. I tried to do it last year and stopped exactly because of this reason... and forgot about it. 😬 
I think manifest-file solution is not that bad for most of the projects. And it's useful not only for storing payload hashes: I used a map of existing pages to check if a link is external in one of my projects. Also, it's useful for implementing breadcrumbs, overriding CMS-slugs (or IDs)... But it should be an option disabled by default, in case the manifest file becomes too big.

lihbr:
@DreaMinder I agree with you, I think I'll try to implement a way to do a hard reload if fetching payload fail once when this feature become available out of nuxt-edge, doing a hard reload here given the edge case seems decent to me :)

J05HI:
@Atinux @pi0 @DreaMinder 
Are there any export hooks? Like [generate hooks](https://nuxtjs.org/api/internals-generator#hooks)?
I would like to save all images in API responses to assets folder and replace the image URL with the local one. I need them locally to process them with webpack ([Nuxt Optimized Images](https://github.com/aceforth/nuxt-optimized-images)).

J05HI:
@Atinux 
> Are there any export hooks? Like [generate hooks](https://nuxtjs.org/api/internals-generator#hooks)?
> I would like to save all images in API responses to assets folder and replace the image URL with the local one. I need them locally to process them with webpack ([Nuxt Optimized Images](https://github.com/aceforth/nuxt-optimized-images)).

Are there any news for this?

husayt:
@Atinux FYI many of projects I have are broken after trying them on this version. I probably need to pass some new parameters. But it would be even better if the new features work as before with previous configuration. Otherwise most of existing nuxt projects will be broken.

For example, SPA with all entry points generated after runnig this version doesn't have any of the other routes generated.

Atinux:
@husayt could you please provide me a reproduction? Nothing should break actually if you don’t specify the target: ‘static’

husayt:
@Atinux I don't have to go far, even the spa example which comes with nuxt  breaks.
https://github.com/nuxt/nuxt.js/tree/dev/examples/spa
add ` "generate": "nuxt generate"`, line to scripts, then  `run generate`. You will see `about` folder in dist.
Now switch to `nuxt-edge` and `yarn run generate` again. You will see there is no `about` folder in dist

Atinux:
@J05HI you can use the generate hooks, they are the same when using `nuxt export` actually (we just renamed to command to avoid breaking changes).

@husayt oh yes this is because it makes no sense to generate all the routes when using `mode: 'spa'` since you have the [SPA fallback](https://nuxtjs.org/api/configuration-generate#fallback) (`200.html`) to catch all the routes or you can directly link to `index.html` to catch all the routes.

husayt:
@Atinux but still it breaks the previous behavior. Many used this to get SPAs with all entry points generated to have routes working on static websites. And what will be similar behavior in the new version?

J05HI:
@Atinux
Thank you very much for your reply!
I was confused of generate. Shouldn't we introduce separate export hooks? It would be much clearer.

Atinux:
@husayt you have valid point, will revert it :)

@J05HI indeed we are going to add `export:*` hooks as well as the `export` key (alias of `generate`)

J05HI:
@Atinux
Yeah, alias is the key ;)

husayt:
@Atinux Thanks. Also to help with testing of this so much anticipated feature it would be great if there was some basic description of how this new mode works. What we have in description here is too little, I ended up breaking everything  and just gave up until it was clear how it all works.

Atinux:
At the moment you can learn more on https://www.notion.so/nuxt/Going-Full-Static-0633ab06d3d34d479947694d457a4817

Will publish it soon on the Nuxt blog

husayt:
> Hey @husayt
> 
> Target is directly linked to the platform you want to host your Nuxt.js App, by default, it's `server` and is a normal node.js server with SSR.
> 
> * `server`: Normal node.js hosting (Heroku, Amazon EC2, etc)
> * `serverless`: Serverless hosting (Now V2, AWS Lambda, Google Cloud Functions, etc)
> * `static`:  Static/CDN hosting (Netlify, GitHub Pages, Surge.sh, Amazon S3, etc)
> 
> They mainly differs when running `nuxt build`, so Nuxt.js knows what kind of hosting platform you are aiming at :)

@atinux thanks for the link to documentation, that helped, but also created a number of questions. I am confused also probably because of changing scope for this feature.
Here is a reply from you last year where you mentioned serverless mode, but seems that now it is being dropped.

So as I understand we have these modes :
 * `server`: Normal node.js hosting (Heroku, Amazon EC2, etc)
with `target:server` and commands  `npm run build` and then `npm run start`
 * `static`:  Static/CDN hosting (Netlify, GitHub Pages, Surge.sh, Amazon S3, etc)
with `target:static` and commands  `npm run build` and then `npm run export`

And both of them have modes `universal`/`SPA`.

Also we still have `nuxt generate` which as I understand does `npm run build` and  `npm run export` as one step. 

Now we have a combination of  two settings target (server/static) and mode(spa/universal), we have
*build-start-server/spa
*build-start-server/universal
*build-export-static/spa
*build-export-static/universal

What about generate, is it being deprecated or is it going to stay?

I am probably confused becaused I think in old terms, but I guess there will be a lot of similar questions.

Atinux:
@husayt thank you for your feedback

> What about generate, is it being deprecated or is it going to stay? 

We plan to deprecate it yes, but because we want to add full static for 2.13, we cannot add a breaking change, so it will be deprecated and removed for Nuxt 3.

SvenC56:
Hey,

I am using `nuxt-edge@2.13.0` and I am trying to use the full static version which is working for me. But if I click on a link and go back again it reaches out to my external headless cms (strapi). I am using it in the conjunction with the `nuxt-community/apollo-module`. Shouldn't it just reuse the payload from the generated page?

This is my AppNavbar.vue (script part) for example:

```js
  apollo: {
    navigation: {
      prefetch: true,
      fetchPolicy: 'cache-and-network',
      query: navigationQuery,
    },
  },
```

Atinux:
Hi @SvenC56 

Actually the full static only mock fetch and asyncData, Apollo is used outside of it actually, I can see quickly how I can support it but may be tricky

SvenC56:
> full static only mock fetch and asyncData

Thanks to this hint I adapted my config the following (same example as above https://github.com/nuxt/nuxt.js/pull/6159#issuecomment-632768070):

```js
async fetch() {
    const response = await this.$apolloProvider.defaultClient.query({
      query: navigationQuery,
    })
    this.navigation = response.data.navigation
  },

  fetchOnServer: true,
```

Now the data is rendered serverside and it does not reach out to my API anymore. :slightly_smiling_face:

emiliobondioli:
Hey all! First of all thanks for the work you're doing on this :) 
I finally managed to get full static generation where all I have to do is upload the dist folder on Netlify and everything works right away, payload and assets included.

Right now I'm still using a little hack to add a `generate:staticAssets` hook to grab the image links and replace them with local versions before saving the json payloads (as discussed [here](https://github.com/nuxt/rfcs/issues/22#issuecomment-623672794)). I wanted to ask the Nuxt team if it's possible to include a hook to modify the payload before it goes through [@nuxt/devalue](https://www.npmjs.com/package/@nuxt/devalue), or if you guys have some better ideas on how to handle this case - I tried to look around the code but I'm still a bit lost on how everything works under the hood.

nilsnh:
Hi! Thanks for all the hard work everyone. This is indeed a great feature. I'm wondering if the [generate route+payload](https://nuxtjs.org/api/configuration-generate#speeding-up-dynamic-route-generation-with-code-payload-code-) feature has been removed?

```javascript
export: {
  routes (): {
     return [{
       route: '/',
       payload: pageData
     }]
  }
}
```

I'm asking because I'm using [Sanity CMS](https://www.sanity.io/) as backend and they have a price cap on **the number of API calls** we make. So, I would like to be able to be able to grab all the page data in one call and feed it to all the pages, instead of making API calls per page on each deploy or content change.

I did some testing with the classic generate route+payload feature just now using `nuxt-edge@2.13.0-26508984.2383bcbd`. I'm calling `npx nuxt build && npx nuxt export`. And it doesn't seem to use my pre-generated route payloads.

Might be bugs in my code off course. Nonetheless, it would be great to hear if generating data for all pages **in one action** is still supported?

Thanks. 🙏 

cc: @Atinux

davidsneal:
Literally just came to as the same thing as @nilsnh :)

rajesh-h:
Hi Team,

I am trying this static generation. I see Lot of 404 on my console whenever subpage has slug which has space in it. 

open chrome inspect/devtools
go to homepage - https://cscom-2019.web.app
you can see many errors.
Here subpage is category. Any category which has a space(%20) in its name is giving 404 error on console for its payload.js.

One more scenario:
Single page - https://cscom-2019.web.app/aloo-puri-recipe-crispy-yet-soft-masala-potato-poori-cookingshooking

This post is linked to 6 categories, apart from one category rest 5 categories payload.js throws 404 error.

Please let me know if you need more details on the same.

rajesh-h:
Ok I have planned to change my backend data for categories list.

But now I am stuck with querying :(

how can i look for slug = 'ABC' in the below categories list/array

`categories = [ {'slug':'123', category:'1 2 3'}, {'slug':'ABC', category:'A B C'}, {'slug':'XYZ', category:'X Y Z'}]`

`qry = this.$content().where({  categories: { $contains : 'How To Manage Slug Here'  }  })`

Earlier I had categories as below, only values
`categories = ['A B C', '1 2 3', 'X Y Z']`
and my query was as below
`qry = this.$content().where({  categories: { $contains :  'A B C'  }  })`

pi0:
@rajesh-h Would you please kindly report this issue to [nuxt/content](https://github.com/nuxt/content) and possibility with proper repro?

I will also try to reproduce and check the issue you mentioned about spaces :)

rajesh-h:
Thank you @pi0. looks like json/netsed filter is not possible based on comments on [this issue. ](https://github.com/nuxt/content/issues/36#issuecomment-635188020)

If space issue is resolved, I can go back to my original design where keeping categories array as simple list. Please let me know if you want any input for the space issue.

Thank you very much.

Davounet:
Hello,

Thank you so much for the static generation mode, I'm trying it through nuxt-edge, great job from the team !
I stumbled upon a weird flash between route changes. The error page appears for a brief time, and displays a `resolveRouteComponents is not defined` message.

In dev mode, it happens on each route change.
In production mode (after `nuxt build && nuxt export && nuxt serve`) it does not happen between generated routes anymore, but still for fallback routes that are excluded from the static generation (the `/profil` route for example).

In both modes, the route loads after the flash. Did you encouter this behaviour before ?
Thank you !

rajesh-h:
> @rajesh-h Would you please kindly report this issue to [nuxt/content](https://github.com/nuxt/content) and with proper repro?
> 
> I will also try to reproduce and check the issue you mentioned about spaces :)

@pi0 I am not sure its the issue with space :( . 
I removed the space on my category names still find the 404 error.

When you go down at this link - https://cscom-2019.web.app/category/cakes-more when more posts are getting loaded you will see lot of 404 for payload.js

this behavior not observed on homepage. https://cscom-2019.web.app you can keep going down and load posts.

Both have the same structure like load 10 posts initially then load further posts when you scroll down. Any lead to resolve this will be helpful.

In case you want to refer:
My Github Repo [is here](https://github.com/rajesh-h/cs_content)
SandBox - [Here](https://codesandbox.io/s/github/rajesh-h/cs_content)

Side Question: Does the static site removes that nice animation of horizontal line page loading on top when you click on a link?

lautiamkok:
Pages with pagination navigations with this "full-static" feature don't work by the way. I assume that it is a bug. The pagination works fine with Nuxt 2.12.2.

rajesh-h:
> > @rajesh-h Would you please kindly report this issue to [nuxt/content](https://github.com/nuxt/content) and with proper repro?
> > I will also try to reproduce and check the issue you mentioned about spaces :)
> 
> @pi0 I am not sure its the issue with space :( .
> I removed the space on my category names still find the 404 error.
> 
> When you go down at this link - https://cscom-2019.web.app/category/cakes-more when more posts are getting loaded you will see lot of 404 for payload.js
> 
> this behavior not observed on homepage. https://cscom-2019.web.app you can keep going down and load posts.
> 
> Both have the same structure like load 10 posts initially then load further posts when you scroll down. Any lead to resolve this will be helpful.
> 
> In case you want to refer:
> My Github Repo [is here](https://github.com/rajesh-h/cs_content)
> SandBox - [Here](https://codesandbox.io/s/github/rajesh-h/cs_content)
> 
> Side Question: Does the static site removes that nice animation of horizontal line page loading on top when you click on a link?

@pi0 did you happen to see what might be causing 404 errors? any lead would be helpful. Functionally no problem, just that too much error on console.

simplenotezy:
I am eager to test this out, but I have two questions:

1) Will it work on a e-commerce platform with roughly 5-6000 routes with payload? Perhaps increase generate time a lot, or something else?
2) Will it work with Vuex? In my `asyncData` I do like so:

	async asyncData ({ store, app, error, route, payload }) {
		if (payload) {
			store.dispatch('product/set', payload);
			return;
		}

simplenotezy:
@Atinux GREAT work! Seems to work out of the box. Even with routes which are meant to be fetched from server. Page switching is waay faster now.

However, I noticed my og:description is now "undefined". I am not sure if it is related to this pull through. I set og:description to `process.env.npm_package_description`. Has anything changed with this? @pi0

simplenotezy:
One question though @Atinux. I am curious, will this have any inital page speed / load time issues? 

How is the implementation done exactly?

I really love the feature, and how fast it now switches pages, but we have a e-commerce platform with 5-6000 routes, so I am curious if this amount of pages will slow down initial page load speed.

You can check here https://beta.thejewelleryroom.com vs https://thejewelleryroom.com

Vout:
> Hello,
> 
> Thank you so much for the static generation mode, I'm trying it through nuxt-edge, great job from the team !
> I stumbled upon a weird flash between route changes. The error page appears for a brief time, and displays a `resolveRouteComponents is not defined` message.
> 
> In dev mode, it happens on each route change.
> In production mode (after `nuxt build && nuxt export && nuxt serve`) it does not happen between generated routes anymore, but still for fallback routes that are excluded from the static generation (the `/profil` route for example).
> 
> In both modes, the route loads after the flash. Did you encouter this behaviour before ?
> Thank you !
> 
> EDIT: I was toying around the options. The weird flash only happened when I had `loading: false` in my nuxt config, I removed it and it is now working perfectly. I'm just leaving this message here in case someone finds themself in the same situation !

Same issue over here, the following error flashes by when changing routes however no error is logged to the console.

```
resolveRouteComponents is not defined
An error occurred while rendering the page. Check developer tools console for details.
```

I have it in dev and production mode.

josepjaume:
I'm finding a funny issue: Routes that return a 404 error still get statically rendered when using `nuxt export`. Shouldn't this halt the rendering and leave it to the `fallback` strategy?

simplenotezy:
> I'm finding a funny issue: Routes that return a 404 error still get statically rendered when using `nuxt export`. Shouldn't this halt the rendering and leave it to the `fallback` strategy?


Why would you generate a route that returns 404?

josepjaume:
> > I'm finding a funny issue: Routes that return a 404 error still get statically rendered when using `nuxt export`. Shouldn't this halt the rendering and leave it to the `fallback` strategy?
> 
> 
> Why would you generate a route that returns 404?

It's kind of a bug in my code (related to i18n and missing translated documents), but I can see some scenarios where the crawler can get confused and try to render a route that returns an error state - like, say, a route that comes from user input. Shouldn't Nuxt avoid statically rendering these routes and instead let the client-side code pick it up? Or, at least, shouldn't it statically render a 404 page? Right now it just falls back to the static SPA version of the page, which doesn't add much value. Funny enough; if you raise an exception, then the route doesn't get rendered (which is how I ultimately did the trick on my scenario).

In any case, this is such a neat feature and I'm enjoying it so much!

Atinux:
Normally if the crawler hits a page that render a 404, it will render the page with no ssr to let the client-side falling back

anjali89r:
@Atinux I want to host the app in was s3, but still want to load data from my API for the dynamic routes. is there a to disable reading the payload.js and read the API response ? I can exclude the dynamic routes, but was s3 will return it as 404 status and can impact the Google indexing

Issue #3504, opened 2018-07-05
Nuxt page fn $forceUpdate exists when running in dev mode but not in prod (build, start) [nuxt-edge]


hartmut-co-uk:
### Version

[v1.4.0](https://github.com/nuxt.js/releases/tag/v1.4.0)

### Reproduction link

[https://github.com/hartmut-co-uk/nuxt-forceupdate-bug-reproduce](https://github.com/hartmut-co-uk/nuxt-forceupdate-bug-reproduce)

### Steps to reproduce

Hi, please follow reproduction link provided.

_) git clone https://github.com/hartmut-co-uk/nuxt-forceupdate-bug-reproduce.git
A.1) start the app via `yarn run dev`
A.2) open start page, use (click) buttons - in particular top button from layout - '[page.$forceUpdate from layout]'
A.3) check console / store / $store.state.random template rendered output
A.4) the button from layout '[page.$forceUpdate from layout]' does invoke the $forceUpdate - triggering the page `fetch` method -> store commit -> state change..

B.1) build and start the app in prod mode via  `yarn run build && yarn start`
B.2) open start page, use (click) buttons - in particular top button from layout - '[page.$forceUpdate from layout]'
B.3) check console / store / $store.state.random template rendered output
B.4) doesn't work...

### What is expected ?

A.4) & B.4) should result in the same behaviour client side.
running nuxt in built / prod mode should allow to call `this.$refs.page.$forceUpdate()`

### What is actually happening?

it's working in dev mode, not in prod mode

### Additional comments?

This has been setup & tested in *nuxt-edge*

Please ask if any questions!

Usage scenario: from layout page header bar - I have a global 'create' which doesn't do a route change but allows creating (http POST) new elements via API - and after success redirect / reload current page - which is not a route change - but should invoke the page's `fetch` method to re-fetch data from API (via store..)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7337">#c7337</a>)</em></sub></div>

hartmut-co-uk:
Note: I'm actually doing sth like the following in addition:
```
let to = '/my/route/to'    // this is dynamic in my code...
if (this.$router.currentRoute.path === to) {
  this.$refs.page.$forceUpdate()
} else {
  this.$router.push(to)
}
```

[...]

PR #6194, opened 2019-08-06
feat(vue-app): add $nuxt.refresh()

atinux:
## Types of changes

## Description

Allo to refresh the page data (calling matched pages `asyncData` & `fetch`).

Will be used for the preview mode, as well as closing #3504

It also add `$nuxt.context` as alias (instead of `$nuxt.$options.context`)

## Checklist:

### DIFF
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -100,6 +100,28 @@ export default {
         }
       }
     },
+    async refreshPageData(showProgressBar) {
+      const context = this.$options.context
+      const pages = this.$route.matched.map((match) => match.instances.default)
+
+      if (!pages.length) {
+        return
+      }
+      this.$loading && this.$loading.start()
+      const promises = pages.map(async (page) => {
+        if (page.$options.fetch) {
+          await page.$options.fetch(context)
+        }
+        if (page.$options.asyncData) {
+          const newData = await page.$options.asyncData(context)
+          for (const key in newData) {
+            Vue.set(page.$data, key, newData[key])
+          }
+        }
+      })
+      await Promise.all(promises)
+      this.$loading && this.$loading.finish()
+    },
     <% if (loading) { %>
     errorChanged() {
       if (this.nuxt.err && this.$loading) {
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -1,4 +1,5 @@
 import Vue from 'vue'
+import { getMatchedComponentsInstances } from './utils'
 <% if (loading) { %>import NuxtLoading from '<%= (typeof loading === "string" ? loading : "./components/nuxt-loading.vue") %>'<% } %>
 <%if (buildIndicator) { %>import NuxtBuildIndicator from './components/nuxt-build-indicator'<% } %>
 <% css.forEach((c) => { %>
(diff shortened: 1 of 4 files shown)


galvez (dismissed):
Cool, that helps phasing out [nuxt-static-render](https://github.com/galvez/nuxt-static-render) too :)

packages/vue-app/template/App.js
```diff
@@ -100,6 +101,28 @@ export default {
         }
       }
     },
+    async refreshPageData(showProgressBar) {
```
pi0: As we call both asyncData and fetch, what do you think naming it one of these:
- `reload`,  `reloadPage` / `reloadRoute`
- `refrash`, `refrashPage` / `refrashRoute`
- `refetch`, `refetchPage` / `refetchRoute`

packages/vue-app/template/App.js
```diff
@@ -100,6 +101,28 @@ export default {
...
+      const context = this.$options.context
+      const pages = getMatchedComponentsInstances(this.$route)
+
+      if (!pages.length) {
+        return
+      }
+      this.$loading && this.$loading.start()
```
pi0: Please avoid one line ifs

packages/vue-app/template/App.js
```diff
@@ -100,6 +101,28 @@ export default {
         }
       }
     },
+    async refreshPageData(showProgressBar) {
+      const context = this.$options.context
```
pi0: Neat: could be const { context }

packages/vue-app/template/App.js
```diff
@@ -100,6 +101,28 @@ export default {
...
+      const pages = getMatchedComponentsInstances(this.$route)
+
+      if (!pages.length) {
+        return
+      }
+      this.$loading && this.$loading.start()
+      const promises = pages.map(async (page) => {
+        if (page.$options.fetch) {
+          await page.$options.fetch(context)
```
pi0: I think during page load we call fetch and asyncData in parallel. Here we can do the same by pushing promises to an array and await on all

packages/vue-app/template/App.js
```diff
@@ -100,6 +101,28 @@ export default {
...
+        }
+        if (page.$options.asyncData) {
+          const newData = await page.$options.asyncData(context)
+          for (const key in newData) {
+            Vue.set(page.$data, key, newData[key])
+          }
+        }
+      })
+      await Promise.all(promises)
+      this.$loading && this.$loading.finish()
```
pi0: Isn't error handling required in case of rejection of above await to finish loading in error state?

packages/vue-app/template/App.js
Atinux: Renamed to `refresh`:)

packages/vue-app/template/App.js
Atinux: Updated :)

kevinmarrec:
@Atinux Could `refresh` take a object param that would be added to the context ? (Would be useful to fetch things differently depending of if it's a refresh or not)

pi0 (dismissed):

Atinux:
We could add this functionality later on key @kevinmarrec, at the moment I prefer to keep it simple :)

kevinmarrec:
@Atinux Yes, surely !

PR #6200, opened 2019-08-08
feat(vue-app): support alias in head

ricardogobbosouza:
This PR allows you to use alias on link property in head

## Types of changes

## Description

Now this is possible
```js
// nuxt.config.js
export default {
  head: {
    link: [
      {
        rel: 'apple-touch-icon',
        sizes: '180x180',
        href: '~/assets/images/favicons/apple-touch-icon.png'
      }
    ]
  }
}
```

## Checklist:


packages/vue-app/template/head.js
```diff
@@ -0,0 +1,29 @@
+<%= isTest ? '/* eslint-disable quotes, semi, indent, comma-spacing, key-spacing, object-curly-spacing, space-before-function-paren  */' : '' %>
+const head = <%= serializeFunction(head) %>
+<%= isTest ? '/* eslint-enable quotes, semi, indent, comma-spacing, key-spacing, object-curly-spacing, space-before-function-paren */' : '' %>
+
+<%
+// Inject link with alias
+if (head.link) {
+  head.link.forEach((link, index) => {
```
pimlie: I believe we prefer to use for..in or for..of in nuxt. Also the reference to link should be kept so you shouldnt need to use index to set the element within the array.

rchl:
I can think of one other tag that would need same treatment (`meta`) but there are possibly more.

```html
<meta rel="icon" type="image/x-icon"  href="~/asssets/favicon.ico">
```

packages/vue-app/template/head.js
```diff
@@ -0,0 +1,29 @@
...
+    if(startsWithRootAlias(link.href) || startsWithSrcAlias(link.href)) {
```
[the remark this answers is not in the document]
pimlie: A href attribute could be empty, maybe do an early return/continue in that case

packages/vue-app/template/head.js
```diff
@@ -0,0 +1,29 @@
...
+head.link[<%= index %>].href = require('<%= relativeToBuild(resolvePath(link.href)) %>')
+<%
+    }
+  })
+}
+
+// Inject script with alias
+if (head.script) {
+  head.script.forEach((script, index) => {
+    if(startsWithRootAlias(script.src) || startsWithSrcAlias(script.src)) {
```
pimlie: see above

pimlie (changes requested):
Besides meta, I just know someone is gonna ask if (s)he can use an aliased import in a style tag. Do you think there is any chance to get that working as well?

Great work btw!

packages/vue-app/template/head.js
```diff
@@ -0,0 +1,29 @@
+<%= isTest ? '/* eslint-disable quotes, semi, indent, comma-spacing, key-spacing, object-curly-spacing, space-before-function-paren  */' : '' %>
+const head = <%= serializeFunction(head) %>
+<%= isTest ? '/* eslint-enable quotes, semi, indent, comma-spacing, key-spacing, object-curly-spacing, space-before-function-paren */' : '' %>
+
+<%
+// Inject link with alias
+if (head.link) {
```
pi0: This process is happening at runtime. We can prepare serialized data (as discussed in discord) for most efficient code

packages/vue-app/template/head.js
```diff
@@ -0,0 +1,29 @@
...
+  head.link.forEach((link, index) => {
+    if(startsWithRootAlias(link.href) || startsWithSrcAlias(link.href)) {
+%>
+head.link[<%= index %>].href = require('<%= relativeToBuild(resolvePath(link.href)) %>')
```
pi0: Simple require around string is enough. Webpack handles aliases

test/fixtures/head-alias/head-alias.test.js
```diff
@@ -0,0 +1,3 @@
+import { buildFixture } from '../../utils/build'
+
+buildFixture('head-alias')
```
pi0: Can't we extend another fixture (like with-config)? Adding fixtures makes CI slower

pi0 (changes requested):

ricardogobbosouza:
Actually there is another point which can be quite problematic if the user uses a head function
how are we going to access link, meta, style, script?

``
export default {
    head() {
        return {
            link: [
                {
                    rel: 'icon',
                    type: 'image/png',
                    href: '~/assets/favicon.png'
                }
            ]
        }
    }
}
```

pimlie:
@ricardogobbosouza We might be able to fully support functions by transforming to AST, manipulate the AST and then transform the AST back to source. This could be a bit tedious to get working properly (and not sure its even possible to get working properly), but if we could implement this it should always work.

There might be other (easier?) tools available, but I only have experience using babel/parser and babel/transform in [nuxt/tib](https://github.com/nuxt/tib/blob/master/src/utils/index.js#L49)

Steps we would need to implement are:
- parse the raw `nuxt.config` file from text to AST
- use [babel/traverse](https://babeljs.io/docs/en/babel-traverse) to find the `head` node in the AST
- find out if there are any imports or variables from a parent scope being used
  - if so, recursively include these
    - include variables by value
    - include functions by definition (e.g. import when possible)
  - _note: this is the big downside of this approach and responsible for the tedious part, there might be a better way and we might not get this too work properly in all use-cases anyway_
- create a new AST program with a default export for the head function AST
- transform the AST back to source and write to head.js file

pi0:
@pimlie I think this approach is a real overengineering over a simple requirement! Extracting a function and it's dependencies from the config file and then transforming back to an ES6 code is subject to countless edge cases and almost means rewriting a JS bundler. Also what if for example a function inside config that should be serialized depends on a dependency (like a vue library)? We are also importing it when loading config.

The `config` (both currently `nuxt.config.js` and upcoming `config/`) should have a clear definition and scope not mixing with runtime IMO.

Like #6055 approach, simply introducing `app/head.js` can resolve this problem, keeping backward compatibility and support for easier use cases and even able to mix both heads (like an object which is filled by PWA and other modules + user-provided ones)

pimlie:
@pi0 Yeah, fully agree thats its probably not worth the time to pursue this. But _theoretically_ it might just work maybe and/or possibly :angel:

pi0:
@pimlie I think even theoretically there is at least (as far as can think) one big problem: When loading `nuxt.config` through ESM, there are imports that are requirements of a function to be serialized. If even not conflicting that adds performance penalty. Also prevents using `~/` aliases for imports as there is no webpack.

pimlie:
@pi0 Maybe another approach could be to extract the head function out of nuxt.config, write a temporary `pre-head.js` that also includes imports from nuxt.config, then create `head.js` using webpack so it resolves the requirements for us. We could still use AST then to reliably determine which imports and parent scope variables are used within the head function.

Performance penalty would be significant probably indeed, but that would be the cost of using a function instead of an object as head in your nuxt.config

pi0:
A reminder that for nuxt3 (and even in a minor) we are also supporting `config/` directory. Having `config/head.js` or `app/head.js` is zero cost ;)

PR #6203, opened 2019-08-09
refactor(config): rename `devModules` to `buildModules`

pi0:
## Types of changes

## Description

`devModules`, initially introduced by #5102 to have a way to only include some modules during dev/build time and removing from production dependencies at runtime. It was named after `devDependencies` section in `package.json` as such modules should not be installed for production environment at all. However the original name is proven to be little confusing and making users thinking it is only for development time. New `buildModules` name tries to make it more clear.
However `devModules` was not documented, backward compatibility with deprecation massage is added to support `devModules` until 3.x.

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -376,7 +376,7 @@ export function getNuxtConfig (_options) {
 
   // Add loading screen
   if (options.dev) {
-    options.devModules.push('@nuxt/loading-screen')
+    options.buildModules.push('@nuxt/loading-screen')
     // Disable build indicator for programmatic users
     if (!options._cli) {
       options.build.indicator = false
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -338,7 +338,7 @@ export function getNuxtConfig (_options) {
 
   // devModules has been renamed to buildModules
   if (typeof options.devModules !== 'undefined') {
-    consola.warn('`devModules` has been renamed to `buildModules` and will be removed in nuxt 3.')
+    consola.warn('`devModules` has been renamed to `buildModules` and will be removed in Nuxt 3.')
     options.buildModules.push(...options.devModules)
     delete options.devModules
   }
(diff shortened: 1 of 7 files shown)


packages/config/src/options.js
```diff
@@ -336,6 +336,13 @@ export function getNuxtConfig (_options) {
     consola.warn('build.extractCSS.allChunks has no effect from v2.0.0. Please use build.optimization.splitChunks settings instead.')
   }
 
+  // devModules renamed to buildModules
+  if (typeof options.devModules !== 'undefined') {
+    consola.warn('`devModules` has been renamed to `buildModules`')
```
manniL: I'd also warn that it's deprecated and will be removed in Nuxt 3.

packages/config/src/options.js
```diff
@@ -336,6 +336,13 @@ export function getNuxtConfig (_options) {
     consola.warn('build.extractCSS.allChunks has no effect from v2.0.0. Please use build.optimization.splitChunks settings instead.')
   }
 
+  // devModules renamed to buildModules
```
```suggestion
  // devModules has been renamed to buildModules. TODO: Remove in Nuxt 3
```

atinux (dismissed):

packages/config/src/options.js
```diff
@@ -336,6 +336,13 @@ export function getNuxtConfig (_options) {
     consola.warn('build.extractCSS.allChunks has no effect from v2.0.0. Please use build.optimization.splitChunks settings instead.')
   }
 
+  // devModules has been renamed to buildModules
+  if (typeof options.devModules !== 'undefined') {
+    consola.warn('`devModules` has been renamed to `buildModules` and will be removed in nuxt 3.')
```
kevinmarrec:
```suggestion
    consola.warn('`devModules` has been renamed to `buildModules` and will be removed in Nuxt 3.')
```
If it seems relevant

kevinmarrec (dismissed):

atinux (approved):

manniL:
Great! It's important that we can somehow track the legacy fallbacks we have to remove for Nuxt 3, so adding it in the error message or as a TODO is the best way (to easily `grep` them out) IMO :relaxed:

pi0:
@manniL Good idea. I believe we can do this for all parts in a separate chore PR.

PR #6205, opened 2019-08-09
feat(vue-app): upgrade `vue-router`  to 3.1.2

pi0:
## Types of changes

## Description

There is currently a known issue with new promise support (vuejs/vue-router#2862) which is when using `$router.push` and `$router.replace` a new promise will be created by default and the `reject()` callback will be passed as `onAbort` to history provider ([src](https://github.com/vuejs/vue-router/blob/d907a13c3dcd43da757fdab2ee02ac3d104eb26f/src/index.js#L157)) during redirects, vue-router calls `abort()` without any params ([src](https://github.com/vuejs/vue-router/blob/dev/src/history/base.js#L163)) this now makes an unhandled promise rejection which resolve to undefined and is a semi-breaking change. We can patch e2e tests to catch the error but still, external-redirect test is failing and if nuxt users programmatically call `$router.push()` to a route which makes redirects will have a console error about unhandled rejection.

## Checklist:

### DIFF
--- a/distributions/nuxt-start/package.json
+++ b/distributions/nuxt-start/package.json
@@ -61,7 +61,7 @@
     "vue": "^2.6.10",
     "vue-meta": "^2.2.1",
     "vue-no-ssr": "^1.1.1",
-    "vue-router": "~3.0.7",
+    "vue-router": "^3.1.2",
     "vuex": "^3.1.1"
   },
   "engines": {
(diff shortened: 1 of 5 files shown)


pi0:
Failing cause: `router.push` is not working on client-side for redirects. Seems **not** a puppeteer bug as I could also reproduce via `basic` fixture (production) on local. Returned promise rejects with `undefined`

image

debugging more...

pi0:
@stale 

<a href="https://imgflip.com/i/3e3gu5"></a>

pi0:
Clossing as of #6722

PR #4257, opened 2018-11-02
fix: infinite loop if error page contains error

TheAlexLichter:
Resolves #4140 (maybe.. soon 🙈)

## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #6217, opened 2019-08-11
fix(vue-app): prevent looping on error during render of error page

### LINKED ISSUE #4140: error.vue infinite loop
### Version

[v2.1.0](https://github.com/nuxt.js/releases/tag/v2.1.0)

### Reproduction link

[https://codesandbox.io/s/vvz3072j15](https://codesandbox.io/s/vvz3072j15)

### Steps to reproduce

1. Create custom error.vue that contains an error
2. trigger error.vue
3. notice the infinte error loop

### What is expected ?

something catches the error

catches and attempts to load the error.vue infinitely

### Additional comments?

The main issue is that the API for the `error` prop being passed to the error.vue page has changed. Before 2.x, it was expected that `error.response` was available. Anybody that was using this, ie: `error.response.data` now have this error. There should be a notice to nuxt users so they are aware of this API change.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7995">#c7995</a>)</em></sub></div>

pimlie:
Follow up of: #4257 
Resolves: #4140 

## Types of changes

## Description

The `$forceUpdate` cannot be called async with eg $nextTick as then the NuxtError component would still be rendered and trigger the infinite reloading.

## Checklist:


pimlie:
Just to make sure: could someone please confirm it is correct that navigation to `/info` renders nothing in the e2e test? When I test this manually with nuxt then in dev mode it returns an error page but with build/start it indeed returns nothing. If you then refresh the page it says server error.

test/e2e/error.test.js
```diff
@@ -0,0 +1,60 @@
...
+  test('/about loads normally', async () => {
+    await page.nuxt.navigate('/about')
+
+    expect(await page.$text('h1')).toBe('About')
+  })
+
+  test('/info prints empty page', async () => {
+    await page.nuxt.navigate('/info')
+
+    expect(await page.$text('#__layout')).toBe('')
```
pimlie: Is this correct behaviour? In dev mode it returns an error page, with build/start it doesnt

pi0 (approved):

kevinmarrec (approved):

galvez (approved):

test/e2e/error.test.js
Atinux: Well it should displayed 'An error occured` error page I believe 🤔

pimlie: I checked with the dev branch and their the page is empty as well.

We could capture this error though as its being passed through errorCaptured. It just seems that by default in production there is no error handler so it just blanks out with an error in the console.

pi0: loading-indicator will go to a bad state in case of fatal errors and SPA mode. Otherwise, that blank screen is somehow unexpected.

Issue #5978, opened 2019-06-24
Add support of watchQuery for nested objects


Koc:
### What problem does this feature solve?

I am building grid for orders list with some filters. Next code used for parse/stringify query:

```js
    router: {
        parseQuery(query) {
            const qs = require('qs');
            const options = {allowDots: true, comma: true};

            return qs.parse(query, options);
        },
        stringifyQuery(query) {
            const qs = require('qs');
            const options = {allowDots: true, arrayFormat: 'comma', skipNulls: true};

            return Object.keys(query).length > 0 ? ('?' + qs.stringify(query, options)) : '';
        },
    },
```

Each filter updates coresponding `filter` property, for example for filter by status next url generates: http://admin.test:3000/orders-list?filter.status=done . And for now there is no support define `watchQuery: ['page', 'filter.status']` or `watchQuery: ['page', 'filter']` to get `asyncData` executed when `filter.status` updated.

`watchQuery: true` could be used as workaround, but seems to be bad idea in terms of performance.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9406">#c9406</a>)</em></sub></div>

danielroe:
Closing as we're not building new features for Nuxt 2, and this is quite specific to v2, but feel free to open a new feature request for Nuxt 3 if you would like to.

PR #6222, opened 2019-08-12
feat(vue-app): support functional `watchQuery`

clarkdo:
## Types of changes

## Description

Resolve issue like: #5978 , make `watchWatch` be function with fromQuery, and toQuery, user can define more customized logic.

## Checklist:


packages/vue-app/template/client.js
```diff
@@ -146,9 +146,12 @@ async function loadAsyncComponents(to, from, next) {
...
       const startLoader = Components.some((Component) => {
         const watchQuery = Component.options.watchQuery
-        if (watchQuery === true) return true
-        if (Array.isArray(watchQuery)) {
+        if (watchQuery === true) {
+          return true
+        } else if (Array.isArray(watchQuery)) {
           return watchQuery.some(key => this._diffQuery[key])
+        } else if (typeof watchQuery === 'function') {
+          return watchQuery(from.query, to.query)
```
pi0: What if calling it with `this` context?

pimlie: Maybe reverse the arguments as well? Is more in line with a Vue watcher which receives `newVal, oldVal`

pi0: We can also use named params :) `{ from, to }`

pimlie: Should be ok as well indeed.

Do you know if (at least in v8) thats optimized exactly the same as using arguments? Eg i know v8 optimizes calls to functions which have destructed args, but does it also optimize that object fully away if the receiving arg only destructures into one property? Sounds logical to do that if you are already optimizing destructures, but dont know for sure

pi0: Suggest to read about inline caches: https://mathiasbynens.be/notes/shapes-ics

clarkdo: The new component hasn’t been created and I think pointing to old instance is not very appropriate

pi0: We can use it to access general context like store, ...

pimlie: Thanks, interesting read. This article answered my question a bit more: https://codeburst.io/es6s-function-destructuring-assignment-is-not-free-lunch-19caacc18137

pimlie: Maybe add it as arg instead `({ from, to, fromComponent })`? Agree with @clarkdo that setting the old component as `this` could be confusing

pi0: Babel makes it different: 

https://babeljs.io/en/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAZmGBeGAKA3vATiAtgGhihBgF8BKZAPhnVKA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.5.5&externalPlugins=

pi0: But it makes sense as we are _watching_ from the current component. Like other watchers.

pimlie: Thats also a good point :)

packages/vue-app/template/client.js
```diff
@@ -151,7 +151,7 @@ async function loadAsyncComponents(to, from, next) {
         } else if (Array.isArray(watchQuery)) {
           return watchQuery.some(key => this._diffQuery[key])
         } else if (typeof watchQuery === 'function') {
-          return watchQuery(from.query, to.query)
+          return watchQuery({from: from.query, to: to.query})
```
pimlie: I guess you're adamant that we shouldnt pass the component in one way or another?

clarkdo: What do you mean?

packages/vue-app/template/client.js
pi0: Why dismissed?

clarkdo: GitHub automatically marked it as outdated? Sorry, I’ll apply change

packages/vue-app/template/client.js
pi0: I think he means not passing route object reference but query. In this case i would suggest keeping consistency with watch API (oldVal, newVal)

pimlie: I meant the same as @pi0 actually about the not-used `this` context. But instead of unresolving the previous thread I created a new one, sorry about that. Will resolve this one in favor of the other

clarkdo: 😿I’m kind of confused, which watch api do you mean?

clarkdo: Oh, you mean vue watch, I got it.

packages/vue-app/template/client.js
clarkdo: Done, please review

pi0 (approved):

packages/vue-app/template/client.js
```diff
@@ -140,15 +140,21 @@ async function loadAsyncComponents(to, from, next) {
...
     if (!this._pathChanged && this._queryChanged) {
       // Add a marker on each component that it needs to refresh or not
-      const startLoader = Components.some((Component) => {
+      const startLoader = Components.some(({Component, instance}) => {
         const watchQuery = Component.options.watchQuery
-        if (watchQuery === true) return true
-        if (Array.isArray(watchQuery)) {
+        if (watchQuery === true) {
+          return true
+        } else if (Array.isArray(watchQuery)) {
```
manniL: No else needed here :thinking: (same for line 156)

clarkdo: What's the benefit ?

manniL: Might be a personal touch, but I perceive it as more readable, especially when the `else` is omitted because there is an early return anyway. :relaxed:

Issue #5960, opened 2019-06-20
Ability to set build.publicPath at runtime


timmw:
### What problem does this feature solve?

This feature relates to apps using `mode: 'universal'`.

Allows for a single build artifact to be used in multiple environments with or without a CDN. 

In fact this used to be possible in Nuxt 2.3 using the following config:

```
  hooks: {
    //Nuxt doesn't allow the publicPath (CDN url) to be overridden at run time - it's baked into the manifest at build time.
    //This hook intercepts the VueRenderer when it has loaded the manifest and updates the publicPath to the current env value.
    render: {
      resourcesLoaded(resources) {
        resources.clientManifest && (resources.clientManifest.publicPath = process.env.PUBLIC_PATH || '/_nuxt/');
      }
    }
  },
```

However this change (https://github.com/nuxt/nuxt.js/commit/0f104aa58890f7b06837a05fc67718bf7822264a#diff-015dff3d45b52153ef5f24ef104e954bL161) in Nuxt 2.4 breaks this as the hook is now called _after_ the the renderer is created.

If anyone knows of a workaround that doesn't require editing the manifest file on disk at run time please let me know.

### What does the proposed changes look like?

`$ CDN_BASE_URL=https://cdn.com/_nuxt/ npm run start`

```
// nuxt.config.js
module.exports = {
  build: {
    publicPath: process.env.CDN_BASE_URL
  }
}
```

This should cause server-rendered script tags to use the URL specified by `CDN_BASE_URL`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9388">#c9388</a>)</em></sub></div>

kams2160:
Hi, I have problem with setting publicPath in build object in my nuxt.config.js at runtime:
```
build: {
    publicPath: process.env.PUBLIC_URL || '/_nuxt/',
}
```
I have to deploy one app to many environments with different publicPath addresses to static files. 
Any ideas? :slight_smile:
It looks like the publicPath is assign at build time and then at runtime the env variable wont be used
I also tried with nuxt-env, but it doesn't help.
Is any different way in nuxt.js to set static files url at runtime?
My nuxt version is 2.10.2.
Thanks :)

edwh:
I came across this useful thread when looking into how to use a CDN.  Thanks.

[...]

PR #6223, opened 2019-08-12
fix(vue-renderer): call `render:resourcesLoaded` hook before `createRenderer`

clarkdo:
## Types of changes

## Description

Resloves #5960, move hook before renderers creation, provide possibility to change resources at runtime.

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -128,13 +128,13 @@ export default class VueRenderer {
     // Load templates
     await this.loadTemplates()
 
+    await this.serverContext.nuxt.callHook('render:resourcesLoaded', this.serverContext.resources)
+
     // Detect if any resource updated
     if (updated.length > 0) {
       // Create new renderer
       this.createRenderer()
     }
-
-    return this.serverContext.nuxt.callHook('render:resourcesLoaded', this.serverContext.resources)
   }
 
   async loadTemplates () {


pi0 (approved):
waiting for CI... thanks :)

pi0:
I was also thinking to make it possible dynamically changing `publicPath` without the hooks hack.

clarkdo:
Function publicPath？

pi0:
@clarkdo no. simply updating the value from loaded config. which can be set differently at production. currently it is loaded from dist manifest.

clarkdo:
Also need a manual updating step at runtime?

pi0:
by nuxt :) WIll create a PR after this :)

timmw:
> I was also thinking to make it possible dynamically changing `publicPath` without the hooks hack.

@pi0 could you ping me when you make this PR. Thanks!

gapipro:
Is changing `publicPath` working only in SSR mode al also somehow in SPA? We are trying to get it to work, but there isn't any effect after this change.

clarkdo:
For spa, since browser has not interaction after first page request, so even you change the publicPath, the client side can't know the changes before a page reload. Could try refreshing the page ? If it still doesn't work, could you provide a repo ?

gapipro:
Well ideally it would be that `publicPath` that is updated in one of the hooks gets injected into final `runtime` bundle. Thats one way.

The other solution would be that Nuxt gets support for `__webpack_public_path__`

PR #6231, opened 2019-08-14
feat(vue-app): export router options

ricardogobbosouza:
Export router options

## Types of changes

## Description

Sometimes we need to use the `@nuxtjs/router` module for more custom creation, in which case `vue-router` is initialized twice.

https://github.com/nuxt-community/router-module/issues/56

## Checklist:

### DIFF
--- a/packages/vue-app/template/router.js
+++ b/packages/vue-app/template/router.js
@@ -98,6 +98,6 @@ export const routerOptions = {
   fallback: <%= router.fallback %>
 }
 
-export async function createRouter() {
+export function createRouter() {
   return new Router(routerOptions)
 }
--- a/packages/vue-app/template/router.js
+++ b/packages/vue-app/template/router.js
@@ -84,26 +84,20 @@ const _routes = recursiveRoutes(router.routes, '  ', _components, 2)
 
 Vue.use(Router)
 
-export function routerOptions() {
-  return {
-    mode: '<%= router.mode %>',
-    base: decodeURI('<%= router.base %>'),
-    linkActiveClass: '<%= router.linkActiveClass %>',
-    linkExactActiveClass: '<%= router.linkExactActiveClass %>',
-    scrollBehavior,
-    <%= isTest ? '/* eslint-disable quotes, object-curly-spacing, key-spacing */' : '' %>
-    routes: [<%= _routes %>],
-    <%= isTest ? '/* eslint-enable quotes, object-curly-spacing, key-spacing */' : '' %>
-    <% if (router.parseQuery) { %>parseQuery: <%= serializeFunction(router.parseQuery) %>,<% } %>
-    <% if (router.stringifyQuery) { %>stringifyQuery: <%= serializeFunction(router.stringifyQuery) %>,<% } %>
-    fallback: <%= router.fallback %>
-  }
+export const routerOptions = {
+  mode: '<%= router.mode %>',
+  base: decodeURI('<%= router.base %>'),
+  linkActiveClass: '<%= router.linkActiveClass %>',
+  linkExactActiveClass: '<%= router.linkExactActiveClass %>',
+  scrollBehavior,
+  <%= isTest ? '/* eslint-disable quotes, object-curly-spacing, key-spacing */' : '' %>
+  routes: [<%= _routes %>],
+  <%= isTest ? '/* eslint-enable quotes, object-curly-spacing, key-spacing */' : '' %>
+  <% if (router.parseQuery) { %>parseQuery: <%= serializeFunction(router.parseQuery) %>,<% } %>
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/router.js
```diff
@@ -97,5 +97,9 @@ export function createRouter() {
     <% if (router.parseQuery) { %>parseQuery: <%= serializeFunction(router.parseQuery) %>,<% } %>
     <% if (router.stringifyQuery) { %>stringifyQuery: <%= serializeFunction(router.stringifyQuery) %>,<% } %>
     fallback: <%= router.fallback %>
-  })
+  }
+}
+
+export function createRouter() {
```
pi0: Should we also allow passing `routerOptions` as an argument and default it to `routerOptions()` ? PS Did you validated this change against router-module? Can you please make a preview PR?

ricardogobbosouza: I actually wanted to use the router options and create my custom router.

ricardogobbosouza: I will pass routerOptions as parameter

packages/vue-app/template/router.js
```diff
@@ -100,6 +100,10 @@ export function routerOptions() {
   }
 }
 
-export function createRouter() {
-  return new Router(routerOptions())
+export async function createRouter(ssrContext, options = routerOptions) {
+  if (typeof options === 'function') {
+    options = await options(ssrContext)
```
pi0: await is unnecessary and costly. we can set default to `options = routerOptions()` and when calling with custom arg, await before calling `createRouter` :)

ricardogobbosouza: Can't pass `ssrContext`? I thought it might be useful for some kind of customization

pi0: Hmm actually what we are doing is:
- passing ssrContext and routerOptions to createRouter
- Let create router call options with ssrContext (with extra await)
- Use it to create router instance

This can be simply:

- Using ssrContext to create options object (can be await or not)
- Pass it to createRouter

```js
createRouter(options = routerOptions)
```

These are identical. Just refactor on function to make it simpler and lighter.

ricardogobbosouza: But this can generate a break change for those who use `ssrContext` in` createRoute`

pi0: I'm not sure. At least __currently__ nuxt accepts no arguments for `createRouter`. Maybe I'm misunderstanding how are you going to use it for router-module. Can you please provide more details about the usage of (new) exported function?

ricardogobbosouza: Today accepts yes, see https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L59

ricardogobbosouza: Today I do not have access to router options without initialization

ricardogobbosouza: I just wanted to have access to router options without having to boot

pi0: Is see. So isn't just exporting `routerOptions` functon enough to import from `defaultRouter`?

ricardogobbosouza: Exactly, I just need to export routeOptions

ricardogobbosouza: When we enable keepDefaultRouter
https://github.com/nuxt-community/router-module#accessing-default-router

The default router is created and I return a new router, ie the router is being started twice, I want to avoid this by using the default router options

ricardogobbosouza: We can simplify by doing this:
```js
...
}

export async function createRouter(ssrContext) {
  return new Router(routerOptions)
}
```

pi0 (dismissed):

packages/vue-app/template/router.js
```diff
@@ -84,18 +84,20 @@ const _routes = recursiveRoutes(router.routes, '  ', _components, 2)
...
+  scrollBehavior,
+  <%= isTest ? '/* eslint-disable quotes, object-curly-spacing, key-spacing */' : '' %>
+  routes: [<%= _routes %>],
+  <%= isTest ? '/* eslint-enable quotes, object-curly-spacing, key-spacing */' : '' %>
+  <% if (router.parseQuery) { %>parseQuery: <%= serializeFunction(router.parseQuery) %>,<% } %>
+  <% if (router.stringifyQuery) { %>stringifyQuery: <%= serializeFunction(router.stringifyQuery) %>,<% } %>
+  fallback: <%= router.fallback %>
+}
+
+export async function createRouter() {
```
pi0: I think async is unnecessary :)

pi0 (approved):

PR #6244, opened 2019-08-18
feat(vue-app): trigger watchParam when param is changed in same route

### LINKED ISSUE #6235: Strange behaviour of nested pages
### Version

### Reproduction link

[https://codesandbox.io/s/example-hello-world-kw5w9](https://codesandbox.io/s/example-hello-world-kw5w9)

### Steps to reproduce

1. Go to page `test`
2. Open Console. You will see console.log message each time when parent page was triggered
3. Press button 1-1-q. -> No messages. Works correct
4. Press button 1-2 -> Parent page was triggered. Incorrect
5. Press button 2-2 -> No messages. Works correct
6. Press button 2-2-q -> No messages. Works correct
7. Press button 2-2 ->  Parent page was triggered. Incorrect
 

### What is expected ?

Parent page should be triggered only on server side

If you change only `params` or only `query` part of route everything works correct.
If you change `params` and `query` part  of routing simultaneously then parent page component will be triggered

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9645">#c9645</a>)</em></sub></div>

clarkdo:
## Types of changes

## Description

Try to resolve #6235 , need more discussion.

Currently when dynamic path is changed (it means same route: like /test1 -> /test2 under /:test), we will always call asyncData instead of reusing, I'm not sure it this is in purpose @Atinux @pi0 .

~~After the pr, asyncData will be called in client side when:~~

- ~~route is changed (checking by from.name !=== to.name)~~
- ~~route is not change, but param is changed and watchQuery is true~~
- ~~route is not change, but query is changed and watchQuery is true~~

~~If current behaviour is correct and by design, how do you think adding a `watchParam` and make default value as `true` which means by default always calling asyncData when changing slug path.~~

I've changed this pr for adding `watchParam`

Behaviour after this pr:

- param is changed with same component, call asyncData
  - if only child param changed, call asyncData of child
  - if `watchParam` is false, `not` call asyncData of child 
- query is changed with same component, `not` call asyncData
- both param and query are changed with same component, call asyncData

## Checklist:


Atinux:
It was on purpose to call `asyncData` when the param changes but the same Component is used (actually it's a feature I missed in vue-router and did not want to people to use `watch: $route`).

I like the idea of introducing `watchParam: false` to disable it (even if I don't see a real use case)

clarkdo:
@Atinux Shared parent compoent may be a use case ? https://github.com/nuxt/nuxt.js/issues/6235

clarkdo:
@Atinux According to `children.patch.browser.test`, it seemed that `asyncData` is not called when param is changed.

Current behaviour is:

- param is changed with same component, not call asyncData
- query is changed with same component, not call asyncData
- both param and query are changed with same component, call asyncData

I don't think it's a correct behaviour, with the current pr, behaviour is:

- param is changed with same component, call asyncData
- query is changed with same component, not call asyncData
- both param and query are changed with same component, call asyncData

What do you think ?

Atinux:
I agree with the behaviour you describe, but we have to keep in mind about the Parent -> Child calls too,

- param is changed with same component, call asyncData
 - if only child param changed, call asyncData of child
- query is changed with same component, not call asyncData (expect if use `watchQuery`)
- both param and query are changed with same component, call asyncData

galvez:
@Atinux @clarkdo is there anything left to do here other than update tests?

galvez (dismissed):

clarkdo:
@Atinux @pi0 Can you review this again ?

msonowal:
@Atinux 
Why this is not merged till now as its a must have like watchQuery

pi0 (dismissed):

atinux (dismissed):
Approve but missing the docs PR :)

pi0 (approved):

PR #6248, opened 2019-08-19
examples(coffeescript): update module

### LINKED ISSUE #5423: Coffeescript Vuex silently stop working on 2.5.1
### Version

### Reproduction link

[https://github.com/nuxt/nuxt.js/tree/dev/examples/coffeescript](https://github.com/nuxt/nuxt.js/tree/dev/examples/coffeescript)

### Steps to reproduce

Build project with `"nuxt":  "2.5.1"`

### What is expected ?

Vuex that works with Coffeescript, just like on 2.4.5

Nothing.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8960">#c8960</a>)</em></sub></div>

devartyom:
## Types of changes

## Description

Resolves [#5423](https://github.com/nuxt/nuxt.js/issues/5423)

## Checklist:


packages/builder/src/builder.js
```diff
@@ -44,7 +44,7 @@ export default class Builder {
       restart: null
     }
 
-    this.supportedExtensions = ['vue', 'js', ...(this.options.build.additionalExtensions || [])]
+    this.supportedExtensions = ['vue', 'js', ...(this.options.extensions || []), ...(this.options.build.additionalExtensions || [])]
```
pi0: Are these options identical now? /cc @kevinmarrec

DevArtyom: Please let me know if I should use nuxt.options.build.additionalExtensions in the module that add support for coffeescript extension.
Like this.nuxt.options.build.additionalExtensions.push ('coffee')

kevinmarrec: @pi0 No
https://github.com/nuxt/typescript/blob/master/packages/typescript-build/lib/module.js#L18
We need to be aware that the 2 options aren't used at the same places in core

kevinmarrec: `options.extensions` is more or less related to runtime, meanwhile the others only buildtime

clarkdo:
Any reason for not using additionalExtensions?

kevinmarrec:
@pi0 @clarkdo If I created a new option it's cause there was a particular need.

All i know is that `options.extensions` is used to resolves files : 
https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L164
https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/resolver.js#L95

And `options.build.additionalExtensions` is used to build files that are not `.vue` or `.js` around Builder, or mostly create the router from files having these extensions.

This requested change https://github.com/nuxt/nuxt.js/pull/6248/files#diff-4f235fc89b94513b32c2cc8b5896cc5aR47 , would make `js` duplicated + make Builder build `.mjs` which Idk if it's relevant.

Having only one `extensions` would mean that the TypeScript module would make Nuxt also trying to resolve `.tsx` which shouldn't be needed ?

Idk

pi0:
@kevinmarrec Seems reasonable to have separated ones because of `tsx` :)

pi0 (dismissed):

clarkdo:
Sorry, I didn’t mean that we need to keep only one config, I mean that why we need to merge extensions into supportedExtensions instead of configuring additionalExtensions. For the coffeescript case, only adding this won’t work, user still needs to setup loader.  And current change also changed the resolving order and files of routes, it’s kind of breaking change for users having files inside pages dir but not meant to be a route.

kevinmarrec:
Exactly @clarkdo, cofeescript support for routes should laverage `build.additionalExtensions` within the cofeescript module.

If somehow cofeescript files should be resolved for other files like serverMiddlewares or local modules, then it should be done like TypeScript module by using both of config extensions options :
https://github.com/nuxt/typescript/blob/master/packages/typescript-build/lib/module.js#L18

pi0:
The breaking change seems introduced by 2.4.5...2.51 -- we are reverting :)

kevinmarrec:
@pi0 AFAIK routes supported extensions never used options. It was hardcoded, what breaking change was it ?

pi0:
According to #5423 CoffeeScript support was working and stopped working. But yeah you are right it was hardcoded. [srec](https://github.com/nuxt/nuxt.js/blob/v2.4.0/packages/builder/src/builder.js#L50] so strange...

kevinmarrec:
According the issue, a partial fix was found making Vuex World but not local modules, which kind of relevant as to have coffee work everywhere it nows need to be added in both extensions options

I think the fix it just about adding `.coffee` extension to `build.additionalExtensions` and nothing more.

There was anyway Indeed something changed in internals. Once upon a time there was a `resolveFiles` that seemed merging `options.extensions` inside `supportedExtensions` of the class that was storint this `resolveFiles` function

kevinmarrec:
@pi0 LGTM now, juste need title update

kevinmarrec (approved):

clarkdo (approved):

PR #6265, opened 2019-08-21
fix(server): treat https: null as https: undefined

aldarund:
Fix situation when null passed to server.https and it will go to protocol HTTP options as null and will lead to unexpected behaviour ( for it was hanging requests when appinsight-js is on )


pi0:
Would you please check failing test? Also a description in PR would be nice 😊

aldarund:
@pi0 test failure wasnt from this PR, it was some  transient failure on running cli test, i have relaunched test and it fine

pi0 (approved):

PR #6271, opened 2019-08-21
refactor(core): use hable

pi0:
## Types of changes

## Description

[jsless/hable](https://github.com/jsless/hable) is an extraction of our internal hooks implementation with the same tests and features. Pros of externalizing:
- Easier maintenance
- Using a consistent solution across different areas (I'm making a PR for using hooks in cli for example)

## Checklist:


pimlie:
Hable & Hookable dont have a fully similar interface and implementation

pi0:
@pimlie hmm I'm not sure. Can you please elaborate more about differences and potentially missing requirements?

PS: Failing test is because hable uses its own consola instance. I'll work on an option to provide custom logger.

atinux (dismissed):
YES!

clarkdo (dismissed):
LGTM

pimlie (dismissed):
Hookable:
+ flatHooks as instance method (probable safe to remove though, but who knows)
+ filter(Boolean) in addHooks

Hable:
+ deprecateHook

When moving to Hable, Nuxt should use the `deprecateHook` method instead of setting the prop directly.

Hable will add _falsy_ hooks in addHooks / flatHooks util. Actually it will also add all other non-function hooks, but thats also the case with Hookable.

pi0:
> flatHooks as an instance method (probable safe to remove though, but who knows)

This is an internal utility. So I think it is safe to not include instance.

> deprecateHook

Yeah. That's an addition of hable that we can leverage :) I'll refactor to use it.

> filter(Boolean) in addHooks

Main [hook](https://github.com/jsless/hable/blob/v2.1.0/lib/hookable.js#L14) function which is called by addHooks already prevents adding falsy hooks.

pimlie (approved):

PR #6274, opened 2019-08-21
chore(cli): accept hooks

pi0:
## Types of changes

## Description

CLI hooks are primary usable for custom CLIs built on top of @nuxt/cli (nuxt-ts). We can pass a `hooks` object as an argument to `run` command to register hooks. Only two useful hooks added for now:

- `setup({ cmd, dev, argv })`: will be called after parsing arguments. Adding possibility to modify args, command, etc and also setup env (like registering ts-node)
- `config(options)`: will be called after the config is loaded. before passing to nuxt. Adding the possibility to modify config. Like adding some extensions or modules.

## Related issues

- nuxt/typescript#79
- nuxt/typescript#80

## Checklist:


packages/cli/src/run.js
```diff
@@ -33,12 +33,20 @@ export default async function run (_argv) {
     cmd = await getCommand('dev')
   }
 
+  // Check for dev
+  const dev = argv[0] === 'dev'
+
+  // Call setup hook
+  if (typeof hooks.setup === 'function') {
+    await hooks.setup({ cmd, dev, argv })
```
clarkdo: If setup hook has been called before hook registration, does it need to be registered again?

clarkdo (commented):
Will cli hooks be able to be configured from nuxt.config?

packages/cli/src/run.js
pi0: Indeed this is a special hook which will be called before command initialization and will be never called again We can remove it from object taught if you want :)

pi0:
> Will cli hooks be able to be configured from nuxt.config?

@clarkdo  no. they are CLI level hooks that will be hooked before nuxt is being ready. As mentioned in PR description that's a internal use-case for better typescript support. (or advanced users that want to build a cli on top of nuxt)

packages/cli/src/run.js
pi0: done

kevinmarrec (commented):
@pi0 Nice ! But how do we easy get rootDir with `setup` Hook ?

pi0:
@kevinmarrec Default value of `rootDir` is `path.resolve(argv._[0] || '.')` before loading nuxt.config.js. We can also access resolved value with `config` hook.

kevinmarrec:
@pi0 but CLI options aren't already parsed so `argv._[0]` doesn't exist yet or am I missing something ?

kevinmarrec:
Nvm I read again NuxtCommand and it should be alright, argv passed to hook is parsed right ?

pimlie:
Not sure what the issue is with typescript, but wasnt the idea of NuxtCommand's that they are easily extendible by themselves? Eg hooks might just be a second way of changing the task of a command?

How much granularity do we want with hooks? Eg in Nuxt we often use `:before` / `:after` suffixes, should we do that here too to keep consistency (even when we dont want to add one of the suffixes)?

pi0:
Not much granularity is required and nuxt-ts is not adding new commands but need a hook to:
- Access parsed args (and rootDir) before command executation to be able to register ts-node
- Extend options and add extensions for runtime

Both are internal use cases and may not be even documented.

kevinmarrec:
Still if we refactor `nuxt-ts` it will only work with next Nuxt version, we'll need to bump `nuxt-ts` to a major version and document it right ?

pi0:
@kevinmarrec the hooks param is an optional addition. It is like we have to say bump to 2.9.2 or 2.10.0 to fix this issue.

kevinmarrec:
@pi0 Yeah I was more likely thinking about the `register` thing that could be really improved with the `setup` hooks. Like only using hooks :)

I mean we can't use only hooks if we want `@nuxtjs/runtime` still work if user doesn't have yet the Nuxt version that use hooks.

pi0:
Yep. Now i see your point why bumping ts module a major version for hooks 😊

kevinmarrec (approved):

Issue #6275, opened 2019-08-22
routes order issue when using `extendRoutes` option


rasikas:
### Version

[2.0.0](https://github.com/nuxt.js/releases/tag/2.0.0)

### Reproduction link

[https://codesandbox.io/s/nuxt-router-issues-5ee9l](https://codesandbox.io/s/nuxt-router-issues-5ee9l)

### Steps to reproduce

1. Have a dynamic page route ex: /products/_id/index.vue (route would be /products/1
2. Now extend a static route which is defined in the extendRoutes section of nuxt.config.js to pointing to index.vue
```
  router: {
    extendRoutes(routes, resolve) {
      routes.push({
        name: "search",
        path: "/products/find",
        component: resolve(__dirname, "pages/index.vue")
      });
    }
  }
```

### What is expected ?

in the generated routes, dynamic route is given priority (listed above in the array) over the static route (`/product/find`)

### What is actually happening?

Static routes should give priority (arrange before the dynamic routes)

### Additional comments?

If you use routes.unshift() to add the static route then it gives the priority
- or - 
if you define a page under /pages/products/find.vue  then it is given priority when generating the router.js. So I think nuxt should give similar priority when extendRoutes is used regardless of whether it is pushed to the array or not.

To see the different behavior
got to
https://5ee9l.sse.codesandbox.io/products/search   (when using routes.unshift)
https://5ee9l.sse.codesandbox.io/products/find        (when using routes.push)
https://5ee9l.sse.codesandbox.io/products/1   

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9671">#c9671</a>)</em></sub></div>

PR #6277, opened 2019-08-22
feat(utils): export sort routes

pimlie:
(Helps to) resolve:
- #6275
- https://github.com/nuxt-community/nuxt-i18n/issues/152

## Types of changes

## Description

Well, not really a feature but also not a bug fix. 

If a user calls `extendRoutes` and adds new routes, he/she is also responsible for maintaining proper order of the routes. This is quite a cumbersome task, by exposing the sortRoutes method a user can import it, add route to his/her liking and then easily sort them properly again.

(This is useful until vue-router fully supports auto path ranking)

## Checklist:


manniL:
Would it make sense to provide it as an argument for `extendRoutes`?

TheAlexLichter (approved):

galvez (approved):

Atinux:
Nice PR @pimlie 

Would it be even better if we call `sortRoutes` just after `extendRoutes` so they won't have to deal with sorting and just push the new route to the array?

pimlie:
It would be better probably, but are we 100% sure that `sortRoutes` doesnt contain any bugs at all and will sort all possible routes always perfectly? Unfortunately I think it doesnt, its too much meant to work with nuxt-routes which are often a bit simpler then the full range or possible routes with vue-router (eg regular expressions). Thats probably also why a lot of people will use extendRoutes anyway.

As vue-router seems to be already adding functionality for this, I think its best if we just provide a temporary fix until auto path ranking is merged in vue-router.

Atinux:
Did not think of this, you are right @pimlie 

Ready to merge 👍

kevinmarrec (approved):

bravokiloecho:
Is there any documentation of how to use this anywhere?

PR #6283, opened 2019-08-22
fix(builder): apply overrides from app dir only

pi0:
## Types of changes

## Description
Fixes issue introduced in #6055. We should only allow overriding from `app/` dir. If a module emits an asset outside of `.nuxt` dir (like `static/sw.js`) nuxt will wrongly override `static/sw.js` to itself and watch wrong files, causing infinity rebuilds.

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -444,7 +444,7 @@ export default class Builder {
 
       // Allow override templates using a file with same name in ${srcDir}/app
       const customAppFile = path.resolve(this.options.srcDir, this.options.dir.app, file)
-      const customAppFileExists = customAppFile.indexOf(appDir) === 0 && await fsExtra.exists(customAppFile)
+      const customAppFileExists = customAppFile.startsWith(appDir) && await fsExtra.exists(customAppFile)
       if (customAppFileExists) {
         src = customAppFile
       }
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -416,19 +416,16 @@ export default class Builder {
     // Sanitize custom template files
     this.options.build.templates = this.options.build.templates.map((t) => {
       const src = t.src || t
-
-      return Object.assign(
-        {
-          src: r(this.options.srcDir, src),
-          dst: t.dst || path.basename(src),
-          custom: true
-        },
-        typeof t === 'object' ? t : undefined
-      )
+      return {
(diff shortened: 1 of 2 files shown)


packages/builder/test/builder.generate.test.js
```diff
@@ -25,8 +25,8 @@ describe('builder: builder generate', () => {
     r.mockImplementation((...args) => `r(${args.join(', ')})`)
     fs.readFile.mockImplementation((...args) => `readFile(${args.join(', ')})`)
     fs.outputFile.mockImplementation((...args) => `outputFile(${args.join(', ')})`)
-    jest.spyOn(path, 'join').mockImplementation((...args) => `join(${args.join(', ')})`)
-    jest.spyOn(path, 'resolve').mockImplementation((...args) => `resolve(${args.join(', ')})`)
+    jest.spyOn(path, 'join')
+    jest.spyOn(path, 'resolve')
```
pi0: I had to prevent spies on this test. It generates artifacts like `r(foo, bar)` which are not comparable to check template being subdir of app dir

packages/builder/src/builder.js
```diff
@@ -437,15 +434,20 @@ export default class Builder {
...
       const customTemplateIndex = customTemplateFiles.indexOf(file)
       const customTemplate = customTemplateIndex !== -1 ? this.options.build.templates[customTemplateIndex] : null
       let src = customTemplate ? (customTemplate.src || customTemplate) : r(this.template.dir, file)
+
       // Allow override templates using a file with same name in ${srcDir}/app
-      const customPath = r(this.options.srcDir, this.options.dir.app, file)
-      const customFileExists = await fsExtra.exists(customPath)
-      src = customFileExists ? customPath : src
+      const customPath = path.resolve(this.options.srcDir, this.options.dir.app, file)
+      const customFileExists = customPath.indexOf(appDir) === 0 && await fsExtra.exists(customPath)
```
pi0: The real fix is actually here: `customPath.indexOf(appDir) === 0`

pi0: I changes `r` to `path.resolve` as aliases are meaningless in this context. We resolve relative to appDir.

pimlie: Using `startsWith` might be just a bit more readable?

kevinmarrec (approved):
Nice ! LGTM :)

pimlie (approved):

PR #6285, opened 2019-08-22
feat(builder): `extendPlugins` option and `builder:extendPlugins'` hook

galvez:
Currently there's no way to determine the final ordering of user-space plugins in relation to plugins registered automatically by modules (and not all modules necessarily use `addPlugin()`, which uses `unshift()`). The problem is complex and likely to have other takes, but as @pi0 said:

Screen Shot 2019-08-26 at 17 19 08

So this introduces a new `build:extendPlugins` configuration hook, similar to `extendRoutes()`, that allows users to completely modify the order of all plugins before the build.


packages/builder/test/builder.plugin.test.js
```diff
@@ -18,16 +18,21 @@ describe('builder: builder plugins', () => {
...
+  test('should normalize plugins', async () => {
     const nuxt = createNuxt()
     nuxt.options.plugins = [
-      '/var/nuxt/plugins/test.js',
       '/var/nuxt/.nuxt/foo-bar.plugin.client.530b6c6a.js',
       { src: '/var/nuxt/plugins/test.server', mode: 'server' },
       { src: '/var/nuxt/plugins/test.client', ssr: false }
     ]
-    const builder = new Builder(nuxt, BundleBuilder)
+    nuxt.options.hooks = {
```
pi0: We could also use `nuxt.hook` :)

packages/builder/src/builder.js
```diff
@@ -225,7 +225,7 @@ export default class Builder {
     }
 
     // Plugins
-    this.plugins = Array.from(this.normalizePlugins())
+    this.plugins = Array.from(await this.normalizePlugins())
```
manniL: Do we have to do sth. with regards to error handling here?

manniL:
Besides my annotation in the code, I'd like to see a more illuminating description for the PR, e.g., why introducing the change, what benefits, maybe even potential drawbacks.

For other contributors (besides the core team), the description isn't helpful as is and not transparent.

pimlie (dismissed):
Haha, +1 for the discord screenshot of Pooya approving the need for this hook

TheAlexLichter (dismissed):

atinux (dismissed):

pimlie:
By request of @galvez I've added support for `options.extendPlugins` which follows the same doctrine as `options.router.extendRoutes`

pimlie (approved):

pi0 (approved):

PR #6287, opened 2019-08-23
feat(vue-app): support configurable features

pimlie:
~~This PR is only a draft because we didn't discuss this yet before. Feature-wise I think its complete, maybe just need to add some tests.~~

## Types of changes

## Description

The Nuxt Vue-app has quite a bit of features which until this PR were _always_ included, whether you used them or not. This PR defines a list of features which are used to either include or omit the corresponding code in the Vue-app templates. This makes it possible for a user to further optimize their bundle sizes.

Here is the list of features which are toggleable:
```js
  features: {
    router: true, // not implemented
    store: true,
    layouts: true,
    meta: true,
    middleware: true,
    transitions: true,
    deprecations: true,
    validate: true,
    asyncData: true,
    fetch: true,
    clientOnline: true,
    clientPrefetch: true,
    clientUseUrl: false, // this is a bit of an odd one, but using URL should eg be ok for modern mode already
    componentAliases: true,
    componentClientOnly: true
  }
```

I've included a copy of the hello world example app with all (supported) features disabled.

App bundle size of Hello world app
image

App bundle size of Minimal-features Hello world app:
image
image

This PR was also partially given in by the benchmark discussions on discord. You could argue that disabling all these settings is not very usefull and I will probably agree with that. But giving users the possiblity to further optimize their bundle sizes is a nice feature I think and worth the extra work with regards to maintaining the vue-app templates.

Disabling features is considered highly experimental and for advanced users only.

Note: The above bundle size of the Hello world app has already improved slightly compared to the current dev branch. This is mainly due to some optimisations in `./utils.js`. Eg `urlJoin` is/was always included in the app bundle but thats only used in `server.js`

## Checklist:


packages/builder/src/builder.js
```diff
@@ -232,6 +232,14 @@ export default class Builder {
       this.resolveMiddleware(templateContext)
     ])
 
+    if (this.options.dev && this.options.build.indicator) {
```
pi0: indicator will be auto disabled for non dev with options.js

packages/vue-app/src/index.js
```diff
@@ -8,15 +8,15 @@ export const template = {
...
     'index.js',
-    'middleware.js',
+    // 'middleware.js',
     'router.js',
     'router.scrollBehavior.js',
     'server.js',
     'utils.js',
     'empty.js',
-    'components/nuxt-build-indicator.vue',
+    // 'components/nuxt-build-indicator.vue',
```
pi0: For long term: I wanted to export a function instead that takes builder instance to dynamically provide template. Refactoring logic of feature related options here.

packages/vue-app/template/App.js
```diff
@@ -67,10 +83,12 @@ export default {
     // add to window so we can listen when ready
     if (process.client) {
       window.<%= globals.nuxt %> = <%= (globals.nuxt !== '$nuxt' ? 'window.$nuxt = ' : '') %>this
+      <% if (features.client.online) { %>
```
pi0: Funny fact: As a nuxt developer, wasn't aware we have it 😂😂😂

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -83,15 +83,18 @@ export default class SSRRenderer extends BaseRenderer {
...
     // Inject head meta
-    const m = renderContext.meta.inject()
-    let HEAD =
-      m.title.text() +
-      m.meta.text() +
-      m.link.text() +
-      m.style.text() +
-      m.script.text() +
-      m.noscript.text()
+    const m = renderContext.meta && renderContext.meta.inject()
```
pi0: Can we also rename m to meta for more readability?

packages/vue-app/template/App.js
manniL: It came together with smart prefetching :relaxed:

manniL:
Amazing idea :+1:

packages/vue-app/template/index.js
```diff
@@ -1,7 +1,7 @@
 import Vue from 'vue'
-import Meta from 'vue-meta'
-import ClientOnly from 'vue-client-only'
-import NoSsr from 'vue-no-ssr'
+<% if (features.meta) { %>import Meta from 'vue-meta'<% } %>
+<% if (features.components.clientOnly) { %>import ClientOnly from 'vue-client-only'<% } %>
+<% if (features.components.noSsr) { %>import NoSsr from 'vue-no-ssr'<% } %>
```
manniL: Doesn't fit `noSsr` under `deprecations`?

pimlie: Yes indeed!

pimlie:
Added the suggestions and implemented store/layout features. Also had some wrong/old features in the nuxt.config of the minimal example, so size is even smaller now (image of app bundle size is updated).

pi0:
Really nice enhancement. I wish we could also support disabling router. But seems not an easy task for initial PR.

pimlie:
I looked into making the router configurable as well but almost every function in `client.js` makes use of `from` / `to` params and it seems just to much integrated into all the logic. Dont think it will be an easy task to support disabling the router in the current code. A complete rewrite of client.js with optional router disabling in mind will probably be faster.

packages/vue-app/template/client.js
```diff
@@ -598,42 +653,62 @@ function addHotReload($component, depth) {
...
+    }
     <% } %>
+
     callMiddleware.call(this, Components, context)
     .then(() => {
+      <% if (features.layouts) { %>
       // If layout changed
-      if (depth !== 0) return Promise.resolve()
+      if (depth !== 0) {
+        return true
```
pimlie: Just a heads-up about this change (and some more below), returning in a thenable chain should always result in a promise.

Try eg: 
```js
new Promise(r => setTimeout(r, 1000)).then(() => true).then((v) => console.log('Done', v))
```

TheAlexLichter (dismissed):

Atinux:
Great work @pimlie, I had this in mind for Nuxt 3.

Ideally, the features enable themselves when used (when detecting the directory associated or defined in the config). Also, I am afraid that some module may not work if the user disable some features 🤔

pimlie:
> Ideally, the features enable themselves when used (when detecting the directory associated or defined in the config).

Well, yes/no. I can imagine there are also times you just want to disable a feature regardless whether you have certain files in your project. Eg with the current store implementation I did you can have a store module in your project, but if you toggle the feature to false it will still be disabled.

> Also, I am afraid that some module may not work if the user disable some features 🤔

True, for the moment thats really up to the user to think about. Thats why I mentioned _experimental and for advanced users only_. 

I dont think we should launch this pr as a great new feature anyway, not as long as we cant provide a router feature. So I would propose to do a silent launch, learn from it and then improve by maybe taking the approach you had in mind for v3

manniL:
> Well, yes/no. I can imagine there are also times you just want to disable a feature regardless whether you have certain files in your project. Eg with the current store implementation I did you can have a store module in your project, but if you toggle the feature to false it will still be disabled.

Definitely, the user prefence should take precedence! But disabling/enabling features if certain conditions are present and the user didn't set the config otherwise would be nice.

examples/minimal-features/nuxt.config.js
```diff
@@ -0,0 +1,35 @@
...
+    validate: false,
+    asyncData: false,
+    fetch: false,
+    client: {
+      online: false,
+      prefetch: false,
+      useUrl: true
+    },
+    components: {
+      aliases: false,
```
pi0: what do you think about flattened options like `compoentAliases`? This way we don't need to always worry about falsy value in the path (Like `components: false`)

pi0 (dismissed):
A REALLY REALLY nice improvement to support minimal builds for nuxt. Can't wait to also have router disabling support. With nuxt 3 we can have nuxt to be used like:

- Create `app.vue`
- `npx nuxt dev`
- TADA! A fully minimal/universal/efficient vue app is ready!

I think we should be better to merge this ASAP to avoid merge conflicts.

pimlie:
The features config is now a flat object.

Another feature we could add is `errorHandling`. I noticed that after merging the dev branch the app size had grown again due to the changes of the loop prevention for the error component (see latest image in start post). Tried to do that quickly but thats also quite a big change.

@clarkdo Would it be possible somehow to instruct webpack to always create a separate chunk for the complete Nuxt vue-app? I've copied the size limit test and used the `unicode-base` fixture as a size test for a minimal app, but I think for the minimal features it would be much more interesting if we could just test the vue-app bundle. I also lowered the allowed threshold to 1%, as big changes to the vue-app should probably be hidden behind new feature flags.

clarkdo:
@pimlie If you mean splitting the chunk only for testing purpose, I think it's a good idea. You can implement it by adding a `cacheGroup` and filtering files inside `.nuxt`(buildDir).

packages/vue-app/template/client.js
```diff
@@ -599,42 +654,62 @@ function addHotReload($component, depth) {
...
+      }
+
       let layout = Component.options.layout || 'default'
       if (typeof layout === 'function') {
         layout = layout(context)
       }
-      if (this.layoutName === layout) return Promise.resolve()
+      if (this.layoutName === layout) {
+        return true
```
pi0: Does `true` value means anything special? Before it was resolving to undefined.

pimlie: No, actually not. Have removed that

pi0 (dismissed):

pimlie:
@clarkdo Thanks, could you please verify if I implemented it correctly?

And I would appreciate your/anyone's help with resolving the error with unicode-base, dont really understand why that now gives an error? The file exists on the fs and is listed in the manifest, does one of you maybe have any idea why jsdom cant find it? Only change for that test is that I added `modern: 'server'` in nuxt.config.

clarkdo:
@pimlie Looks perfect, you can double verify it by having a look at analyze report.

pi0 (approved):

clarkdo (approved):

TheAlexLichter (approved):

PR #6290, opened 2019-08-24
test: disable terser/minify by default

pimlie:
## Types of changes

## Description

We have probably all wasted hours of our time waiting for terser to run on the test fixtures. Thats completely useless in most cases because we dont have any need for it.

This PR speeds up the building of the fixtures with ~40 seconds for me (from 100-110s to 65s)

On CircleCI, the nighty build from last night:
image

This pr:
image

## Checklist:

- [x] All new and existing tests are passing. **YES AND MUCH QUICKER**

### DIFF
--- a/test/utils/nuxt.js
+++ b/test/utils/nuxt.js
@@ -33,5 +33,14 @@ export const loadFixture = async function (fixture, overrides) {
   config.dev = false
   config.test = true
 
+  // disable terser to speed-up fixture builds
+  if (config.build) {
+    if (!config.build.hasOwnProperty('terser')) {
+      config.build.terser = false
+    }
+  } else {
+    config.build = { terser: false }
+  }
+
   return defaultsDeep({}, overrides, config)
 }
--- a/test/utils/nuxt.js
+++ b/test/utils/nuxt.js
@@ -35,7 +35,7 @@ export const loadFixture = async function (fixture, overrides) {
 
   // disable terser to speed-up fixture builds
   if (config.build) {
-    if (!config.build.hasOwnProperty('terser')) {
+    if (!config.build.terser) {
       config.build.terser = false
     }
   } else {
--- a/test/fixtures/async-config/nuxt.config.js
+++ b/test/fixtures/async-config/nuxt.config.js
@@ -1,6 +1,9 @@
 const createData = async () => {
   await new Promise(resolve => setTimeout(resolve, 500))
   return {
+    build: {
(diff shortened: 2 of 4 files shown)


test/utils/nuxt.js
```diff
@@ -33,5 +33,14 @@ export const loadFixture = async function (fixture, overrides) {
   config.dev = false
   config.test = true
 
+  // disable terser to speed-up fixture builds
+  if (config.build) {
+    if (!config.build.hasOwnProperty('terser')) {
```
pi0: isn't config.build.terser enough?

pi0 (commented):
Great! BTW why we need terser for modern and async-config?

pimlie:
- modern needs it because it checks the bundle if it contains modern syntax. If we dont use terser on that then you have to match all kind of whitespace stuff.

This is part of the output from one of the failing tests when I didnt had terser enabled:
```
    /* harmony default export */ var pagesvue_type_script_lang_js_ = ({
      test() {
        var test = {
          build: true,
          test: true,
          arrow: () => {
            return 'build test';
          }
        };
        return {
          test
        };
      }·
    });

```

while the test searches if the output contains `arrow:()=>\"build test\"`. We could change it to just check the `arrow: () => {` line, should be ok I guess?

- async-config needs it due to the `async-config.size-limit` test. Not sure if that could be changed or not

pi0:
Something more general: Why not automatically optimizing defaults when env is test (config.test) This can make module tests faster.

test/utils/nuxt.js
pimlie: You mean instead of `hasOwnProperty`? Probably yes, will change that

pi0:
> async-config needs it due to the async-config.size-limit test. Not sure if that could be changed or not

I'm not sure if that test is even relevant to this test. Please add a comment above it or skip for now. After having perf reports we only build a small fixture with terser to benchmark.

pi0:
For modern, if we update tests that would be great. Saving even more time :)

pimlie:
Yes, we broke the 200s mark on Circle CI!

image

pimlie:
> Something more general: Why not automatically optimizing defaults when env is test (config.test) This can make module tests faster.

I find it sometimes confusing when Nuxt silently changes default settings based on the environment.

Think a setting like this would be much more suited to be changed in one of the test-utils packages.

pi0 (approved):

PR #6292, opened 2019-08-24
feat(vue-app): add `prefetch` prop to `<nuxt-link>`

### LINKED ISSUE #6135: Allow `prefetch` prop for nuxt-link to enable prefetching for specific links

nuxt-link currently allows disabling prefetching with a specific link using a no-prefetch prop, but on the other hand, there is no way to use reversed behavior when links prefetching is disabled globally via nuxt.config with prefetchLinks: false

<nuxt-link prefetch>

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9559">#c9559</a>)</em></sub></div>

lmichelin:
## Type of changes

- ~~Bug fix~~ (a non-breaking change which fixes an issue)
- ~~Breaking change~~ (fix or feature that would cause existing functionality to change)

## Description

Fixes #6135

`<nuxt-link>` currently allows disabling prefetching with a specific link using a no-prefetch prop, but on the other hand, there is no way to prefetch a specific link when links prefetching is disabled globally via `nuxt.config.js` with `prefetchLinks: false`.

This PR adds a `prefetch` prop to `<nuxt-link>` which can be used when `prefetchLinks` is set to `false`.

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1550)


packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -25,18 +25,23 @@ const observer = window.IntersectionObserver && new window.IntersectionObserver(
...
-  props: {
+  props: {<% if (router.prefetchLinks) { %>
     noPrefetch: {
       type: Boolean,
       default: false
-    }<% if (router.linkPrefetchedClass) { %>,
+    }<% } else { %>
+    prefetch: {
+      type: Boolean,
+      default: false
```
pimlie: Instead of adding a second prop, why not change the default value of the existing prefetch prop depending on the global prefetch status?

```suggestion
     default: <%= router.prefetchLinks ? 'true' : 'false' %>
```

lmichelin: the existing prop is `no-prefetch`. If we want to keep only the new `prefetch` prop, it would introduce a breaking change 😕

lmichelin: or maybe we can make the 2 props coexist when `router.prefetchLink` is set to `true`.
`no-prefetch` would have the same effect as `:prefetch="false"` to keep the compatibility.

pimlie: Ah right, sry missed that. In that case I think we should deprecate `no-prefetch` in favor of prefetch. If you set the default value of no-prefetch to undefined then you can check if its set and print a deprecation message (just make sure to do it once for all links and only in dev mode)

But opinions vary so I'd recommended to wait until @pi0 gives his approval on the deprecation :)

manniL: I think it makes sense to be able to toggle both behaviors:

1. Having all links prefetched by default and disable single ones (think of ToS, Privacy and so on)
2. Having all links w/o prefetch and only prefetch a few of them.

pimlie: Maybe Im not understanding this correctly, but cant we achieve that with a single prop?

```js
prefetch: {
  default: true
}
...
// prefetched
<nuxt-link> 
// not prefetched
<nuxt-link :prefetch="false"> 

/* or */

prefetch: {
  default: false
}
..
// not prefetched
<nuxt-link> 
// prefetched
<nuxt-link :prefetch="true"> 
```

pi0: I agree to have `prefetch` option (with a configurable default) and also a `no-prefetch` shortcut for more semantic usage && preventing breaking changes.

lmichelin: @pi0 should the `no-prefetch` prop be deprecated ?

pi0: I would say no as of better DX. At least not annoying users until nuxt3 if we even want to deprecate. /cc @Atinux

pimlie: One prop provides better DX in my mind. In that case if you set the prop explicitly that value will always be used, regardless of the global config.

Also if we have two props and you set them both: `<nuxt-link no-prefetch="true" prefetch="true">`, what happens then?

pi0: Only first takes precedence. BTW I just noticed that vue3 is going to deprecate boolean props. So there are no DX benefits. (`<foo no-prop>` ~> `foo no-pro="true"`)

pi0: So the is only annoying exciting users or not by this deprecation warn :)

lmichelin: Something like that ?

```js
if (this.prefetch === this.noPrefetch) {
  console.warn("The `prefetch` and `no-prefetch` props can't be used simultaneously.")
}
```

I'm not convinced... 😕 And it would be difficult to detect with default values...
I think the deprecation warning is the best option 😉

pi0: Nuxt 2 already comes with a default value of `true` for prefetch and vue 2 which allows boolean props. This has negative DX for _current_ users. Deprecation should have a valid reason. For nuxt 3 we can occentially just unsupport `no-` prop or start warning. But for nuxt 2 it is somehow annoying imo.

pimlie: If they think its annoying they would be able to turn it off with `features.deprecations = false` from the app features pr :stuck_out_tongue: 

Anyway, improving this behaviour in Nuxt 3 is ok with me!

pimlie (changes requested):
My apologies to have to do this, but I have just added a compileTemplate test helper method I would like you to ask to use to add unit tests (using vue-test-utils) instead of adding new fixtures. Adding new fixtures is not preferred as building each fixtures takes at least 10 seconds to the total test time.

You can have a look at [nuxt-loading.test.js](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/test/nuxt-loading.test.js) for an example of an existing component unit test.

For this pr you could use the compileTemplate function as follows:
```js
// nuxt-link.client.prefetch.test.js
const compiledTemplatePath = await compileTemplate('components/nuxt-link.client.vue', 'nuxt-link.client.prefetch.vue', { router: { prefetch: true } })

// nuxt-link.client.noprefetch.test.js
const compiledTemplatePath = await compileTemplate('components/nuxt-link.client.vue', 'nuxt-link.client.noprefetch.vue', { router: { prefetch: false } })
```
For quick reference, here is the link to the [vtu docs](https://vue-test-utils.vuejs.org/)

packages/vue-app/template/client.js
```diff
@@ -17,7 +17,7 @@ import {
   globalHandleError
 } from './utils.js'
 import { createApp, NuxtError } from './index.js'
-import NuxtLink from './components/nuxt-link.<%= router.prefetchLinks ? "client" : "server" %>.js' // should be included after ./index.js
+import NuxtLink from './components/nuxt-link.client.js' // should be included after ./index.js
```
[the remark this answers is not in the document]
lmichelin: I don't completely understand what would be these three states:

- `prefetchLinks: true` (default) -> import `nuxt-link.client.js`
- `prefetchLinks: false` -> import `nuxt-link.server.js` (`prefetch` is completely disabled)
- `prefetchLinks: undefined` -> import `nuxt-link.client.js` (`:prefetch="true"` would work) 

Is that what you mean?

The code would be like that:

```js
import NuxtLink from './components/nuxt-link.<%= router.prefetchLinks === false ? "server" : "client" %>.js'
```

But I think it's really unclear 😕 because __the user would have to set `prefetchLinks: undefined` explicitly__. Your app.feature PR would really help to solve this 🙂.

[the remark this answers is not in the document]
pimlie: In your example I would probably switch the meaning of `false` and `undefined`. But still not optimal indeed, not sure what the best course of action is here. Lets wait for someone else to chime in :)

Atinux:
What about `prefetchLinks: “on-demand”`?

packages/vue-app/template/client.js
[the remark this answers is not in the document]
manniL: I'd also say that switching the behavior and removing the prefetch part completely when set to `undefined` (or maybe even `null` to make it more intentional?) would make sense.

However, we could also work with string here.

lmichelin:
> What about `prefetchLinks: “on-demand”`?

Maybe the best option is to wait for the PR https://github.com/nuxt/nuxt.js/pull/6287 to be merged?

packages/vue-app/template/client.js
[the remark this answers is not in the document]
lmichelin: @pimlie I just rebased my branch to include the latest changes from `dev`. I think the https://github.com/nuxt/nuxt.js/pull/6287 solves this conversation. To disable prefetch completely and import `nuxt-link.server.js`, users will have to set `features.clientPrefetch` to false.

pimlie (dismissed):
Looks good to me.

Thanks for the recurring work you put into this pr!

lmichelin:
> My apologies to have to do this, but I have just added a compileTemplate test helper method I would like you to ask to use to add unit tests (using vue-test-utils) instead of adding new fixtures. Adding new fixtures is not preferred as building each fixtures takes at least 10 seconds to the total test time.
> 
> You can have a look at [nuxt-loading.test.js](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/test/nuxt-loading.test.js) for an example of an existing component unit test.
> 
> For this pr you could use the compileTemplate function as follows:
> 
> ```js
> // nuxt-link.client.prefetch.test.js
> const compiledTemplatePath = await compileTemplate('components/nuxt-link.client.vue', 'nuxt-link.client.prefetch.vue', { router: { prefetch: true } })
> 
> // nuxt-link.client.noprefetch.test.js
> const compiledTemplatePath = await compileTemplate('components/nuxt-link.client.vue', 'nuxt-link.client.noprefetch.vue', { router: { prefetch: false } })
> ```
> 
> For quick reference, here is the link to the [vtu docs](https://vue-test-utils.vuejs.org/)

I tried to do some tests but I faced some issues mounting the `<nuxt-link>` component with vue-test-utils:

image

pimlie:
Its probably because `extends: Vue.component('RouterLink')` fails. Did you add a stub or fake router-link component? See: https://vue-test-utils.vuejs.org/api/components/#routerlinkstub

lmichelin:
I tried but it has no effect 😕

pimlie:
What about something like this?
```js
import { createLocalVue } from '@vue/test-utils'

const localVue = createLocalVue()
// register stub component so nuxt-link extends its render fn
localVue.component('RouterLink', {
  render: h => h('a')
})
```

pi0 (dismissed):
LGTM

lmichelin:
> What about something like this?
> 
> ```js
> import { createLocalVue } from '@vue/test-utils'
> 
> const localVue = createLocalVue()
> // register stub component so nuxt-link extends its render fn
> localVue.component('RouterLink', {
>   render: h => h('a')
> })
> // load nuxt-link
> ```

Still face the issue 😕, I made a WIP commit to show you the code :) (I also needed to fix a bug in the `compileTemplate` helper)

pimlie:
Hmm, could you try to explicitly set the `router-link` component as a stub in the mounting options as described [here](https://vue-test-utils.vuejs.org/api/options.html#stubs)?

lmichelin:
Still facing the warning 😕
I tried : `stubs: ['router-link']`, `stubs: ['RouterLink']`, `stubs: { 'router-link': true }`

pimlie:
@lmichelin Hope you dont mind, but as I didnt knew the correct answer myself I just tried fixing this on your branch. The tests should work now.

Feel free to improve them further if you wish!

lmichelin:
Nice! Thanks a lot for your help 😃

lmichelin:
And btw do you agree with the fix in `compileTemplate` https://github.com/nuxt/nuxt.js/pull/6292/commits/374363a214a37d1c49c45323bcc82b804a9db5b5

pimlie (dismissed):

Atinux:
Great work here @lmichelin 👏 

One thing came to my mind when I read `prefetch` and `no-prefetch` props, what about deprecating `no-prefetch` in favour of `:prefetch="false"`?

lmichelin:
I asked the question here https://github.com/nuxt/nuxt.js/pull/6292#discussion_r317368260 but @pi0 preferred not to deprecate it.

lmichelin:
> @lmichelin Let us know btw if you think this PR is now ready or if you wish to make some other changes, thanks :)

It's ok for me!

pi0 (approved):

PR #6297, opened 2019-08-25
feat(vue-app): function watchQuery

clarkdo:
## Types of changes

## Description

Be merged after 2.9.2
## Checklist:


pimlie:
I fear the watchQuery function indeed very much 😆

clarkdo:
？

pimlie:
See the PR title

clarkdo:
Opps😱

packages/vue-app/template/client.js
```diff
@@ -279,6 +279,7 @@ async function render(to, from, next) {
   // Get route's matched components
   const matches = []
   const Components = getMatchedComponents(to, matches)
+  const instances = getMatchedComponentsInstances(to)
```
pimlie: Could this be moved to the watchQuery `else if` (eg using memoization)?

The need for `instances` is for such a specific use-case only that we shouldnt always spent those cpu cycles retrieving them when there is a reasonable chance we dont need them

clarkdo: Done.

PR #6300, opened 2019-08-25
test: add jest roots for less greedy test search

pimlie:
## Types of changes

## Description

This should save at least 1 second on running tests. By default jest searches for modules/tests on `<rootDir>`. That is normally pretty ok, but (at least on linux) jest uses the following command to saerch for modules: `find <rootDir> -type f -iname *.snap -o -iname *.js -o iname *.json`

This is pretty greedy and also searched all examples, all node_modules dirs (main and from examples). So it will find _a lot_ of files, on my machine that takes about 1 second. On circleci that will probably be multiple seconds because find is quite i/o intensive (thats also how I noticed this command was being run because it was showing on iotop for some time).

Also it appears that the time it takes for the find command to run does not count for the total run time that jest reports. Eg if you run the tests like this `time yarn test` then the time reported by jest and by the time command should be closer to each other now.

## Checklist:

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,8 +6,8 @@ module.exports = {
   forceExit: true,
 
   roots: [
-    "<rootDir>/packages",
-    "<rootDir>/test"
+    '<rootDir>/packages',
+    '<rootDir>/test'
   ],
 
   // https://github.com/facebook/jest/pull/6747 fix warning here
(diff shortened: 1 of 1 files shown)


clarkdo (dismissed):

clarkdo:
LGTM except lint error

pimlie:
Oops! Thanks for the heads-up

PR #6301, opened 2019-08-25
test: remove unnecessary generate

clarkdo:
## Types of changes

## Description

Good catch from @pimlie , some `generate` can be removed since they can be covered by unit test and a default fixture test.

## Checklist:


packages/config/test/options.test.js
```diff
@@ -112,11 +112,21 @@ describe('config: options', () => {
     expect(pageTransition.appear).toEqual(true)
   })
 
-  test('should return 404.html as default generate.fallback', () => {
+  test('should return 200.html as default generate.fallback', () => {
```
clarkdo: fallback is also tested in `generator.gen.test.js` -> `should write fallback html after generate`

TheAlexLichter (dismissed):

test/unit/basic.generate.test.js
```diff
@@ -204,11 +204,6 @@ describe('basic generate', () => {
     })
   })
 
-  test('nuxt re-generating with no subfolders', async () => {
-    generator.nuxt.options.generate.subFolders = false
-    await expect(generator.generate({ build: false })).resolves.toBeTruthy()
-  })
-
   test('/users/1.html', async () => {
```
pimlie: This test can be removed now, but please also explicitly add `nuxt.options.generate.subFolders = false` to one of the unit tests for generator. Otherwise we only have implicit tests and those will only cause problems in the long run

Actually the unit tests are somewhat confusing already because the default setting for subFolders is `true` in Nuxt, but due to how these unit tests have been setup they have different default settings. I would strongly suggest to use a single `createNuxt` method which is shared by all unit tests in all packages which have the default Nuxt settings as base on which we then extend. Similar like the loadFixture method for the fixtures.

The main reason for this is that in the real world these methods will never run without the full config anyway, so by recreating a small subsection of the nuxt config just for these tests you could unwillingly create a different use-case then the real world and thereby (partially) invalidate them.

clarkdo: I added subFolders false in one test.

But for unit test, the test case should care up/downstream functionality, like unit test for `generator` does't depend on the `config` default options. This should be responsibility of fixtures for e2e tests.

pimlie: Thanks, understood :)

pimlie (dismissed):

TheAlexLichter (approved):

PR #6302, opened 2019-08-25
test: add test for modern bundle size

clarkdo:
## Types of changes

## Checklist:


pimlie:
I've wondered for some time, but what exactly is being tested here? Just how big the bundle sizes are (as the title of this pr suggests) or is there some server setting for first/max payload size? Where does the `196` come from, i.e. the test talk about size boundaries but how/where are those size boundaries defined? Is it a config number or is it just an arbitrary number specific for the async-config fixture?
And to continue my questions, if its just about bundle size why do you have to through html/cheerio? Cant you look them up on the fs instead?

clarkdo:
The original test is from @manniL ,I suppose the reason why used `cheerio` is for find all script in html easily. And the test is as an bundle size indicator for new feature came in for monitoring Nuxt basic bundle size.

pimlie:
Thanks for your explanation. Would it maybe be possible to use `modern: 'client'` and just load the client/modern manifests from the dist/server folder and then find the largest file included in the manifest? Saves the trouble of starting a server, especially one on `0.0.0.0`

Also assign the 196 to a variable with a proper defined name so its easier to understand why 196 is 196 and eg not 195 or 197? If I understand correctly its a bit arbitrary though so maybe something like `const maxKnownClientBundleSize = 196`

clarkdo:
Checking manifest with fs api is OK for me, even modern ‘client’ is not necessary since server mode also has an accurate manifest file, but I think we’d better discuss with @manniL , I suppose this test is for async config and because it’s almost an emtpy app, so @manniL put size-limitation inside it. And the value 196 is not a random value, it’s the current bundle size of this almost emtpy app which means the size is equal to nuxt size, I’ll extract them into more readable and meaningful variables.

pimlie:
Not trying to be a know it all, but the largest bundle size seems to be currently `152K`. This means a PR could add 40 kilobyte to the bundle and this test still doesnt trigger. Which doesnt make it really effective I guess.

Maybe we could store the largest bundle size on each Nuxt.js release (some sort of snapshot) and check against that value?

clarkdo:
The bundle size doesn’t mean the largest bundle file,the largest mainly contains node_moudles, there are also other bundle files like app which contains other  nuxt vue-app and some shared components

pimlie:
I ran `./packages/cli/bin/nuxt-cli.js start test/fixtures/async-config/` and checked the Network console. There is no `app.js` etc, just the bundle files from dist/client.

clarkdo:
Try analyze ? I’m on mobile right now, I’ll have a look tomorrow

pimlie:
image

clarkdo:
app.is is above commons.app.js in your screenshot which is 40K

pimlie:
Ah right, it counts the sum. Sorry about that.

Maybe this test should also fail when the bundle is too small, so keep the size within an upper & lower boundary instead of just an upper. Eg that should prevent/detect any issues with my app.features pr

clarkdo:
Yep, a range could be a good idea

pimlie:
Thanks for your patience. While we are at it, should we just test individual bundle file sizes as well? The sizes are distinguishable enough that we can say that we eg expect _a_ file of less then 500 bytes, one between 2.3 and 2.5KB, around 40KB and one around 152KB and in total not more then 196.

Otherwise the increase of one file could be hidden by the decrease of another.

clarkdo:
It could be more accurate as you said, but it could also be a correct case because of webpack optimization, like moving some shared files from app.js into commons or splitting one chunk file into multiple files

clarkdo:
@pimlie I change from server and cherrio to read manifest and fs api

pimlie (dismissed):
Thanks, looks good to me

test/unit/async-config.size-limit.test.js
```diff
@@ -1,41 +1,33 @@
...
+    const LEGACY_JS_RESOURCES_KB_SIZE = 195
+    const legacyResourcesSize = await getResourcesSize('client')
+    expect(legacyResourcesSize).toBeLessThanOrEqual(LEGACY_JS_RESOURCES_KB_SIZE * 1.05)
+    expect(legacyResourcesSize).toBeGreaterThanOrEqual(LEGACY_JS_RESOURCES_KB_SIZE * 0.95)
   })
 
-  it('should stay within the size boundaries', () => {
-    const responseSizeBytes = responseSizes.reduce((bytes, responseLength) => bytes + responseLength, 0)
-    const responseSizeKilobytes = Math.ceil(responseSizeBytes / 1024)
-    // Without gzip!
```
manniL: We should keep that piece of information somewhere around

TheAlexLichter (changes requested):

test/unit/async-config.size-limit.test.js
clarkdo: Do you mean gzip?

manniL: I mean that it should be noted somewhere the we compare ungzipped file sizes :relaxed:

clarkdo: It makes sense when we use http, but now it’s reading from fs directly, is it still necessary? Or we can add test for gzip or brotli?

manniL: I see what you mean but from my experience, most people think of gzipped & minified numbers when it comes to benchmarking bundle sizes. So the comment wouldn't hurt here.

Gzip/brotli tests would help us as they would provide accurate base values for a basic nuxt application (without taking #6287 into consideration here)

clarkdo: I’ll add compression size test

TheAlexLichter (dismissed):

test/unit/async-config.size-limit.test.js
```diff
@@ -1,41 +1,67 @@
...
+    const fileContent = await fs.readFile(file)
+    sizes.gzip += await compressSize(fileContent, gzip) / 1024
+    sizes.brotli += await compressSize(fileContent, brotli) / 1024
+  }
+  return sizes
+}
+
+describe('nuxt basic resources size limit', () => {
+  expect.extend({
+    toBeWithinSize (received, size) {
```
pimlie: Very nice idea / clean solution to extend expect for this ❤️

PR #6312, opened 2019-08-26
chore: upgrade circleci config to v2.1

pimlie:
## Types of changes

## Description

This upgrades the current circleci config to `v2.1`. One of the main benefits of `v2.1` is that builds can be automatically cancelled when a new build is triggered, [see here for more info](https://circleci.com/docs/2.0/build-processing/). This could save some time in some cases.

This config is based on the config I have been using for `nuxt/vue-meta` and `nuxt/press`

## Checklist:


pimlie:
Btw, the tests for this PR are still using the old config

.circleci/config.yml
```diff
@@ -1,48 +1,60 @@
...
-  environment:
-    NODE_ENV: test
-    NODE_OPTIONS: --max_old_space_size=4096
+commands:
+  attach-project:
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/project
+  store-coverage:
```
pimlie: I realize this is probably the wrong name, but I wanted something different then eg `store-test-results`. Let me know if you have a suggestion

.circleci/config.yml
```diff
@@ -1,48 +1,60 @@
-version: 2
+version: 2.1
+
+executors:
+  node:
```
clarkdo: How about node-latest ?

.circleci/config.yml
```diff
@@ -1,48 +1,60 @@
...
+    parameters:
+      browsers:
```
clarkdo: How do you think make this as string parameter `variant` which default value is "" and `-browsers` for specific case ?

.circleci/config.yml
clarkdo: store-test-results ?

.circleci/config.yml
pimlie: If we would do something with the node version then version should be a param on the executor

.circleci/config.yml
pimlie: Not really usefull, there's really only `-browsers`: https://hub.docker.com/r/circleci/node

clarkdo: https://circleci.com/docs/2.0/circleci-images/#language-image-variants

clarkdo: I think it could make `<<# parameters.browsers >>-browsers<</ parameters.browsers >>` more readable with `<< parameters.variant >>`

pimlie: I dont agree. Introducing a text variant only makes it possible to introduce other errors because then you have to make certain to pass a proper string to that parameter.

If you want to support legacy browsers we can add a boolean legacy as well. If you want to support version tagging we can add a version param but with the same risk that it doesnt exists.

The idea behind defining an executor is that we can define it exactly to our needs so we dont have to duplicate settings anywhere else with the risk of them being wrong. Using a boolean for browsers provides that abstraction. With the current config I can guarantee you that however you call the executor the docker image exists. If we are going to pass a string you cannot guarantee that.

clarkdo: I'm not quite understand your concern, anywhere which is using string parameters will have the risk you mentioned, so does it mean that the string parameters shouldn't exist at all ? As this config is for internal CI usage, I think it provide more flexibility and the risk you mentioned can be easily avoided/fixed by running a CI job.

I agree with your point about executor is for removing duplicate settings, but I think it does't mean the executor itself should be bigger and bigger because of more boolean params are added for ensuring stability, we still should care about the size and reusability of executor itself.

Maybe we can listen to the opinions of other members. @nuxt/core-team

pimlie:
@clarkdo Feel free to make changes to this pr yourself if you really believe this is a blocking issue

clarkdo:
I don’t think it blocks anything, it’s just in reviewing, we can hear feedback from others, if no concerns, it can be merged

clarkdo:
Cool

pimlie:
@clarkdo Thank you for merging :hugs: 

@pi0 Pinging you for visibility so you know circleci pipelines are now enabled

PR #6329, opened 2019-08-28
feat(builder): add builder:prepared hook

pimlie:
## Types of changes

## Description

This hook is called just after the `buildDir`'s have been created but before anything else has happened. We need this hook in eg `nuxt/press` so we can add build files before the webpack builders are started.

Currently we use the `build:compile` hook but that runs for each webpack build, so we have added a [check](https://github.com/nuxt/press/blob/develop/src/blueprint.js#L235) that the hook only runs on the first call. We should be able to replace that with the `build:ready` hook.

## Checklist:


TheAlexLichter (changes requested):
The name isn't intuitive IMO. "ready" would mean (to me) that the build is finished. Maybe something more explicit like `dirCreated` or similar (verbositiy level like `generate:distRemoved`)?

pimlie:
Finding good names is hard! This was the first thing that came to mind because of `nuxt:ready`. Eg the description of that hook says `Nuxt is ready to work`. Thats similar to here, `builder:ready: Builder is ready to work` (or get to work).

I'd love to change this hook to `build:before` and change `build:before` to `build:setup` or `build:init` or something. Think thats closer to the truth for the before hook actually but is a breaking change ofc.

`build:beforeGenerate` (1st thing logged after this hook is `Generating nuxt files`), but confusing due to the whole generator stuff.
`build:beforeGenerateRoutesAndFiles` -> too long
`build:afterBefore` -> strange
`build:beforeCompile` -> I'd expect this to be called just before `bundleBuilder.build()`
`build:beforeBuild` -> strange
`build:beforeApp` -> to generic
`build:dirCreated` -> what dir? not sure why but doesnt sound really nice
`build:beforeResolve` -> maybe the best I can think of for now (almost all methods in generateRoutesAndFiles are resolving stuff)

idk

manniL:
Totally agree, finding names *is* hard :D

`builder:ready` would be a way better fit than `build:ready` ("builder is ready to work" vs. "build is ready to be served"), but we don't have that naming scheme :see_no_evil: 	

Maybe

`build:directoriesCreated`? 

As https://github.com/nuxt/nuxt.js/blob/46d219f9c2ecf86c926a1c54d63f97d215e10efc/packages/builder/src/builder.js#L143-L152 will do exactly that, creating the build directories

`build:afterSetup`?

But then we have the question of what "setup" is again.

pimlie:
Maybe `build:ready` is indeed wrong because I keep typing `builder:ready` by mistake already ;)

What about combining yours into `build:dirsCreated`? Technically thats still not fully correct because eg the (optional) `layouts`, `views` dirs are not yet created at that point. 

Or `build:validated`? Thats what happens just before the dirs are created.

manniL:
`build:dirsCreated` is my favorite at the moment, though I'm not sure if pluralizing will cause problems later (when we pick up `dir` for a hook somewhere and have to be consistent :thinking:)

pimlie:
@Atinux @pi0 @clarkdo Does one of you want to make an executive decision about the name of this hook?

pi0:
Some notes regarding hook names:

- For legacy reasons (before mono-repo) hooks were not per package. Newer hooks are with `[package-name]:name` convention so `builder:*` should be a better idea and soon we deprecate and improve hook names to new convention.
- `ready` in nuxt core is often referred to a stage that a class constructor is ready so that would be little confusing using it here.

I would suggest using `builder:prepare` or `builder:initial` or `builder:init`

pimlie:
Would `builder:prepared` be ok as well?

Atinux:
I'm good with `builder:prepared` :)

manniL:
If we can use `builder` then I'd vote for the initial `builder:ready`

pi0:
@mannil see my previous comment regarding ready :)

pi0 (approved):

atinux (approved):

manniL:
We shouldn't forget docs ☺️

PR #6331, opened 2019-08-29
feat(builder, router): enforce uniform trailing slash handling

TheAlexLichter:
## Types of changes

## Description

Trailing slash handling is important when it comes to SEO. For every existing Nuxt application, the content provided by /route and /route/ should be the same (opposed to how it sometimes worked back in the days). But when it comes to how the search engine sees our page, we have to be consistent and choose one of the options, either with or without a trailing slash.

> Be consistent with the preferred version. Use it in your internal links. If you have a Sitemap, include the preferred version (and don’t include the duplicate URL).

With the current setup, it is not entirely possible to enforce trailing slashes (or the removal of them) without substantial effort.

As long as you have simple and fixed URLs, you can always append the slash (or not), but when it comes to resolved named routes (e.g. when using https://github.com/nuxt-community/nuxt-i18n/ ), things become difficult and hard to manage.

**Thus, this PR**

---

It introduces a new router option called `trailingSlashes` set to `undefined` by default. If it's set to `true`, trailing slashes will be appended on every route. If set to `false`, they'll be removed.

Therefore, we use the `pathToRegexpOptions` settings (per route) provided via vue-router (see https://github.com/vuejs/vue-router/issues/1753 and https://github.com/vuejs/vue-router/issues/1443)

---

This option should not be taken lightly and has to be tested thoroughly. When setting `router.trailingSlashes` to something else than undefined, one of the options (either with or w/o trailing slash) will stop working. 301 redirects should be in place before.

---

Related work:

https://github.com/nuxt/nuxt.js/pull/3448/
https://github.com/nuxt/nuxt.js/pull/5746

## TODO

- [ ] check edge cases
	- [ ] index route
    - [ ] nuxt child routes 
	- [ ] (???)
- [ ] docs

## Checklist:


galvez (dismissed):
Good idea.

pimlie:
Nice timing, have been struggling with trailing slashes in nuxt/press a bit the last weeks!

I'd recommend to include a util method so users can easliy normalize paths they use in their project as well, see eg [here](https://github.com/nuxt/press/blob/develop/src/utils/route.js) for the methods currently used in nuxt/press (but they could be improved on probably).

Atinux:
Good call!
Shouldn’t be easier to directly add a meta canonical directly so we fix the duplicate for SEO? (This is what we use for nuxtjs.org)

manniL:
>  Shouldn’t be easier to directly add a meta canonical directly so we fix the duplicate for SEO? (This is what we use for nuxtjs.org)

Well, canonical is one thing and easily realizable via layouts ([example](https://github.com/Developmint/developmint.de/blob/master/layouts/default.vue#L35-L49)). But the internal link structure should reflect trailing slashes (or none) as well

manniL:
> I'd recommend to include a util method so users can easliy normalize paths they use in their project as well

What would be the use case when all (named) routes or those resolved programmatically will go in line with the config and fixed routes would be easy to change by appending/removing the `/`?

pimlie:
@Atinux Isnt that technically more a workaround then a fix? Due to the nature of a SPA the path resolve behaviour is just different then serving static files with a webserver (ie apache/nginx). How close do we want to stay to that 'original' behaviour?

@manniL If your router config lists all routes with a trailing slash, then your nuxt/router-links should point to a link with a trailing slash as well right?

packages/utils/src/route.js
```diff
@@ -173,6 +174,18 @@ export const createRoutes = function createRoutes ({
         }
       }
     })
+    if (typeof trailingSlashes !== 'undefined') {
```
pi0: I think for consistency with other parts, this would be better to handled/normalized in config package. This way, during startup all modules access to correct resolved value.

packages/utils/src/route.js
```diff
@@ -173,6 +174,18 @@ export const createRoutes = function createRoutes ({
...
+      route.pathToRegexpOptions = { strict: true }
+    }
+
+    if (trailingSlashes) {
+      route.path = route.path.endsWith('/') ? route.path : `${route.path}/`
+    }
+
+    if (trailingSlashes === false) {
```
pi0: Maybe we can make at an `else if`? This would save branch check-in cases that `trailingSlashes` is true + makes it more readable.

packages/utils/src/route.js
```diff
@@ -136,7 +136,8 @@ export const createRoutes = function createRoutes ({
   srcDir,
   pagesDir = '',
   routeNameSplitter = '-',
-  supportedExtensions = ['vue', 'js']
+  supportedExtensions = ['vue', 'js'],
+  trailingSlashes = undefined
 }) {
   const routes = []
   files.forEach((file) => {
```
pi0: (Not related to this PR) what do you think of refactoring `forEach` to ES6 `for of`? OR using `files.map` to directly assign the result of `routes` instead of pushing to the end.

pi0 (commented):
Nice improvement 👍

packages/utils/src/route.js
manniL: Plus one for using `map` instead :relaxed:

packages/utils/src/route.js
manniL: I'm afraid it's not that easy because I need access to `route` (to set page options). There is no normalization that is global / doesn't need the route access.

pi0: Oops sorry, i wrongly read next line. That's the totally correct place

packages/utils/src/route.js
clarkdo: How about ?
```js
if(trailingSlashes !== undefined) {
  const trailing = trailingSlashes ? '/' : ''
  route.path = route.path.replace(/\/+$/, '') + trailing
}
```

packages/config/src/config/router.js
```diff
@@ -12,5 +12,6 @@ export default () => ({
   parseQuery: false,
   stringifyQuery: false,
   fallback: false,
-  prefetchLinks: true
+  prefetchLinks: true,
+  trailingSlashes: undefined
```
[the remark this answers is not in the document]
pi0: +1 for `trailingSlash`

manniL:
> If your router config lists all routes with a trailing slash, then your nuxt/router-links should point to a link with a trailing slash as well right?

Correct. If you use fixed URLs as in `to="/simple-string/"` you can do it on your own. If you used named routes `to="{ name: 'foobar' }"` it is handled by vue-router already and will include the trailing slash (or not)

packages/utils/src/route.js
clarkdo: Previous comment was swollen by outdated comment
```suggestion
if (trailingSlashes !== undefined) {
  route.pathToRegexpOptions = { ...route.pathToRegexpOptions, strict: true }
  route.path = route.path.replace(/\/+$/, '') + (trailingSlashes ? '/' : '')
}
```

pimlie:
> If you use fixed URLs as in `to="/simple-string/"` you can do it on your own

Thanks, this is why I recommend a normalization utility. Maybe even within nuxt-link itself, but should also be accessible outside for easy use in eg the sitemap-module.

If we support this feature switching between having a trailing slash or not should be as effortless as possible, preferable even with just toggling the setting if you use the normalization utility everywhere.

packages/utils/src/route.js
```diff
@@ -136,7 +136,8 @@ export const createRoutes = function createRoutes ({
   srcDir,
   pagesDir = '',
   routeNameSplitter = '-',
-  supportedExtensions = ['vue', 'js']
+  supportedExtensions = ['vue', 'js'],
+  trailingSlashes = undefined
```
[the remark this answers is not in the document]
manniL: No :relaxed:

packages/config/src/config/router.js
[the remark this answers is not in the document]
manniL: Changed it :relaxed:

manniL:
> Thanks, this is why I recommend a normalization utility. Maybe even within nuxt-link itself, but should also be accessible outside for easy use in eg the sitemap-module.

Now I got it, that'd be a nice DX improvement and also guarantee that all modules implement the same normalization.

A utility directly included in the `nuxt-link` component is also a nice idea, but could be more difficult than it seems on the first look (taking e.g. query params and programmatic usage into consideration)

clarkdo (approved):

packages/utils/src/route.js
pi0: Let's do this later and not blocking this PR :)

pi0 (approved):

pi0:
TODO for myself: add multi scope handling for release-notes generator!

PR #6342, opened 2019-09-02
feat(builder): support adding custom middlewares

pimlie:
## Types of changes

## Description

It seems that the middleware template in vue-app contains some business logic which should be moved to the builder class. This logic prevents me from adding a custom middleware (using the build:templates hook) in nuxt/press which has a webpack alias as path so it cant/shouldnt be resolved with `relativeToBuild`.

This pr moves that logic into the `resolveMiddleware` method of the builder class. Not sure if its needed, but to prevent a breaking change I have kept the resolve logic in the middleware template for now.

## Checklist:

### DIFF
--- a/packages/vue-app/template/middleware.js
+++ b/packages/vue-app/template/middleware.js
@@ -1,8 +1,10 @@
 const middleware = {}
-<% middleware.forEach(m => {
-   const name = m.src.replace(new RegExp(`\\.(${extensions})$`), '')
+<% for (const m of middleware) {
+  const name = m.name || m.src.replace(new RegExp(`\\.(${extensions})$`), '')
+  const dst = m.dst || relativeToBuild(srcDir, dir.middleware, m.src)
 %>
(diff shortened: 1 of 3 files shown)


galvez (dismissed):
+1

packages/vue-app/template/middleware.js
```diff
@@ -1,8 +1,10 @@
 const middleware = {}
-<% middleware.forEach(m => {
-   const name = m.src.replace(new RegExp(`\\.(${extensions})$`), '')
+<% for (const m of middleware) {
+  const name = m.name || m.src.replace(new RegExp(`\\.(${extensions})$`), '')
```
pi0: Maybe we can add a TODO comment to remove duplicate logic in next major version?

pimlie: done :)

pi0 (approved):

PR #6349, opened 2019-09-03
feat: add hook build:config

clarkdo:
## Types of changes

## Description

I didn't add any param for now since we already have `extend` for customising webpack config and this hook should avoid introducing more side effects.

## Checklist:

### DIFF
--- a/packages/webpack/src/builder.js
+++ b/packages/webpack/src/builder.js
@@ -57,6 +57,12 @@ export class WebpackBundler {
       webpackConfigs.push(this.getWebpackConfig('Server'))
     }
 
+    const additionalConfigs = await this.buildContext.nuxt.callHook('build:config')
+
+    if (additionalConfigs) {
+      webpackConfigs.push(...[].concat(additionalConfigs))
+    }
+
     // Check styleResource existence
     const { styleResources } = this.buildContext.options.build
     if (styleResources && Object.keys(styleResources).length) {


manniL:
I'm wondering whether we should be more explicit (e.g. `webpackConfig`) and use the new patter (`builder:`)?

Also, we need docs (as usual) explaining the difference between the different methods of adding stuff to the webpack config (e.g. via extends).

packages/webpack/src/builder.js
```diff
@@ -57,6 +57,12 @@ export class WebpackBundler {
       webpackConfigs.push(this.getWebpackConfig('Server'))
     }
 
+    const additionalConfigs = await this.buildContext.nuxt.callHook('build:config')
+
+    if (additionalConfigs) {
+      webpackConfigs.push(...[].concat(additionalConfigs))
```
pimlie: Why the `[].concat`? If the array contains objects then after the concat its still an array to the same references right?

clarkdo: The concat here is for supporting return value either array or object as single or multiple builds

clarkdo:
@manniL Sorry, I don’t know the new hook pattern decision, is the new pattern corresponding to package? So maybe use hook name as ‘webpack:config’ ?

packages/webpack/src/builder.js
pimlie: Thanks, actually I didnt know you could return values with [`callHook`](https://github.com/jsless/hable/blob/master/src/hable.js#L45). Maybe I am just overseeing something simple but there is no return statement or whatever, so how does it even return a value?

clarkdo:
Will open another pr

packages/webpack/src/builder.js
clarkdo: Yes, that's true, it has been changed for a while, I've changed in new pr

manniL:
> y, I don’t know the new hook pattern decision, is the new pattern corresponding to package? So maybe use hook name as ‘webpack:config’ ?

https://github.com/nuxt/nuxt.js/pull/6329#issuecomment-526927022

PR #6350, opened 2019-09-03
feat(webpack): add `webpack:config` hook

clarkdo:
## Types of changes

## Description

I didn't add any param for now since we already have extend for customising webpack config and this hook should avoid introducing more side effects.

## Checklist:

### DIFF
--- a/packages/webpack/src/builder.js
+++ b/packages/webpack/src/builder.js
@@ -57,11 +57,9 @@ export class WebpackBundler {
       webpackConfigs.push(this.getWebpackConfig('Server'))
     }
 
-    const additionalConfigs = await this.buildContext.nuxt.callHook('build:config')
-
-    if (additionalConfigs) {
-      webpackConfigs.push(...[].concat(additionalConfigs))
-    }
+    const additionalConfigs = []
+    await this.buildContext.nuxt.callHook('build:config', additionalConfigs)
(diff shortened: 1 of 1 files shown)


packages/webpack/src/builder.js
```diff
@@ -57,6 +57,10 @@ export class WebpackBundler {
       webpackConfigs.push(this.getWebpackConfig('Server'))
     }
 
+    const additionalConfigs = []
+    await this.buildContext.nuxt.callHook('build:config', additionalConfigs)
```
pimlie: Why not just pass webpackConfigs? And also a ref to the builder?

For hacking purposes it would be interesting to just have access to everything.

clarkdo:
@manniL I see, from @pi0 comment, the hook should be corresponding to package, so this should be webpack instead of builder ?

packages/webpack/src/builder.js
clarkdo: We've already provided `extend` for customising, I'm wondering it we should expose here again.

clarkdo: Probably we can use this hook to implement extend directly

clarkdo: I added webpackConfigs for now, let's see further comments

pi0 (approved):

PR #6353, opened 2019-09-03
refactor(cli): call setup hook in run command with more args

kevinmarrec:
## Types of changes

## Description
It's overall a fix (where the setup hook is called) that will make TypeScript Runtime shine by not relying anymore on `process.argv` to find the `rootDir` which was not optimal (glitchy and not secure regarding new potential CLI Options).

See : https://github.com/nuxt/typescript/pull/97

## Checklist:

### DIFF
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -1,4 +1,5 @@
 
+import path from 'path'
 import consola from 'consola'
 import minimist from 'minimist'
 import Hookable from 'hable'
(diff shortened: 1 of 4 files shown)


packages/cli/src/command.js
```diff
@@ -35,6 +36,12 @@ export default class NuxtCommand extends Hookable {
   }
 
   async run () {
+    await this.callHook('setup', {
+      argv: this._argv,
+      cmd: this.cmd,
+      rootDir: path.resolve(this.argv._[0] || '.')
```
pi0: That's a duplicated logic here but at least more consistent to be into the core :)

pi0 (approved):
T.N.X 👍

packages/cli/src/command.js
```diff
@@ -35,6 +36,12 @@ export default class NuxtCommand extends Hookable {
   }
 
   async run () {
+    await this.callHook('setup', {
```
pi0: I hope no one complains about hook name 😆 It is inside run so could be called like `run:before` but with logical perspective, it is really a setup step and internal hook so looks good and also not breaking users.

PR #6363, opened 2019-09-05
chore(cli): rename setup hook to run:before

kevinmarrec:
## Description
Rename `setup` CLI hook to `run:before` to better fit the context where it's called.

## Checklist:

### DIFF
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -36,7 +36,7 @@ export default class NuxtCommand extends Hookable {
   }
 
   async run () {
-    await this.callHook('setup', {
+    await this.callHook('run:before', {
       argv: this._argv,
       cmd: this.cmd,
       rootDir: path.resolve(this.argv._[0] || '.')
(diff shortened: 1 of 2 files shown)


packages/cli/src/command.js
```diff
@@ -36,7 +36,7 @@ export default class NuxtCommand extends Hookable {
   }
 
   async run () {
-    await this.callHook('setup', {
+    await this.callHook('run:before', {
```
pi0: This will be a **breaking** change for ts module users that only upgrade their nuxt. Is this OK? (I think for an internal hook like this setup name is just fine) 

so still merge this? :)

kevinmarrec: @pi0 Well current Nuxt release has `setup` hook not called at the really wanted place, and Nuxt TS with hooks is only draft PR.

We talked about **breaking** change indeed for people upgrading module but not Nuxt, but we said we want to force user to use last hooky Nuxt version , so all good no ? 😁

pi0: Yep! I thought TS module was released with setup hook. So seems good!

PR #6368, opened 2019-09-05
feat(builder): `followSymlinks` option to allow for symlinks

SnirShechter:
## Types of changes

## Description

This PR adds a build.followSymlinks boolean(default: false), and passes it to the builder's glob function inside resolveFiles().

This allows for following symlinks inside layouts/middleware/pages/store, thus including them in the app.
## Checklist:

- [X] I have added tests to cover my changes (if not applicable, please state why)


pimlie:
@SnirShechter Thanks for this pr. Would it maybe be possible to combine the tests with an existing fixture? Adding it to `basic` or `with-config` should be fine.
The reason for this is that every new fixture adds about ~10seconds to the test time. So we try to only add new fixtures for incompatible configs.

Also note that it seems you are using a different emailaddress in your local git config then for github. If you want you can either add that other emailaddress to your github account (Settings -> Emails) or correct the emailaddress & do a force-push (old commits will still be accessible though).

SnirShechter:
> @SnirShechter Thanks for this pr. Would it maybe be possible to combine the tests with an existing fixture? Adding it to `basic` or `with-config` should be fine.
> The reason for this is that every new fixture adds about ~10seconds to the test time. So we try to only add new fixtures for incompatible configs.
> 
> Also note that it seems you are using a different emailaddress in your local git config then for github. If you want you can either add that other emailaddress to your github account (Settings -> Emails) or correct the emailaddress & do a force-push (old commits will still be accessible though).

Hey, sorry for the really late reply. I'm on it.

pimlie (approved):
LGTM. The only thing I was thinking about is to maybe make it possible for a user to set any [glob](https://github.com/isaacs/node-glob#options) option, but I think we probably shouldnt do that as changing a setting like `absolute` or `realpath` could result in unexpected behaviour in nuxt.

SnirShechter:
It is something worth thinking of, but basically followSymlinks is (through my point of view) much more than a simple glob setting, since it allows separation of pages in different folders. For me and my company, this means that we can integrate one project into another, when built right.

Thanks for the approval 👍

pi0 (approved):

PR #6373, opened 2019-09-06
feat: expose monorepo scripts/helpers

renoirb:
(Related to #6380)

As a monorepo maintainer, I want to normalize/synchronize fields in my packages’ *package.json*. **AND** Nuxt.js’ does it already. But it's hidden, and inaccessible for other monorepo maintainers. 

* This started [based on a twitter thread](https://twitter.com/shellscape/status/1169303684976697347)
* And also on [RushJS.io's](https://rushjs.io/) (a monorepo manager, like Lerna. But leveraging PNPM)  [Gitter (link to channel)](https://gitter.im/web-build-tools/web-build-tools?at=5d71af113b1e5e5df15d47bc)

Factor out [scripts/package.js](https://github.com/nuxt/nuxt.js/blob/dev/scripts/package.js) so that Nuxt.js, and other monorepo projects maintainer, can use in their own monorepos themselves. Exporting parts of `scripts/package.js` into a new package `normalize-package-json`.

### Found elsewhere

* **RushJS** does it in their `@microsoft/node-core-library` ([GitHub](https://github.com/microsoft/web-build-tools/tree/master/libraries/node-core-library), [npmjs package](https://www.npmjs.com/package/@microsoft/node-core-library)) in [src/PackageJsonLookup.ts](https://github.com/microsoft/web-build-tools/blob/master/libraries/node-core-library/src/PackageJsonLookup.ts). But doesn’t have file/field syncing logic. In RushJS land, when using PNPM, we could do [same behavior as what `scripts/package` does (link in Gitter)](https://gitter.im/web-build-tools/web-build-tools?at=5d759e41a08e2b4bd2b0af37)

## Types of changes

## Description

- Create new package `normalize-package-json` (suggested name), it could be in a separate project/git repo, and not namespaced (e.g. `@nuxt(|js|-community)/...`) so that other projects that aren’t using Nuxt.js wouldn’t be confusing it as a co-dependent of Nuxt.js
- Support reading `package.json` from a directory, run `sort-package-json` using `Package#autoFix()` method
- Support using `Package#load('relative/package-name')` method to create another reference and then be able to copy fields/files between packages

### Questions

- Refactor hooks outside `Package#build()` method, or make it abstract
- Need help testing `Package#copyFilesFrom()`
- Need help referencing appropriately between new `package/normalize-package-json` and host monorepo `scripts/package.js`.
 

## Checklist:


renoirb:
Ping @pi0, just to tell you I've started here. Gotta write tests and factor things around. Also, I’ve forgot to get through CMTY.

packages/normalize-package-json/src/package.js
```diff
@@ -0,0 +1,157 @@
...
+          rootDir: this.resolvePath(relativePath)
+        },
+        opts
+      )
+    )
+  }
+
+  async writePackage () {
+    if (this.options.sortDependencies) {
+      this.sortDependencies()
```
shellscape: I'd bet that this isn't needed, given that most (all?) package managers will do this for you on install.

renoirb: That’s likely something they do in their own monorepo management.

packages/normalize-package-json/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "license": "MIT",
+  "files": [
+    "dist"
+  ],
+  "main": "dist/index.js",
+  "dependencies": {
+    "consola": "^2.10.1",
+    "fs-extra": "^8.1.0",
+    "lodash": "^4.17.15",
```
renoirb: Maybe we should refactor `sortObjectKeys` to NOT make this package depend on lodash. Maybe inline dependencies?

packages/normalize-package-json/test/package.test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(bar.pkg).toHaveProperty('author')
+
+    // Snapshot, just for visualization
+    foo.autoFix()
+    expect(foo.toString()).toMatchSnapshot()
+    bar.autoFix()
+    expect(bar.toString()).toMatchSnapshot()
+  })
+
+  // @TODO: Test copyFilesFrom. But we’d need to mock filsystem writes.
```
renoirb: Did not yet write test for writing to filesystem. This should be mocked, there must be an utility, or a convention to do such mocking/testing. Somewhere. I didn’t find yet.

scripts/package.js
```diff
@@ -1,69 +1,27 @@
...
+import { existsSync, remove } from 'fs-extra'
 import { rollup, watch } from 'rollup'
 import { glob as _glob } from 'glob'
 import pify from 'pify'
-import sortPackageJson from 'sort-package-json'
+
+// FIXME
+// @renoirb: I’m unsure how we should refer from here.
+//           (It's the kind of thing that made me look elsewhere than lerna)
+import { Package as BasePackage } from '../packages/normalize-package-json'
```
renoirb: @pi0 please help here :)

packages/normalize-package-json/src/package.js
pi0: @shellscape Yep. But this is for linting our code to make package.json files consistent. And certainly can be optional to disable :)

packages/normalize-package-json/package.json
pi0: Rollup looks like a good idea!

scripts/package.js
pi0: Maybe if we have a fancy name like Kitty can call it KittyPackage :)

pi0 (commented):
Thank you @renoirb for initial PR and @shellscape for review.  I think it refactoring this package out of nuxt monorepo makes maintenance easier, less dependency on nuxt requirements and more visible to the world! I'll create it based on your works and notify :)

PS: SORRY for the late review. I had a bad cold this weekend.

packages/normalize-package-json/src/package.js
renoirb: sort-package-json already sort all dependencies.  Would sort-package-json run *before* syncing version between monorepo packages?

galvez:
@pi0 Should we close this PR then?

pi0:
@renoirb @galvez Sorry I was too busy to continue working on this :( Let's keep it open as a TODO until we create a new repo.

renoirb:
It's OK for the delay.
I have something started somewhere ([here][experiment-normalize-package-json]) but stalled about it too.

[experiment-normalize-package-json]: https://github.com/renoirb/experiments-201908-rush-typescript-just-bili-monorepo/tree/master/tools/normalize-package-json

clarkdo (commented):
@renoirb @pi0 I think we can move all the logic for `package.js` to this new module and make the module be `@nuxt/package` or `@nuxt/packager`

renoirb:
Ohai, @clarkdo! (It's been a while!)

If @pi0 is OK, or has some tribal-knowledge and requirements I could do the work.

I guess requirements would look something like:
- in a specific location
- using future friendly/or in transition monorepo utilities 
- add/change feature such and such

I did navigate Nuxt code base for a while now, I wouldn't need much hand holding

Keep me posted

clarkdo:
@renoirb 😸 
I think your current work is a very good start, maybe just need a more generic package name, then @pi0 and myself can work on left migration works on it in another thread.

pi0:
Hi! We are working with @danielroe on the siroc project (private) to move forward this refactor. @renoirb I have sent an invitation for early test and having your opinions :)

renoirb:
Thanks @pi0 will definitely follow up!

PR #6374, opened 2019-09-06
feat: find app.html in root dir as writed in docs

777genius:
## Types of changes

## Description

[Docs](https://nuxtjs.org/guide/views/#app-template):

> To change the template, create an app.html file in the root folder of your project.

But actually the file was searched in the src folder. This makes it work as indicated in the documentation.


manniL:
Often, `rootDir` is equal to the `srcDir`. I'm wondering whether it's a doc issue and `srcDir` is meant or both should be possible.

cc @pi0

pi0:
Hi. Thanks @iliyaZelenko for this PR and attention ❤️ 

The `app.html` file should be in `srcDir`. Being in `rootDir` can make conflicts when more than one project exists. And yeah @manniL is right. by default, these directories are the same.

We should update docs instead.

PR #6399, opened 2019-09-11
feat(cli, config): `cli.bannerColor` option

### LINKED ISSUE #6397: Change banner color

Perhaps this is the strangest feature request in this repository...<br/>
I think we should update the banner color of `nuxt-ts` to match the color scheme :D

### Nuxt TypeScript

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9760">#c9760</a>)</em></sub></div>

danielroe:
## Types of changes

## Description
This exposes a `build.bannerColor` option to change the color of the Nuxt title in the CLI banner.

Resolves: #6397

## Checklist:
   *I think that might be overkill for a simple cosmetic change*


pi0 (changes requested):
We don't include typescript logic here. It may be better an option to customize color.

danielroe:
> We don't include typescript logic here. It may be better an option to customize color.

I've exposed a `build.ui.titleColor` option, with the idea that more ui options could be exposed in future. Is this the right location for this option? If so, I'll prepare a PR updating `@nuxt/typescript` and `@nuxt/types`.

galvez (changes requested):
I'd call it `bannerColor` instead of `ui.titleColor`, since we don't immediately have any other options to place under that key.

danielroe:
> I'd call it `bannerColor` instead of `ui.titleColor`, since we don't immediately have any other options to place under that key.

Done!

packages/cli/src/utils/banner.js
```diff
@@ -20,7 +20,11 @@ export function showBanner (nuxt, showMemoryUsage = true) {
   const messageLines = []
 
   // Name and version
-  titleLines.push(`${chalk.green.bold('Nuxt.js')} ${nuxt.constructor.version}`)
+  let bannerColor = 'green'
```
pi0: Please keep defaults in config package (test snapshots require update)

pi0 (changes requested):

pi0 (approved):
LGTM! Thanks ❤️

kevinmarrec:
@danielroe @pi0 Is there any reason the banner border color hasn't be set to the new option ? :P

kevinmarrec:
image

Exeteres:
I have the same question :flushed:

danielroe:
@kevinmarrec @Exeteres I like the idea! (Mind you, in production mode there will be green in the picture too.)

PR #6414, opened 2019-09-14
feat(config): `serverMiddleware` as a simple key/value object

galvez:
If using only named serverMiddleware routes, this PR makes it possible to do:

```
serverMiddleware: {
   '/route' (req, res, next) {
   }
}
```


packages/config/src/options.js
```diff
@@ -395,5 +395,10 @@ export function getNuxtConfig (_options) {
     options.server.timing = { total: true, ...timing }
   }
 
+  if (!Array.isArray(options.serverMiddleware)) {
```
pimlie: Maybe `isPureObject` from utils would be better suited? It only doesnt check for _null_, if you check for that as well then Object.entries will be guaranteed to only be called on a proper object.

packages/config/test/options.test.js
```diff
@@ -256,6 +256,18 @@ describe('config: options', () => {
   })
 })
 
+describe('config: serverMiddleware', () => {
+  test('should transform serverMiddleware hash', () => {
+    const serverMiddleware = {
+      '/resource' (req, res, next) {
```
pimlie: Although this is fine, I'd prefer to just use a normal key/value syntax with a closure. Putting single quotes around the function name is just weird

pimlie (commented):
Nice feature, certainly makes for a cleaner/simpler to understand config. It would be nice to also support this in `addServerMiddleware`, you'd just have to make sure it doesnt collide with the `path/handler` object.

packages/config/test/options.test.js
galvez: @pimlie heh, noted -- mostly personal preference, either way works, but I'll update it for you.

packages/config/src/options.js
galvez: Nice, forgot about `isPureObject()`, thanks.

packages/config/test/options.test.js
pimlie: You could also just always prepend the first slash, so users dont have to bother with that. Then you could do the above by just using `resource`.

Just have to change the [replace regex here](https://github.com/nuxt/nuxt.js/blob/1b837a1ab41342016800e18c7ac7c05ad11afb9d/packages/server/src/server.js#L201-L204) to read `replace(/\/{2,}/g, '/')`

galvez: I'm opting to keep it literal unless anyone feels strongly otherwise -- I think the syntax can grow on you if you give it time, and it saves us from the confusion.

manniL:
Could you elaborate what you mean with "as hash" exactly?

galvez:
@manniL updated PR title.

pi0 (approved):

PR #6438, opened 2019-09-19
feat(server): add config option to define etag hash function

pimlie:
## Types of changes

## Description

While looking into some benchmarks, @aldarund found out that generating the etag could snoop up a lot of processing time (especially with large html, ie `extractCss: false`). Further investigation found that the etag module uses `SHA1` for generating a hash of the full html body.

This pr introduces a new config option `render.etag.hash` which users can use to implement their own etag hash function. Ie for etags it should often be fine to use a non-cryptographic hash function like murmurhash, which is multiple times faster with larger html body sizes.

Note: the `render.etag.weak` option is ignored when using the hook, user is responsible for implementing that.

Example  for using [murmurhash-native](https://github.com/royaltm/node-murmurhash-native) instead:
```js
// nuxt.config.js
import { murmurHash128 } from 'murmurhash-native'

export default {
  render: {
    etag: {
      hash: html => murmurHash128(html)
    }
  }
}
```

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -38,10 +38,8 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
     // Add ETag header
     if (!error && options.render.etag) {
-      const etagArg = { value: undefined }
-      await nuxt.callHook('render:generateETag', html, etagArg)
-
-      const etag = etagArg.value || generateETag(html, options.render.etag)
+      const { hash } = options.render.etag
+      const etag = (hash && hash(html)) || generateETag(html, options.render.etag)
       if (fresh(req.headers, { etag })) {
         res.statusCode = 304
         res.end()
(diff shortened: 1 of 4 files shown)


packages/server/src/middleware/nuxt.js
```diff
@@ -38,7 +38,10 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
     // Add ETag header
     if (!error && options.render.etag) {
-      const etag = generateETag(html, options.render.etag)
+      const etagArg = { value: undefined }
+      await nuxt.callHook('render:generateETag', html, etagArg)
```
pimlie: Should we pass the context as well here?

I didnt add it as users either already define the hook in nuxt.config and module imlementation can already access nuxt options from the module container

packages/server/src/middleware/nuxt.js
```diff
@@ -38,7 +38,8 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
     // Add ETag header
     if (!error && options.render.etag) {
-      const etag = generateETag(html, options.render.etag)
+      const { hash } = options.render.etag
+      const etag = (hash && hash(html)) || generateETag(html, options.render.etag)
```
pi0:
```suggestion
      const etag = hash ? hash(html, options.render.etag) : generateETag(html, options.render.etag)
```

pi0 (approved):

PR #6445, opened 2019-09-20
chore: add script to generate md for wiki package pages

### LINKED ISSUE #5596: Document in Readme or elsewhere what version of Vue Nuxt is using

Thanks for reviewing this feature request! As Vue is getting ready for 3.0 release, it's going to be more and more common that people really need to know the which version of Vue they're using. This came up for me recently because I was unsure of which slot syntax I was able to use (no big deal to test), but good to be more explicit for userland. Thanks!

Either document which version of Vue is being used in the Readme, or as some field generated by package.json. Those are the two places I would check first. I can find it in yarn.lock but I have to walk through 102 instances of "vue" in the file to find it.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9100">#c9100</a>)</em></sub></div>

pimlie:
Resolves #5596 

## Types of changes

## Description

I used this script to generate the markdown for the following wiki pages:
https://github.com/nuxt/nuxt.js/wiki/Packages and
https://github.com/nuxt/nuxt.js/wiki/Dependencies

Adding it to the repo because the wiki pages should be updated once a new dependency or package is added. See [here](https://github.com/Decathlon/wiki-page-creator-action/blob/master/entrypoint.sh) for the implementation of a github action which could be used for that (but probably easier to implement it just as a release step).

## Checklist:

### DIFF
--- a/scripts/generate-wiki-markdown.js
+++ b/scripts/generate-wiki-markdown.js
@@ -1,0 +1,248 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+const https = require('https')
+
+const repoPath = 'nuxt/nuxt.js'
+const pkgNamespace = '@nuxt'
+const branches = [
+  {
+    name: '2.x',
+    title: 'Stable'
+  },
+  {
+    name: '',
+    title: 'Edge',
+    pkgSuffix: '-edge'
+  }
+]
+
+const template = ({ tables, vars, extra }) => `
+<!-- /* this page is automatically generate */ -->
+
+## Distributions
+${tables.dist}
+
+## Core packages
+${tables.pkgs}
+
+${extra}
+${vars}
+`
+
+const options = {
+  '-d': 'dependencies',
+  '-p': 'packageSize'
+}
+
+const optionKey = process.argv.find(arg => !!options[arg])
+
+if (!optionKey) {
+  // eslint-disable-next-line no-console
+  console.log(`Usage: ${path.basename(__filename)} -d | -p
+  -d    Generate markdown for dependencies page
+  -p    Generate markdown for package sizes page
+`)
+  process.exit(1)
+}
+
+const generateOption = options[optionKey]
+
+function arrayToTable (tableArray) {
+  let table = ''
+  tableArray.forEach((row, index) => {
+    table += row.join('|')
+    table += '\n'
+
+    if (!index) {
+      // header border
+      table += row.map(c => '-').join('|')
+      table += '\n'
+    }
+  })
+
+  return table
+}
+
+function request (url) {
+  return new Promise((resolve, reject) => {
+    https.get(url, (resp) => {
+      let data = ''
+
+      // A chunk of data has been recieved.
+      resp.on('data', (chunk) => {
+        data += chunk
+      })
+
+      // The whole response has been received. Print out the result.
+      resp.on('end', () => resolve(data))
+    }).on('error', (err) => {
+      reject(err)
+    })
+  })
+}
+
+function createPackageSizeBadge ({ branch: { name: branch, pkgSuffix }, name, fullName, type, vars }) {
+  fullName = `${fullName}${pkgSuffix || ''}`
+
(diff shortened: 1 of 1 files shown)


scripts/generate-wiki-markdown.js
```diff
@@ -0,0 +1,248 @@
...
+      // header border
+      table += row.map(c => '-').join('|')
+      table += '\n'
+    }
+  })
+
+  return table
+}
+
+function request (url) {
```
pi0: Why not using axios or node-fetch?

scripts/generate-wiki-markdown.js
```diff
@@ -0,0 +1,248 @@
...
+
+  return table
+}
+
+async function main () {
+  const distPath = path.join(__dirname, '..', 'distributions')
+  const pkgsPath = path.join(__dirname, '..', 'packages')
+
+  const distributions = fs.readdirSync(distPath)
+  const packages = fs.readdirSync(pkgsPath)
```
pi0: Maybe we can use [getWorkspacePackages](https://github.com/nuxt/nuxt.js/blob/dev/scripts/package.js#L169) helper?

scripts/generate-wiki-markdown.js
```diff
@@ -0,0 +1,248 @@
...
+`
+  }
+
+  const md = template({
+    tables,
+    extra,
+    vars: Object.entries(vars).reduce((acc, [key, value]) => `${acc}[${key}]: ${value}\n`, '')
+  })
+
+  // eslint-disable-next-line no-console
```
pi0: Consola can remove need of this :P

scripts/generate-wiki-markdown.js
```diff
@@ -0,0 +1,248 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+const https = require('https')
+
+const repoPath = 'nuxt/nuxt.js'
+const pkgNamespace = '@nuxt'
+const branches = [
```
pi0: Maybe we can move this to the top level `package.json` and using Package class to read?

scripts/generate-wiki-markdown.js
pimlie: Cause I prefered 0 dependencies so this script can easily be used for eg nuxt/press or non-nuxt monorepo's as well

scripts/generate-wiki-markdown.js
pimlie: Same as above :)

scripts/generate-wiki-markdown.js
pimlie: Same as above :)

scripts/generate-wiki-markdown.js
pimlie: Same as above 🙃

scripts/generate-wiki-markdown.js
pimlie: We could integrate this script maybe in the jsless package.js package if thats being extracted, then it would be ok probably

pi0: Yep! I really want to externalize our maintenance at a point and your wiki generator is AWESOME for that!

pi0:
@pimlie How do you think about more integration with other scripts and extenalize all of our tools at a point?

pimlie:
> How do you think about more integration with other scripts and extenalize all of our tools at a point?

You mean like #6373? I have no objections to that but not sure if there is really a big enough demand for that to spent a lot of time on.

pi0:
Yeah but zero dep is quite strange for like making HTTP requests. When we don't even externalize script and already have dep in dev dependencies. How is the plan to use script for other repos like nuxrpress? copy paste entire script?

pimlie:
I made this PR mostly just as a courtesy, dont think its worth your or my time discussing this in so much detail so here it is as a gist: https://gist.github.com/pimlie/7f3b1ebdedcff80080942a20dccc05ca

PR #6460, opened 2019-09-25
fix(webpack): correct sass options for `sass-loader@8`

TheAlexLichter:
When using sass-loader@^8.0.0, the Nuxt defaults will break it's usage due to a validation error.

```
ValidationError: Invalid options object. Sass Loader has been initialised using an
options object that does not match the API schema.
 - options has an unknown property 'indentedSyntax'. These properties are valid:
   object { implementation?, sassOptions?, prependData?, sourceMap?, webpackImporter? }
```

That's because the API has changed and distinct sass options are no top-level keys anymore but belong into the `sassOptions` object.

**Problem**: This will likely break sass-loader<=7.x.y

What should we do regarding this issue?

* Add a note that we implicitly bump sass-loader to ^8.0.0? (We can't control that)?
* Grab the version of the installed sass-loader and set the defaults based on it?
* Add a note that ppl using sass-loader@^8.0.0 have to update the stuff on their own?

## Types of changes

## Checklist:

## Related
https://github.com/vuejs/vue-cli/issues/4513


pi0:
> Grab the version of the installed sass-loader and set the defaults based on it?

That could be NICE!

PR #6475, opened 2019-09-28
feat(cli): catch build errors with `cli:buildError` hook

pimlie:
refactor: errors during initiating/listening nuxt are non-recoverable thus should be fatal. Also some errors during the build are non-recoverable but not sure if thats _some_ or _all_, so dont make those fatal for now

## Types of changes

When you were developing your app and eg added a plugin but made a typo, then the re-build of Nuxt would fail but the loading screen would not reflect that. The build error was only visible on the console, which can be confusing if you are not used to looking at the console for errors.

This pr changes that (together with https://github.com/nuxt/loading-screen/pull/45) to capture build errors in dev mode and print a stracktrace of them on the loading screen.

image

## Checklist:

### DIFF
--- a/packages/cli/src/commands/dev.js
+++ b/packages/cli/src/commands/dev.js
@@ -27,16 +27,25 @@ export default {
   },
 
   async startDev (cmd, argv) {
+    let nuxt
     try {
-      const nuxt = await this._startDev(cmd, argv)
-
-      return nuxt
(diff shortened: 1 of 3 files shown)


pi0 (approved):

Atinux:
YES, thank you @pimlie <3

PR #6477, opened 2019-09-28
feat(webpack): use `vue-meta` browser build for client bundle

pimlie:
## Types of changes

## Description

Similar to consola, explicitly using the vue-meta browser dist for client builds saves ~1.5KB in bundle size

## Checklist:

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -224,7 +224,8 @@ export default class WebpackBaseConfig {
   alias () {
     return {
       ...this.buildContext.options.alias,
-      consola: require.resolve(`consola/dist/consola${this.isServer ? '' : '.browser'}.js`)
+      consola: require.resolve(`consola/dist/consola${this.isServer ? '' : '.browser'}.js`),
+      'vue-meta': require.resolve(`vue-meta${this.isServer ? '' : '/dist/vue-meta.esm.browser.js'}`)
     }
   }
 


clarkdo (approved):

husayt:
@pi0, After this change I am getting error:

> This vue app/component has no vue-meta configuration

pimlie:
@husayt Thanks for the report, could you please share your config / reproduction?

That message should only be printed when initialization of vue-meta failed but a refresh of the meta-info is still called: https://github.com/nuxt/vue-meta/blob/master/src/client/refresh.js#L24

husayt:
I have rolled back on that project to before the change. Will try to reproduce on a fresh project and share. General config wise it's a generate only  SPA with modern mode set  to client.

kevinmarrec:
@pimlie Reproduction :
```
git clone https://github.com/nuxt-community/vuetify-module
yarn
yarn dev
```

frck006:
Since Nuxt 2.10.0, and @nuxtjs/vuetify@1.9.0, Vuetify doesn't work.

In Vuetify, file:/services/theme/index.ts, the method "initVueMeta()" uses "root.$meta()" which returns object with "options" to "undefined"
After, meta.getOptions().keyName throw an exception.

pimlie:
@frck006 thanks but see the vuetify issue that was linked 2 hrs ago

frck006:
Yes I see: https://github.com/vuetifyjs/vuetify/issues/9181
Thanks

Issue #6232, opened 2019-08-14
`config.watch` failed at second restart if use wildcard


TrungRueta:
### Version

[v2.8.1](https://github.com/nuxt.js/releases/tag/v2.8.1)

### Reproduction link

[http://empy](http://empy)

### Steps to reproduce

1. create new project with minimum setup (no add any modules / packages) via `create-nuxt-app`.
2. create new folder at root project named `serverMiddleware`.
3. create new file inside `serverMiddleware` folder, named `test.middleware.js`. add content (see bottom):
4. open `nuxt.config.js`. add into `watch` property : `watch: ['serverMiddleware/test.middleware.js']`. Note: `config.watch` not `config.build.watch`
5. run `yarn dev` to host dev mode, open site in browser and f5 several times. We will see in terminal the console `~this xxx` message.
6. open `test.middleware.js` and edit string from `~ this` to `~ this edited`. save file. We will see in terminal the server auto restarted.
7. After restart done, refresh brower and see message changed. Repeat step 6, 7 to make sure server restart no issue.
8. turn off process (ctrl + c). then open `nuxt.config.js`, edit watch file string from `serverMiddleware/test.middleware.js` to `serverMiddleware/*.js`.
9. start server by `yarn dev`. refresh browser to see message show on.
10. repeat step 6, 7. The watcher will failed with trap 6 nodejs report at second restart.

### test.middleware.js
```js
export default {
  handler: function(req, res, next) {
    // req is the Node.js http request object
    console.log('~ this ', req.path)

    // res is the Node.js http response object

    // next is a function to call to invoke the next middleware
    // Don't forget to call next at the end if your middleware is not an endpoint!
    next()
  }
}
```

### What is expected ?

config.watch with wildcard should work no issue.

### What is actually happening?

config.watch failed when restart second times if watch string contains wildcard syntaxs

### Additional comments?

Ps: also i found that serverMiddleware will cache somewhere outside nuxt processing if in `config.serverMiddlware` property we use string file path.
If we require middleware file use `require()` inside `nuxt.config.js` then import direct to `config.serverMiddleware` and add file path to `config.watch` then project will listen and restart correctly.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9642">#c9642</a>)</em></sub></div>

pimlie:
[...]

PR #6490, opened 2019-10-01
fix(core): clear resolver cache for project files dev mode

pimlie:
## Types of changes

## Description

Partial fix for #6232 (that if serverMiddleware is provided as string then on dev server restart the changes in the middleware are not picked up).

I think this also fixes the issue I have during module development which caused me to having to restart the dev server all the time because changed werent picked up.

Not sure how to add a test for this.

## Checklist:

### DIFF
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -3,7 +3,12 @@ import fs from 'fs-extra'
 import consola from 'consola'
 import esm from 'esm'
 
-import { startsWithRootAlias, startsWithSrcAlias } from '@nuxt/utils'
+import {
+  startsWithRootAlias,
+  startsWithSrcAlias,
+  isExternalDependency,
+  clearRequireCache
+} from '@nuxt/utils'
 
 export default class Resolver {
   constructor (nuxt) {
@@ -138,12 +143,12 @@ export default class Resolver {
       lastError = e
     }
 
-    const isExternal = resolvedPath.includes('/node_modules/')
+    const isExternal = isExternalDependency(resolvedPath)
 
     // in dev mode make sure to clear the require cache so after
     // a dev server restart any changed file is reloaded
-    if (this.options.dev && !isExternal && require.cache[resolvedPath]) {
-      delete require.cache[resolvedPath]
+    if (this.options.dev && !isExternal) {
+      clearRequireCache(resolvedPath)
     }
 
     // By default use esm only for js,mjs files outside of node_modules
--- a/packages/core/src/resolver.js
+++ b/packages/core/src/resolver.js
@@ -147,8 +147,8 @@ export default class Resolver {
     }
 
     // By default use esm only for js,mjs files outside of node_modules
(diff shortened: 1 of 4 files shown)


packages/core/src/resolver.js
```diff
@@ -138,9 +138,17 @@ export default class Resolver {
       lastError = e
     }
 
+    const isExternal = resolvedPath.includes('/node_modules/')
+
+    // in dev mode make sure to clear the require cache so after
+    // a dev server restart any changed file is reloaded
+    if (this.options.dev && !isExternal && require.cache[resolvedPath]) {
+      delete require.cache[resolvedPath]
```
pi0: We have a util (in cjs.js) to recursively cleanup required cache

packages/core/src/resolver.js
```diff
@@ -138,9 +138,17 @@ export default class Resolver {
       lastError = e
     }
 
+    const isExternal = resolvedPath.includes('/node_modules/')
```
pi0: Maybe also nice to export as a utility function

packages/core/src/resolver.js
pimlie: Ah yes, missed that. Using that now

packages/utils/src/cjs.js
```diff
@@ -1,6 +1,10 @@
+export function isExternalDependency (id) {
+  return id.includes('/node_modules/')
+}
+
 export function clearRequireCache (id) {
   const entry = require.cache[id]
-  if (!entry || id.includes('node_modules')) {
+  if (!entry || isExternalDependency(id)) {
```
pimlie: not sure if it was deliberate to use `node_modules` without separators here?

kevinmarrec:
@pi0 Any idea is it wouldn't cause issue to TypeScript runtime ?

packages/utils/src/cjs.js
pi0: Indeed we can use path sep (don't forgot to include both \\ and / in regex ;))

pi0:
@kevinmarrec Good point. We can try to ensure. I remember there was something wrong. But can try :)

kevinmarrec:
Yeah I don't remember well what was going on around `ts-node` require cache.. It may be not an issue anymore 🤷‍♂️

packages/utils/src/cjs.js
```diff
@@ -1,6 +1,12 @@
+import { sep } from 'path'
+
+export function isExternalDependency (id) {
+  return id.includes(`${sep}node_modules${sep}`)
```
pi0: I think it should cover **both** separators. A windows user can also call require with slash :)

pi0 (commented):
-

pi0 (approved):
LGTM! It may (rare) be possible we introduce some perf penalties as we clear cache (so more CPU and Disk read) but from DX perspective seems a good idea.

aldarund (approved):

atinux (approved):

PR #6495, opened 2019-10-01
fix(webpack): enable devtool for dev mode

clarkdo:
## Description

TODO:
- [ ] Test dev mode memory usage
- [ ] Update doc for default value


pimlie:
What exactly is the default value currently? Would it be an idea to just set it to `cheap-module-eval-source-map` by default as for the server build?

I tried this locally but think the errors are still not as helpful as they can be

clarkdo:
I think the default value of webpack may have more consideration and try to keep best practice in both readability and performance. And this config can be easily set via nuxt.config as well.

clarkdo:
The default value is eval in dev mode which includes almost everything in source code

aldarund (dismissed):

pimlie:
Looking at https://webpack.js.org/configuration/devtool/, wouldnt it be best for DX to have at least the `original source` quality by default? `eval` has the quality `generated code` which is still not really helpful right?

As you said, if performance is an issue users can always set it back to `eval` themselves

pimlie (approved):
LGTM, thanks!

TheAlexLichter (approved):

aldarund (approved):

PR #6525, opened 2019-10-07
fix(cli): initialize server for build when there pages should be gene…

aldarund:
…rated

fixes #6519

## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
In spa mode we need to generate page, so hence need server to be initialized

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x ] All new and existing tests are passing.

PR #6515, opened 2019-10-04
fix(utils): error when router.trailingSlash is false

y-temp4:
## Types of changes
- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Fix https://github.com/nuxt/nuxt.js/issues/6501

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #6511, opened 2019-10-03
chore(readme): remove deprecated starter template

dennishzg:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

I removed the deprecated starter template `starter` ([nuxt-community/starter-template](https://github.com/nuxt-community/starter-template)) from the README.md.

### Why?

- The template is deprecated, the repository is archived and the text in the repository also says:

> use create-nuxt-app instead

>  Please use create-nuxt-app to scaffold your Nuxt applications instead!

- > **manniL said:** We have to deprecate the template in favor of create-nuxt-app. Otherwise, **problems will arise when people try to use it**. - https://github.com/nuxt-community/starter-template/issues/80#issue-364357998

- This change also prevents issues like this one: #5010

- The starter templates are not in the official Nuxt.js documentation! This is probably a bit confusing for beginners.

- We have `create-nuxt-app` 😄 ❤️ 🎉 

### 🤔 Question:

Should i remove [nuxt-community/adonuxt-template](https://github.com/nuxt-community/adonuxt-template) too? This template is also marked as deprecated.

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

PR #6504, opened 2019-10-02
fix(vue-app, webpack): remove consola from client

clarkdo:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Fix https://github.com/nuxt/nuxt.js/pull/6298#issuecomment-537563409

~~- Use `consola BrowserReporter` directly which have smaller size and better compatibility~~
~~- Remove unnecessary consoal webpack `alias` config~~
- Remove consola from client side

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #6529, opened 2019-10-08
v2.10.1

pi0:
### 🐞 Fixes

- `vue-app, webpack`
  - #6504 Remove consola from client
- `utils`
  - #6515 Fix error when `router.trailingSlash` is `false`
- `generator`
  - #6362 Remove trailing slash when no subFolders are used
- `webpack`
  - #6520 `chunkhash` and `contenthash` not work with `extractCSS`
- `cli`
  - #6525 Enable server for `nuxt build` when pages should be generated

### 🏡 Chore

- `readme`
  - #6511 Remove deprecated starter template

### ⚓ Dependency Upgrades

- `vue-meta`: Upgraded to [`v2.3`](https://github.com/nuxt/vue-meta/releases) which fixes duplicated tags in SPA mode

### 💖 Thanks to

- Dennis Herzberg (@dennis47528)
- Yuki Terashima (@y-temp4)


pimlie (commented):
#6525 showed we have a serious gap in our tests. The danger of merging the fix without any tests to fill that gap is that adding those tests will be postponed indefinitely

Can we somehow add those tests (somewhat quickly) so we are releasing a fix which we can guarantee shouldnt happen again?

aldarund (approved):

galvez (approved):

TheAlexLichter (approved):

vinayakkulkarni:
Lowkey hoping this gets released sooner than later.

Got few of my projects hanging on v2.9.x

pimlie (approved):
#6537 adds new tests which should mostly close the gap mentioned above

clarkdo:
@pi0 The error comes from `memory-fs@0.5.0`. We're using `^0.4.2` now, so it should be safe to release.

PR #6566, opened 2019-10-14
fix(webpack): normalize devtool for terser compatibility

clarkdo:
## Types of changes

## Description

Some modules are using `#source-map` as devtool, this pr is for supporting those modules usage.

## Checklist:


packages/webpack/src/config/base.js
```diff
@@ -207,6 +207,7 @@ export default class WebpackBaseConfig {
             condition: 'some',
             filename: 'LICENSES'
           },
+          sourceMap: this.devtool && /((inline|hidden|nosources)-)?source-?map/.test(this.devtool),
```
pi0: Can't we normalize and remove # instead?

clarkdo: There should also have other variants, hard to tell if we can cover all

pi0: I'm not sure if we have other variants. But currently, we silently disable sourceMap which is not ideal imo. Maybe we can make a getter that tries to normalize and remove hash and in case it was still invalid, throw a console warn?

rchl: I also think it would be better to just remove `#` as otherwise this code potentially creates extra maintenance (validating supported values that can change at some point)

pi0 (approved):

PR #6586, opened 2019-10-17
examples(with-sockets): fix cannot read property 'app' of undefined on nuxt build

reskume:
The provided example works flawlessly in dev mode but when "nuxt build" is run, the build will fail with an error "Cannot read property 'app' of undefined" since the renderer is not available yet. By using the 'render:before' hook to assure that the renderer is available, it works as expected. This may not be the best solution but worked for me.

## Types of changes

## Checklist:


pi0:
Thanks!

PR #6594, opened 2019-10-20
fix(router): show default child if trailingSlash is false

### LINKED ISSUE #6593: Removing trailingSlash with nested route cause some problem
### Version

[v2.10.1](https://github.com/nuxt.js/releases/tag/v2.10.1)

### Reproduction link

[https://codesandbox.io/s/codesandbox-nuxt-9e9n1?fontsize=14](https://codesandbox.io/s/codesandbox-nuxt-9e9n1?fontsize=14)

### Steps to reproduce

1. Setup a nested route in `pages` directory
2. Add `router: { trailingSlash: false} ` to `nuxt.config.js`

In the codesandbox, I created the following files
```
pages/
  ...
  - users.vue
  - users/
    - index.vue
```

### What is expected ?

In `https://9e9n1.sse.codesandbox.io/users`, the contents of `pages/users.vue` & `pages/users/index.vue` are shown.

In `https://9e9n1.sse.codesandbox.io/users/` (with trailingSlash), `This page could not be found` is shown

In `https://9e9n1.sse.codesandbox.io/users`, the contents of `pages/users/index.vue` is **not** shown. Only `pages/users.vue` is shown.

But in `https://9e9n1.sse.codesandbox.io/users/` (with trailingSlash), the contents of both are shown.

Also, the following error appeared in console.
```
[vue-router] Named Route 'users' has a default child route. When navigating to this named route (:to="{name: 'users'"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead.
[vue-router] Duplicate named routes definition: { name: "users", path: "/users" }
```

### Additional comments?

The following is the summary of generated `.nuxt/router.js` 

```
...
  mode: 'history',
  base: decodeURI('/'),
  linkActiveClass: 'nuxt-link-active',
  linkExactActiveClass: 'nuxt-link-exact-active',
  scrollBehavior,

  routes: [{
    path: "/users",
    component: ...,
    name: "users",
    children: [{
      path: "",
      component: ...,
      name: "users"
    }]
  },
  ...
  ],
  fallback: false
}
...
```

In my opinion, to remove trailing slash, 

1. The path of children which is default root of nested route, should be same as parent's root.
2. We should not set `name` to parent of nested route. It's duplicated in children

So expected router.js may be

```
...
  mode: 'history',
  base: decodeURI('/'),
  linkActiveClass: 'nuxt-link-active',
  linkExactActiveClass: 'nuxt-link-exact-active',
  scrollBehavior,

  routes: [{
    path: "/users",
    component: ...,
    children: [{
      path: "/users",
      component: ...,
      name: "users"
    }]
  },
  ...
  ],
  fallback: false
}
...
```

Currently, my workaround is configuring `nuxt.config.js` as following

```
...
router: {
    trailingSlash: false,

    extendRoutes (routes) {
      routes.forEach(r => {
        if (!r.children) return

        delete r.name

        r.children.forEach(c => {
          if (c.path !== '') return

          c.path = r.path
        })
      })
    }
}
...
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9930">#c9930</a>)</em></sub></div>

TheAlexLichter:
Resolves https://github.com/nuxt/nuxt.js/issues/6593

The bug was introduced by setting `'/'` instead of `''` for child routes when `router.trailingSlash` is `false`:
 https://github.com/nuxt/nuxt.js/blob/e39f54b44e203f6c47100f8fd779f34345bac772/packages/utils/src/route.js#L184.

## Types of changes

## Checklist:


manniL:
I kept the failing tests because I'm not sure about the impact of the introduced change and love to get feedback on the PR.

packages/utils/src/route.js
```diff
@@ -47,7 +47,7 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-')
         paths.shift()
       } // clean first / for parents
       routesIndex.forEach((r) => {
-        const i = r.indexOf('index') - start //  children names
+        const i = r.indexOf('inxdex') - start //  children names
```
pi0: typo?

manniL: oh lol, yeah :see_no_evil: 
```suggestion
        const i = r.indexOf('index') - start //  children names
```

pimlie:
The failing test seems to be an issue, adding the `?` to the route param will now also match the route when no param is given. In the test that might not be an issue because `projects` is already listed before, but if a user doesnt want to match `/projects` at all then this is a breaking change I guess?

manniL:
My check for `trailingSlash` was wrong, thus the weird result. Should be fine with the latest commit.

manniL:
Ready for review :relaxed:

pimlie:
Sorry for asking but why is deleting the name of the route the solution? It isnt something thats immediately obvious to me.

manniL:
@pimlie Valid question! As reported in https://github.com/nuxt/nuxt.js/issues/6593,

pimlie:
Thanks. Isnt deleting the route name then strictly speaking resolving a consequence of an issue? Ie why are the route names duplicate in the first place?

If it isnt solvable by fixing the duplicate route names, could we maybe add a comment explaining why we choose to fix it this way?

manniL:
@pimlie Not sure why there were duped route names initially. Maybe @clarkdo could provide more info 🤔

pimlie:
I think its because of this replace: https://github.com/nuxt/nuxt.js/blob/dev/packages/utils/src/route.js#L64

Because of that `my-route` and `my-route/index` will both be named `my-route`.

RettuSh (approved):

RettuSh:
Hi! Is there any change to fix this bug soon? Could I help you somehow?

ems1985:
Hi, any updates on this?

Just FYI the issue also presents itself when ```router.trailingSlash``` is set to ```true```

This is my folder structure:
image

Without the trailingSlash option my router.js routes looks like this:
```
routes: [{
    path: "/something",
    component: _2eb15426,
    children: [{
      path: "",
      component: _0e04f6b0,
      name: "something"
    }, {
      path: "childcomponent",
      component: _4484dbf3,
      name: "something-childcomponent"
    }]
  }, {
    path: "/",
    component: _d1bba076,
    name: "index"
  }],
```

With ```router.trailingSlash``` set to ```true``` 
```
routes: [{
    path: "/something/",
    component: _2eb15426,
    name: "something",
    children: [{
      path: "childcomponent/",
      component: _4484dbf3,
      name: "something-childcomponent"
    }, {
      path: "",
      component: _0e04f6b0,
      name: "something"
    }]
  }, {
    path: "/",
    component: _d1bba076,
    name: "index"
  }]
```

Of course results in the following warnings:
```
 WARN  [vue-router] Named Route 'something' has a default child route. When navigating to this named route (:to="{name: 'something'"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead.


 WARN  [vue-router] Duplicate named routes definition: { name: "something", path: "/something/" }  
```

atinux (approved):

Issue #6588, opened 2019-10-18
Support case-sensitive patterns in .nuxtignore


chriscalo:
### What problem does this feature solve?

In the `pages` directory, I would like to intermingle `.vue` files that are pages and `.vue` files that are just components.

(I'm aware of the convention to put components in the `components` directory, I just don't agree with it in all situations. There are plenty of situations when a page component becomes quite complex and it's useful to break parts of it into separate components, but those components will never be reused. In those situations, it's best for others trying to understand the codebase to have the component colocated right next to the page that uses it.)

I tried renaming all `.vue` files that are pages to `name.page.vue`, but that makes the URL to access them `name.page` when I want it to be just `name`.

Then I renamed all supporting components using the convention `name.component.vue`, but that resulted in quite verbose file names. Alternatively, the abbreviation `name.cmpt.vue` felt non-obvious, so I didn't love that.

The convention I liked most was to name all page components in lowercase (`name.vue`) to match the all-lowercase convention in URLs and capitalize all supporting components (`LoginForm.vue`) as a nod to the convention of capitalizing classes (another construct that produces instances from a kind of template or blueprint).

So I tried the following:

``` gitignore
pages/**/[A-Z]*.vue
```

Unfortunately, that pattern seemed to match all `.vue` files that started with a letter, so it seems that `.nuxtignore` files are case-insensitive.

### What does the proposed changes look like?

I would love to use the following case-sensitive pattern in `.nuxtignore`:

``` gitignore
# ignore all capitalized .vue files in the pages directory
pages/**/[A-Z]*.vue
```

However, because the `[A-Z]` pattern is case-insensitive, this essentially means any `.vue` file that starts with a letter, which doesn't work as expected.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9927">#c9927</a>)</em></sub></div>

clarkdo:
https://github.com/nuxt/nuxt.js/pull/6597 has been merged and will be published in  next release

PR #6597, opened 2019-10-20
feat(config): add `ignoreOptions` for `node-ignore`

clarkdo:
## Types of changes

## Description

Since node-ignore 4.x, it has support `options` as parameter of constructor, so this pr is adding the config from nuxt.config.js.

With this change, #6588 can be solved by: 
```js
// nuxt.config.js

export default {
  ignoreOptions: {
    ignorecase: false
  }
}
``` 

TODO:

- [ ] Test
- [ ] Doc


galvez (approved):

pi0 (approved):

chriscalo:
Just tried this in the latest release 2.11.0 and it doesn't seem to be working.

`nuxt.config.js`:

``` js
export default {
  // .nuxtignore
  ignoreOptions: {
    ignorecase: false,
  },
};
```

`.nuxtignore`:

``` text
pages/**/[A-Z]*.vue
```

All `pages/**/*.vue` in my project seem to be ignored regardless of how they are named.

clarkdo:
@chriscalo What is the version of ignore in you project? The options is supported from v4

clarkdo:
And also can you try by exact path like pattern abc.vue doesn’t ignore ABC.vue.

chriscalo:
@clarkdo, thanks for the reply.

I'll check on the version when I get back to a computer, I haven't installed ignore myself. It's a dependency of Nuxt, so shouldn't Nuxt control the version rather than me?

> And also can you try by exact path like pattern abc.vue doesn’t ignore ABC.vue.

Not sure what you mean by this. Can you say more? Where should I try exact paths?

clarkdo:
Yes, it’s a dep from nuxt, but it may have conflict or hoisting if your other dep also has ignore package as a dep.
I mean use exact pattern in your ignore file like pages/abc.vue and verify it doesn’t ignore pages/ABC.vue in your repo.

clarkdo:
I’m sorry that I’m in holiday. I’ll check this as soon as I’m back.

chriscalo:
Here's what I found in my `yarn.lock` file:

``` text
ignore@^5.1.1, ignore@^5.1.4:
```

chriscalo:
> I mean use exact pattern in your ignore file like pages/abc.vue and verify it doesn’t ignore pages/ABC.vue in your repo.

Thanks, will try that.

> I’m sorry that I’m in holiday. I’ll check this as soon as I’m back.

Enjoy! Please don't rush on my account. This refactoring can wait. Just wanted to point out that it didn't seem to be working. Happy to provide more debug info if that's useful.

chriscalo:
I think I may have figured out the problem: my macOS HD is case insensitive, which I'm guessing means the `ignoreCase` won't work as expected. I may have to find some other way of solving my problem. Apologies for the runaround on this.

PR #6615, opened 2019-10-25
perf: use flatstr for SSR

galvez:
Seems we can take advantage of [flatstr](https://github.com/davidmarkclements/flatstr) for SSR?

### DIFF
--- a/packages/vue-renderer/package.json
+++ b/packages/vue-renderer/package.json
@@ -11,6 +11,7 @@
     "@nuxt/devalue": "^1.2.4",
     "@nuxt/utils": "2.10.2",
     "consola": "^2.10.1",
+    "flatstr": "^1.0.12",
     "fs-extra": "^8.1.0",
     "lru-cache": "^5.1.1",
     "vue": "^2.6.10",
--- a/packages/vue-renderer/src/renderers/base.js
+++ b/packages/vue-renderer/src/renderers/base.js
@@ -1,3 +1,5 @@
+import flatstr from 'flatstr'
+
 export default class BaseRenderer {
   constructor (serverContext) {
     this.serverContext = serverContext
@@ -16,7 +18,7 @@ export default class BaseRenderer {
     opts.head_attrs = opts.HEAD_ATTRS
     opts.body_attrs = opts.BODY_ATTRS
 
-    return templateFn(opts)
+    return flatstr(templateFn(opts))
   }
 
   render () {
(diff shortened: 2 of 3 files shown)


pi0:
I think before that we need benchmarks with different Node.js versions and platforms :)

```
// You may be tempted to copy and paste this, 
// but take a look at the commit history first,
// this is a moving target so relying on the module
// is the best way to make sure the optimization
// method is kept up to date and compatible with
// every Node version.
```

pi0:
Besides the risks, in their docs it is mentioned: 

> The best place to use flatstr is just prior to passing it to an API that eventually runs non-v8 code (such as fs.WriteStream, or perhaps xhr or DOM apis in the browser).

We don't! We concatenate `HEAD` and `APP` before sending response :)

Also:

> calling flatstr too much can in fact negatively effect performance

Simply with more concurrency, we are calling it too much

galvez:
> We don't!

We don't what? I think HTTP write streams are affected by this if fastify is to be believed :)

> We concatenate HEAD and APP before sending response :)

I think you're supposed to run it on a concatenated string.

Also, re: compatibility, it seems to work for our supported Node versions :)

galvez:
> Simply with more concurrency, we are calling it too much

I'm interested in a benchmark, I'll try this out soon. Let's keep it in standby for now :)

galvez:
@pi0 FWIW it seems `fastify` does use `flatstr` for concurrent requests :)

pi0:
> We don't what? I think HTTP write streams are affected by this if fastify is to be believed :)

We don't directly send these strings to stream. We do a concat before that! This optimization can be done on the final string i think. 

> Also, re: compatibility, it seems to work for our supported Node versions.

Yes and No. Perf improvement is not guaranteed and also about next V8 versions which can be even backported.

> I'm interested in a benchmark, I'll try this out soon. Let's keep it in standby for now :)

Sure :) (however could be a simple issue/RFC or draft PR before asking review on something which is not tested)

> FWIW it seems fastify does use flatstr for concurrent requests.

Fastify is designed/optimized for API. Nuxt is working with SSR rendered content. Of course, both use Node.js.

galvez:
> We don't directly send these strings to stream. We do a concat before that!

Gotcha, I moved it to `renderTemplate()` now and will benchmark :)

galvez:
@pi0 My benchmark didn't show any significant different. Tried in multiple settings and two computers. Withdrawing this PR :)

PR #6623, opened 2019-10-28
fix(vue-app): add `prefetched` class to `<nuxt-link>` after chunk loaded

atinux:
## Types of changes

## Description

Add prefetched class only when the chunk is loaded.

## Checklist:

### DIFF
--- a/packages/vue-app/template/components/nuxt-link.client.js
+++ b/packages/vue-app/template/components/nuxt-link.client.js
@@ -102,7 +102,7 @@ export default {
         }
         Component.__prefetched = true
       }<% if (router.linkPrefetchedClass) { %>
-        Promise.all(promises).then(() => this.addPrefetchedClass())
+        return Promise.all(promises).then(() => this.addPrefetchedClass())
       <% } %>
     }<% if (router.linkPrefetchedClass) { %>,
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -92,15 +92,18 @@ export default {
...
       for (const Component of Components) {
         const componentOrPromise = Component()
         if (componentOrPromise instanceof Promise) {
           componentOrPromise.catch(() => {})
+          <% if (router.linkPrefetchedClass) { %>promises.push(componentOrPromise)<% } %>
         }
         Component.__prefetched = true
       }<% if (router.linkPrefetchedClass) { %>
-      this.addPrefetchedClass()<% } %>
+        Promise.all(promises).then(() => this.addPrefetchedClass())
```
pi0: We can add a return statement here to avoid unhandled promises:
```suggestion
        return Promise.all(promises).then(() => this.addPrefetchedClass())
```

Atinux: Thanks @pi0

pi0 (approved):

PR #6625, opened 2019-10-28
fix(builder): call extendPlugins after normalizing them

atinux:
## Types of changes

## Description

Extending plugins should happen after normalising them.

## Checklist:

- Pending doc PR: https://github.com/nuxt/docs/pull/1698

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -226,7 +226,7 @@ export default class Builder {
   async generateRoutesAndFiles () {
     consola.debug('Generating nuxt files')
 
-    this.plugins = Array.from(await this.normalizePlugins())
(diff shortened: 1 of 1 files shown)


galvez (approved):
Good catch, thanks!

pimlie:
Hmm, but now plugins added in extendPlugins are not normalized anymore? Can we really trust the user to do everything as they supposed to?

Atinux:
Indeed this could be an issue as well @pimlie

Atinux:
Closing since it will make the `extendPlugins` harder to use for end users.

manniL:
We could consider structural changes for the next major instead? :relaxed:

PR #6657, opened 2019-11-05
chore(github): add Chinese translation for issue template

clarkdo:
Add Chinese translation for issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,14 +1,22 @@
 blank_issues_enabled: false
 contact_links:
-  - name: 🚨 Bug report
+  - name: 🚨 Bug report | Bug 提交
     url: https://bug.nuxtjs.org/
-    about: Please report bugs here.
-  - name: 🙋 Feature request
+    about: |
+      Please report bugs here.
+      请在此提交 Bug。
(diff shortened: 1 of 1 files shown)


pimlie (commented):
Why is Chinese so special that it needs to be added here? Should we also add Spanish, Russian, French or other languages with large Nuxt communities?

manniL:
@pimlie 

AFAIK: Vue & Nuxt are quite popular in Asian countries. Learning English isn't as "established" there as it is in e.g. European Countries.

pimlie:
True, but:
- Asia is bigger then just China
- We shouldn't favor a single language like this. Imo we cant reasonably defend adding Chinese but not eg French (3 core-team members), Russian (@aldarund) or Portuguese (@galvez) which are all in the top 10 of most spoken languages according to [wikipedia](https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers)
- Except for @clarkdo nobody can support issues written in Chinese anyway (same would apply ofc to adding Russian or Portugese)

pi0 (approved):
I think there are valid reasons for add Chinese:

- Nuxt.js is based on Vue.js which has the biggest market in china
- In the past, there were no issues other than English but some Chinese ones. So that makes sense to guide Chinese developers in the right direction

Suggestion for later improvements: Maybe we can also request them to please write in English in CMTY

PR #6658, opened 2019-11-06
fix(vue-app): pass router mode to `getLocation`

### LINKED ISSUE #523: Fails to redirect to locale on initial navigation with router in hash mode
### Version

[v6.3.0](https://github.com/nuxt-i18n/releases/tag/v6.3.0)

### Reproduction link

[https://codesandbox.io/s/nuxt-i18n-router-in-hash-mode-tl28z](https://codesandbox.io/s/nuxt-i18n-router-in-hash-mode-tl28z)

### Steps to reproduce

1. In nuxt.config.js add: `router: { mode: 'hash' }`
2. Configure `nuxt-i18n` to redirect to locale:
```
      "nuxt-i18n",
      {
        locales: ["en"],
        detectBrowserLanguage: {
          alwaysRedirect: true,
          fallbackLocale: "en"
        },
        vueI18n: {
          fallbackLocale: 'en',
          messages: {
            en: {
              welcome: "Welcome"
            }
          }
        }
      }
```
3. Go to `http://localhost:3000`

### What is expected ?

Index route is rendered

Blank page is shown

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt-i18n/issues/c341">#c341</a>)</em></sub></div>

rchl:
## Types of changes

## Description
On the client, Nuxt creates "synthetic" route initially, before VueRouter
has initialized so that plugins and middlewares have something to work
with at that early stage already. The problem was that initial route
was created without taking router's `mode` into consideration so it
created route with normal path rather than hash-based path. That confused
nuxt-i18n's logic that tries to redirect user to correct locale on
initial navigation.

As for tests:
 - created new `spa-hash` fixture config so that we can run tests in hash mode
 - new fixture is using files from `spa` fixture to avoid duplication
 - change `spa.test.js` testsuite to run same tests on both fixtures
 - disable `/тест雨 (test non ascii route)` test in hash mode as it is failing
   for unrelated reason (I haven't investigated deeper).
 - add `Initial route has correct fullPath` test to test the bug fixed here

Resolves https://github.com/nuxt-community/nuxt-i18n/issues/523

## Checklist:


rchl:
Much more pleasant to review after enabling `Hide whitespace changes` option in github's diff view.

clarkdo (approved):

test/unit/spa.test.js
```diff
@@ -1,161 +1,174 @@
...
-const url = route => 'http://localhost:' + port + route
-
-const renderRoute = async (_url) => {
-  const window = await nuxt.server.renderAndGetWindow(url(_url))
-  const head = window.document.head.innerHTML
-  const html = window.document.body.innerHTML
-  return { window, head, html }
-}
+// Runs tests in specified router mode (either "hash" or "history").
+function spaTests ({ isHashMode }) {
```
clarkdo: Can we extract the hash mode specific test cases into a separted file like `spa.hash.test.js` for more readable and easier to maintain in future ?

rchl: There are no "hash specific" testcases per se. All are valid testcases for both modes (apart from one test which fails in hash mode for some unknown reason). So it wouldn't make sense to create two identical sets of tests and two identical sets of files in a new fixture.

pi0 (approved):

PR #6683, opened 2019-11-11
fix(vue-app): apply path-to-regexp options to tokensToFunction regexp

pimlie:
## Types of changes

## Description

Route params should be matched case insensitive, this wasnt the case as the path regexp sensitive option. This was originally fixed in https://github.com/pillarjs/path-to-regexp/pull/191 but vue-router is still on path-to-regexp v1, so I've backported that fix which has now been released in https://github.com/pillarjs/path-to-regexp/releases/tag/v1.8.0

Related vue-router issue: https://github.com/vuejs/vue-router/issues/2925

Note: actually the options arent used at all as we never pass options to compile/tokensToFunction. This actually means the default behaviour changes due to this pr from being case sensitive to case insensitive

Note 2: for a full fix its probably required to wait until a vue-router release has been released that uses path-to-regexp v1.8 or v3. Although in my tests that wasnt actually necessary, not sure exactly if/when vue-router will call _tokensToFunction_ in a Nuxt.js app

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -280,7 +280,7 @@ export function getLocation (base, mode) {
  * @return {!function(Object=, Object=)}
  */
 export function compile (str, options) {
-  return tokensToFunction(parse(str, options))
+  return tokensToFunction(parse(str, options), options)
 }
 
 export function getQueryDiff (toQuery, fromQuery) {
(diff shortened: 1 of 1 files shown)


clarkdo (approved):

clarkdo:
@pimlie Could you double check if it works as expected when enabling `sensitive` in IE ? I think IE doesn't support flags for RegExp.

pimlie:
Tested with IE11:
image

PR #6685, opened 2019-11-11
fix: context.route issue in plugins

mshahbazi:
`context.route` always points to the `/`, when `splitChunks.pages == false` in `nuxt.config.js`
it seems sort of an edge case. BTW, the issue is resolved with this change


clarkdo (commented):
Remove function ``() => ${route._name}.default || ${route._name}`` will break route components lazy loading, can you describe more details about your issue ?

mshahbazi:
@clarkdo, I have a plugin in which I need to get the current `context.route.path`. The plugin works perfectly with a  fresh `nuxt.config.js`. 
The problem raises when I set the `splitChunks.pages` to `false` in the `nuxt.config.js`. In this case, the `context` does not provide the right route. It actually returns `/`, no matter what the current page is. Digging into the code led me to the `vue-app` template file. Removed the line, resolved the issue.

clarkdo:
Cool, thanks for the explanation, so your issue is related to `splitChunks.pages: false`, then we should keep left part of conditional operator and add `route._name` as value of `splitChunks.pages: false` condition.

BTW, could you please add proper tests for this case ?

mshahbazi:
@clarkdo, you are right.
Let me add some thoughts:
The left part is already `route._name` and the right part causes the wrong `context.route` issue. My best guess was: at the time of loading the plugin, the async route component has not been loaded yet. To deal with this, I could think of two options:

1. Use an `await` mechanism to make sure that the component is loaded before the plugin. There are negative points in this: (1) requires deep code review; (2) I doubt if there is any point in async loading while all the route components are supposed to be loaded before the plugin.
2. Omit the condition. This disables async component loading.

To save time, I went for the second option and used chunk grouping to make sure that the components are optimally loaded. This resolved my issue, though the first option seems more sustainable.

clarkdo:
Thanks.

I didn't notice that `interopDefault` has been wrapped with the component in `import()`.

clarkdo (approved):

mshahbazi:
@clarkdo thank you so much

PR #6712, opened 2019-11-20
[wip] feat: add speed profiling to client build

eljefedelrodeodeljefe:
In my quest for speeding up webpack builds, I find it useful to add a plugin that can trace it. Due to the nature of said plugin, one cannot easily attach it as a user but would need to wrap the generated config. This PR does exactly that as opt-in. 

This is work in progress (docs and options missing), but I would appreciate it to be considered.

## Types of changes

## Description

Add plugin to client config.

## Checklist:

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -36,6 +36,7 @@
     "postcss-preset-env": "^6.7.0",
     "postcss-url": "^8.0.0",
     "semver": "^6.3.0",
+    "speed-measure-webpack-plugin": "^1.3.1",
     "std-env": "^2.2.1",
     "style-resources-loader": "^1.3.2",
     "terser-webpack-plugin": "^2.2.1",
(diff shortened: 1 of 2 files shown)


Kronhyx (approved):

clarkdo (commented):
Thank you for this great idea!
I think this feature would be better to be implemented as a nuxt module by leveraging hook webpack:config, if it proves to be a common used feature, we can bring it in core then.

pi0:
Hi, thanks for this request. Indeed it can be supported simply with a module. Also, we already have advanced profiling support from [webpackbar](https://github.com/nuxt/webpackbar). Simply use `--profile` or `build.profile: true`.

Docs: https://nuxtjs.org/api/configuration-build/#profile

Results:

image

image

eljefedelrodeodeljefe:
@pi0 thanks. Missed that one

PR #6717, opened 2019-11-22
feat: plugin hooks and parallel loading

pimlie:
## Types of changes

## Description

This PR adds support for:
- parallel loading of plugins (`features.parallelPlugins`, default _false_)
- plugin hooks  (`features.hookablePlugins`, default _false_)

Parallel loading could be a breaking change so its hidden behind a feature flag. Hookable plugins are also hidden but we might want to set that to _true_ by default

Plugin authors can adapt their plugins by using the new/third argument to their exported function. This argument is an instance of jsless/hable. Default hooks that are called are:
- `loaded`, called when all plugins are loaded
- `done`, called immediately after `loaded`. This is split up to support separate `setup` and `build` actions within plugins. Not sure if this is really useful though
- `injected:<key>`, called for any property that gets inserted in the context. It receives the injected value as arg

Because I think this PR will need some discussion I havent added tests yet.

## Checklist:


packages/vue-app/template/plugins.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  <% if (features.parallelPlugins) { %>
+  async function loadPlugin (plugin) {
+    if (!plugin || typeof plugin !== 'function') {
+      return
+    }
+
+    let injectPromises = []
+    const asyncInject = (key, value) => {
```
pimlie: This is not really a nice solution and actually the `injected:<key>` hook doesnt work yet for `parallelPlugins`. It would be much easier if we could change the `inject` function itself to be async, but not sure if thats a BC or not

packages/vue-app/template/plugins.js
```diff
@@ -0,0 +1,124 @@
+import Vue from 'vue'
+<% if (features.hookablePlugins) {
+%>import Hookable from 'hable'
```
pimlie: Including hable adds more functionality then we use/need at the moment. Not sure if its worth it to use the full hable? We could either use our own copy or maybe hable could export a slim version?

packages/vue-app/template/plugins.js
```diff
@@ -0,0 +1,124 @@
...
+
+    key = `$${key}`
+    // Add into app
+    app[key] = value
+    <% if (store) { %>
+    // Add into store
+    store[key] = app[key]
+    <% } %>
+
+    if (Vue.prototype.hasOwnProperty(key)) {
```
pimlie: Could we use `in` here?
```suggestion
    if (key in Vue.prototype) {
```

packages/vue-app/template/plugins.js
pimlie: Or maybe we should support synchronous hooks calling

packages/vue-app/template/plugins.js
galvez: I think keeping hable is ok. 

Not sure it's worth maintaining a separate version just for saving a couple hundred bytes.

packages/config/src/config/_app.js
```diff
@@ -70,6 +70,8 @@ export default () => ({
     validate: true,
     asyncData: true,
     fetch: true,
+    hookablePlugins: false,
+    parallelPlugins: false,
```
pi0: I'm **strongly** against introducing these two new configs as global configs. It fragments entire ecosystem! 

Instead, we should add support per/plugin for hook and parallel. A nuxt hook/module can enable it for all but it is then the risk of __consumers__. Not all plugins/modules are designed to work like this and depending on hooks. -- it is a per-plugin feature :)

examples/plugins-hookable/README.md
```diff
@@ -0,0 +1,3 @@
+# Using parallel and hookable plugins with Nuxt.js
```
pi0: Can we simplify new concept to just hooks? (instead of hookable)

packages/vue-app/template/plugins.js
pi0: We can do it easily if needed to introduce slim browser version. But what exactly you think is extra that worth trim for slim build?

examples/plugins-hookable/plugins/first.js
```diff
@@ -0,0 +1,12 @@
+
+export default function ({ app: { $loadOrder, $hookOrder } }, inject, hooks) {
```
pi0: Maybe we can add `inject` and `hooks` properties to the `inject` function object?

1. Keeping backward support
2. We can use the new syntax of `function(ctx, h)`: `function(..., { inject, hooks })`

pi0 (changes requested):

packages/config/src/config/_app.js
pimlie: ### `parallelPlugins`

- Parallel loading could be a breaking change. Module/plugin authors could make mistakes, this feature could also be used to forcibly run every plugin synchronous in case that happens 
- Consumers are not always in control of plugins, eg when they are added by 3rd party stuff
- Prosumers should be able to override any default behaviour of less-well-maintained 3rd party stuff if they know it should be save
  - This is probably also half a reason for the per-plugin argument, but we have no mechanisms atm in place for prosumers to override module plugin behaviour :)

### `hookablePlugins`
- Thats actually why I mentioned it should probably by default be `true`, we could then also do an extra pass on the used plugins to check if `arguments.length === 3` to enable it on-demand (but that doesnt prevent Hookable from being included in the bundle)

### Both
- You should also see these features as options for prosumers to fully optimize their Nuxt projects like the idea is of the other features. The plugin stuff is also used on the client so we should provide optimization options that just skips all the management logic that is needed for it

examples/plugins-hookable/README.md
pimlie: We could simplify hable to also export individual functions, then have the Hookable class use those functions too? Then we could do here:
```
import { hook, callHook } as Hookable from 'hable'
```
and the unused functionality of hable would be tree-shaked

examples/plugins-hookable/plugins/first.js
pimlie: Isnt using a new syntax a breaking change?

packages/config/src/config/_app.js
pi0: - In the case of "A superpower user that optimizes its project with full responsibility of enabling flag" she/he can also add a hook to enable this flag(s) for all plugins. including 3rd party ones. At least with a chance to partially do it.

- I think hookable plugins are parallel plugins, however, they are toward a single goal, are completely separate addons to nuxt. discussing in a single PR makes it just confusing.

examples/plugins-hookable/README.md
pi0: indeed. functions are always more js-native than classes IMO :D Anyway this comment was just regarding naming...

examples/plugins-hookable/plugins/first.js
pi0: If we attach it to inject, no:

```js
// backward compatible object
const h = inject
h.inject = inject
h.hook = hook
// pass h as second arg
```

packages/config/src/config/_app.js
pi0: Regarding `arguments.length` check, we can instead optionally export an explicit meta-object from plugins :)

pimlie: > instead optionally export an explicit meta-object from plugins

Hmm, I'm not in favor of having to export meta stuff because _describing_ how something works isnt really _kiss_ and always introduces the chance of the description being wrong. It'd be much better if we could deduct this automatically from syntax

examples/plugins-hookable/plugins/first.js
pimlie: Ghehe, I think @Atinux dislikes settings props on functions like this. He shot down a pr of mine because I did this (pr was about hooks too, having the possibility to run a hook only once and remove it afterwards)

packages/config/src/config/_app.js
pi0: Detecting from syntax can be tricky and unstable. Adding separate args for each plugin helper makes it not scalable (like adding more features for plugin export)

examples/plugins-hookable/plugins/first.js
pi0: Yes. Probably because you also skipped entire comments in ideas channel regarding hooks implementation for plugins ;)

PR #6722, opened 2019-11-24
feat(vue-app): upgrade `vue-router` to 3.1.x

clarkdo:
## Types of changes

## Description

As `vue-router@3.1` has a break change for programmatic `router.push`, so this pr is adding a small wrapper for reverting the break change, let's remove the wrapper and ship it in Nuxt 3.

BTW, actually user can still use the break change(promisable push), I think it's 
very rare in current user's code and also a good way for making new feature possible.

```js
router.push('/...', false)
```

## Checklist:


pi0 (approved):
LGTM! Thanks @clarkdo 👍

PR #6736, opened 2019-11-27
fix(nuxt-edge): global name is wrongly set to nuxt-edge

clarkdo:
[release]

Fix https://github.com/nuxt/nuxt.js/pull/6733#issuecomment-559126653

### DIFF
--- a/packages/config/src/config/_common.js
+++ b/packages/config/src/config/_common.js
@@ -12,8 +12,7 @@ export default () => ({
   mode: 'universal',
   modern: undefined,
 
-  // Globals
-  globalName: 'nuxt',
+  globalName: undefined,
   globals: {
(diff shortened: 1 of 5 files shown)


packages/config/src/config/_common.js
```diff
@@ -12,8 +12,7 @@ export default () => ({
   mode: 'universal',
   modern: undefined,
 
-  // Globals
-  globalName: 'nuxt',
+  globalName: undefined,
```
pi0: Can't we keep it here but using template literal to avoid escape?

clarkdo: We can, but it will be re-assign in options.js anyway

pi0 (approved):

PR #6740, opened 2019-11-28
test: use `got` instead of `request`

clarkdo:
As you may already notice, the tests are failing randomly recently.
From node 13.2, the condition becomes even worse, CI is 100% failing now.
Through my investigation, it's related to `request` lib, it seemed that there is reference leak and race condition, so this pr is gonna replace it with `got`


clarkdo:
Looks like issue still persists, let's continue to observe and investigate.

clarkdo:
I disabled `csp` test in `dev` for now, let's track in this pr.

clarkdo:
Update: Test is failed event in v2.8.0 and only happened in Node.js 13.2.0

clarkdo:
Update: Looks like it's related to vue renderer `runInNewContext: false` in prod mode

galvez:
@clarkdo In any case, switching to `got` is a Good Thing, I think. Been using it at work.

PR #6745, opened 2019-11-28
fix(vue-app): sanitize path with trailing slash in `getLocation` (#6744)

### LINKED ISSUE #6744: router.base is sanitized with trailing slash which does not match with path wihout trailing slash (spa mode only)
### Version

[v2.10.2](https://github.com/nuxt.js/releases/tag/v2.10.2)

### Reproduction link

[https://codesandbox.io/s/example-hello-world-no8hn](https://codesandbox.io/s/example-hello-world-no8hn)

### Steps to reproduce

Set nuxt.config
```js
{
  mode: 'spa',
  router: {
    base: "/app"
  }
}
```

Open localhost:3000/app

### What is expected ?

`index.vue` file from pages dir

404 This page could not be found

### Additional comments?

Works fine if open localhost:3000/app/

Works fine in universal mode, but logs ssr error (The client-side rendered virtual DOM tree is not matching...)

It's happening because of `getLocation` function (https://github.com/nuxt/nuxt.js/blob/701f2d033bbc421f85f888f9bf9100b176aac818/packages/vue-app/template/utils.js#L267) and because of `router.base` is sanitizing with trailing slash (https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L93)

So I propose change `getLocation` function

crutch12:
## Types of changes

## Description

Resolves: #6744

## Checklist:


packages/vue-app/template/utils.js
```diff
@@ -260,7 +260,7 @@ export function promisify (fn, context) {
 
 // Imported from vue-router
 export function getLocation (base, mode) {
-  let path = decodeURI(window.location.pathname)
+  let path = decodeURI(window.location.pathname + '/').replace('//', '/') // to get matched with sanitized router.base
```
pi0: Some important note:

- Having double slash inside path in url is valid and if we do replace, may introduce regressions. What if we just ensure pathname is ending with slash and if not appending?

Some  general notes:

1. (not related to this PR) but can we move this after `if(mode ==='hash')`. It can prevent an extra computed var when we don't need it
2. Can you please put comment above and start first chat with upper case? (`// To get`)

pi0 (dismissed):

packages/vue-app/template/utils.js
crutch12: You are right, I didn't think about this case and will fix it

crutch12: Fixed

packages/vue-app/template/utils.js
```diff
@@ -261,6 +261,10 @@ export function promisify (fn, context) {
 // Imported from vue-router
 export function getLocation (base, mode) {
   let path = decodeURI(window.location.pathname)
+  // To get matched with sanitized router.base
+  if (!/\/$/.test(path)) {
```
pi0: We can also use [`endsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith)

crutch12: I'm agree, didn't see this method previously, lol
Fixed

crutch12:
@pi0 fixed getLocation again, because at first I did mutating with path and it failed tests. Now I don't mutate path
Because of mutating a lot of apps could be crushed :D

packages/vue-app/template/utils.js
```diff
@@ -264,7 +264,8 @@ export function getLocation (base, mode) {
   if (mode === 'hash') {
     return window.location.hash.replace(/^#\//, '')
   }
-  if (base && path.indexOf(base) === 0) {
+  // To get matched with sanitized router.base add trailing slash
+  if (base && (path.endsWith('/') ? path : path + '/').startsWith(base)) {
```
pi0: This is an inline patch. Not sure but may introduce inconsistencies. WDYT @pimlie @manniL?

pimlie (commented):
I have nothing to add then what I've said before that we should add some sanitation methods to do this for us. It seems we have seen a rise of issues caused by trailing slashes and this is exactly what I was worried about (even though they are probably not all related)

Elhebert:
Are there any update about the status of this PR?

We can't update Nuxt in our project as this issue is blocking us. At the moment we are stuck on `2.6.1`.

crutch12:
I don't know the proper way how to fix my PR to be merged, so I just still waiting :)

@pi0 could you help me with PR?

pi0:
Hi @crutch12 Sure we should merge this fix just as @pimlie mentioned, it is not only source of inconsistencies for baseURL and trailing slash handling. (like regression introduced by https://github.com/nuxt/nuxt.js/pull/6775 with 2.12). Do you mind adding an e2e test to cover this change? Meanwhile working to make a url utility package to use it on all places instead of depending on sanetization(s)

crutch12:
Ok, I'll try

crutch12:
@pi0 I've added some e2e tests for checking spa mode router.base (in browser, based on spa fixtures)

Basically with/without trailing slash tests.

Could you check it out?

pi0 (approved):
Looks awesome! Thanks for your time @crutch12 :green_heart:

PR #6746, opened 2019-11-28
fix(ssr): segmentation fault (SIGSEGV) in Node.js 13.2

clarkdo:
## Description

Relate to https://github.com/nuxt/nuxt.js/pull/6740 , we found `segmentation fault` in ssr with Node.js 13.2.0.

This error happens when server is started frequently and in same Node.js context, throught my investigation, it'e related to the shared contenxt between ssr render and Node server, so this pr is to change `runInNewContext` to `once` in prod.

This change might be kind of breaking change, but as we're using `runInNewContext : false` in `dev` mode, so if there is any issue caused by running in new context, the issue should have been found in dev, so it should be somehow safe change.

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -401,7 +401,7 @@ export function getNuxtConfig (_options) {
   // Use runInNewContext for dev mode by default
   const { bundleRenderer } = options.render
   if (typeof bundleRenderer.runInNewContext === 'undefined') {
-    bundleRenderer.runInNewContext = options.dev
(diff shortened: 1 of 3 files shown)


galvez:
This explains a lot :) Looking forward to trying this out on nuxt-edge.

galvez (approved):
We should document change and list potential issues (shared exports and stuff like that).

pi0 (changes requested):
(mentioned at discord)

clarkdo:
Closed as @pi0 comment

PR #6768, opened 2019-12-07
fix: avoid parent `dataRefresh` in nested routes

mshahbazi:
in nested routes, when the route is changed, the `asyncData` of both parent and child are called when the child route has `query`. this is the case even if `watchQuery` is `false`.

this change resolved my issue.


clarkdo (commented):
The issue of this is related to https://github.com/nuxt/nuxt.js/pull/6244
Could you please have a look at that pr comments to see if it can solve your problem?

mshahbazi:
@clarkdo, #6244 resolves my issue. Thank you so much.

PR #6775, opened 2019-12-09
fix(vue-app): middleware not executing on double browser back navigation

jsonpoindexter:
## Types of changes

## Description

When a Nuxt app is in Universal mode it is possible that  in some cases a component's middleware does not execute.
This can happen when a user first navigates to a project page (example: Page B), server.js executes and then before client.js executes the user navigates to another page in the project (example: Page A). In this scenario https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/client.js#L810 will evaluate to `TRUE` and not execute the middleware(s) for page A.

Resolves: https://cmty.app/nuxt/nuxt.js/issues/c10049 / https://github.com/nuxt/nuxt.js/issues/6716

## Checklist:

I could not find a way to programmatically reproduce this issue ie. changes routes after server.js has executed, but before client.js executes.


Atinux:
Actually this makes sense, even if I cannot reproduce it on my machine (might be too fast).

Can you reproduce this on your slower machine when running `nuxt start`?

jsonpoindexter:
@Atinux yes, I am able to reproduce this in production mode.

jeremybradbury (dismissed):

pi0 (approved):

pi0:
Thanks @jsonpoindexter

PR #6792, opened 2019-12-16
fix(generator): no copying built files when publicPath is an url

hgl:
## Types of changes

## Description

This is a small change so I didn't create an issue for it first. This is also technically a small breaking change, but the old behavior was never documented.

I have already described the intention in the git commit message, but here it is again:

> When build.publicPath is an url, there is no point in copying over built css and js files, since they should be uploaded to CDN separately and already lives in a separate documented path.

In my case, since I have to upload `.nuxt/dist/client` to the CDN, I have to filter out the built js files in the dist folder, which is uploaded separately (I think this should be the case pretty generally when you use a CDN). And having to perform an extra step to remove these files first feels like a waste.

I don't have a strong opinion on this, if any nuxt team member has an objection, feel free to close the PR.

BTW, I'd like to send two more PRs, but I'm not sure if they will be accepted. So I'll just ask them here if that's ok.

1. Creating `.nojekyll` should be optional, maybe `generate.nojekyll: bool` and default to true?
2. There shouldn't be a `__layout` element in the generated html (there are a few issues asking for this), and technical wise, it only contains one child element, and not as the __nuxt element which might contain several thus is mandatory.

## Checklist:

Not sure how to test this, the generator test doesn't have any test that actually generates files.

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -167,6 +167,7 @@ export default class Generator {
   async initDist () {
     // Clean destination folder
     await fsExtra.remove(this.distPath)
+    await fsExtra.mkdirp(this.distPath)
 
     await this.nuxt.callHook('generate:distRemoved', this)
 
@@ -174,7 +175,9 @@ export default class Generator {
     if (await fsExtra.exists(this.staticRoutes)) {
       await fsExtra.copy(this.staticRoutes, this.distPath)
     }
-    await fsExtra.copy(this.srcBuiltPath, this.distNuxtPath)
+    if (!isUrl(this.options.build.publicPath)) {
+      await fsExtra.copy(this.srcBuiltPath, this.distNuxtPath)
+    }
 
     // Add .nojekyll file to let GitHub Pages add the _nuxt/ folder
     // https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/


Atinux:
Hey @hgl 

To better understand your use-case, you are using `nuxt generate` with a `publicPath` where you upload your `.nuxt/dist/client` to your CDN but the `dist/*.html` to another CDN?

About your `1.` and `2.`, I agree, actually we may have to dig a bit more about `__layout` but I think it's mostly for testing purpose (so it can be env related).

Atinux:
About the `.nojekyll` file, it can be replaced by just have this file in the `static/` directory. But that would be a breaking change so I think it could be for Nuxt 3, as well as the `__layout` removal, but agree for these changes.

hgl:
> To better understand your use-case, you are using nuxt generate with a publicPath where you upload your .nuxt/dist/client to your CDN but the dist/*.html to another CDN?

@Atinux mostly correct. In my case, I use s3 + cloudfront, and I need to upload twice, first for the built static files(eg., js, css, images etc), then for the html files. For two reasons:

1. Static files has hashes in their names, thus can have a far future expiration date.
2. Static files never overwrite but html files do, in order to achieve zero downtime, static files have to be uploaded first.

As for `.nojekyll`, I mostly just concern about omitting its generation, since not everyone deploys to a jekyll environment like GitHub pages, but good to know it can be replaced with the `static/` trick.

So I should wait for the nuxt3 release instead of submitting the PRs?

hgl:
I guess I probably should just use wildcard to match files instead of deleting static files.

Atinux:
Actually I would like to have your PR already so we can merge them in our `next` branch that will be related to Nuxt 3 :)

hgl:
@Atinux OK, will take a look at the `next` branch when I can. :)

Issue #3877, opened 2018-09-11
CSS from multiple layouts is being included on individual pages


markplewis:
### Version

[v1.4.2](https://github.com/nuxt.js/releases/tag/v1.4.2) - v2.4.2 (see below).

### Reproduction link

[https://github.com/markplewis/nuxt-layout-css-issue](https://github.com/markplewis/nuxt-layout-css-issue)

### Steps to reproduce

1. Clone the following example repo.
2. Run `npm install` then `npm run dev`.
3. Navigate between the pages and inspect the CSS that's applied to the `<body>` element via your browser's dev tools.

https://github.com/markplewis/nuxt-layout-css-issue

### What is expected ?

When a page specifies a layout, like this:

**pages/index.vue:**
```
<script>
  export default {
    layout: "home"
  }
</script>
```

Only the CSS from the corresponding layout file should be included in the page:

**layouts/home.vue:**
```
<style>
  body {
    background-color: steelblue;
  }
</style>
```

### What is actually happening?

The CSS from every layout file is being included on every page, regardless of which layout is specified in the page's `.vue` file.

<img width="423" alt="screenshot" src="https://user-images.githubusercontent.com/2489555/45423911-d0be9000-b662-11e8-93c9-3c6def7a6fb7.png">

### Additional comments?

My example repo is using Nuxt Edge.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7723">#c7723</a>)</em></sub></div>

manniL:
Can confirm this bug.

Workaround:

```js
 build: {
    splitChunks: {
      layouts: true
    }
}
```

markplewis:
This is also a problem in Nuxt v1 (see [the `v1` branch of my demo repo](https://github.com/markplewis/nuxt-layout-css-issue/tree/v1)). Styles from multiple layouts are still being pulled into the page, but at least I'm not seeing styles from the `default` layout:

<img width="401" alt="screen shot 2018-09-12 at 7 55 33 am" src="https://user-images.githubusercontent.com/2489555/45423507-8e488380-b661-11e8-98b2-0f9208647952.png">

hareku:
I have same issue.

Legym:
@hareku I was able to solve my issue using

 build: {
    splitChunks: {
      layouts: true
    }
}

hareku:
@Legym
[...]

PR #6806, opened 2019-12-18
feat(vue-app): remove __layout element

hgl:
## Types of changes

## Description

As discussed previously, `__layout` element isn't necessary. This PR removes it. Since it's a breaking change, PR is sent to Nuxt 3.

@Atinux 

## Checklist:


clarkdo (commented):
We can add a temporary option for disabling `__layout` and compatibility in 2.x, then remove both of them in 3.

Atinux:
Sounds good @clarkdo

manniL:
I see an opportunity here to somehow provide a workaround for #3877 by providing a named class/id instead of just `__layout`🤔

Atinux:
Something like this @manniL ?

```js
const templateEl = h('div', {
  domProps: {
     class: 'layout-' + this.layoutName
  },
  key: this.layoutName
}, [ layoutEl ])
```

manniL:
@Atinux Exactly ☺️ (now we just have to ensure that the layout name can be used as valid css class, e.g. by stripping spaces

pi0 (commented):
I think makes sense merging this for v3 :)

pi0:
Hi, @hgl thanks again for your contribution. As we are doing a full rewrite for vue-app, I'm going to close this PR but the unnecessary class will be removed.

BobbyJonas:
It's been two years. and I've met the same problem.
How're things going now

pi0:
Check out Nuxt 3. Layout is an opt-in component: 

- https://v3.nuxtjs.org/guide/directory-structure/layouts
- https://github.com/nuxt/framework/blob/main/packages/nuxt/src/app/components/layout.ts

PR #6829, opened 2019-12-26
fix(router): prevent navigation duplicated error

kevinmarrec:
Prevents error if onAbort is not given

## Types of changes

## Description
Since we updated Nuxt to last `vue-router`, there's now an error when navigating to the same route the user already is, when using `router.push`.

See more info : https://github.com/vuejs/vue-router/issues/2881#issuecomment-520554378

Around Nuxt it can look as a breaking change, cause the error was previously (in previous `vue-router` version) silent, and now it just breaks.

So, until Nuxt 3, we should still silently catch the error, unless the issue specify the `onAbort` argument to `push` function and manually handle the error in case there is a duplicated navigation.

## Checklist:

- [ ] All existing tests are passing.


Atinux:
Sounds good, but weird tests are failing:
Screenshot 2019-12-26 at 11 21 12

kevinmarrec:
Yes, I wanted to be sure tests passed before ticking the test checkbox and ask for review (could have been draft pr though), I need to investigate why the thing is `undefined`

kevinmarrec:
Alright, sounds it returns a Promise only if both `onComplete` & `onAbort` are not defined
https://github.com/vuejs/vue-router/blob/dev/src/index.js#L153

kevinmarrec:
Hmm, in fact I think Nuxt just needs to give default `onComplete` & `onAbort` as empty functions and that's it, gonna make local tests to be sure.

atinux (dismissed):

kevinmarrec:
@Atinux Doesn't seem to work, I still have the "NavigationDuplicated" error locally when navigating to same route, it may be something that can't be change by Nuxt by changing the `push` function :/

kevinmarrec:
As It is intended by `vue-router` itself to throw the error, I'm not sure that we should override such behavior around Nuxt, cause it will make some users not being able to redefine their own `push` or `replace` over original one. As it would go userOverride -> nuxtOverride -> original, making user being caught if something he/she doesn't want is done by Nuxt in the override.

So I think we should not continue trying to find a solution that should be implemented around Nuxt.

If someone is looking to prevent the error and the routing if you're pushing or replacing the same route, you can create a Nuxt plugin and go this way :

```js
// Override Push
const originalPush = VueRouter.prototype.push
VueRouter.prototype.push= function (location) {
  if (this.resolve(location).route.fullPath === this.currentRoute.fullPath) {
    return // You can also if you wish fire an event through `this.app.$nuxt.$emit`
  } else {
    originalReplace.push(this, location)
  }
}

// Override Replace
const originalReplace = VueRouter.prototype.replace
VueRouter.prototype.replace = function (location) {
  if (this.resolve(location).route.fullPath === this.currentRoute.fullPath) {
    return // You can also if you wish fire an event through `this.app.$nuxt.$emit`
  } else {
    originalReplace.call(this, location)
  }
}

liaohui1080:
When can it be solved?

danielroe:
@liaohui1080 This issue was resolved in a different PR. If you are experiencing a bug with Nuxt, please do raise a new issue 🙏

PR #6846, opened 2020-01-04
perf(vue-renderer):  don't serialize session when `injectScripts` is false

404-NF:
## Types of changes

## Description

`serializedSession` is always computed even though not use the result. it use too much cpu in some render options.
now, `serializedSession` will be computed only when this result needed. 

before: 
image

after: 
image

## Checklist:


pi0:
Thanks @404-NF for this PR. Would you please fix lint issue?

PS: This perf change is only effective when `this.options.render.injectScripts` set to `false`

pi0 (approved):

atinux (approved):

PR #6854, opened 2020-01-11
fix(vue-app): add fallback context for error

negezor:
## Types of changes

## Description

Fixes bug https://github.com/nuxt/nuxt.js/issues/6517#issuecomment-573280207

## Checklist:


packages/vue-app/template/index.js
```diff
@@ -120,7 +120,7 @@ async function createApp (ssrContext) {
         err = err || null
         app.context._errored = Boolean(err)
         err = err ? normalizeError(err) : null
-        const nuxt = this.nuxt || this.$options.nuxt
+        const nuxt = (this && (this.nuxt || (this.$options && this.$options.nuxt))) || app.nuxt
```
pi0: What do you think expanding to normal if else syntax now and add a comment on top of it pointing to issue why nuxt cannot be available in some cases?

pi0 (commented):
Thanks for PR 😊

packages/vue-app/template/index.js
manniL: Could also be a candidate for an own function ☺️

negezor: The change allows you to use the new API and does not break backward compatibility.

pi0: Seems fine for me as a simple work-round but having some docs on top of it regarding reason, next plans (refactor to function), etc could be nice :)

Atinux: What about:

```suggestion
        let nuxt = app.nuxt // to work with @vue/composition-api, see https://github.com/nuxt/nuxt.js/issues/6517#issuecomment-573280207
        if (this) {
          nuxt = this.nuxt || this.$options.nuxt
        }
```

Atinux: What about:

```suggestion
        let nuxt = app.nuxt // to work with @vue/composition-api, see https://github.com/nuxt/nuxt.js/issues/6517#issuecomment-573280207
        if (this) {
          nuxt = this.nuxt || this.$options.nuxt
        }
```

negezor: Looks nice

packages/vue-app/template/index.js
```diff
@@ -120,7 +120,10 @@ async function createApp (ssrContext) {
         err = err || null
         app.context._errored = Boolean(err)
         err = err ? normalizeError(err) : null
-        const nuxt = this.nuxt || this.$options.nuxt
+        let nuxt = app.nuxt // to work with @vue/composition-api, see https://github.com/nuxt/nuxt.js/issues/6517#issuecomment-573280207
+        if (this) {
```
pi0: Should we also check `if (!nuxt && this)` ? 

/cc @Atinux

Atinux: Actually I prefer to keep the behaviour with the instance to avoid breaking changes, we will refactor this part with Vue 3 anyway :)

pi0 (approved):

PR #6864, opened 2020-01-13
fix(webpack): provide global `URL` and `URLSearchParams`

### LINKED ISSUE #4914: global.URL is missing
### Version

[v2.4.0](https://github.com/nuxt.js/releases/tag/v2.4.0)

### Reproduction link

[https://codesandbox.io/s/ww5n6p5mll](https://codesandbox.io/s/ww5n6p5mll)

### Steps to reproduce

Accessing URL object like `new URL('http://example.com');` anywhere in server side code.

### What is expected ?

URL object is accessible globally in server side since it is defined as global interface in node.js.
https://nodejs.org/docs/latest-v10.x/api/globals.html#globals_url

Accessing URL object in server side code causes "ReferenceError URL is not defined".

### Additional comments?

It had been accessible before and including v2.3.4.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8582">#c8582</a>)</em></sub></div>

pi0:
## Types of changes

## Description

Fixes #4914 by using built-in url package. Tested against node 8.x too. If we want even more consistency, can switch to https://www.npmjs.com/package/whatwg-url.

Related PRs: vuejs/vue#10414.

## Checklist:


atinux (dismissed):

atinux (approved):

clarkdo (commented):
As ProvidePlugin can be easily configured by build.extend, should we provide this special config only for URL polyfill?

pi0:
@clarkdo I think so because currently it is broken (#4914). If we do a polyfill, it makes sense to also allow disabling/adjusting this fix (some reported only `whatwg-url` works). And BTW this option is not documented.

dhritzkiv:
Is there a trick to having this work?

I have a universal app that under v2.11.0 I was able to get URL working in client and SSR by setting `config.render.bundleRenderer.runInNewContext` to `false`.

Updating to 2.12.0 results in `URL is not a constructor`, no matter if I remove that setting or not. Is there something else I need to enable/disable or install in order to have URL work in both SSR and client?

pi0:
@dhritzkiv ProvidePlugin is a webpack feature and indeed yes the reason is runInNewContext is enabled by default in dev mode. How sad it broke your project :( Would you please help reproducing in a simple repo to quickly investigate?

PR #6868, opened 2020-01-14
feat: improve programatic usage

pi0:
## Types of changes

## Description

**UPDATE:** See [docs](https://nuxtjs.org/api/nuxt)

- Refactor config loading logic into `@nuxt/config` package and abstract from CLI arguments
- Expose `loadNuxtConfig`, `loadNuxt`, `getBuilder`, `build` and `getGenerator`

Goal: Making programmatic usage better for making custom solutions without the need to worry about loading config, handling breaking changes for class constructors. Also making `@nuxt/cli` more lightweight only focusing on its own functionality.

Example:
```js
import { loadNuxt } from 'nuxt'

// Load nuxt instances with config loaded from the current directory
const nuxt  = await loadNuxt({ for: 'dev' })

// Specify rootDir (and/or configFile)
const nuxt = await loadNuxt({ rootDir, for: 'start' })
```

New programmatic usage for using as a middleware: [See Example](https://github.com/nuxt/nuxt.js/tree/feat/get-nuxt/examples/programmatic)

## Checklist:

- [ ] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1811)


pi0:
(some GH steps failing because of cache issues)

kevinmarrec (dismissed):

kevinmarrec:
@pi0  What's wrong with `node-sass` audit ? Doesn't sound related to cache

clarkdo (dismissed):
@pi0 @kevinmarrec The github action failure is because of a new found DoS vulnerability in all versions of node-sass which doesn't have a patch yet.
 https://www.npmjs.com/advisories/961

packages/core/src/index.js
```diff
@@ -1,3 +1,12 @@
+import { loadNuxtConfig } from '@nuxt/config'
+import Nuxt from './nuxt'
```
clarkdo: We have `export { default as Nuxt } from './nuxt'` below, now it can be :

```js
export Nuxt
```

packages/generator/src/index.js
```diff
@@ -1 +1,6 @@
+import Generator from './generator'
```
clarkdo: Same as above

packages/builder/src/index.js
```diff
@@ -1 +1,6 @@
+import Builder from './builder'
 export { default as Builder } from './builder'
+
+export function getBuilder (nuxt) {
```
clarkdo: Same as https://github.com/nuxt/nuxt.js/pull/6868/files#r366453215

packages/builder/src/index.js
```diff
@@ -1 +1,6 @@
+import Builder from './builder'
```
clarkdo: Same as https://github.com/nuxt/nuxt.js/pull/6868/files#r366453215

packages/core/src/index.js
pi0: I was thinking this may ensure less breaking changes and the fact that we don't export something else by accident. Also fewer characters :)

examples/programmatic/scripts/dev.js
```diff
@@ -0,0 +1,9 @@
+const { loadNuxt, build } = require('nuxt')
+
+const main = async () => {
+  const nuxt = await loadNuxt({ for: 'dev' })
+  build(nuxt)
```
clarkdo: Should have await here?

examples/programmatic/scripts/start.js
```diff
@@ -0,0 +1,8 @@
+const { loadNuxt } = require('nuxt-start')
```
clarkdo: Should we expose `loadNuxt` in `nuxt-start` ? If so, is this `loadNuxt` only for `start` ?

packages/builder/src/index.js
```diff
@@ -1 +1,10 @@
...
```
clarkdo: Can be `export default Builder` ?

packages/cli/src/utils/config.js
```diff
@@ -1,67 +1,20 @@
 import path from 'path'
-import consola from 'consola'
 import defaultsDeep from 'lodash/defaultsDeep'
-import { defaultNuxtConfigFile, getDefaultNuxtConfig } from '@nuxt/config'
-import { clearRequireCache, scanRequireTree } from '@nuxt/utils'
-import esm from 'esm'
+import { loadNuxtConfig as _loadNuxtConfig, getDefaultNuxtConfig } from '@nuxt/config'
```
clarkdo: Will `@nuxt/config loadNuxtConfig` be confused with `@nuxt/cli loadNuxtConfig`, maybe change `@nuxt/config loadNuxtConfig` sth else ?

examples/programmatic/scripts/start.js
Atinux: `loadNuxt` is for both `nuxt` and `nuxt-start`, for example, on serverless function where you only want to load nuxt for production, you only include `nuxt-start` to reduce the lambda size

examples/programmatic/scripts/dev.js
Atinux: Not in dev, since it will allow to display our fancy loading screen 😎

packages/cli/src/utils/config.js
Atinux: Actually `loadNuxtConfig` from `@nuxt/cli` should be used only for 

I would rather not export `loadNuxConfig` from `@nuxt/cli` anymore and rename it to `loadNuxtConfigFromArgv(argv, configContext)`, might be a really rare breaking change.

examples/programmatic/scripts/dev.js
clarkdo: It might need a good explanation in doc or call await build after listen for consistent usage

examples/programmatic/scripts/start.js
clarkdo: I understand it’s same api, but as it’s imported from nuxt-start, so for:’start’ makes me feel confusing if it’s duplicate and also if I can put any other value here.

Atinux: I see what you mean, actually we also have the `'dry'` value that could work for `nuxt-start`

packages/cli/src/utils/config.js
pi0: Indeed i considered both points. To avoid breaking changes of `@nuxt/cli` package, export is there but it is not even documented so should be fine. New `loadNuxtConfig` export is exposed by core to distro but i think for now we should only document `loadNuxt`

examples/programmatic/scripts/dev.js
pi0: Well i think no one should use dev programatic usage because it is just a bad DX. It is just to make current usage easier.

packages/builder/src/index.js
Atinux: cc @pi0

packages/cli/src/utils/config.js
Atinux: Agree :)

examples/programmatic/scripts/dev.js
clarkdo: In create-nuxt-app, we provide this usage 😅

Atinux: Using `await build(nuxt)` could be done after `nuxt.server.listen()` indeed.

Anyway we need to remove the backend framework support in create-nuxt-app to suggest the serverMiddleware usage

packages/builder/src/index.js
pi0: The way of exporting classes can be improved in another PR :)

atinux (approved):

clarkdo:
Is this a breaking change?

If not, we’ll have 2 different programmatic ways, are we planning to remove old way in next release?

pi0:
@clarkdo No, this PR doesn't break the current type of usage but adding new recommended method of getting nuxt instance.

crutch12:
Hi, @pi0 
Suddenly found out that `publicRuntimeConfig`/`privateRuntimeConfig` doesn't work with legacy programatic usage:

**server.js**
```js
const { Nuxt, Builder } = require('nuxt');
const nuxt = new Nuxt(config);
await nuxt.ready();
```

**nuxt.config.js**
```js
export default {
    publicRuntimeConfig: ($env) => {
      return {
        NODE_ENV: $env.NODE_ENV,
        wtf: 'WTF',
      };
    },
}
```

The resulting **script**
```
<script>window.__NUXT__={config:{},staticAssetsBase:void 0}</script>
```
instead of
```
<script>window.__NUXT__={config:{NODE_ENV:"development",wtf:"WTF"},staticAssetsBase:void 0}</script>
```

It would be great to **deprecate** legacy syntax or add support for `runtimeConfig`

PR #6880, opened 2020-01-16
feat(vue-app): new fetch syntax

### LINKED ISSUE #3776: Allow layouts or the error layout to use fetch/asyncData

atinux:
## Types of changes

## Description

ℹ️ See RFC: https://github.com/nuxt/rfcs/issues/27
ℹ️ See original pull request first: https://github.com/nuxt/nuxt.js/pull/5118
ℹ️ Documentation about its usage is available on https://github.com/nuxt/nuxt.js/tree/feat/new-fetch/examples/new-fetch

This PR is not a breaking change since it activates this new fetch only when `fetch` is used without expecting any arguments.

### Old fetch

```js
export default {
  fetch(context) {
    // old fetch logic
  }
}
```

With this pr, this syntax is now **deprecated** in favour of `middleware(context)`

### New fetch

```js
export default {
  fetch() {
    // New fetch, access to `this`
  }
}
```

Resolves #3776, #32 and #127

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1823)


MLDMoritz:
If you need any help please let me know.

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,53 @@
...
+      Vue.util.defineReactive(this, '$fetchError', null)
+    }
+  },
+  created () {
+    if (hasFetch(this) && isSsrHydration(this)) {
+      // Hydrate component
+      this._hydrated = true
+      this.$isLoading = false
+      this._lastFetchAt = Date.now()
+
```
galvez: @Atinux perhaps we should avoid using `_` for public properties we expect people to use (and are even part of usage examples). Maybe a per-component `$fetchState` with `{ hydrated, loading, timestamp }` -- wdyt? (Or even attach these properties to the `$fetch` function itself)

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,53 @@
...
+      this._ssrKey = +this.$vnode.elm.dataset.ssrKey
+      const data = nuxtState.data[this._ssrKey]
+
+      // If fetch error
+      if (data && data._error) {
+        return this.$fetchError = data._error
```
galvez: https://eslint.org/docs/rules/no-return-assign :)

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,53 @@
...
+  beforeMount () {
+    if (!this._hydrated && hasFetch(this)) {
+      this.$fetch()
+    }
+  },
+  methods: {
+    async '$fetch' () {
+      this.$nuxt.nbFetching++
+      this.$isFetching = true
+      this.$fetchError = null
```
galvez: @Atinux Likewise here, instead of having `$fetchError`, we could simply have `$fetch.error` or `$fetchState.error` where `$fetchState` would also give access to the other properties.

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -0,0 +1,44 @@
...
+    if (hasFetch(this) && this._fetchOnServer) {
+      const data = Object.assign({}, this.$data)
+
+      try {
+        await this.$options.fetch.call(this)
+      } catch (err) {
+        this.$fetchError = err
+      }
+      // Define and ssrKey for hydration
+      this._ssrKey = this.$ssrContext.nuxt.data.length
```
galvez: @Atinux can you give a little more detail on why this works? At first glance it seems prone to collisions?

galvez (dismissed):

packages/vue-app/template/mixins/fetch.server.js
Atinux: Well it's to generate the `data-ssr-key` to keep `index` of `window.__NUXT__.data` when hydrating, since the **same component** with `fetch` can be used multiple times.

packages/vue-app/template/mixins/fetch.client.js
Atinux: Yes, I will change to have something like `$fetchState` and also introducing `<nuxt-fetch>` component to use it like https://github.com/posva/vue-promised

packages/vue-app/template/mixins/fetch.client.js
Atinux: Actually this is just for layouts here to keep track of actual fetch() running

packages/vue-app/template/mixins/fetch.server.js
bf: But why is it data.length and not some hash of the data?

Atinux: `nuxt.data` is actually an array, not sure to understand what you mean here

DominusKelvin:
Wow. Awesome job you guys

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -0,0 +1,47 @@
...
+  async serverPrefetch () {
+    if (hasFetch(this) && this._fetchOnServer) {
+      const data = Object.assign({}, this.$data)
+
+      try {
+        await this.$options.fetch.call(this)
+      } catch (err) {
+        this.$fetchState.error = normalizeError(err)
+      }
+      // Define and ssrKey for hydration
```
jamesgeorge007:
```suggestion
      // Define an ssrKey for hydration
```

packages/vue-app/template/components/nuxt-child.js
```diff
@@ -42,15 +43,17 @@ export default {
...
-    listeners.beforeEnter = (el) => {
-      // Ensure to trigger scroll event after calling scrollBehavior
-      window.<%= globals.nuxt %>.$nextTick(() => {
-        window.<%= globals.nuxt %>.$emit('triggerScroll')
-      })
-      if (beforeEnter) {
-        return beforeEnter.call(_parent, el)
+    if (process.client) {
+      // Add triggerScroll event on beforeEnter (fix #1376)
+      const beforeEnter = listeners.beforeEnter
```
jamesgeorge007: Maybe use the destructuring assignment syntax.

packages/vue-app/template/mixins/fetch.server.js
galvez: @bf component `asyncData` objects are pushed to `window.__NUXT__` in the order they're loaded, so that's why using `data.length` as a unique per-component key works.

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -0,0 +1,47 @@
...
+    if (hasFetch(this)) {
+      this._fetchOnServer = this.$options.fetchOnServer !== false
+      Vue.util.defineReactive(this, '$fetchState', {
+        pending: !this._fetchOnServer,
+        error: null,
+        timestamp: Date.now()
+      })
+    }
+  },
+  async serverPrefetch () {
```
pimlie: Not sure if this is also possible to add programmatically, if not would it make sense to cache the hasFetch result in beforeCreate? (Probably not)

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -0,0 +1,47 @@
...
+    if (hasFetch(this) && this._fetchOnServer) {
```
pimlie: Prefer to reverse these conditions

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -0,0 +1,47 @@
+import Vue from 'vue'
+import { hasFetch, normalizeError } from '../utils'
+
+function getDataDiff(o1, o2) {
+  return Object.keys(o2).reduce((diff, key) => {
+    if (o1[key] === o2[key]) {
+      return diff
+    }
+    return {
+      ...diff,
```
pimlie: Please see: https://www.richsnapp.com/blog/2019/06-09-reduce-spread-anti-pattern

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,61 @@
+import Vue from 'vue'
+import { hasFetch, normalizeError } from '../utils'
+
+const isSsrHydration = (vm) => vm.$vnode && vm.$vnode.elm && vm.$vnode.elm.dataset && vm.$vnode.elm.dataset.ssrKey
+const nuxtState = window.<%= globals.context %>
+
+export default {
+  beforeCreate () {
+    if (hasFetch(this)) {
```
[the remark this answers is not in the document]
pi0: Done

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,61 @@
...
+      if (data && data._error) {
+        this.$fetchState.error = data._error
+        return
+      }
+      for (const key in data) {
+        this[key] = data[key]
+      }
+    }
+  },
+  beforeMount () {
```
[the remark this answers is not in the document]
pi0: Let's keep conv one place :)

packages/vue-app/template/mixins/fetch.server.js
pi0: Done :)

packages/vue-app/template/mixins/fetch.server.js
pi0: Done (actually using cache)

packages/vue-app/template/mixins/fetch.server.js
pi0: @pimlie I refactored to use cache in beforeCreate. Let's keep programmatically conv at https://github.com/nuxt/nuxt.js/pull/6880#discussion_r372854230

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -0,0 +1,61 @@
...
+          await new Promise(resolve => setTimeout(resolve, delayLeft))
+        }
+        this.$fetchState.error = error
+        this.$fetchState.pending = false
+        this.$fetchState.timestamp = Date.now()
+        this.$nextTick(() => this.$nuxt.nbFetching--)
+      }
+    }
+  },
+  created () {
```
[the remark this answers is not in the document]
pi0: I think it worth to consider. Mixin merge should be quick also call to hasFetch is quick and can be cached too.. But it is important to not add unnecessary hooks like `beforeMount()` to all components. WDYT @Atinux? @pimlie do you have any benchmarks?

examples/new-fetch/layouts/default.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div>
+    <!-- <p>Fetching: {{ $nuxt.isFetching }} ({{ $nuxt.nbFetching }})</p> -->
```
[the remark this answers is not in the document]
Atinux: Yes :)

examples/new-fetch/layouts/default.vue
```diff
@@ -0,0 +1,28 @@
...
+    <nuxt />
+  </div>
+</template>
+
+<script>
+export default {
+  head: {
+    link: [
+      { rel: 'stylesheet', href: 'https://cdn.jsdelivr.net/gh/kognise/water.css@latest/dist/light.min.css' }
```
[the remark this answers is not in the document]
Atinux: Sure we can update to use `1.4.0`

packages/vue-app/template/mixins/fetch.client.js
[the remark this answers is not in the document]
pi0: @pimlie Already done (by approval of @Atinux) :)

packages/vue-app/template/components/nuxt-child.js
```diff
@@ -13,7 +13,8 @@ export default {
       default: undefined
     }
   },
-  render (h, { parent, data, props }) {
+  render (_, { parent, data, props }) {
+    const h = parent.$createElement
```
pimlie: Just wondering, what is the benefit of doing this?

pimlie (commented):
What about supporting fetchOnServer as a function like we discussed on discord?

packages/vue-app/template/components/nuxt-child.js
Atinux: Making `nuxt-child` invisible and removing a virtual node normally. See https://github.com/vuejs/vue-router/blob/dev/src/components/view.js#L19

examples/new-fetch/layouts/default.vue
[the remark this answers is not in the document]
pi0: Done

packages/vue-app/template/components/nuxt-child.js
pimlie: Interesting (somewhat hidden) Vue feature, thanks for sharing

pi0 (dismissed):

pi0 (approved):

husayt:
@pi0 @Atinux 
I have been playing with new fetch for last few days and came across some practical hurdles, the most painful of them is error handling. The way it currently implemented it allows only hiding the component through `$fetchState.error` variable. So if `$fetchState.error===true` you can only hide your component. But often I want to display error page if data is not found, as I would normally do with asyncdata via `error` method. So I try the following and it works fine on client-side, but if error happens during SSR I get ` The client-side rendered virtual DOM tree is not matching server-rendered content.` warning. 


```
 <span v-if="$fetchState.pending" >loading</span>
 <span v-else-if="$fetchState.error" >error</span>
 <Quote :item="quote" />
       
.....

  async fetch() {
     try {
      this.quote = await this.$api.getQuote(this.$route.params.id)
    } catch (e) {
      this.$nuxt.error({
        statusCode: 404,
        message: "Quote not found " + this.$route.params.id
      })
      this.$fetchState.error = true
      // throw new Error("Aforizm tapılmadı: " + this.$route.params.id)
    }
  },
```

So how this most common error handling pattern should implemented via new Fetch hook ?

Atinux:
Hi @husayt 

Actually, since this happens in `serverPrefetch` hook, we cannot change the the root element on SSR anymore. So we are blocked only at the component level.

I will recommend to use `asyncData` when it's about the page data actually @husayt 

I believe that with Vue 3 we may find an alternative to it 🤞

husayt:
Thanks @Atinux ,
since this hook was introduced I was trying to clarify for myself role of new fetch along other hooks. Initially I thought that asyncdata might not be required anymore, but as this example shows that asyncdata has still got a valid usecase.

For now the main benefit for me it seems that new fetch allows to asynchronously provide data  at component level  for both server and client side, which is itself is huge and allows new patterns not available to us before.

Another gotcha with fetch is that it happens at the very end. Would be very interesting to explore firing data request in page middleware/asyncdata and awaiting in fetch

PR #6892, opened 2020-01-20
feat(vue-renderer): Hook VueMeta  SPA / Static to modify VueMeta data while generatating

harlan-zw:
## Types of changes

## Description

For SPA and static site builds, the generated HTML pages all use the meta options from `nuxt.config.js`. This is a problem for SEO (obviously) and for link sharing where the meta content of the HTML will be parsed to give a preview. While a universal or server-sided rendered app would solve this, I'd rather not have to deal with the complexities involved when there is a fairly simple solution (I only need the meta pre-rendered, nothing else).

What I'm proposing is a hook for vue-meta allowing the app to decide on a more granular level what meta to use for each route. I believe this is in-line with existing core functionality and could open up other use cases.

I would hope one day the vue-meta plugin could just work for this for static / spa built sites but I think this can be a good intermediate solution.

Example: with this code block a user can generate all their static HTML to have individual titles without any effort 

```
hooks: {
    'vue-renderer': {
      spa: {
        meta ({ head, renderContext }) {
          const urlParts = renderContext.url.split('/')
          head.title = startCase(urlParts.pop())
        }
      }
    },
  },
```

## Checklist:

- [x] My change requires a change to the documentation. - **No documentation page for `vue-renderer` hooks that I can see?**
- [ ] I have added tests to cover my changes (if not applicable, please state why) - **Wasn't sure exactly where to put these tests, wasn't a SPA generate unit test from what I could see?**

### DIFF
--- a/packages/vue-renderer/src/renderers/spa.js
+++ b/packages/vue-renderer/src/renderers/spa.js
@@ -58,7 +58,7 @@ export default class SPARenderer extends BaseRenderer {
       }
 
       // Allow the head meta passed to VueMeta to be overridden
-      await this.serverContext.nuxt.callHook('vue-renderer:spa:meta', { head, renderContext });
+      await this.serverContext.nuxt.callHook('vue-renderer:spa:meta', { head, renderContext })
 
       const m = VueMeta.generate(head || {}, this.vueMetaConfig)
 
--- a/packages/vue-renderer/src/renderers/spa.js
+++ b/packages/vue-renderer/src/renderers/spa.js
@@ -57,6 +57,9 @@ export default class SPARenderer extends BaseRenderer {
         head = cloneDeep(this.options.head)
       }
 
+      // Allow the head meta passed to VueMeta to be overridden
+      await this.serverContext.nuxt.callHook('vue-renderer:spa:meta', { head, renderContext });
+
       const m = VueMeta.generate(head || {}, this.vueMetaConfig)
 
       // HTML_ATTRS


pimlie (commented):
Thanks for taking the time to explain why you need this.

Have you perchance looked at `mode: 'universal'` as well, could you maybe use that? Universal mode is meant to solve the issue you describe in this pr, it will prerender the SPA for every single page (and additional dynamic routes you've configured in your nuxt.config). If you set `generate.fallback` to true or a filename an non-prerendered SPA will also be saved.

loonpwn:
> Thanks for taking the time to explain why you need this.
> 
> Have you perchance looked at `mode: 'universal'` as well, could you maybe use that? Universal mode is meant to solve the issue you describe in this pr, it will prerender the SPA for every single page (and additional dynamic routes you've configured in your nuxt.config). If you set `generate.fallback` to true or a filename an non-prerendered SPA will also be saved.

Hi @pimlie 

Thanks for the reply and pointing that out. In retrospect this was really obvious. When I was experimenting with universal the first time I somehow ended up with a client / server folder, which I assumed wouldn't work for s3, I guess this was down to me not configuring it correctly. 

Have just tried now and it's all good... So will close this PR

Sford4:
> In retrospect this was really obvious. When I was experimenting with universal the first time I somehow ended up with a client / server folder, which I assumed wouldn't work for s3, I guess this was down to me not configuring it correctly.
> 
> Have just tried now and it's all good... So will close this PR

@loonpwn @pimlie can you explain to me what was obvious? I'm trying to deploy as a static site to S3, but I need SEO, while wanting the cost savings of serving statically from S3, I've explained the issue I'm having with `nuxt generate` in another issue thread [here](https://github.com/nuxt/nuxt.js/issues/7717).

Is it possible to do `nuxt build nuxt start` then deploy that on S3?  What configurations were you alluding to that I can tweak to set this up?

Is there a guide I can consult?  The [deployment guide for S3](https://nuxtjs.org/faq/deployment-aws-s3-cloudfront) only talks about using `nuxt generate`.

Thanks!

loonpwn:
@Sford4 

Is using the universal build mode with the configured routes, not an option? My use case is that using universal deployment (which server-side renders static HTML first) was causing some issues that were too hard to debug and it was increasing build time significantly.

We are now deploying a static s3 build of the site using spa/static mode. The code from this issue was actually deploying in another PR I did and was merged, so you can do the above to hack in `head` into the static HTML when using that spa mode.

The syntax is the same as the above and you would need to map out each individual page in the function, there is no way for it to read the `head` content of the component since it's not actually running the pages. To reiterate, this is a hacky solution and should only be used as a last resort, but it will sort out some SEO issues, our main issue was including the social sharing data.

Sford4:
@loonpwn Thanks for getting back!  I'm more than happy to use universal build mode with the configured routes, is there a guide for deploying this to AWS?  Right now deploying a static build to S3 works smoothly, and costs next to nothing to keep live.  

Can I deploy the universal build to S3, set up for static web hosting, perhaps with the entry point `.nuxt/server/index.ssr.html`?  Pardon my ignorance, If there's documentation I can consult to get this up on S3 please let me know how I might find it!

loonpwn:
@Sford4 It's a bit off-topic from this PR, but I'll try to cover it.

Firstly, you can't run the SSR scripts on the S3 bucket, the S3 web hosting only has support for serving content, not for running it on a server-side level. 

Within your nuxt.config.js you should have your target set to 'static' and your mode set to the default (universal I think).

You should have a deploy script which looks like the following (this is mine with cloudfront support).

```bash
# Install deps
- yarn install --frozen-lockfile --non-interactive
# Generate dist folder
- nuxt generate
# Push Non-HTML assets with cache headers
- aws s3 cp dist s3://$S3_BUCKET --recursive --acl=public-read --include="*" --exclude "*.html" --cache-control max-age=31536000,public
# Push HTML pages with no cache headers
- aws s3 cp dist s3://$S3_BUCKET --recursive --acl=public-read --exclude="*" --include "*.html" --cache-control max-age=0,no-cache,no-store,must-revalidate
# Invalidate caches
- aws cloudfront create-invalidation --distribution-id=$CLOUDFRONT_ID --paths '/*'
```

If you're using cloudfront you set the root element to index.html and handle 404's with the 200.html file. This is probably explained in the nuxt docs somewhere if you dig around enough 

Happy to assist you further over a DM via [Twitter](https://twitter.com/harlan_zw) if you're still stuck.

xueyongg:
Do we have any solution currently for the social sharing data not updated for dynamic routes for an AWS S3 deployment approach?

PR #6922, opened 2020-01-27
fix(vue-app): set scrollRestoration when there is no location hash

### LINKED ISSUE #6921: Scroll position on initial page load does not go to hash/anchor position
### Version

[v2.11.0](https://github.com/nuxt.js/releases/tag/v2.11.0)

### Reproduction link

[https://codesandbox.io/s/amazing-euler-wtyp7](https://codesandbox.io/s/amazing-euler-wtyp7)

### Steps to reproduce

• Go to codesandbox URL
• Click the `About (scrolled)` button.
• This works as you would expect, scrolling to the correct element.

• Open a new tab/window and go to that URL directly: https://wtyp7.sse.codesandbox.io/about#scroll
• This does not work as expected, the scroll position is at the top of the page.

(note: the behavior when refreshing a page is _different_ than going to it directly. this can make it confusing when trying to recreate the issue, so make sure you use a new tab/window each time.)

### What is expected ?

Initial scroll position is at the correct element.

Initial scroll position is at the top of the page.

### Additional comments?

The default scrollBehavior function handles scrolling to a hash position, however this function does not run on initial page load. Since `window.history.scrollRestoration` is being turned to `'manual'` in the router config, browsers will not automatically scroll to the desired position.

Looking through some old PRs the code change appears to have been made to fix scroll issues with route transitions. A later change was made to retain scroll position when leaving a page, which allows page reload and forward/back navigation to use the previous scroll position.

We could move the code that first sets `scrollRestoration` inside of a conditional which looks at the location hash, since we set it to `'manual'` again again after page load.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10240">#c10240</a>)</em></sub></div>

casey6:
## Types of changes

## Description
Move `window.history.scrollRestoration` inside a conditional so that it only applies when there is no hash in the URL. This allows browsers to handle the initial scroll position when navigating directly to these URLs.

Resolves #6921

## Checklist:


clarkdo (commented):
I think the root cause is that `scrollBehavior`  is not handled in client from ssr (direct access via url), we'd be better to make another appropriate fix.

@Atinux Any thought ?

Atinux:
Thank you for the PR @casey6 

@clarkdo you mean we should find a way to call `scrollBehaviour` on client-side before mounting the app?

Would like to have @posva thoughts on it!

posva:
I agree with @clarkdo. Also, not setting the scrollRestoration to manual only because there is a `hash` would break the scroll for applications using the fragment section of the url to store some information as well as other cases with an empty hash `#` to avoid some urls like the ones ending with `.` not being completely considered as URLs in many applications (like Discord)

This might be a bit difficult to add to current vue router, although still possible. It should be quite easy with the next version

clarkdo:
@posva Thanks, look forward to next version

galvez:
Adding this here for ref: https://github.com/vuejs/rfcs/pull/176

posva:
This shouldn't be necessary anymore since vue-router does set the restoration to manual if scrollBehavior is supported and provided (in v3) (https://github.com/vuejs/vue-router/pull/1814)

PR #6936, opened 2020-02-01
feat: Add csp hashes for non component css

larspederamlie:
Non component css gets hashes calculated and added to the csp if a style-src has been specified.

## Types of changes

## Description

Currently we need to add unsafe-inline to the style-src in the csp.  We would like to be able to turn off unsafe-inline for security reasons.  

## Checklist:


lpamlie:
Easier to use the extractcss option

PR #6939, opened 2020-02-03
chore(vue-app): pass error as param to nuxt-loading fail method

ghost:
pass error as param to nuxt-loading.fail method.

## Description

When users switch tabs quickly, the axios request will be canceled. The loading bar will become red (fail). But in our case, we don't want to display red color for canceled request. Therefore, we need the `error` param to check if the error is canceled request or not for our own custom loading.

## Checklist:

here is the pr of docs repos: https://github.com/nuxt/docs/pull/1828

### DIFF
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -178,7 +178,7 @@ export default {
       try {
         await Promise.all(promises)
       } catch (error) {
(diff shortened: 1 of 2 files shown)


pi0 (approved):

pi0:
Thanks!

PR #6803, opened 2019-12-17
fix(vue-app): resolve leave transition on child routes

atinux:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Fixes https://github.com/nuxt/nuxt.js/issues/6788

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #6946, opened 2020-02-07
fix(vue-app): use child transition name when navigating to parent

reegodev:
## Types of changes

## Description

After #6803 fixed the issue i reported about child transition properties, i discovered that also the transition name did not work correctly when leaving a child route to its parent.

I've created a quick demo to show the problem: https://mllwe.sse.codesandbox.io/
If you enter a child route from its parent (ie: B -> BB ) you can see the correct transition applied ( rotation + fade-in ).
If you leave the child route and return to its parent ( ie: BB -> B ), instead of using the child route transition ( rotation + fade-out ), you see the default page transition applied ( swipe to the right ).
This is particularly bad when you use child routes to open modals, as the opening animation works correctly while the leave animation does not.

This PR provides a fix for this scenario, along with some E2E tests to assure that page transitions name and callbacks merging do not break in the future.

This is the same demo with the fix applied: https://ghdkd.sse.codesandbox.io/

## Checklist:


packages/vue-app/template/client.js
```diff
@@ -102,31 +102,29 @@ function componentOption (component, key, ...args) {
...
+
+  const mergedTransitions = []
+  for (let i=0; i<maxDepth; i++) {
+    // Clone original objects to prevent overrides
+    const toTransitions = Object.assign({}, componentTransitions(toComponents[i]))
+    const transitions = Object.assign({}, componentTransitions(fromComponents[i]))
+    
+    // Combine transitions & prefer `enter` callbacks and name of "to" route
+    Object.keys(toTransitions)
+        .filter(key => typeof toTransitions[key] !== 'undefined' && !key.toLowerCase().includes('leave'))
```
matteo-rigon: Reversed the logic so that we start from the "from" properties and merge only existing "to" properties, ignoring the leave properties as before. This allows to keep the transition name from the "from" component in case there is no "to" component at the same depth ( as it happens when going from child to parent ).

Also just checking for truthiness of `toTransitions[key]` was preventing falsy values from being merged, ie: `css: false`, so `typeof toTransitions[key] !== 'undefined'` fixes that as well.

atinux (approved):

PR #6953, opened 2020-02-09
fix: eliminate side effect from modern utils

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/utils/src/modern.js
+++ b/packages/utils/src/modern.js
@@ -1,6 +1,3 @@
-import UAParser from 'ua-parser-js'
-import semver from 'semver'
-
 export const ModernBrowsers = {
   Edge: '16',
   Firefox: '60',
@@ -15,22 +12,29 @@ export const ModernBrowsers = {
   'Mobile Safari': '10.3'
 }
 
-const modernBrowsers = Object.keys(ModernBrowsers)
-  .reduce((allBrowsers, browser) => {
-    allBrowsers[browser] = semver.coerce(ModernBrowsers[browser])
-    return allBrowsers
-  }, {})
+const modernBrowsers = new Proxy(ModernBrowsers, {
+  get (browsers, prop) {
+    if (browsers[prop] && !browsers[prop].version) {
+      const coerce = require('semver/functions/coerce')
+      browsers[prop] = coerce(browsers[prop])
+    }
+    return browsers[prop]
+  }
+})
 
 export const isModernBrowser = (ua) => {
   if (!ua) {
     return false
   }
+  const coerce = require('semver/functions/coerce')
+  const gte = require('semver/functions/gte')
+  const UAParser = require('ua-parser-js')
   const { browser } = UAParser(ua)
-  const browserVersion = semver.coerce(browser.version)
+  const browserVersion = coerce(browser.version)
   if (!browserVersion) {
     return false
   }
-  return Boolean(modernBrowsers[browser.name] && semver.gte(browserVersion, modernBrowsers[browser.name]))
+  return Boolean(modernBrowsers[browser.name] && gte(browserVersion, modernBrowsers[browser.name]))
 }
 
 export const isModernRequest = (req, modernMode = false) => {


packages/utils/src/modern.js
```diff
@@ -15,22 +12,29 @@ export const ModernBrowsers = {
   'Mobile Safari': '10.3'
 }
 
-const modernBrowsers = Object.keys(ModernBrowsers)
-  .reduce((allBrowsers, browser) => {
-    allBrowsers[browser] = semver.coerce(ModernBrowsers[browser])
-    return allBrowsers
-  }, {})
+const modernBrowsers = new Proxy(ModernBrowsers, {
```
pi0: Still, we are always initializing a Proxy. What if we instead use `getModernBrowsers` which in the first call evaluates the value and keep in `_modernBrowsers` for next calls?

packages/utils/src/modern.js
```diff
@@ -15,22 +12,29 @@ export const ModernBrowsers = {
...
+    }
+    return browsers[prop]
+  }
+})
 
 export const isModernBrowser = (ua) => {
   if (!ua) {
     return false
   }
+  const coerce = require('semver/functions/coerce')
```
pi0: Is it ok require called per-request? (TBH no idea of perf impact when we have cache)

packages/utils/src/modern.js
clarkdo: Proxy get won't be called at initialization, it's same as using _modernBrowsers

packages/utils/src/modern.js
clarkdo: This code is called in nodejs and I think require.cache is pretty fast

packages/utils/src/modern.js
pimlie: Would also prefer using memoization here. Seems cleaner then using a Proxy

pi0: Makes sense :)

pi0: I also like Proxy self-memo pattern :)

pi0 (approved):

packages/utils/src/modern.js
pi0: The only thing that should check later is that ensure rollup doesn't count this as a side-effect and tree-shaking works for the const variable.

clarkdo: Actually my first edition is using a local state `__modernBrowsers` and function `getModernBrowsers`, but I feel having a `__modernBrowsers` in scope   
 is still kind of side effect, and proxy can have shorter code and fine-grained control.

pi0: The difference is in `__modernBrowsers` method, side-effect is only an empty object rather than a proxy that also has getter logic itself and less low-level. But bundle size is not really a concern so i'm ok with trying new Proxy approach :)

clarkdo: I checked the built dist/utils.js, it works fine.
@pimlie Maybe you can check with your usage ?

pi0: Mergeed with [release] to beta test :)

pimlie: @clarkdo Doesnt seem to work:
```
 ERROR  Webpack build for lambda failed [                                                                                 11:56:45
  '../node_modules/@nuxt/utils-edge/dist/utils.js\n' +
    "Module not found: Error: Can't resolve 'semver/functions/coerce' in 'nuxt-lambda/node_modules/@nuxt/utils-edge/dist'",
  '../node_modules/@nuxt/utils-edge/dist/utils.js\n' +
    "Module not found: Error: Can't resolve 'semver/functions/gte' in 'nuxt-lambda/node_modules/@nuxt/utils-edge/dist'"
]
```

pimlie: @clarkdo This is because I had still stubbed semver btw, let me check what happens if i dont stub it

pimlie: When I dont stub semver I get a runtime error, have to check it. But nuxt/utils load is as follows:

image

Total time is 8.2ms for loading/parsing @nuxt/utils-edge. The webpack require is signal-exit, the small yellow block is `escapeRegExp` and the other block already shows the function.

clarkdo: From log, the error is from `isModernBrowser`, `semver` is required at hat time for browser detection.

clarkdo: @pimlie I've applied the changes from your and @pi0 review, but it's not released to edge

PR #6954, opened 2020-02-09
refactor: use scope variable memo for modern utils

clarkdo:
I'm not aware that the original pr https://github.com/nuxt/nuxt.js/pull/6953 has been merge.

Address review from @pi0 and @pimlie 
- Use local variable to implement memo
- Avoid calling require for every request

### DIFF
--- a/packages/utils/src/modern.js
+++ b/packages/utils/src/modern.js
@@ -1,6 +1,3 @@
-import UAParser from 'ua-parser-js'
-import semver from 'semver'
-
 export const ModernBrowsers = {
   Edge: '16',
   Firefox: '60',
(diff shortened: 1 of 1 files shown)


pi0 (approved):
BTW I agreed about both cases in the previous PR which was more cleaner. Merge as you wish :)

pimlie:
@clarkdo Functionality wise this is working, but all the refactoring doesnt help with the fact that I just dont want any semver code to be included in my lambda build. These changes still includes the semver code in my lambda build so that code still needs to be parsed (ie its just not executed atm). So at the moment I can still only achieve that by stubbing the semver package.

clarkdo:
As I mentioned in previous pr, semver will be executed when you’re using modern mode.

In server side, we won’t bundle node modules into server dist, so if you don’t even want include semver package, I think you can add it into exclude in your lambda building config.

PR #6959, opened 2020-02-11
fix(core): Provide the correct name of routeContext hook

husayt:
New name of hook is `vue-renderer:ssr:context`

## Types of changes

- [x ] Bug fix (a non-breaking change which fixes an issue)

## Description

This corrects deprecation message and also links to correct documentation (PR: 1832)
No tests are required

## Checklist:

- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly. (PR: 1832)

### DIFF
--- a/packages/core/src/nuxt.js
+++ b/packages/core/src/nuxt.js
@@ -26,7 +26,7 @@ export default class Nuxt extends Hookable {
     // Deprecated hooks
     this.deprecateHooks({
       'render:context': 'render:routeContext',
-      'render:routeContext': 'vue-renderer:afterRender',
+      'render:routeContext': 'vue-renderer:ssr:context',
       showReady: 'webpack:done' // Workaround to deprecate showReady
     })
 
--- a/packages/core/test/nuxt.test.js
+++ b/packages/core/test/nuxt.test.js
@@ -38,7 +38,7 @@ describe('core: nuxt', () => {
 
     expect(nuxt._deprecatedHooks).toEqual({
       'render:context': 'render:routeContext',
-      'render:routeContext': 'vue-renderer:afterRender',
+      'render:routeContext': 'vue-renderer:ssr:context',
       showReady: 'webpack:done'
     })
 


Atinux:
Thanks for the PR @husayt 

Could you fix the failing test as well? :)

husayt:
@Atinux thanks for pointing out. Seems ready for merge now

atinux (approved):
LGTM, waiting for @pi0's review as well :)

clarkdo (approved):

packages/core/src/nuxt.js
```diff
@@ -26,7 +26,7 @@ export default class Nuxt extends Hookable {
     // Deprecated hooks
     this.deprecateHooks({
       'render:context': 'render:routeContext',
-      'render:routeContext': 'vue-renderer:afterRender',
+      'render:routeContext': 'vue-renderer:ssr:context',
```
pi0: These hooks are not identical by parameters! `render:routeContext` accepts `renderContext.nuxt` while `vue-renderer:ssr:context` accepts `renderContext` ([src](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-renderer/src/renderers/ssr.js) 

Ref in v2.4.0: https://github.com/nuxt/nuxt.js/blob/v2.4.0/packages/vue-renderer/src/renderer.js#L373

clarkdo: But I don't see there is a `vue-renderer:afterRender`

pi0: Indeed. This is why it is kept for next major: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-renderer/src/renderers/ssr.js#L94

clarkdo: I mean the error message is `render:routeContext hook has been deprecated, please use vue-renderer:afterRender`, but there is no `vue-renderer:afterRender`, users who migrate to new hook won't get hook worked.

clarkdo: If only diff is param, we need to update doc.

pi0: There are 3 problems:
- `vue-renderer:afterRender` does not exists
- Recursive deprecate not being supported
- deprecated warning is not customizable (for example to mention param changes)

Working on it will make a PR :)

husayt: @pi0 regarding params indeed it accepts different param. I have mentioned that in the [docs pull request](https://github.com/nuxt/docs/pull/1832/files). Since this is merged it makes sense to merge that pull request too.

PR #6967, opened 2020-02-13
chore: migrate to memfs

jamesgeorge007:
## Types of changes

## Description
`memory-fs` is deprecated and not maintained anymore. It is advised to migrate to `memfs`.

## Checklist:


pi0:
Thanks for PR @jamesgeorge007 👍

clarkdo:
Thank you so much for the pr 😸 , several points need to be considered:

- We're delegating memory-fs api from https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/utils/async-mfs.js , it can be removed as `memfs` supports `Promises` already
- I've seen your pr in webpack, great work! Will it be merged into 4.x ? https://github.com/webpack/webpack/pull/10378 , if not, we should merge this into webpack5 branch
- webpack-dev-middleware is still using `memory-fs` as default option. https://github.com/webpack/webpack-dev-middleware/blob/master/lib/utils/setupOutputFileSystem.js

jamesgeorge007:
- https://github.com/webpack/webpack/pull/10378 will be landed on `master`.
- Regarding `webpack-dev-middleware`, active works are happening on the `dev` branch where in which the migration had already happened (`memory-fs` -> `memfs`).

pi0:
Some points:

- It seems the current webpack `master` is releasing to `5.x` (right?)
- Nuxt provides [compiler.outputFileSystem](https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/builder.js#L86) for webpack dev middleware
- This filesystem is being shared with server renderer (dev and prod) and `AsyncMFS` is providing a similar interface to [fs-extra](https://www.npmjs.com/package/fs-extra) for async functions so we need to keep interface or do a refactor to `fs.promises`. Example [usage](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-renderer/src/renderer.js#L96)
- `fs.promises` is supported by Node 10+ (while current version is 8.x)
- Webpack 4 depends on `outputFileSystem.join` which has it's own [join](https://github.com/webpack/memory-fs/blob/master/lib/join.js) implementation

galvez:
Just to be clear, what's missing to merge this? Raising the minimum version to Node 10?

manniL:
> Just to be clear, what's missing to merge this? Raising the minimum version to Node 10?

Which would be a breaking change.

clarkdo:
> Which would be a breaking change.

Yes, that's right, so we can work on `next` branch (v3 release) for avoiding breaking change

pi0:
@manniL @galvez There is nothing wrong with `memfs` for node8 support :)

`fs.promises` I mentioned was a replacement for `fs-extra` as we need a similar interface for using in production and development loading bundle. (promises.exists is not available). AsyncFS is just a polyfill to make them similar... 

Also re landing to master, latest 4.x release of webpack (4.43.0) is still using `memory-fs`.

So agree with @clarkdo would be better doing the upgrade for `next` branch after merging #4616  unless webpack 4 eventually also switches to memfs which we can make backport/create compatible AsyncFS for Nuxt2.

pi0:
Migrated to memfs for webpack5 (#4616) branch (https://github.com/nuxt/nuxt.js/pull/4616/commits/3d6fa0d90c0d2a7e52e7608a0e4a2ec1483971a6).

While this change can be backported to nuxt@2, as wepback@4 (currently 4.43.0) and also webpack-dev-middleware still depending on `memory-fs` and there are no known issues with it (other than being deprecated), I think would be better for stability to keep using it unless either weback@4 migrates or we find a bug. So closing PR.

@jamesgeorge007 Thanks for PR and bringing it to top and contribs to webpack :heart:

PR #6987, opened 2020-02-19
feat(wepack): support `@import` using `~assets` in css files

clarkdo:
## Types of changes

## Description

From Nuxt 2.0 with css-loader upgradation, the `~/` alias can't be resolved correctly in CSS files, so we've updated doc that recommending using ~assets (without a slash) , but @import css is still only working with `~/assets`, this makes @import and normal css url  have different syntax, so this pr is adding `~assets` to postcss-import alias.

In further consideration, we may enhance postcss plugin to support `~${anyPath}` and `@${anyPath}`.

## Checklist:


packages/webpack/src/utils/postcss.js
```diff
@@ -45,15 +45,31 @@ export default class PostcssConfig {
     return this.buildContext.buildOptions.postcss
   }
 
+  get postcssImportAlias () {
+    const alias = { ...this.buildContext.options.alias }
+
+    for (const key in alias) {
+      if ((key.startsWith('~'))) {
```
manniL:
```suggestion
      if (key.startsWith('~')) {
```

clarkdo:
I think we should not support @... in @import as css-loader only supports ~... in url()

pi0 (dismissed):

PR #5425, opened 2019-03-30
feat(webpack): Expose webpack base config class

juliendargelos:
Expose webpack base config class.

## Types of changes
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Add an export of the [`WebpackBaseConfig`](https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/config/base.js) class.

It could be useful for some advanced modules to be able to extend this class for independent webpack build using Nuxt base config:
I'm making a module which hooks Nuxt to add another webpack build with NetlifyCMS, but sharing some source files with Nuxt project (for previews). I would like to have the same base configuration so Nuxt project files can be imported with the same rules, plugins, etc...

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

Issue #4286, opened 2018-11-07
Expose resolved webpack config for external tool


amoshydra:
### What problem does this feature solve?

I am trying to add unit test with ava in this project. However, webpack alias is not understood by ava. This problem with ava could be resolved by providing a webpack config.

As of now, we need to create the webpack config by hand and try to match it with what nuxt is using.

If a resolved webpack config is produced and exposed, we could use this config with our testing, linting and other external tools.

Related to:
- #200
- #2936
- #2170

### What does the proposed changes look like?

Perhaps, it could look something similar to vue's cli-service implementation. Where the resolved webpack.config is exposed as a file.

https://cli.vuejs.org/guide/webpack.html#inspecting-the-project-s-webpack-config

For testing use case, perhaps:
1. Developer call a function provided by nuxt to trigger the resolvation of webpack.config
2. Nuxt expose the resolved webpack.config as a file
3. Developer run test script which rely on the exposed webpack.config

For linting use case,
1. Perhaps a base webpack.config can always be exposed as a file at the begining
2. When developer call `nuxt dev`, a new resolved webpack.config can override the existing webpack.config
3. IDE / Editor's eslint tool could automatically pick up this change and update its rule accordingly?

In summary,
1. Expose a base webpack.config
2. Resolve and update webpack.config as nuxt is building or via a function call by developer
3. Developer / external tool make use of this resolved config

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8098">#c8098</a>)</em></sub></div>

amoshydra:
This issue appear to be semi hackable by making use of nuxt hook and its api.

Let say we are making a build for SSR that will output webpack config for both client and server, this could be done.

## How
Write a script to collect the compiler options via the `build:compile` hook. Once we have collected enough options (we need 2 options in this case), we will call `resolve()` and return all the collected configurations back to caller.

**FILE:** `./scripts/collect-webpack-configs.js`
```js
import { Nuxt, Builder } from 'nuxt';
import config from '../nuxt.config.js';

const stopProcessIfCalledTimes = (times, callback) => {
  let counter = times;
  const compilerOptionMap = {};

  return ({ name, compiler }) => {
    compilerOptionMap[name] = compiler;
    counter--;
    if (counter === 0) {
      callback(compilerOptionMap);
    }
  };
};

[...]

PR #7029, opened 2020-02-28
feat: expose webpack config

pi0:
## Types of changes

## Description

🐑ing a long-time wanted feature!

### `nuxt webpack [query...]`

**Options:**

- `--name`: Bundle name to inspect. (client, server, modern)
- `--dev`: Inspect webpack config for dev mode
- `--depth`: Inspection depth. Defaults to 2 to prevent verbose output.
- `--no-colors`: Disable ANSI colors (disabled by default when TTY is not available or when piping to a file)

Inspired by [`vue inspect`](https://cli.vuejs.org/guide/webpack.html#inspecting-the-project-s-webpack-config) a convenient way to quickly inspect webpack configuration. Some examples:

- `nuxt webpack`
- `nuxt webpack devtool`
- `nuxt webpack resolve alias`
- `nuxt webpack module rules`
- `nuxt webpack module rules test=.jsx`
- `nuxt webpack module rules test=.pug oneOf use.0=raw`
- `nuxt webpack plugins constructor.name=WebpackBar options reporter`
- `nuxt webpack module rules loader=vue-`
- `nuxt webpack module rules "loader=.*-loader"`

DX: Tip to inspect more in-depth and auto-fallback to the upper-level match:

image

#### Why not `nuxt inspect`?

- Reserve namespace for future improvements like inspecting more internals other than webpack
- Ensure we can easily keep backward compatibility for cli while adding new possibilities

### `node_modules/nuxt/webpack.config.js`

Exporting default webpack config for Client and `rootDir` of `cwd()` using `getWebpackConfig`. If `nuxt.config.js` is in current working dir (ide project root for example) it will be also correctly loaded. Some external tools may need access to the resolved webpack config as a file, for example, IDEs or command-line tools that expect a webpack config path.

### `getWebpackConfig(name, options)`

```js
import { getWebpackConfig } from 'nuxt'

getWebpackConfig()
getWebpackConfig('server', { rootDir: '...' })
```

* `options` type is `loadOptions` used for new programatic  #6868 interface

This function returns a Promise which loads project config, creates an instance of nuxt internals, runs modules for extending webpack config with the final result of exactly what Nuxt internally using! Webpack supports [Exporting a Promise](https://webpack.js.org/configuration/configuration-types/#exporting-a-promise) so it is also a valid webpack config.

### History

- #200
- #4286
- #5366
- #5425
- #5378

## Checklist:


packages/cli/src/commands/webpack.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const [l, r] = query.split('=')
+
+  if (Array.isArray(result)) {
+    result = result.filter((i) => {
+      const v = get(i, l)
+      if (v === r) {
+        return true
+      }
+      if (v) {
```
pimlie: early return when `!v`?

packages/cli/src/commands/webpack.js
```diff
@@ -0,0 +1,92 @@
...
+          return true
+        }
+        if (r.match(v)) {
+          return true
+        }
+      }
+    })
+
+    if (result.length === 1) {
+      result = result[0]
```
pimlie: why reassign? return immediately should be ok too?

packages/cli/src/commands/webpack.js
```diff
@@ -0,0 +1,92 @@
...
+function advancedGet (obj = {}, path = '', query = '') {
+  let result = obj
+
+  if (!path) {
+    return result
+  }
+
+  result = get(obj, path)
+
+  if (!query || result === undefined) {
```
pimlie: is checking for undefined required? couldnt we return on any falsy value ?

packages/cli/src/commands/webpack.js
```diff
@@ -0,0 +1,92 @@
...
+export default {
+  name: 'webpack',
+  description: 'Inspect webpack config',
+  usage: 'webpack [path] [prop=value]',
+  options: {
+    ...common,
+    name: {
+      alias: 'n',
+      type: 'string',
+      default: 'Client',
```
pimlie: Could we make this case insensitive? I rly dont want to have to remember to type Client instead of client

packages/cli/src/commands/webpack.js
```diff
@@ -0,0 +1,92 @@
...
+    const config = await cmd.getNuxtConfig({ dev: false, server: false, _build: true })
+    const nuxt = await cmd.getNuxt(config)
+    const builder = await cmd.getBuilder(nuxt)
+    const { bundleBuilder } = builder
+    const webpackConfig = bundleBuilder.getWebpackConfig(name)
+    const match = advancedGet(webpackConfig, path, query)
+
+    if (match === undefined) {
+      throw new Error(`No match in webpack config for path:${path}${query ? ` query:${query}` : ''}. Try inspecting without query and path.`)
+    } else {
```
pimlie: else is redundant when throwing

galvez:
Great use of Nuxt commands! :)

galvez (approved):

atinux (approved):

f3ltron:
Is it any doc for it ? I had to in the core code to check exactly i can use to get dev client config ^^.

{for: 'dev'}

Thank's for your work anyway that help a lot !

pi0:
@f3ltron If you have time to contribute to docs it would be more than welcome :hugs:

f3ltron:
i can to add some yeah. I just don't know where to put it ^^

pi0:
@manniL May give some advice for best place to add :)

PR #7071, opened 2020-03-10
refactor(babel-preset-app):  use `@babel/helper-compilation-targets` instead of deprecated api

clarkdo:
## Types of changes

## Description

Per https://github.com/babel/babel/blob/v7.8.7/packages/babel-preset-env/src/index.js#L33 and https://github.com/babel/babel/blob/v7.8.7/packages/babel-preset-env/src/targets-parser.js#L1, some apis are gonna be removed, we can use the new ones now.

## Checklist:


packages/babel-preset-app/src/index.js
```diff
@@ -32,15 +32,20 @@ function getDefaultPolyfills (corejs) {
 }
 
 function getPolyfills (targets, includes, { ignoreBrowserslistConfig, configPath, corejs }) {
-  const { isPluginRequired } = require('@babel/preset-env')
+  const { default: getTargets, isRequired } = require('@babel/helper-compilation-targets')
```
pi0: Should we add direct `@babel/helper-compilation-targets` dependency?

clarkdo: I think it's find, `@babel/helper-compilation-targets` is also used by `@babel/preset-env` and other packages

pi0: As we directly require this module I think we need to define in package.json. At least it is a requirement of yarn v2 and similar new PMs.

clarkdo: I'll add it.

clarkdo: @pi0 Thanks

PR #7108, opened 2020-03-19
fix(fetch() api): fix multiple fetch() call on same component

### LINKED ISSUE #7107: new fetch() api called multiple times
### Version

[v2.12.0](https://github.com/nuxt.js/releases/tag/v2.12.0)

### Reproduction link

[https://codesandbox.io/s/solitary-rain-2t47v](https://codesandbox.io/s/solitary-rain-2t47v)

### Steps to reproduce

I can't reproduce on codesandbox and the cose is exactly the code of the example. But on a spa mode.
My problem is that, in the first page, i do a fetch calling a remote api with axios. From chrome console I see only one call and all is ok.
Now I navigate to another page, then, I came back to first page and I expect one call to fetch to reload data. But what I see from chrome console is that the fetch method is called twice! now, If i navigate another time to another page and then i came back to first page, the fetch is called three times!! and so on. Is like the fetch is put in an array of calls that increments the call at each navigation and then is run n-times. 
I cannot recreate this problem.. but is a big problem. Keep in mind that, if I put the same call in a monted() event, the problem disappear. Only in asynch fetch() there is this problem! please help me!
nb. If i put a "debug" instruction on top of page like 
<p>Fetching: {{ $nuxt.isFetching }} ({{ $nuxt.nbFetching }})</p>
I see, each time I came back to page, the nbFetching value increased. And it decrease to zero.
each time, nbFetching is one more big.

One possible solution to this problem is to modify fetch.client.js in this way
async function $fetch() {
+  if (this.$fetchState.pending) return 

in this way, all works as expected!

### What is expected ?

On each navigation, fetch is called only one time

fetch is called n-times where n is the number of times the page was navigated to

### Additional comments?

My configuration is nuxt 2.12 with axios module and vuetify. No particular configuration. I've also tried to delete .nuxt folder and rebuild but none works. 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10414">#c10414</a>)</em></sub></div>

asterd:
Closes #7107

## Types of changes

- [x ] Bug fix (a non-breaking change which fixes an issue)

## Description

Resolves: #7107. Just add a return on fetch already pending to prevent multiple calls 

## Checklist:


asterd:
In addition, as suggested from @olddeda we can solve the issue also in this way

// file vue-app/template/utils.js
export function addLifecycleHook(vm, hook, fn) {
  if (!vm.$options[hook]) {
    vm.$options[hook] = []
  }
-  vm.$options[hook].push(fn)
+ if (!vm.$options[hook].includes(fn)) {
+     vm.$options[hook].push(fn)
+ }
}


If you need this patch, I can make another pull request.

Atinux:
The hook option seems better indeed @asterd 

Happy that you make another PR :)

pi0:
Thanks, @asterd for all helps. I've added both proposed suggestions by #7129 (credits preserved of course :heart: )

PR #7144, opened 2020-03-27
feat(babel-preset-app): add `bugfixes` option

clarkdo:
## Types of changes

## Description

Related to https://github.com/nuxt/nuxt.js/issues/6784#issuecomment-605042333

Add new option `bugfixes` which is introduced from babel-preeset-env 7.9.0.

Options can be set by:

```js
export default {
  build: {
    babel: {
      presets: ['@nuxt/babel-preset-app', { bugfixes: true }]
  }
}
```

## Checklist:

### DIFF
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -55,6 +55,7 @@ module.exports = (api, options = {}) => {
   const envName = api.env()
 
   const {
+    bugfixes,
     polyfills: userPolyfills,
     loose = false,
     debug = false,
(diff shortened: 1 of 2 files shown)


packages/babel-preset-app/src/index.js
```diff
@@ -55,6 +55,7 @@ module.exports = (api, options = {}) => {
   const envName = api.env()
 
   const {
+    bugfixes,
```
pi0: What do you think of setting explicit default value here?

clarkdo: I think keeping same default value as babel make it default true until babel 8 is proper.

clarkdo: I can add false to it. And change to true when upgrade to babel 8

pi0: Both LGTM. I let you decide state to merge :)

pi0 (approved):

PR #7267, opened 2020-04-24
fix(core): improve module not found error 

gion:
Update: https://github.com/nuxt/nuxt.js/pull/7267#issuecomment-634946919

## Types of changes

## Description

This PR adds logs the original error to the output in order to aid debugging.

In my case, I was chasing a __FATAL__ _"Module x not found"_ error and it was really hard to debug, since the underlying cause was stripped from the logs.  

Before this commit:
Screenshot 2020-04-24 at 13 31 43

After this commit:
Screenshot 2020-04-24 at 13 38 39

In my particular case, this kind of message would've saved me at least an hour of debugging ⏳.

## Checklist:

### DIFF
--- a/packages/core/src/module.js
+++ b/packages/core/src/module.js
@@ -159,6 +159,9 @@ export default class ModuleContainer {
         }
 
         message = message.replace(/{name}/g, src)
+        
+        // add this extra warning that points to the original error for debuging purposes
+        consola.warn('originalError:', error)
 
         if (this.options._cli) {
           throw new Error(message)
--- a/packages/core/src/module.js
+++ b/packages/core/src/module.js
@@ -150,25 +150,26 @@ export default class ModuleContainer {
           throw error
         }
 
-        let message = 'Module `{name}` not found.'
+        // Hint only if entrypoint is not found and src is not local alias or path
+        if (error.message.includes(src) && !/^[~.]|^@\//.test(src)) {
+          let message = 'Module `{name}` not found.'
 
-        if (this.options.buildModules.includes(src)) {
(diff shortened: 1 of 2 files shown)


packages/core/src/module.js
```diff
@@ -160,6 +160,9 @@ export default class ModuleContainer {
 
         message = message.replace(/{name}/g, src)
 
+        // add this extra warning that points to the original error for debuging purposes
+        consola.warn('originalError:', error)
+
         if (this.options._cli) {
           throw new Error(message)
```
pi0, line 167: Nice improvement! Maybe here instead we can throw original error (stack + message) but only refine error.message to add additional hints?

gion: I didn't dive in too deep to see what conventions do you use or how are you guys handling errors in general, so I guess, there is a lot of room for improvement :). 
Also, I think that there are other cases as well where the original error is shadowed and would be nice to be exposed.

If you feel that this small improvement needs further discussion/refactoring, maybe let's create an issue and move this fix there.  

Meanwhile, it's up to you if this fix can be merged.

Cheers ✌️

pi0: Indeed. Thanks for bringing it to top. I will push changes to this PR if is okay ☺️

gion: Sounds awesome! Thanks!

pi0:
**Update:**

Non-exsistent module (without alias or local import):

image

If `foo.js` trys to import `bar.js` which is not existent:

image

pi0 (approved):

gion:
awesome 🥳

PR #7292, opened 2020-04-29
feat: Progress indication when generating routes

simplenotezy:
Adds a progress indication to the console when generating routes (e.g. `(2/3) Generated /about-us`)

nuxt generate progress

## Types of changes

## Description
As described in this issue: https://github.com/nuxt/nuxt.js/issues/7038 it would be handy with a little progress indication when generating routes.

I generate around 3000 routes for an e-commerce site, and it takes anywhere between 600-3000 seconds (currently investigating why) and it would be quite helpful for my impatience to have some kind of idea where in the process we are.

This pull request will add a simple indication when generating routes, like so:

    (1/3) Generated /home
    (2/3) Generate /about-us
    (3/3) Generated /contact

I would have loved to implement it using https://github.com/nuxt/webpackbar, however, lack of knowledge of how to implement it made me refrain from doing it (the webpackbar could use a little improvement in the documentation ;))

## Checklist:

Not sure how to run the testsuite yet. Doubt this change would have an effect.

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -1,92 +1,80 @@
-/*!
- * @nuxt/generator v2.12.2 (c) 2016-2020
+import path from 'path'
+import Chalk from 'chalk'
+import consola from 'consola'
+import fsExtra from 'fs-extra'
+import htmlMinifier from 'html-minifier'
 
- * - All the amazing contributors
- * Released under the MIT License.
- * Website: https://nuxtjs.org
-*/
-'use strict';
+import { flatRoutes, isString, isUrl, promisifyRoute, waitFor } from '@nuxt/utils'
 
-Object.defineProperty(exports, '__esModule', { value: true });
-
-function _interopDefault (ex) { return (ex && (typeof ex === 'object') && 'default' in ex) ? ex['default'] : ex; }
-
-const path = _interopDefault(require('path'));
-const Chalk = _interopDefault(require('chalk'));
-const consola = _interopDefault(require('consola'));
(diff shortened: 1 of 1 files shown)


simplenotezy:
hey @lihbr - thanks for your feedback and great ideas. How could this be turned into a module?

simplenotezy:
Also I think you already get this "spam" when deploying, just without the `(2/5)`-part

lihbr:
Oof my bad indeed, don't know why I was convinced that Nuxt was not displaying generated routes 🤔 Really sorry to have you puzzled on that!

As for the module you could have come up with something like the following but it doesn't make sense anymore ^^'
```javascript
const consola = require("consola");

module.exports = function () {
  this.nuxt.hook("generate:extendRoutes", routes => {
    const routesTotal = routes.length;
    let currentRoute = 0;

    this.nuxt.hook("generate:routeCreated", ({ route, errors }) => {
      if (!errors.length) {
        consola.success(`(${++currentRoute}/${routesTotal}) Generated ${route}`);
      }
    });
  });
};
```

pi0:
Hi! Maybe we can try something like [ora](https://github.com/sindresorhus/ora)? It might be good to also honor nuxt `quite` option to make CI/CD silent and also `std-env`'s minimal to fall back to plain logs as is now.

pi0:
Hi! As this PR is stalled and there might be different opinions on how to display progress indicator, postponing this enhancement for Nuxt3 but for sure will be improved. Thanks for contrib @simplenotezy ❤️

PR #7295, opened 2020-04-30
feat: transparently add Plug'n'Play support

### LINKED ISSUE #7142: Compatibility with Yarn 2 Plug and Play

It will make developer be able to use Yarn 2. I think this worth working on since Yarn 2 may be used more in the future.

Just compiled without any errors.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10447">#c10447</a>)</em></sub></div>

paul-soporan:
## Types of changes

## Description

This PR transparently adds support for the Plug'n'Play installation strategy that is found in Yarn 2.
CC @arcanis.

The changes have been tested on the default template of `create-nuxt-app`.
Result: It works without any errors.
Also, [`@yarnpkg/doctor`](https://github.com/yarnpkg/berry/tree/master/packages/yarnpkg-doctor) reports far fewer errors on the Nuxt monorepo than before. 🙂 (This is a huge step in case Nuxt ever decides to migrate to Yarn 2) 

### List of changes:
- add undeclared dependencies (this includes all kinds of dependencies: regular, dev, peer, optional peer, transitive peer) - this is required, as PnP is a much stricter installation strategy, so errors are thrown for undeclared dependencies
- add [`pnp-webpack-plugin`](https://github.com/arcanis/pnp-webpack-plugin) - this plugin transparently adds support for PnP. It doesn't have any effect if PnP is not used. It is shipped by default inside Webpack starting from Webpack@5. Note: The `buildDir` has to be bound to `@nuxt/babel-preset-app` and to the template (e.g. `@nuxt/vue-app`), as it imports dependencies of those packages
- change resolver implementation to use `createRequire`. This is required, as the resolution has to be done in the package that calls `createRequire`, not in `@nuxt/core`. A polyfill for `createRequire` has been added, as Nuxt still supports Node 8
- make webpack loaders use `require.resolve`. See https://github.com/yarnpkg/berry/tree/master/packages/yarnpkg-doctor#no-unqualified-webpack-config to understand why it's needed

The existing tests have been updated, but no new tests have been added yet.
Note: We will add an E2E test on the [yarnpkg/berry](https://github.com/yarnpkg/berry) repository. Our E2E tests are scheduled and check regularly if something there's a problem with popular repositories like Nuxt: https://github.com/yarnpkg/berry#current-status.
I'm not sure if Nuxt's testing infrastructure supports running tests under different installation strategies. (If it does, we can add a test for Plug'n'Play.)

Resolves #7142.

## Checklist:


paul-soporan:
🤔 `wp.test.js` seems to fail consistently for me locally, even after reverting to the `dev` branch.

distributions/nuxt-start/package.json
```diff
@@ -55,8 +55,12 @@
     "dist"
   ],
   "dependencies": {
+    "@nuxt/builder": "2.12.1",
```
pi0, line 58: `nuxt-start` is a special package for using in production without webpack dependency. I think we should not list it here as require won't even happen for it :)

packages/cli/package.json
```diff
@@ -22,12 +22,22 @@
...
     "wrap-ansi": "^6.2.0"
   },
+  "devDependencies": {
+    "pretty-format": "^25.4.0"
+  },
+  "peerDependencies": {
+    "@nuxt/builder": "*",
+    "@nuxt/core": "*",
+    "@nuxt/generator": "*",
+    "@nuxt/webpack": "*"
```
pi0, line 39: webpack, generator and builder is optional dependencies

packages/core/src/resolver.js
```diff
@@ -10,6 +11,24 @@ import {
   clearRequireCache
 } from '@nuxt/utils'
 
+// Polyfill Node's `Module.createRequire` / `Module.createRequireFromPath` function if not present
+export const createRequire =
```
pi0, line 15: Nice! Just can we maybe add to utils? (or ideally an npm package for polyfill import :p)

packages/core/src/resolver.js
```diff
@@ -27,9 +46,11 @@ export default class Resolver {
     this._resolve = require.resolve
   }
 
-  resolveModule (path) {
+  resolveModule (path, { requirePath } = {}) {
     try {
-      return this._resolve(path, {
+      const pathResolve = requirePath ? createRequire(requirePath).resolve : this._resolve
```
pi0, line 51: What do you think to cache `createRequire` (per `requirePath`)? (maybe doing it directly with `createRequire` util or `getRequire(requiePath)`?)

packages/utils/package.json
```diff
@@ -11,6 +11,7 @@
     "consola": "^2.11.3",
     "fs-extra": "^8.1.0",
     "hash-sum": "^2.0.0",
+    "lodash": "^4.17.15",
```
pi0, line 14: Lodash is inlined by rollup for production release

packages/vue-renderer/package.json
```diff
@@ -12,6 +12,7 @@
     "@nuxt/utils": "2.12.1",
     "consola": "^2.11.3",
     "fs-extra": "^8.1.0",
+    "lodash": "^4.17.15",
```
pi0, line 15: Lodash is inlined

packages/webpack/src/config/base.js
```diff
@@ -19,6 +20,14 @@ import WarningIgnorePlugin from '../plugins/warning-ignore'
 
 import { reservedVueTags } from '../utils/reserved-tags'
 
+function maybeResolve (name) {
```
pi0, line 23: Nice util too! maybeMoveToUtils? :D

packages/webpack/src/config/base.js
```diff
@@ -184,14 +193,31 @@ export default class WebpackBaseConfig {
...
+    } = this.buildContext
+
     return {
       resolve: {
         extensions: ['.wasm', '.mjs', '.js', '.json', '.vue', '.jsx'],
         alias: this.alias(),
-        modules: webpackModulesDir
+        modules: webpackModulesDir,
+        plugins: [
+          PnpWebpackPlugin.bind(buildDir, module, '@nuxt/babel-preset-app'),
```
pi0, line 211: (just out of curiosity) is there any way that we detect Nuxt is running via PnP and conditionally add plugin?

pi0 (commented):
Thanks for works! I've left some minor comments also will try to check locally why tests failing.

It could be nice having basic berry support with Nuxt 2 but Nuxt 3 is coming with WP5 out of the box and refined mono-repo structure which makes yarn happy :D

packages/webpack/src/config/base.js
arcanis: The plugin already works this way, [it's a noop unless PnP is detected](https://github.com/arcanis/pnp-webpack-plugin/blob/master/index.js#L127-L131) 🙂

pi0: Awesome thanks for clearance @arcanis!

packages/webpack/src/config/base.js
paul-soporan: Done. I've also renamed it to `tryResolve` for more consistency. 👍

packages/core/src/resolver.js
paul-soporan: There's currently no standalone npm package for `createRequire`, but there is [a Polyfill](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-core-utils/src/create-require-from-path.ts) in `gatsby-core-utils`, which, funny enough, was added by @arcanis (https://github.com/cstoddart/gatsby/commit/72e0d6fb874cfd2d355265f08ca08460dcde873c#diff-4e10856144ca00b229a8029e440b9e60) so it is the exact same one I used here :D.
We have 3 options:
- duplicate the Polyfill in `@nuxt/utils`
- depend on `gatsby-core-utils`
- maybe @arcanis would like to publish the polyfill as a standalone npm package (What do you think?)

arcanis: Not sure it's worth adding in the Yarn monorepo since it's a quite "static" tool, which won't need updates going forward 🤔

pi0: Adding it as an external package in few minutes :)

packages/core/src/resolver.js
paul-soporan: 🤔 I don't think it would have a noticeable effect on performance. The function with the highest complexity seems to be [`Module._nodeModulePaths`](https://github.com/nodejs/node/blob/c17dcb32533aa007dfbf507d22c28ef3c7c11c29/lib/internal/modules/cjs/loader.js#L1241), but I don't really think it would affect the performance.

packages/core/src/resolver.js
pi0: https://www.npmjs.com/package/create-require

packages/webpack/src/config/base.js
```diff
@@ -251,7 +269,7 @@ export default class WebpackBaseConfig {
     return [
       {
         test: /\.vue$/i,
-        loader: 'vue-loader',
+        loader: require.resolve('vue-loader'),
```
pi0, line 272: Resolving here introduces a potential breaking change for any nuxt module trying to match loader with an exact match.

currently breaking built-in [@nuxt/components](https://github.com/nuxt/components)):
- Usage: https://github.com/nuxt/components/blob/master/src/index.ts#L83 (`find: use => use.loader === 'vue-loader')
- CI: https://github.com/nuxt/nuxt.js/runs/707443791#step:5:36

We may try resolving right before compilation so all hooks are passed.

clarkdo: Good catch, so other loaders should be also same? We can add the loaders package resolving right before webpack started by sth like: extendBuild, nuxt hook or webpack hook

pi0:
Hi @paul-soporan @arcanis. For an update, nuxt3 is rewritten with typescript and single package (not mono-repo anymore). And will have first-class cherry support. Closing this PR to focus on proper support on nuxt3 and avoid potential regressions. 

This PR helped a lot to understand dependency resolution mistakes. Thanks again ❤️ and stay tuned for good news 🔥

evanrlong:
@paul-soporan @pi0 - Is there a known way to get Nuxt 2 working with Yarn 2 without this merge request?

merceyz:
@evanrlong Yes, you can use the node-modules linker (`yarn config set nodeLinker node-modules`) or you can try with PnP Loose mode (`yarn config set pnpMode loose && yarn config set pnpFallbackMode all`), though I don't know if that will work

https://yarnpkg.com/configuration/yarnrc#nodeLinker
https://yarnpkg.com/configuration/yarnrc#pnpMode
https://yarnpkg.com/configuration/yarnrc#pnpFallbackMode

evanrlong:
@merceyz -Wow I totally missed your reply. Thanks, that is really helpful!

I've played around with this a bit, and have a proof of concept for a thin wrapper around Nuxt 2 that can provide Yarn 2 support with Plug'n'Play.  

https://github.com/evanrlong/nuxt-berry

@pi0 - Would you be open to a narrower merge request that adds the currently undeclared dependencies to Nuxt without any of the webpack and resolver changes?  Specifically, I'd be looking at adding [these dependencies](https://github.com/evanrlong/nuxt-berry/blob/main/.yarnrc.yml). See [this issue](https://github.com/evanrlong/nuxt-berry/issues/2) for more info.

Adding those dependencies should allow a project such as the one I posted to function as expected with Yarn 2 even when published as a library. I wouldn't think it would cause any major regressions that would impact the release of Nuxt 3.

Or if there is a better mechanism through which I should pursue this in Yarn, let me know @arcanis.

Thanks! 😃

merceyz:
If you strip out the PnP plugin the rest of this PR is just bug fixes, really wish it got merged as, if you're unlucky, nuxt wont work correctly https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

> Or if there is a better mechanism through which I should pursue this in Yarn, let me know @arcanis.

The `patch:` protocol is always an option https://yarnpkg.com/features/protocols#patch

evanrlong:
I agree.  If the dependency resolution and resolver fixes from this PR were merged, the PnP functionality could be added through a standard Nuxt module (using [`pnp-webpack-plugin`](https://github.com/arcanis/pnp-webpack-plugin)) instead of needing to fully wrap the Nuxt package like I did here.  

With that being said, I'd be happy with any approach that will make Yarn 2 available for use until Nuxt 3 is out.  Dependency resolutions are the only hard blocker to that.  If we could get the resolver fixes added too, that would make the approach simpler, but I see that as a want, not a need right now.

I don't think that the `patch:` protocol will work in this case.  https://yarnpkg.com/features/protocols#why-cant-i-add-dependencies-through-the-patch-protocol

It looks like the options for making [`nuxt-berry`](https://github.com/evanrlong/nuxt-berry) deployable as an npm package are...

1. **Properly resolve dependencies in Nuxt.**  This is preferred as it would allow users to install and use `nuxt-berry` without requiring additional Yarn config.

2. **Add dependency resolutions to Yarn 2's [`plugin-compat`](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat)**.  This would also work without additional Yarn config. However, it could potentially conflict with other packages in a user's monorepo due to the [`dependenciesMeta`](https://github.com/evanrlong/nuxt-berry/blob/main/package.json) settings I would need to add.  Those settings are needed for Nuxt to work with Zero-Installs.

3. **Make a custom Yarn 2 plugin for Nuxt Berry.** This is not ideal, as it would have the `dependenciesMeta` problem from option 2 and also require the user to install a Yarn plugin along with the `nuxt-berry` package.

evanrlong:
Or a possible option 4.

4. **Make a custom Yarn 2 plugin to fix dependencies and resolvers.** This would requires a `patch:` and some package externsions, but would remove the need for `nuxt-berry` entirely. Instead, the user would just install `nuxt` and a yarn plugin. It could be model off (and maybe rolled in to?) Yarn's [`plugin-compat`](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat).

I haven't really looked into that though and don't know the ramifications.

merceyz:
> I don't think that the patch: protocol will work in this case. https://yarnpkg.com/features/protocols#why-cant-i-add-dependencies-through-the-patch-protocol

I meant for adding `require.resolve` etc

evanrlong:
> I meant for adding `require.resolve` etc

Gotcha that makes sense. The problem I see with that is distribution. We would have to wrap the `patch` in a Yarn plugin and ideally get it added to the official plugin list for easy install. Probably doable, its just an area I am personally less familiar with.

evanrlong:
Also, would those patches potentially break when the user bumps Nuxt versions?

pi0:
Really nice initiative @evanrlong . Honestly I didn't wanna nuxt 2 having breaking changes but also don't wanna keep it an obsolete tool not supporting yarn2. Opening a branch based on this PR and moving changes behind a flag. I hope it helps :)

pi0:
Finally we will have proper support for pnp with #8389 🎉 You can try with `nuxt-pnp` pre-releases and [this demo](https://github.com/pi0/nuxt-pnp-demo). 

It couldn't be possible without your helps @arcanis @paul-soporan @merceyz ❤️

evanrlong:
That is awesome. Thanks @pi0!

PR #7306, opened 2020-05-03
feat(modern): remove inlined safari fix in csp mode

clarkdo:
## Types of changes

## Description

Idea is from https://github.com/vuejs/vue-cli/pull/2741
Related issue: https://github.com/nuxt/nuxt.js/issues/4876

If `csp` is enabled without `unsafe-inline`, let's inject safari no-module fix as an external script.

## Checklist:

### DIFF
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -144,9 +144,17 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
     }
 
     if (modern) {
+      let noUnsafeInline = false
+      const { csp } = this.buildContext.options.render
+      if (csp) {
+        const { policies = {} } = csp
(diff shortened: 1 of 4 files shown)


pi0:
The number of initial parallel requests is [critical](https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser) and [SplitChunksPlugin](https://webpack.js.org/plugins/split-chunks-plugin/#defaults) is optimized to keep with this number.  Is it **necessary** for CSP to not inline scripts?

pi0 (dismissed):

clarkdo:
> Is it necessary for CSP to not inline scripts?

I think `inline scripts` is restricted by most csp users as it's safer.

And this pr is only for `spa`, `modern` and `csp` users.

PR #7307, opened 2020-05-03
feat(server): support csp report-uri

Spice-Z:
## Types of changes

## Description

Now, we cannot set report-uri when setting alloweSources option.
I enable to set report-uri in Content-Security-Policy-Repor-Only header, even if allowedSources is set.

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -70,9 +70,10 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
     if (options.render.csp && cspScriptSrcHashes) {
       const { allowedSources, policies } = options.render.csp
-      const cspHeader = options.render.csp.reportOnly ? 'Content-Security-Policy-Report-Only' : 'Content-Security-Policy'
+      const isReportOnly = !!options.render.csp.reportOnly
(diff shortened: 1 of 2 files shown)


clarkdo (approved):

pi0 (changes requested):
`report-uri` seems deprecated and discouraged. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

Should we provide both `report-uri` and `report-to` and use new recommanded name?

Spice-Z:
@pi0 
Thank you for pointing out it!
I agree with you, it's better to provide both  `report-uri` and `report-to`.
But to use `report-to` directive, server have to set `Report-To` header, like this.

```
Content-Security-Policy-Report-Only: ...; report-to csp-endpoint
Report-To: {
    ...
  }, {
    "group": "csp-endpoint",
    "max_age": 10886400,
    "endpoints": [{
      "url": "https://example.com/csp-reports"
    }]
  }
```
(from  https://developers.google.com/web/updates/2018/09/reportingapi#csp)

Now, Nuxt doesn't support `Report-To` header in CSP options, so adding `report-to` and `Report-To` header  support is totally new feature.
I think it's better to make new issue to support `report-to` directive, and  want to focus on `report-uri` on this PR.

Issue #5100, opened 2019-02-24
dynamic runtime environment variables


janesser:
While striving for immutable build artifacts we came across some possible improvement points.

First let me explain the motivation behind "immutable build artifact". Alongside of the stages like e,g, DEV, TEST, QA, UAT, PROD the configuration changes, as an example nuxt-axios API_URL and API_BROWSER_URL. Ideally this can be injected by env vars, which pretty much fits the kubernetes way of dealing with configuration.

Presently a work-around can be achieved with help of https://github.com/samtgarson/nuxt-env, which does a decent job in providing needed injection mechanism. The down-side is that no nuxt-plugin by default goes for this.$env, which drives one in a lot of customizing of plugins. Not to mention the broken upgrade ease.

Can you imagine a way to support this natively in some of the next nuxt releases?

<div align="right"><sub><em>This question is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8719">#c8719</a>)</em></sub></div>

pi0:
Hi @janesser. I've transferred issue from RFC to here because it is more an enhancement discussion.

Values passed to the `env` section of `nuxt.config` will be _replaced_ by a webpack plugin during the build. Would you please elaborate more what you expect from storing it in build artifact?

vhf:
I also ran into this problem, here is one of my example case:

I'm using nuxtjs auth module with some OAuth API. Here is what the config looks like: https://auth.nuxtjs.org/reference/providers/github , in `nuxt.config.js`:

```js
auth: {
  strategies: {
      github: {
        client_id: '...',
```

I'd like to build my app once and deploy it many times with different configurations. This cannot be done because `client_id: process.env.FOO_CLIENT_ID` will be replaced at build time and the build-time value will be written in the build artifacts.

Here you couldn't build your app in a container and deploy it with different `FOO_CLIENT_ID` values, you would have to build the app each time you start a container.

aldarund:
@pi0 its more about have runtime env vars like in the module that is linked. E.g. build once deploy to different environments same artifact, isntead of building each time for each env ( e.g. staging, test, production etc )

pi0:
I see the point. And auth and axios are probably best examples for such need. However, this is a duplicate topic and issue was discussed here is a short brief of status:

- 1. `process.env.*` from **server side** is runtime configurable. axios module supports it for example for `API_URL`
[...]

PR #7312, opened 2020-05-04
feat: runtime config and built-in dotenv support

pi0:
## Types of changes

## Description

Nuxt supports [env](https://nuxtjs.org/api/configuration-env) config to provide configuration via `process.env`. This is done by webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/). This approach had two downsides:
- Values are read during build time and persisted into webpack bundle. So for a change to `process.env` we need to rebuild which is against [12factor](https://12factor.net) app design
- A simple `env` option can easily mislead users to expose secret keys to client-side bundle
- Needs [dotenv module](https://github.com/nuxt-community/dotenv-module) to load `.env` file

**Related Issue:** dynamic runtime environment variables (#5100)
**Previous works:** [nuxt-env](https://github.com/samtgarson/nuxt-env) module by @samtgarson
**Try locally**: swtich to `feat/runtime-config` branch and use `yarn nuxt test/fixtures/runtime-config -o`

### Runtime Config

Two new options added to `nuxt.config` to allow passing runtime configuration which is then accessible using context `$config`.

Config is added to Nuxt payload (`__NUXT__.config`) so there is no need to rebuild to update runtime configuration. SSR, SPA, and Static targets supported with an exception that for static target, a regenerate is necessary.

Verbose config names are chosen intentionally to avoid users exposing private information to clients. `privateRuntimeConfig` always overrides `publicRuntimeConfig` for server-side.

```js
export default {
  publicRuntimeConfig: {},
  privateRuntimeConfig: {}
}
```

###  Usage

Anywhere in context (pages, store, components, plugins), can use `this.$config` or `ctx.$config` or inside templates, directly using `{{ $config.* }}`

```vue
<script>
export default {
  asyncData ({ $config: { baseURL } }) {
    fetch(`${baseURL}/test`)
  }
}
</script>
```

### `.env` support

Similar to vue-cli (*), `.env` file will be always loaded via [dotenv](https://github.com/motdotla/dotenv) and is accessible via `process.env` and `options._env`. `process.env` is updated so one can use it right inside `nuxt.config` for runtime config. Values are interpolated and expanded with an improved version of [dotenv-expand](https://github.com/motdotla/dotenv-expand). `.env` file is also watched to reload during `nuxt dev`. Path can be set via cli `--dotenv <file>` or disabled by `--dotenv false`.

### Expand/Interpolation Support

Supported both for env (`_env`) and runtime configs. Expand for runtime config happens only if there is already a key (`API_SECRET` example). Interpolation allows nesting env vars (`baseURL` example) (**). It is also possible to use a function for `publicRuntimeConfig` and `privateRuntimeConfig` but not recommended.

`.env`:

```config
BASE_URL=/api
PUBLIC_URL=https://nuxtjs.org
API_SECRET=1234
```

`nuxt.config.js`:

```js
export default {
  publicRuntimeConfig: {
    baseURL: process.env.BASE_URL
  },
  privateRuntimeConfig: {
    baseURL: '${PUBLIC_URL}${BASE_URL}',
    API_SECRET: ''
  },
}
```

### Notes

(*) With the difference that we don't load multiple `.env` files as advised by both [dotenv](https://github.com/motdotla/dotenv#should-i-have-multiple-env-files) and 12-factor [config](https://12factor.net/config) section. ~~Also despite `vue-cli`, it is runtime and nothing will be exposed unless specified in runtime config.~~ We also already expose `NUXT_*` envs to `env` (which is exposed to client if used)

(**) We may need to disable `no-template-curly-in-string` eslint rule

An alternative implementation is available in [runtime-config2](https://github.com/nuxt/nuxt.js/tree/feat/runtime-config2) which allows accessing config without context but importing `~nuxt/config`. But it is much more hacky implementation. (later we can add similar method with limited support)

## Checklist:

- [ ] Docs PR
- [x] Add Tests
- [x] Watch `.env`
- [x] ~~Check CSP for SPA~~ (should also check for static target)
- [x] Make `.env` name customizable via loadOptions


packages/config/src/load.js
```diff
@@ -27,6 +29,16 @@ export async function loadNuxtConfig ({
     configFile = undefined
   }
 
+  // Load env
+  const env = loadEnv(rootDir)
+
+  // Fill process.env so it is accessable for nuxt.config
```
Atinux, line 35:
```suggestion
  // Fill process.env so it is accessible in nuxt.config
```

packages/config/src/load.js
```diff
@@ -66,5 +78,72 @@ export async function loadNuxtConfig ({
...
+  }
+
   return options
 }
+
+function loadEnv (rootDir) {
+  const env = Object.create(null)
+
+  // Read dotenv
+  const dotenvFile = path.resolve(rootDir, '.env')
```
pimlie, line 98: A common approach seem to be to use a different `.env` file for each `NODE_ENV` (or even more like dev, staging, production). Should we support that?

packages/config/src/load.js
```diff
@@ -66,5 +78,72 @@ export async function loadNuxtConfig ({
...
+    if (options[c]) {
+      expand(options[c], options._env)
+    }
+  }
+
   return options
 }
+
+function loadEnv (rootDir) {
+  const env = Object.create(null)
```
pimlie, line 95: Why this and not just `const env = {}` ?

packages/config/src/load.js
```diff
@@ -66,5 +78,72 @@ export async function loadNuxtConfig ({
...
+
+  // Read dotenv
+  const dotenvFile = path.resolve(rootDir, '.env')
+
+  if (fs.existsSync(dotenvFile)) {
+    const parsed = dotenv.parse(fs.readFileSync(dotenvFile, 'utf-8'))
+    Object.assign(env, parsed)
```
pimlie, line 102: Couldnt this just be a `env = parsed`?

packages/config/src/load.js
```diff
@@ -66,5 +78,72 @@ export async function loadNuxtConfig ({
...
+  }
+
+  // Apply process.env
+  Object.assign(env, process.env)
```
pimlie, line 106: Do we really want to always inject the PATH, PWD, __SCRIPT_PID into the env?

If we would pass process.env as second arg to expand a user could choose to inject this by just setting `PATH=` in his/her .env file

packages/config/src/load.js
pi0: Avoid object prototype pollution (null means no proto)

packages/config/src/load.js
pi0: God point but this way we should either filter process.env or not support real environment variables at all! `.env` is sometimes good for environments that cannot or are not easy to use os environments but not a replacement for it.

packages/config/src/load.js
pi0: As mentioned in the readme, it is both not advised by dotenv and 12 factor. Also, this step is before loading nuxt.config. We can make configurable for loadOptions however :)

.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   rules: {
     'no-console': 'error',
     'no-debugger': 'error',
+    'no-template-curly-in-string': 0,
```
manniL, line 19: Is there a reason to disable this check globally instead on a line-by-line base?

packages/config/src/load.js
```diff
@@ -27,6 +29,16 @@ export async function loadNuxtConfig ({
     configFile = undefined
   }
 
+  // Load env
+  const env = loadEnv(rootDir)
+
+  // Fill process.env so it is accessible in nuxt.config
+  for (const key in env) {
+    if (key[0] !== '_' && process.env[key] === undefined) {
```
manniL, line 37: nit for better readability
```suggestion
    if (!key.startsWith('_') && process.env[key] === undefined) {
```

.eslintrc.js
pi0: > (**) We may need to disable no-template-curly-in-string eslint rule

This rule generally prevents using interpolation in `nuxt.config`

manniL: Whoops, missed that. Thanks!

packages/config/src/load.js
pi0: Also, this is dotenv behavior: https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set

packages/config/src/load.js
pi0: e809dbfda93f783dbdb640f0cfe06bad79c24993 (not multi-env but path to file can be customized)

packages/config/src/load.js
pi0: e809dbfda93f783dbdb640f0cfe06bad79c24993 can be disabled with envConfig.systemVars

pimlie: nice, thx

pimlie:
Wdyt about populating env before loadNuxtConfig and passing it within the [nuxt-command-context](https://github.com/nuxt/nuxt.js/blob/feat/runtime-config/packages/cli/src/command.js#L107)?

Eg I sometimes have the need to use env vars in my nuxt config and it would probably be nice to be able to use the normalized `env` object in nuxt.config too.

pi0:
@pimlie seems a good idea. Actually we already have [`configContext`](https://github.com/nuxt/nuxt.js/blob/feat/runtime-config/packages/config/src/load.js#L18) passed by [loadNuxtConfig](https://github.com/nuxt/nuxt.js/blob/feat/runtime-config/packages/cli/src/utils/config.js#L13) when CLI is used. Just had to populate: 6c4762cf353e691233701a8beca10174183b9e83

galvez:
I've found that adding a proxy for access can be helpful in treating some env values, ie.:

```js
env = new Proxy(env, {
  get (_, prop) {
    if (prop.includes('_ENABLED')) {
      return env[prop].trim().match(/^1|(true)$/)
    }
    return env[prop]
  }
})
```

Maybe we can slip this in?

packages/config/src/load.js
```diff
@@ -27,6 +30,21 @@ export async function loadNuxtConfig ({
     configFile = undefined
   }
 
+  // Load env
+  envConfig = defu(envConfig, {
+    dotenv: '.env',
+    systemVars: true,
```
galvez, line 36: This needs to be made optional as it can be problematic in some setups with `pm2`.

I can't find the ref right now but apparently `process.env` has a limit! Couple it with a Kubernetes-based setup with hundreds of pre-populated environment vars and you're left scratching your head for a few days ;)

galvez (dismissed):

packages/config/src/load.js
pi0: I can't understand what's the _limit_? Nothing is exposed it is just an internal object used for expanding. If a k8s deployment is already having hundreds of envs, they are already in process.env

galvez: Actually, sorry, this has no relation to the Nuxt usage. 

This is a problem specific to pm2 and is why I'm using a forked version of it. When forking a new process, it will inject the env for the process, and it will fail if it's "too big". How big it can be apparently varies in different platforms.

Still I think having `systemVars` be made optional might be something of interest to some users, would it be too hard to add it? Maybe expose dotEnvOptions?

pi0: It can be optionally disabled but honestly there is no reason disabling by default. Adviced by both 12 factor and dotenv, environment variables have highest periority and for a container native deployment it is something obviously required ;)

Regarding pm2 issue, if process.env is already too big, this PR makes no difference. Flag only adds it to options._env nothing more 😀

pi0: I will add cli flag too

danielroe:
> I've found that adding a proxy for access can be helpful in treating some env values, ie.:
> 
> ```js
> env = new Proxy(env, {
>   get (_, prop) {
>     if (prop.includes('_ENABLED')) {
>       return env[prop].trim().match(/^1|(true)$/)
>     }
>     return env[prop]
>   }
> })
> ```
> 
> Maybe we can slip this in?

I love Proxies but would be worried about issues with IE11. [The polyfill](https://github.com/GoogleChrome/proxy-polyfill) has significant limitations.

pi0:
@danielroe Actually `_env` is being read from loadConfig (node.js only) object containing both system envs and `.env` file as a utility to easily populate runtimeConfig by users :)

packages/config/src/load.js
pi0: 5cb822135d2e1b9011da9410f0cee8ba84a5178a

manniL:
From a naming perspective, would it make sense to use a nested object á la

```js
runtimeConfig: {
  public: {},
  private: {}
}
```

instead of prefixed separate objects? 🤔 

Or wouldn't that work because public/private are reserved keywords?

pi0:
@manniL There shouldn't be any problem with keywords at least for now (and [alternative impl](https://github.com/nuxt/nuxt.js/blob/feat/runtime-config2/packages/config/src/config/_common.js#L86) is using this) but it was mostly flat because of we only have two possible namespaces and considering autocomplete could be better DX. ([next](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) is also using similar convention)

atinux (approved):

PR #7318, opened 2020-05-05
fix(webpack): allow hmr client path to be set

mathieutu:
## Types of changes

## Description

Hi,
I'm trying to set the path to the webpack hmr client in the config, but as the path is set by hand in the query string, we end with 2 paths in query string: 
```js
// nuxt.config.js

hotMiddleware: {
  client: {
    path: DEV_URL + '/__webpack_hmr/client',
  },
},
```

```js
// query string sent to client by /packages/webpack/src/config/client.js#L200:

"?reload=true&timeout=30000&ansiColors=&overlayStyles=&path=http%3A%2F%2Flocalhost%3A3000%2F__webpack_hmr%2Fclient&name=client&path=/__webpack_hmr/client"
```

The HMR client (https://github.com/webpack-contrib/webpack-hot-middleware/blob/master/client.js#L17) will parse this querystring, and set an object with two paths, and then will try to make the request to a stringified version of an array of theses two paths which will result in a 404:
>XHR GET http://localhost:3000/__webpack_hmr/client,/__webpack_hmr/client

This PR fix that by using the path set in config.

It should not be a breaking change, as nobody could have used the behavior with theses two paths.

PS: I've kept the `&path` way because of [this commit](https://github.com/nuxt/nuxt.js/pull/4796/commits/aefba48e743792547979cdc074c09cfe4d6d9569) message:
>HMR client path should not be encoded 

But actually, I see no problem with encoding it, as the client will decode it afterwards. If ok for you, I will just merge the path with the other options.

Maybe @wangxingkai can help us with that?

Thanks for your amazing work,
Mathieu.

## Checklist:

### DIFF
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -185,7 +185,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
     } = this.buildContext
 
     const { client = {} } = hotMiddleware || {}
-    const { ansiColors, overlayStyles, path, ...options } = client
+    const { ansiColors, overlayStyles, path: pathFromOptions, ...options } = client
 
     const hotMiddlewareClientOptions = {
       reload: true,
@@ -195,7 +195,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       ...options,
       name: this.name
     }
-    const clientPath = path || `${router.base}/__webpack_hmr/${this.name}`
+    const clientPath = pathFromOptions || `${router.base}/__webpack_hmr/${this.name}`
 
     const hotMiddlewareClientOptionsStr =
       `${querystring.stringify(hotMiddlewareClientOptions)}&path=${clientPath}`.replace(/\/\//g, '/')
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -185,20 +185,19 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
     } = this.buildContext
 
     const { client = {} } = hotMiddleware || {}
-    const { ansiColors, overlayStyles, path: pathFromOptions, ...options } = client
+    const { ansiColors, overlayStyles, ...options } = client
 
     const hotMiddlewareClientOptions = {
       reload: true,
       timeout: 30000,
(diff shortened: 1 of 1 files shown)


packages/webpack/src/config/client.js
```diff
@@ -195,7 +195,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       ...options,
       name: this.name
     }
-    const clientPath = path || `${router.base}/__webpack_hmr/${this.name}`
+    const clientPath = pathFromOptions || `${router.base}/__webpack_hmr/${this.name}`
 
     const hotMiddlewareClientOptionsStr =
       `${querystring.stringify(hotMiddlewareClientOptions)}&path=${clientPath}`.replace(/\/\//g, '/')
```
pi0, line 201: Can we instead push `path` to `hotMiddlewareClientOptions` string options and move default value to L190?

mathieutu: Hi pi0!
I'd be glad to, but as I said in my PS, there is a [commit](https://github.com/nuxt/nuxt.js/pull/4796/commits/aefba48e743792547979cdc074c09cfe4d6d9569) from @wangxingkai which is titled `HMR client path should not be encoded`. I don't know why it shouldn't (actually I've tested in local and everything seems to work well) but I'm afraid to create an unwanted BC or a regression by doing it. This is why I kept this way of doing. 

If it's ok for you, I'll do it right now!

pi0: Thanks for precisely checking history @mathieutu. I see but actually there is not any specific reason. The only reason may be if special chars are in path and encoding makes special problems... If you want to keep it as is would you please add a comment above for making it TODO to investigate? Otherwise we can just refactor and if anything was broken and reported revert.

pi0: /cc @clarkdo any reason you think would be?

clarkdo: I suppose it's from fixing unit test instesd of changing unit test, so it should be OK to change as your suggestion.

mathieutu: And done!

pi0 (approved):

PR #7358, opened 2020-05-12
feat(server): add unload method support for serverMiddleware

### LINKED ISSUE #7355: Allow server middleware to export an unload/destroy method

In my application, I am using server middleware to serve an API - My API uses setInterval for some frequent background tasks. When in dev mode and I make some changes, the API reloads yet the old intervals stay alive, creating more each time I reload. A unload/destroy method would allow for properly closing the server middleware.

```js
const interval = setInterval(() => console.log('This is my interval'), 30 * 1000);

module.exports = {
  path: '/api',
  handler: /* express */,
  unload() {
    clearInterval(interval);
  }
};

```

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10646">#c10646</a>)</em></sub></div>

atinux:
## Types of changes

## Description

Resolves #7355

I don't know if we should support `async unload()`

I need to add a way to test it and update the documentation to add a note.

## Checklist:


packages/server/src/server.js
```diff
@@ -369,6 +378,7 @@ export default class Server {
       await this.renderer.close()
     }
 
+    this.app.stack.forEach(this.unloadMiddleware)
```
pi0, line 381: Small note: This might break metal as it doesn't has `stack` prop but anyway it will be broken for HMR functionality.

Atinux: What about `this.app.stack && this.app.stack.forEach(this.unloadMiddleware)` ?

pi0: It would be probably safer :) But also current status is fine because a standard express/connect instance should expose stack

pi0 (approved):

PR #7374, opened 2020-05-16
feat: components

pi0:
## Types of changes

## Description

👀

**Core fix:** Don't use module handler fn name as unique key so we can do unique module check by default. This is important as one may add components module to modules/buildModules section

## Checklist:


kevinmarrec (dismissed):
😍

packages/config/src/options.js
```diff
@@ -448,6 +448,11 @@ export function getNuxtConfig (_options) {
       .map(([path, handler]) => ({ path, handler }))
   }
 
+  // Components module
+  if (options.components) {
+    options.buildModules.push('@nuxt/components')
```
pi0, line 453: Side note: This is safe if user already specify in `modules` or `buildModules` section ([src](https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/module.js#L187)).

atinux (approved):
Please add `[release]` when merging :)

Atinux:
Added the reference in upcoming docs: https://github.com/nuxt/docs/pull/1922/commits/46c6ba8a29ecbc1553d8adf08e8d1a96a796ffab

PR #7408, opened 2020-05-22
feat: experimental nuxt-telemetry

atinux:
## Types of changes

Add anonymous telemetry system to better understand how our users uses Nuxt and keep improving Nuxt for the masses.

## Description

https://github.com/nuxt/telemetry

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/1922)


pi0 (approved):

packages/config/src/options.js
```diff
@@ -473,5 +474,19 @@ export function getNuxtConfig (_options) {
     staticAssets.versionBase = urlJoin(staticAssets.base, staticAssets.version)
   }
 
+  // Nuxt Telemetry
+  if (
+    options.telemetry !== false &&
+    !options.test &&
+    !destr(process.env.NUXT_TELEMETRY_DISABLED) &&
```
clarkdo, line 481: Do we need a new dependecy `destr` here ?

pi0: Dependency is added. It is to check for `true` value not just env existence.

clarkdo: I mean do we need this dependency for just one checking here ?

pi0: No, actually I created it to use in more places including runtimeConfig (convert `"true"` to `true`) :see_no_evil:

clarkdo (approved):

PR #7422, opened 2020-05-27
feat(generator): `export:route` hook and `setPayload` 

pi0:
## Types of changes

## Description

This PR allows providing a shared payload to all `nuxt export` routes similar to [generate payload](https://nuxtjs.org/api/configuration-generate#speeding-up-dynamic-route-generation-with-code-payload-code-)

Thanks for @nilsnh (https://github.com/nuxt/nuxt.js/pull/6159#issuecomment-634565730) for discovering this :)

**Usage:**

`nuxt.config`

```js
export default {
  target: 'static',
  hooks: {
    export: {
      async before ({ setPayload }) {
        const categories = await backend.getCategories()
        setPayload({ categories })
      }
    }
  }
}
```

```pages/index.vue```

```js
export default {
  async asyncData({ params, error, payload }) {
    if (payload) { return { categories: payload.categories } }
    return { categories: await backend.getCategories() }
  }
}
```

It is also possible for modules using `export:route` to fill payload per route.

**Why using `setPayload` and not directly allow setting `payload` props?**

Current usage is checking `payload` is not `null` (default value) [docs](https://nuxtjs.org/api/configuration-generate#speeding-up-dynamic-route-generation-with-code-payload-code-). So it changes `null` to object on first set.

## Checklist:

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -25,7 +25,10 @@ export default class Generator {
     this.generatedRoutes = new Set()
 
     // Shared payload
-    this.payload = { ...this.options.generate.payload }
+    this._payload = null
+    this.setPayload = (payload) => {
+      this._payload = { ...this._payload, ...payload }
+    }
   }
 
   async generate ({ build = true, init = true } = {}) {
@@ -255,7 +258,7 @@ export default class Generator {
     const routeMap = {}
     // Fill routeMap for known routes
     routes.forEach((route) => {
-      routeMap[route] = { route, payload: {} }
+      routeMap[route] = { route, payload: null }
     })
     // Fill routeMap with given generate.routes
     generateRoutes.forEach((route) => {
@@ -263,7 +266,7 @@ export default class Generator {
       const path = isString(route) ? route : route.route
       routeMap[path] = {
         route: path,
-        payload: route.payload || {}
+        payload: route.payload || null
       }
     })
     return Object.values(routeMap)
@@ -273,11 +276,17 @@ export default class Generator {
     let html
     const pageErrors = []
 
-    // Apply shared payload
-    payload = { ...this.payload, ...payload }
+    const setPayload = (_payload) => {
+      payload = { ...payload, ..._payload }
+    }
 
-    await this.nuxt.callHook('generate:route', { route, payload })
-    await this.nuxt.callHook('export:route', { route, payload })
+    // Apply shared payload
+    if (this._payload) {
+      setPayload(this._payload)
+    }
+
+    await this.nuxt.callHook('generate:route', { route, setPayload })
+    await this.nuxt.callHook('export:route', { route, setPayload })
 
     try {
       const renderContext = {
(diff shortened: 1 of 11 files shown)


packages/generator/src/generator.js
```diff
@@ -23,6 +24,12 @@ export default class Generator {
       isUrl(this.options.build.publicPath) ? '' : this.options.build.publicPath
     )
     this.generatedRoutes = new Set()
+
+    // Shared payload
+    this._payload = null
+    this.setPayload = (payload) => {
```
clarkdo, line 30: Would `getter/setter` for `payload` make more sense here ?

clarkdo (dismissed):

packages/generator/src/generator.js
```diff
@@ -47,6 +54,7 @@ export default class Generator {
 
     // Done hook
     await this.nuxt.callHook('generate:done', this, errors)
+    await this.nuxt.callHook('export:done', this, { errors })
```
clarkdo, line 57: Are we gonna deprecate `generate:*` hooks ?

packages/generator/src/generator.js
pi0: We may need something like a proxy then for deep watch. setPayload can ensure it is deep assigned and is simple :)

packages/generator/src/generator.js
pi0: Hmm not sure. But since it is for `nuxt export` and we already have `export` config alias, was thinking makes sense to introduce new hooks so we have backward compatibility if deprecating for Nuxt3

packages/generator/src/generator.js
clarkdo: This make me a little confuse, we have `this.setPayload` which is never called.

We also have another `setPayload` as a [scoped var](https://github.com/nuxt/nuxt.js/pull/7422/files#diff-22690cd173b27a10986baff5efcd4ac4R280)

pi0: The one set in the constructor is used by `export:before` hook to set shared payload.

The one in scoped var is used for `export:route` hook.

clarkdo: Ah, I didn’t realize it’s passed by this.

packages/generator/src/generator.js
Atinux: Indeed we duplicate them now, and we will deprecate nuxt generate (I think we can for Nuxt 2.14) and delete it for Nuxt 3

packages/generator/src/generator.js
clarkdo: I think I might have understood what’s happening here, so we have a global payload, each route has a scoped payload, what we’re doing is to make sure default value is always populated to updated payload.

packages/generator/src/generator.js
Atinux: We have to be careful since it’s not only renaming for module authors :)

Atinux: Can we find a nice way for module author to give the option that export is available? Without needing to check the Nuxt version?

pi0: @Atinux They can check `options.target` to see if using `static` target or not. Also agree about deprecate if we also deprecate `nuxt generate` command.

Atinux: Perfect 🙌

atinux (approved):

PR #7430, opened 2020-05-28
feat(config): support `.nuxtrc`

pi0:
## Types of changes

## Description

Support `.nuxtrc` (per project and user home) to extend `nuxt.config` without making modifications. The use case is for any module that wants to persist a state or config per project on an automated basis like telemetry module to set `telemetry.enabled`. RC format is flattened INI but config is read unflattened with native types (see [rc9](https://github.com/nuxt-contrib/rc9))

Loading priority: configOverrides > nuxtConfig > .nuxtrc > .nuxtrc (global)

## Checklist:


clarkdo (commented):
The change looks good.

I'm wondering if it's doing different things like what most framworks are doing, like babel won't have `babel.config.js --falback-to-->  .babelrc`, eslint wont have `.eslintrc.js --falback-to--> .eslintrc`.

Another point is : Regarding to use rc file, is there anything impossible or hard to be implemented by using `nuxt.config.js` (or maybe a global `nuxt.config.js`) ?

pi0:
> I'm wondering if it's doing different things like what most framworks are doing, like babel won't have babel.config.js --falback-to--> .babelrc, eslint wont have .eslintrc.js --falback-to--> .eslintrc.

Indeed this is same behavior we will have. The current loadConfig fixture for nuxtrc does not have a `nuxtconfig.js` file.  Main different with other (js) frameworks, is that nuxtrc is flat INI that allows easy modifications by automated tools (even a bash script) but loaded as valid json.

> Another point is : Regarding to use rc file, is there anything impossible or hard to be implemented by using nuxt.config.js (or maybe a global nuxt.config.js) ?

The most advantage is updating as for `nuxt.config` we need to parse syntax with different possibilities ({typescript, esm, cjs} x {exported object, exported function}) compared to a dead-simple syntax and using `rc.update()`. It also makes `nuxt.config` more readable by hiding tokens or configs like telemetry into a plain file. Another advantage is that this file can be optionally checked-in into a repository so based on project requirements can be gitignored or not.

clarkdo:
So if we always respect `nuxt.config.js`, `.nuxtrc` and `~/.nuxtrc` at same time, do we need to provide a way to at least disable `~/.nuxtrc` in an app for edge cases ?

pi0:
> So if we always respect nuxt.config.js, .nuxtrc and ~/.nuxtrc at same time, do we need to provide a way to at least disable ~/.nuxtrc in an app for edge cases ?

Agree. If we find a valid problematic case can support CLI flag but I don't think it would be necessary for initial support.

packages/config/src/load.js
```diff
@@ -87,6 +88,15 @@ export async function loadNuxtConfig ({
     options.rootDir = rootDir
   }
 
+  // Load Combine configs
+  // Periority: configOverrides > nuxtConfig > .nuxtrc > .nuxtrc (global)
```
Atinux, line 92: Priority*

atinux (dismissed):

packages/config/src/load.js
pi0: Fixed

PR #7431, opened 2020-05-28
fix(generator): respect `options.router.trailingSlash` for crawler

TheAlexLichter:
## Types of changes

## Description
When crawling and enforcing a trailing slash, all pages will be crawled twice because the trailing slash is skipped. This also leads to massive errors because pages without trailing slashes can't be generated (as they throw a 404 when enforcing a trailing slash)


pi0 (approved):
Thanks! Seems we need a utility for trailing slash asap

PR #7432, opened 2020-05-28
types: move `@nuxt/types` to nuxt mono-repo

pi0:
## Types of changes

## Description

As discussed with @kevinmarrec current status of `@nuxt/types` is almost stable. This refactoring helps to keep types more updated with each change and also publishing `@nuxt/types-edge` (Can be alias install using `"@nuxt/types": "npm:@nuxt/types-edge"`

### Changes

- Base on `@nuxt/types@0.7.7`
- Bump version to `@nuxt/types@2.12.1` to sync (this also allows making slight breaking changes)
- Added `test:types` and CI step
- Export `NuxtOptions` and `NuxtConfig` which is `Partial<NuxtOptions>` so modules can have correct types with `this.nuxt.options` (like `srcDir` which always has value as passes from `options.js`). `Configuration` export is kept as an alias for legacy support
- Simplify `NuxtConfiguration*` to `NuxtOptions*` (also aligning with `nuxt.options`)
- Support `types: ` for changelog
- Add types for `target` 
- Add types for `export`

## Checklist:

### DIFF
--- a/packages/types/config/index.d.ts
+++ b/packages/types/config/index.d.ts
@@ -1,0 +1,63 @@
+import { Transition } from '../app'
+import { NuxtConfigurationBuild } from './build'
+import { NuxtConfigurationCli } from './cli'
+import { NuxtConfigurationEnv } from './env'
+import { NuxtConfigurationFeatures } from './features'
+import { NuxtConfigurationFetch } from './fetch'
+import { NuxtConfigurationGenerate } from './generate'
+import { NuxtConfigurationHead } from './head'
+import { NuxtConfigurationHooks } from './hooks'
+import { NuxtConfigurationGlobals } from './globals'
+import { NuxtConfigurationLoading, NuxtConfigurationLoadingIndicator } from './loading'
+import { NuxtConfigurationModule, Module } from './module'
+import { NuxtConfigurationPlugin } from './plugin'
+import { NuxtConfigurationRender } from './render'
+import { NuxtConfigurationRouter } from './router'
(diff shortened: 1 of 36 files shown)


clarkdo (dismissed):

packages/types/config/index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  modules: NuxtOptionsModule[]
+  modulesDir: string[]
+  plugins: NuxtOptionsPlugin[]
+  render: NuxtOptionsRender
+  rootDir: string
+  router: NuxtOptionsRouter
+  server: NuxtOptionsServer
+  serverMiddleware: NuxtOptionsServerMiddleware[]
+  srcDir: string
+  transition?: Transition
```
kevinmarrec, line 59: Maybe
```suggestion
  transition: Transition
```
?

packages/types/config/index.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  watch: string[]
+  watchers: NuxtOptionsWatchers
+}
+
+export type NuxtConfig = Partial<NuxtOptions>
+
+/**
+ * @deprecated Use NuxtConfig instead
+*/
+export type Configuration = NuxtConfig// Legacy alias
```
kevinmarrec, line 70:
```suggestion
export type Configuration = NuxtConfig // Legacy alias
```

kevinmarrec (changes requested):

kevinmarrec (approved):
Let's go !

Issue #7441, opened 2020-05-31
Page mounts twice causing huge performance issues


simplenotezy:
### Version

[v2.12.2](https://github.com/nuxt.js/releases/tag/v2.12.2)

### Reproduction link

[https://codesandbox.io/s/nuxt-vuetify-0xz91](https://codesandbox.io/s/nuxt-vuetify-0xz91)

### Steps to reproduce

1) Add a new layout
2) Create a new page and assign this new layout
3) Change from the "default" layout to the page with the new layout
4) See that mounted() and created() is fired twice

### What is expected ?

Page should only be created/mounted once

### What is actually happening?

Page mounts itself twice

### Additional comments?

It writes "I have been mounted" twice in the console, see here: https://codesandbox.io/s/nuxt-vuetify-0xz91. I even tried to give a name to the layout file.

When switching to the about page:

```
i have been created. Layout: default 
i have been mounted. Layout: default 
i have been created. Layout: other 
i have been mounted. Layout: other 
```

It's also weird it calls the layout "other" when I call `this.$nuxt.layoutName`, because I explicitly set the name to be `otherlayout`.

I believe it causes page switches to take extra long time - we just launched our production site: https://thejewelleryroom.com and when switching pages to a page that has been cached by Cloudflare, and the response time is 50-100ms, it still takes 1-2 seconds for the page to switch. Perhaps it is because it mounts the page twice?

Related issue: https://github.com/nuxt/nuxt.js/issues/5703

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10715">#c10715</a>)</em></sub></div>

pi0:
Closing as confirmed to be fixed https://github.com/nuxt/nuxt.js/pull/7442#issuecomment-641908054

PR #7442, opened 2020-05-31
fix(vue-app): prevent double layout execution (#5703)

galvez:
In vue-app/client.js, we use `showNextPage()` to set layout, which is ran in an `afterEach` hook. As reported in #5703, this causes a double execution when switching layouts. Perhaps having this occur in `afterEach` is not fast enough -- switching to `beforeResolve` seems to solve it.

Update: using `afterEach` works, but the actual call must happen before `nextTick`.


packages/vue-app/template/client.js
```diff
@@ -611,7 +606,15 @@ function showNextPage (to) {
     layout = layout(app.context)
   }
   this.setLayout(layout)
-  <% } %>
+  next()
+}
+<% } %>
+
+function checkForErrors (to) {
```
pi0, line 613: to argument is unused. Can we use (app) without binding this?

galvez: Sure.

galvez: Let me know if it's okay now.

galvez: Looks like also we got a CI error but seems to be unrelated?

galvez:
Also reported in #7441.

pimlie:
Does this change influence page transitions? Did you check manually? Dont think we have proper tests to make sure they perform well visually?

But Im worried that this might be a breaking change (albeit it probably a low risk one). If a user already adds their own beforeResolve hook _and_ cancels navigation (or redirects) within that, then the new/wrong layout will already be set.

galvez:
I did test manually, seems ok.

Re: `beforeResolve()`, I had an idea to workaround this, will try it next.

galvez:
@pimlie turns out we can use `afterEach` for this, we just can't `await` for `nextTick`.

simplenotezy:
🎉🎉

atinux (approved):

simplenotezy:
Great!! Good job.

@Atinux Can I pull this fix in through nuxt-edge or how can I get this fix in my application?

pi0:
@simplenotezy `nuxt-edge` releases are nightly so you should be able to test if it is fixing issue in your application.

simplenotezy:
I see, thanks @pi0

simplenotezy:
I can confirm it's working on my end. Good job guys

ademyalcin27:
Hi,
I was using Nuxt@2.9.2 and which is mounted triggered 2 times in mobile. after that I upgraded my nuxt version from 2.9.2 to 2.13.3 I still got same error, Also It's same layout calling 2 times in mounted, '**VirtualPageViewAll**', and '**mobileDefault**' calling in mounted

Ekran Resmi 2020-07-07 13 42 15

hjujah:
For me its not page but all page components are mounting twice. Tried Nuxt 2.13 and 2.14 and edge (2.15) but having the same issue everywhere. Please is there any fix for this?

Issue #7451, opened 2020-06-02
render.csp not compatible with csp3 strict-dynamic


P4sca1:
### Version

~~[v2.12.2](https://github.com/nuxt.js/releases/tag/v2.12.2)~~
[v2.13.3](https://github.com/nuxt.js/releases/tag/v2.13.3)

### Reproduction link

[https://github.com/P4sca1/nuxt-csp3](https://github.com/P4sca1/nuxt-csp3)

### Steps to reproduce

1. Run `yarn dev`
2. Visit `http://localhost:3000` and open browser console

### What is expected ?

No csp errors

### What is actually happening?

~~Many `unsafe-eval` errors because `vendors.app.js` consists of many `eval` calls. This should already be fixed (see https://github.com/nuxt/nuxt.js/issues/4876) but it seems like this is still an issue. Of course this could be fixed by allowing `unsafe-eval` but it would be great if it could work without this less safe rule.~~

Several errors trying to load `_nuxt` scripts:
```
[Report Only] Refused to load the script 'http://localhost:3000/_nuxt/runtime.js' because it violates the following Content Security Policy directive: "script-src 'sha256-mT12zZGdDlFLTlEHBdXxk66zRM2yq44zzYnjADqPTU0=' 'nonce-WFzKYWOSvw1gBKUEEfASwQpfV7pCUKXCeUZjsiyy90ReAKCyKaYEdcj1i6met3HY7SFr2kiB1FP2lSU4gW/vDrTUGezdZbfmLelDwZvBF4mdYjzQdhMH4aIifn9ETQ+S4Z2m2q3gO+ruDYnP5At0ogCBfvu66AIH0D0KZYoqgWo=' 'self' 'strict-dynamic' 'unsafe-inline' https: http:". 'strict-dynamic' is present, so host-based whitelisting is disabled. Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

[...]

PR #7454, opened 2020-06-02
fix(server): set `unsafe-eval` CSP in dev mode with custom CSP policies

FloEdelmann:
## Types of changes

## Description

This could relate to #7451. When the `render.csp.policies` config is an object, `unsafe-eval` is not added to the CSP header in development mode, like it is when an array or nothing is provided.

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -134,20 +134,24 @@ const getCspString = ({ cspScriptSrcHashes, allowedSources, policies, isDev, isR
   }
 
   if (policyObjectAvailable) {
-    const transformedPolicyObject = transformPolicyObject(policies, cspScriptSrcHashes)
+    const transformedPolicyObject = transformPolicyObject(policies, cspScriptSrcHashes, isDev)
     return Object.entries(transformedPolicyObject).map(([k, v]) => `${k} ${Array.isArray(v) ? v.join(' ') : v}`).join('; ')
   }
 
   return baseCspStr
 }
 
-const transformPolicyObject = (policies, cspScriptSrcHashes) => {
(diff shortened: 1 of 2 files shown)


P4sca1:
Why is unsafe-eval needed in dev but not in production?

FloEdelmann:
@P4sca1 Actually, I don't know, but it's the same way it happens in https://github.com/FloEdelmann/nuxt.js/blob/8933d375e9d2e082f2f23b491fe8241762fe4080/packages/server/src/middleware/nuxt.js#L129.

packages/server/src/middleware/nuxt.js
```diff
@@ -134,20 +134,24 @@ const getCspString = ({ cspScriptSrcHashes, allowedSources, policies, isDev, isR
   }
 
   if (policyObjectAvailable) {
-    const transformedPolicyObject = transformPolicyObject(policies, cspScriptSrcHashes)
+    const transformedPolicyObject = transformPolicyObject(policies, cspScriptSrcHashes, isDev)
```
clarkdo, line 137: What do you think adding `'unsafe-eval'` to `cspScriptSrcHashes` if it doesn't exist in dev mode ? Then it should work for both string and object cases.

clarkdo:
@P4sca1 It's because eval is only in dev mode because of webpack devtool config https://webpack.js.org/configuration/devtool/  and we should avoid unsafe eval code in produciton.

clarkdo (commented):
Sorry, I just checked the original issue in comment, I think root cause might be something else. I'll look into it now, please hold this pr.

clarkdo:
@FloEdelmann This issue has been fixed in v2.13.0, so this pr can be closed, we'll do some code cleanup shortly https://github.com/nuxt/nuxt.js/issues/7451#issuecomment-653124094

PR #7455, opened 2020-06-02
feat(generator): allow excluding exported routes using `export:page` hook

### LINKED ISSUE #7440: Ability to prevents page generation in `generate:page`

What if we can add a single boolean to the `page` variable to conditionally control page generation?

https://github.com/nuxt/nuxt.js/blob/c14471681c7134a6a08a69f9eabdb7d213feb2f0/packages/generator/src/generator.js#L361-L373

```js
// Call hook to let user update the path & html
    const page = { route, path: fileName, html, generate: true }
    await this.nuxt.callHook('generate:page', { page, errors: pageErrors })
    await this.nuxt.callHook('export:page', { page })
  
    if (page.generate) {
        page.path = path.join(this.distPath, page.path)
        ...
    }

    return page.generate

```

Or for being fully backward compatible we could use `export:page` hook:
```js
    const page = { route, path: fileName, html, generate: true }
...
    await this.nuxt.callHook('export:page', { page, errors: pageErrors })
  
    if (page.generate) {
...
```

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10713">#c10713</a>)</em></sub></div>

farnabaz:
close #7440

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -359,29 +359,30 @@ export default class Generator {
     }
 
     // Call hook to let user update the path & html
-    const page = { route, path: fileName, html, export: true }
+    const page = { route, path: fileName, html, exclude: false }
     await this.nuxt.callHook('generate:page', page)
     await this.nuxt.callHook('export:page', { page, errors: pageErrors })
 
-    if (page.export) {
-      page.path = path.join(this.distPath, page.path)
+    if (page.exclude) {
+      return false
+    }
+    page.path = path.join(this.distPath, page.path)
 
-      // Make sure the sub folders are created
-      await fsExtra.mkdirp(path.dirname(page.path))
-      await fsExtra.writeFile(page.path, page.html, 'utf8')
+    // Make sure the sub folders are created
+    await fsExtra.mkdirp(path.dirname(page.path))
+    await fsExtra.writeFile(page.path, page.html, 'utf8')
 
-      await this.nuxt.callHook('generate:routeCreated', { route, path: page.path, errors: pageErrors })
-      await this.nuxt.callHook('export:routeCreated', { route, path: page.path, errors: pageErrors })
+    await this.nuxt.callHook('generate:routeCreated', { route, path: page.path, errors: pageErrors })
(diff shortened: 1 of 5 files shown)


test/dev/basic.generate.test.js
```diff
@@ -36,6 +36,13 @@ describe('basic generate', () => {
     nuxt.hook('generate:done', () => {
       writeFileSync(changedFileName, '')
     })
+    nuxt.hook('export:page', ({ page, errors }) => {
+      if (errors.length > 0) {
+        if (page.path.indexOf('/skip-on-fail') === 0) {
```
pi0, line 41: Refactor tips:
 - `> 0` can be omitted
- `indexOf === 0` can be `includes` (or simply eqeqeq)
- Both conditions can be combined

packages/generator/src/generator.js
```diff
@@ -359,27 +359,29 @@ export default class Generator {
...
 
     // Call hook to let user update the path & html
-    const page = { route, path: fileName, html }
+    const page = { route, path: fileName, html, export: true }
     await this.nuxt.callHook('generate:page', page)
-    await this.nuxt.callHook('export:page', { page })
+    await this.nuxt.callHook('export:page', { page, errors: pageErrors })
 
-    page.path = path.join(this.distPath, page.path)
+    if (page.export) {
```
pi0, line 366: We can early return here like `if (!page.export) { return false }`

packages/generator/src/generator.js
```diff
@@ -359,27 +359,29 @@ export default class Generator {
     }
 
     // Call hook to let user update the path & html
-    const page = { route, path: fileName, html }
+    const page = { route, path: fileName, html, export: true }
```
pi0, line 362: Maybe `skip: false` makes more sense?

farnabaz: I will go with `exclude` :)

atinux (dismissed):

Atinux:
Approved but missed the failing checks

pi0 (approved):
Great!

PR #7464, opened 2020-06-04
fix(webpack): reduce false negatives in nodeExternals (#7462)

### LINKED ISSUE #7462: node_modules with "." in their module path accidentally whitelisted from nodeExternals (e.g. @sum.cumo/vue-states)
### Version

### Reproduction link

[https://github.com/JohannesLamberts/nuxt-bug-externals-whitelist](https://github.com/JohannesLamberts/nuxt-bug-externals-whitelist)

### Steps to reproduce

Follow steps from https://github.com/JohannesLamberts/nuxt-bug-externals-whitelist/blob/master/README.md

### What is expected ?

Modules are only whitelisted in webpack nodeExternals if manually specified in nuxt.config.js.

Modules are always whitelisted in webpack nodeExternals if they contain a "." in their path.

### Additional comments?

One side-effect of this behaviour is, that the package from node_modules is re-executed whenever the bundle is re-executed. 
That is once per SSR-request in devMode, since the default for runInNewContext is then set to `true`.

The re-execution produces a *new Instance* of the plugin, which effectively disables [the Vue.js internal check if the Plugin has already been installed](https://github.com/vuejs/vue/blob/dev/src/core/global-api/use.js#L8).
Since the Vue module is *not* re-executed, the Plugin is then installed multiple times, which can lead to secondary, Plugin-dependent Bugs.

In case of `@sum.cumo/vue-states` the Plugin (unsuccessfully) tried to redefine a property on a component once per installation.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10729">#c10729</a>)</em></sub></div>

JohannesLamberts:
## Types of changes

This *could* change behaviour in existing SSR-Apps. The behaviour is however undocumented and unintended (afaik). 

## Description

Resolves: #7462 

## Checklist:

I'm not sure how/if the changes can be tested and would be grateful for guiding/support.


packages/webpack/src/config/server.js
```diff
@@ -20,11 +20,23 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
...
+  /**
+   * files *not* ending on js|jsx|json should be processed by webpack
+   *
+   * this might (still) generate false-positives for
+   * - someFile.umd (referring to someFile.umd.js)
+   * - some.folder (some.folder being a folder containing a package.json)
+   */
+  isNonNativeImport (modulePath) {
+    const extname = path.extname(modulePath)
+    return extname !== '' && !/^\.js(x|on)?$/.test(extname)
```
pi0, line 37: This would introduce more edge cases as name.[type].[ext] is not uncommon for node_modules...

What if we remove ^ from regex and simply test against path.basename? Linked issue is for dot inside part of path so i think it would still fix that ☺️

JohannesLamberts: @pi0 thanks for the amazingly fast response. I might be missing something but would that change the behaviour?

```javascript
/^\.js(x|on)?$/.test(path.extname('file.json.md'))  // false
/^\.js(x|on)?$/.test(path.extname('file.json'))        // true

/\.js(x|on)?$/.test(path.basename('file.json.md')) // false
/\.js(x|on)?$/.test(path.basename('file.json'))        // true

since
path.extname('file.json.md')                                     // '.md'
path.extname('file.json')                                           // '.json'
```

pi0: I mean $ should be still there 😅 But not restriction for start to allow patterns like entry.min.js 

Not related but I'm wondering why jsx is considered native? It should be only js/json (aso mjs) i guess

JohannesLamberts: I also wondered about `jsx` but didn't want to change more than necessary. I can however remove jsx and allow mjs if you'd like.

About extname / basename:
```
path.extname('entry.min.js') === path.extname('entry.js')
```

Therefore patterns like `name.[type].[ext]` should not make any difference (node 12.16.1).

JohannesLamberts: I adjusted the comment to make it (hopefully) more clear.

IMHO the whitelisting should be removed completely and left to the user, due to the limitations of the implementation. But that might be a breaking change...

pi0: I agree. That's why we support `--standalone` and `options.build.standalone`

JohannesLamberts:
@pi0 is there anything more to do for me or can this be merged?

pi0:
@JohannesLamberts No, actually it is perfect. Would be nice removing `jsx` as well if you have time :)

JohannesLamberts:
@pi0 done, I also removed the RegEx to reduce complexity

pi0:
Amazing thanks :heart: Waiting for tests to pass

JohannesLamberts:
@pi0 https://github.com/nuxt/nuxt.js/pull/7464/checks?check_run_id=750834024 failed with

```
jest packages --forceExit --coverage
/bin/sh: 1: jest: not found
```

I cannot see how this could be related to the code changes :/

PR #7469, opened 2020-06-05
feat(server): add new hook render:beforeResponse 

### LINKED ISSUE #7468: Can't overwrite Accept-Ranges header
### Version

### Reproduction link

[https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/middleware/nuxt.js#L81](https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/middleware/nuxt.js#L81)

### Steps to reproduce

See code line

### What is expected ?

I happen to need to overwrite the Accept-Ranges header

Can't do that because `res.setHeader('Accept-Ranges', 'none')` is called right before ` res.end(...)` is being called.

### Additional comments?

The line `res.setHeader('Accept-Ranges', 'none')` should be above the call for `render:route` hook.
In that why the header could be rewritten.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10735">#c10735</a>)</em></sub></div>

Dima000:
## Types of changes

## Description

Resolves: #7468
It also allows to overwrite headers set by next renderer.

## Checklist:

### DIFF
--- a/packages/server/src/middleware/nuxt.js
+++ b/packages/server/src/middleware/nuxt.js
@@ -42,6 +42,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
       const etag = hash ? hash(html, options.render.etag) : generateETag(html, options.render.etag)
       if (fresh(req.headers, { etag })) {
         res.statusCode = 304
+        await nuxt.callHook('render:beforeResponse', url, result, context)
         res.end()
         await nuxt.callHook('render:routeDone', url, result, context)
         return
@@ -80,6 +81,7 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
     res.setHeader('Content-Type', 'text/html; charset=utf-8')
     res.setHeader('Accept-Ranges', 'none') // #3870
     res.setHeader('Content-Length', Buffer.byteLength(html))
+    await nuxt.callHook('render:beforeResponse', url, result, context)
     res.end(html, 'utf8')
     await nuxt.callHook('render:routeDone', url, result, context)
     return html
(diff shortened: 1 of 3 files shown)


packages/server/src/middleware/nuxt.js
```diff
@@ -78,7 +79,6 @@ export default ({ options, nuxt, renderRoute, resources }) => async function nux
 
     // Send response
     res.setHeader('Content-Type', 'text/html; charset=utf-8')
-    res.setHeader('Accept-Ranges', 'none') // #3870
     res.setHeader('Content-Length', Buffer.byteLength(html))
     res.end(html, 'utf8')
     await nuxt.callHook('render:routeDone', url, result, context)
```
pi0, line 84: There might be other use-cases to override headers and while `render:route` hook  _can_ do it it is more for context modification not doing response logic. Maybe we can add a new `render:beforeResponse` hook before `res.end` or simply swap L83 with L84?

pi0: Second idea: we can use `res.getHeader` to ensure not overriding any header already set

Dima000: There was a similar discussion a year ago https://github.com/nuxt/nuxt.js/pull/4755
Maybe it is a good idea to go with the first option you mentioned @pi0

I will try to implement the required changes and see how it goes. Thanks for the reply!

pi0 (approved):

clarkdo (approved):

pi0:
Thanks @Dima000 :heart:

manniL:
Please don't forget to alter the documentation as well.

Dima000:
I'll do a PR to the docs this week. 
Thanks for the feedback guys and thanks for mentaining this awesome project

PR #7475, opened 2020-06-07
fix(vue-app): `redirect` no longer strips trailing slash from URL

acu192:
## Types of changes

## Description
When calling `redirect('...')` with an external (non-relative) URL,
and when that URL has a trailing slash, Nuxt will remove the trailing
slash before performing the redirect. This breaks when the external URL
requires that trailing slash!

The cause of this is simple. Nuxt was filtering empty components
form the URL's path. This is over-paranoid and (in this case) was
causing a bug where the trailing slash was being removed (because
there was an empty component path after that trailing slash, which
was being filtered out before re-joining the path back together.

The fix is simple. Don't filter empty components from the path.
This type of paranoid URL cleaning should not be the responsibility
of the `redirect()` function.

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -596,7 +596,7 @@ function formatUrl (url, query) {
   let parts = url.split('/')
   let result = (protocol ? protocol + '://' : '//') + parts.shift()
 
-  let path = parts.filter(Boolean).join('/')
+  let path = parts.join('/')
   let hash
   parts = path.split('#')
   if (parts.length === 2) {


acu192:
Not sure why `test/dev/cli.test.js` is failing. Doesn't seem related. (?)

clarkdo (commented):
The chage looks good, can you please add a test for this scenario?

acu192:
@clarkdo Sure. Will you point me to where the test should be added? Is there a particular existing test file that this new test should go?

acu192:
Thanks @clarkdo for creating tests for this in #7533 .

PR #7476, opened 2020-06-07
feat: `createRequire` option and disable esm for jest

clarkdo:
## Types of changes

### esm disabled by default

As [esm](https://github.com/standard-things/esm) package has been incompatible since [jest 25](https://jestjs.io/blog/2020/01/21/jest-25), this PR disables esm for jest environments by default to prevent regression happening when testing nuxt with jest. You may need to make changes to your project after this PR:

jest config:

```js
    "transform": {
      "^.+\\.js$": "babel-jest"
    },
```

babel config:

```js
module.exports = function (api) {
  if (api.env('test')) {
    return {
      plugins: ['@babel/plugin-transform-modules-commonjs']
    }
  }
  return {}
}
```

### `createRequire` option

Meanwhile, we have started working on [`jiti`](https://github.com/nuxt-contrib/jiti) project for stable esm/typescript runtime support. You can try it by setting `createRequire: 'jiti'` inside `nuxt.config`.

Also, it is possible to use this option to keep using `esm` with jest < 25 or provide a custom factory function. Setting value to false or unsupported value, makes nuxt using native require (via [create-require](https://github.com/nuxt-contrib/create-require))

## Checklist:


clarkdo:
@pi0 Feel free to merge or close if we're gonna remove esm.

pi0:
@clarkdo PR is good we should remove esm more carefully. I was just thinking to provide an option instead of disabling. Can make changes in same PR :)

packages/config/src/load.js
```diff
@@ -52,9 +53,8 @@ export async function loadNuxtConfig ({
   if (configFile) {
     // Clear cache
     clearRequireCache(configFile)
-
-    options = esm(module)(configFile) || {}
-
+    const requirModule = createRequire(module)
```
Atinux, line 56:
```suggestion
    const requireModule = createRequire(module)
```

packages/config/src/load.js
```diff
@@ -52,9 +53,8 @@ export async function loadNuxtConfig ({
...
+    options = requirModule(configFile) || {}
```
Atinux, line 57:
```suggestion
    options = requireModule(configFile) || {}
```

packages/config/src/load.js
pi0: @Atinux PR is still WIP pushing fixes in few minutes :)

PR #7493, opened 2020-06-11
feat(vue-app): support async watchQuery

rwaness:
## Types of changes

## Description

It may be usefull to do async work inside the watchQuery option instead of rerun ALL the fetch option.

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -157,7 +157,7 @@ async function loadAsyncComponents (to, from, next) {
         (Component, instance) => ({ Component, instance })
       )
       // Add a marker on each component that it needs to refresh or not
-      const startLoader = Components.some(async ({ Component, instance }) => {
+      const asyncWatchQueries = Components.map(async ({ Component, instance }) => {
         const watchQuery = Component.options.watchQuery
         if (watchQuery === true) {
           return true
@@ -170,6 +170,7 @@ async function loadAsyncComponents (to, from, next) {
         }
         return false
       })
+      const startLoader = (await Promise.all(asyncWatchQueries)).some(Boolean)
       <% if (loading) { %>
       if (startLoader && this.$loading.start && !this.$loading.manual) {
         this.$loading.start()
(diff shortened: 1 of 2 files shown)


packages/vue-app/template/client.js
```diff
@@ -157,7 +157,7 @@ async function loadAsyncComponents (to, from, next) {
         (Component, instance) => ({ Component, instance })
       )
       // Add a marker on each component that it needs to refresh or not
-      const startLoader = Components.some(({ Component, instance }) => {
+      const startLoader = Components.some(async ({ Component, instance }) => {
```
pi0, line 160: [Array.prototype.sum](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#:~:text=Description,some()%20immediately%20returns%20true%20.) expects callback returning boolean not promise. We may use `map` + `Promise.all` + `some` if want to support this :)

rwaness: i'll do

Atinux:
Thank you for the PR.

I am not sure to understand the need here, what would be the use case for this?

rwaness:
Salut :)

ie: fetch do multiple async call to init the page dependings on multiple query string parameters
then some of thoses params change so I would like to rerun only the call linked to this param and not all the fetch stack ...

with async watchQuery, we may do async call and then return false to prevent all fetch and asyncData content to be fired again

imo it's to prevent the loading state and unecessary calls to server (fetch > store actions > api)
and so improve ux

I stay here if needed

clarkdo (commented):
Nitpicking: With current change, every watchQuery will be executed even if one of first components returns true.

rwaness:
@clarkdo it may be interested to fire fetch & async as soon as 1 watchQuery return true but all watchQuery have to run (even if 1 return true) because all async process in watchQuery have to be executed.

pi0:
Hi @rwaness. Thanks for your contribution. After talking with the team, introducing this feature would be risky for regressions also making it different with other vue watchers that are not async.

You can maybe use a method to be shared with both fetch and watchQuery to do partial data fetching.

PR #7522, opened 2020-06-15
fix: nuxt instead of nuxt-start

777genius:
I think this is more relevant.


Atinux:
Thank you for the PR but actually `nuxt-start` is here for a reason, it is packed with only runtime dependencies, making it smaller and faster to require :)

PR #7533, opened 2020-06-16
fix(vue-app): not strip trailing slash for redirect external domain

clarkdo:
## Types of changes

## Description

Add test for https://github.com/nuxt/nuxt.js/pull/7475

When redirect to external domain, not strip trailing slash.

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -598,7 +598,7 @@ function formatUrl (url, query) {
 
   let path = parts.join('/')
   if (path === '' && parts.length === 1) {
-    path = '/'
+    result += '/'
   }
 
   let hash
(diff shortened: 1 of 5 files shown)


packages/vue-app/template/utils.js
```diff
@@ -597,6 +597,10 @@ function formatUrl (url, query) {
   let result = (protocol ? protocol + '://' : '//') + parts.shift()
 
   let path = parts.join('/')
+  if (path === '') {
+    return result + (parts.length === 1 ? '/' : '')
```
acu192, line 601: This early return is not using incorporating the `query` parameter. I think your new logic here should be included in the logic on line 610 below (`result += path ? '/' + path : ''`).

clarkdo: Here is only for url only has a domain without path, querystring and hash

acu192: What about if you call redirect like this?

```javascript
<script>
export default {
  middleware ({ redirect }) {
    const query = { foo: 'bar' }
    return redirect('https://nuxtjs.org/', query)
  }
}
</script>
```

I mean, if you pass a second parameter to `redirect()`, like above, which is the `query` parameter, which is being ignored now. Please correct me if I'm wrong.

clarkdo: Ah, correct, thanks

pi0 (approved):

PR #7537, opened 2020-06-17
feat(cli): improve banner lines dx

pi0:
## Types of changes

## Description

- Alignments for OCD people like me
- Use cyan to keep consistent with all **informative** messages and avoid drawing attention
- The environment is `development` or `production` with notice of custom `NODE_ENV` if provided.

Before:

image

After (prod with custom `NODE_ENV`):

image

windows: (outdated)

image

## Checklist:


clarkdo (commented):
What will the banner look like if version is a long git string?

pi0:
@clarkdo It is however only for our internal development not end-users :)

image

clarkdo (dismissed):

pimlie:
What is `mode: server-side`? I understand the first `Rendering: server-side` but `mode: server-side` doesnt seem to make much sense to me.

manniL:
Maybe `ssr: enabled` / `ssr: disabled` instead? 🤔

Atinux:
I am not found of using acronyms here since it targets also beginners, Rendering: server-side looks good :)

manniL:
> I am not found of using acronyms here since it targets also beginners

..which should learn what SSR means anyway when they are using it through dynamic SSR or static site generation, right? 😋

atinux (approved):

PR #7590, opened 2020-06-23
types: fix module and typescript compatibility and add missing crawler definition

### LINKED ISSUE #7585: Missed crawler field definition in NuxtOptionsGenerate
### Version

### Reproduction link

[https://jsfiddle.net/z0sLajyw/](https://jsfiddle.net/z0sLajyw/)

### Steps to reproduce

according to docs https://nuxtjs.org/api/configuration-generate#crawler generate definition can have crawler, but its missed in NuxtOptionsGenerate

### What is expected ?

no errors

Type '{ crawler: false; }' is not assignable to type 'NuxtOptionsGenerate'.
  Object literal may only specify known properties, and 'crawler' does not exist in type 'NuxtOptionsGenerate'.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10829">#c10829</a>)</em></sub></div>

kevinmarrec:
## Types of changes

## Description

- `Configuration` needs to be an interface to be extended by modules types and be compatible with old/current modules.
- `NuxtOptions` should now be extended by new modules instead of `Configuration`
  - It is extending `Configuration` to support modules still using it 
  - `NuxtConfig` is the right import to do over `Configuration` in configuration files, and it inherits `Configuration` through `NuxtOptions`.
- Add missing `crawler` config (fixes #7585)
- Do not use `export *` as of issues with typescript < [3.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#export--is-always-retained)

## Checklist:

### DIFF
--- a/packages/types/config/generate.d.ts
+++ b/packages/types/config/generate.d.ts
@@ -10,6 +10,7 @@ type NuxtOptionsGenerateRoutesFunctionWithCallback = (callback: (err: Error, rou
 
 export interface NuxtOptionsGenerate {
   concurrency?: number
+  crawler?: boolean
   devtools?: boolean
   dir?: string
   exclude?: RegExp[]
--- a/packages/types/config/index.d.ts
+++ b/packages/types/config/index.d.ts
@@ -22,7 +22,9 @@ import { NuxtOptionsWatchers } from './watchers'
 export { Module } from './module'
 export { ServerMiddleware } from './server-middleware'
 
-export interface NuxtOptions extends Record<string, any> {
+export interface Configuration extends Record<string, any> {}
+
+export interface NuxtOptions extends Configuration {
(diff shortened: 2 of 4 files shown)


kevinmarrec:
@pi0 I'm unsure, the issue is that :
- It's deprecated cause we want user to use `NuxtConfig` over `Configuration`
- It's also not deprecated cause modules extend it

pi0:
> It's also not deprecated cause modules extend it

Indeed. For me `deprecated` means something was possible, it is currently still possible but is not encouraged and will not be possible in future versions :D

pi0 (approved):

PR #7601, opened 2020-06-24
fix(example): "`head` property in component must be a function"

suhrmann:
Fix example "SEO HTML Head"

The codestyle for ``head`` of [nuxt examples: SEO HTML Head](https://nuxtjs.org/examples/seo-html-head/) and [nuxt api: API: The head Method](https://nuxtjs.org/api/pages-head/) differ.
Further on `Nuxt.js @ v2.12.x` I got the obove mentioned error.

### DIFF
--- a/examples/meta-info/pages/index.vue
+++ b/examples/meta-info/pages/index.vue
@@ -1,30 +1,32 @@
 <template>
   <div class="container">
     <h1>Home page 🚀</h1>
-    <NuxtLink to="/about">
-      About page
-    </NuxtLink>
+    <NuxtLink to="/about">About page</NuxtLink>
   </div>
 </template>
 
 <script>
 export default {
-  head: {
-    title: 'Home page 🚀',
-    meta: [
-      { hid: 'description', name: 'description', content: 'Home page description' }
-    ],
-    noscript: [
-      { innerHTML: 'Body No Scripts', body: true }
-    ],
-    script: [
-      { src: '/head.js' },
-      // Supported since 1.0
-      { src: '/body.js', body: true },
-      { src: '/defer.js', defer: '' }
-    ]
+  head() {
+    return {
+      title: "Home page 🚀",
+      meta: [
+        {
+          hid: "description",
+          name: "description",
+          content: "Home page description"
+        }
+      ],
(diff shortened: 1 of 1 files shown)


pimlie:
Thanks for your contribution. `head` only needs to be a function if you use any (reactive) component properties, as this example doesnt use those it should be fine to just use an object.

When exactly did you get this error? After running lint? Or is it a typescript error?

suhrmann:
It was lining, but with an error that prohibited build.
I was just confused that these 2 docs pages show different notation, one causing an error in my code. 

With your explanation I think, I can close this PR.

pimlie:
If you are using eslint-plugin-nuxt, I have reported this behaviour in this issue: https://github.com/nuxt/eslint-plugin-nuxt/issues/93 which was fixed by removing the rule from `nuxt/recommended`. So either you might need to upgrade eslint-plugin-nuxt or you have added the rule locally?

suhrmann:
To check I completely disabled ``@nuxtjs/eslint-module``: 

**Result**
 - For Nuxt itself (without ``eslint``) it does not matter if I use object or method notation. It works!
 - If I enable ``eslint``, the above mentioned error _"`head` [...] must be a function"_ prohibits the build

Here are my packages of the troubling project - latest with ``yarn upgrade``:
```
$ npm list --depth=0
my-project@1.0.0
├── @fortawesome/fontawesome-free@5.13.1
├── @nuxtjs/eslint-config@2.0.2
├── @nuxtjs/eslint-module@1.2.0
├── @nuxtjs/vuetify@1.11.0
├── babel-eslint@10.1.0
├── eslint@6.8.0
├── eslint-plugin-nuxt@0.5.2
└── nuxt@2.13.1
```

-----

I just tested it on a new test-project: Here is the problem fixed:
```
$ npm list --depth=0
TEST@1.0.0 /TEST
├── @nuxtjs/eslint-config@2.0.2
├── @nuxtjs/eslint-module@1.2.0
├── @nuxtjs/vuetify@1.11.2
├── babel-eslint@10.1.0
├── eslint@6.8.0
├── eslint-plugin-nuxt@1.0.0
└── nuxt@2.13.2
```

PR #7603, opened 2020-06-24
fix(webpack): chunk name normalization for catch-all pages and windows

aldarund:
## Types of changes

## Description
For _.vue SplitChunks will generate route starting with dot and it will lead to 404 for this chunks
Dropped window path replace to be in line with Linux, as seems to work fine on win without it

## Checklist:

### DIFF
--- a/packages/utils/src/resolve.js
+++ b/packages/utils/src/resolve.js
@@ -18,7 +18,7 @@ export const wp = function wp (p = '') {
 }
 
 export const wChunk = function wChunk (p = '') {
-  // workaround for SplitChunksPlugin that generate names starting from . for catchAll pages _.vue 
+  // workaround for SplitChunksPlugin that generate names starting from . for catchAll pages _.vue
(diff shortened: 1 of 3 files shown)


pi0:
Seems windows tests need updating: https://dev.azure.com/nuxt/nuxt.js/_build/results?buildId=7762&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=bd05475d-acb5-5619-3ccb-c46842dbc997&l=109

pi0 (approved):

aaronhuisinga:
@aldarund @pi0 This introduced a regression breaking dynamic routing in AWS API Gateway and likely many other services. According to [RFC3986](https://tools.ietf.org/html/rfc3986?#section-3.2.2), square brackets are not valid in URIs, and several services don't allow them. API Gateway will return a 400 with any request containing them, as an example (that we're currently running into).

PR #7604, opened 2020-06-25
fix(utils): flat routes if child routes have absolute paths

kedrzu:
Fixed `flatRoutes` function to work properly if children routes have absolute paths.
This was working fine in 2.12, it did break in 2.13.

## Types of changes

## Description
I am using `nuxt.js` along with `nuxt-18n`. I use SEO-friendly, localized routes.
For this to work, you have to configure pages to have absolute paths.

Here is a snapshot of my config:
image

However, it does not work well:
image

Routes are not properly concatenated. When I look into `router.js` generated file, it looks fine:
image

The reason behind this is, that `flatRoutes` function concatenates parent path and child path, even if child path is absolute.

## Checklist:

### DIFF
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -13,14 +13,19 @@ export const flatRoutes = function flatRoutes (router, fileName = '', routes = [
       if (fileName === '' && r.path === '/') {
         routes.push('/')
       }
+
       return flatRoutes(r.children, fileName + r.path + '/', routes)
     }
     fileName = fileName.replace(/\/+/g, '/')
-    routes.push(
-      (r.path === '' && fileName[fileName.length - 1] === '/'
-        ? fileName.slice(0, -1)
-        : fileName) + r.path
-    )
+
+    // if child path is already absolute, do not make any concatenations
+    if (r.path && r.path.startsWith('/')) {
+      routes.push(r.path)
+    } else if (r.path === '' && fileName[fileName.length - 1] === '/') {
+      routes.push(fileName.slice(0, -1) + r.path)
+    } else {
+      routes.push(fileName + r.path)
+    }
   })
   return routes
 }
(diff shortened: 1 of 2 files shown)


packages/utils/src/route.js
```diff
@@ -13,14 +13,19 @@ export const flatRoutes = function flatRoutes (router, fileName = '', routes = [
...
-    routes.push(
-      (r.path === '' && fileName[fileName.length - 1] === '/'
-        ? fileName.slice(0, -1)
-        : fileName) + r.path
-    )
+
+    // if child path is already absolute, do not make any concatenations
+    if (r.path && r.path.startsWith('/')) {
+      routes.push(r.path)
+    } else if (r.path === '' && fileName[fileName.length - 1] === '/') {
```
kedrzu, line 24: I think plain old `if-else` is more readable than ternary operator with some parentness.

pi0 (approved):

pi0:
Hi @kedrzu. This fix looks good to me btw why you are not using [alias](https://router.vuejs.org/guide/essentials/redirect-and-alias.html#alias)?

kedrzu:
I don't think this would be an alternative.
What I use is a basic solution when using `nuxt-i18n`, described here: https://nuxt-community.github.io/nuxt-i18n/routing.html#custom-paths

I don't think aliasing would be properly recognized by `nuxt-i18n` to display page in proper language. I think in `nuxt-i18n` routes are recognized by their name:
image

Language is attached to the end of the route name. I don't want to mess with that :)

kedrzu:
@pi0 maybe it would be better to cherry pick the commit to 2.x branch?

pi0:
Then it is a nuxt-i18n limitation to support aliases for CEO friendly version. 

Fix looks good and dev branch is on fix mode so all good! Thanks for contribution :heart:

PR #7626, opened 2020-06-27
types: import types for `@nuxt/telemetry` and `@nuxt/components`

danielroe:
Add missing types for new Nuxt modules `@nuxt/telemetry` and `@nuxt/components`.

## Types of changes

## Description
Added two new properties to `NuxtConfig` type.

## Checklist:

### DIFF
--- a/packages/types/config/index.d.ts
+++ b/packages/types/config/index.d.ts
@@ -1,3 +1,5 @@
+import { ComponentsDir } from '@nuxt/components'
+
 import { Transition } from '../app'
 import { NuxtOptionsBuild } from './build'
 import { NuxtOptionsCli } from './cli'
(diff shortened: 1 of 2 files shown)


pi0 (dismissed):

packages/types/config/index.d.ts
```diff
@@ -32,6 +34,7 @@ export interface NuxtOptions extends Configuration {
   buildDir: string
   buildModules: NuxtOptionsModule[]
   cli: NuxtOptionsCli
+  components: boolean | Array<string | ComponentsDir>
```
pi0, line 37: There is also object format support (which may be used later on to provide more options

danielroe: Ideally the options would be completely defined in `@nuxt/components` and just imported in `@nuxt/types`. I will propose a PR to do that if that sounds good to you.

pi0: Indeed. Just published types for both :see_no_evil: Updating deps in a sec.

pi0: #7621

PR #7633, opened 2020-06-29
feat(vue-renderer): add `vue-renderer:spa:prepareContext` hook

harlan-zw:
## Types of changes

## Description

For SPA and static site builds, the generated HTML pages all use the meta options from `nuxt.config.js`. This is a problem for SEO (obviously) and for link sharing where the meta content of the HTML will be parsed to give a preview. While a universal or server-sided rendered app would solve this, I'd rather not have to deal with the complexities and additional build time when there is a fairly simple solution (I only need the meta pre-rendered, nothing else).

What I'm proposing is a hook for vue-meta allowing the app to decide on a more granular level what meta to use for each route. I believe this is in-line with existing core functionality and could open up other use cases.

I would hope one day the vue-meta plugin could just work for this for static / spa built sites, reading the page component meta, but I think this can be a good intermediate solution.

Example: with this code block a user can generate all their static HTML to have individual titles without any effort 

```
hooks: {
    'vue-renderer': {
      spa: {
        meta ({ head, renderContext }) {
          const urlParts = renderContext.url.split('/')
          head.title = startCase(urlParts.pop())
        }
      }
    },
  },
```

## Checklist:

- [x] My change requires a change to the documentation. - **No documentation page for `vue-renderer` hooks that I can see?**
- [ ] I have added tests to cover my changes (if not applicable, please state why) - **Wasn't sure exactly where to put these tests, wasn't a SPA generate unit test from what I could see?**


pi0 (dismissed):
Thanks for PR ❤️ Makes sense as we already also support hooks for `ssr:` rendering override. Adding some docs would be appreciated if you could.

clarkdo (dismissed):

pimlie (approved):
LGTM with the changes by @pi0, thank you both!

PR #7656, opened 2020-07-02
fix(vue-app): check plugin key on `Vue.prototype` instead of `Vue`

simplesmiler:
Fix for an obscure bug that got introduced in https://github.com/nuxt/nuxt.js/commit/e8aca9eb117851047e82e94948ff8b4bcb464b1a. 
Triggered when you `inject` something under a key, that was already used by other library. 

## Types of changes

## Description

Example (using https://github.com/stalniy/casl/tree/master/packages/casl-vue):
```js
// plugins/casl.js
import Vue from 'vue';
import { Ability, abilitiesPlugin } from '@casl/vue';
export default function(context, inject) {
   const ability = new Ability(...);
   Vue.use(abilitiesPlugin, { ability }); // defines $ability and $can on Vue.prototype
   inject('ability', ability); // error: Can not redefine '$ability'
}
```

## Checklist:

### DIFF
--- a/packages/vue-app/template/index.js
+++ b/packages/vue-app/template/index.js
@@ -189,7 +189,7 @@ async function createApp(ssrContext, config = {}) {
     Vue[installKey] = true
     // Call Vue.use() to install the plugin into vm
     Vue.use(() => {
-      if (!Object.prototype.hasOwnProperty.call(Vue, key)) {
+      if (!Object.prototype.hasOwnProperty.call(Vue.prototype, key)) {
         Object.defineProperty(Vue.prototype, key, {
           get () {
             return this.$root.$options[key]


simplesmiler:
Also, it appears that a proper reentrance check is done above via `installKey`. Because of this, maybe we should warn developer instead of silently skipping?

packages/vue-app/template/index.js
```diff
@@ -189,7 +189,7 @@ async function createApp(ssrContext, config = {}) {
     Vue[installKey] = true
     // Call Vue.use() to install the plugin into vm
     Vue.use(() => {
-      if (!Object.prototype.hasOwnProperty.call(Vue, key)) {
+      if (!Object.prototype.hasOwnProperty.call(Vue.prototype, key)) {
```
pi0, line 192: What do you think to keep both checks or simply Vue[key] by using proto chain?

clarkdo (approved):

PR #7639, opened 2020-06-30
 fix(webpack): use compact name for concatenated modules

pi0:
Update: Additional patch: 2302ffa40b2e83775ca9f7d5ab7d63b38b523333

## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

With #7463 we use `[name]` for chunknames for easier debugging of production assets but it introduced two new issues:

- `pages/_.vue` causing to make structures like `pages/.[hash].js` (#7603)
   - Fix #7625 and https://github.com/nuxt/nuxt.js/pull/7603#issuecomment-651216394 
   - Fixed by setting `automaticNameDelimiter` to default (`~`) instead of `.`
- The strange directory structure for concatenated module chunks (https://github.com/nuxt/nuxt.js/pull/7463#issuecomment-650734494)
  - Using better names like `commons/product.id~index~about.[hash].js`
- Use `.modern.js` instead of `modern-` prefix to co-locate chunks

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

PR #7667, opened 2020-07-03
fix(webpack): fallback for empty chunk name

### LINKED ISSUE #7665: since nuxt@2.13.3 some modern files are missing?
Since nuxt 2.13.3 it seems our app is totally broken in modern mode.
e.g. https://cdn-test.hokify.com/pwa/company/_assets/commons/.2714f92.modern.js <-- see the commons/ followed by a dot immediately.

Browser console log:
image

Nuxt 2.13.2 is NOT affected by this bug as far as I can tell, as the only change from yesterday to today is the nuxt version.
Maybe it's related to #7639 ?

In another app I use a custom "filename" option for the assets, It seems this one is not affecetd by this bug, but still verifying it.
```js
filenames: {
			app: ({ isModern, isDev }) =>
				`${!isModern ? 'legacy-' : ''}${!isDev ? '[contenthash]' : '[name]'}-app.js`,
			chunk: ({ isModern, isDev }) =>
				`${!isModern ? 'legacy-' : ''}${!isDev ? '[contenthash]-' : ''}[name].js`
		}
```

### Versions

- nuxt: nuxt@2.13.3
- node: v12.18.0

### Reproduction

Couldn't find time yet, will deliver a reproducable example later if needed.

simllll:
fixes #7665

## Types of changes

## Description
if all chunks have no name, it results to an empty compactName. Due to #7639 this results into file paths like /common/.xxx.js. As files with a dot at the beginnign are not served, this is a problem ;-)  

## Checklist:

### DIFF
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -74,24 +74,29 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
 
     if (!this.dev && cacheGroups.default && cacheGroups.default.name === undefined) {
       cacheGroups.default.name = (_module, chunks) => {
-        // Use compact name for concatinated modules
-        let compactName = chunks
-          .filter(c => c.name)
+        // Map chunks to names
+        const names = chunks
           .map(c => c.name)
+          .filter(Boolean)
           .sort()
           .map(name => name.replace(/[/\\]/g, '.').replace(/_/g, '').replace('pages.', ''))
-          .join('~')
 
         // Fixes https://github.com/nuxt/nuxt.js/issues/7665
         // TODO: We need a reproduction for this case (test/fixtures/shared-chunk)
-        if (!compactName) {
-          compactName = 'default'
+        if (!names.length) {
+          return 'commons/default'
         }
 
+        // Single chunk is not common
+        if (names.length === 1) {
+          return names[0]
+        }
+
+        // Use compact name for concatinated modules
+        let compactName = names.join('~')
         if (compactName.length > 32) {
           compactName = hash(compactName)
         }
-
         return 'commons/' + compactName
       }
     }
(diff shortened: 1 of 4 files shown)


packages/webpack/src/config/client.js
```diff
@@ -86,6 +82,12 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
           .map(name => name.replace(/[/\\]/g, '.').replace(/_/g, '').replace('pages.', ''))
           .join('~')
 
+        // Fixes https://github.com/nuxt/nuxt.js/issues/7665
+        // TODO: We need a reproduction for this case (test/fixtures/shared-chunk)
```
pi0, line 86: @simllll @aldarund I've added a TODO item as couldn't reproduce the issue. Would be awesome if we can find a reproduction for fixtures. And sorry for issues that happened due to this name change...

pi0 (dismissed):

packages/webpack/src/config/client.js
```diff
@@ -74,10 +74,6 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
 
     if (!this.dev && cacheGroups.default && cacheGroups.default.name === undefined) {
       cacheGroups.default.name = (_module, chunks) => {
-        // Use default name for single chunks
-        if (chunks.length === 1) {
-          return chunks[0].name || ''
-        }
```
simllll, line 80: This might change behavior, because it will now be returned as commons/name instead of just the name without the folder prefix for chunks with size 1. Is this intended?

pi0: 476af4304a6a6624d71b14d27183e5a7a9f59ef2

simllll:
Lgtm

pi0:
Thanks @simllll. Would you please confirm it working using `nuxt-edge` before doing release? (should be published in few minutes)

simllll:
@pi0 just tested it, I can confirm the issue is fixed in `nuxt-edge`. looks good, thanks!

packages/webpack/src/config/client.js
husayt: Nuxt 2.13.13 was broken trying to load chunks like `.xxx.js`. This has fixed it. Thanks @simllll and @pi0

PR #7705, opened 2020-07-12
examples(programmatic): fix package import

edbzn:
Fix programmatic example package import.


pi0:
Hi. This is intentional! You should put nuxt-start in dependencies and nuxt in devDependencies to avoid installing build dependencies (like webpack) for production

PR #7712, opened 2020-07-13
feat: static target DX improvements

pi0:
## Types of changes

## Description

- With 2.13 we were making errors to force using `nuxt export` and `nuxt serve` for static target. Now instead users can keep using `nuxt generate` and `nuxt start`
- `export:*` hooks deprecated to avoid using different names (generate will kept for nuxt3 too)
- `exports` option is deprecated
- Move smart build from [nuxt/static](https://github.com/nuxt/static) to the core for `nuxt generate` for static target only (non breaking change)

## Checklist:


packages/generator/src/generator.js
```diff
@@ -37,7 +37,7 @@ export default class Generator {
 
     // Payloads for full static
     if (this.isFullStatic) {
-      consola.info('Full static mode activated')
+      consola.info('Generating with full static mode')
```
pimlie, line 40: Could this whole if be moved into a separate function (including the if)? For nuxt-generate-cluster Im not calling this method but instead calli each function that this method calls either from the master or worker thread. As this if is not abstracted we'd need to copy this implementation ourselves, see eg this pr: https://github.com/nuxt-community/nuxt-generate-cluster/pull/32

Thanks :)

pi0: @pimlie I'm not sure related to this PR but would you mind directly committing your idea on this branch or new PR?

pimlie: @pi0 Done, it should be fine to just move this if into `initiate`. We call this method in nuxt-generate-cluster already and it would need to be set for both master as workers.

pi0: Thanks, I see :) Is it also ok with  d8d6045 and 51d693d?

pimlie: Yep, should be ok too. Thanks :)

atinux (approved):

Atinux:
I need to update articles & docs now :)

PR #7718, opened 2020-07-14
fix(vue-app): handle fallback on generated page

atinux:
## Types of changes

## Description

This is an advances use case mentioned by @Akryum when you want a static fallback (not SPA) using a `pages/404.vue` and a Netlify redirect. Hitting an unknown page will fetch the wrong payload, this is why I use `NUXT.routePath` to fetch the generated page to fetch the payload.

@pi0 I am wondering if we should no directly remove `_app.context.route.path` directly in `fetchPayload(NUXT.routePath || _app.context.route.path)`

## Checklist:


packages/vue-app/template/client.js
```diff
@@ -827,7 +827,7 @@ async function mountApp (__app) {
   // Load page chunk
   if (!NUXT.data && NUXT.serverRendered) {
     try {
-      const payload = await _app.fetchPayload(_app.context.route.path)
+      const payload = await _app.fetchPayload(NUXT.routePath || _app.context.route.path)
```
pi0, line 830: We may add `NUXT.routePath` to [spa renderer context](https://github.com/nuxt/nuxt.js/blob/f585e46829d7e704b98ca0492350e03d9fde1ce5/packages/vue-renderer/src/renderers/spa.js#L157) and always use it instead of `_app.context.route.path` then we may be able to remove `context.route.path` (just i think we may to do sanetization similar to `vue-router` on baseURL and trailing slash (for `/`)

pi0 (commented):
Also, is it also possible to add a redirect scenario to one of the test fixtures?

Atinux:
Will do :)

packages/vue-app/template/client.js
Atinux: Well here we are in the condition if `NUXT.serverRendered` so it won't change to have it in the SPA renderer right?

Atinux:
One thing also, because `NUXT.routePath` and `_app.context.route.path` are different, it will do a full render on client-side (like SPA fallback) because of https://github.com/nuxt/nuxt.js/blob/fix/fallback-with-404-page/packages/vue-app/template/client.js#L892

PR #7752, opened 2020-07-21
fix(vue-renderer): read `target` from class instead of `serverContext`

timrchavez:
The `renderRoute` method assumes that `this.serverContext.nuxt.options` is a valid path, but that does not seem like a good assumption AFAICT. We could defensively check for its existence, but I think the correct thing to do here, is to use `this.options` instead.

I'm not sure how this was working before, but I hit this particular problem while trying to use `nuxt-lambda` which creates a minimal `serverContext` that doesn't include `serverContext.nuxt.options`. See: https://github.com/pimlie/nuxt-lambda/blob/master/src/renderer.js#L8-L21 -- I freely admit that the bug could lie there. I tried to find some authoratative documentation on `serverContext` but was unsuccessful.

## Types of changes

## Description
Presently `renderRouter` assumes that `serverContext` has a `nuxt.options.target` field. This seems at odds with other areas within the code that need to access `target` and get it directly from `serverContext.options`. To remain consistent and remove (what I think is) an errant path, we should update the assignment of `serverContext.target` using `this.options.target` instead of `this.serverContext.nuxt.options.target`

## Checklist:
I tried to run the tests locally but failed hard. Could not get jest to work.

### DIFF
--- a/packages/vue-renderer/src/renderer.js
+++ b/packages/vue-renderer/src/renderer.js
@@ -276,7 +276,7 @@ export default class VueRenderer {
     // Add url to the renderContext
     renderContext.url = url
     // Add target to the renderContext
-    renderContext.target = this.serverContext.options.target
+    renderContext.target = this.options.target
 
     const { req = {}, res = {} } = renderContext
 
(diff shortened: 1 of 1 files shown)


timrchavez:
I had to amend the author on my commit as I am not representing my employer with this change :)

pi0:
Hi, @timrchavez thanks for the PR and well-detailed explanations ❤️  It totally makes sense to merge this PR as both a fix and simplification refactor. BTW I also opened pimlie/nuxt-lambda#6 to fix the root cause.

PR #7754, opened 2020-07-21
feat(builder): support async `extendRoutes`

MartinN3:
Current behaviour
its impossible to extend routes with APIs

Expected behaviour
Being able to await requests to server that are required for extendingRoutes

## Types of changes

## Description

With this simple change, we are able to extend our nuxt router with async data without need to create custom router.

In documentation is something like this:
```javascript
router: {
  extendRoutes(routes, resolve) {
    routes.push({
      name: 'PageHierarchyPage',
      path: '/test/test/',
      props: {
        id: 12,
      },
      component: resolve(__dirname, 'pages-extended/PageHierarchyPage/_.vue'),
    });
    sortRoutes(routes);
  }
}
```

But we might need something like this:
```javascript
router: {
  extendRoutes: async (routes, resolve) => {
   const dynamicRoutes = await getDynamicRoutes();
   const dynamicRoutesTransformed = dynamicRoutes.map(item => ({
     ...item,
     component: resolve(__dirname, item.component)
   }));
   routes.push(...dynamicRoutesTransformed);
   sortRoutes(routes);

   return routes;
  }
}
```

## Checklist:

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -414,13 +414,13 @@ export default class Builder {
     // router.extendRoutes method
     if (typeof this.options.router.extendRoutes === 'function') {
       // let the user extend the routes
-      const extendedRoutes = this.options.router.extendRoutes(
+      const extendedRoutes = await this.options.router.extendRoutes(
         templateVars.router.routes,
         r
       )
       // Only overwrite routes when something is returned for backwards compatibility
       if (extendedRoutes !== undefined) {
-        templateVars.router.routes = await extendedRoutes
+        templateVars.router.routes = extendedRoutes
       }
     }
 
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -420,7 +420,7 @@ export default class Builder {
       )
       // Only overwrite routes when something is returned for backwards compatibility
       if (extendedRoutes !== undefined) {
-        templateVars.router.routes = extendedRoutes
+        templateVars.router.routes = await extendedRoutes
       }
     }
 


MartinN3:
Can this idea be even considered good or the whole thing should be scraped? Thanks for insights.

btw:
Some unit test wasn't successful?  I can't really see which one and what failed tho.
I'll have to read details of semantic pull requests.

packages/builder/src/builder.js
```diff
@@ -420,7 +420,7 @@ export default class Builder {
       )
       // Only overwrite routes when something is returned for backwards compatibility
       if (extendedRoutes !== undefined) {
-        templateVars.router.routes = extendedRoutes
+        templateVars.router.routes = await extendedRoutes
```
pi0, line 423: I think would be better to put await [here](https://github.com/nuxt/nuxt.js/blob/e10b9a72327d5c85ca93f64da490d666cd55f92b/packages/builder/src/builder.js#L417) because result of the promise can be undefined too

MartinN3: Moved

gianpesto:
Thank you for the merge,
i am having troubles wrapping my head around a similiar problem:

i'd like to extend Routes as well after fetching data form an API, but during runtime.

the scenario is the following:

1) A User navigates to the app (possibly with a specific locale 'en-us')
2) User has to login using nuxt auth module
3) routes are being fetched from the Server (can be different for each user, depending on permissions, locale, ecc...)
4) extend Routes

i am now having one dynamic route that checks the requested route and shoves it into a dynamic component, but somehow it feels wrong to do it that way 🤮

any help would be appreciated

pi0:
Hey @gianpesto maybe opening a [dicsussion](https://github.com/nuxt/nuxt.js/discussions) with more details how are you currently doing stuff?

gianpesto:
[Opened one ☺️](https://github.com/nuxt/nuxt.js/discussions/7941)

Issue #2603, opened 2018-01-15
How to add component to root instance (outside main router-view)


chasebank:
I have need for adding a component in the main instance outside of the main `<router-view>`. The use-case is an element (global navigation, in my case) that needs to **not** re-load on route/layout change due to styling requirements.

A 'normal' Vue implementation would look like

```
<div id="app">
  <Navigation/>
  <router-view></router-view>
</div>
```

It seems like issue #1 (super ironic...) and [this video](https://www.youtube.com/watch?v=wBhia7uBxDA) are what I'm looking for, although that doesn't appear to be working in 1.0

I've tried adding `app.vue` in layouts (per the video description), with

```
<template>
  <nuxt-container>
    <h2>Test</h2>
    <nuxt/>
  </nuxt-container>
</template>
```
But that seems to get ignored altogether.

Thanks for any help!

Oliboy50:
Hello,

I think what you're looking for are layouts, even if it seems that you already tried.

Have you tried to use the simple layouts example here https://nuxtjs.org/examples/layouts ?

I confirm that layouts worked at least using `v1.0.0-RC11`

chasebank:
Hi, thanks for the reply but I don't think normal layouts will work here, unless there's some way to extend the root instance in them (like issue #1 suggests).

As I said, I need this component to not reload when layouts change.

As far as I can tell, everything inside a layout component re-loads, including anything outside the `<nuxt/>` element.

So, extending the example, if I have an Index page with default layout:
```
<template>
  <div>
    <Navigation/>
    <img src="logo.png"/>
    <nuxt/>
  </div>
</template>
```
and About page with dark layout:
```
<template>
  <div class="dark">
    <Navigation/>
    <nuxt/>
  </div>
</template>
```
Correct  me if I'm wrong, but that Navigation component is going to get remounted when the route changes.  I need it to be outside of the routes.

Al-Rozhkov:
I think your Navigation component will stay, if you use `<nuxt-link to="/path">Link</nuxt-link>` for your navigation links.

chasebank:
Hey, thanks for the suggestion.

Here's a live example that shows the component getting re-mounted, if you check console, even when using path links.

The use case for what I'm talking about is the first example found in the [vue-router Getting Started docs](https://router.vuejs.org/en/essentials/getting-started.html). This seems to be a pretty standard router setup.

```html
<div id="app">
  <h1>Hello App!</h1>
  <p>
[...]

PR #7768, opened 2020-07-24
feat: support root components outside layout

### LINKED ISSUE #7753: Components into root element
Initially, i just needed simple toast, i'm thinking how it would be working in another layouts. The first thing came up with the idea to use `Vuex` and place component(list with toasts) in all layout. The idea in itself is not very good, it was also aggravated by the fact that when changing the `layout`, a rerender takes place. After this i check `vue-toasted` source, [link](https://github.com/ankurk91/vue-toast-notification/blob/master/src/js/api.js#L16) nice idea, i try use it, but some components in my toast use `vuex` and i can`t remove it

After this attempts i found #2603 issue, this my problem yeah... without solution

Please add some api for use global(between layouts and outside router-view) components it be usefull for modals, toasts etc

image

P.S. Why `nuxt.config.js` `build.template` not mentioned anywhere?) Found it late, but it's too complicated for me)

Kolobok12309:
Now it's impossible to add component which not rerender when change layout

## Types of changes

## Description
With this prop in config we can add components under root nuxt instance like this
image

Resolves: #7753, #2603
`nuxt.config.js`
### Array variant
```js
  ...
  rootComponents: [
    '@/components/global.vue',
  ],
  ...
```
### Object variant
```js
  ...
  rootComponents: {
    global: '@/components/global.vue',
  },
  ...
```

And use it in components
```js
{
  methods: {
    method() {
      const { global } = this.$nuxt.$rootComponents;
      if (global) {
        global.someMethod();
      }
    }
  }
}
```

## Checklist:

- [x] I have updated the documentation accordingly. (PR: nuxt/docs#2118)

Add fixture, update config snapshot and fix 2 tests in `cli`


pi0:
Hi @Kolobok12309 thanks for PR ❤️ Actually with nuxt3 it is possible to define top-level components via `app.vue`. Since nuxt2 is feature freeze, closing but can see benefit of avoiding re-render by layout changes. If it is making critical issues, we can think of a workaround for #7753

Kolobok12309:
Hi @pi0, ok, when released nuxt3 with vue3 try feature.
Is not critical moment, so we can wait nuxt3.
Mb someone else will ask for this feature :D

9-lives:
i just mount manually that component in middleware once.

lopugit:
Wait, are you telling me that nuxt re-renders all layout components on route change? Isn't that the whole point of layouts, that they don't re-render, only the route specific data in <nuxt> should re-render?

damn 😮

PR #7775, opened 2020-07-26
fix(cli): don't use absolute path of `.cache/nuxt` as transpilation pattern

### LINKED ISSUE #7774: `nuxt-edge` doesn't respect targets in generation
### Versions

- nuxt: v2.14.0-26582020.6ab65fbf
- node: v12.18.1

### Reproduction

https://github.com/danielroe/nuxt-edge-preset

### Steps to reproduce
1. `yarn nuxt generate`
2. `cd dist && grep -RH '=>'`

### What is expected?

No arrow functions should be present in `dist/_nuxt` JS files.

Non-transpiled content is present - and if terser is disabled (as a debugging step) this can be seen in great detail.

danielroe:
closes #7774

If `static.cacheDir` isn't set, this won't add the fully resolved cache directory but instead just `.cache/nuxt` as a transpilation pattern.

## Type of change

## Description
Resolves #7774 

## Checklist


pi0 (dismissed):

pi0 (commented):
If absolute path is problematic, shall we ensure it is always relative pattern or remove static.cacheDir support at all or add a warning for full path?

danielroe:
> If absolute path is problematic, shall we ensure it is always relative pattern or remove static.cacheDir support at all or add a warning for full path?

@pi0 I don't think a relative path will work either (e.g. with a leading `./`) because it's being tested against the file pattern. I think let's remove support for adding transpile for custom staticPath, and leave it (with warning) for the person who uses the custom path.

One approach: 4a3ba0d5a933e57e15ccab087099411fb7f8fe02

packages/cli/src/utils/generate.js
```diff
@@ -110,7 +110,21 @@ async function getNuxt (args, cmd) {
   config.buildDir = (config.static && config.static.cacheDir) || path.resolve(config.rootDir, 'node_modules/.cache/nuxt')
   config.build = config.build || {}
   config.build.transpile = config.build.transpile || []
-  config.build.transpile.push(config.buildDir)
+  if (config.static && config.static.cacheDir) {
+    if (
+      !config.build.transpile.some(path =>
```
pi0, line 115: Transpile also supports functions ([example](https://github.com/nuxt/http/blob/v0.5.0/lib/module.js#L106). Simply puting `.cache/nuxt` without warning could be enough :)

danielroe: Okay - and we can always specify in the docs if/when there are docs for this option :stuck_out_tongue:

pi0 (approved):

PR #7776, opened 2020-07-26
fix(generator): avoid duplicate slashes for routes ending with hash

### LINKED ISSUE #7772: export error with hash link, baseurl and trailingSlash
### Versions

- nuxt: 2.13.3
- node: v12.18.1

### Reproduction

https://github.com/tobigumo/nuxt-hashlink-bug

`nuxt.config.js`
```
  router: {
    base: '/',
    trailingSlash: true
  }
```

`pages/bar.vue`
```
      <div class="links">
        <a
          href="/bar/#target"
          class="button--green"
        >
          Hash Link
        </a>
      </div>
```

**no problem**
- `href="#target"` with no base option
- `href="/bar#target"` with no trailingSlash option

**There is a problem**
- `href="/bar/#target"` with trailingSlash and base option
  - `ERROR  Error generating route "/bar//": This page could not be found`

### Steps to reproduce

`nuxt build && nuxt export`

### What is Expected?

Static export completes successfully

An error occurs in `nuxt export`
```
ℹ Generating output directory: dist/                                                              11:54:03
ℹ Full static mode activated                                                                      11:54:03
ℹ Generating pages                                                                                11:54:03
✔ Generated route "/"                                                                             11:54:03
✔ Generated route "/bar/"                                                                         11:54:03

 ERROR  Error generating route "/bar//": This page could not be found                             11:54:03

✔ Client-side fallback created: 200.html                                                          11:54:03
ℹ Ready to run nuxt serve or deploy dist/ directory   
```
and inside the title tag become "This page could not be found" in exported file `dist/bar/index.html`

farnabaz:
Using links like `/foo/#bar` cause error in generating static pages. This happens because crawler adds a trailing slash to every route. We need to make sure the is no duplicate slashes in route.

close #7772

## Types of changes

## Checklist:

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -303,6 +303,7 @@ export default class Generator {
             .trim()
 
           const route = decodeURI(sanitizedHref + possibleTrailingSlash)
+            .replace(/\/+/g, '/')
 
           if (route.startsWith('/') && !path.extname(route) && this.shouldGenerateRoute(route) && !this.generatedRoutes.has(route)) {
             this.generatedRoutes.add(route)
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -297,13 +297,12 @@ export default class Generator {
         parse(html).querySelectorAll('a').map((el) => {
           const sanitizedHref = (el.getAttribute('href') || '')
             .replace(this.options.router.base, '/')
-            .replace(/\/+$/, '')
             .split('?')[0]
             .split('#')[0]
+            .replace(/\/+$/, '')
             .trim()
 
           const route = decodeURI(sanitizedHref + possibleTrailingSlash)
-            .replace(/\/+/g, '/')
 
           if (route.startsWith('/') && !path.extname(route) && this.shouldGenerateRoute(route) && !this.generatedRoutes.has(route)) {
             this.generatedRoutes.add(route)
--- a/test/fixtures/full-static/layouts/default.vue
+++ b/test/fixtures/full-static/layouts/default.vue
@@ -10,7 +10,7 @@
       <NLink to="/pagination/1">
         Pagination
       </NLink>
-      <NLink to="/dynamic/foo bar#hash">
+      <NLink to="/dynamic/foo bar/#hash">
         Dynamic route 1
       </NLink>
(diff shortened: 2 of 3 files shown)


pi0 (dismissed):
You rock @farnabaz 👍

pi0:
Updating `full-static` fixture with `trailingSlash` enabled, it seems we have other problems with it as well:

> ERROR  Error generating route "/pagination/1/2/": This page could not be found
> ERROR  Error generating route "/pagination/1/0/": This page could not be found    

Also client-side routing is broken. Maybe we should make a warning or force disable trailing slash for static target to avoid this issues?

farnabaz:
> Maybe we should make a warning or force disable trailing slash for static target to avoid this issues?  

This isn't about static target, Using `trailingSlash` and relative urls at the same time will cause the problem. I believe we should warn users about the behavior of relative urls with and without tailing slash.


If you look at `full-static/pages/pagination/_i.vue` you will see that links are relative and with `trailingSlash` enabled it always generates wrong url. It is safer to use absolute paths when `trailingSlash` is enable.
```
<NLink v-if="i>0" :to="`./${i-1}`">
      Previous
</NLink>
```

pi0:
> I believe we should warn users about the behavior of relative urls with and without tailing slash.

Agree. Anyway, this is a bug  :D Was talking with @Atinux we probably going to remove trailingSlash support for nuxt3 since it is adding more and more edge-cases.

papertokyo:
@pio what will removing trailingSlash support mean for Netlify users? We're currently using that config setting because Netlify rewrites urls with a slash and it messes up SEO/loading performance with superfluous redirects otherwise. (At least, that's my understanding.)

mornir:
Yeah, I'm also using the trailing slash because of Netlify:
https://dev.to/mornir/nuxt-netlify-and-the-trailing-slash-3gge

@papertokyo it's possible to disable the redirection: https://github.com/gatsbyjs/gatsby/issues/15317#issuecomment-530048373, but it's not possible to have Netlify automatically redirect routes that end with a trailing slash to their non trailing slash equivalent (but the opposite is possible: non trailing slash --> trailing slash).

TorbjornHoltmon:
A very annoying limitation from netlify...
I have had to go all-in on trailing slashes too.

PR #7780, opened 2020-07-26
types: fix generate types

kevinmarrec:
## Types of changes

## Description
Fix types and setup missing testing of types on workflow `test` (was only present on `nightly`)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -136,6 +136,32 @@ jobs:
           flags: unittests
     needs: setup
 
+  test-types:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -136,6 +136,32 @@ jobs:
...
+      - name: checkout
+        uses: actions/checkout@master
+
+      - name: restore workspace cache
+        uses: actions/cache@v1
+        with:
+          path: ${{ github.workspace }}
+          key: ${{ matrix.os }}-node-v${{ matrix.node }}-nuxt-${{ github.sha }}
+
+      - name: test types
```
pi0, line 161: What do you think to simply add step to lint workflow to avoid extra setup?

kevinmarrec: done :)

PR #7784, opened 2020-07-27
fix(config): allow serving compiled files starting with "."

matthieusieben:
## Types of changes

## Description

From a nuxt page file:
```vue
<template>
  <div><foo-bar v-if="foo" /></div>
</template>

<script>
export default {
  components: {
    FooBar: () => import('~/components/foo-bar.vue'),
  },
  data: () => ({ foo: false }),
  // ...
}
</script>
```

Will generate a chunk with the following name:
`.nuxt/dist/client/commons/.b416c4c.js`

Which will be ignored by `serve-static`, resulting in a 404 when the browser tries to load `/_nuxt/commons/.b416c4c.js`.

## Checklist:

### DIFF
--- a/packages/config/src/config/render.js
+++ b/packages/config/src/config/render.js
@@ -26,6 +26,8 @@ export default () => ({
   },
   csp: false,
   dist: {
+    // Webpack might generate files starting with a dot
+    dotfiles: 'allow',
     // Don't serve index.html template
     index: false,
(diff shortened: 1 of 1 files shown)


pi0:
Hi @matthieusieben Actually this has to be fixed with #7667 since we are manually generating chunk names (you can try with `nuxt-edge`)

PR #7785, opened 2020-07-27
feat(webpack): auto detect core-js version

farnabaz:
Provide easy way to change `core-js` version and auto detect `core-js` installed version

## Types of changes

## Checklist:


packages/webpack/src/config/base.js
```diff
@@ -96,6 +97,8 @@ export default class WebpackBaseConfig {
 
   getBabelOptions () {
     const envName = this.name
+    const { corejs = 'auto' } = this.buildContext.buildOptions
```
pi0, line 100: We should add defaults to `@nuxt/config/src/config/build.json` (also update types with `yarn jest packages/config -u`)

packages/webpack/src/config/base.js
```diff
@@ -96,6 +97,8 @@ export default class WebpackBaseConfig {
...
+    const { rootDir } = this.buildContext.options
```
pi0, line 101: Refactor: we may destructure both in single line :)

clarkdo (dismissed):

packages/webpack/src/config/base.js
```diff
@@ -113,8 +115,11 @@ export default class WebpackBaseConfig {
         }
       )
     }
-
-    const defaultPreset = [require.resolve('@nuxt/babel-preset-app'), {}]
+    const defaultPreset = [require.resolve('@nuxt/babel-preset-app'), {
+      corejs: {
+        version: corejs === 'auto' ? createRequire(rootDir)('core-js').version[0] : corejs
```
pi0, line 120: What do you think to compute `version` outside (L118) and also do a validation to ensure it is either `"2"` or `"3"`?

farnabaz: What we do on invalid values? choose default value or raise an error?

pi0: A warn could be enough :)

https://babeljs.io/docs/en/babel-preset-env#corejs

Added here: https://github.com/nuxt/nuxt.js/pull/7785/commits/4424e9c61556b5005f6955d619ddc99c00349ba5#diff-d8071aa99dfe30c3c1c6135dfd903334R129

farnabaz: 👍 ❤️

pi0 (approved):

pi0:
@farnabaz Would be nice also adding to the docs

rchl:
a) I don't see types being updated
b) It seems like the code is assigning strings rather than numbers to `corejs.version`. Won't this be a problem for babel-preset-env?

packages/webpack/src/config/base.js
```diff
@@ -114,7 +116,26 @@ export default class WebpackBaseConfig {
...
+    let corejsVersion = corejs
+    if (corejsVersion === 'auto') {
+      try {
+        corejsVersion = createRequire(rootDir)('core-js').version.split('.')[0]
+      } catch (_err) {
+        corejsVersion = '2'
+      }
+    }
+
+    if (corejsVersion !== '2' && corejsVersion !== '3') {
```
rchl, line 129: Comparing with strings but user is supposed to provide 2 or 3 as numbers?

pi0:
@rchl It is normalized and passed as a string [src](https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/normalize-options.js#L242) 

Would be nice if can open a PR to support number and also add missing types :)

PR #7787, opened 2020-07-27
fix(webpack): ensure loaders are first searched from `@nuxt/webpack`

matthieusieben:
## Types of changes

## Description

When nuxt is used as part of a mono repo, NPM & Yarn may install `@nuxt/webpack`'s dependencies in `@nuxt/webpack/node_modules` instead of the project's root folder. When that is the case, Webpack is not able to find these loaders.

## Checklist:


packages/webpack/src/config/base.js
```diff
@@ -195,8 +195,12 @@ export default class WebpackBaseConfig {
   }
 
   resolve () {
-    // Prioritize nested node_modules in webpack search path (#2558)
-    const webpackModulesDir = ['node_modules'].concat(this.buildContext.options.modulesDir)
+    const webpackModulesDir = [
+      // Prioritize nested node_modules in webpack search path (#2558)
+      'node_modules',
+      // Ensure that loaders defined in @nuxt/webpack's package.json can be found
+      path.join(path.dirname(require.resolve('@nuxt/webpack/package.json')), 'node_modules')
```
pi0, line 202: - As this core is already inside `@nuxt/webpack/dist`, we can simply use `path.resolve(__dirname, '../node_modules)`
- I think would be better adding it **after** modulesDir as a fallback

pi0 (commented):
Thanks for the PR @matthieusieben. It would be nice if you can provide a reproduction for bug-fixes/reports too so we can ensure this fix works

matthieusieben:
Reproduce like so:

```
- test/package.json
{
  "name": "test",
  "private": true,
  "license": "MIT",
  "workspaces": ["packages/*"],
  "dependencies": {
    "css-loader": "^4.0.0"
  }
}

- test/packages/a/package.json
{
  "name": "a",
  "version": "1.0.0",
  "license": "MIT",
  "dependencies": {
    "nuxt": "^2.13.3"
  }
}
```

Run `yarn install` in the root dir

```
 find . -name css-loader -type d
./node_modules/css-loader // v4.x will break builds because Nuxt expects v3.x
./node_modules/@nuxt/webpack/node_modules/css-loader
```

The "right" `css-loader` is the one in `node_modules/@nuxt/webpack/node_modules`, which is also why the module should be *before* `modulesDir`.

packages/webpack/src/config/base.js
matthieusieben: I used `path.join(path.dirname())` so that this line of code does not depend on how/where the JS file is built

matthieusieben: I see, this line break the tests... I will change accordingly

pi0:
@matthieusieben I made a reproduction accordingly: https://github.com/pi0/nuxt-7787

Issue is reproducible but this change won't fix it! We may have to resolve 'css-loader' like this:

```js
    const cssLoader = { loader: require.resolve('css-loader'), options };
```

matthieusieben:
> @matthieusieben I made a reproduction accordingly: https://github.com/pi0/nuxt-7787
> 
> Issue is reproducible but this change won't fix it! We may have to resolve 'css-loader' like this:
> 
> ```js
>     const cssLoader = { loader: require.resolve('css-loader'), options };
> ```

This was my first approach but there are two issues with it:
- It break backward compatibility of code doing: `urlLoader = loader.use.find(use => use.loader === 'url-loader')`
- Not all the loaders referenced in the code are present in the package.json file.

pi0:
Correct. We cannot do this but is the only possible fix. So probably you should avoid specify `css-loader` in monorepo dependencies (but use it for any package that needs it)

pi0:
@matthieusieben With 93753d3473400bd27b2f604410011b4ae3d090e5 i can confirm it was fixing issue :)

matthieusieben:
> So probably you should avoid specify `css-loader` in monorepo dependencies (but use it for any package that needs it)

Have a look at https://github.com/matthieusieben/nuxt-7787

```
$ find . -name url-loader -type d
./node_modules/url-loader
./node_modules/@nuxt/webpack/node_modules/url-loader
$ find . -name css-loader -type d
./node_modules/css-loader
./packages/ext/node_modules/css-loader
```

matthieusieben:
> @matthieusieben With [93753d3](https://github.com/nuxt/nuxt.js/commit/93753d3473400bd27b2f604410011b4ae3d090e5) i can confirm it was fixing issue :)

Lpook great!

pi0:
Failing windows CI is not related to this PR

PR #7789, opened 2020-07-27
fix: add types for `build.corejs` option and use number

rchl:
Match the documentation https://github.com/nuxt/nuxtjs.org/commit/9301152105f89255813a46c1d5c00286c731df0c

## Types of changes

## Description

A follow-up to comments raised in https://github.com/nuxt/nuxt.js/pull/7785
Correct the check that checks user-provided value for `build.corejs`.
It now matches documentation at https://github.com/nuxt/nuxtjs.org/commit/9301152105f89255813a46c1d5c00286c731df0c

## Checklist:

### DIFF
--- a/packages/types/config/build.d.ts
+++ b/packages/types/config/build.d.ts
@@ -137,6 +137,7 @@ export interface NuxtOptionsBuild {
   analyze?: BundleAnalyzerPlugin.Options | boolean
   babel?: NuxtBabelOptions
   cache?: boolean
+  corejs?: 'auto' | 2 | 3
   crossorigin?: string
   cssSourceMap?: boolean
   devMiddleware?: WebpackDevMiddlewareOptions
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -120,15 +120,15 @@ export default class WebpackBaseConfig {
     let corejsVersion = corejs
     if (corejsVersion === 'auto') {
       try {
-        corejsVersion = createRequire(rootDir)('core-js/package.json').version.split('.')[0]
+        corejsVersion = Number.parseInt(createRequire(rootDir)('core-js/package.json').version.split('.')[0])
       } catch (_err) {
-        corejsVersion = '2'
+        corejsVersion = 2
       }
     }
 
-    if (corejsVersion !== '2' && corejsVersion !== '3') {
-      consola.warn(`Invalid corejs version ${JSON.stringify(corejsVersion)}! Please set "build.corejs" to either "2" or "3".`)
-      corejsVersion = '2'
+    if (![2, 3].includes(corejsVersion)) {
+      consola.warn(`Invalid corejs version ${corejsVersion}! Please set "build.corejs" to either "auto", 2 or 3.`)
+      corejsVersion = 2
     }
 
     const defaultPreset = [require.resolve('@nuxt/babel-preset-app'), {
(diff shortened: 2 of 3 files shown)


rchl:
@pi0 @farnabaz

packages/webpack/src/config/base.js
```diff
@@ -120,15 +120,15 @@ export default class WebpackBaseConfig {
...
-        corejsVersion = '2'
+        corejsVersion = 2
       }
     }
 
-    if (corejsVersion !== '2' && corejsVersion !== '3') {
-      consola.warn(`Invalid corejs version ${JSON.stringify(corejsVersion)}! Please set "build.corejs" to either "2" or "3".`)
-      corejsVersion = '2'
+    if (corejsVersion !== 2 && corejsVersion !== 3) {
+      consola.warn(`Invalid corejs version ${corejsVersion}! Please set "build.corejs" to either "auto", 2 or 3.`)
```
rchl, line 130: I've removed `JSON.stringify` here. Don't use the point of using it. The value should be literal anyway, if it shows as "object" then it's clear enough that it's not one of the allowed values.

pi0: Initial version idea was to only support string format thus using JSON.stringify to show user type. Looks good change now as we support both :)

packages/types/config/build.d.ts
```diff
@@ -137,6 +137,7 @@ export interface NuxtOptionsBuild {
   analyze?: BundleAnalyzerPlugin.Options | boolean
   babel?: NuxtBabelOptions
   cache?: boolean
+  corejs?: 'auto' | 2 | 3
```
pi0, line 140: We also support string '2' | '3' :)

rchl: not according to documentation, see linked commit

packages/webpack/src/config/base.js
rchl: In a template literal, it would work the same regardless if string or number so I don't understand what you mean.

packages/types/config/build.d.ts
pi0: :/  @rchl docs PR was wrong at first place to mention number (reverted by 5e8933417fe809bec46043cce00ed2e73173fd8b to match 2.14) And now we do support both types.

rchl: And you are sure that we want to extend it to string values also? I'd personally prefer more strict set of types but if you insist...

packages/webpack/src/config/base.js
rchl: Oh. I guess it would be quoted when using JSON.stringify...

packages/webpack/src/config/base.js
```diff
@@ -120,15 +120,15 @@ export default class WebpackBaseConfig {
...
       } catch (_err) {
-        corejsVersion = '2'
+        corejsVersion = 2
       }
     }
 
-    if (corejsVersion !== '2' && corejsVersion !== '3') {
-      consola.warn(`Invalid corejs version ${JSON.stringify(corejsVersion)}! Please set "build.corejs" to either "2" or "3".`)
-      corejsVersion = '2'
+    if (corejsVersion !== 2 && corejsVersion !== 3) {
```
danielroe, line 129:
```suggestion
    if (![2, 3].includes(corejsVersion)) {
```

rchl: Thanks but first we have to decide on if we support string numbers :)

packages/types/config/build.d.ts
pi0: Hmm i see. Since we didn't add types on 2.14.0 having types to force number is non-breaking.

pi0 (approved):

pi0:
Docs need to be updated with 2.14.1 release :)

PR #7801, opened 2020-07-28
fix(vue-app): handle server-side routing errors

### LINKED ISSUE #7798: router guard error causes SSR request to never complete
### Versions

- nuxt: v2.14.0
- node: v14.5.0

### Reproduction

https://codesandbox.io/s/inspiring-nash-hknkc?file=/plugins/router-error.js

According to `vue-router` [navigation guards documentation](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards), the `next()` function can be called with an error.

However, calling `next()` with either an `error` or `false` during SSR will cause the server never to respond.

Here are the attempts I made:

<details>
<summary>Using `next(err)`</summary>

```js
export default function plugin({ app, error }, inject) {
  app.router.beforeEach(async (to, from, next) => {
    try {
      await someAsyncTask()
      next()
    } catch (err) {
      next(err) // Will cause the server to hang indefinitely
    }
  })
}
```
</details>
<details>
<summary>Using `onError`</summary>

```js
export default function plugin({ app, error }, inject) {
  app.router.onError(error) // Will cause the server to hang indefinitely
}
```
</details>

<details>
<summary>Using `next(false)`</summary>

```js
export default function plugin({ app, error }, inject) {
  app.router.beforeEach(async (to, from, next) => {
    try {
      await someAsyncTask()
      next()
    } catch (err) {
      next(false) // Will cause the server to hang indefinitely
      error(err) // Does nothing
    }
  })
}
```
</details>

<details>
<summary>Workaround</summary>

```js
export default function plugin({ app, error }, inject) {
  app.router.beforeEach(async (to, from, next) => {
    try {
      await someAsyncTask()
    } catch (err) {
      error(err)
    } finally {
      next() // The route will still be loaded, but never shown...
    }
  })
}
```
</details>

### Steps to reproduce
```js
export default function plugin({ app, error }, inject) {
  app.router.beforeEach(async (to, from, next) => {
    next(new Error("Some async error"));
  });
}
```

### What is Expected?
Nuxt should handle vue-router navigation errors (esp. during SSR)

Nuxt does not handle router errors, causing the requests to hang indefinitely

matthieusieben:
## Types of changes

## Description

Resolves: #7798

https://github.com/nuxt/nuxt.js/blob/4caffb04ab9309fb334f68739c61b73f8c325a05/packages/vue-app/template/index.js#L248-L268

This introduces a fix for 2 use cases:
- The user uses `next(err)` in a router guard. This will generate a 500 error.
- The user uses `next(false)` in a router guard. This will:
  - Either render any error provided through `context.error()`
  - Or generate a 404 (by default)

## Checklist:


matthieusieben:
Still does not handle the `next(false)` case...

matthieusieben:
Will cause the following error when using `next(false)`:
```
Navigation aborted from "/" to "/xxx" via a navigation guard.
```

matthieusieben:
I can't figure out how to do something like "if there was a routing error but `error(err)` was called with an error, then just render the error layout page, otherwise throw".

matthieusieben:
About:
https://github.com/nuxt/nuxt.js/blob/762d49a06d6e94d6faa2e04a2ab76fab527b5475/packages/vue-app/template/index.js#L267-L270

Typically, the user did `next(false)` in a router guard. There are two options here:
- Consider that it is a coding mistake not to being able to resolve a route (which it kinda is) and `reject(err)`, resulting in a 500.
- Use a more user friendly approach and just `resolve()` (as it is the case in my commit) which will result in a 404.

matthieusieben:
These changes will imply that users should never do `next(err)` (or they will get a 500) but do `error(err); next(false)` instead.

Note that in the browser, nothing will happen if a user does `next(err)` except from displaying an error in the console.

packages/vue-app/template/index.js
```diff
@@ -246,16 +246,28 @@ async function createApp(ssrContext, config = {}) {
...
-        const unregister = router.afterEach(async (to, from, next) => {
-          ssrContext.url = to.fullPath
-          app.context.route = await getRouteData(to)
-          app.context.params = to.params || {}
-          app.context.query = to.query || {}
-          unregister()
+      router.push(ssrContext.url, resolve, (err) => {
+        if (!err._isRouter) {
+          // router guard returned an error
+          reject(err)
```
pi0, line 252: we can do early return on this rejection `return reject(err)`

packages/vue-app/template/index.js
```diff
@@ -246,16 +246,28 @@ async function createApp(ssrContext, config = {}) {
...
+        } else if (err.type === 1) {
```
pi0, line 253: Similarry, `if (err.type !== 1) { return resolve() }`

packages/vue-app/template/index.js
matthieusieben: done

packages/vue-app/template/index.js
matthieusieben: done

pi0:
Thanks for PR @matthieusieben :heart:

Issue #7625, opened 2020-06-27
Dynamic routing broken with AWS API Gateway


aaronhuisinga:
### Version

[v2.13.2](https://github.com/nuxt.js/releases/tag/v2.13.2)

### Reproduction link

(Not possible without AWS setup using API Gateway)

### Steps to reproduce

1. Setup Nuxt app running behind AWS API Gateway
2. Create dynamic route
3. Click link to dynamic route

### What is expected ?

Router navigates to page defined in dynamic route

### What is actually happening?

400 error when trying to fetch `/_nuxt/pages/resource/[_]id.js`

### Additional comments?

AWS API Gateway does not allow square brackets "[" and "]" in URLs. Nuxt uses these square brackets in the file names for the compiled JS for dynamic routes. Any time that a request is made to load this file, API Gateway returns a 400 error. The recommended fix for this is encoding the URL before making the request, which is what Next.js did to fix this same issue (https://github.com/vercel/next.js/issues/8013). Another possible fix could be adding a configuration option for changing the name of the generated files for dynamic routes, and giving users the ability to replace the square brackets with something else.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10860">#c10860</a>)</em></sub></div>

cmorbitzer:
Also experiencing this issue. Seems to be introduced by #7603

pi0:
Hi! Should be fixed by #7639 (you can early test using `nuxt-edge`)

aaronhuisinga:
Thanks @pi0! I can confirm this PR fixes the issue. Just tested using `nuxt-edge` and routing works as expected.

PR #7463, opened 2020-06-04
feat: improve production debugging dx

pi0:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Resolve SSR stack trace paths

(outputs modified for readability. consola has a fancy logger that does some similar magic but only for display)

Old:

```
ReferenceError: crashme is not defined                                                                                                                                      16:07:17
    at asyncData (examples/hello-world/pages/index.vue:13:0)
    at promisify (examples/hello-world/.nuxt/utils.js:259:0)
    at server.js:1:<col>
    at Array.map (<anonymous>)
    at module.exports.__webpack_exports__.default (examples/hello-world/.nuxt/server.js:227:0)
```

New:

```
ReferenceError: crashme is not defined                                                                                                                                      
    at asyncData ([path]/examples/hello-world/pages/index.vue:13:0)
    at promisify ([path]/examples/hello-world/.nuxt/utils.js:259:0)
    at [path]/examples/hello-world/.nuxt/dist/server/server.js:2582:23
    at Array.map (<anonymous>)
    at module.exports.__webpack_exports__.default ([path]/examples/hello-world/.nuxt/server.js:227:0)
```

### Make chunk names easy to debug

Old:

```
├── client
│   ├── 421fcb3837a373bcc316.js
│   ├── 43f6a8fec12d37fd9731.js
│   ├── 923c20ce304abf61b7d0.js
│   ├── f4d7116a13cbd5bad38e.js
│   ├── fb4666c8451b74cab31e.js
│   └── LICENSES
└── server
    ├── 7250d0198e376567f352.js
    ├── ae78742f8bb4962f2610.js
    ├── client.manifest.json
    ├── index.spa.html
    ├── index.ssr.html
    ├── server.js
    └── server.manifest.json
```

New:

```
├── client
│   ├── app.43f6a8f.js
│   ├── commons.app.f4d7116.js
│   ├── LICENSES
│   ├── pages
│   │   ├── about.923c20c.js
│   │   └── index.421fcb3.js
│   └── runtime.e630eac.js
└── server
    ├── client.manifest.json
    ├── index.spa.html
    ├── index.ssr.html
    ├── pages
    │   ├── about.js
    │   ├── about.js.map
    │   ├── index.js
    │   └── index.js.map
    ├── server.js
    ├── server.js.map
    └── server.manifest.json
```

[...]

PR #7808, opened 2020-07-29
fix(webpack): avoid grouping with default name

### LINKED ISSUE #7795: [2.14.0] Async components don't get separate bundles
# Versions

- nuxt: v2.14.0
- node: v12.18.1

### Reproduction

https://github.com/danielroe/webpack-async

### Steps to reproduce

1. Clone repo
2. Run `yarn build --analyze` - note that there is a separate async bundle for the huge dependency `twilio-video`
3. Run `yarn add nuxt@2.14 && yarn build --analyze` - note that there is no separate async bundle

### What is expected?

Bundle behaviour should not change in this significant of a way.

Significant performance degradation by bundling all dependencies in just five JS files (even on large sites), meaning we don't benefit from code-splitting or async dependencies.

### Notes
* Regression introduced in b4ffdab790b89b7fe806b2168a70925dd1f5e10b

pi0:
## Types of changes

## Description

Fixes #7795 by letting webpack to use computed cache group key.

Related:
- #7463 (initial PR)
- #7603
- #7639 (fixing #7625 and #7603 and 7463#issuecomment-650734494)
- 2302ffa40b2e83775ca9f7d5ab7d63b38b523333
- #7667 (fixing #7665)

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
     "rollup-plugin-license": "^2.1.0",
     "sass-loader": "^8.0.2",
     "sort-package-json": "^1.44.0",
-    "twilio-video": "^2.7.1",
     "typescript": "~3.9",
     "vue-jest": "^4.0.0-beta.4"
   }
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "rollup-plugin-license": "^2.1.0",
     "sass-loader": "^8.0.2",
     "sort-package-json": "^1.44.0",
+    "twilio-video": "^2.7.1",
     "typescript": "~3.9",
     "vue-jest": "^4.0.0-beta.4"
   }
--- a/packages/config/src/config/build.js
+++ b/packages/config/src/config/build.js
@@ -15,7 +15,7 @@ export default () => ({
   filenames: {
     // { isDev, isClient, isServer }
     app: ({ isDev, isModern }) => isDev ? `[name]${isModern ? '.modern' : ''}.js` : `[name].[contenthash:7]${isModern ? '.modern' : ''}.js`,
-    chunk: ({ isDev, isModern }) => isDev ? `[name]${isModern ? '.modern' : ''}.js` : `[name].[contenthash:7]${isModern ? '.modern' : ''}.js`,
+    chunk: ({ isDev, isModern }) => isDev ? `[name]${isModern ? '.modern' : ''}.js` : `[name]_[contenthash:7]${isModern ? '.modern' : ''}.js`,
     css: ({ isDev }) => isDev ? '[name].css' : '[name].[contenthash:7].css',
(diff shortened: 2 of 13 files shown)


pi0:
I did a bunch of tests to ensure this fix is not causing dot file issue and webpack's default `[name]` generator ([src](https://github.com/webpack/webpack/blob/v4.44.0/lib/optimize/SplitChunksPlugin.js#L162)) returns to fallback into default key still would be nice if we can beta test it on `nuxt-edge`  /cc @danielroe @aldarund @simllll @husayt

clarkdo (commented):
Do we have an exmaple about what the new output dirs and files would be ?

pi0:
For #7795 reproduction:

```
.nuxt/dist/client
├── app.97f1f0a.js
├── LICENSES
├── pages
│   └── index.d590496.js
├── runtime.d718bc0.js
└── vendors
    ├── commons.15ff78a.js
    └── pages
        └── index.dba88f0.js
```

With a vendor shared between two pages:

```
├── app.52c70b1.js
├── LICENSES
├── pages
│   ├── foo.56fb624.js
│   └── index.e9174bf.js
├── runtime.b2c0774.js
└── vendors
    ├── commons.4e61828.js
    └── foo~index.f8232e7.js
```

Bud sadly (or happily :D) no simple reproduction for dotfile issue happened in previous releases

danielroe:
I have projects that can replicate both issues so will test.

package.json
```diff
@@ -76,6 +76,7 @@
     "rollup-plugin-license": "^2.1.0",
     "sass-loader": "^8.0.2",
     "sort-package-json": "^1.44.0",
+    "twilio-video": "^2.7.1",
```
danielroe, line 79: can we replicate fixture without this particular package?

simllll:
Will test it tomorrow!

package.json
pi0: What do you suggest? Initially added to replicate with similar dep

danielroe: I will see if I can find another large dependency already within the Nuxt codebase as I hate the thought of adding unnecessarily to install. 🏋️‍♂️

simllll:
> Will test it tomorrow!

Just tested it with nuxt-next in two of our applications that were affected by the . bug before. Looks good to me! Everything seems to work!

pi0:
Update

```
├── app.37f7ad3.js
├── LICENSES
├── node_modules
│   ├── commons.99cd09c.js
│   └── foo~index.c1279c0.js
├── pages
│   ├── foo.f8abb2f.js
│   └── index.2145558.js
└── runtime.a01e42c.js
````

package.json
pi0: https://github.com/nuxt/nuxt.js/pull/7808/commits/e9f60c3b31b314245155d7b5f41fdb6aa891df74

atinux (approved):

PR #7810, opened 2020-07-30
feat: bind serverMiddleware to nuxt instance

btkostner:
Bind the server middleware functions to the nuxt instance. Ideally I wished this was [the server context](https://nuxtjs.org/api/context/) object because I wanted to load some runtime configuration options, but I couldn't find it. This is at least a little stop gap that makes it easier to access lower level options.

## Types of changes

## Description
Solves the issue of not being able to access nuxt or any lower level settings in server middleware functions.

## Checklist:

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -273,7 +273,7 @@ export default class Server {
 
   useMiddleware (middleware) {
     const { route, handle } = this.resolveMiddleware(middleware)
-    this.app.use(route, handle)
+    this.app.use(route, handle.bind(this.nuxt))
   }
 
   replaceMiddleware (query, middleware) {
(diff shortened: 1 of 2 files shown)


btkostner:
errrr, I might need some help with Jest asserts. The `bind` seems to mess with `toBeCalledWith` and they are not the same now :(

pi0:
Hi @btkostner thanks for the PR ❤️ There is concern that this might lead to security issues since middleware are executed based on your input and `nuxt` object has access to the request-wise context. You can do it easily using a module to inject into `req.nuxt`:

```js
export default function () {
  const { nuxt } = this
  nuxt.options.serverMiddleware.unshift((req, _res, next) => {
    req.nuxt = nuxt
    next()
  })
}
```

This solution can be also published as a package to use across projects.

PR #7839, opened 2020-08-03
fix(core, config): disabe esm cache

clarkdo:
## Types of changes

## Description

Related to https://github.com/nuxt/nuxt.js/issues/7838

We can see there're [some opened issues](https://github.com/standard-things/esm/search?q=cache&type=Issues) related to esm cache and there is no response for a long time, so I think we can disable it for now because we'll remove esm in Nuxt 3 anyway.

## Checklist:

### DIFF
--- a/packages/config/src/load.js
+++ b/packages/config/src/load.js
@@ -18,7 +18,7 @@ export async function loadNuxtConfig ({
   configFile = defaultNuxtConfigFile,
   configContext = {},
   configOverrides = {},
-  createRequire = module => isJest ? _createRequire(module.filename) : esm(module)
+  createRequire = module => isJest ? _createRequire(module.filename) : esm(module, { cache: false, cjs: { cache: false } })
 } = {}) {
   rootDir = path.resolve(rootDir)
 
--- a/packages/config/src/load.js
+++ b/packages/config/src/load.js
@@ -18,7 +18,7 @@ export async function loadNuxtConfig ({
   configFile = defaultNuxtConfigFile,
   configContext = {},
   configOverrides = {},
-  createRequire = module => isJest ? _createRequire(module.filename) : esm(module, { cache: false, cjs: { cache: false } })
+  createRequire = module => isJest ? _createRequire(module.filename) : esm(module, { cache: false })
 } = {}) {
   rootDir = path.resolve(rootDir)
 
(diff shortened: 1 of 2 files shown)


packages/config/src/load.js
```diff
@@ -18,7 +18,7 @@ export async function loadNuxtConfig ({
   configFile = defaultNuxtConfigFile,
   configContext = {},
   configOverrides = {},
-  createRequire = module => isJest ? _createRequire(module.filename) : esm(module)
+  createRequire = module => isJest ? _createRequire(module.filename) : esm(module, { cache: false, cjs: { cache: false } })
```
pi0, line 21: `cjs.cache` option is a flag to indicate where to expose runtime cache not about fs cache: https://github.com/standard-things/esm/blob/950e74186acf7e2e733168b327f8d9747ad13427/src/module/esm/load.js#L58

pi0: I didn't test but as reported by @rchl (https://github.com/nuxt/nuxt.js/issues/7838#issuecomment-667934798) we may set `cache` option to `node_modules/.cache/esm` to avoid regression and still benefit fs based caching.

rchl: I'm not sure about that. The issue seems to be about npm not clearing the cache. I'm not sure I understand how having it in one place help the issue.

clarkdo: I think cache path of esm is calculated by esm package dir which can not be override

rchl: > `cjs.cache` option is a flag to indicate where to expose runtime cache not about fs cache: [standard-things/esm@`950e741`/src/module/esm/load.js#L58](https://github.com/standard-things/esm/blob/950e74186acf7e2e733168b327f8d9747ad13427/src/module/esm/load.js#L58)

Which will be then written to disk on exit no?

https://github.com/standard-things/esm/blob/91a154bedd2700a8fd85d78c1e15161676104419/src/caching-compiler.js#L168

(I don't pretend to understand esm's code. This is just after quite look.)

pi0: @clarkdo You are right it is calculated by esm [__dirname](https://github.com/standard-things/esm/blob/91a154bedd2700a8fd85d78c1e15161676104419/esm.js#L105)

@rchl Yes. It was about `cjs.cache` option which indicates using `require.cache{}` or another internal object for in mem cache (when require same file multiple times)

pi0 (approved):

PR #7860, opened 2020-08-05
fix(cli): improve project dir detection for external commands

farnabaz:
Directories without `nuxt.config.js` are not valid projects, thus we could check for `nuxt.config.(ts|js)` existence in CLI.
This change resolves conflicts between directories and commands.

https://github.com/nuxt-community/storybook/issues/30

## Types of changes

## Checklist:


pi0:
nuxt.config file is optional. we may add a more robust check to also check for pages/ directory existence.

packages/cli/src/run.js
```diff
@@ -28,7 +29,7 @@ export default async function run (_argv, hooks = {}) {
   let cmd = await getCommand(argv[0])
 
   // Matching `nuxt` or `nuxt [dir]` or `nuxt -*` for `nuxt dev` shortcut
-  if (!cmd && (!argv[0] || argv[0][0] === '-' || (argv[0] !== 'static' && fs.existsSync(argv[0])))) {
+  if (!cmd && (!argv[0] || argv[0][0] === '-' || fs.existsSync(`${argv[0]}${path.sep}pages`) || fs.existsSync(`${argv[0]}${path.sep}pages`))) {
```
pi0, line 32: `argv[0]` is `rootDir` (while pages can be different using custom `srcDir`) but it needs a configuration then :P So we should check both with OR maybe with a utility :)

farnabaz: If `nuxt.config` is missing, `rootDir` and `srcDir` is same. If we check both of `nuxt.config` and `pages` we achieve our goal, right?

pi0: One may have `nuxt.config` but no `rootDir/pages`. So existence of either one means it is a valid nuxt directory.

packages/cli/src/run.js
```diff
@@ -28,7 +29,7 @@ export default async function run (_argv, hooks = {}) {
   let cmd = await getCommand(argv[0])
 
   // Matching `nuxt` or `nuxt [dir]` or `nuxt -*` for `nuxt dev` shortcut
-  if (!cmd && (!argv[0] || argv[0][0] === '-' || (argv[0] !== 'static' && fs.existsSync(argv[0])))) {
+  if (!cmd && (!argv[0] || argv[0][0] === '-' || fs.existsSync(`${argv[0]}${path.sep}nuxt.config.js`) || fs.existsSync(`${argv[0]}${path.sep}pages`) || fs.existsSync(`${argv[0]}${path.sep}nuxt.config.ts`))) {
```
farnabaz, line 32: @pi0 What do you think about the order of these 3. Should we put `pages` at first?

pi0: It is fine just would be better if we refactor out into a util function :D

farnabaz: Yeah, I just want to confirm it.
I was thinking about creating `isNuxtDir` or something like it. It will be useful for modules, especially in the Storybook.
https://github.com/nuxt-community/storybook/blob/master/src/cli.ts#L27

pi0: Maybe we cannot reuse as it would require using latest nuxt for storybook but can use same logic :)

farnabaz: Isn't it a feature? 😅

packages/cli/src/utils/dir.js
```diff
@@ -0,0 +1,11 @@
+import fs from 'fs'
+import path from 'path'
+
+export function isNuxtDir (dir) {
+  if (fs.existsSync(`${dir}${path.sep}nuxt.config.js`) ||
```
pi0, line 5: dir may have trailing slash from cli. Can we use path.join instead? 🙈

packages/cli/src/utils/dir.js
```diff
@@ -0,0 +1,11 @@
+import fs from 'fs'
+import path from 'path'
+
+export function isNuxtDir (dir) {
```
pi0, line 4: i would suggest using rootDir to clarify

pi0 (approved):
:ok_hand:

PR #7862, opened 2020-08-05
server(hook): Ability to execute a function before starting the server

loick:
In some use cases, we need to initialize our server with some tools, before even starting the app. The best example is the monitoring. In order to be sure everything is tracked, it has to be the first thing done on the app. 

The problem here, is that I really want to stick as close as possible to the "native" nuxt implementation using the nuxt commands (nuxt dev, nuxt start etc.). Using the server with an express wrapper is not a good option in my opinion because I would create several ways of doing 1 thing (express middleware, nuxt middleware etc.), highly emphasised by the size of the team which could lead to inconsistencies. 

Let me give you an example, it might be easier to follow : ) *Datadog*

https://docs.datadoghq.com/tracing/setup/nodejs/#quickstart

This APM library needs to be (as it is explained) the first library loaded, before the server. 

## Types of changes

## Description
I believe, simply adding a hook on the Server would do the trick (correct me if I'm wrong). That way this is flexible enough to implement anything we want here.

I'm open to suggestions if you believe I didn't look at the right place :) 

## Checklist:

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -18,6 +18,9 @@ export default class Server {
     this.nuxt = nuxt
     this.options = nuxt.options
 
+    // Before server init
+    this.nuxt.callHook('server:prepare', null)
+
     this.globals = determineGlobals(nuxt.options.globalName, nuxt.options.globals)
 
     this.publicPath = isUrl(this.options.build.publicPath)
--- a/packages/server/test/server.test.js
+++ b/packages/server/test/server.test.js
@@ -146,9 +146,10 @@ describe('server: server', () => {
 
     await server.ready()
 
-    expect(server.nuxt.callHook).toBeCalledTimes(2)
-    expect(server.nuxt.callHook).nthCalledWith(1, 'render:before', server, server.options.render)
-    expect(server.nuxt.callHook).nthCalledWith(2, 'render:done', server)
+    expect(server.nuxt.callHook).toBeCalledTimes(3)
+    expect(server.nuxt.callHook).nthCalledWith(1, 'server:prepare', null)
+    expect(server.nuxt.callHook).nthCalledWith(2, 'render:before', server, server.options.render)
+    expect(server.nuxt.callHook).nthCalledWith(3, 'render:done', server)
     expect(ServerContext).toBeCalledTimes(1)
     expect(ServerContext).toBeCalledWith(server)
     expect(VueRenderer).toBeCalledTimes(1)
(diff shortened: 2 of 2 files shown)


pi0:
Hi @loick. Is it too late if registreing datadog with `server:listen` hook?

loick:
> Hi @loick. Is it too late if registreing datadog with `server:listen` hook?

Yes, that's what I tried but sadly it didn't work. Btw, I also opened an issue on the datadog side to get their recommendation, and maybe to have some example of it (even with nextJS I guess it would be similar)

pi0:
> Yes, that's what I tried but sadly it didn't work

Hmm what about `render:before` hook? Because current hook is almost same time as `nuxt:ready` or module entrypoint. We just need to check not using `nuxt build` and init datadog.

loick:
> > Yes, that's what I tried but sadly it didn't work
> 
> Hmm what about `render:before` hook? Because current hook is almost same time as `nuxt:ready` or module entrypoint. We just need to check not using `nuxt build` and init datadog.

Same problems on the `render:before` yeah :/ 

I believe it has to be called before this: 
```
this.app = connect()
```

When is the `nuxt:ready` call? Not sure where to find it

pi0:
The server constructor is called by [_initServer](https://github.com/nuxt/nuxt.js/blob/b32fd650a97f3ef5933984a865c14262be017cde/packages/core/src/nuxt.js#L57) which is before modules being registered. Would be nice if you can share your module's repo working on to test different methods against with.

loick:
> The server constructor is called by [_initServer](https://github.com/nuxt/nuxt.js/blob/b32fd650a97f3ef5933984a865c14262be017cde/packages/core/src/nuxt.js#L57) which is before modules being registered. Would be nice if you can share your module's repo working on to test different methods against with.

Here is an example of the module I'm testing: https://github.com/loick/nuxt-module-datadog-test/blob/master/index.js 

I've seen someone trying something similar without adding a hook on the module: https://github.com/aaronransley/nuxt-datadog-trace/blob/master/index.js, but it didn't work either. 

I'm talking with some people from Datadog, and they are apparently working with the NextJS team to implement properly their APM with the framework. I'm trying to get more infos about it, maybe it would be the opportunity to do an official module for NuxtJS as well

pi0:
Awesome @loick I'll have a look and we can work together on proper support 😊 (meanwhile closing this PR as the hook on constructor doesn't works anyway)

PR #7877, opened 2020-08-07
fix(vue-app): use anchor to `/` if route is undefined in error page

### LINKED ISSUE #7841: Crash with // in the URL
### Versions

- nuxt: v2.14.0
- node: v12.18.1 

### Steps to reproduce

1. Create a new app with $ yarn create nuxt-app test
2. Run $ yarn dev
3. Open http://localhost:3000// (with 2 slashes)

### What is Expected?

To get a 404

image

farnabaz:
When `vue-router` does not detect current route and `$route` is `undefined`, we using `NuxtLink` or `RouterLink` throws an exception and it causes a loop in nuxt error page

close #7841

## Types of changes

## Checklist:


pi0:
What if using a custom error page? Maybe we can handle condition with `NuxtLink` or directly a fix for `vue-router`? /cc @posva

posva:
$route shouldn't be undefined. I think there's an open issue in vue router about the double leading slash
This was fixed in v4

PR #7902, opened 2020-08-13
types: improve definitions for component transition property

mwgamble:
## Types of changes

## Description

There are two issues with supplying a function for the transition
property:

- The 'from' parameter can be undefined in certain circumstances, such
  as the initial page load when a user first arrives at the application.
  In this situation, there is no 'from' route.
- It's perfectly valid for this function to return a Transition object,
  as if that object had been assigned directly to the transition property.
  The underlying code already supports this.

This PR resolves both issues.

## Checklist:


pi0 (commented):
Thanks!

Issue #7915, opened 2020-08-13
Nuxt memory leak universal mode


WilliamDASILVA:
### Versions

- nuxt: 2.14.1
- node: 10.16.0

### Reproduction

<details open>
<summary>Additional Details</summary>
<br>

nuxt.config
```js
module.exports = {
  mode: 'universal',
  srcDir: './src',
  server,
  head: {
    title: pkg.name,
    noscript: [{ innerHTML: 'This website requires JavaScript to run properly.' }],
    htmlAttrs: {
      dir: 'ltr'
    },
    meta: [
      { charset: 'utf-8' },
      { name: 'viewport', content: 'width=device-width, initial-scale=1' },
      { hid: 'description', name: 'description', content: '' }
    ],
    link: [
      { rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }
    ]
  },
  css: [
    { src: '@/assets/scss/style.scss', lang: 'scss' }
  ],
  plugins: [
    '~/plugins/vee-validate.js'
  ],
  modules: [
    'nuxt-stripe-module',
    '@nuxtjs/style-resources',
    '@nuxtjs/axios',
    '@nuxtjs/sitemap',
    '@nuxtjs/sentry',
    '@nuxtjs/robots',
    ['vue-wait/nuxt', { useVuex: true }],
    ['@nuxtjs/google-analytics', {
      id: 'UA-TEST'
    }]
  ],
  stripe: {
    publishableKey: process.env.STRIPE_PUBLISHABLE_KEY || ''
  },
  sentry: {
    dsn: process.env.SENTRY_DSN || ''
  },
  sitemap: {
    path: '/sitemap.xml'
  },
  styleResources: {
    scss: [
      'assets/scss/_variables.scss',
      'assets/scss/_mixins.scss'
    ]
  },
  build: {
    transpile: [
      'vee-validate/dist/rules'
    ]
  },
  env: {
    appUrl: process.env.VUE_APP_APP_URL
  },
  axios: {
    baseURL: process.env.API_BASE_URL,
    browserBaseURL: process.env.API_BROWSER_BASE_URL
  },
  tailwindcss: {
    cssPath: '~/assets/css/tailwind.css'
  },
  buildModules: [
    '@nuxtjs/pwa',
    '@nuxtjs/tailwindcss'
  ]
}
```

store.js
```js
import Vue from 'vue'
import Vuex from 'vuex'

Vue.use(Vuex)

[...]

PR #7916, opened 2020-08-14
fix(vue-app): prevent router.registerModule reallocation (#7915)

galvez:
Prevents reallocating a new function on every request for the `registerModule` override, as reported in #7915.


packages/vue-app/template/index.js
```diff
@@ -66,6 +67,14 @@ const defaultTransition = <%=
 %><%= isTest ? '// eslint-disable-line' : '' %>
 <% } %>
 
+const { assign, create } = Object
+const originalRegisterModule = Vuex.Store.prototype.registerModule
+const baseStoreOptions = { preserveState: process.client }
+
+function registerModule (path, rawModule, options) {
+  originalRegisterModule.call(this, path, rawModule, assign(create(baseStoreOptions), options))
```
pi0, line 75: - Any reason not using { preserveState, ...options } ?
- Currently registerModule isn't returning anything but maybe better if we preserve it's return value?

pi0 (commented):
Thanks for PR @galvez 👍 However we should ensure bug/fix is valid with a nuxt only reproduction (instead of user report) seems also nice refactor and improvement.

packages/vue-app/template/index.js
galvez: I thought `...` produced some horrible transpiled code? Also with create() won't we be further limiting allocation of at least one property? Good call about adding a `return`.

galvez: No idea why the build failed -- will run locally again.

pi0: @galvez We are using spread syntax several places at least in `client.js` and babel helpers (which are not that heavy) already injected once. Also babel rules are different for modern. [example transpile](https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDA3jAZiEAuGUBOArgKYA0MAdFQEZz4wC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cenv&prettier=false&targets=&version=7.11.1&externalPlugins=)

galvez: @pi0 updated! Also reworded the pR title so you can classify this as a general enhancement.

It does seem like the memory leak is valid, since that function allocation is happening repeatedly.

PR #7920, opened 2020-08-14
fix(vue-app): fix regression with `scrollToTop`

pi0:
## Types of changes

## Description

`vue-router@3.4.0` landed with vuejs/vue-router#3199 which is breaking `scrollToTop: false` as we need to call `triggerScroll` event on first page too otherwise it will be delayed until next navigation which causes jumping to top (or saved position of previous page)

PR also introducing some code refactors

## Checklist:

### DIFF
--- a/packages/vue-app/template/router.scrollBehavior.js
+++ b/packages/vue-app/template/router.scrollBehavior.js
@@ -22,7 +22,11 @@ if (process.client) {
   }
 }
 
-export default function (to, from, savedPosition) {
+  export default function (to, from, savedPosition) {
+  if (from === to) {
+    return
+  }
+
   // if the returned position is falsy or an empty object,
   // will retain current scroll position.
   let position = false
--- a/packages/vue-app/template/router.scrollBehavior.js
+++ b/packages/vue-app/template/router.scrollBehavior.js
@@ -23,24 +23,24 @@ if (process.client) {
 }
 
   export default function (to, from, savedPosition) {
-  if (from === to) {
-    return
-  }
+  const isInitialLoad = to === from
 
-  // if the returned position is falsy or an empty object,
-  // will retain current scroll position.
+    if (isInitialLoad) {
+      return
+    }
+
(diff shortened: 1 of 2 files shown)


packages/vue-app/template/router.scrollBehavior.js
```diff
@@ -47,8 +45,12 @@ export default function (to, from, savedPosition) {
 
   const nuxt = window.<%= globals.nuxt %>
 
-  // triggerScroll is only fired when a new component is loaded
-  if (to.path === from.path && to.hash !== from.hash) {
+  if (
+    // Page hash changes
+    (to.path === from.path && to.hash !== from.hash) ||
+    // Initial load (vuejs/vue-router#3199)
+    to === from
```
pi0, line 52: @posva Do you think it is a safe check for detecting initial load?

pi0:
merging to fix ci and test on edge

test/dev/unicode-base.size-limit.test.js
```diff
@@ -20,7 +20,7 @@ describe('nuxt minimal vue-app bundle size limit', () => {
   it('should stay within the size limit range', async () => {
     const filter = filename => filename === 'vue-app.nuxt.js'
     const legacyResourcesSize = await getResourcesSize(distDir, 'client', { filter })
-    const LEGACY_JS_RESOURCES_KB_SIZE = 16.5
```
farnabaz, line 23: 🤣

packages/vue-app/template/router.scrollBehavior.js
posva: not completely sure, we could also check if `from` is the `START` location that is exported in v4.

pi0: I tried exposing [START](https://github.com/CoolCyberBrain/vue-router/blob/v3.2.0/src/util/route.js#L52) ref but checking here, on initial load `from` and `to` are same references but not equal to START:

image

PR #7938, opened 2020-08-17
fix(cjs): fix HMR issue with mongoose

### LINKED ISSUE #7917: App crashes on ServerMiddleware hot reload if Mongoose models are used, gives an error like "ServerMiddleware Error: Cannot overwrite User model once compiled"
### Versions

- nuxt: v2.14.1
- node: v12.18.3

### Reproduction
Use [this project](https://github.com/nuxt/nuxt.js/files/5074173/reproduction.zip) to reproduce, you may need to add a working MongoDB database credentials in .env file.

### Steps to reproduce
npm install
npm run dev
(No error at this point yet)
Make a change in login.js file
Save

### What is Expected?
Normal hot reload of servermiddleware as always (like when Mongoose is not used).

App crashes with this error:

`ERROR ServerMiddleware Error: Cannot overwrite User model once compiled.`

farnabaz:
## Types of changes

## Description
Introduce `isHMRCompatible` function to detect clear incompatible modules cache.  
As we know `mongoose` is incompatible and clearing `mongoose/index.js` and `mongoose/lib/index.js` makes it work.

close #7917

## Checklist:

### DIFF
--- a/packages/utils/src/cjs.js
+++ b/packages/utils/src/cjs.js
@@ -1,7 +1,7 @@
 import { join } from 'path'
 
 export function isHMRCompatible (id) {
-  return !/[/\\]mongoose[/\\/](lib[/\\/])?index.js/.test(id)
+  return !/[/\\]mongoose[/\\/]/.test(id)
 }
 
 export function isExternalDependency (id) {
--- a/packages/utils/src/cjs.js
+++ b/packages/utils/src/cjs.js
@@ -1,11 +1,15 @@
 import { join } from 'path'
 
+export function isHMRCompatible (id) {
+  return !/[/\\]mongoose[/\\/](lib[/\\/])?index.js/.test(id)
+}
+
 export function isExternalDependency (id) {
   return /[/\\]node_modules[/\\]/.test(id)
 }
 
 export function clearRequireCache (id) {
-  if (isExternalDependency(id)) {
+  if (isExternalDependency(id) && isHMRCompatible(id)) {
     return
   }
 


packages/utils/src/cjs.js
```diff
@@ -1,11 +1,15 @@
 import { join } from 'path'
 
+export function isHMRCompatible (id) {
+  return !/[/\\]mongoose[/\\/](lib[/\\/])?index.js/.test(id)
```
pi0, line 4: Why not marking entire mongoose dep as incompatible? A regex matching `node_modules/mongoose` as a part of id

farnabaz: mongoose has lots of files and it does not bundle them. clearing entire deps causes Maximum call stack size exceeded. (Honestly, I don't understand why)

pi0: Do you have a repo to quickly check? Maybe it is a nuxt HMR bug

farnabaz: I'll have one, let me push it to github

farnabaz: @pi0 https://github.com/farnabaz/nuxt-mongoose

clarkdo: Can we create a nuxt module for mongoouse and clear cache via `watchers` or `modules:before` hook ?

pi0: @farnabaz Typescript recursion should be fixed by (#7966)

pi0: @clarkdo Seems nice idea. Also there is an alternative way i used once to use [unload](#7358) method. 

Since it is really minor case (serverMiddleware + HMR + mongoose) but not uncommon setup, it may worth excluding mongoose by default. Nuxt 3 will isolate functions anyway so we won't have such issues for HMR.

farnabaz: @pi0 I have updated the regex based on based on fix #7966

pi0 (approved):

PR #7939, opened 2020-08-17
fix(vue-app): var names shadowing in chrome debugger

### LINKED ISSUE #7731: Shadowing name variables in store.js
[Very low priority]
### Versions

- nuxt: v2.12.2
- node: v12.17.0

### Reproduction

A very small detail that annoyed me a bit while debugging nuxt in the Chrome Devtool: **some variables held the same name in a single code block and, due to hoisting, it results in shadowing variable names that Chrome debugger can't deal with.**

Here is a screenshot that illustrate the problem:
Screenshot 2020-07-17 at 09 08 01

The concerned variables are `storeModule` and `property`. On hover, I should see the value `{ namespace: true, modules: {...} }` but Chrome actually show me `undefined` as it's the first variable's value with that name in the current scope.

Maybe we should consider renaming the variables in the `if` statement so that the debugger won't mix up both values? :)

Kapcash:
Fixes https://github.com/nuxt/nuxt.js/issues/7731 :)
Just renamed a variable in a small block so that another variable in the parent block don't shadow eachother in the Chrome debugger.

## Types of changes

## Checklist:

### DIFF
--- a/packages/vue-app/template/store.js
+++ b/packages/vue-app/template/store.js
@@ -94,10 +94,10 @@ function resolveStoreModules (moduleData, filename) {
   // If src is a known Vuex property
   if (VUEX_PROPERTIES.includes(moduleName)) {
     const property = moduleName
-    const storeModule = getStoreModule(store, namespaces, { isProperty: true })
+    const moduleStore = getStoreModule(store, namespaces, { isProperty: true })
 
     // Replace state since it's a function
-    mergeProperty(storeModule, moduleData, property)
+    mergeProperty(moduleStore, moduleData, property)
     return
   }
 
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/store.js
```diff
@@ -94,10 +94,10 @@ function resolveStoreModules (moduleData, filename) {
   // If src is a known Vuex property
   if (VUEX_PROPERTIES.includes(moduleName)) {
     const property = moduleName
-    const storeModule = getStoreModule(store, namespaces, { isProperty: true })
+    const moduleStore = getStoreModule(store, namespaces, { isProperty: true })
```
clarkdo, line 97: What about `propertyStoreModule` ?

Kapcash: Why not! It's better actually haha
Gonna change that soon

Kapcash: Corrected :)

pi0 (approved):
LGTM. Thanks @Kapcash :+1:

PR #7985, opened 2020-08-28
types: expose `Inject` type

Kapcash:
## Types of changes

## Description

When using Typescript, I usually declare my Nuxt plugins as so:

```
import { Context, Plugin } from '@nuxt/types'
const MyPlugin: Plugin = (ctx: Context, inject: (key: string, value: any) => void) => {})

export default MyPlugin
```

I have to define the `inject` function type myself. I think it would be better to actually export the existing type!
This way, if the type changes one day, I won't have to change the definition in all my plugins.

Now it looks like:
```
import { Context, Plugin, Inject } from '@nuxt/types'
const MyPlugin: Plugin = (ctx: Context, inject: Inject) => {})

export default MyPlugin
```

What do you think?

## Checklist:

### DIFF
--- a/packages/types/app/index.d.ts
+++ b/packages/types/app/index.d.ts
@@ -50,7 +50,8 @@ export interface Context {
 }
 
 export type Middleware = string | ((ctx: Context, cb: Function) => Promise<void> | void)
-export type Plugin = (ctx: Context, inject: (key: string, value: any) => void) => Promise<void> | void
+export type Inject = (key: string, value: any) => void
+export type Plugin = (ctx: Context, inject: Inject) => Promise<void> | void
 
 export interface Transition {
   name?: string


pi0:
Thanks!

danielroe:
@Kapcash Just as a note, you shouldn't need to define the types of the function arguments. This should be equally well-typed:

```ts
import { Plugin } from '@nuxt/types'
const MyPlugin: Plugin = (ctx, inject) => {
  // ctx and inject are both typed here...
})

export default MyPlugin
```

pi0:
@danielroe Agree. Type inference should works too :)

Kapcash:
That is very true! Thanks for pointing it out.
Are you still considering merging it? As it's a bit useless actually haha

pi0:
@Kapcash already merged. At least function is internally named now 😊

Kapcash:
Could have been reverted as the release is not done yet.
Of course, I'll focus on useful types next time ;)

Thanks for your reactivity on PR by the way!

Issue #7990, opened 2020-08-29
process.env.TOKEN is undefined


JoseJuan81:
### Versions

- nuxt: 
- node: 

### Steps to reproduce
set `process.env.TOKEN =$2yr$10$tN3JmuvXZKGwi0l38A.SxUVemGrFJiyLunyqkOqOTE7LHsrSwMEmEBi` in `.env` file
console.log(process.env.TOKEN) in `nuxt.config.js`.
### What is Expected?
$2yr$10$tN3JmuvXZKGwi0l38A.SxUVemGrFJiyLunyqkOqOTE7LHsrSwMEmEBi
![Captura de Pantalla 2020-08-29 a la(s) 13 17 59](https://user-images.githubusercontent.com/21971589/91643572-17495880-e9fa-11ea-9d9a-7fbed47e730d.png)

### What is actually happening?
undefinedundefinedundefinedt.SxUVemGrFJiyLunyqkOqOTE7LHsrSwMEmEBi
![Captura de Pantalla 2020-08-29 a la(s) 13 18 44](https://user-images.githubusercontent.com/21971589/91643581-2f20dc80-e9fa-11ea-834b-8322bda91f00.png)

There was no error in nuxt 2.12.2 just when I update to 2.14.3.

f3ltron:
Hello @JoseJuan81. Show us your nuxt.config.js but i think it's because you don't put it on env config in nuxt

https://fr.nuxtjs.org/api/configuration-env/

danielroe:
Looks like this might be an issue with [the new runtimeConfig](https://nuxtjs.org/guide/runtime-config) interpreting the `$2yr` etc. as variables to be interpolated.

pi0:
Indeed it seems a bug in forked version of `dotenv-expand`. Will make a PR.

JoseJuan81:
![Captura de Pantalla 2020-09-01 a la(s) 09 29 13](https://user-images.githubusercontent.com/21971589/91864265-9f677200-ec35-11ea-9bd7-645134458456.png)

JoseJuan81:
```
require('dotenv').config();

export default {
  mode: 'universal',
  /*
  ** Headers of the page
  */
  head: {
    htmlAttrs: {
      lang: 'es',
    },
    title: '¡ Tu tienda virtual !',
    meta: [
      { charset: 'utf-8' },
      { name: 'viewport', content: 'width=device-width, initial-scale=1' },
[...]

PR #8013, opened 2020-09-03
fix(config): don't interpolate env variables with `undefined` value

pi0:
## Types of changes

## Description

fixing #7990 and #8013 by preserving variables with `undefined` value (note that this is not exactly how `dontenv-expand` works, they do replace with an empty string but since we already have a fork and has bug for `undefined` it can at least resolve current issue)

In rare cases to avoid conflicting part of token with env variables, it is adviced to escape like this:

```
TOKEN=\\$2yr\\$10\\$tN3JmuvXZKGwi0l38A.SxUVemGrFJiyLunyqkOqOTE7LHsrSwMEmEB
```

**Some ideas for future:**

-  support `Token="..."` syntax for `.env` to escape but it is being removed by dotenv ([here](https://github.com/motdotla/dotenv/blob/7301ac9be0b2c766f865bbe24280bf82586d25aa/lib/main.js#L51)) so not much chance unless they support it or we use a fork.
- We could only support `${...}` syntax for interpolation (suggested by @galvez) but not possible since we need a fork of `dotenv` and it is a breaking change
- We could only support interpolation for `runtimeConfig` not `dotenv` which is still a breaking change

## Checklist:


clarkdo (approved):

JoseJuan81:
Hello, How can I use this changes in my project? I have a TOKEN in my `.env` and I can't use it for server http calls

pi0:
@JoseJuan81 Please see https://github.com/nuxt/nuxt.js/issues/7990#issuecomment-690255123

PR #8014, opened 2020-09-03
 fix(config): avoid recursion when interpolating env

pi0:
## Types of changes

## Description

originally reported by @hartmut-co-uk in discord following config makes maximum call stack error:

```js
  privateRuntimeConfig: {
    API_SECRET: '${API_SECRET}'
  },
```

It is also reproducible with dotenv-expand itself when using variable as value. Since we are using forked version there is this PR but have to also fix for upstream.

## Checklist:


clarkdo (dismissed):
What about giving a warning message here?

pi0:
@clarkdo Making sense. I also improved logic to early stop to avoid duplicate values:

image

PR #8022, opened 2020-09-04
feat(generator): generate CSP hashes for inline styles and scripts in SPA mode

### LINKED ISSUE #6592: Issue with generating CSP policy meta tag for (static) SPAs
### Version

[v2.10.1](https://github.com/nuxt.js/releases/tag/v2.10.1)

### Reproduction link

[https://github.com/PedroD/nuxt-csp-test](https://github.com/PedroD/nuxt-csp-test)

### Steps to reproduce

The link presented is of a sample project created with Nuxt using `$ npx create-nuxt-app <project-name>`.

I made no changes to the source code, except for the `nuxt.config.js` file, where I added the `render:csp` property, and for the `package.json` I changed `nuxt` from 2.0.0 to 2.10.1.

In order to reproduce the error:

1. navigate to the project root folder in your terminal
2. run `npm run generate`
3. use `serve` to serve the `dist` folder (`npm i -g serve`, to install it) and then `serve dist/` to serve it)
4. open the page in the browser (usually served on `http://localhost:5000/` by default)

### What is expected ?

When you inspect the HTML, it is expected to see a meta tag in the `<head>` of the page containing the CSP policies defined, with hashes, and also all scripts and styles (incl.  inline) should have a hash referenced in their tags.

No meta tag is created in `<head>`, and no hashes are being referenced in scripts and styles.

It is as if the `render:csp` property is not even set in `nuxt.config.js`.

### Additional Info

Here is the contents of the `nuxt.config.js` for this sample project:

```
const ALLOWED_HOSTS = ['*.google-analytics.com', '*.doubleclick.net', '*.googleusercontent.com']

export default {
  mode: 'spa',

  render: {
    csp: {
      reportOnly: false,
      addMeta: true,
      hashAlgorithm: 'sha256',
      unsafeInlineCompatiblity: true,
      policies: {
        'default-src': ["'self'", 'https:', ...ALLOWED_HOSTS],
        'script-src': ["'self'", "'strict-dynamic'", 'https:'],
        'style-src': ["'self'", "'strict-dynamic'", 'https:'],
        'frame-src': [],
        'object-src': ["'none'"],
        'base-uri': ["'self"]
      /* "report-uri": [
        "https://sentry.io/api/<project>/security/?sentry_key=<key>",
      ], */
      }
    }
  },

  /*
  ** Headers of the page
  */
  head: {
    title: process.env.npm_package_name || '',
    meta: [
      { charset: 'utf-8' },
      { name: 'viewport', content: 'width=device-width, initial-scale=1' },
      { hid: 'description', name: 'description', content: process.env.npm_package_description || '' }
    ],
    link: [
      { rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }
    ]
  },
  /*
  ** Customize the progress-bar color
  */
  loading: { color: '#fff' },
  /*
  ** Global CSS
  */
  css: [
    'element-ui/lib/theme-chalk/index.css'
  ],
  /*
  ** Plugins to load before mounting the App
  */
  plugins: [
    '@/plugins/element-ui'
  ],
  /*
  ** Nuxt.js dev-modules
  */
  buildModules: [
    // Doc: https://github.com/nuxt-community/eslint-module
    '@nuxtjs/eslint-module'
  ],
  /*
  ** Nuxt.js modules
  */
  modules: [
    // Doc: https://axios.nuxtjs.org/usage
    '@nuxtjs/axios',
    '@nuxtjs/pwa'
  ],
  /*
  ** Axios module configuration
  ** See https://axios.nuxtjs.org/options
  */
  axios: {
  },
  /*
  ** Build configuration
  */
  build: {
    transpile: [/^element-ui/],
    /*
    ** You can extend webpack config here
    */
    extend (config, ctx) {
    }
  }
}
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c9929">#c9929</a>)</em></sub></div>

BenceSzalai:
## Types of changes

## Description

It is a modification to `generator.js`, so that it considers the [CSP options](https://nuxtjs.org/api/configuration-render/#csp)  and adds the Content-Security-Policy hashes for inline scripts and styles when building an SPA. 

When SPA is built the only solution to avoid CSP problems is to explicitly allow all inline scripts and styles (using the `unsafe-inline` directive). However this is considered bad practice and makes SPAs more vulnerable to XSS attacks.

Resolves: #6592

## Checklist:


packages/generator/src/generator.js
```diff
@@ -217,6 +218,77 @@ export default class Generator {
       consola.warn('HTML minification failed for SPA fallback')
     }
 
+    const { csp } = this.options.render
```
clarkdo, line 221: I think there is CSP string in response headers of `renderRoute` returned value, we may use it to generate `meta`.

BenceSzalai: I've checked. `this.nuxt.server.renderRoute(...)` returns an object with only two items:
* `html` (string): the rendered HTML
* `preloadFiles` Array of Objects, each representing one of the key JS files

No CSP headers are returned.

Afaics `renderRoute()` calls different renderer for SSR and SPA. While the `SSRRenderer`'s `render()` method [really returns](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-renderer/src/renderers/ssr.js#L285) `cspScriptSrcHashes`, the `SPARenderer`'s `render()` [only returns](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-renderer/src/renderers/spa.js#L174) the html and the files.

clarkdo: Sorry, csp in response only works in ssr request.

I think a proper way is supporting csp meta tag in spa renderer and also extract csp hash generation and getCspString code.

BenceSzalai: That was my first approach, but it is impossible, since the rendered HTML is minified after being returned [here](https://github.com/nuxt/nuxt.js/blob/dev/packages/generator/src/generator.js#L215) so any hashes calculated during generation will become invalid after minification. That is why I've put the CSP generation code into the generator, so it can act on the minified JS and CSS tags.

tewnut (approved):

danielroe:

PR #8030, opened 2020-09-07
fix(config): normalize `build.indicator: true` to default value

remigermain:
## Types of changes

## Description

when you set indicator to true in nuxt.config , the component nuxt-build-indicator build failed.
so when is set to true, we replace it by default value.
nuxt.config.js:
```js
build: {
   indicator: true,
   ...
}
```
output: nuxt-build-indicator.vue
```js
export default {
  ...
  computed: {
     // options is generate by template
     // options: () => (<%= JSON.stringify(buildIndicator) %>),
      options: () => (true),
      indicatorStyle () {
          const [d1, d2] = this.options.position.split('-') // error when access position in bollean
       ...
     }
  }
 ...
}
```

## Checklist:

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -474,7 +474,7 @@ export function getNuxtConfig (_options) {
 
   // Indicator
   // Change boolean true to default nuxt value
-  if (typeof options.build.indicator === 'boolean' && options.build.indicator) {
+  if (options.build.indicator === true) {
     options.build.indicator = nuxtConfig.build.indicator
   }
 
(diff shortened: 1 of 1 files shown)


packages/config/src/options.js
```diff
@@ -472,6 +472,12 @@ export function getNuxtConfig (_options) {
 
   // ----- Builtin modules -----
 
+  // Indicator
+  // Change boolean true to default nuxt value
+  if (typeof options.build.indicator === 'boolean' && options.build.indicator) {
```
clarkdo, line 477: Can be `options.build.indicator === true`

pi0:
Thanks for PR @remigermain :heart:

PR #8055, opened 2020-09-11
fix(vue-app): scrollRestoration hasn't set

Kolobok12309:
Try catch block for setting `window.history.scrollRestoration`

## Types of changes

## Description
We have a lot of bug-reports with error `Cannot set property scrollRestoration of [object History] which has only a getter`, this not critical setter
For fix it, i add try/catch block, while setting this prop

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -667,5 +667,5 @@ export function isSamePath (p1, p2) {
 export function setScrollRestoration (newVal) {
   try {
     window.history.scrollRestoration = newVal;
-  } catch() {}
+  } catch(e) {}
 }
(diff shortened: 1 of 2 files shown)


packages/vue-app/template/utils.js
```diff
@@ -663,3 +663,9 @@ export function stripTrailingSlash (path) {
 export function isSamePath (p1, p2) {
   return stripTrailingSlash(p1) === stripTrailingSlash(p2)
 }
+
+export function setScrollRestoration (newVal) {
```
clarkdo, line 667: Don't think we need this function, any benifis for not setting `scrollRestoration` as we already have `'scrollRestoration' in window.history` checking ?

Kolobok12309: 'scrollRestoration' in window.history` not check for setter of this property.
Angular do this [function](https://github.com/angular/angular/pull/30630/files#diff-bc2334a81b493353fc53a822bec60f8cR255), but i think simple try/catch simpler

clarkdo (approved):

clarkdo:
@Kolobok12309 Thanks!

PR #8068, opened 2020-09-13
fix(vue-app): use mixin to provide `this.$nuxt`

### LINKED ISSUE #7696: Nuxt server $nuxt.context changed by latter request

### Versions

- nuxt: 2.10.2
- node: v8.11.2

### Steps to reproduce

1. create a nuxt project

./pages/index.vue
./pages/B.vue

```js
<template>
  <div>
    <p>{{ $route.name }}</p>

    <Nuxt />
  </div>
</template>
<script>
export default {
  created() {
    setTimeout(() => {
    }, 1000)
  },
}
</script>
```

### What is Expected?

1 second later, print out  
```bash
-------- index index
-------- B B
```

1 second later

```bash
-------- index index
-------- B B
```

```bash
-------- index B
-------- B B
```
### Conclusion

danielroe:
## Types of changes

## Description
avoids prototype pollution on the Vue prototype

closes #7696

## Checklist:

### DIFF
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -100,8 +100,11 @@ export default {
     // Add this.$nuxt in child instances
     const vm = this
     Vue.mixin({
-      created() {
-        this.<%= globals.nuxt %> = vm
+      beforeCreate() {
+        Object.defineProperty(this, '<%= globals.nuxt %>', {
+          get: () => vm,
+          configurable: true,
+        })
       }
     })
     if (process.client) {
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -98,7 +98,12 @@ export default {
   },
   created () {
     // Add this.$nuxt in child instances
-    Vue.prototype.<%= globals.nuxt %> = this
+    const vm = this
+    Vue.mixin({
+      created() {
+        this.<%= globals.nuxt %> = vm
+      }
+    })
     if (process.client) {
       // add to window so we can listen when ready
       window.<%= globals.nuxt %> = <%= (globals.nuxt !== '$nuxt' ? 'window.$nuxt = ' : '') %>this
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -99,12 +99,16 @@ export default {
   created () {
     // Add this.$nuxt in child instances
     const vm = this
+    function setGlobalNuxt() {
+      Object.defineProperty(this, '<%= globals.nuxt %>', {
+        get: () => vm,
+        configurable: true,
+      })
+    }
+    setGlobalNuxt.call(this)
     Vue.mixin({
       beforeCreate() {
-        Object.defineProperty(this, '<%= globals.nuxt %>', {
-          get: () => vm,
-          configurable: true,
-        })
+        setGlobalNuxt.call(this)
       }
     })
     if (process.client) {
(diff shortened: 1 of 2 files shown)


packages/vue-app/template/App.js
```diff
@@ -98,7 +98,12 @@ export default {
   },
   created () {
     // Add this.$nuxt in child instances
-    Vue.prototype.<%= globals.nuxt %> = this
+    const vm = this
+    Vue.mixin({
```
pi0, line 102: Shall we move this mixin to a global scope with a getter proxy? (similar to plugins)

danielroe: Good idea - so only evaluated on access. This is also still not fully working as we need to provide for layout components as well.

danielroe: I've tried various combinations of defining this on the Vue prototype (even the plugin method) but they all reproduce the original bug.

Damianos02 (commented):
~~[http://192.168.100.29](url)~~

packages/vue-app/template/App.js
clarkdo: @danielroe I've an idea about this, can I update your pr and let you review ? Feel free to revert if it's not proper.

danielroe: Yes, do! @clarkdo

clarkdo: @danielroe @pi0 Done, can you review ?

pi0:
@clarkdo @danielroe Can we add a SSR test to ensure this fix works?

danielroe:
Are we okay with prototype pollution outside of a component instance? (I'm also aware that injected objects currently also appear to be shared across requests but I haven't yet verified that...)

clarkdo:
> Are we okay with prototype pollution outside of a component instance?
Actually I don't think is a prototype pollution as it's a very normal way in [Vue plugin](https://vuejs.org/v2/guide/plugins.html#Writing-a-Plugin), the issue is that different ssr request is sharing same Vue reference.

Kronhyx (approved):

clarkdo:
Hi @danielroe , have you verified this fix in your project ? Are we OK to merge ?

danielroe:
@clarkdo Looks good to me - apologies for the delay. (Didn't realise you were waiting on me :facepalm:)

clarkdo:
@danielroe No worries, thanks for the confirmation, I'll merge it now.

pi0:
@clarkdo @danielroe I think we shouldn't merge this PR. Mixin method is probably replaceable with simple getter and SSR tests aren't added

clarkdo:
> SSR tests aren't added

I'll revert the pr.

> Mixin method is probably replaceable with simple getter

Correct me if I misunderstood, I think we can't use getter like below, because it will still get new reference.

```
get nuxt() {
  return Vue.prototype.nuxt
}

danielroe:
> Correct me if I misunderstood, I think we can't use getter like below, because it will still get new reference.
> 
> ```js
> get nuxt() {
>   return Vue.prototype.nuxt
> }
> ```
This is correct - it results in the same original issue if we are using getters in the way that the plugins do.

clarkdo:
Reverted and new pr is here https://github.com/nuxt/nuxt.js/pull/8132

pi0:
@danielroe Not directly returning same reference :see_no_evil: But we can pass and read same reference from instance options (similar to plugins `inject`) this way we don't need a new mixin to be called for every component :)

clarkdo:
@pi0 What about adding `$root....` to `Vue.prototyp.nuxt getter` ?

PR #8085, opened 2020-09-16
fix(babel-preset-app): allow to specify `corejs.version` as string

Yihao-G:
## Types of changes

## Description

According to the [`core-js` documentation](https://github.com/zloirock/core-js#babelpreset-env), it's recommended to specify used minor `core-js` version:
> **Warning!** Recommended to specify used minor `core-js` version, like `corejs: '3.6'`, instead of `corejs: 3`, since with `corejs: 3` will not be injected modules which were added in minor `core-js` releases.

More examples in their documentation: 
> By default, `@babel/preset-env` with `useBuiltIns: 'usage'` option only polyfills stable features, but you can enable polyfilling of proposals by `proposals` option, as `corejs: { version: '3.6', proposals: true }`.

However, with the current implementation of `babel-preset-app`, only integers can be set as they are used as the key of `coreJsMeta`. This PR enables users to specify minor `core-js` versions by handling minor version string/number. 

## Checklist:

  > `babel-preset-app` doesn't have any tests...

### DIFF
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -15,8 +15,16 @@ const coreJsMeta = {
   }
 }
 
+function getMajorVersion (version) {
+  if (typeof version === 'number') {
+    return Math.floor(version)
+  } else {
+    return Number(version.split('.')[0])
+  }
+}
+
 function getDefaultPolyfills (corejs) {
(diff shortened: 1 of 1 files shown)


pi0:
Hi @Yihao-G Actually it is better to use [build.corejs](https://nuxtjs.org/guides/configuration-glossary/configuration-build#corejs) option to specify version and we do it by default. Still your PR seems a good fix :+1:

PR #8132, opened 2020-09-30
fix(vue-app): use mixin to provide `this.$nuxt` #8068

### LINKED ISSUE #7696: Nuxt server $nuxt.context changed by latter request

clarkdo:
## Description

closes #7696

Continue work on #8068 (author @danielroe)

### TODO:

- [ ] Use `defineProperty` with a getter
- [ ] Test

### DIFF
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -24,6 +24,14 @@ if (!Vue.__nuxt__fetch__mixin__) {
 }
 <% } %>
 
+Vue.mixin({
+  beforeCreate() {
+    if(Vue.prototype.<%= globals.nuxt %>) {
+      this.<%= globals.nuxt %> = Vue.prototype.<%= globals.nuxt %>
+    }
+  }
+})
+
 // Component: <NuxtLink>
 Vue.component(NuxtLink.name, NuxtLink)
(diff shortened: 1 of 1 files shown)


galvez:
Wondering if passing `nuxt` as instance option + prototype getter would work...

```js
if (!Vue.prototype.$nuxt) {
    get () {
      return this.$root.$options.nuxt
    }
  })
}
```

clarkdo:
Yes, we talked here https://github.com/nuxt/nuxt.js/pull/8068#issuecomment-701339239
Haven’t got a chance to update and add test.

danielroe:
closes #7696

pi0:
closing in favor of #8170

PR #8170, opened 2020-10-08
fix(vue-app): use getter to provide `this.$nuxt`

### LINKED ISSUE #7696: Nuxt server $nuxt.context changed by latter request

danielroe:
## Types of changes

## Description
This is a redo of #8068 

**TODO**
- [x] Use defineProperty with a getter
- [x] Test

Closes #7696, #8132 

## Checklist:


danielroe:
@clarkdo Very happy for you to make changes if you have a different approach.

packages/vue-app/template/App.js
```diff
@@ -98,7 +98,18 @@ export default {
   },
   created () {
     // Add this.$nuxt in child instances
-    Vue.prototype.<%= globals.nuxt %> = this
+    if (process.server) {
+      this.$root.$options.<%= globals.nuxt %> = this
+      Object.defineProperty(Vue.prototype, '<%= globals.nuxt %>', {
```
pi0, line 103: I think other than options assignment, this part can be moved in a global context for defining getter

packages/vue-app/template/App.js
```diff
@@ -98,7 +98,18 @@ export default {
...
+        get() {
+          return this.$root.$options.<%= globals.nuxt %>
+        },
+        configurable: true
+      })
+    } else {
+      Vue.prototype.<%= globals.nuxt %> = this
```
pi0, line 110: Why different behavior for client-side?

test/dev/context.ssr.test.js
```diff
@@ -0,0 +1,34 @@
...
+  beforeAll(async () => {
+    const options = await loadFixture('context')
+    nuxt = new Nuxt(options)
+    await nuxt.ready()
+
+    port = await getPort()
+    await nuxt.server.listen(port, '0.0.0.0')
+  })
+
+  test('does not share state', async () => {
```
pi0, line 17: Shall we merge it to `ssr` test/fixture?

packages/vue-app/template/App.js
danielroe: I think if we're moving into global context - which is better - then we don't need different behaviour because we won't be reassigning the property.

pi0 (dismissed):
LGTM. Thanks @danielroe @clarkdo :heart:

clarkdo (dismissed):
Thanks @danielroe

bissolli:
Just wanted to add my tip on this one since I am flying around with the issue!
I've tried the proposed solution in a real application suffering with this issue and after a heavy round of tests everything went well!

Thanks for that @danielroe !

The details of my case can be found [here](https://github.com/nuxt/nuxt.js/issues/7696#issuecomment-704771876)

DaniilIsupov:
I used to be able to write `const vm = Vue.prototype.$nuxt;` and access context not from Vue component. (I needed this in order to have access to context from my utils functions, so as not to pass the context through the function's arguments)
But now I get the following error `TypeError: Cannot read property '$options' of undefined`.
Tell me what I should do in this situation?

pi0:
@DaniilIsupov Actually `$nuxt` instance is context dependent hence shouldn't be shared on a global variable and utils need to accept it directly. If you can explain more about your use case in a new issue or discussion, we may find a better solution.

DaniilIsupov:
> @DaniilIsupov Actually `$nuxt` instance is context dependent hence shouldn't be shared on a global variable and utils need to accept it directly. If you can explain more about your use case in a new issue or discussion, we may find a better solution.

I need to have access to `context` from my utils functions. For example to access `vuex` or `i18n` etc. But I don't want to throw the `context` through the `function arguments` all the time, tell me the best solution to this problem.

danielroe:
@DaniilIsupov It's difficult to answer that question because it's really about how to architect your app - and therefore requires more information about your app.

However, in general, one would want to avoid sharing state between requests. Injecting helpers into the Nuxt context is often a good way to achieve that. You could also use classes or curried functions to bind a helper to a given context/request. Note that `$nuxt` is still available on the component instance, just not on the Vue prototype.

pseudo-god:
> @DaniilIsupov It's difficult to answer that question because it's really about how to architect your app - and therefore requires more information about your app.
> 
> However, in general, one would want to avoid sharing state between requests. Injecting helpers into the Nuxt context is often a good way to achieve that. You could also use classes or curried functions to bind a helper to a given context/request. Note that `$nuxt` is still available on the component instance, just not on the Vue prototype.

I also have such confusion, do not know how to use!

TiagoDanin:
`plugin.js`
image

`web`
image


I call `Vue.prototype.$nuxt` and i have the same problem (`TypeError: Cannot read property '$options' of undefined`),
@pseudo-god, @DaniilIsupov, solution:
Replace all `Vue.prototype.$nuxt` to `this.$nuxt`.

or

```
		if (this.$nuxt) {
			// call my function in new version
			doSomething(this.$nuxt)
		} else if (Vue.prototype.$nuxt) {
			// call my function in old version
			doSomething(Vue.prototype.$nuxt)
		}
```

danielroe:
@TiagoDanin What are you trying to access? You might be able to get what you want from `ctx.app`

TiagoDanin:
> @TiagoDanin What are you trying to access? You might be able to get what you want from `ctx.app` - or just `ctx`. [Details here](https://nuxtjs.org/docs/2.x/internals-glossary/context#app).
But he doesn't have all the context, as Vue Meta.

danielroe:
@TiagoDanin What are you trying to access?

TiagoDanin:
> What are you trying to access?

Vue Meta Context, in old version `Vue.prototype.$nuxt` and in new version `this.$nuxt.$options.head`:
I have no problem because I put a conditional to check, but this was not to be removed in a patch version, because it broke applications of me and other members who use other contexts via `Vue.prototype.$nuxt.*`.
As reported by two more people above, was just indicated the solution.

https://github.com/TiagoDanin/Nuxt-SEO/blob/f8a0fab5c533a25cd65b2b9aaf92ce2609f5ee1c/lib/plugin.js#L26-L32

ceilfors:
@danielroe @pi0 I'm experiencing a similar issue where I hit the error `TypeError: Cannot read property '$options' of undefined`. I'd be grateful if you have recommendation on how we could properly retrieve the runtime config without my hack. (Also thanks for maintaining this great framework!)

My use case: I'm trying to retrieve Nuxt runtime config on the server side on static page generation BUT not within the Vue component declaration. The minimum examples can be found below:

**Example 1: Using runtime config to create a mixin**

```js
<script>
const $config = getConfig(this);

const algoliaClient = algoliasearch(
  $config.ALGOLIA_APPLICATION_ID,
  $config.ALGOLIA_SEARCH_API_KEY
);

const algoliaIndex = algoliaClient.initIndex($config.ALGOLIA_INDEX_NAME);

const { instantsearch, rootMixin } = createInstantSearch({
  searchClient: algoliaClient,
  indexName: $config.ALGOLIA_INDEX_NAME,
});


export default {
  mixins: [rootMixin], // This needs to be created with runtimeConfig!
}
</script>
```

**Example 2: Vue.use requires runtime config**:

```js
<script>
const $config = getConfig();

Vue.use(VueGoogleMaps, {
  load: {
    apiKey: $config.SOM_WEBAPP_GOOGLE_MAPS_API_KEY,
    libraries: ["places"],
  },
});

export default {
  props: {
  },
  data() {
  },
};
</script>
```

**The runtime config hack!**

```js
export const getConfig = () => {
  return (
    // Retrieve from Nuxt context if available (server)
    (Vue.prototype.$nuxt && Vue.prototype.$nuxt.context.$config) ||
    // Retrieve from Nuxt runtime config if available (browser)
    (window.__NUXT__ && window.__NUXT__.config)
  );
};
```

Issue #7722, opened 2020-07-16
optional chaining code makes error in SSR step


njh7799:
### Versions

- nuxt: v2.13.3
- node: v14.5.0 

### Reproduction

codesandbox does not work well :(

```vue
// index.vue
<template>
  <div>
    <nuxt-link to="/about">about</nuxt-link>
    <div>{{ fullName }}</div>
  </div>
</template>

<script>
export default {
  computed: {
    fullName() {
      const a = { name: { firstName: 'Jack', lastName: 'Nam' } };
      const fullName = `${a.name?.firstName} ${a.name?.lastName}`;
      return fullName;
    },
  },
};
</script>
```

```vue
// about.vue
<template>
  <nuxt-link to="/">index</nuxt-link>
</template>
```

### Steps to reproduce
1. use create-next-app to create project
2. create file named `index.vue` and `about.vue` in the pages directory
3. write down code above
4. run `npm run dev` in terminal
5. access `127.0.0.1:3000` 

### What is Expected?
<img src="https://user-images.githubusercontent.com/40619551/87642849-08c61b00-c785-11ea-8214-ea8757ba503a.png" width="300px"/>

**[Fig. 1]**

<br/><br/>

### What is actually happening?
<img src="https://user-images.githubusercontent.com/40619551/87641568-40cc5e80-c783-11ea-8294-5d95baffc9b8.png" width="600px"/>

**[Fig. 2]**
<br/><br/>
When I access to `127.0.0.1:3000` by SSR (entering address directly),  the error page(Fig. 2) came out. However  when I access by CSR, the right page(Fig. 1) came out.

I think babel does not work properly in SSR.

\+ when I ran `npm run build` command, similar error came out in terminal
```shell
ERROR in ./pages/index.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib??ref--2-0!./node_modules/@nuxt/components/dist/loader.js??ref--0-0!./node_modules/vue-loader/lib??vue-loader-options!./pages/index.vue?vue&type=script&lang=js&) 26:33
Module parse failed: Unexpected token (26:33)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/@nuxt/components/dist/loader.js
 * ./node_modules/vue-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|         }
|       };
>       const fullName = `${a.name?.firstName} ${a.name?.lastName}`;
|       return fullName;
[...]

PR #8203, opened 2020-10-14
fix(babel-preset-app): always transpile optional chaining and nullish-coalescing for server

pi0:
## Types of changes

## Description

Since node@14 supports optional chaining, `@babel/preset-env` disables transpilation for this target (we use runtime node version as target). But this makes troubles (#7722) as webpack@4 uses acorn 6.x which doesn't support this syntax (webpack/webpack#10227)

### Alternative

We could strictly set node target to 8 or 10 to avoid same happening in the future with acorn and nuxt@2

## Checklist:

### DIFF
--- a/packages/babel-preset-app/package.json
+++ b/packages/babel-preset-app/package.json
@@ -14,6 +14,8 @@
     "@babel/helper-compilation-targets": "^7.10.4",
     "@babel/plugin-proposal-class-properties": "^7.10.4",
     "@babel/plugin-proposal-decorators": "^7.10.5",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.10.4",
+    "@babel/plugin-proposal-optional-chaining": "^7.11.0",
     "@babel/plugin-transform-runtime": "^7.11.5",
     "@babel/preset-env": "^7.11.5",
     "@babel/runtime": "^7.11.2",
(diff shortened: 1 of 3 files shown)


clarkdo (commented):
I think adding this plugin in babel-preset-app only for server env would be better.

pi0:
@clarkdo Indeed but as mentioned i tried this: (on an external project)

image

This doesn't works (probably preset-env somehow takes over?)

And what do you think about fixing target to 10 or 12?

clarkdo:
@pi0 What about add it in `include` of `preset-env` ? I can take a look tonight if it doesn't work.

```js

include: ['proposal-optional-chaining']
```

pi0:
@clarkdo Actually it was babel cache that needed to be manually invalidated :see_no_evil: Seems all fine now

clarkdo (dismissed):
Nice

marcvangend:
I'm running nuxt 2.14.7 now (with node 15.1.0, yarn 1.2.10), but I'm still getting a fatal error when I try to use a nullish coalescing operator or optional chaining in a generate function in my nuxt.config.js:

```
 FATAL  Invalid or unexpected token                                                                                                          15:20:32

  const foo = data?.foo ?? 'bar'
  ^
  
  SyntaxError: Invalid or unexpected token
  at Object.Module._extensions..js (node:internal/modules/cjs/loader:1112:10)


   ╭──────────────────────────────────────────────╮
   │                                              │
   │   ✖ Nuxt Fatal Error                         │
   │                                              │
   │   SyntaxError: Invalid or unexpected token   │
   │                                              │
   ╰──────────────────────────────────────────────╯
```

Is this expected behavior? Am I making a mistake somewhere? Should I open a new issue?

PR #8205, opened 2020-10-14
fix(webpack): allow transpiling package in nested `node_modules`

padinko:
## Types of changes

## Description
When I need to transpile `[PROJECT_PATH]/node_modules/lowlight/node_modules/highlight.js/...` but not transpile `/lowlight/` itself, it is imposible, because split only return text between `node_modules` ( `/lowlight/` in this example)

this change leave full path after first `node_modules`, not only path between 2 `node_modules` ( `/lowlight/node_modules/highlight.js/...` in this example)

## Checklist:

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -316,7 +316,7 @@ export default class WebpackBaseConfig {
       {
         test: /\.m?jsx?$/i,
         exclude: (file) => {
-          file = file.split('node_modules', 2)[1]
+          file = file.split(/node_modules(.*)/)[1]
 
           // not exclude files outside node_modules
           if (!file) {


clarkdo (commented):
The change look OK.

But I think we need to consider it this is a proper fix in your scenario, I think a bigger issue from your description is: your app is having inconsistent version of `highlight.js`, so `highlight.js` is installed under `lowlight/node_modules`.

I suggest adding `highlight.js` to dependencies explicitly and remove inconsistent version, then `highlight.js` will be installled under `node_modules`

padinko:
@clarkdo 
Hi, for me it is imposible to have highlight not nested:
- `@vue/cli-service` required `cli-highlight` which requires `highlight.js@9`
- `tiptap-extensions` requires `lowlight` which requires `highlight.js@10`

we need both this packages, and they are not using same version, so npm install one of them to nested node_modules (standard behavior)

clarkdo:
> for me it is imposible to have highlight not nested:

Sorry, what I mean is we should always add dependencies explicitly.

Like your case, if you're using `highlight.js v9` in you app pages, you should have it in dependencies, and it will be hoisted to root node_modules, `@vue/cli-service` will also work as its dependencies will be nested but you won't use it in your app pages or components.

ghost (approved):

pi0 (approved):
Change seems good. Still it is highly recommended to explicitly include dependencies in `package.json`. Since there is no proper reproduction, if it makes regressions we may need to revert in a patch.

PR #8206, opened 2020-10-14
fix: add better logs for DX with target option

atinux:
## Types of changes

## Description

This is an improvement for developer experience when working with static target vs full static when working with `nuxt build`  or `nuxt generate`.

## Checklist:


packages/cli/src/commands/build.js
```diff
@@ -78,6 +79,7 @@ export default {
     // TODO: remove if in Nuxt 3
     if (nuxt.options.mode === MODES.spa && nuxt.options.target === TARGETS.server && cmd.argv.generate !== false) {
       // Build + Generate for static deployment
+      consola.info(`Enforcing ${chalk.bold.blue('static')} target because of \`nuxt build\` with server-side rendering disabled`)
```
pi0, line 82: Maybe also add a warning to use `nuxt generate` and this will be unsupported?

packages/cli/src/commands/generate.js
```diff
@@ -75,6 +77,7 @@ export default {
 
     // Forcing static target anyway
     config.target = TARGETS.static
+    consola.info(`Enforcing ${chalk.bold.blue('static')} target because of \`nuxt generate\` command`)
```
pi0, line 80: Maybe also changing it to an advice to use `target: 'static'` as it will be default behavior?

packages/generator/src/generator.js
```diff
@@ -16,6 +16,11 @@ export default class Generator {
 
     // Set variables
     this.isFullStatic = isFullStatic(this.options)
+    if (this.isFullStatic) {
```
pi0, line 19: If there is already warning about legacy behavior and that should be avoided, do we still need both infos?

packages/cli/src/commands/generate.js
pi0: Probably also with a short-link to https://nuxtjs.org/blog/going-full-static? :D

packages/cli/src/commands/build.js
Atinux: Good idea

packages/generator/src/generator.js
debs-obrien: this would be nice as enforcing static target because of nuxt generate with full static disabled and target static. Perhaps a hint in order to activate full static update to Nuxt v 2.14

Atinux: We don't display a warning for target server and nuxt generate since we kept the old behaviour to not-mock asyncData & fetch.

packages/cli/src/commands/build.js
```diff
@@ -80,6 +80,7 @@ export default {
     if (nuxt.options.mode === MODES.spa && nuxt.options.target === TARGETS.server && cmd.argv.generate !== false) {
       // Build + Generate for static deployment
       consola.info(`Enforcing ${chalk.bold.blue('static')} target because of \`nuxt build\` with server-side rendering disabled`)
+      consola.warn('This feature will be removed in Nuxt 3, please use the `nuxt generate` command.')
```
debs-obrien, line 83: I like this 💚

pi0 (approved):

PR #8213, opened 2020-10-15
feat(builder): added support for custom layout directories

### LINKED ISSUE #7879: Multiple Layout Directories
At the moment it is possible to override what directory nuxtjs looks for layout components in as demonstrated below:
```js
//nuxt.config.js

export default {
    ...,
    dir: {
        layouts: 'custom_dir/layouts'
    }
}
```

It would be useful in some scenarios to be able to specify multiple directories, for example:
```js
//nuxt.config.js

export default {
    ...,
    dir: {
        layouts: [
            'custom_dir/layouts',
            'other_custom_dir/layouts'
        ]
    }
}
```

Then when specifying a layout in a page component, it should check the first directory for the layout and if it can't be found it will check the next directory.

This could also be useful for having layouts in both the application itself as well as from other modules. Alternatively, a method of creating additional layouts that could be loaded from modules could be useful.

rotem-bar:
## Types of changes

## Description

Resolves #7879,
Adds support for string array type to the dir.layouts setting, added related tests and I assume it requires changes to the nuxt config doc page - even though it does not break any existing setting, just adds support for arrays as value (please confirm and I will create the appropriate PR if needed)

## Checklist:


Atinux:
Hi @rotem-bar and thank you very much for taking the time of making this nice PR!

I believe the best approach to resolve the issue is to use a module to add layouts directly (using `this.addLayout`, see https://github.com/nuxt/nuxt.js/blob/dev/packages/core/src/module.js#L86-L102).

Otherwise it adds more complexity, why could we have multiple layouts dir but not pages, plugins, etc 😬

rotem-bar:
Hey @Atinux, I do agree with your module approach and the inconsistency this PR will make in the `dir` setting, however I do believe that the `dir` setting should support arrays for every setting by default (not only layouts) and not force you to make a module, read the directory, add every layout in directory - just for supporting an extra layout directory where you don't need to manually add every file in the directory yourself.

I can look into expanding this PR to not only layouts but include every setting in the `dir` object if thats something you think would help solve the inconsistencies.

pi0:
Hi @rotem-bar thanks for PR. Actually it totally makes sense to support multiple directories for auto-scanned dirs. 

This will be natively supported with nuxt3 (see https://github.com/nuxt/rfcs/issues/30)  but since then it might add extra complexity (#5818) to nuxt2 code base yet still with limitations.

Issue #8212, opened 2020-10-15
Option to generate route files without ".html"


camhart:
### Is your feature request related to a problem? Please describe.
If you attempt to deploy a nuxt static generated site to AWS s3 & Cloudfront, then Cloudfront *only* gets the correct prerendered page if the route in the URL matches the *exact* file name.

For example, stay I have a nuxt static generated site with the following routes:
* /
* /about
* /faq

If I have generate.subFolders set to true, then when using cloudfront, the follow URLs will load the prerendered page.
* mysite.com/index.html
* mysite.com/about/index.html
* mysite.com/faq/index.html

However when you hit those pages directly, the Vue/Nuxt app doesn't mount/hydrate properly..  There are no error messages--it simply just doesn't work.

However *if* you create a AWS Lambda Cloudfront Origin Request function, and have it redirect mysite.com/about to mysite.com/about/index.html, then the Vue/Nuxt app *does* mount/hydrate properly.  Here's an example of what that function looks like:

```
exports.handler = async (event, context) => {
    
    let request = event.Records[0].cf.request
    let olduri = request.uri
    
    let newuri = olduri.replace(/\/$/, '\/index.html')                  //redirect anything ending with /, to /index.html
    newuri = newuri.replace(/about$/, 'about/index.html')      //redirect /about to /about/index.html
    newuri = newuri.replace(/faq$/, 'faq/index.html')              //redirect /faq to /faq/index.html
    
    console.log(`old = ${olduri}, new = ${newuri}`)
    
    request.uri = newuri


    return request;
};
```

This resolves the issue and allows everything to work properly.  If you navigate to mysite.com/about everything works great.   **However it sucks to have to maintain that lambda function.** 

If you set generate.subFolders to false, the issue still exists.

### Describe the solution you'd like

Add a new generate property within nuxt.config.js, that allows you to specify the file extensions for the pages except for the root index.html.  That way you could do:

```
generate: {
  subFolders: false,
  nonRootPageExtension: '' //overrides from '.html' to nothing
}
```

Then when statically generated, the actual files would look like this
```
/index.html   -> cloudfront/s3 already has easy support for redirecting 403's to index.html.  Leave this as is since we can't have a file with no name for the root.
/about -> this about file would normally be "about.html", but because nonRootPageExtension is set to '', it's just "about"
/faq -> this about file would normally be "faq.html", but because nonRootPageExtension is set to '', it's just "faq"
```

### Describe alternatives you've considered

[...]

PR #8222, opened 2020-10-19
feat(generator): ability to customize non-root page extensions

lielfr:
## Types of changes

## Description
This PR adds a new parameter to the `generate` config section, called `nonRootPageExtension`. When specified and `subfolders` is set to false, it allows setting the extension of all the files, except index.html and 404.html. For example, setting it to an empty string would result in generating the files without any extension, as requested in issue #8212

## Checklist:

- [x] I have updated the documentation accordingly. (PR: 872 in the nuxtjs.org repo)

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -359,7 +359,8 @@ export default class Generator {
       fileName = fileName === '/404/index.html' ? '/404.html' : fileName // /404 -> /404.html
     } else {
       const normalizedRoute = route.replace(/\/$/, '')
-      fileName = route.length > 1 ? path.join(path.sep, normalizedRoute + '.html') : path.join(path.sep, 'index.html')
+      const extension = (typeof this.options.generate.nonRootPageExtension === 'undefined') ? '.html' : this.options.generate.nonRootPageExtension
+      fileName = route.length > 1 ? path.join(path.sep, normalizedRoute + extension) : path.join(path.sep, 'index.html')
     }
 
     // Call hook to let user update the path & html
--- a/packages/config/src/config/generate.js
+++ b/packages/config/src/config/generate.js
@@ -18,5 +18,6 @@ export default () => ({
     versionBase: undefined, // Default: "_nuxt/static/{version}""
     dir: 'static',
     version: undefined // Default: "{timeStampSec}"
-  }
+  },
+  nonRootPageExtension: undefined // Default: '.html'
 })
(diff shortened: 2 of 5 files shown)


clarkdo:
Doc pr: https://github.com/nuxt/nuxtjs.org/pull/872

packages/generator/src/generator.js
```diff
@@ -359,7 +359,8 @@ export default class Generator {
       fileName = fileName === '/404/index.html' ? '/404.html' : fileName // /404 -> /404.html
     } else {
       const normalizedRoute = route.replace(/\/$/, '')
-      fileName = route.length > 1 ? path.join(path.sep, normalizedRoute + '.html') : path.join(path.sep, 'index.html')
+      const extension = (typeof this.options.generate.nonRootPageExtension === 'undefined') ? '.html' : this.options.generate.nonRootPageExtension
```
clarkdo, line 362: `.html` can be directly put in `config/src/config/generate.js` as default option

lielfr: Yes, however for some reason some tests won't pass without that check

clarkdo: I can help you on test when I'm available.

clarkdo: BTW, you can easily solve your problem by using hook `generate:page`:

```js
// nuxt.config.js
export default {
  hooks: {
    generate: {
      page(page) {
        page.path = page.path.replace(/html^/, '...')
      }
    }
  }
}
```

lielfr: The original issue was not mine. Does this mean there is no need for such a configuration parameter?

clarkdo: > The original issue was not mine

I'll leave comment there.

> Does this mean there is no need for such a configuration paramete

I would love to hear opinion from @Atinux @pi0

pi0: I think so too. It is a really rare usage and it is possible to fix using hooks

pi0: Using either hooks or postprocessing seems better option than adding a new feature.

pi0:
Closing this PR as makes no sense adding such option. But let's continue discussion on #8212 for better way for S3 deployments :)

packages/generator/src/generator.js
lielfr: Sure, thanks.

PR #8263, opened 2020-10-29
fix(server): remove virtualConsole option when disabled

Perlover:
It's patch for [this feature](https://github.com/nuxt/nuxt.js/issues/8262)
May be it the fix not a feature

1) The `renderAndGetWindow` method can pass `virtualConsole` option as `false`
2) If we pass to `renderAndGetWindow` method our `virtualConsole` instance it will not be touched by nuxt

All unit tests are passed by `yarn test:unit`

### DIFF
--- a/packages/server/src/jsdom.js
+++ b/packages/server/src/jsdom.js
@@ -40,9 +40,13 @@ export default async function renderAndGetWindow (
   if (options.virtualConsole) {
     if (options.virtualConsole === true) {
       options.virtualConsole = new jsdom.VirtualConsole().sendTo(consola)
+      // Throw error when window creation failed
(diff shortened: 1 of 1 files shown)


packages/server/src/jsdom.js
```diff
@@ -40,9 +40,13 @@ export default async function renderAndGetWindow (
   if (options.virtualConsole) {
     if (options.virtualConsole === true) {
       options.virtualConsole = new jsdom.VirtualConsole().sendTo(consola)
+      // Throw error when window creation failed
+      options.virtualConsole.on('jsdomError', jsdomErrHandler)
```
clarkdo, line 44: This can be still in `if (options.virtualConsole) {` ?

PR #8280, opened 2020-11-02
fix(vue-app): prevent redirection loop with uri encoded path

### LINKED ISSUE #8116: ssrContext.url is not encoded, which leads to redirection loop if the URL has space
### Versions
- nuxt: v2.14.6
- node: v12.18.3

### Reproduction
https://github.com/unicrus/nuxt-ssr-redirection-loop

### Steps to reproduce
1. Run the app from https://github.com/unicrus/nuxt-ssr-redirection-loop
2. On the main page there are 2 buttons: `Redirect with no loop` and `Redirect with loop`.
3. Click `Redirect with no loop` — it works as expected — Nuxt redirection [loop guard](https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/vue-app/template/server.js#L53) works fine.
4. Go back to the main page and click `Redirect with loop`.
5. Observe a redirection loop (`localhost redirected you too many times`).

### What is Expected?
Nuxt should see, that the current route is the same as the target route and do not redirect the user.

Nuxt redirection [loop guard](https://github.com/nuxt/nuxt.js/blob/6ab65fbf4f2b0d2348363c43f5c2fe81ef47cbc6/packages/vue-app/template/server.js#L53) does not work because `ssrContext.url` is not encoded.

danielroe:
## Types of changes

## Description
This change decodes the requested path before comparing it with the ssrContext URL to prevent a redirect loop if the path requested has a uri-encoded string (e.g. `%20`) but the ssrContext URL (owing to redirection) has the decoded value.

Resolves: #8116

## Checklist:

### DIFF
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -50,7 +50,7 @@ const createNext = ssrContext => (opts) => {
     opts.path = urlJoin(routerBase, opts.path)
   }
   // Avoid loop redirect
-  if (opts.path === ssrContext.url) {
+  if (decodeURIComponent(opts.path) === ssrContext.url) {
     ssrContext.redirected = false
     return
   }
--- a/test/e2e/basic.browser.test.js
+++ b/test/e2e/basic.browser.test.js
@@ -319,6 +319,11 @@ describe('basic browser', () => {
     page.close()
   })
 
+  test('/redirection/no loop', async () => {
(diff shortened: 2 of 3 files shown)


pi0 (dismissed):

danielroe:
e2e tests should be passing once #8282 is merged

clarkdo (approved):

PR #8282, opened 2020-11-02
test: fix redirect tests

danielroe:
## Types of changes
- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
e2e and ssr redirect tests depended on a now-moved page on the nuxt docs ... which ironically was redirected, thereby breaking the test

## Checklist:
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

PR #8289, opened 2020-11-03
fix(vue-app): use nuxt globalName correctly in nuxt-link and fetch mixin

### LINKED ISSUE #8118: Reference in nuxt-link.client.js to this.$nuxt causes error when custom globalName is set
### Versions

- nuxt: v2.14.5
- node: v12.18.3

### Reproduction

I'm getting an error under circumstances it's hard to reproduce, but I'm pretty sure I see the the cause of the problem. 

If I set `globalName` in nuxt.config.js to a custom name, I get an error (cannot get property `fetchPayload` of undefined). 

The error is caused by the 2 lines that reference `this.$nuxt`: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt-link.client.js#L114-L115

I believe that changing `this.$nuxt` to `this.<%= globals.nuxt %>` (as it is in other places within that file) should fix the problem.

It's strange that it gets past the condition, though. Maybe webpack resolve `this.$nuxt` to `true` when it's being built, thus exposing `this.$nuxt.fetchPayload`?

danielroe:
## Types of changes

## Description
**History**: `this.$nuxt` accidentally crept in in the static target PR (https://github.com/nuxt/nuxt.js/pull/6159). This caused an error which was solved by testing for `this.$nuxt` rather than using `globalName` (https://github.com/nuxt/nuxt.js/pull/7766) as is done elsewhere in the component (see https://github.com/nuxt/nuxt.js/pull/4743)

This PR reverts to using `globalName`.

Resolves #8118

## Checklist:

### DIFF
--- a/packages/vue-app/template/components/nuxt-link.client.js
+++ b/packages/vue-app/template/components/nuxt-link.client.js
@@ -111,8 +111,8 @@ export default {
       // Preload the data only if not in preview mode
       if (!this.$root.isPreview) {
         const { href } = this.$router.resolve(this.to, this.$route, this.append)
-        if (this.$nuxt)
-          this.$nuxt.fetchPayload(href).catch(() => {})
+        if (this.<%= globals.nuxt %>)
+          this.<%= globals.nuxt %>.fetchPayload(href).catch(() => {})
       }
       <% } %>
       <% if (router.linkPrefetchedClass) { %>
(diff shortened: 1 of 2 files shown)


danielroe:
e2e tests should pass when #8282 is merged

clarkdo (approved):

kswedberg:
thank you!

PR #8314, opened 2020-11-07
fix(vue-app): handle missing payloads on full static

### LINKED ISSUE #7717: Full static mode dynamic pages payload 404 errors (i18n module)
### Versions

- nuxt: v2.13.3
- node: any

### Reproduction

- https://codesandbox.io/s/throbbing-feather-v7nnx
- https://v7nnx.sse.codesandbox.io/about/

<details open>
<br>
When using nuxt-i18n with full static generated mode there are still 404 errors for payload.js on dynamic pages.
</details>

Related https://github.com/nuxt/nuxt.js/issues/7559 https://github.com/nuxt/nuxt.js/issues/7648 https://github.com/nuxt/nuxt.js/issues/7622

### Steps to reproduce
Check codesandbox.

About page tries to pre-fetch "_nuxt/static/1594718554/**fr/test/payload.js**" and gets 404 error in console log.

### What is Expected?
Payload shouldn't be fetched because it wasn't generated at all.

Still 404 errors.

atinux:
## Types of changes

## Description

In full static, we generate a `payload.js` for each route, expect for ignored pages.

When navigating on client-side to a page not statically generated, it will anyway try to fetch the payload, leading to a 404 error in the console:

94172662-0ffb5a80-fe93-11ea-89b1-3f060d1ee988

This PR will create a `manifest.js` grouping all routes generated, so we can avoid fetching payloads for not generated routes.

Left to do:
- [x] Generate a `manifest.js` instead with `routes` key inside (will be useful later on to add more fields)
- [x] Check behaviour with `router.base` defined
- [x] Add an option `generate.manifest` to disable the behaviour when in serverless (payload can be generated on-demand)
- [ ] Update the full static fixture and test the behaviour (instead of manual testing)

Fixes #7717

## Checklist:


packages/vue-app/template/App.js
```diff
@@ -295,18 +295,37 @@ export default {
...
+      }
+      return (route.replace(/\/+$/, '') || '/').split('?')[0].split('#')[0]
+    },
+    getStaticAssetsPath(route = '/') {
+      const { staticAssetsBase } = window.<%= globals.context %>
+
+      return urlJoin(this.getRouterBase(), staticAssetsBase, this.getRoutePath(route))
+    },
+    async fetchStaticManifest() {
+      console.log(urlJoin(this.getStaticAssetsPath(), 'manifest.js'))
```
pi0, line 314: TODO to remove

packages/vue-app/template/App.js
```diff
@@ -295,18 +295,37 @@ export default {
...
     async fetchPayload(route) {
-      const { staticAssetsBase } = window.<%= globals.context %>
-      const base = (this.$router.options.base || '').replace(/\/+$/, '')
-      if (base && route.startsWith(base)) {
-        route = route.substr(base.length)
+      const manifest = await this.fetchStaticManifest()
+      const path = this.getRoutePath(route)
+      if (!manifest.routes.includes(path)) {
+        this.setPagePayload(false)
+        throw new Error(`Route ${path} is not pre-rendered`)
```
pi0, line 326: By throwing error, we should check manifest in NuxtLink as well to avoid prefetch of routes that are not pre-rendered

packages/vue-app/template/App.js
```diff
@@ -295,18 +295,37 @@ export default {
...
+      return (this.$router.options.base || '').replace(/\/+$/, '')
+    },
+    getRoutePath(route = '/') {
+      const base = this.getRouterBase()
+      if (base && route.startsWith(base)) {
+        route = route.substr(base.length)
+      }
+      return (route.replace(/\/+$/, '') || '/').split('?')[0].split('#')[0]
+    },
+    getStaticAssetsPath(route = '/') {
```
pi0, line 308: We may refactor `urlJoin` and `encodeURI` to `getStaticAssetFile(...path)` (usage: `getStaticAssetFile (route, 'payload.js')` || `getStaticAssetFile ('manifest.js')`

packages/vue-app/template/App.js
```diff
@@ -295,18 +295,37 @@ export default {
...
+      const { staticAssetsBase } = window.<%= globals.context %>
+
+      return urlJoin(this.getRouterBase(), staticAssetsBase, this.getRoutePath(route))
+    },
+    async fetchStaticManifest() {
+      console.log(urlJoin(this.getStaticAssetsPath(), 'manifest.js'))
+      return window.__NUXT_IMPORT__('manifest.js', encodeURI(urlJoin(this.getStaticAssetsPath(), 'manifest.js')))
```
pi0, line 315: Even if unnecessary, i would create `nuxtImport` method to wrap `window.__NUXT_IMPORT__` call

pi0:
One more suggestion: We shall preload `manifest.json` chunk in renderer (SPA and SSR)

packages/vue-app/template/App.js
Atinux: I think it could be move if used elsewhere (and I think could be used more as a module util)

packages/vue-app/template/App.js
Atinux: Same here, could be expose as a module util.

packages/vue-app/template/App.js
pi0: jsonp is mostly an internal API...

packages/generator/src/generator.js
```diff
@@ -31,6 +32,9 @@ export default class Generator {
       const { staticAssets } = this.options.generate
       this.staticAssetsDir = path.resolve(this.distNuxtPath, staticAssets.dir, staticAssets.version)
       this.staticAssetsBase = this.options.generate.staticAssets.versionBase
+      this.manifest = {
+        routes: []
```
pi0, line 36: Suggestion: `precached` or so?

pi0: Suggestion: We may also persist `staticAssets.version` here

Atinux: Actually the version is already in the staticAssetsBase

pi0: I know it is also in page payload. Keeping version here could be beneficial to simplify on demand payload generation but nothing necessary :)

danielroe:
It would be good to add the ability to check there is a newer version of code available on the server and (if so) refresh in the background.

pi0 (dismissed):

PR #8325, opened 2020-11-09
fix: handle route encoding

pi0:
## Types of changes

## Description

vuejs/vue-router#3350 introduced a breaking change which dropped supporting to encode uri paths. So we need to update internals doing it:

- Ensure `router.base` is always encoded
- Encode `path` param in route entries (generated code) (update #8421)
- Decode URIs for CLI messages for DX and avoid conflicting with util.format
- Override `router.resolve` with `encodeURI(decodeURI(to))` (props to @farnabaz)
- Update rest of internals
- New encoding test (local): `yarn nuxt dev test/fixtures/encoding`
  - Would be nice if we have more e2e tests later on specially for SPA and CSR

## Checklist:


clarkdo (dismissed):

pi0:
merging to try on edge

yarn.lock
```diff
@@ -13403,6 +13403,11 @@ vue-router@3.4.8:
   resolved "https://registry.npmjs.org/vue-router/-/vue-router-3.4.8.tgz#2c06261d35d8075893470352d42d70b6287b8194"
   integrity sha512-3BsR84AqarcmweXjItxw3jwQsiYNssYg090yi4rlzTnCJxmHtkyCvhNz9Z7qRSOkmiV485KkUCReTp5AjNY4wg==
 
+vue-router@^3.4.9:
+  version "3.4.9"
+  resolved "https://registry.npmjs.org/vue-router/-/vue-router-3.4.9.tgz#c016f42030ae2932f14e4748b39a1d9a0e250e66"
+  integrity sha512-CGAKWN44RqXW06oC+u4mPgHLQQi2t6vLD/JbGRDAXm0YpMv0bgpKuU5bBd7AvMgfTz9kXVRIWKHqRwGEb8xFkA==
+
```
danielroe, line 13410: should replace router above

packages/vue-app/template/server.js
```diff
@@ -7,7 +7,8 @@ import {
   <% if (features.middleware) { %>middlewareSeries,<% } %>
   <% if (features.middleware && features.layouts) { %>sanitizeComponent,<% } %>
   getMatchedComponents,
-  promisify
+  promisify,
+  ensureURIEncoded
```
danielroe, line 11:
```suggestion
  promisify
```

danielroe (approved):

yarn.lock
pi0: (handled in renovate PR)

packages/vue-app/template/server.js
pi0: Nice catch! a8cc430574ef02a7817da357fb9daf22de74778f

packages/server/src/server.js
```diff
@@ -167,7 +167,7 @@ export default class Server {
         prefix: false,
         handler: (req, res) => {
           const to = urlJoin(this.nuxt.options.router.base, req.url)
-          consola.info(`[Development] Redirecting from \`${req.url}\` to \`${to}\` (router.base specified).`)
+          consola.info(`[Development] Redirecting from \`${decodeURI(req.url)}\` to \`${decodeURI(to)}\` (router.base specified).`)
```
danielroe, line 170: I'm sure there's a reason - but is there any merit in normalising on the `req` object?

pi0: Actually it is a bad-request error for any http server with non-encoded URI. It is mostly to ensure not happening if `req` is not real `req` :)

PR #8346, opened 2020-11-14
feat: update all dependencies that require node 10.x

clarkdo:
Checklist: https://github.com/nuxt/nuxt.js/issues/8327


packages/webpack/src/builder.js
```diff
@@ -168,15 +168,19 @@ export class WebpackBundler {
     const buildOptions = this.buildContext.options.build
     const { client, ...hotMiddlewareOptions } = buildOptions.hotMiddleware || {}
 
+    compiler.options.watchOptions = this.buildContext.options.watchers.webpack
+    compiler.hooks.infrastructureLog.tap('webpack-dev-middlewre-log', (name) => {
```
kevcodez, line 172: Misspelled middleware, fyi

w3bdesign (dismissed):

pi0 (dismissed):

clarkdo:
@pi0 This pr will result in some config change because of the packages breaking change like `css-loader`, I know you mentioned release note, it would be a great idea that list all the points.

pi0:
@clarkdo Sure thing. Mainly merged to try on edge for a while and prepare list on same time. We may try to normalize some options probably?

clarkdo:
I’m not sure if we should as some packages has renamed or reorganized configs in a better way, I think aligning to their official config may be more consistent

pi0:
Just for backwards compatibility with a warning :)

clarkdo:
Good idea, I can try to add some.

PR #8347, opened 2020-11-14
feat(vue-renderer): allow adding async attribute to `<script>` tags

### LINKED ISSUE #8242: Add `async`  attribute to webpack scripts

Webpack seems to have fixed the issue when using `async`.

Add `defer` **and** `async`  to webpack scripts.

### Additional context

I tested on nuxtjs.org using hooks in `nuxt.config.js`:

```js
export default {
  hooks: {
    'render:route': function (url, result, context) {
      result.html = result.html.replace(/ defer>/g, ' defer async>')
    },
    'generate:page': (page) => {
      page.html = page.html.replace(/ defer>/g, ' defer async>')
    }
  }
}
```

An it works perfectly :)

danielroe:
## Types of changes

## Description
This allows adding an `async` attribute to `<script>` tags for Nuxt bundles, enabling them to be fetched in parallel to parsing. [More information](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).

You can enable this in your nuxt config:
```js
export default {
  render: {
    asyncScripts: true
  }
}
```

This may ultimately be implemented in `vue-server-renderer`.

Resolves #8242

## Checklist:
- [x] I have updated the documentation accordingly. (https://github.com/nuxt/nuxtjs.org/pull/989)


Atinux:
We encountered an issue with nuxtjs.org and Nuxt content lately, having “cannot read property dbHash of undefined”.

Do you mind checking this PR with Nuxt Content and SSG?

danielroe:
> We encountered an issue with nuxtjs.org and Nuxt content lately, having “cannot read property dbHash of undefined”.

Will look into it @atinux

Do you have a repro? Did it occur every time you generated nuxtjs.org?

pi0:
I think related issue was accessing `$config` within plugins in full-static mode (since we await on JSONP after app creation and without async it was implicitly being awaited)

BTW even this feature is really nice would suggest supporting behind a flag to avoid unwanted regressions on nuxt2.

danielroe:
Linking https://github.com/nuxt/content/issues/493 and

danielroe:
(Am I right to think that the `dbhash` issue is unrelated, if [this issue](https://github.com/nuxt/content/issues/493)?)

I propose `render.asyncScripts` and will update PR - other ideas welcome...

Atinux:
Sounds good, could you open a PR to nuxtjs.org to add it there: https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-render

atinux (approved):

pi0:
(merge is pending for 2.15)

Timkor:
What about chunks loaded dynamically after page load? Can those also be loaded using the `async` attribute?

pi0 (approved):

PR #8348, opened 2020-11-15
fix(vue-app): warn if promises and functions are in fetch state

### LINKED ISSUE #8338: Functions declared in data are cleared when fetch is present
### Versions

- nuxt: 2.12.2

### Reproduction

[https://codesandbox.io/s/distracted-sky-tdi6b](https://codesandbox.io/s/distracted-sky-tdi6b)

<details open>
<br>

</details>

### Steps to reproduce

Declare a function as variable inside `data` in a page

### What is Expected?
Access to that variable and use it

The variable is empty if I use fetch hook of nuxt. The problem occurs only in ssr mode, if go to the page with the navigator all is working as expected.

danielroe:
## Types of changes

## Description
Resolves #8338

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -26,10 +26,18 @@ export function hasFetch(vm) {
   return vm.$options && typeof vm.$options.fetch === 'function' && !vm.$options.fetch.length
 }
 export function purifyData(data) {
+  if (process.env.NODE_ENV === 'production') {
+    return data
+  }
+
   return Object.entries(data).filter(
-    ([_, value]) =>
-      !(value instanceof Function) &&
-      !(value instanceof Promise)
+    ([key, value]) => {
+      const valid = !(value instanceof Function) && !(value instanceof Promise)
+      if (!valid) {
+        console.warn(`${key} is not able to be stringified. This will break in a production environment.`)
+      }
+      return valid
+    }
     ).reduce((obj, [key, value]) => {
       obj[key] = value
       return obj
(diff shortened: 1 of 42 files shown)


packages/vue-app/template/utils.js
```diff
@@ -25,6 +25,16 @@ export function interopDefault (promise) {
 export function hasFetch(vm) {
   return vm.$options && typeof vm.$options.fetch === 'function' && !vm.$options.fetch.length
 }
+export function purifyData(data) {
```
pi0, line 28: Since this adds a runtime overhead, maybe we can warn users in development mode to not use non POJO in state/data?

danielroe: I think a lot of people don't understand the POJO warning - and indeed the side effects (that the entire component state is stringified, for example!)

Are you thinking of warning in dev mode and not checking in prod?

Another option: use a Proxy for `this` for fetch to detect changed properties so not everything will be stringified?

pi0: > I think a lot of people don't understand the POJO warning - and indeed the side effects (that the entire component data is stringified, for example!)

Indeed fully agree. We need better DX to guide users about this.

> Are you thinking of warning in dev mode and not checking in prod?

Yeah. Like a `process.env` guard similar to other vue packages (vuex for example only checks `stat` to be a function in dev env)  this can avoid adding overhead to production

> Another option: use a Proxy for this for fetch to detect changed properties so not everything will be stringified?

Nuxt 3 PR welcome :D

pi0 (approved):

PR #8389, opened 2020-11-22
feat: support plug-and-play, typescript runtime and native modules

pi0:
## Summary

- Using [nuxt-contrib/jiti](https://github.com/nuxt-contrib/jiti) to for imports for better ESM support
- Support typescript runtime out of the box with jiti
- Use explicit dependencies for plug-and-play compatibility (with exception of `@nuxt/cli` and `@nuxt/config`)
- Resolving dependencies with proper path for plug-and-play compatibility
- Allow overriding dependencies from project with higher priority (like postcss@8)

## Supporting plug-and-play and yarn v2 (berry)

Berry requires every package have explicitly defining their dependencies and also passing require-path for resolving on behalf of another module. After awesome works by @paul-soporan and @arcanis to support berry (#7295)  ( ❤️ )  we finally had to close PR since it was risking breaking changes for nuxt2 users (specially for loaders matching) and was requiring to change way of nuxt2 mono-repo packages are connected with distro dependencies and relying on `node_modules` implicit structure. (nuxt3 rewrite is already considered this)

Since @evanrlong started another initiative with [nuxt-berry](https://github.com/evanrlong/nuxt-berry) ( ❤️ )  ~I got an idea to support a new nuxt distro `nuxt-pnp` fixing mono-repo issues. Technically a nuxt _distro_ is build from source. With only difference that this special distro, squashes all packages in core (400Kbs in total) and automatically syncs dependencies. This means final install size is exactly the same but with single package (similar to nuxt3) which is fully compatible with berry without any compat mode or patches~ (Update: solution is simplified with `__NUXT_PATHS__` for dynamic imports in Config and CLI)

## Changing default require from `esm` to `jiti`

We started to support ESM syntax long time ago using [standard-things/esm](https://github.com/standard-things/esm). This helped a lot adopting syntax for runtime places like `nuxt.config` and `serverMiddleware`. But over the time this package started to introduce [countless issues](https://github.com/nuxt/nuxt.js/search?q=esm&type=issues) and lack of maintenance. 

We had to disable it for typescript-runtime and since regression with jest, we had to also auto detect jest environment and auto-disable it. At that point, I felt we really need a stable solution supporting syntax and replace esm so started developing in-house solution [nuxt-contrib/jiti](https://github.com/nuxt-contrib/jiti) which is a 1MB build of babel to support both esm modules and typescript on the fly with caching and proper traces.

Even though currently it is possible to switch require engine using `createRequire: 'jiti'` from nuxt config, I think it is finally time to switch off using esm package. One plus point with this is also that we can support **typescript runtime out of the box** for config and middleware!

**What about native esm?** Jiti bypasse `.mjs` requires when native support is available. Yet as it is not possible to do hot reload for modules (cache is intentionally not exposed to user level)., changing config or middleware, would require a full server reload when using native modules...

### Noticeable changes:

- Using explicit dependencies (with exception of `@nuxt/cli` which is used for `nuxt-start` and ad-hoc modules 
- Removing unused dependencies
- `lodash` dependency is not bundled anymore to simplify flow
`@nuxt/components`/ `@nuxt/telemetry` which are discovered by `@nuxt/config` which is also used for `nuxt-start`)
- `__NUXT_PATHS__`: Importing resolve mechanism to extend paths used for CLI
- Resolve loader paths (**potential breaking**)
- Build scripts (for nuxt packaging) will fail if there are unused for implicit dependencies
- Removed `@rollup/plugin-node-resolve` to avoid silent inlining of dependencies (we was doing this in past without even notice)

## TO...DO

- [x] More tests for native esm support (@danielroe @rchl need your help)
- [x] More tests for typescript (both runtime and build module) (/@kevinmarrec need your help)
- [ ] Update docs for `createRequire` option and new features (@Atinux @debs-obrien need you help)
- [x] Double check to ensure there are no breaking changes (@clarkdo need your help) (using nuxt-edge)


merceyz:
It would be better if this was released under a tag instead of a different package name as it would allow peer dependencies to still work and when nuxt 3 is released you don't have to change any imports

clarkdo:
I assigned to me as well for reminding me this issue to be working on.

evanrlong:
This is amazing, thanks for adding this!

One addition I would suggest would be to support [Zero-Installs](https://yarnpkg.com/features/zero-installs). I opened a PR for `nuxt-pnp-demo` with an example of how it can be done.

https://github.com/pi0/nuxt-pnp-demo/pull/2

Its mostly straightforward... just version control `.yarn/cache`.  However, there are some native modules required by `nuxt-pnp` that complicate things.  See my PR for a more detailed description.  

Short version... adding the following snippet to a project's (root) `package.json` will make Zero-Installs work once `.yarn/cache` is version controlled.

```json
  "dependenciesMeta": {
    "core-js": {
      "built": false,
      "unplugged": false
    },
    "ejs": {
      "built": false,
      "unplugged": false
    },
    "fsevents": {
      "built": false,
      "unplugged": false
    },
    "nan": {
      "built": false,
      "unplugged": false
    },
    "node-gyp": {
      "built": false,
      "unplugged": false
    },
    "nuxt": {
      "built": false,
      "unplugged": false
    },
    "term-size": {
      "built": false,
      "unplugged": false
    }
  }
```

I don't know if resolving the install of those native dependencies inside `nuxt-pnp` itself would be an option or if that could cause conflicts with other packages installed in a project.  If that can't be done, I'd suggest that instructions could be added to the `nuxt-pnp` readme, telling people how to use Zero-Installs inside their project, along with any possible side-effects.

clarkdo (dismissed):

SirWindfield:
@evanrlong I tried to setup my project the same as your demo PR with zero-installs but I get errors for native extensions:

```
Error: Qualified path resolution failed - none of those files can be found on the disk.

Source path: .yarn/cache/nuxt-pnp-npm-2.14.12-3-3820474557-1d93ba144a.zip/node_modules/nuxt-pnp/dist/nuxt
Not found: .yarn/cache/nuxt-pnp-npm-2.14.12-3-3820474557-1d93ba144a.zip/node_modules/nuxt-pnp/dist/nuxt
Not found: .yarn/cache/nuxt-pnp-npm-2.14.12-3-3820474557-1d93ba144a.zip/node_modules/nuxt-pnp/dist/nuxt.js
Not found: .yarn/cache/nuxt-pnp-npm-2.14.12-3-3820474557-1d93ba144a.zip/node_modules/nuxt-pnp/dist/nuxt.json
Not found: .yarn/cache/nuxt-pnp-npm-2.14.12-3-3820474557-1d93ba144a.zip/node_modules/nuxt-pnp/dist/nuxt.node
```

Did you come across this error? If so, how did you solve it?

pi0:
Hi @SirWindfield. Thanks for giving it a try. If it works with `nuxt-edge` that's perfect. `nuxt-pnp` was a prototype idea to make it working without breaking nuxt 2 but finally it is working out of the box with main distro :) (i will tear down both nuxt-pnp and my old repo)

nekdolan:
Hi @pi0 which nuxt version with which yarn version does it work out of the box specifically? I tried with `yarn set version berry` and most versions above nuxt 2.15 .

PR #8398, opened 2020-11-25
fix(cli): respect `router.trailingSlash` when serving static files

### LINKED ISSUE #8350: Router trailing slash not behaving as expected.
### Versions

- nuxt: v2.14.7
- node: v14.15.0

### Reproduction

Code:
https://codesandbox.io/s/nuxt-glitches-eb37j

Preview:
https://csb-eb37j.netlify.app/

### Steps to reproduce

1. Go to the preview link.
2. Click the mountains button (this takes you to a new page via router push).
3. Notice the URL is without a slash at the end (correct).
4. Hard refresh the page and it's forward to a URL with a slash (incorrect).

### What is Expected?

There should never be a slash at the end of the URL.

In the `nuxt.config.js` file we have specified:

```
router: {
  trailingSlash: false
}
```

A 301 redirect is occurring and placing a slash at the end.

This is not unique to Netlify as it also happens locally when running `npm run generate` followed by `npm run start`. **Note** that this does **NOT** occur when running `npm run dev`. It seems to only be in a production build.

The docs state that this should not be the case if setting the value to `false` and not `undefined`.
https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router#trailingslash

Tested in **full static** mode using `target: 'static'`

danielroe:
## Types of changes

## Description
`serve-static` defaults to *redirect* directories to `trailingSlash` ([see docs](https://github.com/expressjs/serve-static#redirect)). This PR respects the user's choice (defaulting not to redirect).

This does change behaviour because it will no longer redirect by default (i.e. if `trailingSlash` is undefined), but this will almost certainly only be true on the local machine as it's unlikely users will serve their static sites in production with `nuxt start`

closes #8350

## Checklist:

### DIFF
--- a/packages/cli/src/utils/serve.js
+++ b/packages/cli/src/utils/serve.js
@@ -30,7 +30,8 @@ export async function serve (cmd) {
   app.use(
     options.router.base,
     serveStatic(options.generate.dir, {
-      extensions: ['html']
+      extensions: ['html'],
+      redirect: !!options.router.trailingSlash
     })
   )
   if (options.generate.fallback) {
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -92,7 +92,10 @@ export default class Server {
     // For serving static/ files to /
     const staticMiddleware = serveStatic(
       path.resolve(this.options.srcDir, this.options.dir.static),
-      this.options.render.static
+      {
+        redirect: !!this.options.router.trailingSlash,
+        ...this.options.render.static
+      }
     )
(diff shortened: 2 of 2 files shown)


pi0 (approved):

TheAlexLichter (approved):

atinux (approved):

B-starboy (commented):
I have had a lot of trouble with this router/stash program. Personally I’m glad you guys are working on this. Thanks 🙏

PR #8408, opened 2020-11-29
feat(postcss): upgrade to postcss v8

### LINKED ISSUE #8087: PostCSS 8

PostCSS 8 came with a new plugin API. Right now Nuxt.js users can use new PostCSS plugins because Nuxt.js uses PostCSS 7.

Update PostCSS from `^7.0.0` to `^8.0.0`

### Additional context

[PostCSS 8 changelog](https://github.com/postcss/postcss/releases/tag/8.0.0)

The only breaking changes important for Nuxt.js is dropping Node.js 6 and 8 support.

clarkdo:
## Types of changes

## Description

**NOTE: Do not merge until https://github.com/csstools/postcss-preset-env/issues/191 is closed.**

- Update postcss to v8
- Update postcss-loader to latest
- Update css-loader to v5

**NOTE:** `postcss-loader` has moved `PostCSS` (plugins/syntax/parser/stringifier) options to the `postcssOptions` option, so `build.postcss` in `nuxt.config.js` will be used to set `postcssOptions` which is `plugins/preset...` 

## Checklist:

- [ ] postcss-preset-env is not fully supported postcss v8 https://github.com/csstools/postcss-preset-env/issues/191
- [ ] update [nuxt doc](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-build#postcss) to refer https://github.com/webpack-contrib/postcss-loader#postcssoptions
- [ ] Update cssnano (https://github.com/nuxt/nuxt.js/pull/9140)
- [ ] Update optimize-css-assets-webpack-plugin (https://github.com/nuxt/nuxt.js/pull/9300)


ai:
It is also important to move `postcss` to `peerDependencies` to avoid Nuxt.js PostCSS 9 migration delay in the future.

pi0:
Hi @ai. I'm afraid this is not possible for `nuxt@2` at least since it requires end-users to specify package in dependencies (even tough latest npm installs them by default). And in case of breaking changes with v9 and hosting by 3rd party packages, users need to manually install either v8 or v9.

Nuxt upgrade for postcss was mostly delayed because of dropping node@8 and `postcss-preset-env` support.

Alternatively, what do you think resolving postcss from rootDir with higher priority or allow an option to specify resolved version?

ai:
> Nuxt upgrade for postcss was mostly delayed because of dropping node@8 and postcss-preset-env support.

Next PostCSS will have Node.js version changes as well 😅 (Node.js 10 will be deprecated next year).

> Alternatively, what do you think resolving postcss from rootDir with higher priority or allow an option to specify resolved version?

Am I right that users can add their own plugins to Nuxt.js (for instance, Tailwind CSS)? In this case, they will have a warning about missing `postcss` peer dependency.

But I like the idea of having built-in default `postcss` in Nuxt.js with an option to replace `postcss` in the project. It looks like a very good DX.

pi0:
> Next PostCSS will have Node.js version changes as well 😅 (Node.js 10 will be deprecated next year).

Thanks for heads up. Totally makes sense not supporting Node.js 10.From next minor release we start warning to use supported versions (12, 14+) and eventually drop 10 as well (10 is only supported by `engine` field as long as core-deps allow)

> Am I right that users can add their own plugins to Nuxt.js (for instance, Tailwind CSS)? In this case, they will have a warning about missing postcss peer dependency.

Indeed. With having an option, modules like tailwind can enforce a specific version of postcss.

andorfermichael:
I hope this will come soon as it will enable usage of latest versions of many other packages.

andorfermichael:
@pi0 when do you plan to move on with this? :)

clarkdo:
I’ll make a pr covering it

darakanoit:
ping

clarkdo:
This pr will be done in https://github.com/nuxt/nuxt.js/pull/9671

PR #8421, opened 2020-12-01
fix: only encode non dynamic path params

### LINKED ISSUE #8420: Custom routes not working after upgrading to v2.14.8
### Versions
nuxt: v2.14.8
node: v12.18.3

**Routes:**
image

**Error:**
image

pi0:
## Types of changes

## Description

Only encode non dynamic route paths. Resolves #8420 

Context: #8325 (with this PR, custom router is responsible to encode routes)

## Checklist:

### DIFF
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -201,9 +201,10 @@ export const createRoutes = function createRoutes ({
       } else if (key === 'index' && i + 1 === keys.length) {
         route.path += i > 0 ? '' : '/'
       } else {
-        route.path += '/' + getRoutePathExtension(key)
+        const isDynamic = key.startsWith('_') && key.length > 1
+        route.path += '/' + getRoutePathExtension(isDynamic ? key : encodeURIComponent(decodeURIComponent(key)))
 
-        if (key.startsWith('_') && key.length > 1) {
+        if (isDynamic) {
           route.path += '?'
         }
       }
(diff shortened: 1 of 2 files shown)


clarkdo (approved):

atinux (approved):

ydnar:
We just encountered a similar bug in our @domainr website frontend. We have tests that caught a URI-encoded param in a route. Previously, a `,` was decoded correctly. This is in the current latest version of Nuxt and vue-router.

rchl:
What about routes like this one: https://github.com/vuejs/vue-router/blob/2c757a1cdc3aa8d55c3e3e22788a89f2351e58a0/examples/route-matching/app.js#L24
?

From a quick look at the code, it seems like it encodes all routes that are not non-dynamic but routes can be non-dynamic and contain regexp which will break when encoded.

pi0:
@rchl Kindly please use a reproduction/issue. Last change only tries to encode auto scanned directories which do not support this convention.

rchl:
Sounds good. Was just pointing out a potential issue without looking much at the code as I don't have time to dig into it now.

ydnar:
> Released as hotfix [v2.14.9](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.9). @ydnar if still had issue, please open an issue :)

Still broken. I’ll file an issue!

PR #8423, opened 2020-12-01
fix: `trailingSlash` and child routes not working

### LINKED ISSUE #8422: trailingSlash=true and child routes generates a VueRouter warning
### Versions

- nuxt: 2.14.8
- node: v12.19.0

### Reproduction

```js
  router: {
    trailingSlash: true
  },
```

Directory structure:
Screenshot 2020-07-31 at 13 40 45

Repo: https://codesandbox.io/s/nuxt-trailingslash-true-and-child-routes-zp5tt (best to download and try locally).

### Steps to reproduce

Visit the root route.

### What is Expected?

No warning in the console

With a directory structure like above and `router.trailingSlash` set to `true` the generated routes in `.nuxt/router.js` look like this:

```js
{
    path: "/posts/",
    component: _17b9cb62,
    name: "posts",
    children: [{
      path: "",
      component: _e4367f36,
      name: "posts"
    }, {
      path: ":slug/",
      component: _e7a3c1c6,
      name: "posts-slug"
    }]
  }, {
    path: "/",
    component: _6f29e0e1,
    name: "index"
  }
```

and visiting the root route triggers a VueRouter warning:

> WARN  [vue-router] Named Route 'posts' has a default child route. When navigating to this named route (:to="{name: 'posts'"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead.
> WARN  [vue-router] Duplicate named routes definition: { name: "posts", path: "/posts/" }

I believe this is due to `name` being duplicated on two routes.

pi0:
## Types of changes

## Description
Resolves #8422 also adding dev test for 3 possible values of trailing slash

`[vue-router] Named Route 'posts' has a default child route. When navigating to this named route (:to="{name: 'posts'"), the default child route will not be rendered. Remove the name from this route and use the name of the default child route for named links instead.`

## Checklist:


rchl:
Maybe you could also add trailingSlash=false version of that testsuite? I guess we don't have any existing testing for it?

pi0:
@rchl I believe there is a test suit for children without trailing slash:
- https://github.com/nuxt/nuxt.js/tree/dev/test/fixtures/children
- https://github.com/nuxt/nuxt.js/blob/dev/test/dev/children.test.js

rchl:
`trailingSlash=undefined` and `trailingSlash=false` are not handled the same though.

packages/utils/src/route.js
```diff
@@ -80,15 +80,14 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-',
     }
     route.name = route.name.replace(regExpIndex, '')
     if (route.children) {
-      const indexRoutePath = trailingSlash === false ? '/' : ''
-      const defaultChildRoute = route.children.find(child => child.path === indexRoutePath)
+      const defaultChildRoute = route.children.find(child => child.path === '/' || child.path === '')
       const routeName = route.name
       if (defaultChildRoute) {
         if (trailingSlash === false) {
```
farnabaz, line 86: I think we can remove this condition now. We are going to remove `route.name` therefore we can use its name for the default child

farnabaz (dismissed):

pi0:
@rchl There is also a unit/snapshot test for routes utility: https://github.com/nuxt/nuxt.js/blob/dev/packages/utils/test/route.test.js

packages/utils/test/route.test.js
```diff
@@ -210,7 +210,6 @@ describe('util: route', () => {
       'pages/subpage/_param.vue',
       'pages/snake_case_route.vue',
       'pages/another_route/_id.vue',
-      'pages/another_route/_id.vue',
```
pi0, line 213: @clarkdo I think duplicate line was typo right?

rchl:
> @rchl There is also a unit/snapshot test for routes utility: [https://github.com/nuxt/nuxt.js/blob/dev/packages/utils/test/route.test.js](https://github.com/nuxt/nuxt.js/blob/dev/packages/utils/test/route.test.js?rgh-link-date=2020-12-01T16%3A59%3A41Z)

I don't see it testing the `trailingSlash: false` case which, as I've said, is not entirely identical to it being undefined.

pi0:
Actually [it is](https://github.com/nuxt/nuxt.js/blob/34359811f67a1ad62d4f705475652e26aa44a2d9/packages/utils/test/route.test.js#L237) :D Anyway agreed that snapshot couldn't cover this issue so added `trailing-slash-false` dev tests :)

test/dev/trailing-slash.test.js
```diff
@@ -0,0 +1,88 @@
...
+  })
+
+  test('/', async () => {
+    const { html } = await nuxt.server.renderRoute('/')
+    expect(html).toContain('[pages/index]')
+  })
+
+  test('/posts', async () => {
+    const { html } = await nuxt.server.renderRoute('/posts')
+    expect(html).toContain('[pages/posts]')
```
pi0, line 64: This is kinda wired behavior tough...

danielroe:
I think we should pair this PR with another one for the docs to explain what the behaviour should be as I think there are a variety of different expectations in the related issues...

Happy to pick that up.

test/fixtures/trailing-slash/trailish-slash.test.js
```diff
@@ -0,0 +1,5 @@
+import { buildFixture } from '../../utils/build'
```
danielroe, line 1: should be trailing-slash.test.js (?)

pi0: Nice fine! trailish is seems likely name for current status 😆

PR #8430, opened 2020-12-02
fix: normalize routes and decode resolved query

### LINKED ISSUE #8429: Query string parameters are not URI-decoded in route.query
### Versions

- nuxt: v2.14.8 and v2.14.9
- node: v15.3.0

### Reproduction

Since updating to nuxt v2.14.8, query string parameters with encoded characters are passed un-decoded in `route.query` in `context` to `asyncData`. The v2.14.9 hotfix did not fix this issue.

<details>

```js

  async asyncData ({ app, route, error }) {
    let data = {
      search: {
        query: route.query.q,
        location: route.query.gl,
        registrar: route.query.r,
        terms: route.query.t,
        defaults: route.query.d,
        results: []
      },
      metadata: {
        domain: route.params.domain
      }
    }

    // BUG: data.search.terms == 'coffee%2Cfood' instead of 'coffee,food'

    // ...

    return data
  }
```
</details>

### Steps to reproduce

- Visit `/?t=coffee,food`
- In `asyncData`, extract the `t` query string parameter

### What is Expected?

`route.query.t == 'coffee,food'`

`route.query.t == 'coffee%2Cfood'`

pi0:
## Types of changes

## Description

- History: vuejs/vue-router#3350 > #8325 > #8421 
- Resolves #8429, #8442

Since `vue-router@2.14.9` makes it mandatory to use `encodeURI` when using `RouterLink`, `NuxtLink` and  `router.push`, we introduced a fix by automatically encoding URIs. But seems it makes regression when a query param (and probably path param) is encoded and not being properly decoded.

Changes:
- Use `normalizeURL` from [`@nuxt/ufo`](https://github.com/nuxt-contrib/ufo)
- Decode `route.resolved.query` with `decodeURIComponent`
- Encode redirect path with `normalizeURL` since http headers should be encoded

--- 

Foot note: Case of #8429 only happens when manually using `encodeURIComponent` on query parameters instead of `encodeURI`  and at least chrome auto fixes this. (still this is a regression that we should fixed)

record

## Checklist:


ydnar:
Note: we see this issue when a browser visits the encoded URL (via typing, for instance), without any router push.

clarkdo (approved):

clarkdo (approved):

packages/utils/src/route.js
```diff
@@ -280,3 +279,11 @@ export const promisifyRoute = function promisifyRoute (fn, ...args) {
   }
   return promise
 }
+
+export function safeEncodeComponent (str) {
+  return /%[0-9a-fA-F]{2}/.test(str) ? str : encodeURI(str)
+}
+
+export function safeEncode (str) {
+  return str.split('/').map(safeEncodeComponent).join('/')
```
clarkdo, line 288: What if url includes %.. as path or parameter ?

Like/rate/%30?max=%90

packages/utils/src/route.js
```diff
@@ -280,3 +279,11 @@ export const promisifyRoute = function promisifyRoute (fn, ...args) {
   }
   return promise
 }
+
+export function safeEncodeComponent (str) {
+  return /%[0-9a-fA-F]{2}/.test(str) ? str : encodeURI(str)
```
clarkdo, line 284: What if url includes %.. as path or parameter ?

Like: /rate/%30?max=%90

pi0: `/%[0-9a-fA-F]{2}/.test('/rate/%30?max=%90')` > true

In this case we avoid encoding

clarkdo: So if url also includes special characters, we’ll skip encoding, is that expected?

Like: /rate/%30?max=%90&name= тест

pi0: Last commit: `/rate/%30?max=%90&name= тест` => `/rate/%30?max=%90&name=%20%D1%82%D0%B5%D1%81%D1%82`

pi0: Last update: `/rate/%30?max=%90&name=%20%D1%82%D0%B5%D1%81%D1%82`

packages/server/src/server.js
```diff
@@ -162,12 +162,16 @@ export default class Server {
...
-    if (this.options.dev && this.nuxt.options.router.base !== '/') {
+    const routerBase = this.nuxt.options.router.base
+    if (this.options.dev && routerBase !== '/') {
       this.useMiddleware({
         prefix: false,
-        handler: (req, res) => {
-          const to = urlJoin(this.nuxt.options.router.base, req.url)
-          consola.info(`[Development] Redirecting from \`${decodeURI(req.url)}\` to \`${decodeURI(to)}\` (router.base specified).`)
+        handler: (req, res, next) => {
+          if (decodeURI(req.url).startsWith(decodeURI(routerBase))) {
```
danielroe, line 170: shouldn't we assume that user-provided data like `router.base` is already decoded?

pi0: routerBase is intentionally encoded and normalized in `options.js`

packages/utils/src/route.js
danielroe: Is there any merit in using something like this as our util - not reinventing url parsing?
```js
const url = '/rate/%30?max=%90&name= тест'
const { pathname, search } = new URL('http://localhost${url}')
return `${pathname}${search}`
// "/rate/%30?max=%90&name=%20%D1%82%D0%B5%D1%81%D1%82"
```

pi0: I would like to at least use a URL aware solution instead of splitting (and finally we should move url utils to a package). But depending on environment URL, might be tricky and inconsistent. At least here is IE 11 behavior:

image

danielroe: In fact we also get benefit of leading slash normalisation if we do:
```js
function safeEncode (str) {
  const { pathname, search } = new URL(str, 'http://localhost')
  return `${pathname}${search}`
}
```

About IE - it is one of the few exceptions - could we polyfill in this case?

pi0: Indeed we can but it costs bundle size for modern browsers and actually it is not enough to make url utils (for joining or ensure there is no leading slash for example) -- not specific to this PR's case

danielroe: In fact IE11 works fine for me - just have to wrap in brackets.
Screenshot from 2020-12-04 16-23-20

pi0: Fair enough :) Switching to URL

pi0: There is just one downside that it is implicit it is full url or URI :))

pi0: Thanks for bringing it up. Working on util before merge ;)

PR #8460, opened 2020-12-09
fix(cli): share nuxt instance for building and generating full static

### LINKED ISSUE #7855: nuxt generate memory leak
### Versions

- nuxt: 2.14.1
- node: v14.7.0

### Reproduction

I'm using `nuxt generate` with the memory benchmarking from https://github.com/nuxt/nuxt.js/issues/7412#issuecomment-643195374.

I've reduced the generate.concurrency down to 1 and I can see in the logs that after "Generating pages with full static mode" it start's going through my pages and the heap memory just keeps on growing.

After about 100 pages generated it runs out of heap memory `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`.

I'm using 3GB of heap memory with `node --max_old_space_size=3072 ./node_modules/.bin/nuxt generate`, sure I could increase this but with hundreds of pages it's not scalable.

This wasn't happening in the older version of nuxt 2.12.2

### What is Expected?

I expect that generating pages would have roughly flat memory usage and not keep on growing with each generated page.

Memory usage keeps growing with each generated page.

danielroe:
## Types of changes

## Description
closes #7855

## Checklist:


pi0:
As far as remember [nuxt/static](http://github.com/nuxt/static) had issues with modules when reusing same instance. Could you track down memory leak root cause?

danielroe:
@pi0 As far as I can tell the issue is caused by sharing the Vue instance (via `require.cache`) between build & generate. So this workaround also solves it:

```js
  hooks: {
    'build:done'() {
      const modulesToClear = [
        'vue',
        'vue/dist/vue.runtime.common.prod'
      ]
      modulesToClear.forEach(entry => {
        delete require.cache[require.resolve(entry)]  
      })
    }
  }
```

I wonder if it would be a reasonable thing simply to flush the entire cache between build & generate steps - ie. not just vue? Happy to amend the above PR to do so.

lastthyme (approved):

pi0:
We were missing `nuxt.close()` when skipping build step, that is probably the root cause. If any module like nuxt-vite or core internal is leaking some refs or opening resources, they should ensure to listen on `'close'` hook and cleanup. Using shared instance: 1) hides root causes 2) introduces other issues when modules are initialized for the wrong env (build vs generate) 3) is not solving leak issue for use cases that we actually need multiple instances like in parallel tests.

Issue #8193, opened 2020-10-12
Static generation with initial negative v-if or condition in components renders out the component


davydnorris:
### Versions

- nuxt: v2.14.6
- node: v12.18.3

### Reproduction
This component illustrates the issue - it's using Vuetify for formatting but basically just presents a message at the bottom of the screen regarding cookies

myGDPRNote.vue
```
<template>
  <v-snackbar
    :value="showAccepted"
    :timeout="-1"
    bottom
  >
    This website uses cookies to ensure you get the best experience.
    <a
      class="text-decoration-underline"
      @click="dialog=true"
    >
      Learn more
    </a>
    <v-dialog
      v-model="dialog"
      width="600"
    >
      <v-card>
        <v-card-actions>
          <v-spacer />
          <v-btn
            @click="dialog = false"
          >
            OK
          </v-btn>
        </v-card-actions>
      </v-card>
    </v-dialog>
    <template v-slot:action="{ attrs }">
      <v-btn
        v-bind="attrs"
        @click="acceptCookies"
      >
        Got it!
      </v-btn>
    </template>
  </v-snackbar>
</template>

<script>
export default {
  data: () => ({
    dialog: false,
    showAccepted: false   // this is the crucial setting
  }),
  created () {
    if (process.browser) {
      this.showAccepted = !localStorage.acceptCookies;
    }
  },
  methods: {
    acceptCookies () {
      this.showAccepted = false;
      if (process.browser) {
        localStorage.acceptCookies = true;
      }
    }
  }
};
</script>
```

This is the resulting HTML in the browser:
```
<div class="v-snack v-snack--active v-snack--bottom v-snack--has-background" style="padding-bottom:0px;padding-top:108px;">
  <div class="v-snack__wrapper v-sheet theme--dark" style="display:none;">
    <div role="status" aria-live="polite" class="v-snack__content">
      This website uses cookies to ensure you get the best experience.
      <a class="text-decoration-underline">
        Learn more
      </a>
      <div role="dialog" class="v-dialog__container">
        <!---->
      </div>
    </div>
    <div class="v-snack__action ">
      <button type="button" class="v-btn v-btn--contained theme--dark v-size--default v-snack__btn">
        <span class="v-btn__content">
          Got it!
        </span>
      </button>
    </div>
  </div>
</div>
```

### Steps to reproduce
When this component is included in a project, it runs fine in dev mode and universal mode, but when a full static site is generated the note never appears

### What is Expected?
[...]

Issue #8150, opened 2020-10-05
async fetch() breaks v-for in child component


samlfair:
### Versions

- nuxt: ^2.14.6
- node: 14.7.0

### Reproduction

`~/layouts/default.vue` contains an `async fetch()` to fetch content from Prismic.

`~/pages/portfolio/index.vue` contains a component `~/components/slices/Featured.vue` (a post list), which contains a `v-for` loop that cycles through a list of posts and renders them with another component, `~/components/FeaturedPost.vue`. `~/components/FeaturedPost.vue` makes an API call to get the title and image for the post.

When the `async fetch()` method is included in the layout, the API data for FeaturedPost[n] gets applied to FeaturedPost[n+1]. FeaturedPost[0] is empty.

Does not happen when `/portfolio` is the first page loaded — only when it is navigated to subsequently.

<details open>
<summary>Additional Details</summary>
<br>

</details>

### Steps to reproduce

To reproduce, download this branch:

https://github.com/samlfair/claire-v4/tree/async-vfor

Run `npm run generate` and then `npm run start`. Start at any page, and then navigate to `/portfolio`. You should see the missing post. To fix the error, open `~/layouts/default.vue`, and comment out the `async fetch()` method and uncomment the `config` data, lines 38-70 (to prevent breaking errors).

The post list is rendered with this component: https://github.com/samlfair/claire-v4/blob/async-vfor/components/slices/Featured.vue

Each item in the list is rendered with this component: https://github.com/samlfair/claire-v4/blob/async-vfor/components/FeaturedPost.vue

### What is Expected?

Posts should render in order.

### What is actually happening?

First post is blank, then posts are rendered in order.

lihbr:
I had a look with Sam this afternoon at this issue and inspected bundle output `payload.js` for the portfolio page.

I think the fetch resolver for static payload is struggling at matching payload with components here and not sure it's either a legit bug or a stupid error somewhere we failed at figuring out 🤔

Issue #8107, opened 2020-09-22
Nuxt Generate - Page not showing any content client side, works fine server side.


drewbaker:
### Versions

- nuxt: 2.14.3
- node: 10.13.0

### Reproduction

This repo: https://github.com/funkhaus/fuxt

Demo is here: https://fuxt-funkhaus.netlify.app

### Steps to reproduce

1. Rename `.example.env` to `.env`
1. Use `nuxt generate` on above repo and deploy to Netlify (or `nuxt start` locally)
2. Load any page on the site (https://fuxt-funkhaus.netlify.app), content loads off server fine. 
3. Browse to any other page via `nuxt-link` and page new will be empty of any dynamic content.

### What is Expected?

Client side browsing should load content from payload onto the page.

### What is actually happening?

So it seems server side is working fine. But client side is not. If you look at the network inspector in your browser as you move around the site, you'll see that `payload.js` is loading and looks to contain the correct code, but it never makes it onto the page for some reason.

The site only contains a few `fetch()` methods here:
https://github.com/funkhaus/fuxt/blob/master/pages/_page.vue
https://github.com/funkhaus/fuxt/blob/master/components/WpMenu.vue
https://github.com/funkhaus/fuxt/blob/master/components/WpSeo.vue

And a request in `nuxtServerInit` here: 
https://github.com/funkhaus/fuxt/blob/master/store/index.js#L60

It all seems to be inline with the way you recommend the use of `fetch()`, so I'm not sure what is not working? Perhaps it doesn't like GQL? Or the use of lodash `_get` in the `fetch()`? 

I wish there was more explanation of what `nuxt generate` and `payload` is doing behind the scenes so I could troubleshoot this myself.

Atinux:
Hi @drewbaker 

It is hard to test your reproduction without any token to test actually.

Also, it seems you depend on `context.req` where actually it is `undefined` with `target: 'static'`

drewbaker:
@Atinux sorry, what token do you need? They are all in `.env.example`. Just rename that `.env`.

drewbaker:
And I don't think we use `context.req` for anythign critical, it's just used to make URL's into relative paths from our CMS. So it shouldn't break if it comes back empty, it's in a lodash _get.

drewbaker:
@Atinux just pushed an update that removes use of `req` for you, [it's the only way to be sure](https://www.youtube.com/watch?v=eGOtZDk6wRc). 

[...]

Issue #7799, opened 2020-07-28
[2.14.0] Static Payload not applying to <nuxt-child> children components in some cases


FreekVR:
May be related to #7791

Nuxt payloads do not always apply correct in generated apps with `target: static`, if `nuxt-child` is used, and both a child component of the higher-level component, and one of the children of `nuxt-child` implement `fetch()` (see diagram below)
### Versions

- nuxt: 2.14.0
- node: 10.23.0

### Reproduction
Source: https://github.com/FreekVR/nuxt-static-repro/tree/nested-fetch-issues

Demo: https://wonderful-ardinghelli-e41a6b.netlify.app/magazine/category/bar/

This will have correctly generated data on first (hard) load, but clicking the link-links present on the page will show that in SPA mode, the payload data is not applying afterwards.

This seems to be caused by the fetch() in [this component](https://github.com/FreekVR/nuxt-static-repro/blob/nested-fetch-issues/components/sidebar.vue), which is included in the nuxt-child parent component.

### Steps to reproduce
Have an application implementing a structure like below (see repro source for practical implementation)

![image](https://user-images.githubusercontent.com/417416/88689764-2edfa980-d0fb-11ea-8e58-51949be04a53.png)

### What is Expected?
Data is correctly applied from the payload when navigating through SPA links

### What is actually happening?
Data is not applied / lost

f3ltron:
Same error here

honzabilek4:
I can confirm the issue, same behaviour as stated in description. 
The page is generated correctly, even the payload data exist, but when navigating using `<nuxt-link>` the nested fetch hook is not being invoked.

elisiondan:
Until this issue is resolved, we've opted for calling `fetch` in `beforeMounted` hook which seems to work properly with the static target.
```
  async fetch () {
    // fetching data
  },
  async beforeMount () {
    await this.$fetch();
  }
```

FreekVR:
I can confirm that this issue still persists with the latest release, unfortunately.

agursoy:
I also faced this bug. Is there any update?

FreekVR:
Issue is still occurring with the latest versions, unfortunately.

JustinRvt:
I'm so much following this as I'm experiencing the same issues.

[...]

PR #8466, opened 2020-12-10
feat(vue-app): support custom `fetchKey` for full static generation

### LINKED ISSUE #7563: Full-static payload gets mixed up using nested fetch() hooks inside dynamically loaded components
### Version

### Reproduction link

[https://codesandbox.io/s/cocky-glitter-t6w26?file=/components/Builder.vue:445-454](https://codesandbox.io/s/cocky-glitter-t6w26?file=/components/Builder.vue:445-454)

### Steps to reproduce

- Create a page that uses its own fetch() hook
- Make that page use a wrapper/builder component which in turn uses dynamic component imports
- Make the components imported within the builder use their own fetch() hooks
- The data as displayed on the page (https://t6w26.sse.codesandbox.io/) will be offset by one (note this only occurs within the payload! The data on the HTML/static version of the page is correct)

### What is expected ?

Data is correctly mapped to each individual component

Data per component is offset by one - the first component incorrectly being empty and each component after that is using the incorrect payload

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10806">#c10806</a>)</em></sub></div>

danielroe:
### Previous behaviour

Previously, fetch payloads were assigned to the components that needed them through a sequential method. When a page changed, the payload array was loaded and a global `fetchKey` was initialised to `0`. As components are created and need data, this key increases and the relevant array item is injected into each component.

This posed a problem when components persist between pages, such as when there is a parent/child page, or if a user would like to use `fetch` in a global layout. In this case, the persistent component never rerenders, and so the component that would normally be keyed `1` was assigned the `0`-indexed payload, and so on.

### New behaviour

We are improving hydration by:

1. **Keying fetch payloads with a non-numeric fetchKey.** This will default to the scope ID (if the component uses scoped CSS), the component name, if there is one, or an auto-incrementing number (as at present).

2. **Allow the user to manually take control of the fetch hydration process, by specifying their own `fetchKey`**
   ```js
   export default {
     async fetch () {
       this.title = await fetch('/api/title').then(r => r.json())
     },
     // multiple components can return the same `fetchKey` and Nuxt will track them both separately 
     fetchKey: 'site-sidebar',
     // alternatively, for more control, a function can be passed with access to the component instance
     // It will be called in `created` and obviously must not depend on fetched data
     fetchKey (getCounter) {
       // getCounter is a method that can be called to get the next number in a sequence
       // as part of generating a unique fetchKey. 
       return this.someOtherData + getCounter('sidebar')
     }
   } 
   ```

### Additional context

Will close #7563, #7791, #7799, #8107, #8150, #8193.

#### Types of change

#### Checklist:
- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/nuxtjs.org/pull/1087)
  - [x] generate tests


packages/types/app/index.d.ts
```diff
@@ -65,6 +65,7 @@ export interface Context {
       layout: string
       data: Array<Record<string, any>>
       fetch: Array<Record<string, any>>
+      fetchKey: number
```
pi0, line 68: We might also allow strings for namespacing fetch key (like `shopify:foo`)

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -64,7 +64,14 @@ function createdFullStatic() {
     return
   }
   this._hydrated = true
-  this._fetchKey = this.<%= globals.nuxt %>._payloadFetchIndex++
+
+  if (typeof this.$options.fetchKey === 'function') {
+    this._fetchKey = this.$options.fetchKey.call(this)
```
pi0, line 69: We might still increase and pass `_payloadFetchIndex` as second argument to distinguish multiple instances of one component. (or `getKey` as second arg)

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -19,14 +19,15 @@ async function serverPrefetch() {
...
-  this._fetchKey = this.$ssrContext.nuxt.fetch.length
+  this._fetchKey = this._fetchKey || this.$ssrContext.nuxt.fetchKey++
 
   // Add data-fetch-key on parent element of Component
   const attrs = this.$vnode.data.attrs = this.$vnode.data.attrs || {}
   attrs['data-fetch-key'] = this._fetchKey
 
   // Add to ssrContext for window.__NUXT__.fetch
-  this.$ssrContext.nuxt.fetch.push(this.$fetchState.error ? { _error: this.$fetchState.error } : purifyData(this._data))
+  this.$ssrContext.nuxt.fetch[this._fetchKey] =
```
pi0, line 29: Shouldn't we guard against overriding same key? (a warn at least)

Atinux:
I do think we will need tests to ensure the correct behaviour for this and avoid any regression in the future, so far great work @danielroe 👏

packages/vue-app/template/mixins/fetch.client.js
danielroe: If we want to support defaulting to the component name as a fetch key, as in linked issue, we will need a nested global fetchKey object to keep track of individual indices of each key. That is, the fetch state might look like:
```js
{
  '0': { a: 'mystate' },
  'MyComponent': { a: 'mystate' },
  'MyComponent': [{ a: 'mystate' }, { a: 'mystate2' }],
  '1': { a: 'mystate' },
}
```

If we don't keep track of these indices separately, the same issue that triggered the original bug will reappear because the global numeric fetch key will get out of sync.

packages/vue-app/template/mixins/fetch.server.js
```diff
@@ -26,6 +26,9 @@ async function serverPrefetch() {
   attrs['data-fetch-key'] = this._fetchKey
 
   // Add to ssrContext for window.__NUXT__.fetch
+  if (process.env.NODE_ENV === 'development' && this.$ssrContext.nuxt.fetch[this._fetchKey] !== undefined) {
```
pi0, line 29: We have options.debug (template option) to allow debugging in production/staging

packages/vue-app/template/mixins/fetch.client.js
pi0: That's true. We might use a counter table for `getKey` implementation using component id as key..

danielroe:
**Updates**:

`. Signature is now `fetchKey (keys: Object<string, any>): string | number` (key is a global object for keeping track of fetch index state).

2. If a `fetchKey` function is passed, it is expected to use this object to deduplicate its fetchKey (e.g. to set a unique key and increment it as appropriate). If a `fetchKey` *string* is passed instead, an increment unique to this string will be managed by Nuxt, and added on automatically.

3. Further, if not passed anything, `fetchKey` now defaults to the component name (though is this a potential problem when the component name might be a long filename?).

packages/vue-app/template/mixins/fetch.client.js
danielroe: What would `getKey` API look like? Would it just return a valid incremented number unique to the string passed in?
```js
getKey (keyToLookup: string): number
```

pi0: Ideally `getKey (id: string = ThisConstructor.id): number` but by default indeed using internal table to get a name spaced incrementor.

packages/vue-app/template/utils.js
```diff
@@ -10,6 +10,15 @@ if (process.client) {
   }
 }
 
+export function createGetCounter (counterObject, defaultKey = '') {
+  return function getCounter (id = defaultKey) {
+    if (counterObject[id] === undefined) {
```
pi0, line 15: Since we have this check, do we still need default `'': 0`?

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -64,7 +64,17 @@ function createdFullStatic() {
...
-  this._fetchKey = this.<%= globals.nuxt %>._payloadFetchIndex++
+
+  const defaultKey = this.$options._scopeId || this.$options.name || ''
+  const getCounter = createGetCounter(this.<%= globals.nuxt %>._fetchCounters, defaultKey)
+
+  if (typeof this.$options.fetchKey === 'function') {
+    this._fetchKey = this.$options.fetchKey.call(this, getCounter)
+  } else {
+    const key = 'string' === typeof this.$options.fetchKey ? this.$options.fetchKey : defaultKey
+    this._fetchKey = key + getCounter(key)
```
pi0, line 75: - Do we need to add counter as suffix when one provided manual one?
- We can use `|| getCounter(key)` for `defaultKey`

packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -38,7 +38,7 @@ function created() {
 
   // Hydrate component
   this._hydrated = true
-  this._fetchKey = +this.$vnode.elm.dataset.fetchKey
+  this._fetchKey = this.$vnode.elm.dataset.fetchKey
```
pi0, line 41: What happens in case of hydration mismatch? (key not available). What if throwing a clear error about this?

danielroe: I think we are guaranteed for key to exist: https://github.com/nuxt/nuxt.js/blob/c4bc0c9cd83e483954d7c218c19a236c40c4718a/packages%2Fvue-app%2Ftemplate%2Fmixins%2Ffetch.client.js#L4

packages/vue-app/template/utils.js
danielroe: Good point :+1:

pi0 (approved):

drewbaker:
PLEASE PLEASE PLEASE merge this! Solves so many problems for us at https://github.com/funkhaus

atinux (approved):
Amazing work done 🔥

pi0:
Merged. Should be available soon on `nuxt-edge` and `2.15.0` release.

drewbaker:
@pi0 Thanks for doing this! Is there a guide or tips on how to use `nuxt-edge`? Because I get get all sorts of webpack build errors using it (for the first time just now).

pi0:
@drewbaker Simply remove `nuxt` from package.json and use `yarn add @nuxt/core-edge@2.14.13-26803446.5318172`. It is possible BTW that edge release is broken with some dependencies. Would be nice if can make an issue with errors and current config.

drewbaker:
@pi0 Thanks. Just opened this issue about the build error I'm seeing. Doubt it's related to this PR, but figured I'd mention it: https://github.com/nuxt/nuxt.js/issues/8508

drewbaker:
This no longer works in `2.15.0-26876193.fffb1a19`.

katerlouis:
OMG this solved my issue (and probably many others more where I had to go for workaround)! Thank you so much @danielroe 

I've now spend 3 full days on solving this, because even after many learnings nuxt SSG and SSR is pretty much still a blockbox for me. 

So feedback from a noob:
One central resource or deep dive on what acutally happens in SSR and SSG.

That includes questions like: 
* Why, how and when is the fetch() hook called and cached?
* What are all these js files like `dist/_nuxt/4d03682.js`?
* And whats `dist/_nuxt/static/my-site/payload.js`? and `state.js`?! 
* Does my vuex store get cached/rendered/payloaded (you see the terminology is an issue for me :D)? Is there maybe even some kind of tree-shaking happening so only the used store data gets cached?

If there is a resource going in depth on this, I apologize and would kindly suggest to make it more prominent. I personally expected something like `Build` on the same level in the docs as `Features` and `Concepts`.

codetheorist:
> PLEASE PLEASE PLEASE merge this! Solves so many problems for us at https://github.com/funkhaus

That was an absolutely shameless plug.

imzedi:
@danielroe @Atinux 
If my component is imported using lazy import, fetch hook inside this component is called on both sever and client. i tried to add fetchKey but fetchHook called twice.

chz:
> @danielroe @Atinux If my component is imported using lazy import, fetch hook inside this component is called on both sever and client. i tried to add fetchKey but fetchHook called twice.

Same problem... @danielroe @pi0 any help?

PR #8475, opened 2020-12-11
fix(vue-app): add catchFetchError to throw on fetch error

### LINKED ISSUE #7742: "--fail-on-error" flag not working with fetch
### Versions

- nuxt: v2.13.0
- node: v12.12.0
- mode: universal
- target: static

### Reproduction
https://github.com/andre-brdoch/nuxt-fail-on-error

### Steps to reproduce
- run `$ npm run prerender`, or `$ nuxt build --fail-on-error && nuxt export --fail-on-error`  --> build fails
- go to `pages/index.vue`, comment out `asyncData()`, and uncomment `fetch()` 
- run `$ npm run prerender` --> build succeeds

### What is Expected?
Builds should fail if an error occurs either in `asyncData()` or in `fetch()`.

Build only fails with `asyncData()` throwing an error. It succeeds if `fetch()` throws an error.

### Additional Information
We want our builds to fail if anything goes wrong in either asyncData or fetch. Under no circumstances would we want to deploy our website, if for example the blog articles did not load 🙂 
Since we sometimes also request CMS data on component level, we have to use fetch instead of asyncData.

atinux:
## Types of changes

## Description

Fixes #7742

This introduce a new key in order to control the behaviour of catching or not an error in `fetch`.

`nuxt.config.js`
```js
export default {
  features: {
    catchFetchError: false
  }
}
```

This will throw an error in development (instead of adding the error into `$fetchState.error`), can be useful when you want the same behaviour as `asyncData` by default.

## Checklist:

### DIFF
--- a/packages/config/src/config/_app.js
+++ b/packages/config/src/config/_app.js
@@ -67,6 +67,7 @@ export default () => ({
     validate: true,
     asyncData: true,
     fetch: true,
+    catchFetchError: true,
     clientOnline: true,
     clientPrefetch: true,
     clientUseUrl: false,
--- a/packages/vue-app/template/mixins/fetch.client.js
+++ b/packages/vue-app/template/mixins/fetch.client.js
@@ -96,6 +96,7 @@ async function $_fetch() {
   let error = null
   const startTime = Date.now()
 
+  <% if (features.catchFetchError) { %>
   try {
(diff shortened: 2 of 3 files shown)


Atinux:
Closing until we find a nice way into Nuxt 3 to make the transition as smooth as possible.

PR #8489, opened 2020-12-14
fix(vue-app, vue-renderer, utils):  respect `trailingSlash` setting for payloads 

### LINKED ISSUE #8476: trailingSlash:true corrupts payloads
### Versions

- nuxt: 2.14.11
- node: 12.18.2

### Steps to reproduce

I'm using `generate` with `{ target: 'static' }` in my config to get a fully-static website.

I'm using `asyncData` to retrieve content from Prismic API.

And I'm using `{ router: { trailingSlash: true } }` in Nuxt config to add trailing slash all URLs.

Problem is, the generated payload files don't have a trailing slash in their keys, so pages's data remains blank, and new calls to API are made on each route change.

### LINKED ISSUE #8488: catchall routes with `trailingSlash: true` lead to 404s when generated
### Reproduction

https://github.com/danielroe/trailingslash-payload-repro

### Steps to reproduce

`yarn generate`

### What is Expected?

`index.html` should be generated

`ERROR   Error generating route "/": This page could not be found`

danielroe:
## Types of changes

## Description
This respects the route URL so payloads won't be refreshed when `trailingSlash: true`. Note this will still refresh payloads if the *link* doesn't respect the `trailingSlash` settings - there is still a need for a `<NuxtLink>` that respects the trailingSlash settings.

closes #8488
closes #8476 

## Checklist:

- [ ] ~~I have added tests to cover my changes~~ (will add tests in #8423)


pi0 (dismissed):
Notes for later:
- We can expose `withoutQuery`  from ufo later on (or simply using new ponyfill `.pathname`
- We may add tests for full-static to trailing-slash PR

packages/vue-app/template/App.js
```diff
@@ -296,19 +297,23 @@ export default {
...
     getRoutePath(route = '/') {
       const base = this.getRouterBase()
       if (base && route.startsWith(base)) {
         route = route.substr(base.length)
       }
-      return (route.replace(/\/+$/, '') || '/').split('?')[0].split('#')[0]
+      let path = route.split('?')[0].split('#')[0]
+      <% if (!nuxtOptions.router.trailingSlash) { %>
+        path = withoutTrailingSlash(path)
+      <% } %>
```
danielroe, line 310: We could normalise to no trailing slashes, which would also solve this problem, but would have the downside of breaking the distinction between parent/child routes (see #8423 - thoughts welcome @pi0).

packages/vue-app/template/App.js
```diff
@@ -296,19 +297,23 @@ export default {
...
-      return (this.$router.options.base || '').replace(/\/+$/, '')
+      return withoutTrailingSlash(this.$router.options.base)
     },
     getRoutePath(route = '/') {
       const base = this.getRouterBase()
       if (base && route.startsWith(base)) {
         route = route.substr(base.length)
       }
-      return (route.replace(/\/+$/, '') || '/').split('?')[0].split('#')[0]
+      let path = route.split('?')[0].split('#')[0]
```
Atinux, line 307: Could we expose à utiliser from ufo for this @pi0 ?

pi0: yes. see comment above 😁

pi0: FYI it is possible with latest version of ufo to do `parsePath(route).pathname`

pi0:
Should be available shortly with `nuxt-edge` and landed with `2.15.0` release.

Issue #8493, opened 2020-12-15
Having "https:/" in my url causes internal NodeJS error


gkatsanos:
Hey folks, whenever there's a `https:/` in my URL, I get this error (instead of my 404 page):
Invalid URL: /https:/

```
next@183:5
internal/url.js:258:9
onParseError
internal/url.js:334:5
new URL
node_modules/@nuxt/vue-renderer/dist/vue-renderer.js:2106:15
parseURL
node_modules/@nuxt/vue-renderer/dist/vue-renderer.js:2110:56
normalizeURL
node_modules/@nuxt/vue-renderer/dist/vue-renderer.js:4436:25
VueRenderer.renderRoute
node_modules/@nuxt/server/dist/server.js:918:38
Server.renderRoute
node_modules/@nuxt/server/dist/server.js:158:26
nuxtMiddleware
node_modules/connect/index.js:239:7
call
node_modules/connect/index.js:183:5
next
node_modules/serve-placeholder/src/index.js:29:16
servePlaceholderMiddleware
node_modules/connect/index.js:239:7
call
```

gkatsanos:
Thank you @danielroe. I struggled thinking it's coming from our app but I realized it's something from core. By the way any part of the URL will have the same result, even as a part of a query (as redirect query for example)

pi0:
Hi @gkatsanos thanks for reporting. Would you please try [v2.14.12](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.12)?

gkatsanos:
Thanks guys! amazing speed !

PR #8462, opened 2020-12-09
fix: fix issues with route normalization

pi0:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

History: vuejs/vue-router#3350 > #8325 > #8421 > #8430

- Add `url-polyfill` (2.1 kB min/gzip) to fix IE11 issue (URL is not a constructor) (resolves #8458)
- Update to [@nuxt/ufo@0.1.0](https://github.com/nuxt-contrib/ufo/blob/main/CHANGELOG.md#010-2020-12-09) preserving relative links (resovles #8457, #8459)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

Issue #8458, opened 2020-12-09
"export 'normalizeURL' was not found in '@nuxt/ufo' [IE]


andorfermichael:
<!-- 💚 Thanks for your time to make Nuxt better with your feedbacks 💚

**IMPORTANT** Before reporting a bug:

- Please make sure that you have read through Nuxt documentation: https://nuxtjs.org
- If issue is related to a module please create the issue in corresponding repository
- Ensure using latest version of nuxt dependencies using `yarn upgrade nuxt` or `npm upgrade nuxt`

👍 A properly detailed bug report can save a LOT of time and help fixing issues as soon as possible.
-->

### Versions

- nuxt: v2.14.10
- node: v14.15.0 

<!--### Reproduction-->

<!--
Link to a minimal test case based on one of:
- A fork of https://template.nuxtjs.org
- A GitHub repository that can reproduce the bug
Without a reproduction, it is so hard to address problem :(
-->

<!--<details open>
<summary>Additional Details</summary>
<br> -->
<!-- Attaching `nuxt.config`, dependencies, logs or code snippets would help to find the issue -->
</details>

### What is actually happening?

When running `nuxt` with the latest release 2.14.10, the following warning occurs:

`"export 'normalizeURL' was not found in '@nuxt/ufo'`

#### Command Line Output
```
$ nuxt
ℹ Using Tailwind CSS from ~/assets/css/tailwind.css                                                                                                                                                                      nuxt:tailwindcss 10:30:21
ℹ Merging Tailwind config from ~/tailwind.config.js                                                                                                                                                                      nuxt:tailwindcss 10:30:21
ℹ [HPM] Proxy created: /  -> https://staging.example.com                                                                                                                                                                                 10:30:22
ℹ [HPM] Proxy rewrite rule created: "^/api/" ~> "/spa/"                                                                                                                                                                                   10:30:22

   ╭───────────────────────────────────────╮
   │                                       │
   │   Nuxt @ v2.14.10                     │
   │                                       │
   │   ▸ Environment: development          │
   │   ▸ Rendering:   server-side          │
   │   ▸ Target:      server               │
   │                                       │
   │   Listening: http://localhost:3000/   │
   │                                       │
   ╰───────────────────────────────────────╯

ℹ Preparing project for development                                                                                                                                                                                                       10:30:23
ℹ Initial build may take a while                                                                                                                                                                                                          10:30:23
[...]

Issue #8457, opened 2020-12-09
V2.14.10, statically generated router.base root path error.


xldsdr124:
### Versions

- nuxt: v2.14.10
- node: v14.15.1

### Reproduction

<details open>
<summary>Additional Details</summary>
<br>

</details>

### Steps to reproduce
 this is part of my configuration:
`{
    ssr: false,
    router: {
        mode: "hash",
        base: "./"
    },
}`

Next execute "yarn run build"

### What is Expected?
The root path of the dist/index.html script should point to router.base.

### What is actually happening?
Upgrade from 2.14.9 to 2.14.10, build output static website, dist/index.html script root path always points to "/".

pi0:
Hi. Thanks for report. This should be resolved with [v2.14.11](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.11).

xldsdr124:
> Hi. Thanks for report. This should be resolved with [v2.14.11](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.11).

I confirm that this problem still exists in 2.14.11.

maximilianschmid:
Problem still exists in 2.14.11

fuxianwei:
Problem still exists in 2.14.11

pi0:
Would you please try [v2.14.12](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.12)?

maximilianschmid:
v2.14.12 works, all my tests green again. THX!

PR #8494, opened 2020-12-15
hotfix: fix issues with url handling

pi0:
## Types of changes

## Description

🚂 History: vuejs/vue-router#3350 > #8325 > #8421 > #8430 > #8462

✅ Resolves: #8497, #8493, #8458, #8457 

ℹ️  Update ufo to 0.5.x:

- Use custom parser instead of depending on global URL parser
- Remove url-polyfill
- Encode with same utils of vue-router-next
- `browser` filed removed as UMD cannot be used for named exports and we have same implementation for both envs
- Add missing `@nuxt/ufo` dependency to `@nuxt/vue-renderer` to avoid inlining

## Checklist:

### DIFF
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -63,7 +63,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       cacheGroups.commons === undefined
     ) {
       cacheGroups.commons = {
-        test: /node_modules[\\/](vue|vue-loader|vue-router|vuex|vue-meta|core-js|@babel\/runtime|axios|webpack|setimmediate|timers-browserify|process|regenerator-runtime|cookie|js-cookie|is-buffer|dotprop|url-polyfill|nuxt\.js)[\\/]/,
+        test: /node_modules[\\/](vue|vue-loader|vue-router|vuex|vue-meta|core-js|@babel\/runtime|axios|webpack|setimmediate|timers-browserify|process|regenerator-runtime|cookie|js-cookie|is-buffer|dotprop|url-polyfill|@nuxt[\\/]ufo|ufo|nuxt\.js)[\\/]/,
         chunks: 'all',
         name: true,
         priority: 10
--- a/packages/webpack/src/config/client.js
+++ b/packages/webpack/src/config/client.js
@@ -205,11 +205,6 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       )
     }
 
-    // Add URL polyfill for IE11 support with ufo
-    if (!this.isModern && !this.isServer) {
-      config.entry.app.unshift(require.resolve('url-polyfill/url-polyfill.min.js'))
-    }
-
     // Add friendly error plugin
     if (this.dev && !quiet && friendlyErrors) {
       config.plugins.push(
--- a/test/dev/async-config.size-limit.test.js
+++ b/test/dev/async-config.size-limit.test.js
@@ -23,7 +23,7 @@ describe('nuxt basic resources size limit', () => {
     const LEGACY_JS_RESOURCES_KB_SIZE = 217
     expect(legacyResourcesSize.uncompressed).toBeWithinSize(LEGACY_JS_RESOURCES_KB_SIZE)
 
-    const LEGACY_JS_RESOURCES_GZIP_KB_SIZE = 70
+    const LEGACY_JS_RESOURCES_GZIP_KB_SIZE = 75
     expect(legacyResourcesSize.gzip).toBeWithinSize(LEGACY_JS_RESOURCES_GZIP_KB_SIZE)
 
     const LEGACY_JS_RESOURCES_BROTLI_KB_SIZE = 64
@@ -36,7 +36,7 @@ describe('nuxt basic resources size limit', () => {
     const MODERN_JS_RESOURCES_KB_SIZE = 180
     expect(modernResourcesSize.uncompressed).toBeWithinSize(MODERN_JS_RESOURCES_KB_SIZE)
 
-    const MODERN_JS_RESOURCES_GZIP_KB_SIZE = 60
+    const MODERN_JS_RESOURCES_GZIP_KB_SIZE = 64
     expect(modernResourcesSize.gzip).toBeWithinSize(MODERN_JS_RESOURCES_GZIP_KB_SIZE)
 
     const MODERN_JS_RESOURCES_BROTLI_KB_SIZE = 55
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -10,7 +10,7 @@
     "index.d.ts"
   ],
   "dependencies": {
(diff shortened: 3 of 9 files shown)


danielroe:
A few issues I've found so far:

1. IE behaviour in encoding fixture to non-URL-encoded direct dynamic link doesn't match on server side but this may be IE issue rather than ufo:
   ```
   /%C3%B6/dynamic/سلام چطوری?q=cofee,food,دسر
   ```
   <details><summary>Screenshot</summary>

   Screenshot from 2020-12-16 10-40-11

   </details>

2. Edge 18 has a hydration issue on `/тест` direct and doesn't recognise matches for the non-URL-encoded direct dynamic link 
3. Safari 5.1 has an issue with most URLs on client-side (indicating differing behaviour between node/client code) - but this is likely too old for us to support

However, evergreen browsers all appear to work with consistent behaviour, and I think the issues highlighted here would probably have been present without `@nuxt/ufo`. With a browser testing suite, we may also be able to identify and address within the ufo project.

packages/webpack/src/config/client.js
```diff
@@ -63,7 +63,7 @@ export default class WebpackClientConfig extends WebpackBaseConfig {
       cacheGroups.commons === undefined
     ) {
       cacheGroups.commons = {
-        test: /node_modules[\\/](vue|vue-loader|vue-router|vuex|vue-meta|core-js|@babel\/runtime|axios|webpack|setimmediate|timers-browserify|process|regenerator-runtime|cookie|js-cookie|is-buffer|dotprop|url-polyfill|nuxt\.js)[\\/]/,
+        test: /node_modules[\\/](vue|vue-loader|vue-router|vuex|vue-meta|core-js|@babel\/runtime|axios|webpack|setimmediate|timers-browserify|process|regenerator-runtime|cookie|js-cookie|is-buffer|dotprop|url-polyfill|@nuxt[\\/]ufo|ufo|nuxt\.js)[\\/]/,
```
clarkdo, line 66: `url-polyfill` here can be removed

test/dev/async-config.size-limit.test.js
```diff
@@ -23,7 +23,7 @@ describe('nuxt basic resources size limit', () => {
     const LEGACY_JS_RESOURCES_KB_SIZE = 217
     expect(legacyResourcesSize.uncompressed).toBeWithinSize(LEGACY_JS_RESOURCES_KB_SIZE)
 
-    const LEGACY_JS_RESOURCES_GZIP_KB_SIZE = 70
+    const LEGACY_JS_RESOURCES_GZIP_KB_SIZE = 75
```
clarkdo, line 26: We've removed url-polyfill, any specific reason that the size is increased ?

packages/webpack/src/config/client.js
pi0: It kept intentionally if some dependency added it to be in commons...

test/dev/async-config.size-limit.test.js
pi0: It is mostly because of standalone encoding utils / ponycode added to ufo bundle (in 0.5.0 i've removed dependency on it when tree-shaking but still normalizeURL would need it)

pi0: FYI this is entire implementation: https://unpkg.com/@nuxt/ufo@0.5.0/dist/index.js

clarkdo: The ufo bundle doesn't look bigger than url-polyfill + 5KB at my first sight.

pi0: ~2.5kB untreeshaked [bundlephobia](https://bundlephobia.com/result?p=@nuxt/ufo@0.5.0) . Legacy bundle:

image


Modern:

image


Conclusion: Overhead is mostly babel transpilation. I will try to see if we can use some hacks to avoid or reduce it

pi0: BTW gzip diff is +1KB. It is strange with tests too....

clarkdo: You can remove the range and see real diff https://github.com/nuxt/nuxt.js/blob/7d18b06f15bcbd9f466335cc695f71177a6add6e/test/dev/async-config.size-limit.test.js#L9

pi0:
@danielroe Thanks for investigation.

Quickly checking IE issue, it seems that is original URL sent by IE is wrong (i think it was an issue before vue-router update as well)

- ctx.url `/dynamic/سلام چطوری?q=cofee,food,Ø¯Ø³Ø±` (which is incorrect)
- normalized: `/dynamic/%D8%B3%D9%84%D8%A7%D9%85%20%DA%86%D8%B7%D9%88%D8%B1%DB%8C?q=cofee,food,%C3%98%C2%AF%C3%98%C2%B3%C3%98%C2%B1` (which is correct but input was incorrect :D) 

For encoding fixture, `(direct)` items are directly using `<a>` tag (without ufo normalization) while we do normalize for `NuxtLink` items so clicking on them or opening in new tab always works since ufo normalized links. So for all browsers support users have to use RouterLink/Nuxtlink or encode themselves for `<a>` tags

test/dev/async-config.size-limit.test.js
pi0: @clarkdo Regardless of tests, i think _overhead_ is correct from bundle analyzer :D 2-3KB/gzip overhead of ufo is almost unavoidable and by using polyfills inside ufo to reduce transpilation overhead, we might add overhead to modern build.

clarkdo: >  2-3KB/gzip overhead of ufo

This looks good to me. I agree that not adding polyfills in ufo package.

gkatsanos:
hey guys, thank you for fixing this! What is the release process for this?

pi0:
@gkatsanos Hotfixes are immediate :) Would be nice if can confirm fix.

gkatsanos:
Confirmed.

alvinindra:
image

Hello, are uuid already fix on new release?

pi0:
@alvinindra Quickly checking seems yes :) (https://runkit.com/pi0/5fdb46cd9a8519001ac7e9b5). You can upgrade using `yarn upgrade nuxt` and please make an issue with sandbox if not working.

awronski:
@pi0 I just installed nuxt 2.14.12 and have problems with the urls handling.

Let's say I hava a url like this: `http://127.0.0.1/unsunbscribe?email=some@email.com&token=DvtiwbIzry319e6KWimopA%3D%3D`

```js
    async asyncData ({ store, query, error }) {
      const email = query.email
      const token = query.token 
      [...]
    }
```

Nuxt should decode it to:
- email: `some@email.com`  - OK!
- token: `DvtiwbIzry319e6KWimopA==` - ERROR

Instead I get `DvtiwbIzry319e6KWimopA`

Could you help?

Issue #8229, opened 2020-10-20
dynamic publicPath


rstormsf:
I have a following problem:
When I build the app, it should be working on my primary domain and also as ipfs hash
example: `ipfs.myapp.com` OR `ipfs.io/ipfs/myappcom`

the problem is with generating the nuxt build. For main domain, the default works just fine
```
build: { publicPath: '/_nuxt/' } 
```
but if I need to generate it for ipfs.io/ipfs/xxx, it has to be 
```
build: { publicPath: './_nuxt/' } 
```

is there a way to make a function that will try to fetch `/_nuxt/` folder by defining how.
example:
```js
publicPath: (resolve) => {
    let pathToNuxt = await resolve('/_nuxt/')
    if(pathToNuxt.failed) {
        pathToNuxt = await resolve('./_nuxt/')
    }
    return pathToNuxt
}
```
something like this

P.S.
`build: { publicPath: './_nuxt/' } ` still not ideal cause it breaks the app if it's not accessed from the root path. But it's better than nothing

If there is any other way to archive this goal, please let me know

jpsc:
I think you can pass a [variable from your CI](https://nuxtjs.org/api/configuration-env#automatic-injection-of-environment-variables) to you `nuxt.config.js` for this.

`NUXT_ENV_PUBLIC_PATH="./nuxt/" nuxt build`

rstormsf:
> I think you can pass a [variable from your CI](https://nuxtjs.org/api/configuration-env#automatic-injection-of-environment-variables) to you `nuxt.config.js` for this.
> 
> `NUXT_ENV_PUBLIC_PATH="./nuxt/" nuxt build`

That won't work because when I deploy to IPFS, my link of `ipfs.mydomain.com` is using ipfs-hash to load the content and some users should also be allowed to access it from `ipfs.io/ipfs/mydomain.com` so I can't have 2 builds, 1 build should work for all.

rstormsf:
bump :-(

rotemrevivo91:
joining on the request

pi0:
Thanks for making issue @rstormsf. It is indeed a must have feature. I'm closing in favor of #8509 to track.

Issue #8033, opened 2020-09-09
Dynamic webpack_public_path working but not on dynamic routes


digitalfuel:
### Versions

- nuxt: v2.14.4 
- node: v13.12.0 

### Steps to reproduce

Add global variable __webpack_public_path__ and documented. https://webpack.js.org/guides/public-path/#on-the-fly.
Run NUXT Generate

directory eg
```
-| pages/
---| users/
-----| _id.vue
-----| index.vue
```

nuxt.config.js
```
plugins: [{ src: "~/plugins/cdn", mode: "client" }],
```

cdn.js
```
export default async ({ store }) => {
  const cdn = store.state["globals"].cdn;
  __webpack_public_path__ = cdn + "/_nuxt/";
};
```

### What is Expected?
Dynamic publicPath for different CDN on different instances.

### What is actually happening?
This works for standard routes on entry and navigation. 

For dynamic routes it works for navigation but on entry(refresh) it reverts to the publicPath default or defined.

In the webpack doc it says _'Be aware that if you use ES6 module imports in your entry file the __webpack_public_path__ assignment will be done after the imports. In such cases, you'll have to move the public path assignment to its own dedicated module and then import it on top of your entry.js:'_

So it maybe an issue with the __webpack_public_path__ not running early enough. I have tried everything I can think of but I cannot get it to work for dynamic routes. 

The chunk is loaded correctly using the dynamic __webpack_public_path__ for standard routes and when navigating to the dynamic route but not when the page is refreshed for the dynamic route. Also it is only the chunk of the dynamic route that fails and reverts to the build.publicPath all the other chunks on the dynamics routes page use __webpack_public_path__ correctly.

Is there anyway to get this fixed working for dynamic routes with generate.

rotemrevivo91:
any luck? this is a bit ridiculous tbh. nuxt is perfect for development but trash for production

danielroe:
I think this feature request is probably what you want: https://github.com/nuxt/nuxt.js/issues/8229

PR #8520, opened 2020-12-19
feat(config, vue-app, vue-renderer): support dynamic `base` and `publicPath`

### LINKED ISSUE #7287: Support dynamic router base path

We're currently building a universal Nuxt-based web app that will be available via multiple reverse proxies with distinct URLs, for example https://project1.com/abc/def and https://project2.com/def/qwerty. That doesn't work out of the box, because Nuxt configures the vue-router with a static base path, so all the reverse proxies would need to pass the user-visible URL as-is.

As a workaround, we're currently configuring Nuxt with `router: { base: '/' }`, let the reverse proxy rewrite the `<base href>` tag and use a patch to let the router actually use rewritten URL (not that this doesn't support IE11 and is probably not the best way to do it anyway).

```
diff --git a/node_modules/@nuxt/vue-app/template/router.js b/node_modules/@nuxt/vue-app/template/router.js
index bcbc109..24627cb 100644
--- a/node_modules/@nuxt/vue-app/template/router.js
+++ b/node_modules/@nuxt/vue-app/template/router.js
@@ -93,7 +93,9 @@ Vue.use(Router)
 
   mode: '<%= router.mode %>',
-  base: decodeURI('<%= router.base %>'),
+  base: process.client ? new URL(document.baseURI).pathname : decodeURI('<%= router.base %>'),
   linkActiveClass: '<%= router.linkActiveClass %>',
   linkExactActiveClass: '<%= router.linkExactActiveClass %>',
   scrollBehavior,
```

I think Nuxt should automatically use the configured `<base href="...">` to initialize the vue-router instead of the build-time constant from the `nuxt.config.js`.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10587">#c10587</a>)</em></sub></div>

### LINKED ISSUE #8509: supporting runtime baseURL/publicPath
### Description

Nuxt doesn't allow the publicPath (CDN url) to be overridden at run time - it's baked into the manifest at
build time.  This hook intercepts the VueRenderer when it has loaded the manifest and updates the publicPath
to the current env value.

The problem is that i have a single build (i.e 1 docker image) that is used on multiple environments. Each user should be able to specify their own router base as an .env variable and just run the container without the need to rebuild the image.
Nuxt resources are then can't be found (nuxt .js files).

When building and running for the first time (nuxt build && nuxt start) everything works normally. Then changing the BASE value in the .env file breaks the app.

### Solution description

Setting the baseURL anew on runtime

### Describe alternatives you've considered
.env:
```
BASE=\rotem\
```
nuxt.config.js:
```
  publicRuntimeConfig: {
    baseURLSubDirectory: process.env.BASE_URL_SUB_DIRECTORY,
  },
  router: {
    base: process.env.BASE_URL_SUB_DIRECTORY
  },
  plugins: [
    {
      src: '@/plugins/cdn.js',
      ssr: false
    },
  ],
  hooks: {
    /* Nuxt doesn't allow the publicPath (CDN url) to be overridden at run time - it's baked into the manifest at
    ** build time.  This hook intercepts the VueRenderer when it has loaded the manifest and updates the publicPath
    */ to the current env value.
    render: {
      resourcesLoaded(resources) {
        const path = `${process.env.BASE_URL_SUB_DIRECTORY}_nuxt/`
        resources.clientManifest && (resources.clientManifest.publicPath = path)
        resources.modernManifest && (resources.modernManifest.publicPath = path)
        resources.serverManifest && (resources.serverManifest.publicPath = path)
      }
    }
  }
```
plugins\cdn.js:
```
export default function({$config}) {
  if (process.client) {
    __webpack_public_path__ = `${$config.baseURLSubDirectory}_nuxt/`
  }
}
```
Build nuxt for production and then change .env:
```
BASE=\xrotem\
```
Then run the nuxt app.
The code replaces all resources on run-time! which is important but still i get a 404 page even though nuxt is able to find all the needed resources.

danielroe:
## Types of changes

## Description
This PR allows [`router.base`](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router#base) and [`build.publicPath`](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-build#publicpath) to be configured at runtime. At runtime, these values will be evaluated. The router will be configured to reflect the new `router.base`, and injected scripts will then be updated with the correct `publicPath` (such as a CDN), which includes setting `__webpack_public_path__` so that the built JS assets point to the correct location. 

### Caveats
* Other modules may not yet support this capability. For example, `@nuxtjs/pwa` does not _yet_ use the runtime values of these options.
*  This does not affect statically generated sites - as there is no server to inject environment variables into the static HTML.

closes #8509, closes #7287
see also #8229, #8033

## Checklist:
- [x] I have updated the documentation accordingly: https://github.com/nuxt/nuxtjs.org/pull/1092


packages/config/src/load.js
```diff
@@ -117,6 +118,23 @@ export async function loadNuxtConfig ({
     }
   }
 
+  if (options.publicRuntimeConfig) {
```
pi0, line 121: - This logic could be better in `options.js` after applying defaults
- We may set value of `options.router.base` to `publicRuntimeConfig._nuxt.routerBase` so **not introducing a new option** but changing router.base in production applies to runtime
- As you suggested it might be better to use `nuxt` / `_nuxt` namespace or something unique inside `$config`
- We can simplify names to something like `routerBase` and `publicPath` in nuxt's runtimeConfig namespace since they are internal for directly reading from `$config`

packages/vue-app/template/client.js
```diff
@@ -45,6 +47,19 @@ let router
 // Try to rehydrate SSR data from window
 const NUXT = window.<%= globals.context %> || {}
 
+let publicPath = NUXT.config.build && NUXT.config.build.publicPath
```
pi0, line 50: ~> It would be simply `const nuxtConfig = NUXT.config.nuxt || {}`

packages/vue-app/template/router.js
```diff
@@ -114,8 +114,12 @@ function decodeObj(obj) {
...
 }
 
-export function createRouter () {
-  const router = new Router(routerOptions)
+export function createRouter (ssrContext, config) {
+  let base = config.router && config.router.base
+  if (typeof base === 'string') {
+    base = withTrailingSlash(normalizeURL(base))
+  }
+  const router = new Router(Object.assign({}, routerOptions, { base: base || routerOptions.base }))
```
pi0, line 122:
```suggestion
  const router = new Router({ ...routerOptions, base: base || routerOptions.base })
```

packages/vue-app/template/router.js
```diff
@@ -114,8 +114,12 @@ function decodeObj(obj) {
   }
 }
 
-export function createRouter () {
-  const router = new Router(routerOptions)
+export function createRouter (ssrContext, config) {
+  let base = config.router && config.router.base
```
pi0, line 118: Not sure but don't we directly have access to `$config` from ssrContext?

packages/vue-app/template/client.js
```diff
@@ -45,6 +47,19 @@ let router
...
+  
+if (typeof publicPath === 'string') {
+  let base = NUXT.config.router && NUXT.config.router.base
+  if (typeof base === 'string') {
+    base = withTrailingSlash(normalizeURL(base))
+  }
+  publicPath = withTrailingSlash(normalizeURL(publicPath))
+  __webpack_public_path__ = ['http', '//'].some(str => publicPath.startsWith(str))
```
pi0, line 58: Can be simplified with `/^http|\/\//` ?

packages/vue-app/template/router.js
danielroe: Yes, but only on server side.

packages/config/src/load.js
danielroe: Great idea. That lets us simplify this considerably. I've left the logic in `load.js` for the moment (thinking to ensure we have the last say on content of `_nuxt`) but happy to move into `options.js` if you feel that's more logic.

packages/config/src/options.js
```diff
@@ -452,17 +452,27 @@ export function getNuxtConfig (_options) {
       .map(([path, handler]) => ({ path, handler }))
   }
 
+  // App config (internal for nuxt2 at this stage)
+  const useCDN = isUrl(options.build.publicPath) && !options.dev
+  options.app = defu(options.app, {
+    basePath: options.router.base,
+    assetsPath: useCDN ? '/' : joinURL(options.router.base, options.build.publicPath),
+    cdnURL: useCDN ? options.build.publicPath : undefined
```
danielroe, line 460: is there any merit in defaulting to `'/'`?

pi0: It is something i still doubt about but at least this ensures we always check it's existence and when it is fasly, useCDN is false

danielroe: Then I would go for null, perhaps?

pi0: Makes sense. updated.

rstormsf:
@pi0 when this PR will be released?

pi0:
Hi @rstormsf. You can try it by changing nuxt to nuxt-edge in dependencies. 2.15 will be released jan.

mrkamel:
I'm trying out nuxt-edge to be able to set publicPath dynamically and use a cdn url, but i have issues with e.g. img tags, as `` is not rendered with runtime publicPath, but build time public path. Same for all assets referenced in css, such that `extractCSS: true` is neccessary and fixes it (but publicPath must match absolute cdn path, i.e. without hostname, at build time to have the absolute cdn paths inside the built css files). However, i was not able to get img tags to work, as the cdn hostname is missing in the final markup. Do you have any suggestions how to make those work as well?

wtagain:
Hi @danielroe  
How can I set publicPath dynamically? Can you give a sample code? Thank you very much.

wensaint:
Is there any example.

my code below not working. 
/***************/
.env
`PUBLIC_PATH = '/test'`

.nuxt.config.js
`router: {
    // base: process.env.PUBLIC_PATH
  },
publicRuntimeConfig: {
    router: {
      base: process.env.PUBLIC_PATH || '/'
    },
    build: {
      publicPath: process.env.PUBLIC_PATH || '' + '/_nuxt/'
    }
  }
`
with these settings.
http://localhost:3000/ works fine, but when visit http://localhost:3000/test. i got 404 error.

wensaint:
Is there any example?

my code below not working. 
/***************/
.env
`PUBLIC_PATH = '/test'`

.nuxt.config.js
`router: {
    // base: process.env.PUBLIC_PATH
  },
publicRuntimeConfig: {
    router: {
      base: process.env.PUBLIC_PATH || '/'
    },
    build: {
      publicPath: process.env.PUBLIC_PATH || '' + '/_nuxt/'
    }
  }
`
with these settings.
http://localhost:3000/ works fine, but when visit http://localhost:3000/test. i got 404 error.

hassan-jahan:
Same problem here, Is there any working example here?

jlkalberer:
@hassan-jahan @wensaint  did you ever figure out a solution for this?

YutaMoriJP:
@wensaint Did you figure out a solution for this? Or could @pi0 please chime in...

PR #8554, opened 2020-12-26
POC of orchestrated components (multi-app)

brophdawg11:
This is a quick POC based on [this comment](https://github.com/nuxt/rfcs/issues/30#issuecomment-744647345) in https://github.com/nuxt/rfcs/issues/30.  I initially looked into doing this with a module but didn't think I would have the right ability to hook in and make changes to the templates without providing custom templates (which would quickly become out of sync with general nuxt upgrades).  However, if I missed something and this could be done via a sustainable module approach, I'd be happy to try to convert it over.

## Types of changes

## Description
The implementation here provides 2 new related features:

* The ability to orchestrate Nuxt server-rendered components into a different application in which Nuxt/Vue may not be in control of the full page or routing - and thus would not want a full `html`/`head`/`body` structure in the SSR'd HTML
* The ability to instantiate _multiple_ Vue/Nuxt apps on a single page

The specific example implementation here is the use-case that my team is currently doing with our own `vue-server-renderer` build.  We're looking to migrate to Nuxt but this seems to be the only thing standing in our way.  We are able to get most of the way there with plugins, a custom `app.html`, etc - but the main thing we cannot get past is the `_app.$mount('#__nuxt')` aspect which sets a global limitation of a single app per-page.  Beyond that, there introduce some minor changes to the router and the ability to share a Vuex instance across the apps.

In addition to this type use-case of "plugging" Nuxt/Vue components into existing SSR apps, I think these new features could also potentially have use-cases in the micro-frontend landscape, but I haven't spent too much time thinking through those.

This addresses I think a slightly different issue that what is posed in https://github.com/nuxt/rfcs/issues/30, but looking for confirmation on that and happy to open a new RFC for discussion.

### Testing Steps
* Checkout this branch
* Tab #1: `yarn nuxt examples/orchestrated-apps/`
* Tab #2: `cd examples/orchestrated-apps && npm install && node server.js`
* http://localhost:3000/ - Loads the normal Nuxt app, rendering header/homepage/footer content
* http://localhost:3001/ - Loads the Express non-Nuxt app with the orchestrated Nuxt header/footer

## Checklist:

(I haven't looked at tests or documentation yet, but I will if this gains any traction)

### DIFF
--- a/examples/orchestrated-apps/app.html
+++ b/examples/orchestrated-apps/app.html
@@ -1,0 +1,12 @@
+{% if (!orchestrated) { %}<!DOCTYPE html>
+<html {{ HTML_ATTRS }}>
+    <head {{ HEAD_ATTRS }}>
+        {{ HEAD }}
+    </head>
+    <body {{ BODY_ATTRS }}>
+        {{ APP }}
+    </body>
+</html>
+{% } else { %}
+{{ APP }}
+{% } %}
--- a/examples/orchestrated-apps/layouts/orchestrated.vue
+++ b/examples/orchestrated-apps/layouts/orchestrated.vue
@@ -1,0 +1,3 @@
+<template>
+  <Nuxt />
+</template>
--- a/examples/orchestrated-apps/nuxt.config.js
+++ b/examples/orchestrated-apps/nuxt.config.js
@@ -1,0 +1,19 @@
+export default {
+  components: true,
+  features: {
+    store: true,
+    layouts: true,
+    meta: false,
+    middleware: false,
+    transitions: false,
+    deprecations: false,
+    validate: false,
+    asyncData: false,
+    fetch: false,
+    clientOnline: false,
+    clientPrefetch: false,
+    clientUseUrl: false,
+    componentAliases: false,
+    componentClientOnly: false
+  }
+}
--- a/examples/orchestrated-apps/nuxt.config.js
+++ b/examples/orchestrated-apps/nuxt.config.js
@@ -1,5 +1,29 @@
 export default {
   components: true,
+  plugins: [
+    '~/plugins/orchestrated.server'
+  ],
+  hooks: {
+    'vue-renderer': {
+      ssr: {
+        context (renderContext) {
+          // Disable style/script injection based on query params parsed
+          // in plugins/orchestrated.server.js
+          if (renderContext.excludeStyles) {
+            renderContext.renderStyles = () => '<!-- styles disabled -->'
+          }
+          if (renderContext.excludeScripts) {
+            renderContext.renderResourceHints = () => '<!-- resource hints disabled -->'
+            renderContext.renderScripts = () => '<!-- scripts disabled -->'
+          }
+        },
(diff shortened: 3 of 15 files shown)


examples/orchestrated-apps/app.html
```diff
@@ -0,0 +1,12 @@
+{% if (!orchestrated) { %}<!DOCTYPE html>
```
brophdawg11, line 1: Use a `templateParam` to opt-out of the `html`/`head`/`body` wrapper markup

examples/orchestrated-apps/layouts/orchestrated.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+  <Nuxt />
```
brophdawg11, line 2: Pass-through layout for single-component rendering

examples/orchestrated-apps/nuxt.config.js
```diff
@@ -0,0 +1,43 @@
...
+  components: true,
+  plugins: [
+    '~/plugins/orchestrated.server'
+  ],
+  hooks: {
+    'vue-renderer': {
+      ssr: {
+        context (renderContext) {
+          // Disable style/script injection based on query params parsed
+          // in plugins/orchestrated.server.js
```
brophdawg11, line 11: This is used to avoid double-inclusion when multiple components are orchestrated into the page

examples/orchestrated-apps/pages/orchestrated/_component.vue
```diff
@@ -0,0 +1,22 @@
...
+}
+export default {
+  layout: 'orchestrated',
+  render (h) {
+    const component = process.server
+      ? this.$route.params.component
+      : this.$root.$options.orchestrated
+    return h(componentMap[component], {
+      attrs: {
+        'data-orchestrated-app': component
```
brophdawg11, line 17: This is what we hook into to detect this scenario client side and hydrate apps individually

examples/orchestrated-apps/pages/orchestrated/_component.vue
```diff
@@ -0,0 +1,22 @@
...
+const componentMap = {
+  footer: FooterComponent,
+  header: HeaderComponent
+}
+export default {
+  layout: 'orchestrated',
+  render (h) {
+    const component = process.server
+      ? this.$route.params.component
+      : this.$root.$options.orchestrated
```
brophdawg11, line 14: This isn't ideal, but client-side we can't rely on the route since it's unrelated to our orchestrated component.  There is probably a cleaner way to do this

packages/vue-app/template/router.js
```diff
@@ -110,6 +110,12 @@ function decodeObj(obj) {
 }
 
 export function createRouter (ssrContext, config) {
+  if (config.orchestrated) {
+    routerOptions.routes = [{
+      ...routerOptions.routes.find(r => r.name === "orchestrated-component"),
+      path: '*',
```
brophdawg11, line 116: Client-side - have all routes render the orchestrated component

PR #8557, opened 2020-12-26
fix(vue-app): scroll to top on route changes only

### LINKED ISSUE #8553: Reloading the page causes the scroll position to be reset to the top
### Previous Issue
https://github.com/nuxt/nuxt.js/issues/3471

### Versions

- nuxt: v2.14.12
- node: v12.20.0

### Reproduction

minimal nuxt installation
OR
https://nuxtjs.org/docs/2.x/get-started/installation

### Steps to reproduce

1. Load nuxt site.
2. Scroll down.
3. Refresh the page.

### What is Expected?

The scroll position should be the same as it was before reloading.

The scroll is reset to the top of the page.

### Additional comments?

The default scroll behavior always scrolls to the top of page and happens during page load.

https://github.com/nuxt/nuxt.js/blob/79440f3ce5b90b39fc1fa68fe1736cac6edae491/packages/vue-app/template/router.scrollBehavior.js#L31-L39

The current workaround is adding custom scroll behavior and implementing a page load check.
```js
// Page load
if (to !== from) {
  position = { x: 0, y: 0 }
}
```

beornf:
## Types of changes

## Description
This change fixes the default scroll behavior by only scrolling to the top of page when the route changes.
Resolves #8553.

## Checklist:

### DIFF
--- a/packages/vue-app/template/router.scrollBehavior.js
+++ b/packages/vue-app/template/router.scrollBehavior.js
@@ -31,9 +31,11 @@ export default function (to, from, savedPosition) {
   // Scroll to the top of the page if...
   if (
       // One of the children set `scrollToTop`
-      Pages.some(Page => Page.options.scrollToTop) ||
+      (Pages.some(Page => Page.options.scrollToTop) ||
       // scrollToTop set in only page without children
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks!

PR #8611, opened 2021-01-09
fix(vue-app): don't remove req/res/query unless in full static mode

### LINKED ISSUE #7965: Universal app built via nuxt generate no longer has correct query params when running
### Versions

- nuxt: 2.14.3, 2.14.4 2.13.3 (any version > 2.12.2)
- node: 12.18.3

### Reproduction

https://github.com/jonthemiller/nuxt-generate-query

<details open>
<br>
This repo is for a universal app that has a router middleware that takes a query param, and if present, sets its value as a header for the axios plugin. When run in dev mode, everything works as expected. But, when run via `nuxt start`, after building with `nuxt generate`, the `context` contains an empty object for the key `query`. This produces different behavior in prod vs dev that has caused a couple days of headaches trying to track down this behavior.

This is behavior that changed starting in version 2.13.x. In previous versions (2.12.x), everything worked as expected with the query context and generate.

Further - we're using `nuxt generate` in our current Nuxt project to create a static 404/500 page at build time. This file is then placed in `.nuxt/views/error.html` to replace the default Nuxt SSR error page. (It is also use by nginx in case of an unrecoverable error but thats unrelated). This is the only page being generated and only use case we have for `generate`.
</details>

### Steps to reproduce

1. Clone the repo, and build/run the app `NODE_ENV=production npm run generate && NODE_ENV=production npm run start`
2. Visit `http://localhost:3000/?foo=bar`
3. Observe in the console, that when logging out `context.query`, it is an empty object.

### What is Expected?

A universal app built via `nuxt generate` or `nuxt build` should behave the same when running in universal mode. The `query` object on the context should contain the query params for the current request.

The query object contains no params, as if it is holding onto the context for when the generate command was run.

danielroe:
## Types of changes

## Description
In full static mode, we remove any query parameters from the URL before parsing with router. But this isn't necessarily desirable if generating files in legacy universal mode. This PR lets users with legacy universal mode access req/res/query objects within the server build.

closes #7965

## Checklist:

### DIFF
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -80,9 +80,11 @@ export default async (ssrContext) => {
   <% } %>
 
   // Remove query from url is static target
-  if (process.static && ssrContext.url) {
+  <% if (isFullStatic) { %>
+  if (ssrContext.url) {
     ssrContext.url = ssrContext.url.split('?')[0]
   }
+  <% } %>
   // Public runtime config
   ssrContext.nuxt.config = ssrContext.runtimeConfig.public
   // Create the app definition and the instance (created for each request)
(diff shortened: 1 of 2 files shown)


atinux (approved):

pi0:
Thanks!

PR #8612, opened 2021-01-09
fix(vue-app): check for Page.options before looking within

### LINKED ISSUE #8234: scrollToTop throws `Page.options is undefined`
### Versions

- nuxt: v2.14.6
- node: v12.18.3

### Reproduction

Not sure how to reproduce.

### Issue

This line: `Pages.some(Page => Page.options.scrollToTop)` in router.scrollBehaviour sometimes throws an error: `Page.options is undefined`. This prevents the page from loading properly. Not sure how this can happen. Consol error is shown both in dev mode and prod mode, but only breaks the page in dev mode.

danielroe:
## Types of changes

## Description
Difficult to reproduce the issue but the fix is lightweight.

closes #8234

## Checklist:

### DIFF
--- a/packages/vue-app/template/router.scrollBehavior.js
+++ b/packages/vue-app/template/router.scrollBehavior.js
@@ -31,9 +31,9 @@ export default function (to, from, savedPosition) {
   // Scroll to the top of the page if...
   if (
       // One of the children set `scrollToTop`
-      (Pages.some(Page => Page.options.scrollToTop) ||
+      (Pages.some(Page => Page.options && Page.options.scrollToTop) ||
       // scrollToTop set in only page without children
-      (Pages.length < 2 && Pages.every(Page => Page.options.scrollToTop !== false))) &&
+      (Pages.length < 2 && Pages.every(Page => !Page.options || Page.options.scrollToTop !== false))) &&
       // route changes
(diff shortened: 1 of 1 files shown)


packages/vue-app/template/router.scrollBehavior.js
```diff
@@ -31,9 +31,9 @@ export default function (to, from, savedPosition) {
   // Scroll to the top of the page if...
   if (
       // One of the children set `scrollToTop`
-      (Pages.some(Page => Page.options.scrollToTop) ||
+      (Pages.some(Page => Page.options && Page.options.scrollToTop) ||
```
Atinux, line 34: Could we use this?
```suggestion
      (Pages.some(Page => Page.options?.scrollToTop) ||
```

clarkdo: This may add extra polyfill if user is not using optional chaining.

clarkdo (approved):

packages/vue-app/template/router.scrollBehavior.js
clarkdo: Let's merge this and I may make some refactoring to this file,

PR #8638, opened 2021-01-13
types: add `enablePreview` to context

TheAlexLichter:
Somehow the [`enablePreview`](https://nuxtjs.org/docs/2.x/features/live-preview/) function was missing in the Context type.
As it is only available in plugins, I'm not sure if there is a better way to add it. Open for feedback here!

### DIFF
--- a/packages/types/app/index.d.ts
+++ b/packages/types/app/index.d.ts
@@ -87,7 +87,7 @@ export interface Context {
   error(params: NuxtError): void
   nuxtState: NuxtState
   beforeNuxtRender(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
-  enablePreview(): Function | undefined
+  enablePreview?: (previewData?: Record<string, any>) => void
   $preview?: Object
 }
 
--- a/packages/types/app/index.d.ts
+++ b/packages/types/app/index.d.ts
@@ -87,7 +87,8 @@ export interface Context {
   error(params: NuxtError): void
   nuxtState: NuxtState
   beforeNuxtRender(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
-  enablePreview(): void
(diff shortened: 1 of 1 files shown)


Atinux:
Thanks @manniL 

It is a bit tricky indeed since it lives only for static target also.
We also inject `$preview`, see https://github.com/nuxt/nuxt.js/blob/9681a8937d0c973bc685b757faa686023a4536fc/packages/vue-app/template/index.js#L235-L240

manniL:
True! Should add that in the same PR as well.
Unfortunately, you can't mark functions as optional while objects can be.

packages/types/app/index.d.ts
```diff
@@ -87,6 +87,7 @@ export interface Context {
   error(params: NuxtError): void
   nuxtState: NuxtState
   beforeNuxtRender(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
+  enablePreview(): void
```
pi0, line 90: This might be `enablePreview: Function | undefined` ?

packages/types/app/index.d.ts
```diff
@@ -87,6 +87,8 @@ export interface Context {
   error(params: NuxtError): void
   nuxtState: NuxtState
   beforeNuxtRender(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
+  enablePreview(): Function | undefined
```
danielroe, line 90:
```suggestion
  enablePreview?: (previewData?: Record<string, any>) => void
```

packages/types/app/index.d.ts
```diff
@@ -87,6 +87,8 @@ export interface Context {
...
+  $preview?: Object
```
danielroe, line 91:
```suggestion
  $preview?: Record<string, any>
```

pi0 (approved):

pi0:
Thanks @manniL @danielroe!

PR #8640, opened 2021-01-13
chore: update webpack-dev-middleware to ^4.0.4

clarkdo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/webpack/src/builder.js
+++ b/packages/webpack/src/builder.js
@@ -185,18 +185,6 @@ export class WebpackBundler {
         })
     )
 
-    // disable webpack-dev-middleware stas printing
-    Object.defineProperty(
-      this.devMiddleware[name].context,
-      'stats',
-      {
-        configurable: true,
-        get () {
-          return 'none'
-        }
-      }
-    )
-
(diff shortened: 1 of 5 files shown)


packages/webpack/src/builder.js
```diff
@@ -185,6 +185,18 @@ export class WebpackBundler {
         })
     )
 
+    // disable webpack-dev-middleware stas printing
```
pi0, line 188: Shall we have a comment about this line linking to PR?

pi0 (dismissed):

packages/webpack/src/builder.js
clarkdo: I found we're not configuring `stats`, so I changed the fix. 
Does this make sense to you ?

clarkdo: BTW, should we remove [stats in webpackbar options ](https://github.com/nuxt/nuxt.js/blob/b4663d2bfb03237a429b7fa348770c831ed5e744/packages/webpack/src/config/base.js#L448)? I can't find any stats option usage in wepbackbar

PR #8644, opened 2021-01-14
feat(vue-app): add dark mode for the loadingIndicator

RubenVanEldik:
Add the option to use different colors for the loadingIndicator when the user has 'prefers-color-scheme: dark' enabled. (https://github.com/nuxt/nuxt.js/issues/8580)

## Types of changes

## Description
This update enables users to add two different color schemes for the loading indicator.

When the developer just uses strings as the for the `color`, `color2`, and `background`, nothing changes, but the developer can now also use an object with a `light` and `dark` key:

```js{}[nuxt.config.js]
export default {
  loadingIndicator: {
    name: 'circle',
    color: '#3B8070',
    background: {
      light: 'white',
      dark: 'darkgray'
    }
  }
}
```

The dark values are only used when a user has `prefers-color-scheme: dark` enabled.

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/nuxtjs.org/pull/1145)

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -238,13 +238,50 @@ export function getNuxtConfig (_options) {
       options.loadingIndicator = { name: options.loadingIndicator }
     }
 
+    const defaultColor = (options.loading && options.loading.color) || '#D3D3D3'
+    const defaultColor2 = '#F5F5F5'
+    const defaultBackground = (options.manifest && options.manifest.theme_color) || 'white'
+
+    if (isPureObject(options.loadingIndicator.color)) {
+      options.loadingIndicator = Object.assign(
+        options.loadingIndicator,
+        {
+          color: options.loadingIndicator.color.light || defaultColor,
+          colorDark: options.loadingIndicator.color.dark || defaultColor
+        }
+      )
+    }
+
+    if (isPureObject(options.loadingIndicator.color2)) {
+      options.loadingIndicator = Object.assign(
+        options.loadingIndicator,
+        {
+          color2: options.loadingIndicator.color2.light || defaultColor2,
+          color2Dark: options.loadingIndicator.color2.dark || defaultColor2
+        }
+      )
+    }
+
+    if (isPureObject(options.loadingIndicator.background)) {
+      options.loadingIndicator = Object.assign(
+        options.loadingIndicator,
+        {
+          background: options.loadingIndicator.background.light || defaultBackground,
+          backgroundDark: options.loadingIndicator.background.dark || defaultBackground
+        }
+      )
+    }
+
     // Apply defaults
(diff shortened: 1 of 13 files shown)


Atinux:
Thank you for the PR @RubenVanEldik 

I think we shall find a way to reduce the amount of code added to `packages/config/src/options.js` 🤔

RubenVanEldik:
Thank you for taking time to look at my PR! This is my first PR, but I have been working with Nuxt.js for about a year now and I absolutely adore it! 

I completely agree that the code can be shorter, but I wrote it this way to make it more maintainable. There are two ways to make it shorter:

1. Don't use objects for in `nuxt.config.js`, so instead of:
```js[nuxt.config.js]
{
  loadingIndicator: {
    color: {
      light: 'blue',
      dark: 'lightblue'
    },
    ...
  }
}
```
do this:
```js[nuxt.config.js]
{
  loadingIndicator: {
    color: 'blue',
    colorDark: 'lightblue',
    ...
  }
}
```
This makes the code behind the scenes a lot simpler, but makes the nuxt.config.js a bit uglier. 

2. Make the code I've written less DRY. This is the easiest solution, but I didn't do this originally since I thought it would make the code less readable/maintaible.


Since its your project I would love to hear your thoughts on what strategy I should use.

Atinux:
Thank you for the time answering and participating on this PR @RubenVanEldik 

Actually, I think this could lead to un-intended behaviour too (if the website does not have dark mode, it will flash from dark to light for example).

For your use case, could you consider using a custom `loadingIndicator` by specifying a path? See https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-loading-indicator#custom-indicators

RubenVanEldik:
I don't think that it will lead to unintended behaviour, because the dark variants will default to light variant. Therefore, if you don't explicitly set a dark color (for either `color`, `color2`, or `background`) that color will just default to the light variant, so you will see no difference.

RubenVanEldik:
I understand that this might not be the most urgent feature request you're dealing with, but I think that having a dark loading screen is more needed than one might think. More and more websites provide a dark version of their site and popular frameworks like Tailwind make it possible to add a dark mode with just a few lines.

It's no deal breaker if this isn't possible (like you said, you can always add a manual page), but the loadingIndicator is such a cool and easy way to add a nice loading indicator to your site that (I think) it would be a shame if there is no dark-mode available. 

I'll take a look tomorrow to see if I can clean up to code a bit further, but I if you think that this feature will clutter up the Nuxt.js repo I would totally understand it. :)

pi0:
Closing due to above reasons and should be straight forward to override spa loading template :)

PR #8675, opened 2021-01-21
perf: global intersection observer for prefetching

### LINKED ISSUE #8672: [Performance] Change smart prefetch strategy to use global IntersectionObserver
Right now, every `<NuxtLink>` instantiates a new `IntersectionObserver` to enable smart prefetching for the linked page (unless disabled via `no-prefetch` or through the config).

https://github.com/nuxt/nuxt.js/blob/3bba4c6a7fd086aaa434e0abf92bdc9480fd3f4e/packages/vue-app/template/components/nuxt-link.client.js#L18

While this works well with a small to medium number of links, performance degrades on larger numbers (e.g. think of a page with a long table).

Sources: [[1]](https://www.bennadel.com/blog/3954-intersectionobserver-api-performance-many-vs-shared-in-angular-11-0-5.htm#main-content), [[2] (related to Mutation Observer + in german though but same principle)](https://twitter.com/sir_pepe/status/1348539744599363584)

Therefore I suggest to instantiate a global `IntersectionObserver`, e.g. accessible via `window.$nuxt.prefetchObserver` and observer all `NuxtLink`s through this single instance. This should improve performance, especially on sites with a huge number of links.

TheAlexLichter:
Still in draft. Feedback welcome!

Resolves #8672 one day :D

Open Questions:

## "General purpose" Observer (similar to [nuxt-image](https://github.com/nuxt/image/blob/dev/src/runtime/utils/observer.ts)) or dedicated to smart prefetching only?

Pros of a general purpose one:

* can be used by module authors (e.g. nuxt image) + developers to avoid multiple IntersectionObservers

Cons:

* Harder to draft out API - more scenarios to consider
* Has to be documented properly then 😜

## Tests?

* Now the smart prefetching capabilities cannot be unit-tested (I think) and a new fixture must be included.

### DIFF
--- a/packages/vue-app/template/components/nuxt-link.client.js
+++ b/packages/vue-app/template/components/nuxt-link.client.js
@@ -15,15 +15,6 @@ const cancelIdleCallback = window.cancelIdleCallback || function (id) {
   clearTimeout(id)
 }
 
-const observer = window.IntersectionObserver && new window.IntersectionObserver((entries) => {
-  entries.forEach(({ intersectionRatio, target: link }) => {
-    if (intersectionRatio <= 0 || !link.__prefetch) {
-      return
-    }
-    link.__prefetch()
-  })
-})
-
 <%= isTest ? '// @vue/component' : '' %>
 export default {
   name: 'NuxtLink',
@@ -42,29 +33,36 @@ export default {
       default: '<%= router.linkPrefetchedClass %>'
     }<% } %>
   },
+  data () {
+    return {
+      __observed: false,
+      __unobserve: () => {}
+    }
+  },
   mounted () {
     if (this.prefetch && !this.noPrefetch) {
-      this.handleId = requestIdleCallback(this.observe, { timeout: 2e3 })
+
+      this.handleId = requestIdleCallback(this.observe, { timeout: 2000 })
     }
   },
   beforeDestroy () {
     cancelIdleCallback(this.handleId)
 
     if (this.__observed) {
-      observer.unobserve(this.$el)
+      this.__unobserve(this.$el)
       delete this.$el.__prefetch
     }
   },
   methods: {
     observe () {
       // If no IntersectionObserver, avoid prefetching
-      if (!observer) {
+      if (!window.IntersectionObserver) {
         return
       }
       // Add to observer
       if (this.shouldPrefetch()) {
         this.$el.__prefetch = this.prefetchLink.bind(this)
-        observer.observe(this.$el)
+        this.__unobserve = this.$nuxt.context.observer.use(this.$el)
         this.__observed = true
       }<% if (router.linkPrefetchedClass) { %> else {
         this.addPrefetchedClass()
@@ -95,7 +93,7 @@ export default {
         return
       }
       // Stop observing this link (in case of internet connection changes)
-      observer.unobserve(this.$el)
+      this.__unobserve(this.$el)
       const Components = this.getPrefetchComponents()
       <% if (router.linkPrefetchedClass) { %>const promises = []<% } %>
 
--- a/packages/vue-app/template/components/nuxt-link.client.js
+++ b/packages/vue-app/template/components/nuxt-link.client.js
@@ -49,7 +49,7 @@ export default {
     cancelIdleCallback(this.handleId)
 
     if (this.__observed) {
-      this.__unobserve(this.$el)
+      this.$options.__unobserve(this.$el)
       delete this.$el.__prefetch
     }
   },
@@ -62,7 +62,7 @@ export default {
       // Add to observer
       if (this.shouldPrefetch()) {
         this.$el.__prefetch = this.prefetchLink.bind(this)
-        this.__unobserve = this.$nuxt.context.observer.use(this.$el)
+        this.$options.__unobserve = this.$nuxt.context.observer.use(this.$el)
         this.__observed = true
       }<% if (router.linkPrefetchedClass) { %> else {
         this.addPrefetchedClass()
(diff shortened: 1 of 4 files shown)


packages/vue-app/template/observer.js
```diff
@@ -0,0 +1,49 @@
...
+    delete el[OBSERVER_ID_KEY]
+    intersectObserver.unobserve(el)
+  }
+
+  const add = (el) => {
+    el[OBSERVER_ID_KEY] = el[OBSERVER_ID_KEY] || ++observerIdCtr
+    intersectObserver.observe(el)
+  }
+
+  observerInstance = { supported: true, add, remove }
```
pi0, line 38: When observer is unsupported, we already return false (ie, non falst value implies supported flag)

packages/vue-app/template/observer.js
```diff
@@ -0,0 +1,49 @@
...
+  return observerInstance
+}
+
+export function use (el) {
```
pi0, line 42: I would suggest using `useObserver` to readability when imported by name

packages/vue-app/template/observer.js
```diff
@@ -0,0 +1,49 @@
...
+  const observer = get()
+  observer.add(el)
+  if (!observer) {
+    return () => {}
```
pi0, line 46: We may return a falsy value or in case of having callback for useObserver, immediately calling otherwise util consumer (nuxt-link) needs to again check `window.IntersectionObserver`

packages/vue-app/template/utils.js
```diff
@@ -240,6 +241,11 @@ export async function setContext (app, context) {
     }
     if (process.client) {
       app.context.nuxtState = window.<%= globals.context %>
+      // Global IntersectionObserver
+      app.context.observer = {
```
pi0, line 245: If keeping instance in context, why not using factory function like `createObserver`? (this way we can also make util side-effect free)

packages/vue-app/template/observer.js
```diff
@@ -0,0 +1,49 @@
...
+    return false
+  }
+
+  if (observerInstance) {
+    return observerInstance
+  }
+
+  const onIntersect = (entries) => {
+    entries.forEach(({ intersectionRatio, target: link }) => {
+      if (intersectionRatio <= 0 || !link.__prefetch) {
```
pi0, line 18: Target is not necessarily link if util is refactored

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -42,29 +33,36 @@ export default {
       default: '<%= router.linkPrefetchedClass %>'
     }<% } %>
   },
+  data () {
+    return {
+      __observed: false,
+      __unobserve: () => {}
```
pi0, line 39: - Having `this.__unobserve` can imply it is observed
- We don't need to keep function as reactive data

packages/vue-app/template/observer.js
manniL: If we only export a factory function, I'd keep name as is (so we always call `ctx.observer.use` for example) :)

packages/vue-app/template/observer.js
manniL: WDYT - should we keep the observer more general or specifically for prefetching only?

packages/vue-app/template/components/nuxt-link.client.js
manniL: Moved to `$options` so it is not reactive anymore ☺️

packages/vue-app/template/observer.js
manniL: I agree, there should be a nicer solution for it so the check only happens once.

manniL: (updated)

packages/vue-app/template/components/nuxt-link.client.js
manniL: Or should I move it to the element itself? 🤔

packages/vue-app/template/observer.js
pi0: Yes in case of factory function makes sense

packages/vue-app/template/components/nuxt-link.client.js
pi0: Unless there is a reason, i think instance should be fine :D

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -42,29 +33,37 @@ export default {
...
     cancelIdleCallback(this.handleId)
 
-    if (this.__observed) {
-      observer.unobserve(this.$el)
-      delete this.$el.__prefetch
+    if (!this.__observed) {
+      return
     }
+    this.$options.__unobserve(this.$el)
+    delete this.$el.__prefetch
```
pi0, line 54: Should we also unsed `this.__observed` (mostly to ensure flag not leaking)

packages/vue-app/template/components/nuxt-link.client.js
```diff
@@ -42,29 +33,37 @@ export default {
...
-        return
-      }
       // Add to observer
       if (this.shouldPrefetch()) {
+        const unobserveFunctionOrUnavailable = this.$nuxt.context.observer.use(this.$el)
+        if(!unobserveFunctionOrUnavailable) {
+          return
+        }
+
+        this.$options.__unobserve = unobserveFunctionOrUnavailable
```
pi0, line 65: If it is falsy, can we simply assign like `this.$options.__unobserve = this.$nuxt.context.observer.use(this.$el)`? (it is a `false` reference but anyway for older browsers. I think simplicity worth over it)

packages/vue-app/template/observer.js
pi0: We are exposing it to global context namespace so i think makes sense (to ensure forward api compatibility) :D

Atinux:
It starts to look pretty good, thank you for your current work on it @manniL 😊

manniL:
Closing here as #8672 is closed + focusing on Nuxt 3 development.

PR #8683, opened 2021-01-23
fix(cli): disable parallel build behaviour when generating

### LINKED ISSUE #7390: How to fix Nuxt Warning :Starting with Nuxt version 3 this will be a fatal error
### Version

### Reproduction link

[https://github.com/fpl1104/nuxttest.git](https://github.com/fpl1104/nuxttest.git)

### Steps to reproduce

npx create-nuxt-app ..
...
...
...
Edit nuxt.config.js =>buid:{        cache: true,parallel: true,}

run generate => OK
run generate again =>Nuxt Warning

 
   │                                                                                      
   │   ⚠ Nuxt Warning                                                                     │
   │                                                                                      │
   │   The command 'nuxt generate' finished but did not exit after 5s                     │
   │   This is most likely not caused by a bug in Nuxt.js                                 │
   │   Make sure to cleanup all timers and listeners you or your plugins/modules start.   │
   │   Nuxt.js will now force exit                                                        │
   │                                                                                      │
   │   DeprecationWarning: Starting with Nuxt version 3 this will be a fatal error        │
   │                                                                                      │

### What is expected ?

generate ok & no warning

I get a warning 

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10680">#c10680</a>)</em></sub></div>

danielroe:
## Types of changes

## Description
* prevents leftover timers and listeners after build

closes #7390

## Checklist:

### DIFF
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -81,6 +81,7 @@ export default {
     // Set flag to keep the prerendering behaviour
     config._legacyGenerate = true
     if (config.build) {
+      // https://github.com/nuxt/nuxt.js/issues/7390
       config.build.parallel = false
     }
 
--- a/packages/cli/src/commands/generate.js
+++ b/packages/cli/src/commands/generate.js
@@ -80,6 +80,9 @@ export default {
 
     // Set flag to keep the prerendering behaviour
     config._legacyGenerate = true
+    if (config.build) {
+      config.build.parallel = false
+    }
 
     const nuxt = await cmd.getNuxt(config)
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/generate.js
```diff
@@ -80,6 +80,9 @@ export default {
 
     // Set flag to keep the prerendering behaviour
     config._legacyGenerate = true
+    if (config.build) {
+      config.build.parallel = false
```
pi0, line 84: - It would be nice if we can investigate more what is root cause and possibly report/fix in upstream
- It seems a webpack bug (?) but only happening for generate?
- Linking related issue to workaround at least would be nice :)

danielroe: Issue only happens when generate includes build step before, like also #7855. Skipping build in full static solves issue.

pi0: But only one instance of nuxt initializes webpack right?

pi0: Update: Added comments for fast merge still think worth checking rootCause instead of patch in longer term :)

PR #8738, opened 2021-02-01
fix: fix encoding issues with payload paths

### LINKED ISSUE #8679: encoded url causing error when accessing via static mode
### Versions

- nuxt: 2.14.12
- node: 15.6.0

### Reproduction
https://github.com/c53hzn/c53hzn.github.io/tree/master/_nuxt/static/1611255759/blog/language
This is how the generated paths look like
image
And the encoded format will cause error when relevant pages try to extract data from the payload.

<details open>
<br>

</details>

### Steps to reproduce
if I go to a page with Chinese characters in path, the payloads generated won't work because the paths generated for payloads with encoded characters need to be encoded again before the payloads can be accessed
eg:
If I want to access the payloads generated for this path: 
c53hzn.github.io/blog/language/中文

The payload is stored in this path:
c53hzn.github.io/_nuxt/static/1611321880/blog/language/%E4%B8%AD%E6%96%87/payload.js
Here "%E4%B8%AD%E6%96%87" is the encoded Chinese character "中文"

But in order to access this payload, the actual URL should be
c53hzn.github.io/_nuxt/static/1611321880/blog/language/%25E4%25B8%25AD%25E6%2596%2587/payload.js

### What is Expected?
Static mode should generate payload path as is, rather than generate encoded paths.

Encoded paths cause errors when fetching data.

I think it's the same issue as this one [https://github.com/nuxt/nuxt.js/issues/8680](https://github.com/nuxt/nuxt.js/issues/8680), and I followed @pi0 's advice and updated Nuxt, but the issue remains.

danielroe:
## Types of changes

## Description
closes #8679

- fix(generator): decode full static filesystem paths
- fix(vue-app): check decoded path against manifest
- fix(vue-app): prevent double encoding for urls

## Checklist:


packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -180,7 +180,11 @@ export default class SSRRenderer extends BaseRenderer {
     if (renderContext.staticAssetsBase) {
       const preloadScripts = []
       renderContext.staticAssets = []
-      const { staticAssetsBase, url, nuxt, staticAssets } = renderContext
+      const { staticAssetsBase, nuxt, staticAssets } = renderContext
+      let url = new $URL(renderContext.url).pathname
```
pi0, line 184: We can instead use lighter `parseURL().pathname` utility (also if prop is of path type, variable may be named pathname as well instead of url)

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -180,7 +180,11 @@ export default class SSRRenderer extends BaseRenderer {
...
+      if (!this.options.router.trailingSlash) {
+        url = withoutTrailingSlash(url) || '/'
```
pi0, line 186: Note for later: we might handle empty string situation with ufo later since empty string is not valid path...

pi0 (commented):
Nice! I would suggest we also decode url in generator when computing filesystem paths to ensure this does not implicitly happens again

packages/generator/src/generator.js
```diff
@@ -161,14 +162,22 @@ export default class Generator {
     return requireModule(path.join(this.options.buildDir, 'routes.json'))
   }
 
+  normalizeRoute (url) {
+    let pathname = decode(parseURL(url).pathname)
+    if (!this.options.router || !this.options.router.trailingSlash) {
```
pi0, line 167: We was already decoding on L171. Any reason trainling slash handling is related to this fix? (also `route` is already of type path not full url). I was refering to [this](https://github.com/nuxt/nuxt.js/blob/fa4e5a9e1a01225b86ddda845706aa51ca277992/packages/generator/src/generator.js#L355) for nomalizing asset path and ensure is not encoded for fs

packages/vue-renderer/src/renderers/ssr.js
```diff
@@ -180,7 +180,13 @@ export default class SSRRenderer extends BaseRenderer {
     if (renderContext.staticAssetsBase) {
       const preloadScripts = []
       renderContext.staticAssets = []
-      const { staticAssetsBase, url, nuxt, staticAssets } = renderContext
+      const { staticAssetsBase, nuxt, staticAssets } = renderContext
+
+      let pathname = decode(parseURL(renderContext.url).pathname)
```
pi0, line 185: I think also here `ctx.url` is of kind path and decode should be enough? :D

pi0:
I've updated fix implementation to generator since rest of url generations, are by encoding (browsers also anyway will encode request for http headers)

Also discovered a second issue (client side) that manifest check is by decoded paths so we was skipping payload fetch for unicode routes...

PR #8759, opened 2021-02-03
fix(vue-app): enable dynamic publicPath on server-side

### LINKED ISSUE #8662: dynamic publicPath not working with img tags and assets within css (unless extractCSS: true)
As stated in #8520, dynamically setting publicPath to use a cdn url doesn't work for img tags and assets referenced in css.
While i can make assets referenced in css work using `extractCSS: true` and setting publicPath to the absolute cdn path (i.e. without the cdn hostname and protocol), i'm unable to make img tags work referencing image assets.

### Versions

- nuxt: nuxt-edge@^2.15.0-26848868.eae3acb7
- node: 14

### Reproduction

Link to codesandbox:

https://codesandbox.io/s/inspiring-river-xp8xb?file=/pages/index.vue

As you can see there, the images referenced via img tag as well as css use the build-time publicPath,
even though the referenced javascript files use the correct dynamic cdn url publicPath.

### Steps to reproduce

E.g. specify `build.publicPath` to equal `process.env.CDN_URL` and add an `img` tag to a component referencing e.g. `~/assets/file.jpg`.
Run `CDN_URL=/_nuxt/ yarn build` and then `CDN_URL=https://cdn.host/path/to/assets/ yarn start`

### What is Expected?

The `img` tag should be referenced via `https://cdn.host/path/to/assets/file.jpg` 

The `img` tag will still be referenced via `/_nuxt/file.jpg`

danielroe:
## Types of changes

## Description
closes #8662

## Checklist:
- [x] All new and existing tests are passing. - **note** - not sure why tests failed earlier


test/dev/basic.dynamic.test.js
```diff
@@ -62,6 +62,14 @@ describe('basic ssr', () => {
     expect(fetchCount).toBeGreaterThan(0)
   })
 
+  test('dynamic publicPath is used in ssr html', async () => {
+    const { html } = await nuxt.server.renderRoute('/', {
+      resources: resourceLoader
+    })
+
+    expect(html).toContain('<img src="https://cdn.nuxtjs.org/img')
```
danielroe, line 70: currently producing `https:/cdn.nuxtjs.org/img`

danielroe: fyi: issue is `urlJoin` - which doesn't handle full urls - would have caused bugs elsewhere that were being normalised by browser

packages/vue-app/template/server.js
```diff
@@ -51,8 +51,8 @@ Vue.component(NuxtLink.name, NuxtLink)
 
 const noopApp = () => new Vue({ render: h => h('div', { domProps: { id: '<%= globals.id %>' } }) })
 
-function urlJoin () {
-  return Array.prototype.slice.call(arguments).join('/').replace(/\/+/g, '/')
+function urlJoin (base, ...args) {
+  return joinURL(base, ...args)
```
pi0, line 55: Is this an alias? :D Then maybe we can directly rexport it to prevent refactoring in other usages?

danielroe: I will prepare separate PR with more utils from `ufo`

pi0 (approved):

test/dev/basic.dynamic.test.js
hippiechick60: Thank you

hippiechick60 (commented):
25c927bd62f010b6e8355127a9a8e3b70f8a176e

PR #8765, opened 2021-02-04
refactor(utils, vue-app): use `ufo` to parse queries and join urls

danielroe:
## Types of changes
- [x] Refactor

## Description
As we're already importing and using `ufo`, we can eliminate some of our homegrown URL tools. This PR (so far):
* replaces `urlJoin()` functions (duplicates) with the `ufo` equivalent
* replaces the extensive URL + query normalisation function (and so obsoletes the `clientUseUrl` feature flag from #6287)
* replaces a custom trailingSlash implementation

## Checklist:


packages/vue-app/template/utils.js
```diff
@@ -683,20 +603,12 @@ export function addLifecycleHook(vm, hook, fn) {
...
+export const urlJoin = joinURL
 
-export function stripTrailingSlash (path) {
-  return path.replace(/\/+$/, '') || '/'
-}
+export const stripTrailingSlash = withoutTrailingSlash
 
 export function isSamePath (p1, p2) {
-  return stripTrailingSlash(p1) === stripTrailingSlash(p2)
+  return withoutTrailingSlash(p1) === withoutTrailingSlash(p2)
```
pi0, line 611: We may implement this in ufo :p

danielroe: Great. I think perhaps keep this PR pending until those changes are released in `ufo`?

packages/vue-app/template/server.js
```diff
@@ -58,8 +57,8 @@ const createNext = ssrContext => (opts) => {
     ssrContext.nuxt.serverRendered = false
     return
   }
-  opts.query = stringify(opts.query)
-  opts.path = opts.path + (opts.query ? '?' + opts.query : '')
+  opts.path = withQuery(opts.path, opts.query)
+  opts.query = stringifyQuery(opts.query || {})
```
pi0, line 61: Checking `createNext ` it seems we no longer would need stringifing `opts.query` (not used until res.end)

danielroe, line 61:
```suggestion
```

packages/vue-app/template/server.js
```diff
@@ -1,6 +1,5 @@
-import { stringify } from 'querystring'
 import Vue from 'vue'
-import { normalizeURL, joinURL } from 'ufo'
+import { joinURL, normalizeURL, stringifyQuery, withQuery } from 'ufo'
```
danielroe, line 2:
```suggestion
import { joinURL, normalizeURL, withQuery } from 'ufo'
```

packages/vue-app/template/server.js
danielroe: That is very satisfying

packages/vue-app/template/server.js
```diff
@@ -57,19 +57,19 @@ const createNext = ssrContext => (opts) => {
     ssrContext.nuxt.serverRendered = false
     return
   }
-  opts.path = withQuery(opts.path, opts.query)
+  let fullPath = withQuery(opts.path, opts.query)
```
danielroe, line 60: good call

pi0 (approved):

PR #8780, opened 2021-02-06
fix(utils): `serializeFunction` fails with certain functions

### LINKED ISSUE #8779: Fails to serialize certain functions using serializeFunction in the template
### Versions

- nuxt: v2.14.12
- node: v14

### Reproduction

https://codesandbox.io/s/serializefunction-bug-3elqy?file=/pages/index.vue

### Steps to reproduce

From a template, call
```ejs
<%
const obj = {
  fn () { const x = rule => rule }
};
const serialized = serializeFunction(obj.fn)
%>
<%= serialized %>
```

### What is Expected?

That the serialized function starts with a `function` keyword.

The serialized function starts with a `fn` identifier.

rchl:
A function shorthand with certain statements inside failed to
serialize properly using serializeFunction.

Resolves #8779

## Types of changes

## Description
The issue was due to the replacement patterns not considering there being
a space between the function name and the parenthesis before the
arguments.

Changed the simple string pattern to regexp to be able to match any
number of whitespace characters between those.

Also changed the test file to not be transformed by babel-jest as that
has affected the input passed to "serializeFunction" and made some tests
invalid. For example, the new test didn't even fail with babel-jest as
it was transforming the test input in a way that made it pass.

## Checklist:

### DIFF
--- a/packages/utils/test/serialize.test.js
+++ b/packages/utils/test/serialize.test.js
@@ -3,6 +3,8 @@ const { serializeFunction, normalizeFunctions } = require('../src/serialize')
 // NOTE: This test file is on purpose using CommonJS syntax to avoid code being
 // transformed by 'babel-jest' which affects the results.
 
+const RE_LINE_BREAKS = /[\r\n]+/
+
 describe('util: serialize', () => {
   test('should normalize arrow functions', () => {
     const obj = {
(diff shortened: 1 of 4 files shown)


packages/utils/test/serialize.test.js
```diff
@@ -1,4 +1,9 @@
-import { serializeFunction, normalizeFunctions } from '../src/serialize'
+const { serializeFunction, normalizeFunctions } = require('../src/serialize')
+
+// NOTE: This test file is on purpose using CommonJS syntax to avoid code being
```
pi0, line 3: Shall we move implementation of function to be serialized into fixtures or a sibling file without `.test`?

rchl: Wouldn't using a fixture would make testing significantly slower and a bit more complicated.

As for separating the test functions, I would think that babel-jest would also transform those on importing. I think, haven't tested.

Any specific concerns about the current approach?

pi0: Actually it is fine per-se but it has a side effect that test file should be pure js and inconsistent with others (imagine if we want to move it to nuxt3 repo where tests are typescript...). By fixture i meant only a single js to import and exclude from babel transform

rchl: I'll invesitage that.

rchl: That has worked.

One note that `normalizeFunctions` used by `serializeFunction` mutates the passed function so I had to copy imported test objects to avoid earlier tests affecting later tests.

I wonder if that's expected.

pi0 (approved):
Thanks!

packages/utils/test/serialize.test.js
pi0: @rchl Seems good :)

PR #8785, opened 2021-02-07
ci: use build artifacts and run dev tests in band

danielroe:


.github/workflows/test.yml
```diff
@@ -191,11 +191,11 @@ jobs:
       - name: build
         run: yarn test:fixtures -i
 
-      - name: cache fixtures
-        uses: actions/cache@v2
+      - name: Save fixtures
+        uses: actions/upload-artifact@v2
```
pi0, line 195: Good idea 👍

danielroe: :blush:

.github/workflows/test.yml
```diff
@@ -225,10 +225,9 @@ jobs:
           key: ${{ matrix.os }}-node-v${{ matrix.node }}-nuxt-${{ github.sha }}
 
       - name: restore fixtures
-        uses: actions/cache@v2
+        uses: actions/download-artifact@v2
         with:
-          path: test/fixtures
-          key: ${{ matrix.os }}-node-v${{ matrix.node }}-fixtures-${{ github.sha }}
+          name: fixtures-${{ matrix.os }}-node-v${{ matrix.node }}
```
pi0, line 230: I think we are missing path here :)

danielroe: yes :-)

danielroe: but it does confirm the other bug - I think likely cache wasn't ready to be downloaded and technically there is no contract that it will be in GHA

pi0: Indeed and cache download issue was frequently happening before too... I hope it works good now :D

package.json
```diff
@@ -25,7 +25,7 @@
     "nuxt": "jiti ./packages/cli/bin/nuxt-cli.js",
     "pkg": "jiti ./scripts/pkg.js",
     "test": "yarn test:fixtures && yarn test:dev && yarn test:unit && test:types",
-    "test:dev": "jest test/dev --forceExit",
+    "test:dev": "jest test/dev --forceExit --runInBand",
```
pi0, line 28: I wish we didn't need this (and cause to me is at least unknown) but lets have fix since is blocking all PRs

pi0 (approved):
Awesome work @danielroe <3

package.json
danielroe: agreed - I wonder if somehow there is a conflict between two tests depending on which runs first - but will have to investigate further

PR #8775, opened 2021-02-05
fix(vue-renderer): fix csp policy when using hash + custom `script-src` policy.

josepjaume:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Nuxt's CSP support currently generates a hash for inline scripts. Unfortunately, if you specify a custom `script-src` policy, it will override the hash - so they don't work together.

This change makes sure both approaches are compatible by adding the generated tags to any custom script-src policy.

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests are passing.

PR #8790, opened 2021-02-08
fix(vue-renderer): fix csp policy when using hash + custom `script-src` policy

pi0:
Follow up #8775 by @josepjaume

> Nuxt's CSP support currently generates a hash for inline scripts. Unfortunately, if you specify a custom script-src policy, it will override the hash - so they don't work together.
> This change makes sure both approaches are compatible by adding the generated tags to any custom script-src policy


danielroe:
@josepjaume Could you explain more about what you mean by `Unfortunately, if you specify a custom script-src policy, it will override the hash - so they don't work together.`?

This should be the case already unless you are specifying `unsafe-inline` in which case you can override it by setting `unsafeInlineCompatibility: true`.

danielroe:
I think we can close this PR.

PR #8792, opened 2021-02-08
feat(cli): warn if incompatible node/package versions detected

### LINKED ISSUE #8783: incompatible versions warning
We have this feature for `@nuxt/vue-app` dependencies if user-installed version differs from the one with current nuxt version. Now that nuxt2 cannot have upgrades like webpack5 ([sass-loader](https://github.com/nuxt/nuxt.js/issues/8771)), we should create a warning system for well known deps:
- Node.js (to use 12+)
- sass-loader <= 10
- [generic implementation to add more]

Linked: #8782

danielroe:
## Types of changes

## Description
closes #8783

## Checklist:
- [x] All new and existing tests are passing


packages/cli/src/utils/dependencies.js
```diff
@@ -0,0 +1,49 @@
+import consola from 'consola'
+import { satisfies } from 'semver'
+
+import { getPKG } from '@nuxt/utils'
+
+export function checkDependencies() {
+  const packagesToTrack = ['postcss', 'webpack']
+  const webpackDeps = getPKG('@nuxt/webpack').dependencies
```
pi0, line 8: An inline list would be more than enough since exceptions are manual anyway (and the fact that some deps are optional like sass-loader)

packages/cli/src/commands/generate.js
```diff
@@ -61,6 +62,7 @@ export default {
     }
   },
   async run (cmd) {
+    checkDependencies()
```
pi0, line 65: We shall run check in setup.js instead of commands

packages/cli/src/utils/dependencies.js
```diff
@@ -0,0 +1,49 @@
...
+        consola.warn(`Expected ${name}@${requiredVersion} to be installed.`)
+      }
+    }
+  })
+
+  // Check Node version
+  if (satisfies(process.version, '^10.0.0')) {
+    consola.warn('Node.js 10 is deprecated and will shortly not be supported.')
+  } else if (!satisfies(process.version, '>12.0.0')) {
+    consola.error('Please use a supported version of Node.js (12+).')
```
pi0, line 47: IMO thin can be still a warning for You are using an incompatible NodeJS version. An error is not necessarily happening by using node 10

pi0: Also no exception will be needed anymore for 10.x (version constraints are always updating)

packages/cli/src/utils/dependencies.js
pi0: I would use an object syntax like `{ 'sass-loader':  { range: '^10.1.1' }, 'node': { range: '>12', message: 'you are using unsupported NodeJS version' }`

danielroe: Node.js version checking has to be done differently so I think we don't need a single object.

packages/cli/src/utils/dependencies.js
danielroe: I've moved the <10 to a warn (as 10 is). Not sure what you mean by last comment about 10.x.

pi0: I mean when `>12.0.0` constraint did not met warning can be generic like _You are using an unsupported version of Node.js (x.y.z) and nuxt is likely to be broken. It is recommended to use latest LTS version (https://nodejs.org)_

note that node@10 is only supported by engine. With 2.15 release notes we are also asking to use 12+. And this might be changed in the future too depending on what happens to nuxt dependencies. So there instead of exception for 10.x, we can have a generic advice and single semver constraint.

packages/cli/src/utils/dependencies.js
pi0: Actually it was idea of #8783 for generic version checking system. And we might make smarter warnings by checking if package is explicitly installed by `{rootDir}/package.json` or hoisted. But seems okay for now we might do it later on as a nuxt-contrib project :)

packages/cli/src/utils/dependencies.js
danielroe: Very helpful - yes, much better approach. I've made this more generic but happy to refactor further either now or in a further PR.

packages/cli/src/utils/dependencies.js
```diff
@@ -0,0 +1,50 @@
...
+      webpack: { range: '^4.46.0' }
+    },
+    optionalDependencies: {
+      'sass-loader': { range: '^10.1.1' }
+    },
+    engines: {
+      node: {
+        range: '>12.0.0',
+        message:
+          'You are using an unsupported version of Node.js (x.y.z) and Nuxt is likely to be broken. It is recommended to use the latest LTS version (https://nodejs.org)'
```
pi0, line 21: 👀 👀  🗡️

packages/cli/src/utils/dependencies.js
```diff
@@ -0,0 +1,50 @@
...
+        range: '^7.0.32'
+      },
+      webpack: { range: '^4.46.0' }
+    },
+    optionalDependencies: {
+      'sass-loader': { range: '^10.1.1' }
+    },
+    engines: {
+      node: {
+        range: '>12.0.0',
```
danielroe, line 19:
```suggestion
        range: '>=12.0.0',
```

packages/cli/src/utils/dependencies.js
```diff
@@ -0,0 +1,50 @@
...
+      webpack: { range: '^4.46.0' }
+    },
+    optionalDependencies: {
+      'sass-loader': { range: '^10.1.1' }
+    },
+    engines: {
+      node: {
+        range: '>=12.0.0',
+        message:
+          'You are using an unsupported version of Node.js (x.y.z) and Nuxt is likely to be broken. It is recommended to use the latest LTS version (https://nodejs.org)'
```
danielroe, line 21: Add link to LTS release schedule.
```suggestion
          'You are using an unsupported version of Node.js (x.y.z) and Nuxt is likely to be broken. It is recommended to use the latest LTS version (https://nodejs.org/en/about/releases/)'
```

packages/cli/src/utils/dependencies.js
pi0: @danielroe Still missing something 😆 (x.y.z). I need to do some refactors before merge will fix it too

pi0: Didn't noticed replace in bottom :)

packages/cli/test/unit/dependencies.test.js
```diff
@@ -0,0 +1,25 @@
...
+  afterEach(() => jest.resetAllMocks())
+
+  test('checkDependencies', () => {
+    checkDependencies()
+    expect(consola.warn).toHaveBeenCalledWith(
+      expect.stringMatching(
+        /webpack@.+ is installed but .+ is expected/
+      )
+    )
+    expect(consola.warn).toHaveBeenCalledTimes(1)
```
danielroe, line 18: I think this can be dropped @pi0

PR #8794, opened 2021-02-08
fix: ignore trailing slash in static payloads manifest (#8794)

danielroe:
## Types of changes

## Checklist:


packages/generator/src/generator.js
```diff
@@ -349,7 +350,7 @@ export default class Generator {
         }
         // Add route to manifest (only if no error and redirect)
         if (this.manifest && (!res.error && !res.redirected)) {
-          this.manifest.routes.push(route)
+          this.manifest.routes.push(withoutTrailingSlash(route))
```
pi0, line 353: I think we are missing `|| '/'` fallback here. It should be finally fixed back in ufo :D

packages/vue-app/template/App.js
```diff
@@ -328,14 +324,14 @@ export default {
       <% if (nuxtOptions.generate.manifest) { %>
       const manifest = await this.fetchStaticManifest()
       const path = this.getRoutePath(route)
-      if (!manifest.routes.includes(decodeURI(path))) {
+      if (!manifest.routes.includes(decode(path))) {
```
pi0, line 327: What if we normalize route once in line before? (decode and strip trailing-slash). This way we ensure consistency and little bit perf for not decoding multiple times...

danielroe: In fact it looks like there was a bug in payload fetching with routerBase that this fixes :see_no_evil:

packages/generator/src/generator.js
pi0: Added in 0.6.3

pi0 (approved):

PR #8802, opened 2021-02-09
fix(vue-app): add leading slash for payload hydration

danielroe:
## Types of changes

## Description
Fixes regression reported https://www.github.com/nuxt/nuxt.js/issues/8107#issuecomment-776227825 and https://www.github.com/nuxt/nuxt.js/issues/8107#issuecomment-776222703.

## Checklist:

### DIFF
--- a/packages/vue-app/template/App.js
+++ b/packages/vue-app/template/App.js
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-import { decode, parsePath, withLeadingSlash, withoutTrailingSlash, normalizeURL } from 'ufo'
+import { decode, parsePath, withLeadingSlash, withoutBase, withoutTrailingSlash, normalizeURL } from 'ufo'
 <% utilsImports = [
   ...(features.asyncData || features.fetch) ? [
     'getMatchedComponentsInstances',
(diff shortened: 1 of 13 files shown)


packages/vue-app/template/App.js
```diff
@@ -304,7 +304,7 @@ export default {
       if (base && route.startsWith(base)) {
         route = route.substr(base.length)
       }
-      return withoutTrailingSlash(parsePath(route).pathname)
+      return withLeadingSlash(withoutTrailingSlash(parsePath(route).pathname))
```
pi0, line 307: I think somewhere else we are wrongly stripping leading slash from `route`. This fixes issue not the root cause

danielroe: That's a good point. Will track it down

danielroe: I think it's the `route.substr` call in the previous line - it subtracts the default router base which is `/` so we lose the leading slash. Not sure why this wasn't causing us issues before.

danielroe: So I think this PR is good to go.

pi0: Yep :) We may use substr(base.lengh, -1) since router base is normalized with trailing slash in nuxt options.

... Adding `withBase` `withoutBase` utils to ufo

PR #8809, opened 2021-02-10
fix(vue-app): align client fetch key with server

danielroe:
## Types of changes

## Description
fixes a mismatch between server/client fetch keys:
https://github.com/nuxt/nuxt.js/blob/92018e586b766b6e2728925c9a324d9863bd5e77/packages/vue-app/template/mixins/fetch.server.js#L57

## Checklist:


packages/vue-app/template/mixins/fetch.client.js
```diff
@@ -72,7 +72,7 @@ function createdFullStatic() {
     this._fetchKey = this.$options.fetchKey.call(this, getCounter)
   } else {
     const key = 'string' === typeof this.$options.fetchKey ? this.$options.fetchKey : defaultKey
-    this._fetchKey = key + ':' + getCounter(key)
+    this._fetchKey = key + getCounter(key)
```
pi0, line 75: Wouldn't better if adding `:` to server generated keys? key=[number] can potentially allow _fetchKey conflicts

PR #8835, opened 2021-02-16
fix(types): `build.html.minify` can be false

harlan-zw:
## Types of changes

Types

## Description

The `build.html.minify` option isn't typed correctly.

 It's passed to:
-  the `html-webpack-plugin` which allows for a boolean option OR an object, see: https://github.com/jantimon/html-webpack-plugin#minification
- generator::minifyHtml which checks if it's not false before minifying

Because it's also used in the generate code, if it was typed to be able to use `'auto'` or `true` it may cause some issues, so we only type the false. Guessing this discrepancy will be solved in Nuxt3.

## Checklist:
- [ ] I have added tests to cover my changes (if not applicable, please state why) - no logic changes
- [] All new and existing tests are passing. - tests are throwing a lot of weird unrelated errors, maybe environment-dependent


danielroe (approved):

pi0:
Thanks!

Issue #8839, opened 2021-02-16
Error: Expected identifier, string or number (Nuxt 12.15.0)


Sendoo:
### Versions

- nuxt: v2.15.0
- node: v14.15.4
- OS : Windows 10 20H2
- browser: Edge 18

### Reproduction

`Expected identifier, string or number`

After upgrading Nuxt to version 2.15.0 in Edge 18 and IE 11, I got an error like this which blocks the application 

![изображение](https://user-images.githubusercontent.com/46229254/108105570-762c8800-7095-11eb-9974-e7c17cb1a3f2.png)

danielroe:
@Sendoo Would you try adding `'ufo'` to `build.transpile`?

Sendoo:
> 
> 
> @Sendoo Would you try adding `'ufo'` to `build.transpile`?

It helped, thanks.
There is no more error.
Will this be fixed in future releases? 🙂

radum:
This is just a temporary fix, I am getting other errors after fixing UFO. The promise polyfill is not added anymore.

PR #8846, opened 2021-02-17
fix(webpack): transpile ufo

### LINKED ISSUE #8838: Error 'SCRIPT1053' occurs on IE11 (nuxt: v2.15.0)
### Versions

- nuxt: v2.15.0
- node: v14.15.4
- IE11 (Windows 10 Pro 20H2)

### Reproduction

This is a minimal example works with no error on IE11.
https://github.com/voidproc/example-nuxt-ie

However, an error 'SCRIPT1053: Const must be initialized' occurs when remove `transpile: ['ufo']` setting in `nuxt.config.js`.

kn3ny:
Add `ufo` module, instead of `@nuxt/ufo`, to the default target of transpile.

## Types of changes

## Description

There is a commit https://github.com/nuxt/nuxt.js/commit/7529d65d75a1f943817cd5683e21335d48891b6b that migrates from `@nuxt/ufo` to `ufo`. However, it seems migration of the configuration of transpiling was not included.
Since `ufo` will not be transpiled by default, user's nuxt.js app will be broken on old browsers such as IE.

Resolves: #8838, #8839

## Checklist:

### DIFF
--- a/packages/webpack/src/config/base.js
+++ b/packages/webpack/src/config/base.js
@@ -73,7 +73,7 @@ export default class WebpackBaseConfig {
     return [
       /\.vue\.js/i, // include SFCs in node_modules
       /consola\/src/,
-      /@nuxt[/\\]ufo/, // exports modern syntax for browser field
+      /ufo/, // exports modern syntax for browser field
       ...this.normalizeTranspile({ pathNormalize: true })
     ]
   }


packages/webpack/src/config/base.js
```diff
@@ -73,7 +73,7 @@ export default class WebpackBaseConfig {
     return [
       /\.vue\.js/i, // include SFCs in node_modules
       /consola\/src/,
-      /@nuxt[/\\]ufo/, // exports modern syntax for browser field
+      /ufo/, // exports modern syntax for browser field
```
kn3ny, line 76: Should I need to change `/^ufo$/`?

pi0: ufo should be fine. All dependant packages except tailwind seem migrated already

pi0 (approved):
Thanks!

packages/webpack/src/config/base.js
kn3ny: I see. Thank you!

PR #8832, opened 2021-02-15
fix(builder): use path.resolve for pages dir

atinux:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)

## Description

This allow to give a `pages/` directory outside the `srcDir`. It should not break anything.

## Checklist:

- [ ] All new and existing tests are passing.

PR #8855, opened 2021-02-19
v2.15.1

### LINKED ISSUE #8820: Deep Merge of Runtime Configs
Currently when Nuxt merges the private and public configs, if a key is public, it's totally replaced by the private value if it exists.  This is fine for simple key/val pairs, but for larger apps where you have more complex configs, this leads to needing redundant info inside the config.

```
 publicRuntimeConfig:{
   algolia:{
     appId: 'AAAAAAA',
     apiKey:'PUBLIC-KEY',
  }
},
privateRuntimeConfig:{
  algolia:{
    apiKey: 'SUPER-SECRET-KEY',
  },
}
```

On server side, $config.algolia just contains the secret API key, not the appId.  So in order to have both, I have to resort to duplicating the appId in 2 places which isn't ideal.

It would be nice for Runtime configs to either

a) detect if something is an object and merge lower keys OR
b) have a flag somewhere to tell Nuxt to allow deep merges

### LINKED ISSUE #8838: Error 'SCRIPT1053' occurs on IE11 (nuxt: v2.15.0)

pi0:
### 🐛 Bug Fixes

- `renderer`
  - Deep merge runtime config for server (resolves #8820)
- `webpack`
  - #8846 Transpile `ufo` (resolves #8838)
- `cli`
  - #8841 Don't early return when a dependency isn't installed
- `types`
  - #8835 `build.html.minify` can be false
- `builder`
  - #8832 Use `path.resolve` for pages dir

### 💖 Thanks to

- Harlan Wilton (@loonpwn)
- Kenny (@kn3ny)


packages/vue-app/template/server.js
```diff
@@ -103,7 +103,7 @@ export default async (ssrContext) => {
     __webpack_public_path__ = joinURL(ssrContext.nuxt.config.app.cdnURL, ssrContext.nuxt.config.app.assetsPath)
   }
   // Create the app definition and the instance (created for each request)
-  const { app, router<%= (store ? ', store' : '') %> } = await createApp(ssrContext, { ...ssrContext.runtimeConfig.public, ...ssrContext.runtimeConfig.private })
+const { app, router<%= (store ? ', store' : '') %> } = await createApp(ssrContext, ssrContext.runtimeConfig.private)
```
zoellner, line 106: @pi0 was this an intentional change? it looks like the server went from receiving the public and private runtimeConfig to just receiving the private one.

pi0: https://github.com/nuxt/nuxt.js/pull/8855/files#diff-895d4548da1de2d3970ec23e8b55f69e181a9f28296b1c96504b4230ebfa4336R300

They get merged here

pi0: If nothing as intended would you please open an issue with reproduction?

PR #8953, opened 2021-03-08
fix(types): export interface for `NuxtConfig` instead of type

danielroe:
## Types of changes

## Description
* Seems accidentally changed in https://github.com/nuxt/nuxt.js/pull/8630
* Context: https://github.com/nuxt-community/module-template/pull/65#issuecomment-705654456, https://github.com/nuxt/nuxt.js/pull/8169
* allows it to be extended rather than overwritten

## Checklist:


clarkdo (approved):

t10-13rocket (approved):
Approval for test.

PR #8955, opened 2021-03-08
feat(generator): add ignoreEnv generate option during ensureBuild(cmd)

wlarch:
## Types of changes

## Description
Added a build option `ignoreEnv` (boolean) to ignore `env` and `process.env`  fields within the ensureBuild(cmd) quick diff function of the generate.js utils.

We are using Nuxt.js to generate multiple static websites with incremental builds. The same Nuxt project (code base) is used for different Websites. The `nuxt generate` command receives `themeOptions` parameters in our build server thus, changing the environment every time the application needs to achieve a partial build. When the environment changed (env or process.env) webpack needs to rebuild the entire project, this increases `yarn generate` time by +30 seconds (compared to 3 seconds when webpack build is not needed). In our scenario, a webpack rebuild is needed when the code base changes, or Nuxt version, or when `--force-build` is specified.

Example config : 

```js
build: {
    ignoreEnv: true,
},
```

Another reason this change is needed is because there was no other way to modify the static `cacheDir` prior to `nuxt ready` hook. As we edit nuxt.config.js using a custom module, no appropriate hook could be used to edit the `'static: { cacheDir }` option.

There are no open issues regarding this PR, but it could be related to the one related to single route generation in which I participated in the discussion : 
https://github.com/nuxt/nuxt.js/issues/6138

(More information to come regarding our incremental generation with Nuxt.js).

## Checklist:
- [ X] I have updated the documentation accordingly.
https://github.com/nuxt/nuxtjs.org/pull/1323
- [X] I have added tests to cover my changes (if not applicable, please state why)


clarkdo (commented):
Nitpicking: I think this may be more readble:

```js
    const fields = ['nuxtVersion', 'ssr', 'target']

    if (nuxt.options.build.ignoreEnv !== true) {
      fields.push('env', 'process.env')
    }
```

clarkdo (commented):
Let's give it a defualt value `false` and update types defination as well.

wlarch:
Thanks for your review @clarkdo. Type definition and default value have been added. Any follow-up on this ?

clarkdo:
Looks good to me, just waiting for other members review. Cc @nuxt/framework

test/fixtures/full-static-with-ignore-env/full-static-with-ignore-env.test.js
```diff
@@ -0,0 +1,3 @@
+import { buildFixture } from '../../utils/build'
```
danielroe, line 1: I might be missing something, but is this used to test the feature?

wlarch: To be fully honest I wasn't quite sure if this was the right way to test the feature using the suggested `ignoreEnv` option in `nuxt.config.js`.

papac25 (approved):

wlarch:
@pi0 We are wondering when can this fix be merged in a minor version ? It will be critical to an imminent release.

pi0 (approved):

pi0:
Hi @wlarch sorry for the delayed merge. You can have it (in few minutes) on `nuxt-edge`. Since its a new feature will land in 2.16 (next month releasee)

wlarch:
This is wonderful, thanks ! :sun_with_face:

PR #8978, opened 2021-03-12
fix(vue-app): reload page once after loading chunk error

### LINKED ISSUE #3389: [edge] Deployment of new version breaking site for live users. ( Loading chunk {n} failed error)

rinu:
## Types of changes

## Description
Resolves: #3389
Changes made in #3940 should maybe be deleted.
There's probably a nicer way of doing this, all comments are welcome.

## Checklist:


clarkdo (dismissed):

packages/vue-app/template/utils.js
```diff
@@ -146,7 +146,19 @@ export function resolveRouteComponents (route, fn) {
...
       if (typeof Component === 'function' && !Component.options) {
-        Component = await Component()
+        try {
+          Component = await Component()
+        } catch (error) {
+          // Handle webpack chunk loading errors
+          // This may be due to a new deployment or a network problem
+          if (error && error.name === 'ChunkLoadError' && window.location.hash !== '#retry') {
+            // mark the page not to reload infinitely
+            window.location.hash = '#retry'
```
pi0, line 156: Modification of URL, might interference with application logic. We may use [navigation type](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation/type) guard. (safari workaround needed)

pi0, line 156: Modification of URL, might interference with application logic. We may use [navigation type](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation/type) guard. (safari workaround needed)

pi0 (changes requested):

packages/vue-app/template/utils.js
rinu: There's no clean way to solve this, can't think of anything at least. Navigation type seems unreliable. We could rename this retry to #__nuxt_retry or something. Or try to use sessionStorage for this?

We could not have this check at all but there's the unlikely possibility of a currupt build where a critical file is actually missing from the file system. Infinite reloads will not help when debugging that.

pi0: Do you have any reference regarding the unreliability of nav type? But of course we can detect incompatible UA and switch to localStorage (or sessionStorage) method.   Modification of `#` overrides one if set in real application.

rinu: As you pointed out navigation type is not supported by Safari so I can't rely on it. I'm working on a sessionStorage method.

pi0 (approved):

pi0:
Thanks :)

PR #8980, opened 2021-03-12
feat(config): support for skipping base tag when router.base is set

### LINKED ISSUE #7573: Hash paths in sub-pages routes to `/`
### Version

### Reproduction link

[https://deploy-preview-948--postwoman.netlify.app/settings](https://deploy-preview-948--postwoman.netlify.app/settings)

### Steps to reproduce

1. Goto https://deploy-preview-948--postwoman.netlify.app/settings
2. Click on any secondary navigation links from sidebar (last 4 icons)

3. `<a href="#account"><i class="material-icons">person</i></a>` should route to `/sub_page#account` but instead it transpiled to  `/#account` resulting opening new link.

### What is expected ?

Hash paths inside all pages should be relative to current path not `/` absolute path.

 `<a href="#account"><i class="material-icons">person</i></a>` should route to `/sub_page#account` but instead it transpiled to  `/#account` resulting opening new link.

### Additional comments?

This was introduced in Nuxt v2.13.0.

I guess it might be erlated using `base` path in `nuxt.config.js`

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10822">#c10822</a>)</em></sub></div>

internetztube:
## Types of changes

Resolves: https://github.com/nuxt/nuxt.js/issues/7573
There's probably a nicer way of doing this, all comments are welcome.

https://stackoverflow.com/questions/659120/url-hash-with-html-base-tag

## Checklist:

Example:
```
export default {
  router: {
    base: '/app/',
    baseTag: false,
  }
}
```

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -31,7 +31,11 @@ export function getNuxtConfig (_options) {
     options.router.middleware = [options.router.middleware]
   }
 
-  if (options.router && typeof options.router.base === 'string') {
+  if (typeof options.router.baseTag === 'undefined') {
+    options.router.baseTag = true
+  }
+
+  if (options.router && typeof options.router.base === 'string' && options.router.baseTag) {
     options._routerBaseSpecified = true
   }
 
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -31,7 +31,7 @@ export function getNuxtConfig (_options) {
     options.router.middleware = [options.router.middleware]
   }
 
-  if (typeof options.router.baseTag === 'undefined') {
+  if (options.router && typeof options.router.baseTag === 'undefined') {
     options.router.baseTag = true
   }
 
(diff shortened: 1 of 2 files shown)


danielroe:
@internetztube You shouldn't use `<a>` if your intent is just to trigger an action but remain on the same route. Instead, use a button.

internetztube:
@danielroe Unfortunately I have no control over what the link looks like. It is in the shadow root of an external web component.

danielroe:
@internetztube I think this is an unusual enough case that I would consider solving this with a Nuxt hook rather than adding an configuration option.

You should be able to excise the base within 'render:route' - see https://nuxtjs.org/docs/2.x/internals-glossary/internals-renderer.

What do you think?

clarkdo (commented):
Agree with @danielroe , I think you can use `render:route` hook to remove base tag in ssr.

**NOTE: the code is not tested which may need adjustment**

```js
// nuxt.config.js
export default {
  hooks: {
    render: {
      route(url, result) {   
        result.html = result.html.replace(/<base [^>]*>/, '')
      }
    }
  }
}
```

internetztube:
This has indeed worked. Thank you very much! @clarkdo

PR #9000, opened 2021-03-17
fix/load dotenv

chbaefront:
## Types of changes

## Description
in nuxt.config.js
When accessing the env property as process.env.variables, you were looking at dotenv unconditionally.

Depending on the value of NODE_ENV I wanted to switch dotenv. This is because nuxt.config.js can get values ​​for the desired environment from process.

## Checklist:

### DIFF
--- a/packages/config/src/load.js
+++ b/packages/config/src/load.js
@@ -35,7 +35,7 @@ export async function loadNuxtConfig ({
 
   // Load env
   envConfig = {
-    dotenv: '.env',
+    dotenv: (process.env.NODE_ENV.trim() === 'development') ? '.env.development' : (process.env.NODE_ENV.trim() === 'production') ? '.env.production' : '.env',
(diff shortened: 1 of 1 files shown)


pi0:
Hi, @katanazero86 thanks for PR. But at least with the current version, this is intentional behavior. Please see [here](https://www.npmjs.com/package/dotenv#should-i-have-multiple-env-files) for why.

You can specify alternative dotenv file when starting with `--dotenv` argument. And can also use it in package.json:

```json
"scripts": {
   "dev": "nuxt dev --dotenv .env.dev
}
```

katanazero86:
@pi0 

Hi.
Thanks for the quick feedback, and thanks for sharing the link to the article!

I am glad that I am learning one thing like this.

Watch out for COVID-19 and hope you have a happy day.

Thank you!

PR #9008, opened 2021-03-18
fix(vue-app): fallback to global nuxt instance of `$root` is not available

### LINKED ISSUE #8995: custom components in vuetify-dialog not working
### Versions

- nuxt: v2.15.3
- node: v12.20.2

### Steps to reproduce

Use vuetify-dialog plugin with custom components 

### What is Expected?

components should work in dialogs as they do everywhere else

$nuxt is not accessible because vuetify-dialog uses a seperate Vue instance. Because of this for example fetch is not working. 

### Proposed fix

I fixed it by modifying the $nuxt property in index.js:

```
  get() {
    return this.$root.$options.$nuxt || window.$nuxt
  },
  configurable: true
})
```

vcordes79:
## Types of changes
return window.$nuxt if $root is not set

## Description
$nuxt was not always accessible (if $root was not set), so return the property from window object

fixes #8995

## Checklist:

### DIFF
--- a/packages/vue-app/template/index.js
+++ b/packages/vue-app/template/index.js
@@ -47,7 +47,7 @@ Vue.component(Nuxt.name, Nuxt)
 Object.defineProperty(Vue.prototype, '<%= globals.nuxt %>', {
   get() {
     const globalNuxt = this.$root.$options.<%= globals.nuxt %>
-    if(process.client && !globalNuxt && typeof window !== 'undefined') {
+    if (process.client && !globalNuxt && typeof window !== 'undefined') {
       return window.<%= globals.nuxt %>
     }
     return globalNuxt
--- a/packages/vue-app/template/index.js
+++ b/packages/vue-app/template/index.js
@@ -46,7 +46,7 @@ Vue.component(Nuxt.name, Nuxt)
 
 Object.defineProperty(Vue.prototype, '<%= globals.nuxt %>', {
   get() {
-    return this.$root.$options.<%= globals.nuxt %> || window.<%= globals.nuxt %>
+    return this.$root.$options.<%= globals.nuxt %> || (process.client && typeof window !== 'undefined' && window.<%= globals.nuxt %>)
   },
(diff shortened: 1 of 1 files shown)


vcordes79:
fixes #8995

packages/vue-app/template/index.js
```diff
@@ -46,7 +46,7 @@ Vue.component(Nuxt.name, Nuxt)
 
 Object.defineProperty(Vue.prototype, '<%= globals.nuxt %>', {
   get() {
-    return this.$root.$options.<%= globals.nuxt %>
+    return this.$root.$options.<%= globals.nuxt %> || (process.client && typeof window !== 'undefined' && window.<%= globals.nuxt %>)
```
danielroe, line 49: I'm worried this will lead to inconsistent behaviour between client/server rendering.

pi0: Indeed we have to double-check this. Check-related issue #8995 is due to usage of the second app by vuetify.

vcordes79: I don't see any inconsistant behaviour between client/server. 

In the server it will be `this.$root.$options.<%= globals.nuxt %>` if it's defined. If it's not defined, it will still be undefined, because window is not defined server side. 

On the client it won't change anything if `$root` exists - only if it does not. Could you please explain, what the prolem is?

clarkdo: It looks OK, but as @pi0 mentioned, let double check, a test for this would be test.

BTW, I think we can split this complex expression into if/else, sth like:

```js
const globalNuxt = this.$root.$options.<%= globals.nuxt %>

if(process.client && !globalNuxt && window) {
  return window.<%= globals.nuxt %>
}

return globalNuxt
```

PayelKarmakar06:
I recently tried upgrading to 2.15.4, but I am still facing $root not available issue.

`Cannot read property '$options' of undefined`

PR #9014, opened 2021-03-18
fix(config): compile type compatibility in css-loader v4

EvanMaFYH:
set the css-loader's compileType option to icss

## Types of changes

## Description
when nuxt upgrade the css-loader library to v4, it results in a break change
if you export your sass variables and import it in your vue file before, it works fine
however when the css-loader library is upgraded to v4, the css-loader's compileType option is module by default which causes the problem
for example:
in your variable.scss file
```
$primaryColor: blue;
:export{
  primaryColor: $primaryColor;
}
```
in your vue file
```
import variables from 'variable.scss'
```

before: we can get the expected value
now : the variables we get is '{}'

## Checklist:


clarkdo (dismissed):

clarkdo:
Can you please update test snapshot ?

EvanMaFYH:
> Can you please update test snapshot ?

I have updated it

pi0 (approved):

PR #9034, opened 2021-03-23
fix(server): avoid caching `.js` assets in development to fix HMR for safari

### LINKED ISSUE #3828: safari: infinite loop when hot reloading

danielroe:
## Types of changes

## Description
This PR turns off caching for webpack-processed JS files in dev mode.

<details><summary><strong>Fixes reload loop on safari (closes #3828)</strong>.</summary>
<br>
Previously, `runtime.js` was being cached in-memory by Safari and triggering infinite HMR updates in the following situation:
* server started
* change made to file (triggering HMR)
* server restarted
* safari hard reload -> starts infinite loop (because `runtime.js` from first server is loaded from memory)

</details>

**Note**: if you have previously experienced an infinite reload loop, you will need to clear your Safari cache once using `Develop...Empty Caches` and then reload.

## Checklist:

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -114,6 +114,11 @@ export default class Server {
         if (!this.devMiddleware) {
           return next()
         }
+        // Safari over-caches JS (breaking HMR) and the seemingly only way to turn
+        // this off in dev mode is to set Vary: * header
+        if (req.url.startsWith(this.publicPath) && req.url.endsWith('.js')) {
+          res.setHeader('Vary', '*')
+        }
         this.devMiddleware(req, res, next)
       })
 
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -116,6 +116,7 @@ export default class Server {
         }
         // Safari over-caches JS (breaking HMR) and the seemingly only way to turn
         // this off in dev mode is to set Vary: * header
+        // #3828, #9034
         if (req.url.startsWith(this.publicPath) && req.url.endsWith('.js')) {
           res.setHeader('Vary', '*')
         }


clarkdo (dismissed):

packages/server/src/server.js
```diff
@@ -114,6 +114,11 @@ export default class Server {
         if (!this.devMiddleware) {
           return next()
         }
+        // Safari over-caches JS (breaking HMR) and the seemingly only way to turn
+        // this off in dev mode is to set Vary: * header
+        if (req.url.startsWith(this.publicPath) && req.url.endsWith('.js')) {
+          res.setHeader('Vary', '*')
```
clarkdo, line 120: Would cache-control: no-store be better?

danielroe: You would think so, but it's a known issue with Safari's cache mechanism (= a bug, imo). See https://stackoverflow.com/a/54337073/9614848

pi0: @danielroe Can you please also confirm it only happens for js? (ie changing an imported asset like image accepts hmr without cache issue)

danielroe: @pi0 I've confirmed there is no issue with HMR image assets.

The issue isn't to do with all .js, just the runtime chunk. I believe the issue could be resolved by simply adding `Vary` header to `/_nuxt/runtime.js`, as that file is what determines the HMR update URLs. But I've made it more general in this PR in case of `runtimeChunk: 'multiple'`, for example.

pi0 (approved):

PR #9050, opened 2021-03-25
fix(vue-app): use `app.context.route` for resolving components

### LINKED ISSUE #9049: (vue-app) - route navigation guards are not processed on client side when calling `createApp`
When the client side script starts, the navigation guards are skipped, preventing the route to reflect any change made through these guards. Adding to that, when resolving components later in the `client.js` script, the context route is not passed to the `resolveComponents` method. Like in the `server.js` previous issue #9009 a new route resolution is created without using the navigation guards that could have been added.

This complete the #9009 issue. Sorry I missed that earlier.

In my specific case, which can be tested with the reproduction link below, I create two routes with the same path but different components. We have specific rules for our website's urls. I use a route navigation guard to test the url parameters to decide if the url belongs to one page or the other. When the url should belong to the other page I use the `next` method with the new route parameters like documented here : https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards.

When `createApp` is called client side, navigation guards are skipped: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/index.js#L267

When the client gets the matched component for the given url. The component resolution using `router.match(path)` is not picking up the changes made by the navigation guards resulting in the wrong components to be matched and rendered. See https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/client.js#L235

We get a DOM mismatch error when the client renders the page. (`[Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside <p>, or missing <tbody>. Bailing hydration and performing full client-side render.`)

### Versions

- nuxt: v2.15.3

### Reproduction

Codesandbox: https://codesandbox.io/s/interesting-perlman-4v1b9

Direct test urls:
- https://4v1b9.sse.codesandbox.io/category
- https://4v1b9.sse.codesandbox.io/content

### Steps to reproduce
1. Navigate to https://4v1b9.sse.codesandbox.io/category with the web console open
```
asyncData category
```

4. Navigate to https://4v1b9.sse.codesandbox.io/content with the web console open
```
asyncData category
```

### What is Expected?
When navigating to https://4v1b9.sse.codesandbox.io/content the 3 `console.log` messages should be saying:
```
asyncData content
```

The client side is matching the `CategoryPage` component instead of the `ContentPage` page component because it uses a new route resolver instead of using the already resolved route that takes into account any navigation guard redirection that could have happened

enwin:
I'll need your feedback on this fix since I'm calling the async component resolution used for the SSR. It might not be the best idea but it works so far 🤷

## Types of changes

## Description

Resolves: #9049 
Update `vue-app/template/index.js` to call async component resolver on client side too.
Update `vue-app/template/client.js` to use `app.context.route` in the `resolveComponents` method to match the components based on the final route.

## Checklist:


enwin:
If someone could have a look at the tests suite?

On my side, most of the time I get a timeout from puppeteer on each e2e tests: `TimeoutError: waiting for function failed: timeout 30000ms exceeded`. I reran `yarn && yarn build` restarted multiple times but still cannot debug whats failing.

pi0 (approved):
LGTM. Thanks!

patrioticcow:
Hi.

Why was the if statement removed `if (process.server && ssrContext && ssrContext.url)` ?

enwin:
Hi @patrioticcow!

The if statement was moved further down so when `createApp` is called from the client side, the component resolution is made using the potential navigation guards that you can have added to your site using `router.beforeEach`. Without that, you could end up with the wrong page component being executed.

From what I'm seing from your code example, the issue resides in the fact that the component resolution is triggered using `router.push` and that in your code you override the push method so it can do a `window.location.href` instead. By doing that every time the page loads `router.push` is called creating a navigation to the same page again and again with `window.location.href`.

The best of both world would be for nuxt to use `router.replace` instead of `router.push`. It shouldn't have any incidence on the component resolution. I also noticed that it would avoid the creation of a new history entry in the browser. That alone should be a nice fix.

Will work on that ASAP 👍

PR #9058, opened 2021-03-27
fix(types): use extensible options for `ModuleThis`

### LINKED ISSUE #8985: [types] Options accessed from module not extendable
The `Module` type has an `options` property whose type is `NuxtOptions`:

https://github.com/nuxt/nuxt.js/blob/f08f4efcde7bc174fd9b68a2a5a785d504659391/packages/types/config/module.d.ts#L24

Since it's `NuxtOptions` and not `NuxtConfig`, if a module extends `NuxtConfig` with an extra property, that won't be accessible type-wise from the module. For example:

```ts
declare module '@nuxt/types' {
  interface NuxtConfig {
    i18n?: Options
  }
}
```

```js
/** @type {import('@nuxt/types').Module} */
export default function () {
  const moduleOptions = this.options.i18n  // `this.options.i18n` has type `any` here
}
```

### Versions

- nuxt: v2.15.2

danielroe:
## Types of changes

## Description
closes #8985

It's also more appropriate as `this.options` within a module isn't fully resolved config.

## Checklist:

### DIFF
--- a/packages/types/config/module.d.ts
+++ b/packages/types/config/module.d.ts
@@ -6,7 +6,7 @@
 
 import { Configuration as WebpackConfiguration } from 'webpack'
 import { NuxtOptionsLoaders } from './build'
-import { NuxtOptions } from '.'
+import { NuxtConfig } from '.'
 
 interface ExtendFunctionContext {
   isClient: boolean
@@ -21,7 +21,7 @@ type ExtendFunction = (config: WebpackConfiguration, ctx: ExtendFunctionContext)
 
 interface ModuleThis {
   extendBuild(fn: ExtendFunction): void
-  options: NuxtOptions
+  options: NuxtConfig
   nuxt: any // TBD
   [key: string]: any // TBD
 }


packages/types/config/module.d.ts
```diff
@@ -21,7 +21,7 @@ type ExtendFunction = (config: WebpackConfiguration, ctx: ExtendFunctionContext)
 
 interface ModuleThis {
   extendBuild(fn: ExtendFunction): void
-  options: NuxtOptions
+  options: NuxtConfig
```
pi0, line 24: Modules receive resolved/normalized options which is `NuxtOptions`  (`Configuration`). If it is desired, the module author should directly extend `NuxtOptions` interface with normalized types or extending `Configuration` interface.

danielroe: TIL. For some reason I had it in my head that this.nuxt.options was the normalised config. There's no point in this PR then - and in fact if following module template module authors should be fine.

rchl: That recommendation raises some questions in my head...

I understand the difference between unresolved and resolved options for **built-in Nuxt options**, but from the module's perspective, the types for a given module are never normalized/resolved because whatever user sets in the given module's options never gets updated to resolved/normalized values. Or rather they might, but only internally in the module, and the Nuxt config setting is never updated itself (at least not in the modules I know or maintain).

So from the module's perspective, it seems like the recommendation should be to always extend `NuxtOptions` rather than `NuxtConfig`. Is that true? That will make the types work both in MouduleThis and the nuxt.config.js itself.

Then I wonder what's the point of `NuxtConfig` interface... In which case will it make sense for a module to extend `NuxtConfig` rather than `NuxtOptions`? 

Also, if you are considering extending `Configuration` interface a viable option at all, then it shouldn't be deprecated probably. But then, I don't quite see the use case for doing that in the first place.

rchl: > For some reason I had it in my head that this.nuxt.options was the normalised config.

Doesn't really change my points but this statement is confusing. `this.nuxt.options` **is normalized**. Normalized built-in Nuxt options at least.

danielroe: @rchl I think the point is that *both* are normalised, rather than just `this.nuxt.options`, which was my mistake.

pi0: from module perspective, this.nuxt.options.i18n is directly user input without applied defaults or normalization. We should only extend configuration (nuxtConfig) interface. I think confusion part is module type we exposed.

rchl: (note that I prefer that we talk about `this.options` rather than `this.nuxt.options` since `this.options` has well-defined type while `this.nuxt` is currently `any`)

> from module perspective, this.nuxt.options.i18n is directly user input without applied defaults or normalization.

Exactly. So `this.options.i18n` and `i18n` in `nuxt.config.js` should both have the same type. I can only achieve that currently if I extend `NuxtOptions` rather than `NuxtConfig`.

> We should only extend configuration (nuxtConfig) interface. I think confusion part is module type we exposed.

I don't think there is any confusion here. Rather either a bug in types or somehow conflicting recommendation that says that we should extend `NuxtConfig` always.

rchl: > I can only achieve that currently if I extend `NuxtOptions` rather than `NuxtConfig`.

Or extend both `NuxtOptions` and `NuxtConfig`, of course. But what would be the point of that if extending `NuxtOptions` alone have the same effect?

danielroe: @rchl Because the `NuxtConfig` interface extends `NuxtOptions`, if you have augmented `NuxtOptions` you shouldn't need to augment `NuxtConfig` as well.

rchl: I know. At this point I'm only arguing with "We should only extend configuration (nuxtConfig) interface".

If we all agree that extending `NuxtOptions` is the way to go for modules so that the same types are applied to (for example) `this.options.i18n` and `i18n` in `nuxt.config.js` then I have no more questions.

pi0: @rchl You are right. This is an issue with module typing. Extending nuxtConfig interface for i18n is correct. BTW this.options and this.nuxt.options are same references (if they don't have same type thats also an issue in typings)

pi0: (let's continue thread in issue)

PR #9077, opened 2021-03-30
fix(vue-app): refresh data in development when navigating

### LINKED ISSUE #8501: AsyncData isn't merged when triggered by watchQuery after HMR in Dev
### Versions

- nuxt: locally 2.14.6 / codestandbox-2.14.12
- node: locally 10.21.0 / codesandbox-?

### Reproduction
https://codesandbox.io/s/asyncdatawatchqueryhmr-problem-8q3cl

### Steps to reproduce
Fork the sandbox, press the button to update the query string - works great

Trigger a hot module reload, press the button - notice the console log shows asyncData firing but the data on the screen no longer is updated.  If you were to inspect it using Vue Devtools you'll see the state is the old state instead of the new one returned from asyncData.
**It may take more than 1 HMR update to trigger the problem**

### What is Expected?
Data returned from asyncData should be merged with data() and the page state should be updated.

updates from asyncData when triggered by watchQuery after a HMR are not applied.

### NOTE
I couldn't get this to break in a production build, so it appears to be isolated to HMR.

danielroe:
## Types of changes

## Description
This makes the requirement in development to refresh data more lax, which is designed to address the issue that after HMR, we lose the match between component & instance.

There is likely a better way of doing this - suggestions welcome.

closes #8501

## Checklist:


clarkdo (commented):
I feel like we can re-register intances in HMR, but I may need more investigation and debug.

pi0:
@danielroe @clarkdo Would you please give an update on this? Thanks :)

clarkdo:
I may have a look this week

PR #9086, opened 2021-03-31
fix(types): use options from `sass-loader` rather than `sass`

### LINKED ISSUE #9085: [types] Type '{ sassOptions: {}; }' is not assignable to type 'Options'.
While upgrading to the latest version of Nuxt, i've encountered a typings problem while customizing the loader options for SASS/SCSS. 

```
TS2322: Type '{ sassOptions: { includePaths: string[]; }; }' is not assignable to type 'Options'.
  Object literal may only specify known properties, and 'sassOptions' does not exist in type 'Options'.
     8 |     loaders: {
     9 |       scss: {
  > 10 |         sassOptions: {
       |         ^^^^^^^^^^^^^^
  > 11 |           // include assets path as absolute path for scss imports
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 12 |           includePaths: [path.resolve(__dirname, "./src/assets")],
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 13 |         },
       | ^^^^^^^^^^
    14 |       },
    15 |     },
    16 |   },
```

SassOptions is referencing [@types/sass](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sass/index.d.ts) directly and since `build.scss.sassOptions` is a property added by [sass-loader](https://github.com/webpack-contrib/sass-loader/blob/master/README.md#options), it doesn't exist in the interface and thus gives an error.

https://github.com/nuxt/nuxt.js/blob/558227f628cc5453be4cec16bf15dcb397baacae/packages/types/config/build.d.ts#L11
https://github.com/nuxt/nuxt.js/blob/558227f628cc5453be4cec16bf15dcb397baacae/packages/types/config/build.d.ts#L69-L70

There seems to be a change between `@nuxt/types@2.14.12` and `@nuxt/types@2.15.0` which caused the issue to plop up for me, since downgrading to `2.14.12` removed the error message for me.

It should be an easy fix though, since there are typings for [`sass-loader`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/sass-loader) which has the correct typings for the loader and the required modules included.

### Versions

- nuxt: v2.15.3
- node: v14.15.4

### Reproduction

https://github.com/marcopixel/nuxt-sassoptions-repro

### Steps to reproduce

1. Clone repo
2. `npm install`
3. `npm run dev`

### What is Expected?

The example should run without any errors caused by TS and should not show any errors in VSCode.

The example builds & runs fine, but shows the error while issue checking in the console & inside VSCode.

danielroe:
## Types of changes

## Description
* pins dependency in case of breaking change (-> webpack 5) in future

closes #9085

## Checklist:


clarkdo (approved):

packages/types/package.json
```diff
@@ -22,7 +22,7 @@
     "@types/node": "^12.20.7",
     "@types/optimize-css-assets-webpack-plugin": "^5.0.3",
     "@types/pug": "^2.0.4",
-    "@types/sass": "^1.16.0",
+    "@types/sass-loader": "8.0.1",
```
clarkdo, line 25: We can also put the package into renovate ingoreDeps for not getting auto upgradation pr

danielroe: Currently [`@types/webpack@^4` is dependency](https://unpkg.com/browse/@types/sass-loader@8.0.1/package.json) and as long as that remains we should be fine with upgrades but we'll need to examine lockfile before approving them.

PR #9078, opened 2021-03-30
types: inline `@types/terser-webpack-plugin` 

clarkdo:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Due to latset @types/terser-webpack-plugin change, webpack@5 will be added to our depednecy tree and types will be broekn as well, see build log of https://github.com/nuxt/nuxt.js/pull/9067, this pr is moving @types/terser-webpack-plugin into nuxt codebase as a quick fix

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #9053, opened 2021-03-26
fix(types): css-loader v4 type definition

zzzmisa:
## Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

I fixed the type definition for CssLoaderOptions to suit css-loader v4.

Also see:
https://github.com/webpack-contrib/css-loader/tree/v4.3.0#options

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [x] All new and existing tests are passing.

PR #9088, opened 2021-03-31
v2.15.4

### LINKED ISSUE #3389: [edge] Deployment of new version breaking site for live users. ( Loading chunk {n} failed error)

### LINKED ISSUE #3828: safari: infinite loop when hot reloading

### LINKED ISSUE #8979: ExtendRoutes/webpack/jiti do not process Vuex route.redirect...
### Versions

- nuxt: 2.15.3

### Steps to reproduce
```
	router:{
		trailingSlash:false,
		extendRoutes(routes,resolve){
			routes.unshift({
				name:'login',
				path:'/login',
				redirect(to,...a){
					return {name:'loginelse'}
				}
			})
		}
	}
```
is processed in .nuxt /router.js to:
```
  mode: 'history',
  base: '/',
  linkActiveClass: 'nuxt-link-active',
  linkExactActiveClass: 'nuxt-link-exact-active',
  scrollBehavior,
  routes: [{
    path: "/login",
    redirect: undefined,
    name: "login"
  }, {
    path: "/",
    component: _4ad08996,
    name: "index"
  }],
  fallback: false
}
```

### What is Expected?
as described in docs : [https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router#extendroutes](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router#extendroutes):
`The schema of the route should respect the vue-router schema.` ([vue-router schema](https://router.vuejs.org/en/))

redirect is not processed by webpack and/or jiti...

### LINKED ISSUE #8995: custom components in vuetify-dialog not working

### LINKED ISSUE #9009: server.js - getMatchedComponents is called with wrong route

### LINKED ISSUE #9016: fetch is not being called for components added on the client side
When adding components on the client side `fetch` is not being called or is linked to the wrong payload data. For instance when showing a dialog that fetches CMS content.

### Versions

- nuxt: v2.15.3
- node: v14.15.1

### Reproduction
https://github.com/woutervdm/nuxt-fetch-test-case

```vue
<template>
  <div class="container">
    <button
      type="button"
      @click="click"
    >
      Add component
    </button>
    <hello-world
      v-for="i of number"
      :key="i"
    />
  </div>
</template>

<script>
export default {
  data: () => ({
    number: 1
  }),
  methods: {
    click () {
      this.number++
    }
  }
}
</script>
```

Using the HelloWorld component:
```vue
<template>
  <div>
    {{ data }}
    <pre>{{ $fetchState }}</pre>
  </div>
</template>

<script>
export default {
  name: 'HelloWorld',
  data: () => ({
    data: ''
  }),
  fetch () {
    this.data = 'Hello world!'
  }
}
</script>
```

### Steps to reproduce
Run `yarn generate` and `yarn start` and click on the "add component" button multiple times.

### What is Expected?
`fetch` should be called for each new component and each added component should contain the text "Hello world!".

`fetch` is not being called and only the firstly added component contains the text. Probably because it is using the same fetch key as the staticly generated component (which also seems incorrect behavior).

### LINKED ISSUE #9017: CSP Policy Names List should not be mandatory
### Versions

- nuxt: 2.15.3
- node: 14.16.0

### Reproduction

https://codesandbox.io/s/condescending-turing-ihmzr?file=/nuxt.config.js

<details open>
<br>

nuxt.config.js
`render: {
    csp: {
      policies: {
        "base-uri": ["'self'"]
      }
    }
  }`
</details>

### Steps to reproduce
* In SSR context add render.csp.policies to your nuxt.config.js.
* then add just a single csp policy name, f.e. "base-uri" with whatever value you like as string[]
* run `npm run build` or `nuxt build`

### What is Expected?
The build is successful since the amount of policies is not mandatory. I should be able to start the server by running `npm run start` and see how everything works well an the csp is executed as well.

Eventhough some of the names are deprecated (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/referrer) or simply in some scopes not needed at all, i am forced to add every single name and give it a value. If I don't do that, the build fails because there is a specific Record List added in file https://github.com/nuxt/nuxt.js/blob/0a98866894f018019d26926f1846396c4be71c0d/packages/types/config/render.d.ts#L37

### Issue can be solved by
Surrounding the `<Record>` with an `<Partial>, like
`policies?: Partial<Record<CspPolicyName, string[]>>`

### LINKED ISSUE #9030: @nuxt/types should be a valid JS package (w/ an index.js) so that TS emitDecoratorMetadata works
### Versions

- nuxt: 2.15.3
- node: 14.15.4
- typescript: 4.2.3

### Reproduction

No reproduction, but I've shown an example and explained it here: https://github.com/microsoft/TypeScript/issues/18008#issuecomment-804299674

For more info on the entire issue and how emitDecoratorMetadata see the entire issue: https://github.com/microsoft/TypeScript/issues/18008

### Steps to reproduce

image

Create a TypeScript class in Nuxt.js that is decorated with a TS decorator & with emitDecoratorMetadata turned on, and have one of the constructor params be one of the types exported from `@nuxt/types`.

### What is Expected?

The code should compile.

Since emitDecoratorMetadata: true leaves the import of `@nuxt/types` in JavaScript in the compiled build, webpack throws an error about not being able to resolve `@nuxt/types`, because the package doesn't have an index.js file.

image

### LINKED ISSUE #9049: (vue-app) - route navigation guards are not processed on client side when calling `createApp`

### LINKED ISSUE #9074: Allow specific names in publicRuntimeConfig
Hi and thanks for your work on Nuxt.

**Request**

As I mentioned in the official Nuxt `Discord`, can you please consider allowing users to use words/names such as `app` in project configuration?

It's short, strong meaning and definitely one of the most common for a configuration within a project. I think this approach that doesn't allow the use of such common names in the project configuration can be very confusing and problematic for debugging.

**Code example**

This worked fine until version `2.15.0`, so it's kind of a `breaking-change`:

```js
// nuxt.config.js

export default {
  publicRuntimeConfig: {
    // Doesn't work after version '2.14.12'
    app: {
      title: 'New Title',
      titleSeparator: '-',
      titleTemplate: 'Title Template',
      description: 'New Description',
    },
  },
}
```

**Suggestion**

As Daniel said in Discord chat:

> Nuxt is using the `app` namespace for internal dynamic config.

It might make more sense to use alternatives under the hood (in Nuxt core such as `_app` or `__app`) and allow users to use the normal versions (such as `app`).

**Other examples**

This is allowed in `Next.js`.

### LINKED ISSUE #9085: [types] Type '{ sassOptions: {}; }' is not assignable to type 'Options'.

pi0:
### 🐛 Bug Fixes

- `vue-app`
  - #8978 Reload page once after loading chunk error (resolves #3389)
  - #9008 Fallback to global nuxt instance of `$root` is not available (resolves #8995)
  - #9010 Use `app.context.route` to match components in server (resolves #9009)
  - #9024 Serialize redirect functions in routes  (resolves #8979)
  - #9026 Re-call `$fetch` if data is missing in payload (resolves #9016)
  - #9050 Use `app.context.route` for resolving components (resolves #9049)
  - #9075 Move internal app runtime config to `_app` namespace (resolves #9074)
- `config`
  - #9014  Set `compileType` to `icss` for `css-loader` v4
- `server`
  - #9034 Avoid caching `.js` assets in development to fix HMR in safari (resolves #3828)
- `cli`
  - #8996 Remove warning if postcss 8 is installed (resolves https://github.com/nuxt/nuxt.js/issues/8087#issuecomment-800167111)

### 👕 Types

- #9078 Inline `@types/terser-webpack-plugin`
- #9086 Use options from `sass-loader` rather than `sass` (resolves #9085)
- #9053 Css-loader v4 type definitions
- #9018 Don't require all csp policies as keys (resolves #9017)
- #9059 Add stub entrypoint to allow node resolution of `@nuxt/types` (resolves #9030)

### 💖 Thanks to

- ChangJoo Park (박창주) (@ChangJoo-Park)
- Evan Ma (@EvanMaFYH)
- Thomas Beduneau (@enwin)
- @rinu
- @zzzmisa


mohmmad-w:
Nnkk

mohmmad-w:
>  Nnkk

mohmmad-w (commented):
#9050

mohmmad-w (approved):

PR #9143, opened 2021-04-12
feat(generator): allow setting up a folder when public path has url

NigmaX:
## Types of changes

## Description

Allows to separate the pages and the content into different folders when publicPath property is set to a url.
(e.g. - I would like to put all the content in a _nuxt folder and the pages outside of it)
so the settings would be:
```
 publicPath: 'cdn.something.com/_nuxt/'
 publicPathFolder: '/_nuxt/'
```
it allows the ability to cdn the payload and the state and all assets in a folder, thus not contaminating the root folder that holds only pages. 
that way i can cdn the "_nuxt" folder, and i can be sure that it will contain all the assets and data.

## Checklist:

- [ x] My change requires a change to the documentation.

Im not sure how to test though


clarkdo (dismissed):

packages/generator/src/generator.js
```diff
@@ -26,7 +26,7 @@ export default class Generator {
     this.distPath = this.options.generate.dir
     this.distNuxtPath = path.join(
       this.distPath,
-      isUrl(this.options.build.publicPath) ? '' : this.options.build.publicPath
+      isUrl(this.options.build.publicPath) ? this.options.build.publicPathFolder : this.options.build.publicPath
```
pi0, line 29: What if instead of introducing a new option, using `pathname` of `publicPath`? (so if `publicPath` is `https://cdn.com/something/_nuxt`, we set it to `dist/something/_nuxt`)

NigmaX: Sound better and cleaner.

NigmaX:
@pi0 thanks for the feedback,
I pushed a fix.
let me know if we need anything else, 
thanks!

danielroe:

PR #9165, opened 2021-04-17
chore(ci): collect coverage for each matrix run

rlam3:
Remove conditional run for codecov

## Types of changes
- [ ] CI fix to run codecov without conditions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -152,7 +152,6 @@ jobs:
 
       - name: codecov
         uses: codecov/codecov-action@v1
-        if: matrix.os == 'ubuntu-latest' && matrix.node == '10'
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unittests


rlam3:
Link: https://github.com/nuxt/nuxt.js/issues/9166

.github/workflows/test.yml
```diff
@@ -152,7 +152,6 @@ jobs:
 
       - name: codecov
         uses: codecov/codecov-action@v1
```
danielroe, line 154: Good spot. 💯 But we probably don't want to run on every run, just once per commit.
```suggestion
        uses: codecov/codecov-action@v1
         if: matrix.os == 'ubuntu-latest' && matrix.node == '14'
```

rlam3: Thank you. I was wondering why coverage hasn't run for so long. Good to know. Should we create a future task to remind us to update this setting to the next LTS node?

pi0: Removing the conditional step makes sense. Purpose of running unit tests in a matrix, is to ensure covering differences between node versions and platforms. (a condition like `if (windows) { something }` is only covered in one. And this step is not adding overhead but collecting coverage results (so that codecov finally merges them for final report)

pi0 (approved):

PR #9180, opened 2021-04-22
fix(webpack): use `javascript/auto` for js rule

danielroe:
## Types of changes

## Description
Webpack 4 + `.mjs` dependencies currently causes an issue where _that_ package's sub-dependencies are not imported in the correct context and the following error is generated (for example):
```
Can't reexport the named export [...] from non EcmaScript module (only default export is available)
```

This PR turns off webpack 4's .mjs handling to avoid broken dependencies. I expect this will be a temporary fix whilst the ecosystem moves to native ESM. Note that there is no such issue in webpack 5.

## Checklist:


pi0 (approved):

danielroe:
(Note, this is not sufficient to solve this issue when using with PNP.)

dargmuesli:
Will there be a release including this change soon? :) Just wondering, I've been waiting for that since early June :star_struck:

pi0:
Soon @dargmuesli with 2.16 (since is a risky change). In the meantime, you can try using `nuxt-edge` and feedback if working :)

PR #9225, opened 2021-05-01
chore(ssr): opt-out nuxt.js ssr app from Google's FLoC

soub4i:
## Types of changes

## Context

FLoC lets advertisers use behavioral targeting without cookies. It runs in Google’s Chrome browser and tracks a user’s online behavior. [more info](https://github.com/WICG/floc) by default your website will allow FLoC (categorizing you app users in Cohorts)

## Description

This PR is a way to opt-out of your app create by nuxt.js from being sending your user's data to this Google service. The new code check `config.nuwt.js` file for a property `floc` if it `true` will not do anything if it's `false` it will add meta-data to prevent Google chrome from sending the data to FLoC service 

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderers/ssr.js
+++ b/packages/vue-renderer/src/renderers/ssr.js
@@ -153,6 +153,10 @@ export default class SSRRenderer extends BaseRenderer {
       HEAD += this.renderResourceHints(renderContext)
     }
 
+    // Inject meta to disable Google's FLoC
+    if (!this.options.floc) {
+      HEAD += `<meta http-equiv="Permissions-Policy" content="interest-cohort=()"/>`
+    }
     // Inject styles
     HEAD += this.renderStyles(renderContext)
 


danielroe:
@AbderrahimSoubaiElidrissi Thank you for the PR ❤️ However, this is probably not the right time for it. Here are a few reasons:

1. FLoC is a proposal that hasn't yet been accepted. If it is, it is likely to be opt-in only. At the moment there is an opt-out Chrome study gathering data on how it might work.
2. The header wouldn't actually disable FLoC - it would simply ensure that the site it's on isn't included in the cohort calculation.
3. The header is likely unnecessary for most users (for example, if you aren't actively serving ads).

There is some good information out there, including [in this blog post](https://seirdy.one/2021/04/16/permissions-policy-floc-misinfo.html). However, this is something we'll be keeping a close eye on and we'll revisit later.

PR #9238, opened 2021-05-04
feat: add http/2.0 support through https (spdy)

Upd4ting:
## Types of changes

## Description
Add http/2.0 support with `sdpy` module instead of `https`. Will only work through https as browser like Firefox and Google chrome implemented http/2.0 only for https.

This is a must have for google lighthouse performance test!

## Checklist:
- [x] All new and existing tests are passing. => I haven't been able to test as the test seems broken but not by my changes.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -24,6 +24,7 @@
     "serve-placeholder": "^1.2.3",
     "serve-static": "^1.14.1",
     "server-destroy": "^1.0.1",
+    "spdy": "^4.0.2",
     "ufo": "^0.7.2"
   },
   "publishConfig": {
--- a/packages/server/src/listener.js
+++ b/packages/server/src/listener.js
@@ -1,5 +1,5 @@
 import http from 'http'
-import https from 'https'
+import https from 'spdy'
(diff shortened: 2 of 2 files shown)


danielroe (commented):
This is an elegant PR but I'm a bit wary of using `spdy` which adds [753 kB install size](https://packagephobia.com/result?p=spdy) and is not undergoing active development: https://github.com/spdy-http2/node-spdy.

(The issue with the test, by the way, was just the need to update mocking `https` -> `spdy`.)

pi0:
SPDY protocol (as well as `node-spdy`) is deprecated and dropped in most major browsers (chrome, safary, safari).

We might use [http2.createSecureServer](https://nodejs.org/api/http2.html#http2_http2_createsecureserver_options_onrequesthandler) in [listener](https://github.com/nuxt/nuxt.js/blob/dev/packages/server/src/listener.js#L68) for supporting h2.

pi0:
Closing to continue http2 support in [listhen](https://github.com/unjs/listhen/issues/10) which is used for upcoming CLI (usable for nuxt2 users too without need to maintain several places)

Issue #7884, opened 2020-08-10
Async data return typed data with TypeScript


deozza:
### Versions

- nuxt: ^2.14.1
- node: 14.7.0
 
### Reproduction

``` ts
// ~/class/user/user
export class User{
/*
properties instanciation + constructor
*/
	public getCleanedForRequest() {
                return 'foo';
	}
}

```

```vue
// ~/pages/user.vue
<script lang="ts">
import { Component, Vue } from "nuxt-property-decorator";

import {User} from "~/class/user/user";

interface UserSpecificVueInterface{
	user: User,
	loading: boolean,
	error: any
}

@Component({
	async asyncData({$axios, $auth, route}): Promise<UserSpecificVueInterface>{
		const link = '/link/to/resource';
		const token = $auth.getToken('local');
		const userFromApi = await $axios.$get(link, {
			headers: {
				Authorization: token
			}
		});

		return {
			user: new User(userFromApi),
			loading: false,
			error: null
		};
	}
})
export default class UserSpecificVue extends Vue implements UserSpecificVueInterface{
	user: User;
	loading: boolean = false;
	error: any = null;

	async updateUser(){

		this.loading = true;
                console.log(this.user instanceof User); //should be true, returns false
                console.log(this.user.getCleanedForRequest()); // should echo 'foo', returns `getCleanedForRequest() is not a function of user`    
	}
}

</script>

```

<details open>
<summary>Additional Details</summary>
<br>
</details>

### What is Expected?

I want to use asyncData function to fetch a resource from an API at the page loading. This resource is then used as a typed property with a custom TypeScript class, with its properties and its functions. 

### What is actually happening?

The property is not typed according to the custom TypeScript class

odanado:
same issue: https://github.com/nuxt/typescript/issues/48

rx-837:
@deozza what mode are you using?

if mode is "universal" (ssr + client navigation) then behavior is normal

how it works:
 - on server: calls asyncData then you returns instance of User 
 - on server: render with asyncData result
 - on server: serialize asynData result and write to script(window.__NUXT__ = ...), property user transforms from class instance to object
 - on client: starting hydration - create your component from raw html and render it with data from window.__NUXT__

deozza:
Sorry for the late answer. Yes my mode is `universal`. I wasn't aware of the particular way the rendering was made in this mode.

[...]

PR #9239, opened 2021-05-04
feat(types): add `asyncData` return types to component instance type

carbotaniuman:
## Types of changes

## Description

This resolves the closed issues #7884 and nuxt/typescript#48. using a variant of @odanado work [here](https://github.com/nuxt/typescript/issues/48#issuecomment-672967770). This allows TypeScript users to use the return value of asyncData directly in their `computed` and `methods` without having to duplicate things again in `data`.

## Checklist:

I'm not sure where to add these kind of tests, or even if we have tests on the type definitions themselves.


danielroe (approved):
Looks good to me but I'd love @kevinmarrec's eyes on it too.

kevinmarrec:
Thanks for the PR @carbotaniuman 

@danielroe I didn't find time yet to check everything cause it needs to perfectly know Vue types to see if there's any mistake.

Also, I thought `fetch` was used over `asyncData` since some versions now ? But it may differ depending on user needs.

Finally, before merge I'd like something that show/prove me everything still work on medium to big projects with these changes.

Radiergummi:
> Also, I thought `fetch` was used over `asyncData` since some versions now ? But it may differ depending on user needs.

They do slightly different things - `asyncData` blocks rendering (there's a bit of irony in that) and allows redirecting before finishing the navigation, while fetch is invoked in a component context, so navigation is already complete.

carbotaniuman:
Good point. This would probably be better when Nuxt 3 comes around.

rchl:
Can you please check the follow up at #19526? This doesn't seem to cause errors (at least with Vue 2.7, haven't checked 2.6).

rchl:
Also this change doesn't even work. At least with vue 2.7:

Screenshot 2023-03-09 at 08 48 44

PR #9245, opened 2021-05-06
fix(server): unregister error event listener

matthieusieben:
## Types of changes

## Checklist:


packages/server/src/listener.js
```diff
@@ -75,8 +75,16 @@ export default class Listener {
     // Call server.listen
     try {
       this.server = await new Promise((resolve, reject) => {
-        this._server.on('error', error => reject(error))
-        const s = this._server.listen(listenArgs, error => error ? reject(error) : resolve(s))
+        this._server.on('error', reject)
+        const s = this._server.listen(listenArgs, (error) => {
+          s.removeListener(reject)
```
danielroe, line 80: I don't think this will work as written.

How about this?
```suggestion
          this._server.off('error', reject)
```

pi0: Alternatively, we can use server.once i guess

danielroe: Yes, that's nicer.

clarkdo (commented):
Can you please fix the CI error?

danielroe (approved):

pi0 (approved):
Thanks!

PR #9262, opened 2021-05-10
fix(build):  change hash filename to query hash filename for ChunkLoa…

icchis:
I think it's better to change the naming format of the generated files of build to prevent ChunkLoadError.

```
# example error
ChunkLoadError: Loading chunk 269 failed
```

## Types of changes

## Description
With the current file naming, the file itself is recognized as a new file, which causes a 404 error for existing users (users who are staying), which in turn causes a ChunkLoadError.
The file name itself is not changed, I think we'd better adding the query hash to end of file name without changing the file name itself.

Also, Currently, when using multiple containers and building in each container, created hash file naming in each container.
When that cases, Sometimes it connects to a different container, 404 file error when using load balancers.

Therefore, we need to generate the same file name in every build, and I think the way to update the cache with the query hash value at the end file name is better.

## Checklist:

### DIFF
--- a/packages/config/src/config/build.js
+++ b/packages/config/src/config/build.js
@@ -14,12 +14,12 @@ export default () => ({
   serverURLPolyfill: 'url',
   filenames: {
     // { isDev, isClient, isServer }
-    app: ({ isDev, isModern }) => isDev ? `[name]${isModern ? '.modern' : ''}.js` : `[contenthash:7]${isModern ? '.modern' : ''}.js`,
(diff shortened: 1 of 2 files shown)


clarkdo (commented):
Filenames used to have name on them, but it [got reverted](https://github.com/nuxt/nuxt.js/pull/8012)  because of some regression issues, so I suggest setting these names in your config specifically. https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-build#filenames

icchis:
@clarkdo Thanks for checking. So that's how it happened. I'll close this pull request.

PR #9270, opened 2021-05-10
fix(types): pin to `@types/webpack-dev-middleware` 4.1.2

### LINKED ISSUE #9268: [2.15.5] Rule can only have one resource source
I have upgraded nuxt to v.2.15.5. I'm using Node v14 :

```
 ERROR  Rule can only have one resource source (provided resource and test + include + exclude) in {                        11:48:55  
  "use": [
    {
      "loader": "C:\\Users\\xxx\\xxx\\node_modules\\babel-loader\\lib\\index.js",
      "options": {
        "configFile": false,
        "babelrc": false,
        "cacheDirectory": true,
        "envName": "client",
        "presets": [
          [
            "C:\\Users\\xxx\\xxx\\node_modules\\@nuxt\\babel-preset-app\\src\\index.js",
            {
              "corejs": {
                "version": 3
              }
            }
          ]
        ]
      },
      "ident": "clonedRuleSet-4[0].rules[0].use[0]"
    }
  ]
}

  "use": [
  {
  "loader": "C:\\Users\\xxx\\xxx\\node_modules\\babel-loader\\lib\\index.js",
  "options": {
  "configFile": false,
  "babelrc": false,
  "cacheDirectory": true,
  "envName": "client",
  "presets": [
  [
  "C:\\Users\\xxx\\xxx\\node_modules\\@nuxt\\babel-preset-app\\src\\index.js",
  {
  "corejs": {
  "version": 3
  }
  }
  ]
  ]
  },
  "ident": "clonedRuleSet-4[0].rules[0].use[0]"
  }
  ]
  }
  at checkResourceSource (node_modules\@nuxt\webpack\node_modules\webpack\lib\RuleSet.js:167:11)
  at Function.normalizeRule (node_modules\@nuxt\webpack\node_modules\webpack\lib\RuleSet.js:198:4)
  at node_modules\@nuxt\webpack\node_modules\webpack\lib\RuleSet.js:110:20
  at Array.map (<anonymous>)
  at Function.normalizeRules (node_modules\@nuxt\webpack\node_modules\webpack\lib\RuleSet.js:109:17)
  at new RuleSet (node_modules\@nuxt\webpack\node_modules\webpack\lib\RuleSet.js:104:24)
  at new NormalModuleFactory (node_modules\@nuxt\webpack\node_modules\webpack\lib\NormalModuleFactory.js:115:18)
  at Compiler.createNormalModuleFactory (node_modules\@nuxt\webpack\node_modules\webpack\lib\Compiler.js:636:31)
  at Compiler.newCompilationParams (node_modules\@nuxt\webpack\node_modules\webpack\lib\Compiler.js:653:30)
  at Compiler.compile (node_modules\@nuxt\webpack\node_modules\webpack\lib\Compiler.js:661:23)
  at node_modules\@nuxt\webpack\node_modules\webpack\lib\Watching.js:77:18
  at AsyncSeriesHook.eval [as callAsync] (eval at create (node_modules\@nuxt\webpack\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:42:1)
  at AsyncSeriesHook.lazyCompileHook (node_modules\@nuxt\webpack\node_modules\tapable\lib\Hook.js:154:20)
  at Watching._go (node_modules\@nuxt\webpack\node_modules\webpack\lib\Watching.js:41:32)
  at node_modules\@nuxt\webpack\node_modules\webpack\lib\Watching.js:33:9
  at Compiler.readRecords (node_modules\@nuxt\webpack\node_modules\webpack\lib\Compiler.js:529:11)
```

danielroe:
* avoid webpack 5 being inserted in project

## Types of changes

## Description
`@types/webpack-dev-middleware` added dependency on webpack 5 in [patch release (4.1.3)](https://unpkg.com/browse/@types/webpack-dev-middleware@4.1.3/package.json).

closes https://github.com/nuxt/nuxt.js/issues/9268

## Checklist:

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -27,7 +27,7 @@
     "@types/terser-webpack-plugin": "^4.2.1",
     "@types/webpack": "^4.41.28",
     "@types/webpack-bundle-analyzer": "^3.9.3",
-    "@types/webpack-dev-middleware": "^4.1.2",
+    "@types/webpack-dev-middleware": "4.1.2",
     "@types/webpack-hot-middleware": "^2.25.4",
     "sass-loader": "^10.1.1"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2584,7 +2584,7 @@
   dependencies:
     "@types/webpack" "^4"
 
-"@types/webpack-dev-middleware@^4.1.2":
+"@types/webpack-dev-middleware@4.1.2":
   version "4.1.2"
   resolved "https://registry.npmjs.org/@types/webpack-dev-middleware/-/webpack-dev-middleware-4.1.2.tgz#c683adc6a44d0b66e98f1932ab33d1c32d558142"
   integrity sha512-SxXzPCqeZ03fJ2dg3iD7cSXvqZymmS5/2GD9fANRcyWN7HYK1H3ty6q7IInXZKvPrdUqij831G3RLIeKK6aGdw==


clarkdo (approved):

rchl:
I think this should be fixed in `@types` (eventually).

There was a followup change there that claimed to restore the webpack 4 compatibility but I'm not sure if it succeeded. I've asked a question there: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52863

pi0:
@rchl Indeed. It was a kind of an essential hotfix since npm users were broken. All types are now pinned to the exact version to avoid the same regression happening also from 2.16 we probably start rolling up all types this way don't need to depend on actual dependencies for types only package.

rchl:
Interesting to hear that you've pinned all dependencies. That could end up badly though, I would think. Could easily end up project using multiple versions of types which would result in errors. Just a thought...

pi0:
It depends in pm behavior but @nuxt/types at least exports expected types (based on expected dependency versions). We usually have to deal with this tradeoffs in a  meta fromwork like nuxt (and specially nuxt2 that is depending on not latest versions)

rchl:
> (also i think typescript is not instance based. Two compatible interfaces would work together unless being extended)

But Typescript would fail if two different versions of type definitions would be loaded in the project due to type redefinition.

In any case, we can just wait and see if there are any issues with that.

pi0:
> But Typescript would fail if two different versions of type definitions would be loaded in the project due to type redefinition.

That's why rolling-up is planned :D

rchl:
I'm not sure I understand what "rolling-up" means but I'm thinking about local development so not sure if that's relevant.

To be clear, I'm thinking about a case like:
1. `@nuxt/types` installs dependency `@types/foo` with pinned version.
2. The user installs dependency `@types/bar` that depends on `@types/foo` in a different version
=> types re-definition issue

pi0:
Bundling 3rd party types :) Then this won't happen

PR #9280, opened 2021-05-11
feat(config): support `nuxtrc` in dist directory

danielroe:
## Types of changes

## Description
One major weakness of `buildModules` is that they can't manipulate runtime Nuxt configuration (for example, by injecting a runtime module into the Nuxt config), leaving the invidious choice of having to be installed as a module, asking a user to manually add runtime configuration or other non-optimal choice.

This PR allows manipulation of `nuxt.config` by placing a `nuxtrc` file within the `.nuxt/dist` directory.

## Checklist:


pi0 (dismissed):

packages/config/src/load.js
```diff
@@ -88,10 +88,13 @@ export async function loadNuxtConfig ({
...
 
   // Load Combine configs
-  // Priority: configOverrides > nuxtConfig > .nuxtrc > .nuxtrc (global)
+  // Priority: configOverrides > nuxtConfig > .nuxt/dist/.nuxtrc > .nuxtrc > .nuxtrc (global)
+  const dev = configOverrides.dev ?? options.dev ?? configContext.dev
+  const buildDir = configOverrides.buildDir ?? options.buildDir ?? configContext.buildDir ?? '.nuxt'
   options = defu(
     configOverrides,
     options,
+    !dev ? rc.read({ name: '.nuxtrc', dir: path.resolve(buildDir, 'dist') }) : {},
```
pi0, line 97: One edge case of using dotfiles in `dist` is previously (during the time was using [name] in assets) we had lots of reports that CI/CD pipeline is ignoring them for deployment. It might be probably safer to use `nuxtrc` (not `.nuxtrc`)

clarkdo (approved):

PR #9322, opened 2021-05-21
fix(config): remove unsupported nullish coalescing

### LINKED ISSUE #9321: [edge] Nullish operator available with node 14 only
I think issue is related with that change: 
https://github.com/nuxt/nuxt.js/commit/186df335d926b56f8eec5249d6330363622390b3#diff-cba57dce8689b0b15bac4cd779357605a9c6ce76da7d424eedf8c4eb385b75d8R92 

I think it should be transpilled, but is not, because I get an error:

```bash
/Users/merc/Projects/OpenSource/mymodule/node_modules/@nuxt/config-edge/dist/config.js:1119
  const dev = configOverrides.dev ?? options.dev ?? configContext.dev;
                                   ^

SyntaxError: Unexpected token '?'
```

Or should we use node 14 ? But I don't think so it is required at the moment. 

### Versions

- nuxt: edge 
- node: 12.20 

### Reproduction

just run nuxt-edge on node 12.20

danielroe:
## Types of changes

## Description
* issue introduced in https://github.com/nuxt/nuxt.js/pull/9280

closes #9321

## Checklist:

### DIFF
--- a/packages/config/src/load.js
+++ b/packages/config/src/load.js
@@ -8,6 +8,14 @@ import destr from 'destr'
 import * as rc from 'rc9'
 import { defaultNuxtConfigFile } from './config'
 
+function coalesce(...values) {
+  const value = values.shift()
+  if (value !== null && value !== undefined) {
+    return value
+  }
+  return coalesce(...values)
+}
+
 export async function loadNuxtConfig ({
   rootDir = '.',
   envConfig = {},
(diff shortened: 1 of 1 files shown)


clarkdo (dismissed):

packages/config/src/load.js
```diff
@@ -8,6 +8,14 @@ import destr from 'destr'
 import * as rc from 'rc9'
 import { defaultNuxtConfigFile } from './config'
 
+function coalesce (...values) {
```
Atinux, line 11: Might be worth adding a comment above explaining what this function does ☺️

danielroe, line 11:
```suggestion
// Returns the first value that isn't null or undefined (as replacement for ??)
function coalesce (...values) {
```

danielroe:
```suggestion
// Returns the first value that isn't null or undefined (as replacement for ??)
function coalesce (...values) {
```

pi0:
(pending to check possibility of adding esbuild to build scripts to avoid accidental syntax usage in the future)

PR #9328, opened 2021-05-24
chore(pkg): build for es2019 target

### LINKED ISSUE #9321: [edge] Nullish operator available with node 14 only

pi0:
## Types of changes

## Description

Use `esbuild` with `taget: es2019` (using `node*` transform to CJS which breaks rollup chunks!) to build dist files. Fixes #9321 Closes #9322

## Checklist:


clarkdo (dismissed):

clarkdo:
Nice change

clarkdo:
OOC, what will be broken when using node ?

pi0:
@clarkdo cli dist (using dynamic imports) was getting broken

clarkdo:
Is dynamic import not transpiled ?
I guess a version<12.20 will work ? But es2019 look better to me as well.

pi0:
> Is dynamic import not transpiled ?

Actually it  transpiled to cjs by esbuild when using node while shouldn't (rollup is esm native and supports cjs with CommonJS plugin). Maybe we could use node if there was a node-es target...

clarkdo:
Right, so node version >= 12.20 in target should work as it fully supports dynamic import.

clarkdo:
OK, it seems not https://github.com/evanw/esbuild/issues/1084#issuecomment-810004991

PR #9332, opened 2021-05-24
feat(vue-app): `context.beforeSerialize` method

### LINKED ISSUE #352: Mismatching childNodes vs. VNodes
### Version

[4.0.1-rc.3](https://github.com/apollo-module/releases/tag/4.0.1-rc.3)

### Reproduction link

[https://github.com/mlsen/nuxt-apollo-report](https://github.com/mlsen/nuxt-apollo-report)

### Steps to reproduce

1. yarn install
2. yarn dev
3. Visit the index page and have a look at the console

### What is expected ?

No errors and data being loaded on the server (SSR)

It's giving me a warning

`Mismatching childNodes vs. VNodes`

and an error 

`The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside <p>, or missing <tbody>. Bailing hydration and performing full client-side render.`

The data is then fetched on the client again.

This only happens using the "apollo" object in my component, asyncData works fine.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/apollo-module/issues/c333">#c333</a>)</em></sub></div>

### LINKED ISSUE #8620: beforeNuxtRender is called too soon
### Versions

- nuxt: v2.14.0
- node: 10.16.0

### Reproduction

https://github.com/MartinMalinda/before-nuxt-render

### Steps to reproduce

npm install in the repo and open the dev server, open `layouts/default.vue`.

### What is Expected?

When I change pinia state on the server, it is passed on `beforeNuxtRender` to `nuxtState` and then the state should be picked on the client.

There is an initialization of pinia store:
```
    const store = useDataStore();
    onServerPrefetch(async () => {
      await wait(400);
      store.state.name = "changed name";
    });
```

And a slight change to the state on the server side. Where `initial name` is changed to `changed name` after 400ms. Because pinia should transfer state from the server to the client these things should happen:

1) Server waits 400ms ✅ 
2) Server sends SSR HTML with `changed name` ✅ 
3) Client pinia should pick up state ❌ 
4) Client should render `changed name` ❌ 

`changed name` turns into `initial name` on the client.

### The issue: beforeNuxtRender (probably)

To pass the state from server to client, pinia uses `nuxtState` together with `beforeNuxtRender`:

https://github.com/posva/pinia/blob/v1/nuxt/plugin.js#L17

When I put a bunch of console logs around, it seems like the `context.beforeNuxtRender` gets called almost immediately and not in fact, just before render.

Not only it does not wait the promise from `useServerPrefetch` to resolve, in this case it's called just right away so it's called even before the pinia stores initialize so not even default store state is passed.

I think something had to change on the Nuxt side, because I remember this working for me when I checked it some months ago, probably on Nuxt < 2.12.

### LINKED ISSUE #8852: Plugin state is lost on hydration
### Versions

- nuxt: 2.15.0
- node: v14.15.2

### Reproduction

https://codesandbox.io/s/confident-johnson-gghmq?file=/pages/index.vue

The only relevant files are `pages/index.vue`, `plugins/stateTest.js` and `nuxt.config.js` (to include the plugin).

### Steps to reproduce

1. Write a plugin that has some injected state
2. Modify the state inside `fetch`
3. Observe that the state is not updated on the client side.
4. Observe that the state **is** updated in the server side rendered output (view-source:https://gghmq.sse.codesandbox.io/).

### What is Expected?

The client state should match the server side state

The client state is never updated, while the server side state is.

---

I encountered this issue while creating a plugin for managing state from the server. By doing `await this.$myPlugin.users.load()` and then accessing that data using `this.$myPlugin.users.items` I could greatly simplify both templates and data loading. While moving this to the `fetch` hook however, It didn't work. I found out that this was the root cause.

I'm not sure if this is necessarily a bug, but if nothing else it seems like something that might've been overlooked (or I've missed some important documentation).

### LINKED ISSUE #9155: Need a new way for plugins to inject their state into `nuxtState`

It is related to issue #8620 ("beforeNuxtRender is called too soon").

Since Vue’s `serverPrefetch` (`onServerPrefetch`) and Nuxt’s [new fetch](https://ja.nuxtjs.org/blog/understanding-how-fetch-works-in-nuxt-2-12/) (which utilizes Vue’s serverPrefetch) were introduced, `beforeNuxtRender` is no longer the proper hook for plugins to inject their state into `nuxtState`.

Nuxt < 2.12
→ asyncData() & fetch() → **beforeNuxtRender()** → start rendering App → **serverPrefetch()** → render HTML

Nuxt >= 2.12
→ asyncData() → **beforeNuxtRender()** → start rendering App → **serverPrefetch()** → **fetch()** → render HTML

Some plugins such as vue-apollo currently rely on `serverPrefetch` (onServerPrefetch) to perform the component-level prefetching, and therefore these plugins cannot do the client hydration properly with the current Nuxt.

Some users seems to do a temporary hack as a workaround https://github.com/nuxt/nuxt.js/issues/8852#issuecomment-789946319.

I think Nuxt should provide plugins with a new hook that is suitable for updating `nuxtState`.

### Describe alternatives you've considered

Instead of adding a new hook, we could make Nuxt call beforeNuxtRender() just after the App is rendered if this doesn't cause serious backward compatibility issues.

dovca:
Edited by @Atinux

## Types of changes

## Description

This PR adds a `beforeSerialize` method in the context that can be used to modify `window.__NUXT__` after waiting for `serverPrefetch`, `fetch` and `asyncData`, store actions and so on. This will solve hydration issues with plugins like @nuxtjs/apollo and others.

⚠️ Note that `beforeSerialize` expects a function to be synchronous now since the `rendered` method is synchronous in Vue: https://github.com/vuejs/vue/blob/dev/src/server/create-renderer.js#L89

The official documented way to wait until the ssr renderer has finished and inject the final state is using [ssrContext.rendered](https://ssr.vuejs.org/guide/data.html#final-state-injection).

Resolves: #9155, resolves #8620, resolves #8852 and possibly resolves nuxt-community/apollo-module#352 after some more work in that repo.

## Checklist:

- [x] I have updated the documentation accordingly. (PR: nuxt/nuxtjs.org#1475)

Will be happy to write up the docs if/when this gets accepted.


Atinux:
Hi @dovca 

Thank you for the pull request, I believe this is a nice implementation and needed since the new `fetch`.

I let Pooya and Daniel take a look too at it, if good, you can start adding a PR to nuxt/nuxtjs.org repo to document it.

atinux (dismissed):

danielroe (dismissed):

Atinux:
Another thing in mind, we may also just move `beforeNuxtRender` so it works for any scenario.

`afterNuxtRender`method is not really good: we are still before Nuxt render the HTML

Atinux:
Ideally, we should have `beforeNuxtRender` inside `rendered` and force it to be synchronous, but this would lead to a breaking change.

One solution might be to rename `afterNuxtRender` to something like `beforeStateSerialize` (suggestion from @danielroe)

pi0:
We can keep `beforeNuxtRender` since name is valid, asynchronous hook and already used by many plugins (deprecating is not necessary) but agree that `afterNuxtRender` naming is not good. What about `beforeSerialize()` ?

packages/vue-app/template/utils.js
```diff
@@ -258,7 +258,11 @@ export async function setContext (app, context) {
       }
     }
     if (process.server) {
-      app.context.beforeNuxtRender = fn => context.beforeRenderFns.push(fn)
+      app.context.beforeNuxtRender = fn => {
+        console.warn('`context.beforeNuxtRender(fn)` is deprecated in favour of `context.beforeRender(fn)`. The function updating the `nuxtState` has to be synchronous. Learn more on https://nuxtjs.org/docs/2.x/internals-glossary/context#beforenuxtrender')
```
pi0, line 262: There is no need to deprecate :)

packages/vue-app/template/utils.js
```diff
@@ -258,7 +258,11 @@ export async function setContext (app, context) {
...
+        context.beforeRenderFns.push(fn)
+      }
+      app.context.beforeRender = fn => context.beforeRenderHooks.push(fn)
```
pi0, line 265: We are still during server rendering process. Using beforeSerialize or similar new name would probably be more clear (and less confusing between old beforeNuxtRender) 😊

danielroe (dismissed):

packages/types/app/index.d.ts
```diff
@@ -87,6 +88,7 @@ export interface Context {
   error(params: NuxtError): void
   nuxtState: NuxtState
   beforeNuxtRender(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
+  beforeSerialize(fn: (params: { Components: VueRouter['getMatchedComponents'], nuxtState: NuxtState }) => void): void
```
danielroe, line 91:
```suggestion
  beforeSerialize(fn: (nuxtState: NuxtState) => void): void
```

packages/vue-app/template/server.js
```diff
@@ -120,16 +122,21 @@ export default async (ssrContext) => {
   const beforeRender = async () => {
     // Call beforeNuxtRender() methods
     await Promise.all(ssrContext.beforeRenderFns.map(fn => promisify(fn, { Components, nuxtState: ssrContext.nuxt })))
-    <% if (store) { %>
+
     ssrContext.rendered = () => {
+      // Call beforeSerialize() hooks
+      ssrContext.beforeSerializeFns.forEach(fn => fn({ Components, nuxtState: ssrContext.nuxt }))
```
danielroe, line 128:
```suggestion
      ssrContext.beforeSerializeFns.forEach(fn => fn(ssrContext.nuxt))
```

test/fixtures/basic/pages/before-serialize.vue
```diff
@@ -0,0 +1,26 @@
...
+<script>
+export default {
+  data () {
+    return {
+      nuxtState: null
+    }
+  },
+  fetch () {
+    if (process.server) {
+      this.$root.context.beforeSerialize(({ nuxtState }) => {
```
danielroe, line 17:
```suggestion
      this.$root.context.beforeSerialize((nuxtState) => {
```

atinux (dismissed):

danielroe (dismissed):

Atinux:
Added the PR for docs: https://github.com/nuxt/nuxtjs.org/pull/1475

clarkdo (dismissed):

pi0 (approved):

fabis94:
Hi, when will this be released?

pi0:
Hi @fabis94 it will be released with 2.16 (next minor) but in the meantime you can try switching from `nuxt` to `nuxt-edge`

PR #9337, opened 2021-05-26
fix(babel-preset-app): respect explicit options.targets for modern preset

### LINKED ISSUE #9315: Some corejs modules are imported regardless of target browsers
### Versions

- nuxt: 2.15.6
- node: 16.1.0

### Reproduction

https://github.com/Solant/nuxt-corejs-issue

This is fresh `yarn create nuxt-app` with some config changes:

<details open>
<summary>nuxt.config.js changes</summary>
<p>

```js
  build: {
      corejs: 3,
      babel: {
        presets({ isModern }, [preset, options]) {
            if (isModern) {
                options.targets = 'last 1 Chrome versions';
            }
        },
      },
  },
    modern: 'client',
```

</p>
</details>

### Steps to reproduce
Run `yarn build -a`, check `modern.html` output

### What is Expected?
`core-js` modules like `web.url`, `es.promise` should not be included in build (as per compat data https://github.com/zloirock/core-js/blob/master/packages/core-js-compat/src/data.js#L1580)

core-js modules are still there
image

clarkdo:
## Types of changes

Fix https://github.com/nuxt/nuxt.js/issues/9315

We used to limit targets of modern preset can only be `{esModules: true}` for not breaking matching between browsers which support `<script type="module">` and corresponding modern syntax, but it makes sense that let users specify a more modern targets manually if they know exactly what is the target browsers.

@pi0 Do you think we need to add a cli info message here about overrding default modern preset targets may break browsers which support `<script type="module">` but not implement more modern ecma syntax yet ?

## Checklist:

### DIFF
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -97,12 +97,7 @@ module.exports = (api, options = {}) => {
     modern: { esmodules: true }
   }
 
-  let { targets = defaultTargets[envName] } = options
-
-  // modern mode can only be { esmodules: true }
-  if (envName === 'modern') {
-    targets = defaultTargets.modern
-  }
+  const { targets = defaultTargets[envName] } = options
 
   const polyfills = []
 


pi0 (approved):

pi0:
Having a way to modernize more than esmodule baseline is a nice idea. 

> Do you think we need to add a cli info message here about overrding default modern preset targets may break browsers which support <script type="module"> but not implement more modern ecma syntax yet ?

We can mention this possibility in documentation and add notes about browser compatibility. But generally i think it is not a bad choice for targeting real modern browsers.

ms-fadaei:
@pi0 

I have a question based on this PR.

With this fix, we just configure the babel. but what about modern mode with server detection? [The source code here](https://github.com/nuxt/nuxt.js/blob/dev/packages/utils/src/modern.js) shows the browsers that can be used in modern mode  (on server detection). but when we change the target in babel, the server serves the wrong file, and everything broke! Am I wrong?

pi0:
You are correct @ms-fadaei. In fact, we should also make server detection configurable otherwise ESM...modern browsers will be broken. /cc @clarkdo

clarkdo:
Good catch, I’ll look into this and try to make a pr

PR #9365, opened 2021-06-01
feat(server): allow disabling `serve-static` middleware entirely

### LINKED ISSUE #9362: `server-static` middleware generating a lot of file system access while it should not
👋 Hello Nuxt team!

### Versions

- nuxt: v2.15.6 (latest version as of today)
- node: v16.1.0

While working on performance improvement, we've been monitoring (very) slow responses from our Nuxt server. And, while digging deeper, we managed to isolate one possible root cause: unexpected file system (FS) `stat` calls.

It seems that for each page request (I mean pages registered through Vue Router), a FS `stat` call is fired to first figure out whether a static file should be returned. See the attached screenshot below.

CleanShot 2021-06-01 at 10 57 42@2x

While 75% of those `stat` calls are faster than 1ms, some may take up to 5 seconds (!). And obviously, when the server waits for 5+ seconds before sending the response, other requests start stacking and pressuring our infrastructure. This phenomenon is even more impacting during high traffic peaks, which results in bad user experience, and unnecessary infrastructure scale up (meaning higher costs overall).

When investigating even further, it seems that those `stat` calls are fired from the connect middleware `serve-static`. See screenshot below.

CleanShot 2021-06-01 at 11 10 37@2x

### Side note.

In production, we do not rely on our node.js server to serve static files. All of them are compiled and then uploaded to Amazon S3 at build time, and then requested from Cloudfront (acting as a CDN in front of S3). 

To be 100% clear: from the browser, we _never_ call `GET /path/to/image.jpg` for example. We _always_ request the asset from `GET https://cloudfront.net/path/to/image.jpg`.

Hence, the `serve-static` middleware is not useful _to us_.

### What is expected?

So, based on the knowledge explained above, I've been looking for a solution to configure the `serve-static` middleware.

I'm aware of the [`render.static`](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-render#static) configuration property, but it seems that it does not solve my issue. I'm not sure about the `prefix` property, but it does not seem to be used to filter out static paths.

I've also taken a look at the [`serve-static` configuration](https://www.npmjs.com/package/serve-static) itself. And again, I didn't find anything useful to us.

So, I finally tried to find a solution to completely disable the `serve-static` middleware. And according to the code I found in this repository, disabling the middleware does not seem possible. See code below.

https://github.com/nuxt/nuxt.js/blob/2ec62617ced873fef97c73a6d7aa1271911ccfd5/packages/server/src/server.js#L90-L94

### Conclusion

So here is my question: How would you solve this issue? Would it be possible to add a configuration property to completely disable the `serve-static` middleware? 

Maybe something like:

```
{
  render: {
    static: false,
  }
}
```

danielroe:
## Types of changes

## Description
Allows explicitly disabling `serve-static` middleware by setting `render.static` to `false`.

closes #9362

## Checklist:

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -87,13 +87,15 @@ export default class Server {
       this.useMiddleware(createTimingMiddleware(this.options.server.timing))
     }
 
-    // For serving static/ files to /
-    const staticMiddleware = serveStatic(
-      path.resolve(this.options.srcDir, this.options.dir.static),
-      this.options.render.static
-    )
-    staticMiddleware.prefix = this.options.render.static.prefix
-    this.useMiddleware(staticMiddleware)
+    if (this.options.render.static !== false) {
+      // For serving static/ files to /
+      const staticMiddleware = serveStatic(
+        path.resolve(this.options.srcDir, this.options.dir.static),
+        this.options.render.static
+      )
+      staticMiddleware.prefix = this.options.render.static.prefix
+      this.useMiddleware(staticMiddleware)
+    }
 
     // Serve .nuxt/dist/client files only for production
     // For dev they will be served with devMiddleware


clarkdo (approved):
Nice, it would be nice to update doc https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-render#static as well.

antoinerey:
Should we also disable the second `serve-static` middleware instance defined a bit later in the code (https://github.com/nuxt/nuxt.js/pull/9365/files#diff-b8c2d26ed3fd075596a838c69b36c537536ae6973cafffdbf481fec464c9b6cfR104-R110)?

pi0 (approved):
LGTM

pi0:
@antoinerey Ideally users shouldn't hit `/_nuxt/` requests and reverse proxy should cache/handle them.

Matausi29 (commented):
Not bad

pi0:
TODO: 
- Docs
- Types for `false` value: https://github.com/nuxt/nuxt.js/blob/9a32fa18afeda72244c50e6a916fa2bc3e9c7ca1/packages/types/config/render.d.ts#L75

antoinerey:
👋 By any chance, any update about when this will be released?

pi0:
Hi @antoinerey there is no specific time but you can use nuxt-edge or for types, in package.json installing `@nuxt/types-edge`

antoinerey:
Thanks for the prompt answer! 

I'm not sure we want to move from the stable channel to the edge one, so we'll probably work on our patience skill. 😁

PR #9407, opened 2021-06-08
fix(vue-renderer): use vue meta in ssr renderer

### LINKED ISSUE #9402: SSRRenderer does not use VueMeta
### Versions

- nuxt: 2.15.0
- node: 14.17.0

### Reproduction

https://github.com/rinu/codesandbox-nuxt

### Steps to reproduce

* `yarn build`
* `TITLE=Nuxt yarn start`
* Compare source of `http://localhost:3000` and `http://localhost:3000/?ssr=1`

To reproduce the issue it is very important not to use dev mode and to only provide the title at start time.

### What is Expected?

Meta title is Nuxt in both SPA and SSR mode.

In SPA mode title is set from env. In SSR mode title is always the default.

In SSR mode the head is generated at build time. The issue of SSR not using the head data is for all elements, title is just an example. In practice I needed this for a script.

rinu:
## Types of changes

## Description

Resolves #9402

While it does fix the original issue, it will almost certainly create new ones instead. This is more like pseudo code to give even more context to the original issue. I hope someone more familiar with the rendering process will help with this.

## Checklist:

### DIFF
--- a/packages/vue-renderer/src/renderers/ssr.js
+++ b/packages/vue-renderer/src/renderers/ssr.js
@@ -8,8 +8,23 @@ import { parsePath, withoutTrailingSlash } from 'ufo'
 import devalue from '@nuxt/devalue'
 import { createBundleRenderer } from 'vue-server-renderer'
 import BaseRenderer from './base'
+import { cloneDeep } from 'lodash'
+import VueMeta from 'vue-meta'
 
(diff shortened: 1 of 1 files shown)


danielroe:
We're approaching end-of-life for Nuxt v2. I think as this is not a fully-worked-out PR, I'm going to close it.

Let me know if you still want to work on this.

PR #9424, opened 2021-06-10
fix(types): add nuxt.config alias type

unrevised6419:
https://nuxtjs.org/docs/2.x/directory-structure/nuxt-config#alias


danielroe (approved):
Thank you! ❤️

clarkdo (approved):

PR #9431, opened 2021-06-14
fix(vue-app): check whether route exists within Nuxt app before replacing

danielroe:
## Types of changes

## Description
Fixes redirection issue introduced in https://github.com/nuxt/nuxt.js/pull/9050

## Checklist:

### DIFF
--- a/packages/vue-app/template/index.js
+++ b/packages/vue-app/template/index.js
@@ -270,7 +270,12 @@ async function createApp(ssrContext, config = {}) {
 
   // Wait for async component to be resolved first
   await new Promise((resolve, reject) => {
-    router.replace(app.context.route.fullPath, resolve, (err) => {
+    const { route } = router.resolve(app.context.route.fullPath)
+    // Ignore 404s rather than blindly replacing URL
+    if (!route.matched.length && process.client) {
+      return resolve()
+    }
+    router.replace(route, resolve, (err) => {
       // https://github.com/vuejs/vue-router/blob/v3.4.3/src/util/errors.js
       if (!err._isRouter) return reject(err)
       if (err.type !== 2 /* NavigationFailureType.redirected */) return resolve()


pabloalvzal:
Hello @danielroe @pi0 

I am migrating our company's web client to Nuxt and I may have found a bug on the latest version (bug fix). I was working on Nuxt 2.15.6 and just did the upgrade to 2.15.7. 

It seems that on normal route navigation, all things work well! But on page reload the forward slash `/` gets replaced in the route by `%2F`. In our case, this leads to 404's because we use the URL to build the API call. 

I can probably work around it by validating the API call everytime, but I thought I should let you know.

danielroe:
@pabloalvzal Could you provide a bit more info on the situation in a new issue?

Sendoo:
Hello @danielroe @pi0
I have a similar problem, only the trailing slash is simply removed from the url on the first load.
With further routing, the trailing slash is not removed.
It started after updating to version 2.15.17.
Version 2.15.16 does not have this problem.

joaquinvexels:
> Hello @danielroe @pi0
> 
> I am migrating our company's web client to Nuxt and I may have found a bug on the latest version (bug fix). I was working on Nuxt 2.15.6 and just did the upgrade to 2.15.7.
> 
> It seems that on normal route navigation, all things work well! But on page reload the forward slash `/` gets replaced in the route by `%2F`. In our case, this leads to 404's because we use the URL to build the API call.
> 
> I can probably work around it by validating the API call everytime, but I thought I should let you know.

same

curtgrimes:
I'm experiencing a redirect loop on this update compared to the last patch release (redirecting from no slash to slash to no slash etc. until browser stops the redirect loop) but I do not have time at the moment to create a minimal reproduction.

mvandiest:
```
Hello @danielroe @pi0

I am migrating our company's web client to Nuxt and I may have found a bug on the latest version (bug fix). I was working on Nuxt 2.15.6 and just did the upgrade to 2.15.7.

It seems that on normal route navigation, all things work well! But on page reload the forward slash / gets replaced in the route by %2F. In our case, this leads to 404's because we use the URL to build the API call.

I can probably work around it by validating the API call everytime, but I thought I should let you know.
```

Also seeing the `%2f` issue on page reload.

darioferderber:
Hi @pi0, @danielroe,

I'm having trouble with routes that have defined aliases after upgrading to 2.15.7 from 2.15.6.

Route example:
```
{
  name: 'Homepage',
  path: '/hr/',
  component: 'pages/Homepage.vue',
  chunkName: 'pages/Homepage',
  pathToRegexpOptions: {
    strict: true
  },
  alias: ['/en/']
},
```

If I navigate to '/en/' (with anchor tag) location gets replaced to '/hr/' (only on client side).

ChristianAlbrecht:
> Hello @danielroe @pi0
> 
> I am migrating our company's web client to Nuxt and I may have found a bug on the latest version (bug fix). I was working on Nuxt 2.15.6 and just did the upgrade to 2.15.7.
> 
> It seems that on normal route navigation, all things work well! But on page reload the forward slash `/` gets replaced in the route by `%2F`. In our case, this leads to 404's because we use the URL to build the API call.
> 
> I can probably work around it by validating the API call everytime, but I thought I should let you know.

We are also seeing this issue with slashes being replaced on page reload or direct navigation.

A few more details.
This only seems to happen for a specific route, which has a named parameter, that can contain multiple levels of path segments. The corresponding path expression:

`path: '/space/:repository([^]*)?/'`

with 2.15.6 we get for `repository` something like `folder/subfolder` but after the update to 2.15.7 it is `folder%2Fsubfolder`. This also replaces our URL at that point, which leads to subsequent errors in our web application so that for now we reverted back to 2.15.6

PR #9499, opened 2021-06-30
feat: browserslist config file support

ms-fadaei:
Add browserslist config file support

## Types of changes

## Description

Resolves: #9337
After supporting modern browsers config in the Nuxt, there were two things that must be worked on it too:
1. Retrieve targets browsers from browserslist config file (.browserlistrc): With this RP, now the Nuxt users can configure their target browsers from a separate file. (that is used by other packages like Autoprifixer, etc too)
2. Target browsers detection on the server, which I reported in #9337 and I think @clarkdo working on it currently

## Checklist:

### DIFF
--- a/packages/babel-preset-app/src/index.js
+++ b/packages/babel-preset-app/src/index.js
@@ -56,6 +56,17 @@ function getPolyfills (targets, includes, { ignoreBrowserslistConfig, configPath
   ))
 }
 
+function getTargetsFromBrowserslistRc (browserslistEnv = 'client') {
+  const { default: getTargets } = require('@babel/helper-compilation-targets')
+  const targets = getTargets(undefined, { browserslistEnv })
+
+  if (Object.keys(targets).length !== 0) {
+    return targets
+  } else {
+    return undefined
+  }
+}
+
 module.exports = (api, options = {}) => {
   const presets = []
   const plugins = []
@@ -70,7 +81,7 @@ module.exports = (api, options = {}) => {
     useBuiltIns = 'usage',
     modules = false,
     spec,
-    ignoreBrowserslistConfig = envName === 'modern',
+    ignoreBrowserslistConfig,
     configPath,
     include,
     exclude,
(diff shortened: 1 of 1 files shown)


clarkdo (approved):
Looks good. I think we can hold this pr for now until we support customized targets for SSR

clarkdo:
> Target browsers detection on the server, which I reported in fix(babel-preset-app): respect explicit options.targets for modern preset #9337 and I think @clarkdo working on it currently

Thanks for the reminder, I’ll find some time this week to look into it.

pi0:
Thanks for pull request @ms-fadaei <3. I think for properly solving the issue in https://github.com/nuxt/nuxt.js/pull/9337#issuecomment-862878228, we should dynamically generate [modernBrowsers](https://github.com/nuxt/nuxt.js/blob/0742f59e2478bd3998fd4d49b6200584ee5e2721/packages/utils/src/modern.js#L4) based on updated nuxt config. Relying on external config files is anti-pattern in nuxt (since we loose single source of trust) and also it needs more boilerplate to update modern target.

pi0:
Thanks for PR @ms-fadaei <3. I think for properly solving the issue mentioned in https://github.com/nuxt/nuxt.js/pull/9337#issuecomment-862878228, we should dynamically generate [modernBrowsers](https://github.com/nuxt/nuxt.js/blob/0742f59e2478bd3998fd4d49b6200584ee5e2721/packages/utils/src/modern.js#L4) based on specified server target. Relying on external config file like `.browserlist` is an anti-pattern in nuxt (since we lose a single source of trust) and also it needs more hassle to update the modern target. Pending for @clarkdo additions.

ms-fadaei:
Thanks for your review @pi0. I will be happy to be part of developing this issue based on the Nuxt principles, so please let me know what can I do for improving this feature?

gkatsanos:
I run into this while looking at the documentation:

> The default targets of [@nuxt/babel-preset-app ](https://github.com/nuxt/nuxt.js/blob/dev/packages/babel-preset-app/src/index.js)are ie: '9' in the client build, and node: 'current' in the server build. 

Do I get this correctly that unless we use our own preset config, we're shipping IE9 compatible javascript?
So the `browserslist` config in `package.json` is ignored? 

Any examples on changing this to target more modern browsers?

danielroe:

PR #9500, opened 2021-06-30
fix(vue-app): redirect to external url replaces current history entry

### LINKED ISSUE #5032: [SPA] Redirect to external URL from asyncData replaces current history entry
### Version

### Reproduction link

[https://codesandbox.io/s/8yl9852pv9](https://codesandbox.io/s/8yl9852pv9)

### Steps to reproduce

1. Create `pages/fail.vue` with:
```
<template>
  <div>
    FAIL
  </div>
</template>
```

2. Create `pages/redirect.vue` with:

```
<script>
export default {
  asyncData({ redirect }) {
    redirect("https://elg.no");
  }
};
</script>
```

3. Create `pages/index.vue` with:

```
<template>
  <nuxt-lint to="/redirect">Navigate to external page</nuxt-link>
</template>
```

4. Navigate to /fail
5. Navigate to /
6. Click the link
7. Navigate back in history

### What is expected ?

Return to the index page where link was clicked from.

Navigates to /fail page

### Additional comments?

What happens is, when navigating in SPA mode, the code that handles redirect() call is using `location.replace` to navigate to external URL. The problem is that that happens while URL is still set to `/` so it replaces that navigation entry with external URL. Navigating back thus goes back to page visited before `/` which is unexpected.

The solution would be to either allow the URL to change `/redirect` and then trigger `location.replace` or use `location.assign` instead. Former is probably more risky as it would trigger extra routing events and potentially some more code that could possibly fail.

This is only an issue with redirects to external URLs.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c8658">#c8658</a>)</em></sub></div>

ms-fadaei:
Fix redirect to external URL replaces current history entry

## Types of changes

## Description
fix #5032

## Checklist:

### DIFF
--- a/packages/vue-app/template/utils.js
+++ b/packages/vue-app/template/utils.js
@@ -249,8 +249,8 @@ export async function setContext (app, context) {
           })
         }
         if (process.client) {
-          // https://developer.mozilla.org/en-US/docs/Web/API/Location/replace
-          window.location.replace(path)
+          // https://developer.mozilla.org/en-US/docs/Web/API/Location/assign
+          window.location.assign(path)
 
           // Throw a redirect error
           throw new Error('ERR_REDIRECT')


clarkdo (approved):

pi0 (approved):
Thanks <3

PR #9505, opened 2021-07-01
fix(utils): Router config `trailingSlash` causes error with dynamic nuxt-child routes

### LINKED ISSUE #9472: Router config `trailingSlash` causes error with dynamic nuxt-child routes
### Versions

- nuxt: tested with v2.15.6 & v2.15.7
- node: v14.16.0

### Reproduction

Following you see a still working example:

https://codesandbox.io/s/great-river-qg38y?file=/nuxt.config.js

The route structure is as following:

-- | parent.vue
-- | parent
-----| _child
-------| index.vue

### Steps to reproduce

I hope the following video helps. It shows first the working example and then the not working where the nuxt-child shows up on the parent route where it shouldn't. That happens after adding the `trailingSlash` (with `true` or `false`) option to the router config.

https://user-images.githubusercontent.com/3109921/123074295-74921880-d417-11eb-9571-b28dd905622c.mp4

Because of #9442 I added the trailingSlash router config to avoid getting the slash removed in nuxt v2.15.7. After that we recognized that this causes that dynamic nuxt-child routes stopped working correctly. We changed back to nuxt v2.15.6 but the bug was already present there.

### What is Expected?

That the dynamic nuxt-child routes work like without the router config. Means the parent route shows only the parent but the child routes show the parent including the nuxt-child.

The router `trailingSlash` option causes that the dynamic nuxt-child gets shown on the parent route.

ms-fadaei:
## Types of changes

## Description
fix #9472 
This bug existed because `trailingSlash = true` not been considered in a specific condition and exists regex missed it!

## Checklist:

### DIFF
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -56,11 +56,7 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-', 
     route.path = isChild ? route.path.replace('/', '') : route.path
     if (route.path.includes('?')) {
       if (route.name.endsWith(`${routeNameSplitter}index`)) {
-        if (trailingSlash) {
-          route.path = route.path.replace(/\?\/$/, '/')
-        } else {
-          route.path = route.path.replace(/\?$/, '')
-        }
+        route.path = route.path.replace(/\?\/?$/, trailingSlash ? '/' : '')
       }
       const names = route.name.replace(regExpParentRouteName, '').split(routeNameSplitter)
       const paths = route.path.split('/')
--- a/packages/utils/src/route.js
+++ b/packages/utils/src/route.js
@@ -56,7 +56,11 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-', 
     route.path = isChild ? route.path.replace('/', '') : route.path
     if (route.path.includes('?')) {
       if (route.name.endsWith(`${routeNameSplitter}index`)) {
-        route.path = route.path.replace(/\?$/, '')
+        if (trailingSlash) {
+          route.path = route.path.replace(/\?\/$/, '/')
+        } else {
+          route.path = route.path.replace(/\?$/, '')
+        }
       }
       const names = route.name.replace(regExpParentRouteName, '').split(routeNameSplitter)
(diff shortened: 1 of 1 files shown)


packages/utils/src/route.js
```diff
@@ -56,7 +56,11 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-',
     route.path = isChild ? route.path.replace('/', '') : route.path
     if (route.path.includes('?')) {
       if (route.name.endsWith(`${routeNameSplitter}index`)) {
-        route.path = route.path.replace(/\?$/, '')
+        if (trailingSlash) {
+          route.path = route.path.replace(/\?\/$/, '/')
```
clarkdo, line 60: The change looks good.

Just double check, will this affect logic inside below route.path.split(/) ?

packages/utils/src/route.js
```diff
@@ -56,7 +56,11 @@ function cleanChildrenRoutes (routes, isChild = false, routeNameSplitter = '-',
     route.path = isChild ? route.path.replace('/', '') : route.path
     if (route.path.includes('?')) {
       if (route.name.endsWith(`${routeNameSplitter}index`)) {
-        route.path = route.path.replace(/\?$/, '')
```
pi0, line 59: Maybe can be simplified with something like this? (not tested)

```js
route.path = route.path.replace(/\?\/?$/, trailingSlash ? '/' : '')

File: packages/utils/src/route.js (hunk 1 above)
Comment (ms-fadaei) [reply packages/utils/src/route.js:60]: No, it dosn't.

before change:
```js
[
  ':child?',
   ''
]
```

after change:
```js
[
  ':child',
  ''
]
```

it just removes `?` same as without `trailingSlash` option

packages/utils/src/route.js
ms-fadaei: Yes, it can be.
I just do that because of more readability

pi0: I think it will cover are 4 possible cases (with/without trailingSlash flag x path containing trailing/not) so would be safer too :)

pi0 (commented):
Thanks! <3

PR #9530, opened 2021-07-07
fix(vue-app): this.$nuxt.refresh doesn't refresh every fetch hooks

### LINKED ISSUE #9439: `this.$nuxt.refresh` doesn't refresh every `fetch` hooks (^2.12.0)
Hey friends 👋

## Versions

- nuxt: 2.15.7
- node: v14.15.3

### Steps to reproduce

<details>
<summary>1. Create <code>~/components/AppHeader.vue</code>, a component that uses the <code>fetch</code> hook</summary>
<br>

```html
<template>
	<header>Header from: {{ timestamp }}</header>
</template>

<script>
export default {
	fetch() {
		this.timestamp = Date.now();
	},
	fetchKey: "appHeader",
	data() {
		return {
			timestamp: 0
		};
	}
};
</script>
```

</details>

<details>
<summary>2. Update your <code>~/layouts/default.vue</code> file to use your new fancy header</summary>
<br>

```html
<template>
	<div>
		<button @click="refresh">Nuxt Refresh</button>
		<app-header />
		<nuxt />
	</div>
</template>

<script>
export default {
	methods: {
		refresh() {
			this.$nuxt.refresh();
		}
	}
};
</script>
```

</details>

<details>
<summary>3. Build your application and start it in production</summary>
<br>

```bash
yarn generate && yarn start
```

</details>

<details>
<summary>4. Fiddle with the Nuxt refresh button</summary>
<be>
Mouse clicking gif
</details>

### What is expected?

Displayed timestamp in header should be refreshed/updated.

Displayed timestamp in header remains the same.

## Context

When Nuxt refresh is called here's what happen regarding the `fetch` hook feature (vulgarized):
1. Nuxt gets the current page component;
2. Nuxt then crawls this component's children, looking for every fetch hooks;
3. Nuxt calls them.

Using our previous example, this leads to `this.$nuxt.refresh` successfully refreshing those `fetch` hooks:
```diff
  DefaultLayout
- ├── AppHeader, uses the fetch hook
  └── Nuxt (the current page)
+    ├── HeroComponent, uses the fetch hook
     ├── NewsletterComponent
     └── ArticleList
+        └── ArticleCard, uses the fetch hook
```

And as you can see, our app header's hook is left ignored 😢

Looking at the code in charge of that (https://github.com/nuxt/nuxt.js/blob/HEAD/packages/vue-app/template/App.js#L180-L193), we can clearly see that the "crawling components looking for `fetch` hooks" part starts from the page, not the layout, leading to the described-above issue.

Voilà! Let me know if anything, cheers~

ms-fadaei:
With this fix, layout fetch, layout child components fetch (including the page), and page child component fetch will be called during the Nuxt refresh helper.
## Types of changes

## Description
In the current version of the Nuxt when we call $nuxt.refresh(), the page fetch will be called if the current page has a fetch, otherwise, the child components fetch method will be called.

with this fix:
* if the layout is enabled:
  * if the layout itself has the fetch, it will be pushed to the call list
  * if the layout child components (including the page) have the fetch, they will be pushed to the call list
* if the layout is disabled:
  * if the page itself has the fetch, it will be pushed to the call list
  * if the page child components have the fetch, they will be pushed to the call list

Caution: What about nested routes? layout fetch will be inserted twice? can use `set` instead of `array` to prevent duplication?
fix: #9439 

## Checklist:


packages/vue-app/template/App.js
```diff
@@ -182,10 +182,25 @@ export default {
         if (page.$options.fetch && page.$options.fetch.length) {
           p.push(promisify(page.$options.fetch, this.context))
         }
-        if (page.$fetch) {
-          p.push(page.$fetch())
-        } else {
-          // Get all component instance to call $fetch
+        // get layout component
```
clarkdo, line 185: Thanks for the pr!

This change seems a breaking change for layouts and fetch users. @pi0 Do you think it should be enabled with an config option flag ?

packages/vue-app/template/App.js
```diff
@@ -182,10 +182,25 @@ export default {
...
+        const layout = this.$children.find((component) => component._name?.toLowerCase().includes("layouts"))
```
pi0, line 186: Refactor suggest: Maybe we can use `let root = page` and change it to parent layout instance if exists?

packages/vue-app/template/App.js
pi0: I think this change kinda makes sense since at least current intended behavior seems refreshing every fetch and layout is not (also source of call is app). /cc @Atinux @danielroe wdyt?

I guess if it is not intended behavior, we shaol simply not do this change at all and instead document layout needs manual fetch.

pi0:
> What about nested routes? layout fetch will be inserted twice? can use set instead of array to prevent duplication?

Can you please ensure about this? If is a case, we might make use of [WeakSet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) to dedup.

Also can you please check behavior when page does not having $fetch but layout does?

ms-fadaei:
> Can you please ensure about this? If is a case, we might make use of [WeakSet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) to dedup. (Note: IE support 🤷)
> 
> Also can you please check behavior when page does not having $fetch but layout does?

I checked that right now. it not working properly with nested routes (then I should refactor the code)
I checked that too and it seems everything is right!

packages/vue-app/template/App.js
danielroe: I would expect `refresh()` to refresh all fetch hooks, so I think this is a bug fix rather than a breaking change.

ms-fadaei:
@pi0 
what you thinking about `getChildrenComponentInstancesUsingFetch`
https://github.com/nuxt/nuxt.js/blob/HEAD/packages/vue-app/template/utils.js#L56-L68

If a child has the fetch, children of it will not be considered!

ms-fadaei:
And I realized the sequence call of fetch (first parent then child) is not guaranteed! (in the current version this is just about nested routes because of limited fetch calls but in my version, this occur in every parent-child relations)

ms-fadaei:
@pi0 @clarkdo I'm done with this PR.

We still have two edge cases that I mentioned above with refresh handler and they aren't new! they exist from the last versions

packages/vue-app/template/App.js
lihbr: I think this also makes sense to only crawl everything once, also keeping `getChildrenComponentInstancesUsingFetch` simple. What was your opinion on that @ms-fadaei?

lihbr:
Hey friends, what's up on this one?
What remains to be changed/discussed to lead to a merge? or is it just awaiting a review?

Cheers!

danielroe (commented):
I think the main thing we still need to handle is:
- [ ] ensure that we don't duplicate fetches (for example if a component with fetch is included in both a page/layout) - see https://github.com/nuxt/nuxt.js/pull/9530#issuecomment-875922938 for one approach

lihbr:
I'll try to have a look at the last bit @danielroe mentioned this week 💪 Keeping you up-to-date!

lihbr:
OK, I'm having a bit of trouble understanding the "duplicate fetch" maybe issue mentioned above.

Looking at the changes, 3 things have been achieved:

1. `getChildrenComponentInstancesUsingFetch()` has been modified to handle an `exclude` array, stopping the crawling if a node from `exclude` is encountered during the crawling process;
2. On the `pages.map()`, the callback now loop only over the children component of the current page, theoretically deduplicating redundant calls (? not 100% confident on that one, but that's what I understand);
3. After the `pages.map()`, the `refresh()` method now looks for a layout, if one it adds its own fetch and crawls its children component, excluding pages and their offspring.

Taking all of those changes individually they seem pretty straightforward to me. Basically, the `refresh()` method now achieve that kind of resolution for fetch hooks (I had budget for an animation today):

refresh

I don't really see how we can end up with duplicate calls here, or at least how we can end up with duplicate calls we didn't have already before(?)

Now, maybe where I'm wrong is on my understanding of the `getMatchedComponentsInstances()` function. To me, that function gets each page component actually displayed on the current route, so from my gif above `~/pages/blog.vue` & `~/pages/blog/index.vue`? I'm not sure about that function behavior as it seems to access properties of Vue Router current route which I'm not aware of/are Nuxt specific(?): https://router.vuejs.org/api/#route-object-properties

If I'm wrong here and there's definitely something leading to duplicated behaviors, I'm not sure however a set would solve the issue? Or maybe it would? I'd be afraid only some of the deduplicated hooks get called and that the others remain stale (a.K.a. no new data?)

Let me know!

pi0:
@ms-fadaei @lihbr Can you please check the latest `nuxt-edge`? I believe #10489 should address the issues by refetching on all children. If not happy to reopen and improve 👍🏼

lihbr:
Hey there,

Unfortunately, this doesn't appear to fix the issue as the component-crawling function still starts from the page component instead of the layout.

Maybe I'm doing something wrong, here's my repro on edge: [nuxt_nuxt.js_9530_1.zip](https://github.com/nuxt/nuxt.js/files/8974462/nuxt_nuxt.js_9530_1.zip)

pi0:
Thanks for checking @lihbr. @ms-fadaei Would you please rebase your branch to dev?

packages/vue-app/template/App.js
```diff
@@ -214,6 +221,24 @@ export default {
 
         return Promise.all(p)
       })
+
+      <% if (features.fetch) { %>
```
lihbr, line 225: Should this be considered again here? From #10489

https://github.com/nuxt/nuxt.js/pull/10489/files#diff-45fb7db1f2da36defea7d4cf1f2a5f4db0a8f9f320b889808f204e51e4df8259R201-R202

danielroe:
My apologies for the long delay on reviewing this. We're approaching Nuxt v2 end-of-life. Is this PR still needed?

PR #9539, opened 2021-07-11
fix(vue-app): re-register components construtor in HMR

### LINKED ISSUE #8501: AsyncData isn't merged when triggered by watchQuery after HMR in Dev

clarkdo:
## Types of changes

## Description

Fix https://github.com/nuxt/nuxt.js/issues/8501
closes https://github.com/nuxt/nuxt.js/pull/9077

We can see component instance and Component are not matching after hot reload, this pr is re-assigning correct constructor to component instance.

## Checklist:

### DIFF
--- a/packages/vue-app/template/client.js
+++ b/packages/vue-app/template/client.js
@@ -738,9 +738,8 @@ function hotReloadAPI(_app) {
 
   if (_app.context.isHMR) {
     const Components = getMatchedComponents(router.currentRoute)
-    const instances = getMatchedComponentsInstances(router.currentRoute)
-    instances.forEach((instance,index) => {
-      instance.constructor = Components[index]
+    Components.forEach((Component) => {
+      Component.prototype.constructor = Component
     })
   }
 }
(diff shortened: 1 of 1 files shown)


ghost (dismissed):

danielroe (approved):
LGTM 👍

PR #9574, opened 2021-07-20
fix(generator): allow passing `builder` to `getGenerator`

mrazauskas:
## Types of changes

## Description
It seems like constructor of `Generator` class has to receive two arguments (`nuxt` and `builder`). Currently `getGenerate` helper is passing only `nuxt` to the class. Therefor calling `getGenerate(nuxt)` throws 'TypeError: Cannot read property 'forGenerate' of undefined...'. It originates [from this line](https://github.com/nuxt/nuxt.js/blob/8e725d34b3ffbbf5b4a94e1b86f45e6ffcc61f3d/packages/generator/src/generator.js#L88), because `this.builder` is undefined.

## Checklist:

`getGenerate` is rather minor and obvious function. I can’t spot it in the test suite. Hope it is acceptable to keep it untested.


mrazauskas:
@pi0 Thanks for the review.

Fixed it. `getGenerator` now is slim and has similar API as Generator class. That’s better solution, right?

Will send separate PR for `if (this.builder)` check.

pi0 (approved):
Thanks!

PR #9629, opened 2021-08-04
feat(vue-app): pass `store` to `createRouter`

### LINKED ISSUE #9594: Get the store instance in the Vue router createRouter function
Hi, 

I'm currently using the router plugin (https://www.npmjs.com/package/@nuxtjs/router) in order to do the following behavior: 
- I'm importing several modules on the fly to compose my router regarding the user context. For example, a US user will have different routes available than a IT user. 
- I'm also adding conditions inside of the router to enable or not some routes depending on feature flags, which are in my Vuex store. 

Here is actually the problem: I can't have any access to the Vuex store at this point. Since I already have it (for sure) instantiated before, I wonder if this is possible to retrieve the store object from the `createRouter` method in order for me to support my conditional imports.

https://github.com/nuxt/nuxt.js/blob/2ec62617ced873fef97c73a6d7aa1271911ccfd5/packages/vue-app/template/router.js#L104

I guess this is the place to look at, but it's weird to see the router initialized before the store here 🤔. Do you think it would have a big impact to create the store beforehand?
https://github.com/nuxt/nuxt.js/blob/eb6bd9febc63146622f56f9cb20511510400e6c5/packages/vue-app/template/index.js#L94-L105

Thanks!

loick:
Resolves #9594

If this PR is validated, I will also create PR on the `@nuxt/router` module in order to support this parameter as well: 
https://github.com/nuxt-community/router-module/blob/18bb2d39b55a9da8abec0a10a9f4552291dc0b39/templates/plugin.js#L10-L11

## Types of changes


packages/vue-app/template/index.js
```diff
@@ -92,10 +92,10 @@ function registerModule (path, rawModule, options = {}) {
 <% } %>
 
 async function createApp(ssrContext, config = {}) {
-  const router = await createRouter(ssrContext, config)
+  const store = <%= store ? %> createStore(ssrContext) <% : undefined %>
+  const router = await createRouter(ssrContext, config, store)
```
loick, line 96: Just wondering here. We did not pass the context to the `createRouter` function, is it a different context expected than the one provided to the createApp method?

packages/vue-app/template/router.js
```diff
@@ -101,7 +101,7 @@ export const routerOptions = {
   fallback: <%= router.fallback %>
 }
 
-export function createRouter (ssrContext, config) {
+export function createRouter (ssrContext, config, store) {
```
clarkdo, line 104: It seemed the new parameter is not used anywhere, can you provide more details about the usage of the store parameter?

packages/vue-app/template/index.js
pi0: > Is it a different context expected than the one provided to the createApp method?

Yes. ssrContext is different than nuxt context (and also slightly from app one).

I think would be safer to pass store as an object to avoid adding 4th param if we had any similar requirement in the future:

```suggestion
  const router = await createRouter(ssrContext, config, { store })
```

packages/vue-app/template/router.js
pi0: @clarkdo It seems usable for custom integrations with nuxt/router module. Indeed we can simply omit param here as is not used by default nuxt router.
```suggestion
export function createRouter (ssrContext, config) {
```

PR #9668, opened 2021-08-13
chore(utils): improve `stripWhitespace` utility

mrazauskas:
## Types of changes

## Description

This is a minor fix. I was playing with plugin templates and found that `stripWhitespace` utility is not removing empty lines if these appear at the beginning of a file. For instance, empty lines will be left if a template begins like this one:

```js
<% if (options.debug) { %>
import { debug } from '<%= options.import.path %>'
<% } %>
// rest of the file
```

A regexp was added to remove empty lines if they appear at the beginning of a file. `g`, global search flag seemed to be unnecessary, because globally there is just one beginning (or end) of a string. Or I missed something important? `g` is not included in the new regexp. For consistency I removed it in the regexp which was targeting the end of a string. (Alternatively `g` flag could be added to both.)

#### Test

After the change, the `expect` with `'\n\n\n\f\r\f}'` string was failing. I improved this and other tests by adding `foo` before (or `bar` after) `\n` as if these samples are taken from the middle of a string. Also after `strip multiple blank lines` regexp there must be only double `\n\n`s left in the string, so two last `expect`s simply checks double `\n\n`s.

## Checklist:

### DIFF
--- a/packages/utils/src/lang.js
+++ b/packages/utils/src/lang.js
@@ -29,7 +29,8 @@ const WHITESPACE_REPLACEMENTS = [
   [/{\n{2,}/g, '{\n'], // strip start padding from blocks
   [/\n{2,}([ \t\f\r]*})/g, '\n$1'], // strip end padding from blocks
   [/\n{3,}/g, '\n\n'], // strip multiple blank lines (1 allowed)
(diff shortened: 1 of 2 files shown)


clarkdo (approved):

pi0 (approved):
Thanks!

PR #9701, opened 2021-08-22
fix(utils): nuxt.config.js is not in cwd

### LINKED ISSUE #9640: /nuxt.config.js is not in cwd
## Greetings,

While I am trying to generate the NUXT app I encountered an issue. 

bug

### Versions
-nuxt : 2.15.7
-node : v14.17.1
-npm :  7.20.1

harishsg99:
## Types of changes

## Description
this resolves https://github.com/nuxt/nuxt.js/issues/9640

## Checklist:


clarkdo (dismissed):

packages/cli/src/utils/generate.js
```diff
@@ -169,7 +169,7 @@ export async function snapshot ({ globbyOptions, ignore, rootDir }) {
   const files = await globby('**/*.*', {
     ...globbyOptions,
     ignore,
-    cwd: rootDir,
+    cwd: path.normalize(rootDir),
```
pi0, line 172: I think it would be better normalizing the path directly in `@nuxt/config` ([here](https://github.com/nuxt/nuxt.js/blob/3c10f176e69b29a88daaae6fb353b9c5fc198031/packages/config/src/options.js#L70)) otherwise we will have inconsistent behavior.

danielroe: I would also check whether it works with `upath.normalize` as that is what we use elsewhere.

pi0: One downside of upath, is that it seems not supporting windows network mounts (that might be currently used by nuxt2 users). If `path.normalize` solving issue with `globby` is less risky change (till we have `unpath` for better normalization)

danielroe:
@harishsg99 Would you confirm whether this change is still needed on the latest 2.16.2 (which has a newer version of globby than when you submitted this PR)?

danielroe:
We're approaching end-of-life for Nuxt v2, and I haven't heard back, so I'm going to close this PR. But let me know if this change is still needed. 🙏

PR #9762, opened 2021-09-05
feat(cli): early return on static generation with --fail-on-error

### LINKED ISSUE #9704: Fails immedeately when `nuxt generate --fail-on-error` encounters first error

Static Generation with option `--fail-on-error` fails with error after all routes generation even with 10,000+ routes.
I think this behavior wastes the rest of time after the first error.
It'd be better if it fails when the first error occurs.

I think it'd be better if it can do `nuxt generate --fail-on-error` fails with the first error.
But considering of backward compatibility, the new CLI option like `--fail-on-first-error`
or add new config option like `failOnFirstError`.

aewshyae:
static generation with command `nuxt generate --fail-on-error` exits when the first error occurs.

## Types of changes

## Description

The command `nuxt generate --fail-on-error` returns error after all routes generation has completed even with 10,000+ routes.
However even if only 1 error occurs with any other generation succeeded the command will fail.
Thus, the command should exit early when the first error occurs.

In detail I have changed to return after the concurrency encountered the first error.
This change will save the time of user.(Resolves: #9704)

## Checklist:

I've tried to execute `nuxt generate --fail-on-error` with my commits locally but I can't and don't know how to do it.
any comments are appreciated!


clarkdo (commented):
Looks we already have fail-on-errror which is throwing error after generation, so this looks a slightly breaking change, what about supporting number on fail-on-error for specifying number of errors, then your requirement will be done by fail-on-error=1

aewshyae:
@clarkdo thanks for your suggestion! I'll try to implement it.

BTW, my previous commits comments are not semantic and occurs CI error.
so I close this PR and re-issue with new implement.

PR #9858, opened 2021-09-29
feat(vue-app): forward custom props from `<nuxt>` to the page component

matthieusieben:
## Types of changes

## Description

The following will apply a `foo` class to the route's root element:
```html
<nuxt class="foo" />
```

However, if the page's component declares `props`, it is not possible to pass them along:
```html
<nuxt my-prop="foo" />
```

It *does* work if we use `<nuxt-child>` directly from within the the layout, but then we loose the error handling logic.
```html
<nuxt-child my-prop="foo" />
```

This change allows defining page `props` through the `<nuxt>` component.

## Checklist:


atinux (dismissed):

danielroe (dismissed):

matthieusieben:
Adapted last commit to also forward `$scopedSlots`  and `$listeners`.

pi0 (approved):

pi0:
Love this solution!

PR #9956, opened 2021-10-17
feat(cli): early return on static generation by error count

### LINKED ISSUE #9704: Fails immedeately when `nuxt generate --fail-on-error` encounters first error

aewshyae:
improve https://github.com/nuxt/nuxt.js/pull/9762 by [review](https://github.com/nuxt/nuxt.js/pull/9762#pullrequestreview-746625291)

## Types of changes

## Description
The command `nuxt generate --fail-on-error` returns error after all routes generation has completed even with 10,000+ routes.
However even if only 1 error occurs with any other generation succeeded the command will fail.
Thus, the command need to the way to exit early when the first error occurs.

In detail I have changed to return after the concurrency occurs specified counts of error.

This change will save the time of user.(Resolves: #9704)

## Checklist:

- [x] My change requires a change to the documentation. ()
- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/docs/pull/2135)

### DIFF
--- a/packages/generator/src/generator.js
+++ b/packages/generator/src/generator.js
@@ -47,7 +47,7 @@ export default class Generator {
     }
   }
 
-  async generate ({ build = true, init = true } = {}) {
+  async generate ({ build = true, init = true, failOnError = false } = {}) {
     consola.debug('Initializing generator...')
     await this.initiate({ build, init })
 
@@ -55,7 +55,7 @@ export default class Generator {
     const routes = await this.initRoutes()
 
     consola.info('Generating pages' + (this.isFullStatic ? ' with full static mode' : ''))
-    const errors = await this.generateRoutes(routes)
+    const errors = await this.generateRoutes(routes, failOnError)
 
     await this.afterGenerate()
 
@@ -168,7 +168,7 @@ or disable the build step: \`generate({ build: false })\``)
     return requireModule(path.join(this.options.buildDir, 'routes.json'))
   }
 
-  async generateRoutes (routes) {
+  async generateRoutes (routes, failOnError = false) {
     const errors = []
 
     this.routes = []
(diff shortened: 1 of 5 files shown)


aewshyae:
any reviews appreciated. thanks.

danielroe (dismissed):
This looks good to me. wdyt @pi0?

packages/generator/src/generator.js
```diff
@@ -184,6 +187,9 @@ or disable the build step: \`generate({ build: false })\``)
     // Start generate process
     while (this.routes.length) {
       let n = 0
+      if (typeof failOnError === 'number' && errors.length >= failOnError) {
```
pi0, line 190: I think would be best if we support this as an additional argument for type safety and better CLI ags definition (`--max-fails` or something similar)

aewshyae: Thanks for your review.

I'm agree with your idea, so we can leave `--fail-on-error` option as it is.
How about `--max-errors` instead of `--max-fails` because  `--fail-on-error` means generations fail when error occurs.

danielroe: Yes, that's good wording 👍

aewshyae: Thx! I'll get work on it this weekend👍

also fix https://github.com/nuxt/docs/pull/2135 too

aewshyae:
@pi0 @danielroe I replaced `--fail-on-error=[num]` with `--max-errors=[num]` but the GitHub Action `test / test-dev` has failed I don't know why it happens.
Would you mind helpling me to solve it?

danielroe:
My apologies for never getting back to you.

We're approaching end-of-life for Nuxt v2 - and as this is a feature rather than a bug fix I think I'm going to close it. Let me know if it's still a high-priority feature for you and I'll happily rethink. 🙏

PR #10001, opened 2021-10-31
feat: support multple dotenv files

leon0399:
## Types of changes

## Description

This feature allows to have multiple dotenv files in projects. This will not only allow developers more flexibility in customizing their environment, but also provide parity between Nuxt and [Vue CLI](https://cli.vuejs.org/guide/mode-and-env.html#environment-variables) or [Vite](https://vitejs.dev/guide/env-and-mode.html#env-files), since they already has such a feature. By default, it will scan for following files:

```
.env                # loaded in all cases
.env.local          # loaded in all cases, ignored by git
.env.[mode]         # only loaded in specified mode
.env.[mode].local   # only loaded in specified mode, ignored by git
```

## Checklist:

- [x] I have updated the documentation accordingly. (PR: https://github.com/nuxt/nuxtjs.org/pull/1905)


pi0:
Hi @leon0399 and thanks for PR. Despite that vite and vue-cli support this convention, it is [not a good practice](https://github.com/motdotla/dotenv#should-i-have-multiple-env-files) to use multiple `.env` files in your setup. You probably can instead set dev defaults in `nuxt.config` (runtimeConfig)

> No. We strongly recommend against having a "main" .env file and an "environment" .env file like .env.test. Your config should vary between deploys, and you should not be sharing values between environments.
> In a twelve-factor app, env vars are granular controls, each fully orthogonal to other env vars. They are never grouped together as “environments”, but instead are independently managed for each deploy. This is a model that scales up smoothly as the app naturally expands into more deploys over its lifetime.

PR #10132, opened 2021-12-13
fix(vue-app): call ssrContext.unsetMutationObserver only if it exists

### LINKED ISSUE #10131: calling `error` from a middleware generates a TypeError in full static mode
### Versions

- nuxt: v2.15.0
- node: any

### Reproduction

[https://github.com/reegodev/nuxt-middleware-error](https://github.com/reegodev/nuxt-middleware-error)

### Steps to reproduce
- Enable full static mode
- Enable Vuex
- Create a middleware that calls the `error` function exposed through the context
- Run `yarn generate`

### What is Expected?
Route generation should fail with the error exposed through the `error` call

Route generation fails with `TypeError: ssrContext.unsetMutationObserver is not a function`.

### Additional info
Although this does not sound like a big deal, when using Bridge + Nitro and edge rendering, if the middleware calls the `error` function the page returns a 500 error instead of our standard error page.

reegodev:
## Types of changes

## Description
Resolves: #10131
`ssrContext.unsetMutationObserver` is set after router middleware are evaluated. This causes a problem because if a middleware calls `error`, then the app calls `renderErrorPage` which in turns calls `ssrContext.unsetMutationObserver()` in the `beforeRender` function, causing an error.

[https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L137](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L137)
[https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L209](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#209)
[https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L216](https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/server.js#L216)

## Checklist:

### DIFF
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -134,7 +134,10 @@ export default async (ssrContext) => {
 
       <% if (isFullStatic && store) { %>
       // Stop recording store mutations
-      ssrContext.unsetMutationObserver && ssrContext.unsetMutationObserver()
+      // unsetMutationObserver is only set after all router middleware are evaluated
+      if (ssrContext.unsetMutationObserver) {
+        ssrContext.unsetMutationObserver()
+      }
       <% } %>
     }
   }
--- a/packages/vue-app/template/server.js
+++ b/packages/vue-app/template/server.js
@@ -134,7 +134,7 @@ export default async (ssrContext) => {
 
       <% if (isFullStatic && store) { %>
       // Stop recording store mutations
-      ssrContext.unsetMutationObserver()
+      ssrContext.unsetMutationObserver && ssrContext.unsetMutationObserver()
       <% } %>
     }
   }


packages/vue-app/template/server.js
```diff
@@ -134,7 +134,7 @@ export default async (ssrContext) => {
 
       <% if (isFullStatic && store) { %>
       // Stop recording store mutations
-      ssrContext.unsetMutationObserver()
+      ssrContext.unsetMutationObserver && ssrContext.unsetMutationObserver()
```
danielroe, line 137:
```suggestion
      // unsetMutationObserver is only set after all router middleware are evaluated
      if (ssrContext.unsetMutationObserver) {
        ssrContext.unsetMutationObserver()
      }
```

danielroe (approved):

pi0 (approved):
Thanks

PR #10193, opened 2022-01-06
fix(core): accidentally adding an extra comma silently disables all modules

IvanEh:
I have recently been working on vuestorefront and found an issue with their public repository https://github.com/vuestorefront/template-commercetools (pull request created). There was an extra comma that made the application to infinitely loop because the server middleware wasn't handling the request due to a missing nuxt module.

The small typo led to an embarrassingly frustrating result that was not easy to solve. 

With my change, I'm skipping the handler validation if there was no handler which seems very sensible to me

## Types of changes

## Checklist:


clarkdo (dismissed):

danielroe (dismissed):

packages/core/src/module.js
```diff
@@ -149,7 +149,7 @@ export default class ModuleContainer {
     }
 
     // Prevent adding buildModules-listed entries in production
-    if (this.options.buildModules.includes(handler) && this.options._start) {
+    if (handler && this.options.buildModules.includes(handler) && this.options._start) {
```
pi0, line 152: We are silently ignoring an important issue here. If handler is not resolved at this point, at least a warning should be shown (it is not same situation of using a build-module but a module not loaded from config because of typo)

pi0 (changes requested):

danielroe:
We're approaching end-of-life for Nuxt v2.

If you want to move ahead, I think we need to at least show a warning as suggested by @pi0. Let me know if that's something you want to investigate 🙏

PR #10380, opened 2022-04-13
fix: catch sessionStorage exception

DKhalil:
One of the things that pop up from time to time is a QuotaExceededError, which would be caught by this


packages/vue-app/template/utils.js
```diff
@@ -162,8 +162,10 @@ export function resolveRouteComponents (route, fn) {
 
             // check for previous reload time not to reload infinitely
             if (!previousReloadTime || previousReloadTime + 60000 < timeNow) {
-              window.sessionStorage.setItem('nuxt-reload', timeNow)
-              window.location.reload(true /* skip cache */)
+              try {
+                window.sessionStorage.setItem('nuxt-reload', timeNow)
+                window.location.reload(true /* skip cache */)
+              } catch (sessionstorageError) {}
```
DraftProducts, line 168: I propose to remove error arg since it's not used.
```suggestion
              } catch {}
```

DraftProducts (changes requested):

packages/vue-app/template/utils.js
```diff
@@ -162,8 +162,10 @@ export function resolveRouteComponents (route, fn) {
 
             // check for previous reload time not to reload infinitely
             if (!previousReloadTime || previousReloadTime + 60000 < timeNow) {
-              window.sessionStorage.setItem('nuxt-reload', timeNow)
-              window.location.reload(true /* skip cache */)
+              try {
+                window.sessionStorage.setItem('nuxt-reload', timeNow)
+                window.location.reload(true /* skip cache */)
+              } catch {}
```
gkatsanos, line 168: dont you want to log this?

PR #10489, opened 2022-06-09
fix(vue-app): preview mode fetch

mdoesburg:
## Types of changes

## Description

Changes:

**1. Make sure all `fetch` hooks get called properly in preview mode**

Prior to this change, if a page component contained a `fetch` hook, all underlying component `fetch` hooks would not be called in preview mode. In addition to that, if a component contained a `fetch` hook and sub components of the component also contained a `fetch` hook, only the root component `fetch` hook would be called in preview mode. 

**2. Make sure that potential `asyncData` and old `fetch` hooks resolve prior to any new `fetch` hook calls in preview mode in accordance with the [Nuxt lifecycle diagram](https://nuxtjs.org/docs/concepts/nuxt-lifecycle#nuxt-lifecycle)**

Prior to this change, if you dynamically rendered components based on data from the `asyncData` hook, and the dynamically rendered components contained a `fetch` hook, those `fetch` hooks would not be called, because `asyncData`, old `fetch`, and new `fetch` were all resolving in parallel in preview mode.

Resolves: 
- https://github.com/nuxt/nuxt.js/issues/10483
- https://github.com/nuxt/nuxt.js/issues/8172

## Checklist:


atinux (dismissed):
Looks good to me, will need a second eye on this from @danielroe

danielroe (dismissed):
This looks good to me - nice changes.

The one question I have is about the comment on https://github.com/nuxt/nuxt.js/blob/0742f59e2478bd3998fd4d49b6200584ee5e2721/packages/vue-app/template/utils.js#L61 but if you're happy with this @Atinux then I am too.

Atinux:
I think it's fine to reload the templates when we do a full refresh on preview mode. @pi0 do you want to take a look at it or happy to merge and let people use it with nuxt-edge?

packages/vue-app/template/App.js
```diff
@@ -203,6 +196,20 @@ export default {
           )
         }
         <% } %>
+
+        <% if (features.fetch) { %>
+        // Wait for asyncData & old fetch to finish
+        await Promise.all(p)
```
pi0, line 202: Does it make sense to empty all resolved promises after this line to cleanup refs?

pi0 (dismissed):
Generally looks good to me. We are introducing a behavior change that but it makes lifecycle more explicit can only slightly affect performance in cases usage was depending on the parallel run.

pi0:
Looks good to me. Pending for #10510 before merging new PRs.

packages/vue-app/template/App.js
pi0: The performance point is memory management. Promises are stored as link lists. As fast as you free them up, garbage collector can empty the dependencies from memory. Still not huge only a nitpick :)

mdoesburg: Ah, makes sense! Learned something new today. Should be addressed in the most recent commit. Let me know if this can be resolved or if you have any other improvements/suggestions etc. :)

pi0 (approved):
LGTM

PR #10577, opened 2022-07-18
Change to render client-side + Update versions to latest stable + Remove unnecessary code + Add basic enhancements

swalx:
## Types of changes

## Description
We realize that whenever the page loads/reloads the icons takes a second to load while the rest of the content is still present. This example uses SSR by default, the page renders on the server but the icons don't.
The problem this solves is better user experience, by setting the app to render client-side the page content and icons loads correctly. Additionally, UIkit has other useful functions: the global $uikit variable will allow one to use these functions across the entire application. 

We have been trying to develop a solution to get this to work properly with SSR but have not been able to efficiently accomplish this. The best we could do is this work-a-round template [nuxt-uikit](https://github.com/edu-fedorae/nuxt-uikit/tree/ssr).

## Checklist:
- There seems to be none at the moment.
- N/A: The changes added does not affect Nuxt itself. Additionally, the example project was ran without any unusual side effects.

### DIFF
--- a/examples/uikit/nuxt.config.js
+++ b/examples/uikit/nuxt.config.js
@@ -1,6 +1,7 @@
 export default {
+  ssr: false,
   css: ['uikit/dist/css/uikit.css'],
   plugins: [
-    { src: '~/plugins/uikit.js', ssr: false }
+    { src: '~/plugins/uikit.js' }
   ]
 }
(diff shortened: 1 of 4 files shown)


pranavkumar389 (approved):

danielroe (commented):
I do apologise for the delay in reviewing this.

On its own, this PR does not result in a working example, and I'm not familiar enough with uikit to propose a solution.

I'm closing the PR for now given the length of time it's been waiting, but just let me know if you want to update the PR; if so, I'll happily reopen.

PR #10734, opened 2022-10-02
ci: add minimum GitHub token permissions for workflows

ashishkurmi:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using [secure-workflows](https://github.com/step-security/secure-workflows).

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes. Here is an example of the permissions in one of the workflow runs:
https://github.com/nuxt/nuxt.js/actions/runs/3169583990/jobs/5161596015#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflows:

- nightly.yml
- test.yml
- windows.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- [GitHub recommends](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) defining minimum GITHUB_TOKEN permissions.
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron:  '0 0 * * *'
 
+permissions:
+  contents: read
+
 jobs:
   nightly:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,9 @@ on:
       - dev
       - 2.x
 
+permissions:
+  contents: read
+
 jobs:
   setup:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -10,6 +10,9 @@ on:
       - dev
       - 2.x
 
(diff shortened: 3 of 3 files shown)


TheAlexLichter (approved):

danielroe:
This is a helpful prompt but I think we should instead enable restrictive permissions by default for the `nuxt` organisation rather than adding these lines to every repo workflow file we have.

See https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#setting-the-permissions-of-the-github_token-for-your-organization.

CleanShot 2022-10-03 at 10 15 21

If this causes any problems we can then grant more explicit permissions in individual repo workflows.

PR #10878, opened 2023-01-04
Revert "fix(webpack): allow transpiling package in nested `node_modules`"

ghost:
Reverts nuxt/nuxt.js#8205


danielroe:
Would you explain more about your motivations for this PR? 🙏

PR #15884, opened 2023-01-19
fix(kit): resolve group syntax of `ignore`

### LINKED ISSUE #12352: Plugin test files get autoloaded
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `yarn@1.22.17`
- Builder:          `vite`
- User Config:      `components`, `modules`, `css`, `vite`
- Runtime Modules:  `@pinia/nuxt@0.4.6`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ayjwvj?file=plugins%2Fmy-plugin.spec.js

### Describe the bug

As plugins are automatically registered, it seems to be autoloading test files as well as the plugins themselves. For example:

```
plugins/my-plugin.js <-- plugin file
plugins/my-plugin.spec.js <-- test file for plugin
```

This throws an error as the test files will be including functions from test runners

Is it possible to exclude `.(spec|test).js` files from being automatically registered?

### Additional context

_No response_

### Logs

_No response_

TheAlexLichter:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/12352

### ❓ Type of change

### 📚 Description

As [node-ignore](https://github.com/kaelzhang/node-ignore/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) doesn't handle the curly group glob syntax (`{js,ts}`) because it is not part of the [gitignore spec](https://git-scm.com/docs/gitignore), this PR introduces a function which resolves the group syntax before passing the rules to `node-ignore`.
Of course, including tests for the resolve function and a test that verifies it is integrated correctly into Nuxt

### 📝 Checklist


manniL:
The problem is that `ignore` can't handle `{}` rules properly it seems...

* `'**/*.{spec,test}.{js,ts,jsx,tsx}'` - `{ ignored: false, unignored: false }`
* `  '**/*.spec.{js,ts,jsx,tsx}'` - `{ ignored: false, unignored: false }`
* `  '**/*.spec.js'` - `{ ignored: true, unignored: false }`

Test script

```
import ignore from 'ignore'
const rules = [
  '**/*.spec.{js,ts,jsx,tsx}'
]
const manager = ignore()
manager.add(rules)
console.log(manager.test('plugins/my-plugin.spec.js'))
```

manniL:
CI is still failing for unknown reasons 🙈 Help appreciated 🙏🏻

packages/kit/src/ignore.ts
```diff
@@ -32,3 +34,29 @@ export function isIgnored (pathname: string): boolean {
...
+
+/**
+ * This function turns string containing groups '**\/*.{spec,test}.{js,ts}' into an array of strings.
+ * For example will '**\/*.{spec,test}.{js,ts}' be resolved to:
+ * ['**\/*.spec.js', '**\/*.spec.ts', '**\/*.test.js', '**\/*.test.ts']
+ *
+ * @param group string containing the group syntax
+ * @returns {string[]} array of strings without the group syntax
+ */
+export function resolveGroupSyntax (group: string): string[] {
```
pi0, line 48: We should update [index.ts](https://github.com/nuxt/nuxt/blob/ae9a98b942fc43f92ed939c3b96d275f3be43907/packages/kit/src/index.ts#L15) to use named exports from this file (excpluding internal util) to avoid exposing internals to final package.

manniL: Agreed 👍🏻

danielroe (approved):
Thank you ❤️

PR #18411, opened 2023-01-20
fix(nuxt): use query for hashing the fetch key

alanpoulain:
### 🔗 Linked issue

Sorry there is no issue but I can create it if needed.

### ❓ Type of change

### 📚 Description

`params` is an alias for `query`. If `query` is used instead of `params`, the key for fetch will not change if the query param changes.
It results in SSR issues (for instance a page change (with a `page` query param) will not load the current page data).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -47,7 +47,7 @@ export function useFetch<
   arg2?: string
 ) {
   const [opts = {}, autoKey] = typeof arg1 === 'string' ? [{}, arg1] : [arg1, arg2]
-  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof request === 'string' ? request : '', unref(opts.params), unref(opts.query)])
+  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof request === 'string' ? request : '', unref(opts.params || opts.query)])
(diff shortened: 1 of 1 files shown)


pi0 (approved):

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -47,7 +47,7 @@ export function useFetch<
   arg2?: string
 ) {
   const [opts = {}, autoKey] = typeof arg1 === 'string' ? [{}, arg1] : [arg1, arg2]
-  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof request === 'string' ? request : '', unref(opts.params)])
+  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof request === 'string' ? request : '', unref(opts.params), unref(opts.query)])
```
pi0, line 50: Alternatively, we could also use || to reduce hash complexity (both shouldn't be in same options!

```suggestion
  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof request === 'string' ? request : '', unref(opts.params || opts.query)])
```

alanpoulain: Sure, do you want me to change it?

danielroe (approved):
Nice catch, thanks!

PR #18416, opened 2023-01-21
fix(nuxt): provide types for modules as array

### LINKED ISSUE #18415: Modules as arrays don't take advantage of ModuleOptions
### Describe the feature

Nuxt configuration allows modules to either be passed as a `NuxtModule` function, a string (usual case), and an array (tuple technically) with the first element as either of the two mentioned and the second as options - quite similar to many other configuration systems like ESLint, etc. However, the schema would only give the type definition if the module option is added as a property instead.

```ts
  modules: [
    '@nuxtjs/tailwindcss', // OK
    ['@nuxtjs/web-vitals', { } ], // no types 🙁
  ],
  tailwindcss: { exposeConfig: true },
});
```

### Additional information

### Final checks

ineshbose:
### 🔗 Linked issue

resolves #18415 

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Extends `modules` definition in `.nuxt/types/schema.d.ts`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -122,19 +122,24 @@ export const schemaTemplate: NuxtTemplate<TemplateContext> = {
 
     const relativeRoot = relative(resolve(nuxt.options.buildDir, 'types'), nuxt.options.rootDir)
     const moduleMap: Record<string, any> = {}
+    const getImportName = (name: string) => name.startsWith('.') ? './' + join(relativeRoot, name) : name
+    const getSchemaProps = (meta: any) => {
+      const [configKey, importName] = [genString(meta.configKey), getImportName(meta.importName)]
+      moduleMap[importName] = configKey
+
+      return { configKey, importName };
+    }
 
     return [
       "import { NuxtModule } from '@nuxt/schema'",
       "declare module '@nuxt/schema' {",
       '  interface NuxtConfig {',
       ...moduleInfo.filter(Boolean).map((meta) => {
-        const configKey = genString(meta.configKey)
-        const importName = meta.importName.startsWith('.') ? './' + join(relativeRoot, meta.importName) : meta.importName
-        moduleMap[importName] = configKey
+        const { configKey, importName } = getSchemaProps(meta)
 
         return `    [${configKey}]?: typeof ${genDynamicImport(importName, { wrapper: false })}.default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>`
       }),
-      Object.keys(moduleMap).length > 0 ? `    modules?: (NuxtModule | string | ${Object.keys(moduleMap).join(' | ')} | [NuxtModule | string, Record<string, any>] | ${Object.entries(moduleMap).map(([importName, configKey]) => `[${genString(importName)}, NuxtConfig[${genString(configKey)}]]`).join(' | ')})[]` : '',
+      Object.keys(moduleMap).length > 0 ? `    modules?: (NuxtModule | string | ${Object.keys(moduleMap).map(m => genString(m)).join(' | ')} | [NuxtModule | string, Record<string, any>] | ${Object.entries(moduleMap).map(([importName, configKey]) => `[${genString(importName)}, NuxtConfig[${genString(configKey)}]]`).join(' | ')})[]` : '',
       '  }',
       generateTypes(await resolveSchema(Object.fromEntries(Object.entries(nuxt.options.runtimeConfig).filter(([key]) => key !== 'public'))),
         {
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -121,25 +121,17 @@ export const schemaTemplate: NuxtTemplate<TemplateContext> = {
     })).filter(m => m.configKey && m.name && !adHocModules.includes(m.name))
 
     const relativeRoot = relative(resolve(nuxt.options.buildDir, 'types'), nuxt.options.rootDir)
-    const moduleMap: Record<string, any> = {}
-    const getImportName = (name: string) => name.startsWith('.') ? './' + join(relativeRoot, name) : name
-    const getSchemaProps = (meta: any) => {
-      const [configKey, importName] = [genString(meta.configKey), getImportName(meta.importName)]
-      moduleMap[importName] = configKey
-
-      return { configKey, importName };
-    }
+    const getImportName = (name: string) => (name.startsWith('.') ? './' + join(relativeRoot, name) : name).replace(/\.\w+$/, '')
+    const modules = moduleInfo.map((meta) => [genString(meta.configKey), getImportName(meta.importName)])
 
     return [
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/core/templates.ts
```diff
@@ -121,14 +121,20 @@ export const schemaTemplate: NuxtTemplate<TemplateContext> = {
...
-        `    [${genString(meta.configKey)}]?: typeof ${genDynamicImport(meta.importName.startsWith('.') ? './' + join(relativeRoot, meta.importName) : meta.importName, { wrapper: false })}.default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>`
-      ),
+      ...moduleInfo.filter(Boolean).map((meta) => {
+        const configKey = genString(meta.configKey)
+        const importName = meta.importName.startsWith('.') ? './' + join(relativeRoot, meta.importName) : meta.importName
+        moduleMap[importName] = configKey
+
+        return `    [${configKey}]?: typeof ${genDynamicImport(importName, { wrapper: false })}.default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>`
+      }),
+      Object.keys(moduleMap).length > 0 ? `    modules: (NuxtModule | string | ${Object.keys(moduleMap).join(' | ')} | [NuxtModule | string, Record<string, any>] | ${Object.entries(moduleMap).map(([importName, configKey]) => `[${genString(importName)}, NuxtConfig[${genString(configKey)}]]`).join(' | ')})[]` : '',
```
danielroe: could we move the shared code (getting import name and config key) into the section above (L124)? that way we can consume it a bit more cleanly in both places below

danielroe: You will likely need to stringify the keys as well, something like:

```js
Object.keys(moduleMap).map(r => JSON.stringify(r)).join(' | ')
```

danielroe (changes requested):
Thank you! This is a nice improvement. Just a couple of suggestions, I think.

packages/nuxt/src/core/templates.ts
ineshbose: I've made changes based on these suggestions.

> Actually, is there any reason we need these explicitly typed?
My answer to that would be 1) DX and 2) ensuring configurations match the desired module inputs to avoid errors.
Based on how developers intend to configure the `modules` property, all that Nuxt would do is provide IntelliSense for it - but I'm mostly making use of `NuxtConfig` interface itself!

In my opinion, it won't make a difference if someone configures to use absolute paths, because 1) they can continue passing options as a property than an array element or 2) if they are to use an array, at least options would be typed and it won't make difference between these two:

```ts
  modules: [
   '/node_modules/.pnpm/@nuxtjs+color-mode@3.2.0/node_modules/@nuxtjs/color-mode/dist/module.mjs',
   // ['/node_modules/.pnpm/@nuxtjs+color-mode@3.2.0/node_modules/@nuxtjs/color-mode/dist/module.mjs', {}]
  ],
  colorMode: {},
})
```

The resulting schema would still include this as:

```ts
import { NuxtModule } from '@nuxt/schema'
declare module '@nuxt/schema' {
  interface NuxtConfig {
    ["colorMode"]?: typeof import("/node_modules/.pnpm/@nuxtjs+color-mode@3.2.0/node_modules/@nuxtjs/color-mode/dist/module.mjs").default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>,
    // modules: (NuxtModule | string | "/node_modules/.pnpm/@nuxtjs+color-mode@3.2.0/node_modules/@nuxtjs/color-mode/dist/module.mjs" | [NuxtModule | string, Record<string, any>] | ["/node_modules/.pnpm/@nuxtjs+color-mode@3.2.0/node_modules/@nuxtjs/color-mode/dist/module.mjs", NuxtConfig['colorMode']])[]
  }
}
```

danielroe: Because `string` is a valid type of `modules` I don't think we'll get any benefit here by using a literal type, added to which you have to have added the string first, then saved the file, before you'll get auto-completion for the string you've just typed.

danielroe:
Would you mind reopening this PR (or updating it) and giving me permission to push some changes? I think we can simplify this a bit further, and I've also spotted a pre-existing bug in the schema generation we can fix at the same time.

ineshbose:
Sorry! Please let me know if you have permission to push on here.

packages/nuxt/src/core/templates.ts
ineshbose: That is a very fair point, and there is the point to realise that the schema validation occurs after the initial load.
I went for literal types because configurations would have to correspond to the modules - like we can't have the tailwind module options for color-mode.

danielroe:
nope, but never mind - I've opened a PR to this PR 😆 

https://github.com/inesh-forks/nuxt/pull/1

packages/nuxt/src/core/templates.ts
danielroe: ah, but if we look at the below type:

```ts
modules: (
  NuxtModule |
  string |
  // here we do not need the literal type (and in fact it has no effect on type hinting because we also
  // allow string as a valid type)
  "@nuxtjs/color-mode" |
  [NuxtModule | string, Record<string, any>] |
  // here we _do_ need the literal type
  ["@nuxtjs/color-mode", NuxtConfig['colorMode']]
)[]
```

ineshbose: Yes, I agree with you - that one is redundant. 🙂

test/fixtures/basic/nuxt.config.ts
```diff
@@ -55,7 +55,16 @@ export default defineNuxtConfig({
     }
   },
   modules: [
-    '~/modules/example',
+    [
+      '~/modules/example',
+      {
+        typeTest(val) {
```
danielroe, line 61:
```suggestion
        typeTest (val) {
```

packages/nuxt/src/core/templates.ts
```diff
@@ -121,14 +121,17 @@ export const schemaTemplate: NuxtTemplate<TemplateContext> = {
     })).filter(m => m.configKey && m.name && !adHocModules.includes(m.name))
 
     const relativeRoot = relative(resolve(nuxt.options.buildDir, 'types'), nuxt.options.rootDir)
+    const getImportName = (name: string) => (name.startsWith('.') ? './' + join(relativeRoot, name) : name).replace(/\.\w+$/, '')
+    const modules = moduleInfo.map((meta) => [genString(meta.configKey), getImportName(meta.importName)])
```
danielroe, line 125:
```suggestion
    const modules = moduleInfo.map(meta => [genString(meta.configKey), getImportName(meta.importName)])
```

danielroe (approved):

danielroe:
Lovely work - thank you! ❤️

PR #18418, opened 2023-01-21
fix(nuxt): deprecate scanning directory index plugins

ineshbose:
Transferred from https://github.com/nuxt/framework/pull/10090

> 

### 🔗 Linked issue

nuxt/nuxt.js#12414 
https://github.com/nuxt/nuxt.js/issues/14358
(Work by https://github.com/nuxt/framework/pull/6544)

### ❓ Type of change

- [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Reopens the closed PR and aims to have consistent behaviour for scanning directories.

### 📝 Checklist


packages/nuxt/src/core/app.ts
```diff
@@ -111,8 +111,7 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
       ...(config.plugins || []),
       ...config.srcDir
         ? await resolveFiles(config.srcDir, [
-          `${config.dir?.plugins || 'plugins'}/*.{ts,js,mjs,cjs,mts,cts}`,
-          `${config.dir?.plugins || 'plugins'}/*/index.*{ts,js,mjs,cjs,mts,cts}`
+          `${config.dir?.plugins || 'plugins'}/*.{ts,js,mjs,cjs,mts,cts}`
```
danielroe, line 114: I appreciate this PR and the desire to add consistency here. But this would be a breaking change, and this early in Nuxt 3 release cycle it will take some time before we're ready to do that. We could possibly update the PR to show a deprecation warning if anyone is doing this, or even just add a 'TODO' comment to remind us to remove this option in future.

ineshbose: I do agree and had the same concern - hence your feedback is important! 🙂

Option B would be to go for the `{composables,utils}/*/index` route (https://github.com/nuxt/framework/pull/10084), but after seeing the previous context to that (https://github.com/nuxt/framework/pull/6025, and antfu's valid point - https://github.com/unjs/unimport/pull/86#issuecomment-1178724831), I'd go for the current approach (even though I'd like my index files to be scanned by default).

Still, both would have side effects - and I'm all for planning removal in a minor/major release after a release with a deprecation warning (along with a TODO comment), but I'm trying to brainstorm how the warning would be generated (figuring out if a file is scanned from `plugins/*/index.*` and it is not explicitly in `nuxt.config` - I may find a solution in a while).

danielroe:
Marking PR as draft as I'm aware you're considering the right approach. Ping me if I can help at any point.

ineshbose:
### Public Notes on working through PR

Potential points of interest -
https://github.com/nuxt/nuxt/blob/ac682fd32c8b734f8ba9140df33a659c26ac2049/packages/nuxt/src/app/nuxt.ts#L241
https://github.com/nuxt/nuxt/blob/ac682fd32c8b734f8ba9140df33a659c26ac2049/packages/nuxt/src/core/templates.ts#L87
https://github.com/nuxt/nuxt/blob/ac682fd32c8b734f8ba9140df33a659c26ac2049/packages/kit/src/plugin.ts#L72

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -21,14 +21,15 @@ For example:
 
 ```bash
 plugins
- | - myPlugin.ts
+ | - myPlugin.ts  // scanned
+ | - mySecondPlugin.ts  // scanned
```
danielroe, line 25:
```suggestion
```

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -21,14 +21,15 @@ For example:
...
  | - myOtherPlugin
- | --- supportingFile.ts
- | --- componentToRegister.vue
- | --- index.ts
+ | --- supportingFile.ts   // not scanned
+ | --- componentToRegister.vue   // not scanned
+ | --- index.ts  // not scanned
```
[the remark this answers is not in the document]
danielroe:
```suggestion
 | --- index.ts  // currently scanned but deprecated
```

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -21,14 +21,15 @@ For example:
...
 ```
 
-Only `myPlugin.ts` and `myOtherPlugin/index.ts` would be registered.
+Only `myPlugin.ts` and `mySecondPlugin.ts` would be registered. You can configure [`plugins`](/docs/api/configuration/nuxt-config#plugins-1) to include unscanned files.
```
danielroe, line 32:
```suggestion
Only `myPlugin.ts` and `myOtherPlugin/index.ts` would be registered. You can configure [`plugins`](/docs/api/configuration/nuxt-config#plugins-1) to include unscanned files.
```

antfu (approved):

Issue #18457, opened 2023-01-23
Add `routeNameSplitter` option from Nuxt 2


adamndev:
### Discussed in https://github.com/nuxt/nuxt/discussions/15906

<div type='discussions-op-text'>

<sup>Originally posted by **adamndev** January 17, 2023</sup>
Hey guys

I'm in the process of migrating a fairly large app over to Nuxt 3 and can't find any mention of the `routeNameSplitter` option from Nuxt 2.

Our routes match the naming conventions from other frameworks including Laravel, so instead of `events-attendees-create`, we'd generate this route as `events.attendees.create`. I think this is much cleaner as routes are likely to contain dashes themselves.

Is there any way to replicate this at the moment? And is this likely to be added back in?

Thanks for all the hard work 👌🏻</div>

I got a bit further in and realised that the simple regex used in a hook isn't gonna work and would need quite a bit of logic to figure out. This is a blocker to us migrating our Nuxt 2 app to Nuxt 3. Do you think we could get this added, please?

manniL:
Added a draft PR. From my POV it could make sense to add the feature back as a normal regex might not be enough here to "fix" it, e.g. via hooks.

danielroe:
On reflection, it seems to be quite an involved piece of coding (that we will need to maintain within Nuxt) for an edge case. Something like this might work if you want to keep the Nuxt page name algorithm:

```js
import { createResolver } from '@nuxt/kit';

// https://v3.nuxtjs.org/api/configuration/nuxt.config
export default defineNuxtConfig({
  hooks: {
    'pages:extend'(pages) {
      const root = createResolver(import.meta.url).resolve('./pages');
      function updateName(pages) {
        if (!pages) return;
        for (const page of pages) {
          let name = page.name.replace(/-/g, '.');
          for (const match of page.file.matchAll(/[\w\-]+/g)) {
            name = name.replace(match[0].replace(/-/g, '.'), match[0]);
          }
          page.name = name;
          updateName(page.children);
        }
      }
      updateName(pages);
    },
  },
});
```

https://stackblitz.com/edit/github-h24slc

Otherwise you can just set whatever name you want for your pages, based on the file path.

pi0:
ّI also agree that adding customizable conventions adds the burden of long-term maintenance for Nuxt and also makes it harder to integrate with toolings outside of Nuxt lifecycle to consider customized conventions (ie: a devtools or CLI extension that generates pages).

[...]

PR #18461, opened 2023-01-23
feat: add route separator

TheAlexLichter:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/18457

### ❓ Type of change

### 📚 Description

This PR introduces an optional config option called `router.routeSeparator` which will work similarly to the `routeNameSplitter` options from Nuxt 2.

(Happy to add docs if the decision to add it has been made if more than the one's added are necessary)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -199,10 +199,11 @@ function parseSegment (segment: string) {
 }
 
 function prepareRoutes (routes: NuxtPage[], routeNameSeparator: string, parent?: NuxtPage) {
+  const separatorWithIndex = new RegExp(`${escapeRE(routeNameSeparator)}index$`)
   for (const route of routes) {
-    // Remove <routeNameSeparator>index
     if (route.name) {
-      route.name = route.name.replace(new RegExp(`${escapeRE(routeNameSeparator)}index$`), '')
+      // Remove <routeNameSeparator>index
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/pages/utils.ts
```diff
@@ -70,7 +70,8 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
       const segmentName = tokens.map(({ value }) => value).join('')
 
       // ex: parent/[slug].vue -> parent-slug
-      route.name += (route.name && '-') + segmentName
+      console.log((route.name && routeSeparator) + segmentName)
```
danielroe, line 73:
```suggestion
```

DamianGlowala:
Do you think renaming it to `routeNameSeparator` would be clearer to the user?

manniL:
> Do you think renaming it to `routeNameSeparator` would be clearer to the user?

You mean to not accidentally mix it up with the "/" between routes? 
That would also fit to the initial implementation in Nuxt2.

DamianGlowala:
What I mean is, that just by the name of the option `routeNameSeparator`, as a user I can already tell what this option most likely does (adds a separator between the routes' *names*), whether the `routeSeparator` might be slightly less obvious. Subtle DX improvement :) Can already see the change you've made, awesome! ✨

manniL:
Closing, see https://github.com/nuxt/nuxt/issues/18457 for discussions ☺️

PR #18464, opened 2023-01-23
docs: updated public directory explanation

felixfavour:
Added explanation about the similarities between previous static directory and current public directory.

### ❓ Type of change

### 📝 Checklist


TheAlexLichter (changes requested):
This should be already explained in https://nuxt.com/docs/guide/directory-structure/public (when clicking on the "`public/` directory" link in the paragraph before. IMO that should be enough (but happy to hear other opinions here)

Instead, we might add it to the [migration guide](https://nuxt.com/docs/migration/overview) though.

danielroe:
Agreed, I think it would make sense for this to be in the migration guide.

TheAlexLichter (approved):

PR #18466, opened 2023-01-24
feat(kit): `exposeRuntimeModuleConfig` util

harlan-zw:
### ❓ Type of change

### 📚 Description

A common use case that I've come across with modules is needing to expose the module config to the app and nitro runtimes. There are a few common ways people do this, all involving quite a bit of boilerplate with no TypeScript support.

This function allows you to simply expose the values in an opinionated way.

```ts
export interface ModuleOptions {
  myOption: { foo: string }
}

export default defineNuxtModule<ModuleOptions>({
  meta: {
    name: 'nuxt-my-module',
  },
 defaults: {
   myOption: { foo: 'bar' }
 },
 async setup(config, nuxt) {
    exposeRuntimeModuleConfig('nuxt-my-module', config)
    // config now available at #nuxt-my-module/config
  }
})
```

**.nuxt/modules/config/nuxt-my-module.d.ts**

```ts
// generated by nuxt-my-module
import type { ModuleOptions } from 'nuxt-my-module'

declare module '#nuxt-my-module/config' {
  export const myOption: 'bar' | ModuleOptions['myOption']
}
```

**composables/foo.ts**

```ts
import { myOption } from '#nuxt-my-module/config'

export function useFoo() {
   return myOption.foo // typed to 'bar' or ModuleOptions['myOption']
}
```

Currently testing this in production on most of my modules.

Ideally, this function would use another more general function such as `exposeObjectExport(alias, fileName, object)`, which could be used to expose arbitrary data that is needed in the runtime.

This is a draft until I can write some doc but open to feedback on any the opinionated decisions:
- the function name
- the export name `#{modulename}/config`
- the types location, currently `.nuxt/modules/config/${moduleName}.d.ts`
- type generation, is there a better way? knitwork seems it would do the job but also a bit overkill 
- primitive values jsonified

### 📝 Checklist


packages/kit/src/module/expose.ts
```diff
@@ -0,0 +1,44 @@
+import { resolve } from 'pathe'
+import { useNuxt } from '../context'
+import { addTemplate } from '../template'
+
+/**
+ * Expose virtual module config to the NuxtApp and nitro with TypeScript support.
+ */
+export function exposeRuntimeModuleConfig (moduleName: string, config: Record<string, any>) {
```
lihbr, line 8: I'm wondering if `config` could be typed to only support serializable values (i.e. a function is not) to prevent confusion/mistakes

lihbr (commented):
Love it so far! A nice common pattern used by modules~

packages/kit/src/module/expose.ts
```diff
@@ -0,0 +1,44 @@
...
+  nuxt.hooks.hook('nitro:config', (nitroConfig) => {
+    nitroConfig.virtual![alias] = jsExports
+  })
+
+  // need to create a ts def for the file
+  const typeDefName = `modules/config/${moduleName}.d.ts`
+  const tsExports = Object.keys(config).map((k) => {
+    // expose actual values in dev
+    if (nuxt.options.dev) { return `  export const ${k}: ${JSON.stringify(config[k])} | ModuleOptions['${k}']` }
+    return `  export const ${k}: ModuleOptions['${k}']`
```
lihbr, line 26: What if the options exposed by the module aren't similar to `ModuleOptions[k]` (i.e. they were transformed, parsed, resolved)? Not sure if this is out of scope for this functions and meant for a lower level alternative :thinking:

pi0:
I really do love the idea that we are providing a kit utility to automate injecting config. 

This method of injecting (using a generated template) is mainly what app config layer does which i plan to also include in Nitropack. We already reserved `appConfig.server` namespace so that's a possibility to instead of generating them.

Alternatively, if we keep it, i would suggest to call it `exposeBundledConfig`/`addBundledConfig` that avoids confusing with runtime config which is handled completely differently (env based, string, runtime assignable)

harlan-zw:
Going to close for now, while we did discuss a solution of using runtime config I think the examples provided in the new module author guide are enough.

PR #18441, opened 2023-01-22
feat(head): `useSeoMeta` composable

harlan-zw:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Seems like the old PR didn't transfer: https://github.com/nuxt/framework/pull/9617

With the migration to @vueuse/head v1, there are new composables available that may be useful for Nuxt users. Specifically the `useSeoMeta` composable has been requested.

You can see details on it here: https://unhead.harlanzw.com/guide/guides/useseometa

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #18476, opened 2023-01-24
perf(nuxt): add tree-shaken `useServerSeoMeta` composable

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/18441

### ❓ Type of change

### 📚 Description

This PR automatically excludes the `useSeoMeta` composable from the client-side bundle.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/head/runtime/composables.ts
+++ b/packages/nuxt/src/head/runtime/composables.ts
@@ -24,8 +24,7 @@ export function useHead<T extends HeadAugmentations> (input: UseHeadInput<T>, op
  * This function will have no effect (and will return nothing) if called on the client.
  */
 export const useSeoMeta: typeof _useSeoMeta = (meta) => {
-  // SEO meta is
-  if (process.server || process.dev) {
+  if (process.server) {
     return _useSeoMeta(meta)
   }
 }
--- a/packages/nuxt/src/head/runtime/composables.ts
+++ b/packages/nuxt/src/head/runtime/composables.ts
@@ -1,5 +1,6 @@
 import type { HeadEntryOptions, UseHeadInput, ActiveHeadEntry } from '@vueuse/head'
 import type { HeadAugmentations } from '@nuxt/schema'
+import { useSeoMeta as _useSeoMeta } from '@vueuse/head'
 import { useNuxtApp } from '#app'
 
 /**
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/head/runtime/composables.ts
```diff
@@ -12,3 +13,19 @@ import { useNuxtApp } from '#app'
...
+ * as a flat object with full TypeScript support.
+ *
+ * This helps you avoid typos and common mistakes, such as using `name`
+ * instead of `property`.
+ *
+ * This function will have no effect (and will return nothing) if called on the client.
+ */
+export const useSeoMeta: typeof _useSeoMeta = (meta) => {
+  // SEO meta is
+  if (process.server || process.dev) {
```
pi0: We might at least also use it for when ssr globally disabled with `ssr: false`.

antfu (approved):

packages/nuxt/src/head/runtime/composables.ts
```diff
@@ -12,3 +13,18 @@ import { useNuxtApp } from '#app'
...
+ * The `useSeoMeta` composable lets you define your site's SEO meta tags
+ * as a flat object with full TypeScript support.
+ *
+ * This helps you avoid typos and common mistakes, such as using `name`
+ * instead of `property`.
+ *
+ * This function will have no effect (and will return nothing) if called on the client.
+ */
+export const useSeoMeta: typeof _useSeoMeta = (meta) => {
+  if (process.server) {
```
pi0, line 27: I think we need to bypass this at least when global ssr is disabled with `ssr:false`.

danielroe: Why would we want to run this on client-side when `ssr: false`?

pi0: Because public websites still get rendered crawled by search engines and benefit from seo meta as such. Obviously not best practice but a valid deployment option with Nuxt.

pi0: Yes, and needs to be a build-time flag to work with tree-shaking. That's why I'm suggesting relying on global `ssr: false` (can be passed similar to `process.server`). global `ssr: false` mode is different.

`useServerSeoMeta` seems good idea too!

packages/nuxt/src/imports/presets.ts
```diff
@@ -6,7 +6,8 @@ const commonPresets: InlinePreset[] = [
   defineUnimportPreset({
     from: '#head',
     imports: [
-      'useHead'
+      'useHead',
+      'useSeoMeta'
```
danielroe, line 10:
```suggestion
      'useSeoMeta',
      'useServerSeoMeta'
```

harlan-zw:
`useServerSeoMeta` - good idea :+1:

PR #18516, opened 2023-01-25
fix(schema): show `payloadExtraction` warning only when unset

hslee2008:
### 🔗 Linked issue

No linked issue.

### ❓ Type of change

### 📚 Description

This is only a minor change and I am not sure if it is a good one.

This change should be made because when users sets `experimental.payloadExtraction` to `true` on purpose, they are hinting that either they are doing this on purpose or wanting the warnings to be gone. 

This code change only changes the code so that if the `enabled` property is `undefined` (meaning the user did not set any configuration), it shows the warnings, but if the user sets it to `true` there will be no warnings.

### 🧪 Tests / Examples

notdefined

setfalse
settrue


packages/schema/src/config/experimental.ts
```diff
@@ -79,11 +79,10 @@ export default defineUntypedSchema({
      */
     payloadExtraction: {
       async $resolve(enabled, get) {
-        enabled = enabled ?? false
-        if (enabled) {
+        if (enabled == null) {
```
danielroe, line 82: I prefer to make it explicit:

```suggestion
        if (enabled === null || enabled === undefined) {
```

danielroe (approved):
This is a nice change, thank you!

danielroe (dismissed):
Oops, my apologies; I forgot how we are doing this. In fact we set this here:

https://github.com/nuxt/nuxt.js/blob/6e78f9dbb03a059814d1284511f732b901dea4f6/packages/nuxi/src/commands/build.ts#L30-L34

So we don't want to show warning always if it is null or undefined because it is only on prerendering that it is in fact enabled, and in that case it is because it is _set_ to `true`.

HyunseungLee-Travis:
I pressed the wrong button. Sorry.

HyunseungLee-Travis:
> Oops, my apologies; I forgot how we are doing this. In fact we set this here:
> 
> https://github.com/nuxt/nuxt.js/blob/6e78f9dbb03a059814d1284511f732b901dea4f6/packages/nuxi/src/commands/build.ts#L30-L34
> 
> So we don't want to show warning always if it is null or undefined because it is only on prerendering that it is in fact enabled, and in that case it is because it is _set_ to `true`.

I want to work on this, but I am not exactly sure how it works.

pi0 (approved):

PR #18553, opened 2023-01-26
feat(kit): add `addRouteMiddleware` method

### LINKED ISSUE #3934: add `addRouteMiddleware` utility from `@nuxt/kit`

Simplifies adding middlewares from modules.

Currently (assuming that middleware is added programmatically from a module, without user interaction), to add a middleware from a module, one has to:
a) add a middleware using either `this.addTemplate` or `this.addPlugin`.
b) add a plugin using `this.addPlugin`
c) from the plugin added in previous point, one has to import middleware added in point 'a' just to trigger a side effect of it registering itself in nuxt's middleware array

This is how official `auth-module` does it* at least and I couldn't find an easier way.

So basically we have to add a plugin just so that it can trigger registration of our custom middleware. It would make sense to be able to skip the plugin step and just add a middleware using function exposed in ModuleContainer api (so just something like `this.addMiddleware()` from a module.

* https://github.com/nuxt-community/auth-module/blob/abfa084184240b76d7071c20002a5e96821916f9/lib/module/plugin.js

Add `this.addMiddleware()` function in ModuleContainer API (`this` object in a module) for having an easy way of adding a middleware without using extra plugin.

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c7796">#c7796</a>)</em></sub></div>

xtoolkit:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/3934

### ❓ Type of change

### 📚 Description

This method help to modules for add middlewares

```ts
addRouteMiddleware({
  name: 'test',
  path: 'path-to-middleware'
})
```

### 📝 Checklist

### DIFF
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -55,7 +55,7 @@ export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], op
         if (options.override === true) {
           app.middleware[find] = middleware
         } else {
-          console.warn(`'${middleware.name}' middleware exists with '${app.middleware[find].path}' path. You can set override option 'true' for replace with your middleware path.`)
+          console.warn(`'${middleware.name}' middleware already exists at '${app.middleware[find].path}'. You can set \`override: true\` to replace it.`)
         }
       } else {
         app.middleware.push(middleware)
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -35,7 +35,7 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
   }
 }
 
-export function extendMiddleware(cb: (middleware: NuxtMiddleware[]) => void) {
+export function extendMiddleware (cb: (middleware: NuxtMiddleware[]) => void) {
   const nuxt = useNuxt()
   nuxt.hook('app:resolve', app => cb(app.middleware))
 }
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -46,9 +46,8 @@ export interface AddRouteMiddlewareOptions {
 
 export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], options: AddRouteMiddlewareOptions = {}) {
   const nuxt = useNuxt()
+  const middlewares = Array.isArray(input) ? input : [input]
   nuxt.hook('app:resolve', (app) => {
-    const middlewares = Array.isArray(input) ? input : [input]
-
     for (const middleware of middlewares) {
       const find = app.middleware.findIndex(item => item.name === middleware.name)
       if (find >= 0) {
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -49,7 +49,7 @@ export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], op
   nuxt.hook('app:resolve', (app) => {
     const middlewares = Array.isArray(input) ? input : [input]
 
-    middlewares.forEach((middleware) => {
+    for (const middleware of middlewares) {
       const find = app.middleware.findIndex(item => item.name === middleware.name)
       if (find >= 0) {
         if (options.override === true) {
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -1,4 +1,4 @@
-import type { NuxtHooks } from '@nuxt/schema'
+import type { NuxtHooks, NuxtMiddleware } from '@nuxt/schema'
 import type { NitroRouteConfig } from 'nitropack'
 import { defu } from 'defu'
 import { useNuxt } from './context'
@@ -33,4 +33,9 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
       ? defu(rule, opts.routeRules[route])
       : defu(opts.routeRules[route], rule)
   }
+}
+
+export function extendMiddleware(cb: (middleware: NuxtMiddleware[]) => void) {
+  const nuxt = useNuxt()
+  nuxt.hook('app:resolve', app => cb(app.middleware))
 }
--- a/packages/kit/src/pages.ts
+++ b/packages/kit/src/pages.ts
@@ -1,5 +1,6 @@
 import type { NuxtHooks, NuxtMiddleware } from '@nuxt/schema'
 import type { NitroRouteConfig } from 'nitropack'
+import consola from 'consola'
 import { defu } from 'defu'
 import { useNuxt } from './context'
 import { isNuxt2 } from './compatibility'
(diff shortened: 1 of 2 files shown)


packages/kit/src/pages.ts
```diff
@@ -34,3 +34,8 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
       : defu(opts.routeRules[route], rule)
   }
 }
+
+export function extendMiddleware (cb: (middleware: NuxtMiddleware[]) => void) {
+  const nuxt = useNuxt()
+  nuxt.hook('app:resolve', app => cb(app.middleware))
```
pi0, line 40: It would be nice to:
- Support input argument being directly an array to inject
- Support returning new middleware from cb and push/normalize/dedup them into `app.middleware` <-- this would be much safer than just exposing an array and let developers push into it

xtoolkit: you want like this:

```ts
function extendMiddleware(newMiddleware: NuxtMiddleware[]): void;
function extendMiddleware(cb: (middleware: NuxtMiddleware[]) => NuxtMiddleware[]): void;
function extendMiddleware (input: any) {
  const nuxt = useNuxt()
  nuxt.hook('app:resolve', (app) => {
    if (Array.isArray(input)) {
      app.middleware = [...app.middleware, ...input]
    } else {
      app.middleware = input(app.middleware)
    }
  })
}
```

or add option argument for override middleware by name?

danielroe: This is my suggestion for the API:

```ts
function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], options: { override: boolean }): void
```

Under the hood, if override is not set, then we will check existing middleware for any of the same name, and warn the user if so (and not replace the existing middleware).

danielroe: (Feel free to ping me on Discord if I can be of any help.)

xtoolkit: hi again

In `addRouteMiddleware` api, module can only add middleware. why not use [Adding middleware Dynamically](https://nuxt.com/docs/guide/directory-structure/middleware#adding-middleware-dynamically) helper.

but in `extendMiddleware` api, module can manage all middleware. like delete, sort and etc

an this moment, nuxt core team must decide which one to implement

tnx <3

danielroe: I would prefer `addRouteMiddleware`. Modules that need to do more advanced things can directly interact with the hook.

xtoolkit: you are right.
for advance change better to use hooks. 

if you are not comment or change this https://github.com/nuxt/nuxt/pull/18553#discussion_r1090448690 api, i can implement that.

danielroe: Yes, that would be perfect, thank you ❤️

packages/kit/src/pages.ts
```diff
@@ -34,3 +35,35 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], options: AddRouteMiddlewareOptions = {}) {
+  const nuxt = useNuxt()
+  nuxt.hook('app:resolve', (app) => {
+    const addMiddleware = (middleware: NuxtMiddleware) => {
+      const find = app.middleware.findIndex(item => item.name === middleware.name)
+      if (find >= 0) {
+        if (options.override === true) {
+          app.middleware[find] = middleware
+        } else {
+          consola.warn(`'${middleware.name}' middleware exists with '${app.middleware[find].path}' path. You can set override option 'true' for replace with your middleware path.`)
```
danielroe, line 57: We can use console.warn. It should be wrapped in consola automatically by Nuxt.


```suggestion
          console.warn(`'${middleware.name}' middleware exists with '${app.middleware[find].path}' path. You can set override option 'true' for replace with your middleware path.`)
```

packages/kit/src/pages.ts
```diff
@@ -34,3 +35,35 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+        }
+      } else {
+        app.middleware.push(middleware)
+      }
+    }
+    if (Array.isArray(input)) {
```
danielroe, line 63: I wonder if we can just normalize input first to an array and then iterate over it. Then no need to define an inline addMiddleware function as well.

packages/kit/src/pages.ts
```diff
@@ -34,3 +34,32 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
       : defu(opts.routeRules[route], rule)
   }
 }
+
+export interface AddRouteMiddlewareOptions {
+  /**
+   * Override middleware path
```
danielroe, line 40:
```suggestion
   * Override existing middleware with the same name, if it exists
```

packages/kit/src/pages.ts
```diff
@@ -34,3 +34,32 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+   */
+  override?: boolean
+}
+
+export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], options: AddRouteMiddlewareOptions = {}) {
+  const nuxt = useNuxt()
+  nuxt.hook('app:resolve', (app) => {
+    const middlewares = Array.isArray(input) ? input : [input]
+
+    middlewares.forEach((middleware) => {
```
danielroe, line 52: In general we prefer for...of rather than forEach in Nuxt.

```suggestion
    for (const middleware of middlewares) {
```

packages/kit/src/pages.ts
```diff
@@ -34,3 +34,32 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+      if (find >= 0) {
+        if (options.override === true) {
+          app.middleware[find] = middleware
+        } else {
+          console.warn(`'${middleware.name}' middleware exists with '${app.middleware[find].path}' path. You can set override option 'true' for replace with your middleware path.`)
+        }
+      } else {
+        app.middleware.push(middleware)
+      }
+    })
```
danielroe, line 63:
```suggestion
    }
```

packages/kit/src/pages.ts
```diff
@@ -34,3 +34,32 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+  nuxt.hook('app:resolve', (app) => {
+    const middlewares = Array.isArray(input) ? input : [input]
+
+    middlewares.forEach((middleware) => {
+      const find = app.middleware.findIndex(item => item.name === middleware.name)
+      if (find >= 0) {
+        if (options.override === true) {
+          app.middleware[find] = middleware
+        } else {
+          console.warn(`'${middleware.name}' middleware exists with '${app.middleware[find].path}' path. You can set override option 'true' for replace with your middleware path.`)
```
danielroe, line 58:
```suggestion
          console.warn(`'${middleware.name}' middleware already exists at '${app.middleware[find].path}'. You can set \`override: true\` to replace it.`)
```

packages/kit/src/pages.ts
```diff
@@ -34,3 +34,32 @@ export function extendRouteRules (route: string, rule: NitroRouteConfig, options
...
+   * @default false
+   */
+  override?: boolean
+}
+
+export function addRouteMiddleware (input: NuxtMiddleware | NuxtMiddleware[], options: AddRouteMiddlewareOptions = {}) {
+  const nuxt = useNuxt()
+  nuxt.hook('app:resolve', (app) => {
+    const middlewares = Array.isArray(input) ? input : [input]
+
```
danielroe, line 51: We can realise a small performance improvement by normalizing immediately rather than every time the app is resolved.

```suggestion
  const middlewares = Array.isArray(input) ? input : [input]
  nuxt.hook('app:resolve', (app) => {
```

danielroe (approved):
Beautiful! Thank you.

Issue #18584, opened 2023-01-27
Unhandled promise rejection when `asyncData` throws


rchl:
### Environment

```
  Binaries:
    Node: 16.19.0 - ~/.nvm/versions/node/v16.19.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.19.0/bin/yarn
    npm: 8.19.3 - ~/.nvm/versions/node/v16.19.0/bin/npm
  npmPackages:
    @nuxt/postcss8: ^1.1.3 => 1.1.3
    @nuxt/types: ^2.15.8 => 2.15.8
    nuxt: ^2.15.8 => 2.15.8
```

### Reproduction

https://stackblitz.com/edit/github-bxdb1b?file=pages%2Findex.vue

### Describe the bug

When exception is thrown during handling of `asyncData`, Nuxt catches it and shows server error but it also chains returned promise without using `catch`. This results in unhandled promise rejection.

https://github.com/nuxt/nuxt/blob/e5efc259401204888d7bd9aebbc4a7cf646851b6/packages/vue-app/template/server.js#L320-L325

One note: this issue does not manifest in any way whatsoever in plain Nuxt. Node doesn't report unhandled promise error. I'm not sure if that's because vue-server-renderer captures that output or maybe Node 16 just doesn't report those. That said, the promise is in fact unhandled and causes issues when in combination with `@nuxtjs/sentry` where it will make the app crash (whether that's a Sentry issue or not is not really relevant IMO).

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/18585

PR #18585, opened 2023-01-27
fix(vue-app): handle promise rejection from `asyncData`

rchl:
When exception is thrown during handling of `asyncData`, Nuxt catches it and shows server error but it also chains returned promise without using `catch`. This results in unhandled promise rejection.

### 🔗 Linked issue

[](https://github.com/nuxt/nuxt/issues/18584)

### ❓ Type of change

### 📚 Description

The original code

https://github.com/nuxt/nuxt/blob/e5efc259401204888d7bd9aebbc4a7cf646851b6/packages/vue-app/template/server.js#L320-L325

gets the `asyncData` promise and chains it to `applyAsyncData`. That `then` doesn't run when `asyncData` has thrown but since this promise is not further chained (especially to `catch`) it effectively makes it an unhandled promise rejection error.

### 📝 Checklist


packages/vue-app/template/server.js
```diff
@@ -321,8 +321,7 @@ const render404Page = () => {
         promise.then((asyncDataResult) => {
           ssrContext.asyncData[Component.cid] = asyncDataResult
           applyAsyncData(Component)
-          return asyncDataResult
-        })
+        }).catch(() => {})
```
rchl, line 324: removed `return` since this promise is not chained with anything so it doesn't need to return anything.

rchl:
any prospects for getting this merged?

danielroe (dismissed):

packages/vue-app/template/server.js
rchl, line 324: After discussing this a bit with @danielroe, I realized then maybe the intention here was to push this chained promise to `promises` rather than the original one? So instead of the current code do:

```js
        let promise = promisify(Component.options.asyncData, app.context)
        promise = promise.then((asyncDataResult) => {
          ssrContext.asyncData[Component.cid] = asyncDataResult
          applyAsyncData(Component)
          return asyncDataResult
        })
        promises.push(promise)

```

(with or without extra `.catch()`).

But that's more risky change as crash in `.then` handler would then affect other code up the chain. Maybe that would be the right thing to do but the change I made is at least safe from the perspective of not changing the behavior apart from catching unhandled rejection.

rchl: Yeah, I think that's a better fix. And it seems quite safe to me as `applyAsyncData` is very simple.

danielroe (approved):

TheAlexLichter (approved):

PR #18586, opened 2023-01-27
fix(nuxt): provide fallback values for undefined runtime config

### LINKED ISSUE #13845: better way to set `runtimeConfig` fallbacks
### Environment

Production

### Reproduction

https://stackblitz.com/edit/github-dnbk4c

### Describe the bug

On my staging + production environment, we use Kubernetes ConfigMaps to store runtime configs and use them in Nuxt app.

I found a problem with Nuxt `^3.0.0-rc.1` that it cannot find runtime configs from Kubernetes since we don't create `.env` file.
- You can see in Reproduction, we have a `development` directory to store Kubernetes ConfigMaps
- We set env variables values on Jenkins pipeline & not create `.env` file.

The last version where Nuxt + Kubernetes config maps still work is `"nuxt3": "^3.0.0-27460489.53fbca7"`. This is how I refer config in server middleware:
```ts
import express from 'express'
import config from '#config'
import createProxy from '~/services/server/proxy'

const app = express()
app.use('*', createProxy(config.API_HOST))

export default app
```

Thank u.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/13845

### ❓ Type of change

### 📚 Description

Whilst working on a PR to improve docs, spotted an opportunity to remove one of the non-intuitive pieces of the docs (the fact that you _have_ to give everything a value - it's not enough to point it to an undefined env variable.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -423,14 +423,17 @@ export default defineUntypedSchema({
    * @type {typeof import('../src/types/config').RuntimeConfig}
    */
   runtimeConfig: {
-    $resolve: async (val: RuntimeConfig, get) => defu(val, {
-      public: {},
-      app: {
-        baseURL: (await get('app')).baseURL,
-        buildAssetsDir: (await get('app')).buildAssetsDir,
-        cdnURL: (await get('app')).cdnURL,
-      }
-    })
+    $resolve: async (val: RuntimeConfig, get) => {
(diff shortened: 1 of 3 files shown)


pi0 (approved):
Seems a good idea. We could also add it to upstream nitro for default and consistent beahvior

Issue #15462, opened 2022-11-13
Support for unified trailing slash handling


manniL:
### Describe the feature

Yes, here we go again. I'd like to propose uniform trailing slash handling - as I did [for Nuxt 2](https://github.com/nuxt/nuxt.js/pull/6331) - again. But I firmly believe that the implementation can be better, less error prone and even easier with Nuxt 3.

## Why?

Trailing slash handling is crucial when it comes to SEO. Right now, the content shown for `/my-route` and `/my-route/` is already the same. But ideally, one of these versions should not exist (and instead redirect to the other). If that is not possible, a fitting `canonical` URL should be set to define the preferred URL.

Now it seems that it is rather easy, we could go and simply use the `pages:extend` hook and apply/remove the trailing slash and add some nitro page rules (or a middleware) for redirects.

But there is another important part: **the internal linking**.  Often neglected, it **is key for a good on-page SEO**. Ideally, the implementation of this feature would cover correct internal linking too. Ideally, also working with any module, like the `i18n` module.

## Proposal

1. Have a global option to enforce a trailling slash `/route/` or *no* slash `/route` for every internal Nuxt route that exists, ideally including Nitro server-routes (API routes are not that important in this case)
2. The "disallowed" version should redirect to the correct version
3. `<NuxtLink>` should be aware of this choice and adapt linking correctly so the preferred version is the one that will be rendered.

### Additional information

- [X] Would you be willing to help implement this feature?
- [X] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

Atinux:
This is needed for sure 👍

harlan-zw:
Definitely important, it's a bit of a tricky one because the scope for what needs to know in the trailing slashes option is quite broad.

It's one of the reasons I built a few of the submodules in https://github.com/harlan-zw/nuxt-seo-kit. 

Some of the things that require it:
- sitemap
- schema.org
- robots
- canonical urls
- link checking

I was going to add a custom NuxtLink component but I thought the DX of having to use a different link component wasn't worth it, so would be great to see it in the core.

## Config

The way I've implemented it in Nuxt SEO Kit was a public runtime called `trailingSlash`. I feel quite strongly that the config should be easily accessible to modules and end users. 

[...]

PR #18593, opened 2023-01-28
fix(nuxt): respect redirects which differ only by trailing slash

### LINKED ISSUE #18561: SSR redirect to same route without trailing slash returns 302 but no Location header
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.0.1`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.0.0`
- Package Manager:  `pnpm@7.26.0`
- Builder:          `vite`
- User Config:      `modules`, `vite`, `runtimeConfig`, `css`, `app`, `hooks`, `content`, `i18n`
- Runtime Modules:  `@nuxt/content@2.3.0`, `@nuxtjs/i18n@8.0.0-beta.7`
- Build Modules:    `-`

### Reproduction

Minimal reproduction middleware that fails to produce a valid HTTP 301 or 302 response during server-side rendering:

```js
// Redirect trailing slashes, preserving query string: /foo/ -> /foo
	if (to.path.length > 1 && to.path.slice(-1) === '/') {
		const path = to.path.replace(/\/+$/, '') || '/'
		return navigateTo({ path, query: to.query, hash: to.hash }, { redirectCode: 301 })
	}
}
```

### Describe the bug

The current Nuxt version (3.1.1) will return an invalid HTTP response during SSR if a redirect to the same route happens within `defineNuxtRouteMiddleware`. The HTTP response contains a status code of `301` or `302`, but no `Location` header.

### Additional context

It seems like the issue is here, where the router plugin uses `ufo.isEqual` in non-strict mode, which considers `/about/` and `/about` to be the same:

https://github.com/nuxt/nuxt/blob/64b835c70701c8f27debe8a534ec99034bda812b/packages/nuxt/src/app/plugins/router.ts#L251-L256

### Logs

```shell-script
❯ curl -vvv http://localhost:3000/about/
*   Trying 127.0.0.1:3000...
* Connected to localhost (127.0.0.1) port 3000 (#0)
> GET /about/ HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/7.86.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Access-Control-Allow-Origin: *
< content-type: text/html;charset=utf-8
< x-powered-by: Nuxt
< date: Thu, 26 Jan 2023 16:27:36 GMT
< connection: close
< content-length: 22456
<
<!DOCTYPE html>
<html >
...
```

### LINKED ISSUE #18592: Location header not set when navigateTo is used for redirection
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.0.0`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.1.0`
- Package Manager:  `yarn@1.22.11`
- Builder:          `vite`
- User Config:      `app`, `runtimeConfig`, `imports`, `components`, `modules`, `i18n`, `css`, `postcss`, `routeRules`, `nitro`, `vite`
- Runtime Modules:  `@nuxtjs/i18n@8.0.0-beta.8`, `@pinia/nuxt@0.4.6`
- Build Modules:    `-`

### Reproduction

Add the following to the nuxt.config file:
```
  modules: [
    '@nuxtjs/i18n',
  ],
  i18n: {
    strategy: 'prefix',
    customRoutes: 'config',
    detectBrowserLanguage: {
      useCookie: true,
      cookieKey: 'LANGUAGE',
      alwaysRedirect: true,
    },
    locales: [
      {
        code: 'nl',
        name: 'Nederlands',
      },
      {
        code: 'fr',
        name: 'Français',
      },
      {
        code: 'en',
        name: 'English',
      },
    ],
    defaultLocale: 'en',
    vueI18n: {
      legacy: false,
      locale: 'en',
      fallbackLocale: 'nl',
      messages: {
        nl: { ... },
        fr: { ... },
        en: { ... },
      },
    },
  },
```

Surf to the root route `/`

### Describe the bug

**tldr;** Using navigateTo for redriection causes redirection (Status 302) without Location header.
As is the case when using @nuxtjs/i18n module with `alwaysRedirect: true`

Going to the root route `/` we get a 302 response without Location header.
The presence of the Location header is no absolute requirement.
The absence of this header would have questionable practicability though.

I've came across this with healthy check software that did not cope with this.

I think a solution would be to add the header before this line: https://github.com/nuxt/nuxt/blob/de4086f6edabefc1a46ee08a099c3065f24f6fe5/packages/nuxt/src/app/composables/router.ts#L104

### Additional context

_No response_

### Logs

_No response_

mathieumagalhaes:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/18592
Also probably fixes https://github.com/nuxt/nuxt/issues/18561

### ❓ Type of change

Adds the missing location header on redirect

### 📚 Description

Added a composable in ssr.ts `setResponseHeader`
Used it where needed in `router.ts`

This adds the missing location header on redirect

### 📝 Checklist

Documentation changes not needed


packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -25,3 +25,10 @@ export function setResponseStatus (code: number, message?: string) {
     }
   }
 }
+
+export function setResponseHeader (name: string, value: string) {
```
mathieumagalhaes, line 29: Not sure if we should add other types possibilities to the `value` argument, besides `string`.
For now this should be enough though.

manniL: Could be maybe use [`setResponseHeader`](https://www.jsdocs.io/package/h3#setResponseHeader) right away?

danielroe (changes requested):
This seems reasonable and I like the new `setResponseHeader` utility but I think we instead need to change the behaviour of this line if we want to support redirecting with a trailing slash.

https://github.com/nuxt/nuxt.js/blob/9751c9dab7423d6df8046e68660bcc2615807b87/packages/nuxt/src/pages/runtime/plugins/router.ts#L174

Otherwise even if we set the `location` header, we will still end up rendering the full page's HTML.

mathieumagalhaes:
> This seems reasonable and I like the new `setResponseHeader` utility but I think we instead need to change the behaviour of this line if we want to support redirecting with a trailing slash.
> 
> https://github.com/nuxt/nuxt.js/blob/9751c9dab7423d6df8046e68660bcc2615807b87/packages/nuxt/src/pages/runtime/plugins/router.ts#L174
> 
> Otherwise even if we set the `location` header, we will still end up rendering the full page's HTML.

Not sure how to achieve that with the inner workings of Nuxt / Nuxt utilities.

**From Vue Router docs**:
_To make trailing slash behavior consistent: by default all routes allow a trailing slash.
It can be disabled by using the strict option and manually appending (or not) a slash to the routes._

Knowing that, all non trailing slash formats URL's would be fine to use, would this not be the best way to implement it then?
Unless I'm missing something.

Vue router also leans towards non trailing slashes format heavily:
_"Navigating or resolving to the named route `{ name: '...' }` will **now produce a URL without a trailing slash**"_

mathieumagalhaes:
As of Nuxt 3 the `router.trailingSlash` configuration is gone from `nuxt.config` file.

If that ever comes back, or if there is a similar configuration:
We could indeed use `isEqual('/foo/', 'foo', { trailingSlash: true })` , after reading the `ufo` docs.
And replace `true` with the value of the  `router.trailingSlash` configuration.

manniL:
> If that ever comes back, or if there is a similar configuration

* https://github.com/nuxt/nuxt/issues/15462

mathieumagalhaes:
@manniL how does https://github.com/nuxt/nuxt/issues/15462#issuecomment-1418944206 affect this pull request?

manniL:
> @manniL how does [#15462 (comment)](https://github.com/nuxt/nuxt/issues/15462#issuecomment-1418944206) affect this pull request?

Wanted to make sure to link the discussion regarding trailing slash behavior here FWIW - as it won't come back as config but more related to `createNuxtLink` (see follow-ups there) ☺️

danielroe (approved):

TheAlexLichter (approved):

PR #18641, opened 2023-01-31
feat(nuxi): cli wrapper for self restart

antfu:
### ❓ Type of change

### 📚 Description

This would allow Nuxt to do hard restart, for self upgrading or other stuffs.

### 📝 Checklist


packages/nuxi/bin/nuxi.mjs
```diff
@@ -1,3 +1,33 @@
 #!/usr/bin/env node
-process._startTime = Date.now()
-import('../dist/cli.mjs').then(r => (r.default || r).main())
+import { fileURLToPath } from 'node:url'
+import { execa } from 'execa'
```
danielroe, line 3: This file isn't bundled. Could we perhaps use `child_process`?

danielroe:
This is lovely 💯

packages/nuxi/bin/nuxi.mjs
antfu: IIRC, using `child_process` we will need to handle quite a lot of edge cases across platforms (path, windows, etc.)

Refactored to use the bundled execa. Thanks

danielroe (approved):

pi0 (changes requested):
Nice idea to support restart, however, we can use the semi-standard signal `SIGHUP` to be handled for dev command and any other sub-command that needs to listen to it for restarting instead of restarting the entire CLI.

Atinux:
> we can use the semi-standard signal SIGHUP to be handled for dev command

You mean listening in https://github.com/nuxt/nuxt/blob/main/packages/nuxi/src/commands/dev.ts ?

But the issue is that this file is already importing Nuxt dependencies where in the wrapper we don't have this caching issue

pi0:
> But the issue is that this file is already importing Nuxt dependencies

That's also a huge performance gain to keep cache. What dependencies you are thinking would be cached and useful to be invalided?

danielroe:
I think performance isn't the issue here; a full restart would be an exceptional situation and most likely _in order to_ flush the cache, for example a programmatic module version upgrade + restarting of nuxt.

antfu:
The problem we are trying to solve here is the support for "Hard restart", where the cache is explicitly not desired (e.g. upgrade Nuxt itself). This shouldn't be a common operation as we already support "Soft restart" by config changes, etc. Performance is relevant here I think.

antfu:
Oh, and I forgot to reference, Vitest has been using this pattern for months to capture Node crashes and handle the restart: https://github.com/vitest-dev/vitest/blob/4fc492c61d6e01bc34bc807a6b57197a1914f823/packages/vitest/src/node/cli-wrapper.ts#L134-L135

It works quite nicely and we don't see issues with this approach.

danielroe (commented):
This looks good to me. What do you think @pi0? We would then have both the soft restart in `dev` (from https://github.com/nuxt/nuxt/pull/19084) and also this 'hard' restart for exceptional circumstances.

antfu:
By the discussion with @danielroe, I just pushed a commit to extend the `restart` hook to be:

```ts
await nuxt.hooks.callHook('restart', { hard: true })
```

Later, with the new CLI integration, we could keep the interface the same, making the transition invisible.

Atinux:
Looks great 🙏 

https://volta.s3.fr-par.scw.cloud/Clean_Shot_2023_03_02_at_15_57_35_005f167975.mp4

atinux (approved):

packages/nuxi/src/commands/dev.ts
```diff
@@ -89,6 +90,14 @@ export default defineNuxtCommand({
         }
 
         currentNuxt = await loadNuxt({ rootDir, dev: true, ready: false })
+        // Hard restart
+        if (process.env.NUXI_CLI_WRAPPER) {
+          currentNuxt.hooks.hook('restart', (options) => {
+            if (options?.hard) {
+              process.exit(EXIT_CODE_RESTART)
+            }
```
pi0, line 98: Refactor: We could merge it with the `restart` hook logic. Use hook once, and only hard reset when `opts.hard` ++ `NUXI_CLI_WRAPPER`

packages/nuxi/src/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Special exit code to restart the process
+ *
+ * Usage:
+ * ```ts
+ * if (process.env.NUXI_CLI_WRAPPER) {
+ *   process.exit(EXIT_CODE_RESTART)
+ * }
+ * ```
+ */
+export const EXIT_CODE_RESTART = 44
```
pi0, line 11: Although I assume this implementation is temporary, code 44 is reserved for UNIX (Channel number out of range). As we don't (and shouldn't) use code 0, i think we can simply use that code as an indicator that child was killed and expects to be restarted (otherwise use a higher number)

packages/nuxi/src/cli-wrapper.ts
```diff
@@ -0,0 +1,40 @@
...
+ * This file is used to wrap the CLI entrypoint in a restartable process.
+ */
+import { fileURLToPath } from 'node:url'
+import { execa } from 'execa'
+import { EXIT_CODE_RESTART } from './constants'
+
+const cliEntry = fileURLToPath(new URL('../dist/cli-run.mjs', import.meta.url))
+
+async function startSubprocess (preArgs: string[], postArgs: string[]) {
+  const child = await execa(
```
pi0, line 11: Perf: I guess we can use `fork` from `nod:child_process`. This is also consistent with the final nuxi-ng implementation and provides a communication channel.

packages/nuxi/src/cli-wrapper.ts
```diff
@@ -0,0 +1,40 @@
...
+    process.exit(child.exitCode)
+  }
+}
+
+const args = process.argv.slice(2)
+// only enable wrapper in dev command
+if (args[0] === 'dev') {
+  startSubprocess([], args)
+} else {
+  import(cliEntry)
```
pi0, line 39: Error is not caught for unexpected import errors

pi0 (approved):
Nice idea with introducing `restart({ hard })` hook 👍🏼

Other than mentioned comments above, i think we can move forward with it unblocking features like https://github.com/nuxt/nuxt/pull/19394 (also adding hook types in next release!)

As I have drafted in https://github.com/nuxt/cli/pull/22, the final new implementation of hard start can be completely different with the CLI server being alive across reloads. This means hard restart will have behavior differences and it might be internally aware of it.

packages/nuxi/src/constants.ts
antfu: I was checking this table: https://tldp.org/LDP/abs/html/exitcodes.html says 3-125 are free to use.

And yes, 44 was a number I picked randomly.

Go a bit deeper I found this:
https://www.cyberciti.biz/faq/linux-bash-exit-status-set-exit-statusin-bash/
It seems to me the codes are specially for bash to exit, I suppose it shouldn't have much impact on our side (+ the code is only be captured by our wrapper)

So anyway I changed it to 85 as it's defined in the previous link as `Interrupted system call should be restarted`

packages/nuxi/src/cli-wrapper.ts
antfu: I see we use `execa` in many places already, and the chunk is shared. I guess the performance won't be a big issue here.

I am just worried about using `fork` or `exec` might create more issues to handle on different OS, etc.

pi0: We could use 1 btw too. Auto restarting with child unwanted exit codes (also consistent with sighup used for restart)

pi0: Main usecase i added execa was search path compatibility which isn’t a case here. Also forke copies mem in unix then forks which by theory at least should be faster than loading node to the mem again.

packages/nuxi/src/constants.ts
pi0: We could use 1. Auto restarting with child unwanted exit codes too as well as signaling explicitly (also consistent with sighup used for restart)

antfu: I am not sure if I get what you mean - are you suggesting we exit with `process.exit(1)`? Wouldn't that be indistinguishable with the normal error exit?

packages/nuxi/src/cli-wrapper.ts
Atinux: Could we mesure the performance difference here?

I like the simplicity of code and execa has some good features:
CleanShot 2023-03-03 at 10 12 41@2x

packages/nuxi/src/constants.ts
pi0: A child normally doesn't exits with (1) but yeah it might introduce edge-cases.

Well, btw if we use fork, we can use the process communication channel to explicitly send a nuxt event (https://github.com/nuxt/cli/pull/22/files#diff-34d76a96faa930c4b9909c055beb422e601e2267f72ce12d6f15a1c2cc6ea914R57)

packages/nuxi/src/cli-wrapper.ts
pi0: Most of those benefits aren't relevant here and main benefit is using fork over exec with fine gained control. we are forking a known entry as process (not a on-off executation).

Finally this impl will be part of citty's dev server so this change is to make sure we are aligned :)

pi0:
Please next times, wait on comments to be resolved before moving forward!

PR #18642, opened 2023-01-31
fix(vite): ensure `__publicAssetsURL` set before loading assets

danielroe:
### ❓ Type of change

### 📚 Description

With upgrade of vite/rollup, we now sometimes output separate chunks for assets which reference globally registered `__publicAssetsURL` function. But it's possible via race condition for this to be executed before the function is set (typically within entry - it's a side-effect of `#build/paths.mjs`). This PR ensures that this implicit dependency is made explicit.

Here's an example of the diff for a generated chunk for a logo url.

```diff
- const o=""+globalThis.__publicAssetsURL("logo.svg");export{o as _};
+ import"./entry.68b62826.js";const o=""+globalThis.__publicAssetsURL("logo.svg");export{o as _};
```

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/paths.ts
+++ b/packages/vite/src/plugins/paths.ts
@@ -1,0 +1,27 @@
+import MagicString from 'magic-string'
+import type { Plugin } from 'vite'
+
+export interface RuntimePathsOptions {
+  sourcemap?: boolean
+}
+
+export function runtimePathsPlugin (options: RuntimePathsOptions): Plugin {
+  return {
+    name: 'nuxt:runtime-paths-dep',
+    enforce: 'post',
+    transform (code, id) {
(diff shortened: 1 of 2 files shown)


antfu (approved):

packages/vite/src/plugins/paths.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface RuntimePathsOptions {
+  sourcemap?: boolean
+}
+
+export function runtimePathsPlugin (options: RuntimePathsOptions): Plugin {
+  return {
+    name: 'nuxt:runtime-paths-dep',
+    enforce: 'post',
+    transform (code, id) {
+      if (code.includes('__VITE_ASSET__') || code.includes('__VITE_PUBLIC_ASSET__')) {
```
pi0, line 13: Small perf: this technically iterates over characters of code (how long it is), twice while a compiled regex can iterate once and being faster...

Issue #15739, opened 2022-12-01
Track install time for modules


Atinux:
This can be used later for debugging in the devtools.

Basically could be added here: https://github.com/nuxt/framework/blob/54ac7dd481355a795c2d2b3a2a8f8e06b48bbbc7/packages/kit/src/module/install.ts#L19-L22

And we calculate the time by wrapping around  `await nuxtModule(inlineOptions, nuxt)`

PR #18647, opened 2023-01-31
fix(kit): avoid adding already installed modules to internal `_installedModules`

pi0:
### 🔗 Linked issue

Reproductions:
- https://stackblitz.com/edit/github-g1pkrv?file=nuxt.config.ts
- https://github.com/Atinux/content-wind/tree/devtools

### ❓ Type of change

### 📚 Description

In situations that same module is added twice (it can be result of merging layers mainly), kit already smartly deduplicates and avoids second install of module using unique key stored in `nuxt.options._requiredModules`. However the newer `nuxt.options._installedModules` containing meta, ignores this fact since setup was not signaling it's result.

This fixes issues both for toolings depending on new `_installedModules` API and also our generated types

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/module.ts
+++ b/packages/schema/src/types/module.ts
@@ -38,7 +38,7 @@ export interface ModuleDefinition<T extends ModuleOptions = ModuleOptions> {
 /** Nuxt modules are always a simple function. */
 type Awaitable<T> = T | Promise<T>
 export interface NuxtModule<T extends ModuleOptions = ModuleOptions> {
-  (this: void, inlineOptions: T, nuxt: Nuxt): Awaitable<void|false>
+  (this: void, inlineOptions: T, nuxt: Nuxt): Awaitable<void | false>
   getOptions?: (inlineOptions?: T, nuxt?: Nuxt) => Promise<T>
   getMeta?: () => Promise<ModuleMeta>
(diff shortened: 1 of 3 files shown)


packages/schema/src/types/module.ts
```diff
@@ -36,8 +36,9 @@ export interface ModuleDefinition<T extends ModuleOptions = ModuleOptions> {
 }
 
 /** Nuxt modules are always a simple function. */
+type Awaitable<T> = T | Promise<T>
 export interface NuxtModule<T extends ModuleOptions = ModuleOptions> {
-  (this: void, inlineOptions: T, nuxt: Nuxt): void | Promise<void>
+  (this: void, inlineOptions: T, nuxt: Nuxt): Awaitable<void|false>
```
danielroe, line 41:
```suggestion
  (this: void, inlineOptions: T, nuxt: Nuxt): Awaitable<void | false>
```

danielroe (approved):

Atinux:
Could we also take the chance to add the timings?

Atinux:
Referring to #15739

pi0:
Good idea! I think would be better in a separate PR to introduce module install result object with timing info.

PR #18648, opened 2023-01-31
feat(kit, schema): measure module setup timings

### LINKED ISSUE #15739: Track install time for modules

pi0:
### 🔗 Linked issue

resolves #15739

### ❓ Type of change

### 📚 Description

- Adds new `ModuleReturn` interface for Nuxt modules with `timings` subkey
- Automatically measure kit-defined modules setup time and store in `_installedModules`
- [DX] Show a warning message if module takes more than 1 seconds to initialize

```
 WARN  Module test took 1502.4ms to setup!  
```

### 📝 Checklist


packages/kit/src/module/define.ts
```diff
@@ -67,7 +67,25 @@ export function defineNuxtModule<OptionsT extends ModuleOptions> (definition: Mo
...
     // Call setup
-    await definition.setup?.call(null as any, _options, nuxt)
+    const setupStart = process.hrtime()
+    const res = await definition.setup?.call(null as any, _options, nuxt) ?? {}
+    const setupEnd = process.hrtime(setupStart)
+
+    // Measure setup time
+    const setupTime = Math.round((setupEnd[0] * 1000000000 + setupEnd[1]) / 10000) / 100 /* round by two digits */
+    if (setupTime > 1000) {
+      console.warn(`Module \`${uniqueKey || '<no name>'}\` took \`${setupTime}ms\` to setup!`)
```
Atinux, line 77:
```suggestion
      logger.warn(`Module \`${uniqueKey || '<no name>'}\` took \`${setupTime}ms\` to setup!`)
```

Atinux: Why not using our logger directly?

pi0: They are actually same channels since we mock.

Atinux:
It seems that using an anonymous module in the config does not trigger a warn:

```ts
  modules: [
    async () => {
      // wait 2 sec
      await new Promise(resolve => setTimeout(resolve, 2000))
    }
  ]
})
```

I guess it's safe to ignore but we should advice to use `@nuxt/kit` to define Nuxt modules:

```ts
import { defineNuxtModule } from '@nuxt/kit'

  modules: [
    defineNuxtModule({
      meta: {
        name: 'my-module'
      },
      async setup () {
        // wait 2 sec
        await new Promise(resolve => setTimeout(resolve, 2000))
      }
    })
  ]
})
```

CleanShot 2023-01-31 at 18 34 45@2x

atinux (approved):
Looks good to me! Waiting for @danielroe review as well and @antfu regarding devtools integration

TheAlexLichter (approved):
LGTM. Should we document it somewhere (besides auto-generated docs)?

packages/schema/src/types/module.ts
```diff
@@ -26,19 +26,32 @@ export interface ModuleMeta {
...
+   */
+  timings?: {
+    /** Total time took for module setup in ms */
+    setup?: number
+    [key: string]: number | undefined
+  }
+}
+
+type Awaitable<T> = T | Promise<T>
+export type _ModuleReturn = Awaitable<void | false | ModuleReturn>
```
danielroe, line 42: Is there a reason we are exporting this internal type?

packages/kit/src/module/define.ts
```diff
@@ -67,7 +67,25 @@ export function defineNuxtModule<OptionsT extends ModuleOptions> (definition: Mo
...
-    await definition.setup?.call(null as any, _options, nuxt)
+    const setupStart = process.hrtime()
+    const res = await definition.setup?.call(null as any, _options, nuxt) ?? {}
+    const setupEnd = process.hrtime(setupStart)
+
+    // Measure setup time
+    const setupTime = Math.round((setupEnd[0] * 1000000000 + setupEnd[1]) / 10000) / 100 /* round by two digits */
+    if (setupTime > 1000) {
+      logger.warn(`Module \`${uniqueKey || '<no name>'}\` took \`${setupTime}ms\` to setup`)
+    }
```
danielroe, line 78: * What about printing time of each module (however short) if `nuxt.options.debug` is set?

* Playing devil's advocate, we might also consider suppressing this warning when we first release the feature until we have a chance to test with the bigger ecosystem. In patch release, we could print warning only if module is _not_ located in `node_modules`, which might give authors a chance to rewrite. (Counter-point: all modules taking this long _should_ rewrite and we don't want users thinking _nuxt_ is slow.)

Atinux: I like the idea of displaying the timing for each module when having `DEBUG=1`, we could highlight the time I guess or add a `(slow)` next to it.

packages/kit/src/module/define.ts
```diff
@@ -67,7 +67,25 @@ export function defineNuxtModule<OptionsT extends ModuleOptions> (definition: Mo
     }
 
     // Call setup
-    await definition.setup?.call(null as any, _options, nuxt)
+    const setupStart = process.hrtime()
```
antfu, line 70: Maybe we could use https://nodejs.org/api/process.html#processhrtimebigint here since it's supported in v10?

packages/kit/src/module/define.ts
antfu: I also prefer not to log them by default，it could be useful to module authors but not too much for regular users, but just make the output verbose.

In devtools, we could present them in UI with different levels of colors (yellow, orange, red) and be able to sort them - this way, authors could find out the bottleneck easier.

packages/schema/src/types/module.ts
pi0: I saw a recommendation somewhere before to always export public interfaces. ModuleSetupReturn seems good idea.

packages/kit/src/module/define.ts
pi0: Good idea will refactor! (initially was lazy to import ahah :D)

packages/kit/src/module/define.ts
pi0: Thanks for the feedback. I think we can show a little bit more verbose output for debug mode too (less threshold than 1000). Most modules are initialized less than few/1ms.

harlan-zw:
PR looks good overall timing. My two cents:

1.  Utilise debug output

Consider the debug output atm. Context is lost without knowing what module setup is taking place.

image

It would be good to add a debug on the start of module setup and on the finish. There are a couple of modules that do globs and they have a habit of accidentally scanning the entire node_modules and stalling indefinitely. (pinceau, windicss, etc). Or making network requests which are using large timeouts. 

2. Warning for slow modules

I think this just encourages module authors to "hide" logic that takes more than 1s within hooks, making it hard to debug, so I am against this (unless debug is enabled).

danielroe:
@pi0 feel free to let me know if I can be helpful here 🙏

packages/kit/src/module/define.ts
pi0: Migrated to performance api

packages/kit/src/module/define.ts
pi0: Moved threshold to 5 seconds. It is really unlikely (and undesired for me) a module alone takes this long for setup. We could probably show much better inspection from devtools from meta anyway.

pi0:
I have made a few changes. And also increased the minimum amount to 5 seconds which is unlikely to happen unless a module is doing something really wrong in which case, "end user" should be aware of why their development experience is slow. And also default hook logs, we have explicit module name as part of warning message so it is tracable. 

I think anyway using this new exported meta devtools can give better experience. And we might try it for a while across ecosystem to first **find the evil** and decide to not warn if there are modules that absolutely cannot be optimized! But without a small push to the ecosystem and user awareness by testing against different environments, it is unlikely we could ever find those kind of modules.

danielroe:
This approach seems good to me.

🚨 Note to readers: I think it is not too strong to say we are now engaged in a fight against evil across the galaxy. 🚨

danielroe (approved):

PR #9671, opened 2021-08-13
feat(webpack)!: update postcss to v8

clarkdo:
## Types of changes

- [X] New feature (a non-breaking change which adds functionality)

## Description

Update postcss to v8:

- Move to postcss-loader options
- Update css-loader, postcss-import, postcss-url and postcss-loader
- ~~Disable postcss-preset-env as it doesn't support postcss8~~
- ~~Warn message if user is using `build.postcss.preset`~~
- ~~Add autoprefixer as default plugins of postcss~~

**NOTE: this is a breaking change for `build.postcss` config:**

- ~~`postcss-preset-env` and `build.postcss.preset` have been removed as it doesn't support postcss v8~~
- `build.postcss` has been moved to [postss-loader options](https://github.com/webpack-contrib/postcss-loader#options) format:
- `build.postcss.plugins` to `build.postcss.postcssOptions.plugins`, the object format is still same
- `build.postcss.order` to `build.postcss.postcssOptions.order`

```diff
  {
+   postcssOptions: {
      plugins: {
        'postcss-import': {},
        'postcss-url': {},
        'postcss-preset-env': this.preset,
        'cssnano': { preset: 'default' } // disabled in dev mode
      },
      order: 'presetEnvAndCssnanoLast',
+   },
    preset: {
      stage: 2
    },
    // Other postcss-loader options https://github.com/webpack-contrib/postcss-loader#options
  }
```

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] I have added tests to cover my changes (if not applicable, please state why)
- [ ] All new and existing tests are passing.

PR #18704, opened 2023-02-02
feat: add back `postcss-preset-env`

danielroe:
### 📚 Description

This PR updates https://github.com/nuxt/nuxt/pull/9671 to add back `postcss-preset-env`, which is now updated for PostCSS8. This reduces the breaking change surface of the linked PR.

We might also consider reverting the breaking changes to postcss options in order to make this more manageable as a migration step. What do you think @pi0 - is there a reason we need to make the following change?

- `build.postcss` has been moved to [postss-loader options](https://github.com/webpack-contrib/postcss-loader#options) format:
  - `build.postcss.plugins` to `build.postcss.postcssOptions.plugins`, the object format is still same
  - `build.postcss.order` to `build.postcss.postcssOptions.order`

---

**Note**: to reduce the clutter of commits this is marked as a PR against `2.x` but before merge I will update to merge into https://github.com/nuxt/nuxt/pull/9671 and thence into `2.x`.


pi0 (dismissed):
Nice work on adding back compatibility with env. since we are introducing other breaking changes anyway in 2.16, postcss8 shouldn’t be a problem.

pi0:
We need to test edge against tailwindcss module.

PR #18728, opened 2023-02-03
docs: Improve routing `validate` example

aimer-fan:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/1.getting-started/5.routing.md
+++ b/docs/1.getting-started/5.routing.md
@@ -132,7 +132,7 @@ If you have a more complex use case, then you can use anonymous route middleware
 <script setup>
 definePageMeta({
   validate: async (route) => {
-    const nuxtApp = useNuxtApp()
+    const route = useRoute()
     // Check if the id is made up of digits
     return /^\d+$/.test(route.params.id)
   }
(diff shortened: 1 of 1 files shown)


docs/1.getting-started/5.routing.md
```diff
@@ -132,7 +132,7 @@ If you have a more complex use case, then you can use anonymous route middleware
 <script setup>
 definePageMeta({
   validate: async (route) => {
-    const nuxtApp = useNuxtApp()
+    const route = useRoute()
```
manniL, line 135: We can fully omit the line here as `route` is passed as argument already
```suggestion
```

TheAlexLichter (changes requested):

docs/1.getting-started/5.routing.md
aimer-fan: Thanks, I didn't notice that route is a parameter, because the previous part of the documentation mentions using the useRoute method to get the route object.

manniL: Yes, that's correct. For `validate`, the route is given though so validation operations can be performed (as described in the doc text above the example) ☺️

TheAlexLichter (approved):

PR #18743, opened 2023-02-03
docs: improve swr documentation

alanpoulain:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/603#issuecomment-1415832639

### ❓ Type of change

### 📚 Description

The current documentation of `swr` is not correct: it is not a static build but SSR + cache headers.

### 📝 Checklist


docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,8 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr`: Server-side renders sections of your app and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL.
```
danielroe, line 92: On Netlify & Vercel this uses their output APIs (Builder Functions and Build Output API, respectively) to add it to the CDN layer (so it's not just adding cache headers). And within the node server preset this caches the response _within_ the server, so it's not just adding cache headers there either.

alanpoulain: Sure I can change the documentation accordingly then,

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,8 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr`: Server-side renders sections of your app and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL. For Netlify and Vercel, the response is also added to the CDN layer. Note that the node-server preset of Nitro is able to cache the response.
+- `static`: Enables a single (on-demand) build (currently enables full incremental static generation on Netlify, with Vercel coming soon)
```
danielroe, line 93: swr and static have the same behaviour of on-demand generation - it's just that static doesn't revalidate after the first request. We could add `prerender` also as another route rule option.

alanpoulain: It's very confusing to call it `static` then IMHO. I will try to modify this part too.

alanpoulain: Are you sure it's the case? In node the only header I see is `Cache-Control: public` without `max-age`.

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,8 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr`: Server-side renders sections of your app and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL. For Netlify and Vercel, the response is also added to the CDN layer. Note that the node-server preset of Nitro is able to cache the response.
```
pi0, line 92: Node.js/Server is the default behaving and also works for any server preset including workers, etc we should move to the beginning and mention it is cached on storage layer.

Behavior of Vercel/Netlify presets that map swr to a similar implementation is exceptional part!

docs/2.guide/1.concepts/3.rendering.md
danielroe: It's somewhat more complicated, but it's definitely not _just_ adding headers. Some routes are cached internally by handler; others aren't. See:

https://github.com/unjs/nitro/blob/2c52c62f1e4adbce4f57e28ec1d4929cd2051928/src/runtime/cache.ts#L160-L333

https://github.com/unjs/nitro/blob/ed3a48026e8a27ecfb86e406ba39bdeb418f8573/src/runtime/app.ts#L55-L64

alanpoulain:
I need to add `prerender` as well too, don't you think?

docs/2.guide/1.concepts/3.rendering.md
pi0: Static is more a permanent cache strategy per deployment.

docs/2.guide/1.concepts/3.rendering.md
alanpoulain: Updated.

docs/2.guide/1.concepts/3.rendering.md
danielroe: ~> https://github.com/unjs/nitro/pull/906

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,9 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr` - Server-side renders sections of your app and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL. The node-server preset of Nitro is able to cache the response. For Netlify and Vercel, the response is also added to the CDN layer.
```
manniL, line 92: Some wording + styling suggestions

```suggestion
- `swr` - Server-side renders routes of your app and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL. The `node-server` preset of Nitro is able to cache the full response. For Netlify and Vercel, the response is also added to the CDN layer.
```

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,9 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
...
+- `static` - The behavior is the same as `swr` except that there is no TTL: the response is cached until the next deployment. On Netlify, it enables full incremental static generation. Vercel is coming soon.
+- `prerender` - Fetches your app sections and copies them as static assets during the build
```
manniL, line 94:
```suggestion
- `prerender` - Fetches your app routes and copies them as static assets during the build
```

mcrapts: What is the result of using the static property when using Node instead of Vercel/Netlify? It's not clear to me what the static property does by reading the current and proposed documentation.

alanpoulain:
Shouldn't this PR be merged?

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,9 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr` - Server-side renders routes and add cache headers in the response for the server or the reverse proxy to cache it for a configurable TTL. The `node-server` preset of Nitro is able to cache the full response. For Netlify and Vercel, the response is also added to the CDN layer.
```
danielroe, line 92:
```suggestion
- `swr` - Add cache headers to the server response and cache it in the server or reverse proxy for a configurable TTL. The `node-server` preset of Nitro is able to cache the full response. For Netlify and Vercel, the response is also added to the CDN layer.
```

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,9 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
...
+- `static` - The behavior is the same as `swr` except that there is no TTL: the response is cached until the next deployment. On Netlify, it enables full incremental static generation. Vercel is coming soon.
```
danielroe, line 93:
```suggestion
- `static` - The behavior is the same as `swr` except that there is no TTL; the response is cached until the next deployment. On Netlify and Vercel, it enables full incremental static generation.
```

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -89,7 +89,9 @@ Nuxt 3 starting from rc.12 comes with the public beta for route rules and hybrid
 - `ssr` - Disables server-side rendering for sections of your app and make them SPA-only with `ssr: false`
 - `cors` - Automatically adds cors headers with `cors: true` - you can customize the output by overriding with `headers`
 - `headers` - Add specific headers to sections of your site - for example, your assets
-- `static` and `swr` - `static` enables a single (on-demand) build; `swr` enables a static build, that lasts for a configurable TTL. (currently enables full incremental static generation on Netlify, with Vercel coming soon)
+- `swr` - Add cache headers to the server response and cache it in the server or reverse proxy for a configurable TTL. The `node-server` preset of Nitro is able to cache the full response. For Netlify and Vercel, the response is also added to the CDN layer.
+- `static` - The behavior is the same as `swr` except that there is no TTL; the response is cached until the next deployment. On Netlify and Vercel, it enables full incremental static generation.
+- `prerender` - Prerenders routes at build time and includes them in your build as static assets
```
danielroe, line 94:
```suggestion
- `prerender` - Prerenders routes at build time and includes them in your build as static assets.
```

danielroe (approved):

PR #18783, opened 2023-02-05
feat(nuxi): warn when prerendering routes with `ssr: false`

### LINKED ISSUE #12393: Documentation is not very clear regarding static site generation in Nuxt 3
### Environment

- Node Version:     `v16.19.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Builder:          `vite`
- User Config:      `ssr`, `routeRules`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/jackdomleo7/nuxt3-simple-static

### Describe the bug

The Nuxt 3 documentation regarding Nuxt 3 states that static site generation should be as simple as setting `ssr: false`, using `nuxi generate` and potentially using `routeRules`. Despite doing this, the site doesn't seem to be rendering statically.

image

This is my Nuxt config:

```ts
  ssr: false,
  routeRules: {
    '/': { static: true },
    '/**': { static: true }
  }
})
```

When you run the reproduction repo and run `npm run generate` (`nuxi generate -m`), it successfully creates the `dist` directory, but I'd expect there to be static content inside the `dist/index.html` file, e.g.

```html
<div>
  <h1>Hello, world!</h1>
  <p>I am a static site built with Nuxt 3.</p>
</div>
```

Instead, it does not.

image

Either, my understanding has defeated me or there is something missing or not quite correct in the Nuxt 3 documentation.

### Additional context

I've tried with Node 16 & Node 18, and with npm 8/9 and pnpm 7 and I get the same results.

### Logs

```shell
> nuxi generate -m

Nuxi 3.0.0                                                                                                                            08:51:03
Nuxt 3.0.0 with Nitro 1.0.0                                                                                                           08:51:03

 WARN  Using experimental payload extraction for full-static output. You can opt-out by setting experimental.payloadExtraction to false.

i Client built in 1681ms                                                                                                              08:51:07
i Building server...                                                                                                                  08:51:07
√ Server built in 888ms                                                                                                               08:51:08
√ Generated public .output/public                                                                                               nitro 08:51:08
i Initializing prerenderer                                                                                                      nitro 08:51:08
i Prerendering 3 initial routes with crawler                                                                                    nitro 08:51:09
  ├─ /index.html (18ms)                                                                                                         nitro 08:51:09
  ├─ /200.html (3ms)                                                                                                            nitro 08:51:09
  ├─ /404.html (5ms)                                                                                                            nitro 08:51:09
√ You can now deploy .output/public to any static hosting!
```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/12393

### ❓ Type of change

### 📚 Description

We can warn when users have `ssr: false` enabled but are running `nuxi generate`. This is almost always the wrong choice although it may be desirable in a small subset of cases.

I've also updated the documentation to make this clearer.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -52,6 +52,9 @@ export default defineNuxtCommand({
     await buildNuxt(nuxt)
 
     if (args.prerender) {
+      if (!nuxt.options.ssr) {
+        consola.warn('HTML content not prerendered because `ssr: false` was set.')
+      }
       // TODO: revisit later if/when nuxt build --prerender will output hybrid
       const dir = nitro?.options.output.publicDir
       const publicDir = dir ? relative(process.cwd(), dir) : '.output/public'
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -53,7 +53,7 @@ export default defineNuxtCommand({
 
     if (args.prerender) {
       if (!nuxt.options.ssr) {
-        consola.warn('HTML content not prerendered because `ssr: false` was set.')
+        consola.warn('HTML content not prerendered because `ssr: false` was set. You can read more in `https://nuxt.com/docs/getting-started/deployment#static-hosting`.')
       }
       // TODO: revisit later if/when nuxt build --prerender will output hybrid
       const dir = nitro?.options.output.publicDir
--- a/docs/1.getting-started/10.deployment.md
+++ b/docs/1.getting-started/10.deployment.md
@@ -76,8 +76,8 @@ By default, the workload gets distributed to the workers with the round robin st
 
 There are two ways to deploy a Nuxt application to any static hosting services:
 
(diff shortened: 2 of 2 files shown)


packages/nuxi/src/commands/build.ts
```diff
@@ -52,6 +52,9 @@ export default defineNuxtCommand({
     await buildNuxt(nuxt)
 
     if (args.prerender) {
+      if (!nuxt.options.ssr) {
+        consola.warn('HTML content not prerendered because `ssr: false` was set.')
```
manniL, line 56: Should we link to the docs so they can read more about that?

pi0, line 56: Small note: This log is for the future (the content "that will be prerendered")

danielroe: This is printed out after the content is already prerendered.

TheAlexLichter (approved):

PR #18800, opened 2023-02-06
test(nuxt): add `BaseTransitionPropsValidators` to auto-import exclude list

yyx990803:
This new export is added in 3.3.0-alpha and is needed to fix a tree-shaking issue.


danielroe (approved):
Thank you ❤️

PR #18862, opened 2023-02-07
fix(nuxt): defer adding route path to preloaded record

Mini-ghost:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

Currently, when there are more than 5 pages that need to be preload on the screen at the same time, only the first 5 pages will be preload, but all pages will be marked as preloaded.

That's because we marked the path for preloaded too early, once it gets delayed because there are more than 5 pages waiting to be preloaded, those pages won't be prefetched.
https://github.com/nuxt/nuxt/blob/34b58d143c17c5169a6561ea538d404ea5b12ad9/packages/nuxt/src/app/composables/preload.ts#L41-L49

Maybe we can delay marking preloaded after the promise length condition.

### 📝 Checklist


danielroe (approved):
Nice find!

PR #18864, opened 2023-02-07
feat(nuxi): add initial support for enabling/disabling devtools

danielroe:
### ❓ Type of change

### 📚 Description

This adds `nuxi enable` and `nuxi disable` commands. For extensibility, we defer the execution to a CLI provided by a core Nuxt package. We may well revisit the implementation of this - for example, not requiring a hardcoded list of features <-> packages.

### 📝 Checklist


antfu (approved):

pi0:
Overall this looks good to me really nice idea 👍🏼 

My vision for the upcoming next-gen of nuxi is to make it fully modular and if we implemented  as `nuxt devtools enable|disable` it could be much better for modularity and consistency with upcoming subcommands. (also same DX and fewer additions in this PR)

But also there are no issues if we want to use an alternative syntax as shortcuts. Only it adds two more commands to maintain as cost.

packages/nuxi/src/commands/devtools.ts
```diff
@@ -0,0 +1,25 @@
...
+    name: 'enable',
+    usage: 'npx nuxi devtools enable|disable [rootDir]',
+    description: 'Enable or disable features in a Nuxt project'
+  },
+  async invoke (args) {
+    const [command, _rootDir = '.'] = args._
+    const rootDir = resolve(_rootDir)
+
+    if (!['enable', 'disable'].includes(command)) {
+      console.error(`Unknown command \`${command}\`.`)
```
pi0, line 17: Suggestion: Do we have to do this check here? If for any reason we support an additional subcommand, it requires the nuxi to be updated while same check could happen inside devtools exposed CLI. (Sorry for nitpicking :D)

pi0: If is for usage, a fail with status code could be used if purpose is showHelp after execa. But i will do my best to make it supported directly from devtools cli. Making life much easier soon.

danielroe: Very much looking forward to seeing it! 🙏

You're right, the check here is to display usage information. When we refactor to a submodule approach it will be fully in control of devools CLI. Or we could even remove this check sooner in a patch if we had any idea of other CLI command. For now, I think it's safe.

pi0 (approved):

pi0:
Maybe we should add to the docs too? (as edge feature)

danielroe:
Let's add docs as a separate PR that we can merge after release - as this feature won't work until we release `@nuxt/devtools`, which is aimed for same time as 3.2.0. (I'll prepare docs PR.)

PR #18431, opened 2023-01-21
fix(nuxt): provide node16-style type exports

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/19606

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR makes some updates to Nuxt internals to improve support for Node16-style module resolution.

### 👉 Current status

<details>
<summary><strong>nuxt v3.2.3</strong></summary>

<table id="resolutions" class="">
<thead>
<tr>
<th></th>
<th><code>"nuxt"</code></th><th><code>"nuxt/config"</code></th><th><code>"nuxt/schema"</code></th><th><code>"nuxt/app"</code></th><th><code>"nuxt/package.json"</code></th>
</tr>
</thead>
<tbody>

<tr>
<td><code>node10</code></td>
<td>✅ </td><td>✅ </td><td>✅ </td><td>✅ </td><td>✅ (JSON)</td>
</tr>
<tr>
<td><code>node16</code> (from CJS)</td>
<td>❌ No types<br>⚠️ ESM (dynamic import only)</td><td>👺 Masquerading as ESM<br>⚠️ ESM (dynamic import only)</td><td>⚠️ ESM (dynamic import only)</td><td>❌ No types<br>⚠️ ESM (dynamic import only)</td><td>✅ (JSON)</td>
</tr>
<tr>
<td><code>node16</code> (from ESM)</td>
<td>❌ No types</td><td>✅ (ESM)</td><td>✅ (ESM)</td><td>❌ No types</td><td>✅ (JSON)</td>
</tr>
<tr>
<td><code>bundler</code></td>
<td>❌ No types</td><td>✅ </td><td>✅ </td><td>❌ No types</td><td>✅ (JSON)</td>
</tr>
</tbody></table>

</details><details>
<summary><strong>nuxt (edge)</strong></summary>

<table id="resolutions" class="">
<thead>
<tr>
<th></th>
[...]

PR #18880, opened 2023-02-08
fix: reorder the types field in package.json

HiChen404:
### ❓ Type of change

### 📚 Description
According to the typescript documentation: Entry-point for TypeScript resolution must occur first.

Same as Vue repo: https://github.com/vuejs/core/commit/957722c4185ea87aabc711f6f8997e528dd3ba1b `fix(build): ensure type exports is first`
>Reference:  
> https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

### 📝 Checklist

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -13,9 +13,9 @@
   "exports": {
     ".": "./dist/index.mjs",
     "./config": {
+      "types": "./config.d.ts",
       "import": "./config.mjs",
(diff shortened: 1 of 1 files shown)


danielroe (approved):
Thanks. Just linking here another significant fix which will apply this change more widely across the monorepo, coming in https://github.com/nuxt/nuxt/pull/18431

Issue #18879, opened 2023-02-08
nuxt preview path error


xjccc:
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.18.0`
- Nuxt Version:     `3.1.2`
- Nitro Version:    `2.1.1`
- Package Manager:  `pnpm@7.24.3`
- Builder:          `vite`
- User Config:      `typescript`, `app`, `nitro`, `experimental`, `sourcemap`, `build`, `runtimeConfig`, `css`, `vite`, `ignore`, `hooks`, `modules`, `errorCacheConfig`
- Runtime Modules:  `nuxt-error-and-cache@1.0.4`, `nuxt-custom-fetch@1.0.7`, `@vueuse/nuxt@9.12.0`, `@pinia/nuxt@0.4.6`
- Build Modules:    `-`

### Reproduction

in nuxt.config.ts
`
nitro: {
    output: {
      dir: '~/node-server'
    }
  }
`

### Describe the bug

the server code in node-server
then use preview
the error is
Cannot find nitro.json. Did you run nuxi build first? Search path:
 [
  '~/.output/nitro.json',
  '~/nitro.json'
]

danielroe:
I think you should be able to run `nuxi preview node-server` in this case - does this work for you?

xjccc:
> 

yes . i know this thanks😁

xjccc:
> I think you should be able to run `nuxi preview node-server` in this case - does this work for you?

yes, i know this thanks😁

PR #18882, opened 2023-02-08
fix(nuxi): preview nitro build with custom dir config

xjccc:
### 🔗 Linked issue
#18879 

### ❓ Type of change

### 📚 Description

when config nitro.output.dir
command nuxi preview is error the path

i have config the path from nitro.output.dir in preview

is this the right way to deal with the problem?

### 📝 Checklist


danielroe:
I'm reluctant to merge this as-is because it adds a bit of overhead to the preview command, which is really just wrapper anyway to running a command specified in the output folder's `nitro.json`.

Given that there's already a way to do this (https://github.com/nuxt/nuxt/issues/18879#issuecomment-14222718500) I wonder if this is needed. cc: @pi0

xjccc:
> I'm reluctant to merge this as-is because it adds a bit of overhead to the preview command, which is really just wrapper anyway to running a command specified in the output folder's `nitro.json`.
> 
> Given that there's already a way to do this ([#18879 (comment)](https://github.com/nuxt/nuxt/issues/18879#issuecomment-14222718500)) I wonder if this is needed. cc: @pi0

yes.it has a way to deal the problem.(nuxi preview dir)
it is my simple idea.
i hope deal better.
thank u

pi0:
Preview (same as `dev`) is local. Overhead is probably not much noticeable. However it won't cover all custom presets, it is also not a bad idea to load configuration for preview improving DX of it with custom config support.

packages/nuxi/src/commands/preview.ts
```diff
@@ -16,8 +17,11 @@ export default defineNuxtCommand({
   async invoke (args) {
     process.env.NODE_ENV = process.env.NODE_ENV || 'production'
     const rootDir = resolve(args._[0] || '.')
+    const { loadNuxtConfig } = await loadKit(rootDir)
+    const config = await loadNuxtConfig({ cwd: rootDir })
+    const serverOutputDir = config.nitro.output?.dir?.replace(/^~\//, '') || '.output'
```
pi0, line 22: Probably not supporting `~` (it is srcDir while config is supposed to be from rootDir/workspaceDir)

pi0 (approved):

danielroe (approved):

xjccc:
@danielroe @pi0 
thank you very much. this is my first pr(however i worked less [smile]),
but its very encouraging for me. i hope i can make more contributions in the future.
once again, thank you for your support and appreciate your efforts and hard work.

Issue #18567, opened 2023-01-27
show 404 error page when client-side navigation fails due to `validate()` function


noriyuki-shimizu:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.1.0`
- Package Manager:  `pnpm@7.26.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

In the pages vue file, I used definePageMeta and declared validate.

* pages/example/index.vue
```ts
definePageMeta({
  validate: (route) => {
    return false
  }
})
```

At that time, if you try to page transition (NuxtLink) with a URL that violates the validate constraint, there is no response.

* pages/index.vue
```vue
<NuxtLink to="example">Example page</NuxtLink>
```

Also the middleware defined in the pages vue file and the global middleware doesn't work either.

* pages/index.vue
```ts
definePageMeta({
  validate: (route) => {
    return false
  },
  middleware: (to) => {
    console.log('Do not work.')
  }
})
```

* middleware/example.global.ts
```ts
export default defineNuxtRouteMiddleware((to) => {
  console.log('Do not work.')
})
```

### Describe the bug

I think that the processing in `packages/nuxt/src/pages/runtime/validate.ts` is probably doing something wrong.

```ts
import { defineNuxtRouteMiddleware } from '#app'

export default defineNuxtRouteMiddleware(async (to) => {
  if (!to.meta?.validate) { return }

  const result = await Promise.resolve(to.meta.validate(to))
  if (result === true) {
    return
  }
  return result
})
```

I think the problem will be solved if showError is returned only when there is an error.

```ts
import { showError, defineNuxtRouteMiddleware } from '#app'
export default defineNuxtRouteMiddleware(async (to) => {
  if (to.meta?.validate) {
    const result = await Promise.resolve(to.meta.validate(to))
    if (result === false) {
      return showError({ statusCode: 404, message: 'This page could not be found' })
    }
  }
})
```

If there is no problem with this fix, I will file a PR, is there any problem?

### Additional context

I am currently upgrading from Nuxt2 to Nuxt3.
If there is a problem that the screen transition is not possible and the middleware does not work when the validate limit is violated, the amount of modification of the project will be very large.:bow:

danielroe:
[...]

PR #18886, opened 2023-02-08
fix(nuxt): validate

noriyuki-shimizu:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/18567

### ❓ Type of change

### 📚 Description

In the vue file of pages, this change is made so that the validate of `definePageMeta` will function properly.

The current problems are considered to be the following three points.

* When the validate limit is violated, the transition from `NuxtLink` to the 404 screen is not possible.
* If the validate of `definePageMeta` is defined in the vue file of pages, the global middleware and the middleware of `definePageMeta` do not work.
* The existing `packages/nuxt/src/pages/runtime/validate.ts` process returned false if the validate limit was violated in the middleware, which caused the screen transition to fail.

This is a solution to the above problems.

https://github.com/nuxt/nuxt/issues/18567

### 📝 Checklist

  - Currently, it was not working as described in the documentation, so this modification is to make it work as described in the documentation. Therefore, no checks have been made.

### DIFF
--- a/packages/nuxt/src/pages/runtime/validate.ts
+++ b/packages/nuxt/src/pages/runtime/validate.ts
@@ -1,11 +1,9 @@
-import { defineNuxtRouteMiddleware } from '#app'
-
+import { showError, defineNuxtRouteMiddleware } from '#app'
 export default defineNuxtRouteMiddleware(async (to) => {
-  if (!to.meta?.validate) { return }
-
-  const result = await Promise.resolve(to.meta.validate(to))
-  if (result === true) {
-    return
+  if (to.meta?.validate) {
+    const result = await Promise.resolve(to.meta.validate(to))
+    if (result === false) {
+      return showError({ statusCode: 404, message: 'This page could not be found' })
+    }
   }
-  return result
 })


danielroe (changes requested):
Thank you for spending time on this PR.

At the moment this will always show an error page if `validate` returns false. But if you have multiple pages matching a route, then this will wrongly show an error page even when there is a valid route:

```
pages/[product].vue // we have a validate function here to check
pages/[...slug].vue
```

In this case, if `[product].vue` returns false, we will never move on to `[...slug].vue`, which would be the correct behaviour.

I believe everything is working correctly at the moment, except for the issue of client-side navigation to an invalid route where no other routes match, as [my comment here](https://github.com/nuxt/nuxt/issues/18567#issuecomment-1415815122) states (note that we did ship a fix in 3.1.0 (https://github.com/nuxt/framework/pull/9180). (If not, it would be very helpful if you would comment in that issue with an example of what isn't working.)

Would you like to update this PR with a different method, for example by using a router lifecycle hook or maybe updating the plugin where we run middleware instead?

noriyuki-shimizu:
Thanks for your comment!

I did not have a good grasp of what the issue was about.
Sorry about that.
I will organize the content and comment again on the following issue!

https://github.com/nuxt/nuxt/issues/18567

danielroe:
@noriyuki-shimizu Fantastic! I have put together a possible solution at https://github.com/nuxt/nuxt/pull/18978 and would value your feedback. If you think it still needs more work, perhaps we could do a joint PR, or you could make a PR to my branch?

noriyuki-shimizu:
@danielroe 
Thank you for responding!
I just checked the operation with the specified PR and this issue was cured!
Therefore, we have no additional requests, etc.
Thanks for the quick fix!

Having confirmed that this issue has been fixed, I am closing this PR!

Issue #18849, opened 2023-02-07
Page middleware is ignored when using an app/router.options.ts based router


VladStepanov:
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.20.0`
- Nuxt Version:     `3.1.2`
- Nitro Version:    `2.1.2`
- Package Manager:  `yarn@1.22.10`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `app`, `css`, `modules`, `build`, `vite`
- Runtime Modules:  `@nuxtjs/device@3.0.0`, `nuxt-proxy@0.4.1`, `@pinia/nuxt@0.4.6`, `@vueuse/nuxt@9.12.0`
- Build Modules:    `-`

### Reproduction

`app/router.options.ts` based routing https://stackblitz.com/edit/nuxt-starter-ds9s8d

same example with fs based routing https://stackblitz.com/edit/nuxt-starter-aamthy

### Describe the bug

When using `app/router.options.ts` based routing, middleware used in pages are not even called.
In reproduction example we have simple middleware in `middleware/privateRoute.ts` with only `console.log` inside.
`pages/index.vue` uses middleware above, but when we enter that page, nothing logged.

But if we rename `app -> _app` (just to enable fs routing) and restart project, then every enter to page is logged.

VladStepanov:
I've found a cause and workaround.
`definePageMeta` writes meta information in every route like
```js
definePageMeta({
  middleware: ['private-route']
})
```
In final routes config looks like:
```js
{
  path: '/',
  name: '',
  component: () => import(''),
  meta: {
    middleware: ['private-route'],
  },
```.

So i write a workaround for merging original route and our.

```js
import { merge } from 'lodash-es'

export default <RouterOptions>{
  routes (originalRoutes) {
    return [
      {
        path: '/',
        name: '',
        component: () => import(''),
      },
    ].map(route => withMeta(originalRoutes, route as RouteRecordRaw))
  },
}

function withMeta (originalRoutes: Readonly<RouteRecordRaw[]>, route: RouteRecordRaw) {
  const originalRoute = routes.find(r => r.path === route.path)

  return merge(originalRoute, route)
}
```

This code will copy all original fields for every matched route.
I'm ready for contributing, but i need help where the router config is constructing.

PR #18887, opened 2023-02-08
fix(nuxt): add route.meta for custom routing

dsfcoll:
### 🔗 Linked issue
#18849 

### ❓ Type of change

### 📚 Description

When adding routes through `app/router.options.ts`, meta information about page is lost. Even if page has `definePageMeta`
```js
definePageMeta({
  middleware: 'foo'
})
```
All metadata is ignored. Middleware won't be called.

Issue: #18849 

### 📝 Checklist

- [ ]  I have updated the documentation accordingly.

### DIFF
--- a/packages/nuxt/src/pages/runtime/plugins/router.ts
+++ b/packages/nuxt/src/pages/runtime/plugins/router.ts
@@ -7,7 +7,8 @@ import {
   createRouter,
   createWebHistory,
   createMemoryHistory,
-  createWebHashHistory
+  createWebHashHistory,
+  RouteRecordRaw
 } from 'vue-router'
 import { createError } from 'h3'
 import { withoutBase, isEqual } from 'ufo'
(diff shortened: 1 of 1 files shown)


danielroe (commented):
My main concern here is that it is possible to implement this in userland as you've done (so it's not necessary to implement in core) and this also makes assumptions that the returned routes will match 1:1 with the existing route paths. But this may be the thing that changes (as in an i18n routing implementation), or a user may wish to swap routes - for example, updating which page is the new home page.

Particularly as this is already an escape hatch for full customisation, I think it's safer to provide the existing routes with metadata to the user and let them decide what they will be doing with that metadata.

(One final comment: if we wanted to do this, we probably would also need to search and augment the child routes also.)

VladStepanov:
:danielroe Well, then, it should be written in docs that `definePageMeta` does not work with custom routing and developer have to provide route.meta fields

danielroe:
Thanks for flagging this up! I've raised a PR at https://github.com/nuxt/nuxt/pull/18901.

PR #18908, opened 2023-02-09
fix(types): import `Location` from `vue-router`

h-kishi:
### ❓ Type of change

### 📚 Description

Fixed `vue-router`'s `Location` type was not imported and global `Location` type was used for `redirect` argument type.

https://github.com/nuxt/nuxt/blob/2.x/packages/types/app/index.d.ts#L60-L61

### 📝 Checklist

### DIFF
--- a/packages/types/app/index.d.ts
+++ b/packages/types/app/index.d.ts
@@ -1,7 +1,7 @@
 import type { ServerResponse } from 'http'
 import type { IncomingMessage, NextFunction } from 'connect'
 import type Vue from 'vue'
-import type { Route } from 'vue-router'
+import type { Location, Route } from 'vue-router'
 import type { Store } from 'vuex'
 import VueRouter from 'vue-router'
 


danielroe (approved):
Thanks!

tjhiggins:
Having an issue now with the redirect type not supporting another route:
image

danielroe:
It should already be published.

tjhiggins:
> It should already be published.

My bad. For whatever reason it didn't show up when I ran `npm outdated`, but I see on npm page. Thanks.

blimmer:
I was also confused about the type packages not being published - it looks like they're not being marked as `latest` anymore, so tools like `npm outdated`, `npm-check-updates`, `yarn upgrade-interactive`, etc. don't show an upgrade past 2.16.0 as available.

Screenshot 2023-03-13 at 18 09 12

Since v3 doesn't require the types package, could we keep publishing types for 2.x as `latest`?

danielroe:
Indeed. That's already been fixed in the release script and the next version will be tagged as latest.

Issue #14146, opened 2022-06-10
pnpm support without `--shamefully-hoist`


pi0:
Update: Since v3.0 release, pnpm without `shamefully-hoist` flag works out of the box but there are still known issues to be addressed (in nuxt or upstream):

- When pages enabled, `vue-router` dependency is implicit
- Multiple instances of vue are can be resolved and used
- Generated tyeps are not fully functional without hoisted `node_modules`

----

History: nuxt/nuxt.js#11823

We have progressively improved plug-and-play support over time and almost all underlying packages including nitro support and are tested against pnpm. 

Last step is making nuxt 3 working with pnpm out of the box before migrating modules and templates from first-class yarn to pnpm.

Using pnpm without `--shamefully-hoist` is possible today 🥳 but requires explicitly installing `vue` and `ufo` dependencies and we don't want this since Nuxt was always a zero-config package and we need to add deps of all nuxt modules as well this way.

One obvious solution is that same as Nitro, we internally (also for `unimport` ~> https://github.com/unjs/unimport/issues/69) resolve them to their full path but we also need to support non-resolved implicit imports in templates as modules might have similar implicit dependencies in their scope and they need to be resolved as well. I think this can be done using a custom resolver for vite (and webpack).

kettanaito:
I've opened a pull request that fixes an issue with the internal `Module.createRequire()` call in https://github.com/nuxt/nuxt/pull/18917. It may also help with this issue as creating the right require context will likely cause Node to resolve 4th party dependencies from the PNPM store instead of always looking them up from the root-level `node_modules`, where they won't exist when using PNPM.

Dante-dan:
From my test results, it appears that if you do not want to use `--shamefully-hoist` (especially in MonoRepo), you can specify certain packages that need to be hoisted.

example
```
// .npmrc
public-hoist-pattern[]=*vue*
public-hoist-pattern[]=*nuxt*
public-hoist-pattern[]=defu
public-hoist-pattern[]=ufo
public-hoist-pattern[]=h3
public-hoist-pattern[]=nitropack
```

nicolaspayot:
We are successfully using `pnpm` without `--shamefully-hoist` in our Monorepo by explicitly adding the following `dependencies` in each nuxt application:
```json
"@vue/reactivity": "3.2.47",
"@vue/runtime-core": "3.2.47",
"@vue/runtime-dom": "3.2.47",
"@vue/shared": "3.2.47",
[...]

PR #18917, opened 2023-02-09
fix: use "import.meta.url" for internal require

kettanaito:
### 🔗 Linked issue

There's no related issue currently open in this repo. Given the scope of this change, I propose to treat this pull request as both the issue report and the proposed resolution. 

This change, however, may partially address #14146, since using `import.meta.url` will create a proper parent module context when requiring modules, which will let Node lookup 4th party dependencies from the PNPM store. 

### ❓ Type of change

I don't have enough context of Nuxt to assert whether this change is breaking or not. The intention behind this change is to keep whichever internal intention there was behind this `createRequire()` call. From the consumer's standpoint, this shouldn't be a breaking change, so I'm marking it as a bug fix. 

### 📚 Description

The internal usage of `Module.createRequire()` in `cjs.ts` is incorrect. Per [Node.js documentation](https://nodejs.org/api/module.html#modulecreaterequirefilename), the `filename` argument must be a _file name_. Passing `process.cwd()` does not yield a file name but a _directory path_. Providing incorrect value may affect the module resolution as the parent module filename will be incorrect. 

To respect the Node specification, I propose to use `import.meta.url` instead, which would point to the current module of `cjs.ts`. 

I suspect that the intention here is to create a `require()` that would always resolve modules from the project's root level (thus the CWD). If that's the case, we should still consider using a proper filename, such as a reference to `package.json` perhaps. 

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/cjs.ts
+++ b/packages/nuxi/src/utils/cjs.ts
@@ -15,7 +15,7 @@ export function getModulePaths (paths?: string | string[]): string[] {
     .filter(Boolean) as string[]
 }
 
-const _require = createRequire(process.cwd())
+const _require = createRequire(import.meta.url)
 
 export function resolveModule (id: string, paths?: string | string[]) {
   return normalize(_require.resolve(id, { paths: getModulePaths(paths) }))


packages/nuxi/src/utils/cjs.ts
```diff
@@ -15,7 +15,7 @@ export function getModulePaths (paths?: string | string[]): string[] {
     .filter(Boolean) as string[]
 }
 
-const _require = createRequire(process.cwd())
+const _require = createRequire(import.meta.url)
```
kettanaito, line 18: Despite the name `cjs`, I can clearly see that Nuxt ships this chunk as a shared module imported in `cli.mjs` so I assume this is meant to run as ESM.

Let me know if this is a wrong assumption and I should use `__filename` here instead.

pi0 (changes requested):
This doesn't look a fix. `import.meta.url` in the context of `nuxi` package, which refers to where the CLI module exists. While we need to require/import from user's directory which we change cwd to rootDir

kettanaito:
@pi0, could you please share more context on why you need to force this require from the user's root? All that the custom `filename` will affect is the lookup tree of any dependencies you require via this created `require` function. By specifying nuxi CLI as the filename, Node will be able to resolve dependencies even in stores that do not have flat structure, like PNPM. Forcing require's from the root break that, as PNPM doesn't install 4th party dependencies in the root-level node_modules (nuxi (3rd) -> `createRequire()` -> always requires 4th-party dependencies). That's the reason #14146 happens. 

From my limited understanding of nuxi, there isn't any change to the behavior of this require apart from allowing Node to work its way up the dependency tree to resolve 4th-party dependencies, which would be correct.

pi0:
Nuxi CLI is a zero-dependency package that can also be installed globally. Only use case of internal `_require` inside nuxi, passes and overrides with explicit search paths ([in `resolveModule`](https://github.com/nuxt/nuxt/blob/00087b014455b719f30cd135e3f37bceb91be879/packages/nuxi/src/utils/cjs.ts#L21)). Using `import.meta.url` isn't effective at all nor relevant to #14146. (nuxi works as expected byitself. other issues are related to nuxt builder namespace)

I would appreciate to investigate any possible changes if there is a specific issue with reproduction.

PR #18948, opened 2023-02-10
fix(nuxt): avoid recursive import in nitro renderer

### LINKED ISSUE #18789: Circular dependency in nuxt with nitropack
### Environment

- Node Version:     `v16.17.0`
- Nuxt Version:     `3.1.2`
- Nitro Version:    `2.1.1`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

and stackblitz 

### Reproduction

https://stackblitz.com/edit/nuxt-starter-kqf57y?file=README.md

### Describe the bug

Hi :wave:  !
Nuxt build in 3.1.2 logs a warning about circular dependency

### Additional context

_No response_

### Logs

```shell-script
`
Export "useNitroApp" of module "node_modules/nitropack/dist/runtime/app.mjs" was reexported through module "node_modules/nitropack/dist/runtime/index.mjs" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
Either change the import in "node_modules/nuxt/dist/core/runtime/nitro/renderer.mjs" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
````
```

pi0:
### 🔗 Linked issue

resolves #18789

### ❓ Type of change

### 📚 Description

Nuxt `renderer.ts` depending on `#internal/nitro` exporting both `useNitroApp` (from nitro/runtime/app) and `defineRenderHandler` (from nitro/runtime/renderer which is ALSO depending on runtime/app). This makes a recusive dependency which rollup was supposed to solve, but well, it doesn't for time being :} 

### 📝 Checklist


danielroe (approved):
This looks good to me.

One comment: within nuxt we are now bound to the file structure of nitro despite the `#internal` prefix, which means we probably need to use a `~` constraint for nitropack unless we can guarantee not to change this structure until the next major release.

A second concern about this warning more generally - what about auto-imports for `useNitroApp` (or Nuxt composables from `#app`) - could that cause this to fail, again, in an end-user project, in certain situations? Options to consider:

* Auto-importing instead from the specific file rather than the index.
* Bundling some runtime directories into a single file.

wdyt?

pi0:
I tested, it won't happen for user-land usecases. Both auto-imported `useNitroApp` and importing from `#imports` works fine (nitro has fallbacked `#imports` too with auto imports disabled). 

`#internal/app` is safe but we should indeed think to switch to `#imports` once was sure is safe enough. It also gives typescript features.

PR #18959, opened 2023-02-11
fix(nuxi): pass through exit code from test errors

### LINKED ISSUE #18958: Nuxi Test will not fail on CI
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.2.0`
- Nitro Version:    `2.2.1`
- Package Manager:  `pnpm@7.25.0`
- Builder:          `vite`
- User Config:      `modules`, `nitro`, `build`, `css`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Reproduction: https://stackblitz.com/edit/nuxt-starter-el5s4f

Call pnpm test:ci to see nuxi test exits with exit code 0 instead of 1

### Describe the bug

Running my nuxt3 tests with `nuxi test` on github actions with vitest will not fail. The problem is that the process always exists with exit code 0 

Example of a failed test but CI passing: https://github.com/piscis/nuxt-3-nuxi-fails-ci/actions/runs/4151622464/jobs/7182035314#step:6:52

### Additional context

vitest-node which nuxi tests calles sets the exitCode to 1 like process.exitCode = 1 this is ignored by nuxi test cli

### Logs

_No response_

piscis:
### 🔗 Linked issue

resolves #18958 

### ❓ Type of change

### 📚 Description

I noticed that my CI will always pass even if I have errors in my nuxt test scripts. Digging through the code of nuxi test command and vitest/node I discovered that errors are displayed but nuxi test will not propagate the test outcome and always exit with exitCode 0  which makes CI pass. 

This should resolve #18958 

Reproduction: https://stackblitz.com/edit/nuxt-starter-el5s4f

Example of a failed test but CI passing: https://github.com/piscis/nuxt-3-nuxi-fails-ci/actions/runs/4151622464/jobs/7182035314#step:6:52

### 📝 Checklist


packages/nuxi/src/cli.ts
```diff
@@ -74,7 +74,11 @@ export function main () {
       if (result === 'error') {
         process.exit(1)
       } else if (result !== 'wait') {
-        process.exit(0)
+        if (process.exitCode && process.exitCode >= 1) {
+          process.exit(process.exitCode)
+        } else {
+          process.exit(0)
+        }
```
manniL, line 81: Is there any reason not to always all `.exit` with the exitCode?

```suggestion
        process.exit(process.exitCode)
```

piscis: I thing setting the exit code can be emitted all together because the default is 0 and then other processes can overwritten by process.exitCode. That’s what vitest tries todo.

piscis: From the manual: Specifying a code to [process.exit(code)](https://nodejs.org/api/process.html#processexitcode) will override any previous setting of process.exitCode.

TheAlexLichter (approved):

danielroe (approved):

Baroshem (approved):

danielroe (approved):

PR #18967, opened 2023-02-11
docs: flatten top-level project in layers example

AloisSeckar:
### 🔗 Linked issue
#18954

### ❓ Type of change

### 📚 Description
As it shows up in the linked discussion, the current example for [Nuxt layers](https://nuxt.com/docs/guide/going-further/layers#basic-example) may cause a bit confusion, because the prefix `app/` used to identify "files in the root directory" may be understood as an actual sub-directory instead.

Simply omitting `app/` in the code examples would make clear what are the docs talking about.

This is proably the only place in the docs, where such prefix is used, files are adressed as `nuxt.config.ts` and `app.vue` without anything else all around, so it will also normalize the usage.

### 📝 Checklist


atinux (approved):

TheAlexLichter (approved):

docs/2.guide/3.going-further/7.layers.md
```diff
@@ -25,15 +25,15 @@ Additionally, certain other files in the layer directory will be auto-scanned an
 
 ::code-group
 
-  ```ts{}[app/nuxt.config.ts]
+  ```ts{}[nuxt.config.ts]
   export default defineNuxtConfig({
     extends: [
       '../base'
```
pi0, line 31: This should be `./base` then since we assume app is flattened

pi0:
```suggestion
      './base'
```

danielroe (approved):

PR #19001, opened 2023-02-13
fix(nitro): ensure ssr error `statusCode` is a number

### LINKED ISSUE #18992: SSR errors in `error.vue` have string statusCode but client errors are numbers
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.2.0`
- Nitro Version:    `2.2.1`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ugcjap?file=error.vue

### Describe the bug

Props Error on error page has String type in statusCode after abortNavigation, but must be Number

### Additional context

_No response_

### Logs

_No response_

clarkdo:
### ❓ Type of change

### 📚 Description

Fix https://github.com/nuxt/nuxt/issues/18992.

For keeping consistent type on ssr error, statusCode should be number, so this pr is normalizing the string value from the error getting by querystring.

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -157,17 +157,6 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
   return ctx
 }
 
-const normalizeSsrError = (ssrErrorQuery: ReturnType<typeof getQuery>) => {
-  const { statusCode, ...ssrErrorStringProps } = ssrErrorQuery
-  const ssrError = ssrErrorStringProps as unknown as Exclude<NuxtApp['payload']['error'], Error>
-
-  if (ssrError && typeof statusCode === 'string') {
-    ssrError.statusCode = parseInt(statusCode)
-  }
-
-  return ssrError
-}
-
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -157,6 +157,17 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
   return ctx
 }
 
+const normalizeSsrError = (ssrErrorQuery: ReturnType<typeof getQuery>) => {
```
clarkdo, line 160: The test is failing because the code added extra bundle size, do you want me:

- Reduce the amount of code
  - Remove the new added function and just parse statusCode in defineRenderHandler
- Or modify the threshold

danielroe: Let's try just parsing the statusCode in the render handler 🤞

If that doesn't work, then yes, we can bump the threshold...

clarkdo: It seems that 90000 is matching the minimum size of the current bundle, I've moved statusCode casting inline and increase threshold slightly.

danielroe: sorry for the pain there - this looks great 👌

danielroe (approved):

PR #19008, opened 2023-02-13
perf(nuxt): don't include side-effects from `#components`

### LINKED ISSUE #18997: NuxtWelcome is present in build
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.6.0`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.1.1`
- Package Manager:  `pnpm@7.11.0`
- Builder:          `vite`
- User Config:      `alias`, `extends`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Didn't manage to reproduce on https://stackblitz.com/

### Describe the bug

Component [NuxtWelcome](https://github.com/nuxt/ui/blob/main/packages/templates/templates/welcome/index.html) uses nowhere. But it is present in build. 
image
image

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/18997

### ❓ Type of change

### 📚 Description

When importing directly from `#components` we were not tree-shaking out unused components. ~~There is one remaining issue - there are extra styles generated in the server build but this is less significant and we can resolve in a follow-up PR.~~

~~We might create an eslint rule to _require_ the use of PURE before `defineAsyncComponent` and `defineComponent` - or maybe there is another solution?~~

~~**TODO**: investigate whether this happens with user components and if there is a way to tell rollup directly about these functions rather than annotating with content.~~

### 📝 Checklist


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -265,7 +265,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
   }) as unknown as DefineComponent<NuxtLinkProps>
 }
 
-export default defineNuxtLink({ componentName: 'NuxtLink' })
+export default /* @__PURE__ */ defineNuxtLink({ componentName: 'NuxtLink' })
```
manniL, line 268: Out of curiosity, is the `@` instead of the `#` intended here? ☺️

danielroe: it is not! thank you!

antfu (approved):

TheAlexLichter (approved):

Baroshem (approved):

packages/nuxt/src/app/components/nuxt-link.ts
danielroe: (though either is valid)

danielroe:
Would appreciate a re-review if you have a chance as I've significantly changed the implementation here. (The issue was that although the previous approach of manually adding comments was successful within Nuxt repo, if users or modules add more components, the comment would need to be added manually, and that's a pain.)

I've added a plugin to automatically inject comments before `defineComponent`, `defineAsyncComponent`, `defineNuxtLink` and `createClientOnly` which were otherwise having the effect of preserving the entire object passed to them within the build.

Moreover, I also mark vue files as not generating side effects within the server build, which allows the server styles to be tree-shaken out if the vue file is not present in the final bundle, which was the one todo I still hadn't previously addressed.

Comments welcome.

antfu:
This is smart! I would even love to have a similar standalone plugin, might be helpful for library building like VueUse 👀

danielroe:
\<building it now\>

danielroe:
~> https://github.com/danielroe/rollup-plugin-pure

PR #19019, opened 2023-02-14
feat(nuxi): add quiet mode support

### LINKED ISSUE #18497: Quiet mode
### Describe the feature

In Nuxt 2 you could make the logs quiet by setting `build.quiet` to `true` in `nuxt.config.js`. But I can't find such an option for Nuxt 3. I use it programmatically a lot and want to mute the logs.

### Additional information

### Final checks

### LINKED ISSUE #18656: CLI commands quiet mode
### Describe the feature

Nuxt v2 had the `-q` flag on the `build` command which only outputed errors if happened.
That would be nice to have for Nuxt v3 - usefull for CI/CD builds.

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves #18656
resolves https://github.com/nuxt/nuxt/issues/18497

### ❓ Type of change

### 📚 Description

This adds a basic `quiet` option that can be set in config or passed via `-q` or `--quiet` option in nuxi that will set vite logLevel to silent.

**TODO**:
- (unrelated to this PR) investigate why we have no logging in webpack at the moment

**For review**:
- consider whether to default _always_ to false (vs enabling in CI/test/without TTY)

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/build.ts
+++ b/packages/schema/src/config/build.ts
@@ -1,7 +1,7 @@
 import { defineUntypedSchema } from 'untyped'
 import { defu } from 'defu'
 import { join } from 'pathe'
-import { isCI, isTest, hasTTY } from 'std-env'
+import { isMinimal } from 'std-env'
 
 export default defineUntypedSchema({
   /**
@@ -49,7 +49,7 @@ export default defineUntypedSchema({
    */
   quiet: {
     $resolve: async (val, get) => {
-      return val ?? (isTest || isCI || !hasTTY)
+      return val ?? (isMinimal)
     }
   },
 
--- a/packages/schema/src/config/build.ts
+++ b/packages/schema/src/config/build.ts
@@ -1,6 +1,7 @@
 import { defineUntypedSchema } from 'untyped'
 import { defu } from 'defu'
 import { join } from 'pathe'
+import { isCI, isTest, hasTTY } from 'std-env'
 
 export default defineUntypedSchema({
   /**
@@ -36,6 +37,20 @@ export default defineUntypedSchema({
         client: await get('dev')
       })
     },
+  },
+
+  /**
+   * Whether or not to suppress non-essential logs when building Nuxt.
+   *
+   * Defaults to true when running in CI or when a TTY is not available.
+   *
+   * @type {boolean}
+   * @default {false}
+   */
+  quiet: {
+    $resolve: async (val, get) => {
+      return val ?? (isTest || isCI || !hasTTY)
+    }
   },
 
   /**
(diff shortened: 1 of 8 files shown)


packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,20 @@ export default defineUntypedSchema({
...
 
+  /**
+   * Whether or not to suppress non-essential logs when building Nuxt.
+   *
+   * Defaults to true when running in CI or when a TTY is not available.
+   *
+   * @type {boolean}
+   * @default {false}
+   */
+  quiet: {
```
pi0, line 50: We could generalize this option to something like `logLevel` (defaulting to quite for CI or when CLI specifies)

packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,20 @@ export default defineUntypedSchema({
...
+    $resolve: async (val, get) => {
+      return val ?? (isTest || isCI || !hasTTY)
```
pi0: Lack of TTY can happen for many fallback CLIs as well (originally, it was because of WebStorm console output). We probably shouldn't assume no-tty as silent at least for default

packages/schema/src/config/build.ts
danielroe: agreed 👍

packages/schema/src/config/build.ts
Baroshem: What are your thoughts about defaulting to `false` then? If user sets the value, then this value will be set globally, otherwise the false will be set

pi0: Makes sense. We could also only disable for test environments

packages/schema/src/config/build.ts
pi0: Yes! We can define levels using an internal enum

Baroshem: We talked about it with @danielroe and we came up with an idea to only work with quiet. If the user wants to have different type of logs, he can still use it through vite/webpack configuration in nuxt.config.ts.

What do you think Pooya?

packages/schema/src/config/build.ts
Baroshem: I am not sure about it. I would keep it disabled globally by default and just give users the option to enable it when needed (for example based on the value of environment variable)

pi0: We do disable logs for built-in tests currently by default. `isTest` only makes sense to be only for consistent behavior.

```suggestion
      return val ?? (isTest)
```

packages/schema/src/config/build.ts
danielroe: (This conversation was early on.) I think supporting multiple log levels is good, with escape hatch of direct vite/webpack configuration.

danielroe:
For next steps, as we planned when we paired, @Baroshem will be either opening a fresh PR with further changes or opening a PR to this one...

danielroe:
closing in favour of https://github.com/nuxt/nuxt/pull/19369

PR #19038, opened 2023-02-14
feat(nuxt): add experimental `app:chunkError` hook and reload strategy

### LINKED ISSUE #14594: Auto reload page when chunk assets are not found
References: nuxt/nuxt.js#14591

We had this feature in Nuxt 2. Basically what happens is with a new deployment, hashes of old js chunks will be changed. And if users already loaded the page with manifest and chunks of the last deployment, they fail when navigating on client-side that tries to load dynamic chunks.

There are several ways to resolve this issue but probably the safest and easiest, is that same as Nuxt 2 we make use of a global error handler that identifies such errors and reload pages. (Note for implementation: We should set a LocalStorage key to avoid infinity page reloads)

Bonus: We could new offer a new `nuxtApp` hook such as `app:chunkError` that allows graceful HMR alike handling by end-user.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14594

### ❓ Type of change

### 📚 Description

This PR adds a feature that can be enabled with the `experimental.emitRouteChunkError` option. This will catch chunk loading errors and calls the `app:chunkError` page. This can be handled with something like this:

```js
  nuxtApp.hook('app:chunkError', ({ error }) => {
    console.log('caught chunk load error'))
  })
})
```

You can also set `experimental.emitRouteChunkError` to `reload` to enable an automatic hard reload when navigating if the new route triggers a chunk loading error.

**Notes**:
 - In webpack, a chunk error is not thrown when fetching payloads, although it does with vite. I think this is fine - errors fetching payload aren't really the kind of error we're trying to catch as payloads aren't hashed.
 - this approach does not _suppress_ chunk errors. perhaps we can find a way for user to indicate that an error is 'handled' so we can then avoid throwing it.
 - the `reload` strategy does not handle all chunk errors, just ones on route navigation. For example, issues with _prefetching_ chunks won't reload the page, though the error will still be thrown and it's possible to handle this intentionally

### 📝 Checklist


packages/nuxt/src/app/plugins/chunk-reload.client.ts
```diff
@@ -0,0 +1,23 @@
+import { defineNuxtPlugin } from '#app/nuxt'
+import { useRouter } from '#app/composables/router'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  const router = useRouter()
+
+  let chunkError: TypeError | null = null
+  nuxtApp.hook('app:chunkError', ({ error }) => { chunkError = error })
```
danielroe, line 8: We could also catch this directly here:

```suggestion
  nuxtApp.hook('app:chunkError', ({ error }) => {
    if (nuxtApp._processingMiddleware) {
      // handle error here
    }
  })
```

I've not done that here in case the user has handled the error, but it's worth keeping in mind if someone has a better idea.

packages/vite/src/client.ts
```diff
@@ -68,7 +69,10 @@ export async function buildClient (ctx: ViteBuildContext) {
       runtimePathsPlugin({
         sourcemap: ctx.nuxt.options.sourcemap.client
       }),
-      viteNodePlugin(ctx)
+      viteNodePlugin(ctx),
+      chunkErrorPlugin({
```
clarkdo, line 73: Should we also add the plugin conditionally according to `reloadOnRouteChunkError` option ?

clarkdo (approved):
Nice!

packages/vite/src/client.ts
danielroe: Good thought! I was thinking with it false they might handle it themselves. Maybe support instead via configuring it as either `true | false | 'manual'` (the last one raises events but doesn't handle)?

clarkdo: > the last one raises events but doesn't handle

Sounds great!

danielroe: ~> refactored to `emitRouteChunkError` with `true | false | 'reload'` signature.

danielroe: This also means we can potentially merge in a patch release for testing. 👀

packages/nuxt/src/app/plugins/chunk-reload.client.ts
pi0, line 8: It might be nice if we can keep all chunk errors (otherwise only latest is available, loosing errors)

packages/nuxt/src/app/plugins/chunk-reload.client.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  router.onError((error, to) => {
+    if (chunkError !== error) { return }
+
+    let handledPath: Record<string, any> = {}
+    try {
+      handledPath = JSON.parse(localStorage.getItem('n:chunk:reload') ?? '{}')
```
pi0, line 15: I would suggest to rename it to `nuxt` for readability since is not transformed over network key size should be okay.. (also empty string cannot be parsed)

```suggestion
      handledPath = JSON.parse(localStorage.getItem('nuxt:reload') || '{}')
```

packages/nuxt/src/app/plugins/chunk-reload.client.ts
```diff
@@ -0,0 +1,23 @@
...
+    } catch {}
+
+    if (handledPath?.path !== to.fullPath || handledPath?.expires < Date.now()) {
+      localStorage.setItem('n:chunk:reload', JSON.stringify({ path: to.fullPath, expires: Date.now() + 10000 }))
```
pi0, line 19:
```suggestion
      localStorage.setItem('nuxt:reload', JSON.stringify({ path: to.fullPath, expires: Date.now() + 10000 }))
```

packages/nuxt/src/app/plugins/chunk-reload.client.ts
```diff
@@ -0,0 +1,23 @@
...
+      window.location.href = to.fullPath
```
pi0, line 20: Can you please double-check if it is double URLs are not remaining in the history API?

pi0:
Nice work! For me it even looks good to be out of experimental flag testing on edge only.

One note about implementation it would be nice if the "reload" functionality extracted as a composable utility `reloadNuxtApp({ error? })`. It can be useful for outside integrations too such as handling app updates.

danielroe:
Great! 😊 I think we can include in next minor, along with extracting reload helper. 👍 For now I'll keep it under flag so we can merge in a patch release.

packages/nuxt/src/app/plugins/chunk-reload.client.ts
pi0: Check MDN docs: https://developer.mozilla.org/en-US/docs/Web/API/Location/href

Assigning makes a navigation (vs replacing as reload)

danielroe: But it's _meant_ to be a navigation.

pi0 (approved):

scofield-ua:
@danielroe so this is how I can set it in `nuxt.config.ts`:

```
   experimental: {
      emitRouteChunkError: 'reload'    
   }
})
```

Is that correct?

danielroe:
@scofield-ua The API changed a little bit in https://github.com/nuxt/nuxt/pull/19086.

The jsdocs should give the possible configuration options for this. It is enabled by default in v3.3+.

PR #19053, opened 2023-02-15
fix(vite): respect `isCustomElement` config for jsx transform

### LINKED ISSUE #19024: add support for configuring @vitejs/plugin-vue-jsx's isCustomElement option
### Describe the feature

there is currently no way to configure `@vitejs/plugin-vue-jsx` via nuxt's config. see this line from `@nuxt/vite-builder` https://github.com/nuxt/nuxt/blob/523b495e1d808c3a1e57e190d73f460ddf3b3d8b/packages/vite/src/client.ts#LL63C23-L63C23

my use case is that i need to pass an `isCustomElement` function to the vue jsx plugin. the jsx plugin needs its own `isCustomElement` function (as well as the usual one you'd pass to the more common vue plugin). without this, i get "Failed to resolve component" warnings logged in the console.

there are two possible approaches to solving this that i see

1. internally pass the same `isCustomElement`  function to both vue and jsx plugins. i can't imagine why you'd ever need them to differ, and this means things would "just work" if you configure it once. however i understand this is a little magical, and may not be paletable to you
2. a simpler option is to add a new config section specifically for the jsx plugin. this could be top-level alongside `vue` like `vueJsx`, or a child key of vue like `vue.jsx`, or whatever makes sense

i am more than happy to work on a PR immediately for this following agreement from maintainers.

thanks!

### Additional information

### Final checks

WickyNilliams:
### 🔗 Linked issue

resolves #19024 

### ❓ Type of change

### 📚 Description

Adds ability to configure `@vitejs/plugin-vue-jsx` via the nuxt config. This config will live at config path `vite.vueJsx`. 

Without this, it is impossible to silence console warnings pertaining to custom elements usage in JSX. However opening up this config is more generally useful for configuring other aspects of `@vitejs/plugin-vue-jsx`.

To simplify the common use case, `vue.comilerOptions.isCustomElement` is used as a smart default for `vite.vueJsx.isCustomElement`.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -44,8 +44,11 @@ export default defineUntypedSchema({
       }
     },
     vueJsx: {
-      isCustomElement: {
-        $resolve: async (val, get) => val ?? (await get('vue')).compilerOptions?.isCustomElement
+      $resolve: async (val, get) => {
+        return {
+          isCustomElement: (await get('vue')).compilerOptions?.isCustomElement,
+          ...(val || {})
+        }
       }
     },
     optimizeDeps: {
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -43,6 +43,11 @@ export default defineUntypedSchema({
         }
       }
     },
+    vueJsx: {
+      isCustomElement: {
+        $resolve: async (val, get) => val ?? (await get('vue')).compilerOptions?.isCustomElement
+      }
+    },
     optimizeDeps: {
       exclude: {
         $resolve: async (val, get) => [
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -2,6 +2,7 @@ import type { KeepAliveProps, TransitionProps } from 'vue'
 import type { ConfigSchema } from '../../schema/config'
 import type { ServerOptions as ViteServerOptions, UserConfig as ViteUserConfig } from 'vite'
 import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
+import type { Options as VueJsxPluginOptions } from '@vitejs/plugin-vue-jsx'
 import type { AppHeadMetaObject } from './meta'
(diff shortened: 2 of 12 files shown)


WickyNilliams:
i struggled to check whether the docs were ok, since the [instructions for running the docs](https://nuxt.com/docs/community/framework-contribution#set-up-documentation-website-in-local-environment) seem to be incorrect/out of date... they recommend running `yarn` commands, but the repo is configured for `pnpm`. also the `dev` command it asks you to run starts the playground rather than the docs. 

so i'm not sure how to test. but it seems config docs are auto-generated, so perpaps i don't actually need to do anything here anyway? please let me know what's required here :) if nothing, i will cross off the item in the checklist

packages/schema/src/types/config.ts
```diff
@@ -95,10 +96,16 @@ export interface NuxtOptions extends Omit<ConfigSchema, 'builder'> {
 export interface ViteConfig extends ViteUserConfig {
   /**
    * Options passed to @vitejs/plugin-vue
-   * @see https://github.com/vitejs/vite/tree/main/packages/plugin-vue
+   * @see https://github.com/vitejs/vite-plugin-vue/tree/main/packages/plugin-vue
```
WickyNilliams, line 99: the url here was out of date, so i updated

packages/schema/src/config/vite.ts
```diff
@@ -43,6 +43,11 @@ export default defineUntypedSchema({
         }
       }
     },
+    vueJsx: {
+      isCustomElement: {
+        $resolve: async (val, get) => val ?? (await get('vue')).compilerOptions?.isCustomElement
```
WickyNilliams, line 48: i wasn't sure if needed the optional chaining here?

WickyNilliams: that makes a lot of sense, thanks

danielroe (approved):

WickyNilliams:
thank you so much @danielroe! really appreciate all the extra work you put in to get this over the line 🚀

PR #19055, opened 2023-02-15
fix(nuxi): read `devServer` options from nuxt config

### LINKED ISSUE #15710: devServer config do not work in nuxt.config
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@8.19.2`
- Builder:          `vite`
- User Config:      `debug`, `ssr`, `srcDir`, `serverDir`, `dir`, `alias`, `app`, `css`, `devServer`, `runtimeConfig`, `router`, `modules`, `plugins`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

In nuxt config add :
`devServer` attribute like in the doc
https://nuxt.com/docs/api/configuration/nuxt-config/#devserver
Try to customise `port` and `host` with custom `port` and/or custom `ip`.
Run dev command.
It will show Local: `http://localhost:3000/ `in Terminal no  matter which custom port or ip you set.

https://stackblitz.com/edit/github-upsjtw?file=nuxt.config.ts

### Describe the bug

The configuration of devserver is in the docs, but do not seems to work :
 https://nuxt.com/docs/api/configuration/nuxt-config/#devserver
 All related issues say to use `NITRO_PORT` or `PORT`; `NITRO_HOST` or`HOST` if we want to customize those values.
 I feel confused to not be able to use nuxt.config. Setting the host port is not part of the configuration 

### Additional context

_No response_

### Logs

_No response_

Verequies:
### 🔗 Linked issue

Related issue #15710 

### ❓ Type of change

### 📚 Description

nuxi has been updated to read `devServer` options from the Nuxt config.
The default for `devServer.host` and `devServer.url` has been changed in schema to avoid issues.

The changes are required as nuxi currently does not read any `devServer` options from the nuxt config.
This is a problem as it means we cannot dynamically change, for example, the HTTPS certificate and private key.

To support the change, schema has been updated to set the default for `devServer.host` to `0.0.0.0`, and `devServer.url` to `http://0.0.0.0:3000`. If this wasn't changed, the default Nuxt host setting would override listhens default settings when the user has not specified any host option, and potentially break existing developers setups.

Current documentation is misleading, stating that the `devServer` configuration options work, when in reality they are never read at all. This would save a lot of developers from wasting time figuring out why the options don't work, as I did.

Resolves #15710 

### 📝 Checklist

Unsure about updating the documentation, I believe it gets automatically updated as I could not find the relating file.

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -46,7 +46,10 @@ export default defineNuxtCommand({
 
     const { loadNuxt, loadNuxtConfig, buildNuxt } = await loadKit(rootDir)
 
-    const config = await loadNuxtConfig({})
+    const config = await loadNuxtConfig({
+      cwd: rootDir,
+      overrides: { dev: true }
+    })
 
     const listener = await listen(serverHandler, {
       showURL: false,
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -44,19 +44,23 @@ export default defineNuxtCommand({
 
     await setupDotenv({ cwd: rootDir, fileName: args.dotenv })
 
+    const { loadNuxt, loadNuxtConfig, buildNuxt } = await loadKit(rootDir)
+
+    const config = await loadNuxtConfig({})
+
     const listener = await listen(serverHandler, {
       showURL: false,
       clipboard: args.clipboard,
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT,
-      hostname: args.host || args.h || process.env.NUXT_HOST,
-      https: args.https && {
-        cert: args['ssl-cert'],
-        key: args['ssl-key']
-      }
+      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+      hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
+      https: (args.https !== false && (args.https || config.devServer.https))
+        ? {
+            cert: args['ssl-cert'] || (config.devServer.https && config.devServer.https.cert) || undefined,
+            key: args['ssl-key'] || (config.devServer.https && config.devServer.https.key) || undefined
+          }
+        : false
     })
-
-    const { loadNuxt, buildNuxt } = await loadKit(rootDir)
 
     let currentNuxt: Nuxt
     const showURL = () => {
--- a/packages/schema/src/config/dev.ts
+++ b/packages/schema/src/config/dev.ts
@@ -28,10 +28,13 @@ export default defineUntypedSchema({
     port: process.env.NUXT_PORT || process.env.NITRO_PORT || process.env.PORT || 3000,
 
     /** Dev server listening host */
-    host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || 'localhost',
+    host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || '',
 
     /**
-     * Listening dev server url
+     * Listening dev server URL.
+     *
+     * This should not be set directly as it will always be overridden by the
+     * dev server with the full URL (for module and internal use).
      */
     url: 'http://localhost:3000',
   }


packages/nuxi/src/commands/dev.ts
```diff
@@ -44,20 +44,22 @@ export default defineNuxtCommand({
...
       clipboard: args.clipboard,
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT,
-      hostname: args.host || args.h || process.env.NUXT_HOST,
-      https: args.https && {
+      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+      hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
+      https: (args.https && {
         cert: args['ssl-cert'],
         key: args['ssl-key']
```
danielroe, line 59:
```suggestion
        cert: args['ssl-cert'] || config.devServer.cert,
        key: args['ssl-key'] || config.devServer.key
```

packages/nuxi/src/commands/dev.ts
```diff
@@ -44,20 +44,22 @@ export default defineNuxtCommand({
 
     await setupDotenv({ cwd: rootDir, fileName: args.dotenv })
 
+    const { loadNuxt, loadNuxtConfig, buildNuxt } = await loadKit(rootDir)
+
+    const config = await loadNuxtConfig({})
```
danielroe, line 49:
```suggestion
    const config = await loadNuxtConfig({ rootDir, dev: true, ready: false })
```

packages/schema/src/config/dev.ts
```diff
@@ -28,11 +28,11 @@ export default defineUntypedSchema({
     port: process.env.NUXT_PORT || process.env.NITRO_PORT || process.env.PORT || 3000,
 
     /** Dev server listening host */
-    host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || 'localhost',
+    host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || '0.0.0.0',
```
danielroe, line 31:
```suggestion
    host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || '',
```

packages/schema/src/config/dev.ts
```diff
@@ -28,11 +28,11 @@ export default defineUntypedSchema({
...
 
     /**
      * Listening dev server url
```
danielroe, line 34:
```suggestion
     * Listening dev server URL.
     *
     * This should not be set directly as it will always be overridden by the
     * dev server with the full URL (for module and internal use).
```

packages/schema/src/config/dev.ts
```diff
@@ -28,11 +28,11 @@ export default defineUntypedSchema({
...
      */
-    url: 'http://localhost:3000',
+    url: 'http://0.0.0.0:3000',
```
danielroe, line 36:
```suggestion
    url: 'http://localhost:3000',
```

packages/nuxi/src/commands/dev.ts
Verequies: I did do this originally, however remove it as the following error occurs:

```
Argument of type '{ rootDir: string; dev: boolean; ready: boolean; }' is not assignable to parameter of type 'LoadConfigOptions<NuxtConfig>'.
  Object literal may only specify known properties, and 'rootDir' does not exist in type 'LoadConfigOptions<NuxtConfig>'.ts(2345)
```

`dev` and `ready` also do not exist.

packages/nuxi/src/commands/dev.ts
Verequies: It seems `config.devServer.https.cert` and `config.devServer.https.key` are not directly accessible without casting.

Is the following acceptable?
```TypeScript
    const listener = await listen(serverHandler, {
      showURL: false,
      clipboard: args.clipboard,
      open: args.open || args.o,
      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
      hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
      https: args.https && ({
        cert: args['ssl-cert'],
        key: args['ssl-key']
      } || config.devServer.https)
    })
```

That should fix the issue that HTTPS is always enabled if the configuration exists in the Nuxt config.
With that said, should we add an option to enable HTTPS from the Nuxt config as well?
I assumed with `devServer.https` being `false | { key: string, cert: string` it would default to true if a HTTPS configuration exists.

packages/schema/src/config/dev.ts
pi0: Good idea defaulting to listhen

packages/nuxi/src/commands/dev.ts
pi0: We can disable https if `config.devServer.https === false` in check

danielroe: I think if user passes `--https` that should override nuxt.config. So logic for https piece should likely be something like:

```js
      https: (args.https || config.devServer.https)
        ? {
            cert: args['ssl-cert'] || (config.devServer.https && config.devServer.https.cert) || undefined,
            key: args['ssl-key'] || (config.devServer.https && config.devServer.https.key) || undefined
          }
        : false
    })
```

Verequies: Currently it does work like that as I made sure to test all terminal arguments override the nuxt config. I put the `--https`, `--ssl-cert` and `--ssl-key` first so that terminal arguments had priority, as I did with the host and port variables. It uses `config.devServer.https` if terminal arguments are not specified. This would just be a more verbose way of implementing it, which in my opinion isn't required to meet the same functionality.

danielroe: As currently written, it wouldn't work if the user passes `--https` but has the key set in the config.

We might make it simpler with `defu` which can merge objects for us.

Verequies: I have just tested this again and when specifying just `--https`, listhen uses its own certificate rather than the options in the nuxt config. If the user doesn't specify any terminal arguments it will use the nuxt config options. 

If the developer would like to use the options in the nuxt config, I believe they should just remove the `--https` arguments as specifying HTTPS options in nuxt config enables HTTPS anyway.

danielroe: I understand that that makes sense. But I think for a minimal extra amount of code here we handle some edge cases that might otherwise confuse people. Otherwise we'll need to document that if you pass `--https` it will override your nuxt config, and if you only pass a key or cert it will be ignored, even if you have ssl entabled in your nuxt config.

packages/nuxi/src/commands/dev.ts
danielroe: quite right - pushed the correct syntax for `loadNuxtConfig`

packages/nuxi/src/commands/dev.ts
Verequies: No worries, I'll fix this up now.

danielroe: thanks - I completely understand ❤️

Verequies:
Everything should now be fixed and good to merge.

packages/nuxi/src/commands/dev.ts
```diff
@@ -44,20 +44,27 @@ export default defineNuxtCommand({
...
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT,
-      hostname: args.host || args.h || process.env.NUXT_HOST,
-      https: args.https && {
-        cert: args['ssl-cert'],
-        key: args['ssl-key']
-      }
+      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+      hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
+      https: (args.https || config.devServer.https)
```
pi0, line 60: Here we should check `args.https !== false` as well. Disabling https if explicit `--no-https` flag was passed to CLI otherwise it is impossible to disable https without updating config.

pi0 (commented):
nice work 💯 (only need supporting `--no-https`)

Verequies:
Okay cool, added checking if `--no-https`  was specified to disable https entirely.

danielroe (approved):

danielroe:
Thank you so much for this, particularly as we've gone back and forth - it's a much-needed fix. ❤️

Verequies:
No worries, glad I could help out. I'm sure a lot of people will be happy with this fix 🙂

PR #19064, opened 2023-02-15
feat(nuxt): add `versions` to runtime nuxtApp

danielroe:
### ❓ Type of change

### 📚 Description

This adds a runtime versions object, similar to the one available on `nuxtApp.vueApp.version` (which corresponds to the Vue version). This can be used for debugging and feature detection. It can be accessed with `useNuxtApp().versions` within user code.

### 📝 Checklist


TheAlexLichter (approved):

manniL:
Should this be documented somewhere?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -119,6 +120,7 @@ export function createNuxtApp (options: CreateOptions) {
   const nuxtApp: NuxtApp = {
     provide: undefined,
     globalName: 'nuxt',
+    version: __NUXT_VERSION__,
```
pi0, line 123: Maybe we could expose it as `versions: { nuxt, vue }`

packages/nuxt/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare global {
+  const __NUXT_VERSION__: string
```
pi0, line 2: We probably can introduce `window.$nuxt` back only with limited information `{ versions }` to reduce side-effects of nuxt on environment

danielroe: this file is only for typing development in this monorepo - it's not shipped to end user

packages/nuxt/src/app/nuxt.ts
danielroe: Sounds nice - and maybe also allow other modules/libraries to expose their versions too

packages/webpack/src/presets/base.ts
```diff
@@ -230,6 +230,7 @@ function getEnv (ctx: WebpackConfigContext) {
     'process.env.NODE_ENV': JSON.stringify(ctx.config.mode),
     'process.mode': JSON.stringify(ctx.config.mode),
     'process.dev': options.dev,
+    __NUXT_VERSION__: JSON.stringify(ctx.nuxt._version),
```
pi0, line 233: Maybe we could use `process.versions.nuxt`? (i know it is not best but since we are already practicing `process.` static shims, and also Node.js uses `process.versions` seems simpler also future proof to also shim runtime if needed)

image

danielroe: What would be the benefit of this?

pi0: `import.meta` seems a good idea to be replaced (however it also inherits the same concerns when we do!). Syntax of  `__NUXT_VERSION__` injection is almost the same as `globalThis. __NUXT_VERSION__` although it is internal even i was confused first time checking PR code.

danielroe: I understand it can be confusing. If you think of a better wording, I'm interested. But this is an internal setting - and, for example, here we already use non-existing env variables and a similar vue devtools flag:

https://github.com/nuxt/nuxt/blob/32f6cb27b7306ec3ff3f4ce7fbd1d60eaac47efd/packages/nuxt/src/core/nitro.ts#L141-L150

Listening to your comments, I think your concern isn't about technical implementation but just about the format of the string that gets replaced, which is an internal factor we aren't exposing publicly. I can see points in favour of both ` __NUXT_VERSION__` and `process.versions.nuxt` but would prefer not to piggyback on `process.versions` as we do nothing with `process.versions` so far and it further ties us to Node APIs.

atinux (approved):

PR #19085, opened 2023-02-16
fix(nuxt): validate no `//` in path when constructing payload url

danielroe:
### ❓ Type of change

### 📚 Description

At the moment we fetch payload URLs using a parsed version of the URL, but we don't need to do this as we've already validated the path passed, and this avoids any issues with improperly constructed URLs like `//localhost//some`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/payload.ts
+++ b/packages/nuxt/src/app/composables/payload.ts
@@ -36,7 +36,12 @@ export function preloadPayload (url: string, opts: LoadPayloadOptions = {}) {
 
 // --- Internal ---
 
+const LOCAL_PATH_RE = /^\/[^/]/
+
 function _getPayloadURL (url: string, opts: LoadPayloadOptions = {}) {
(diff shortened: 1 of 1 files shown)


pi0:
url can be from various sources and less validated. (We introduced URL parsing to validate it). Probably pathname is safer as is but maybe we can add a check that `url` should not start with protocol or `//` before constructing URL? (using `ufo.hasProtocol(url, { allowRelative })`). This disallows the `//` bypass (which currently is possible and bypassing checks before)

danielroe:
Yes, it can be invalid in many ways even after our checks, but the main point is that we have already validated _host_, so the worst situation is that we trigger a 404.

pi0 (approved):

pi0:
I liked it before https://github.com/nuxt/nuxt/pull/19085/commits/bda8b030bdbca2f785856016d1f132e09db8a32b 😆

PR #19086, opened 2023-02-16
feat(nuxt): enable chunk error handling by default

### LINKED ISSUE #19182: emitRouteChunkError: 'reload' redirects incorrectly with hashMode: true
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.5.0`
- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.2`
- Builder:          `vite`
- User Config:      `ssr`, `router`, `experimental`, `components`, `modules`, `css`, `runtimeConfig`, `vite`, `imports`, `vue`
- Runtime Modules:  `@pinia/nuxt@0.4.6`, `bootstrap-vue-next/nuxt`
- Build Modules:    `-`

### Reproduction

In description.

### Describe the bug

set `router: {
        options: {
            hashMode: true
        },
    }`, make a build with `nuxt generate`, deploy .output/public to any static hosting, open some page, make changes to another page, rebuild and re-deploy, and then try to navigate to another page with changes from already opened page from previous build, you'll have redirect without `#` which in my case lead to 404 error.
Path saved in local storage `nuxt:reload` does not contains `#` and `NUXT_APP_BASE_URL` prefix.
Instead of redirect to `https://example.com/app/#/directory/page` I got redirected to `https://example.com/directory/page` 

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/19038#issuecomment-1431462008
https://github.com/nuxt/nuxt/issues/14594
resolves https://github.com/nuxt/nuxt/issues/19182

### ❓ Type of change

### 📚 Description

This is the next step to enabling chunk error handling by default in Nuxt apps.

#### Progress

- [x] add `reloadNuxtApp` composable.
- [x] add documentation for composable and chunk error handling
- [x] remove `experimental` flag for emitting errors (and enable reload strategy by default)
- [x] resolve https://github.com/nuxt/nuxt/issues/19182
- [x] allow preserving nuxtApp state during reload (requires opt-in)
- [x] Add more documentation for the `reload` strategy in **Error Handling** docs section

#### Future enhancements

- [ ] possibly handle chunk errors outside route navigation?
- [ ] allow users to _suppress_ chunk errors

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/chunk.ts
+++ b/packages/nuxt/src/app/composables/chunk.ts
@@ -1,17 +1,35 @@
 export interface ReloadNuxtAppOptions {
+  /**
+   * Number of milliseconds in which to ignore future reload requests
+   *
+   * @default {10000}
+   */
+  ttl?: number
+  /**
+   * Force a reload even if one has occurred within the previously specified TTL.
+   *
+   * @default {false}
+   */
+  force?: boolean
+  /**
+   * The path to reload. If this is different from the current window location it will
+   * trigger a navigation and add an entry in the browser history.
+   *
+   * @default {window.location.pathname}
+   */
   path?: string
 }
 
-export function reloadNuxtApp (options?: ReloadNuxtAppOptions) {
-  const path = options?.path || window.location.pathname
+export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
+  const path = options.path || window.location.pathname
 
   let handledPath: Record<string, any> = {}
   try {
     handledPath = JSON.parse(localStorage.getItem('nuxt:reload') || '{}')
   } catch {}
 
-  if (handledPath?.path !== path || handledPath?.expires < Date.now()) {
-    localStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + 10000 }))
+  if (options.force || handledPath?.path !== path || handledPath?.expires < Date.now()) {
+    localStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
     if (window.location.pathname !== path) {
       window.location.href = path
     } else {
--- a/packages/nuxt/src/app/composables/chunk.ts
+++ b/packages/nuxt/src/app/composables/chunk.ts
@@ -1,0 +1,21 @@
+export interface ReloadNuxtAppOptions {
+  path?: string
+}
+
+export function reloadNuxtApp (options?: ReloadNuxtAppOptions) {
+  const path = options?.path || window.location.pathname
+
+  let handledPath: Record<string, any> = {}
+  try {
+    handledPath = JSON.parse(localStorage.getItem('nuxt:reload') || '{}')
+  } catch {}
+
+  if (handledPath?.path !== path || handledPath?.expires < Date.now()) {
+    localStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + 10000 }))
+    if (window.location.pathname !== path) {
+      window.location.href = path
+    } else {
+      window.location.reload()
+    }
+  }
+}
--- a/packages/nuxt/src/app/composables/chunk.ts
+++ b/packages/nuxt/src/app/composables/chunk.ts
@@ -37,7 +37,7 @@ export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
     } catch {}
 
     try {
-      // TODO: handle serializing/deserializing complex states: https://github.com/nuxt/nuxt/issues/12831
+      // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
       sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
     } catch {}
 
--- a/packages/nuxt/src/app/composables/chunk.ts
+++ b/packages/nuxt/src/app/composables/chunk.ts
@@ -14,6 +14,12 @@ export interface ReloadNuxtAppOptions {
    */
   force?: boolean
   /**
+   * Whether to dump the current Nuxt state to sessionStorage (as `nuxt:reload:state`).
+   *
+   * @default {false}
+   */
+  persistState?: boolean
+  /**
    * The path to reload. If this is different from the current window location it will
    * trigger a navigation and add an entry in the browser history.
    *
@@ -36,10 +42,12 @@ export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
       sessionStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
     } catch {}
 
-    try {
-      // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
-      sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
-    } catch {}
+    if (options.persistState) {
+      try {
+        // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
+        sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
+      } catch {}
+    }
 
     if (window.location.pathname !== path) {
       window.location.href = path
--- a/packages/nuxt/src/app/composables/chunk.ts
+++ b/packages/nuxt/src/app/composables/chunk.ts
@@ -25,11 +25,14 @@ export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
 
   let handledPath: Record<string, any> = {}
   try {
-    handledPath = JSON.parse(localStorage.getItem('nuxt:reload') || '{}')
+    handledPath = JSON.parse(sessionStorage.getItem('nuxt:reload') || '{}')
   } catch {}
 
   if (options.force || handledPath?.path !== path || handledPath?.expires < Date.now()) {
-    localStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
+    try {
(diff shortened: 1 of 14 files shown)


mcrapts:
I tried it out, and I like it! 

Specifically I'm trying to handle the situation where the network connection is interrupted. This especially can happen often in corporate environments with VPNs. Without any handling, interrupted connections with Nuxt 3 result in a fatal error that is non-recoverable, i.e. any subsequent clicks when the connection is restored won't do anything. I simulate this behavior by setting the throttling to "No connection" in the network tab in Chrome developer tools.

1. The behavior is different in dev vs production. In the above scenario the chunkError hook doesn't seem to be called in dev.
2. The hook will trigger in the production build, but now I encounter the problem that the hook is already called during prefetching. Ideally it would be possible to discard prefetch situations, or at least add the trigger to the error context, so that errors can be handled differently for prefetching vs regular navigation.

danielroe:
@mcrapts Would you explain more about this:

> The hook will trigger in the production build, but now I encounter the problem that the hook is already called during prefetching. Ideally it would be possible to discard prefetch situations, or at least add the trigger to the error context, so that errors can be handled differently for prefetching vs regular navigation.

The `reload` strategy currently does not handle prefetch errors, so I take it you mean that it would be helpful to have more information in the `app:chunkError` hook about the situation that caused the chunk error?

mcrapts:
> @mcrapts Would you explain more about this:
> 
> 
> 
> > The hook will trigger in the production build, but now I encounter the problem that the hook is already called during prefetching. Ideally it would be possible to discard prefetch situations, or at least add the trigger to the error context, so that errors can be handled differently for prefetching vs regular navigation.
> 
> 
> 
> The `reload` strategy currently does not handle prefetch errors, so I take it you mean that it would be helpful to have more information in the `app:chunkError` hook about the situation that caused the chunk error?

Yes, exactly. Then I can differentiate between prefetch and navigation errors. Navigation errors are more critical and I want to handle them differently.

danielroe:
Interesting idea. Currently we get that information this way in the `reload` strategy:

```js
  const router = useRouter()

  const chunkErrors = new Set()

  router.beforeEach(() => { chunkErrors.clear() })
  nuxtApp.hook('app:chunkError', ({ error }) => { chunkErrors.add(error) })

  router.onError((error, to) => {
    if (chunkErrors.has(error)) {
      // it was a navigation error
    }
  })
})
```

We can't really annotate the error (which is emitted from vite's preload function) without adding runtime code to do so, so that might be a future issue/PR.

packages/nuxt/src/app/composables/chunk.ts
```diff
@@ -0,0 +1,48 @@
...
+   * The path to reload. If this is different from the current window location it will
+   * trigger a navigation and add an entry in the browser history.
+   *
+   * @default {window.location.pathname}
+   */
+  path?: string
+}
+
+export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
+  const path = options.path || window.location.pathname
```
pi0, line 26: I guess we can add this to optimizations to be treeshaked from server bundle or directly protect with `if (process.server) { return }`

packages/nuxt/src/app/composables/chunk.ts
```diff
@@ -0,0 +1,48 @@
...
+    handledPath = JSON.parse(sessionStorage.getItem('nuxt:reload') || '{}')
+  } catch {}
+
+  if (options.force || handledPath?.path !== path || handledPath?.expires < Date.now()) {
+    try {
+      sessionStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
+    } catch {}
+
+    try {
+      sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
```
pi0, line 39: Would be nice to add a todo, using state serializer when possible because this way we also drop some non-pojo types even if JSON.stringify succeded

packages/nuxt/src/app/composables/chunk.ts
danielroe:
```suggestion
  if (process.server) { return }
  const path = options.path || window.location.pathname
```

packages/nuxt/src/app/composables/chunk.ts
danielroe:
```suggestion
      // TODO: handle serializing/deserializing complex states: https://github.com/nuxt/nuxt/issues/12831
      sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
```

packages/nuxt/src/app/composables/chunk.ts
```diff
@@ -0,0 +1,50 @@
...
+    handledPath = JSON.parse(sessionStorage.getItem('nuxt:reload') || '{}')
+  } catch {}
+
+  if (options.force || handledPath?.path !== path || handledPath?.expires < Date.now()) {
+    try {
+      sessionStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
+    } catch {}
+
+    try {
+      // TODO: handle serializing/deserializing complex states: https://github.com/nuxt/nuxt/issues/12831
```
pi0, line 40: I think this is updated one :) 

```suggestion
      // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
```

packages/nuxt/src/app/composables/chunk.ts
```diff
@@ -0,0 +1,50 @@
...
+  } catch {}
+
+  if (options.force || handledPath?.path !== path || handledPath?.expires < Date.now()) {
+    try {
+      sessionStorage.setItem('nuxt:reload', JSON.stringify({ path, expires: Date.now() + (options.ttl ?? 10000) }))
+    } catch {}
+
+    try {
+      // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
+      sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
```
pi0, line 41: Small suggestion: We could support persistState: false as an escape hatch for when it is unwanted or problematic (but i guess we cannot customize it for chuck auto reload behavior…)

pi0 (approved):

packages/nuxt/src/app/composables/chunk.ts
danielroe: Nice idea 👌

packages/nuxt/src/app/composables/chunk.ts
```diff
@@ -36,10 +42,12 @@ export function reloadNuxtApp (options: ReloadNuxtAppOptions = {}) {
...
-    try {
-      // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
-      sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
-    } catch {}
+    if (options.persistState) {
+      try {
+        // TODO: handle serializing/deserializing complex states as JSON: https://github.com/nuxt/nuxt/pull/19205
+        sessionStorage.setItem('nuxt:reload:state', JSON.stringify({ state: useNuxtApp().payload.state }))
+      } catch {}
+    }
```
pi0, line 50: (I think it might be safe to remove `nuxt:reload:state` otherwise to ensure there is no stalled data)

pi0 (approved):

packages/nuxt/src/app/composables/chunk.ts
danielroe: If `experimental.restoreState` is set, then we will do that on reload. Otherwise the user is handling it manually and will need to clean it up themselves.

toniengelhardt:
I'm not 100% sure if it is related to this, but since I turned it on I get these errors with `kurkle` in `vue-chartjs`. Cannot reproduce, but I see a lot of errors in Sentry:

https://github.com/chartjs/Chart.js/issues/11029#issuecomment-1465203796

Any clues why those would be triggered with

```ts
experimental: {
  emitRouteChunkError: 'reload',
}
```

Maybe a teardown issue somehow?

danielroe:
Nuxt isn't injecting that `querySelectorAll` into the code (so I think that's from another dependency - it doesn't seem to be in chart.js either).

The reload is a hard reload (`window.location.href = path` or `window.location.reload()`). I don't know if that helps at all.

toniengelhardt:
I think it is in the `kurkle` dependency of `chartjs`. It is probably an issue with that package then, that it doesn't handle the reload properly... Thanks for checking 🙏🏽

sondh0127:
Would it be possible to extract it as a vite plugin for use in vite base SPA applications

danielroe:
Indeed. I've already opened a PR as the first step: https://github.com/vitejs/vite/pull/12084

edwh:
@danielroe Is there a rationale for only handling errors on the route?

If you're using async components then you'll hit missing chunks when deploying on something like Netlify, where the old chunks are gone.  

Would the solution to that be to always reload when you hit a chunk error, along the lines of:

`import { joinURL } from 'ufo'
import { defineNuxtPlugin, useRuntimeConfig } from '#app/nuxt'
import { reloadNuxtApp } from '#app/composables/chunk'

  name: 'nuxt:chunk-reload',
  setup (nuxtApp) {

    nuxtApp.hook('app:chunkError', ({ error }) => { 
        const isHash = 'href' in to && (to.href as string).startsWith('#')
        const path = isHash ? config.app.baseURL + (to as any).href : joinURL(config.app.baseURL, to.fullPath)
        reloadNuxtApp({ path, persistState: true })
    })
  }
})`

I've not tested this - I'm wondering if I'm missing something Very Bad that might happen.

danielroe:
No, I think that should be a reasonable approach :+1:

danielroe:
Try getting the route outside the hook - i.e. in the body of the plugin?

edwh:
@danielroe Thanks, I think that now works.  But I think we might be able to do better, at least on Netlify. I've opened a separate issue https://github.com/nuxt/nuxt/issues/20950.

Issue #10844, opened 2022-12-03
Nuxt2 default setup doesn't work with latest node LTS version 18.12.1


michael-pfister:
The default Nuxt2 project setup runs into an error `Error: error:0308010C:digital envelope routines::unsupported` when starting the development environment (`yarn dev`) using node v18.12.1.

> **Reason For The Error**
In Node.js v17, the Node.js developers closed a security hole in the SSL provider. This fix was a breaking change that corresponded with similar breaking changes in the SSL packages in NPM. [Webpack 4.46.0](https://www.npmjs.com/package/webpack/v/4.46.0) (a module currently used by nuxt2 for serving) is using a deprecated hashing algorithm `md4` for it's SSL encryption, which I assume was patched out in Node.js v17.

# ⚠️ DANGER 
Almost all threads found online regarding this issue are suggesting to either downgrade Node.js to pre v17 or to use the legacy SSL provider. Both of those solutions are hacks that leave your builds open to security threats.

This is a big issue because the [prerequisites](https://nuxtjs.org/docs/get-started/installation#prerequisites) in the current nuxt installation guide are telling newcomers [to get the latest node LTS version](https://nuxtjs.org/docs/get-started/installation#:~:text=node%20%2D%20We%20recommend%20you%20have%20the%20latest%20LTS%20version%20installed), which is currently 18.12.1. Googling the error message (which they will definitely get) will lead them to either set `NODE_OPTIONS=--openssl-legacy-provider`  or downgrade Node.js to pre v17.

### Versions

- nuxt: v2.15.8
- node: v18.12.1

### Reproduction

GitHub repository: https://github.com/michael-pfister/nuxt2-node-demonstrations

<details open>
<summary>Additional Details</summary>
<br>
Make sure you have the currently latest node LTS version (v18.12.1) installed before running any build commands.
</details>

### Steps to reproduce
```bash
# clone git repository
$ git clone https://github.com/michael-pfister/nuxt2-node-demonstrations

# cd into the repository
$ cd nuxt2-node-demonstrations

# install dependencies
$ yarn install

# serve with hot reload at localhost:3000
$ yarn dev
```

### What is Expected?

[...]

PR #19108, opened 2023-02-17
fix(config): upgrade md4 -> md5 on node > 16

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/10844
https://github.com/nuxt/nuxt/discussions/19078

### ❓ Type of change

### 📚 Description

We have a tricky situation upcoming, with Node 16 due to be EOL on September 11th, 2023, but Nuxt 2 is not EOL until the end of the year.

Webpack 4 hard-codes `md4` algorithm in several places in the code base meaning our options to resolve this for Nuxt 2 users are somewhat limited. In build/dev mode, this change monkey patches the `crypto.createHash` function if running on Node > 16 to upgrade `md4` to `md5`. Would welcome other ideas too 🙏

> **Warning**
> This should not be taken for an endorsement of continuing to run with Webpack 4, which is out of date and has a number of dependencies with issues. I expect that number to continue to grow, and we will not be able to resolve all of them. I would strongly urge migrating to Nuxt 3 if possible and the team will do our best to make this possible over the course of the year ❤️ 

### 📝 Checklist


pi0 (approved):
LGTM (i hope not introducing side-effects, but seems won't!)

Issue #18845, opened 2023-02-07
`preloadRouteComponents` allows passing values of type `RouteLocationRaw`


Mini-ghost:
### Describe the feature

Currently, `preloadRouteComponents()` only allows the first parameter to be a string, but sometimes we may pass an object for `to` of `<NuxtLink />`. This can cause many of the same routes path to be considered not preloaded, like bellow:

```html
<!-- Passing a string -->
<NuxtLink to="/home"> Home </NuxtLink>

<!-- Passing a string with query string -->
<!-- `router.resolve('/home?year=2023').path` is `/home` -->
<NuxtLink to="/home?year=2023"> Home </NuxtLink>

<!-- Passing a object -->
<!-- `router.resolve({ name: 'home', query: { year: '2023' } }).path` is `/home` -->
<NuxtLink :to="{ name: 'home', query: { year: '2023' } }"> Home </NuxtLink>
```

These routing paths are all '/home', but all of them will run preload without early return, although the request will not actually be sent.

---

Maybe we can allow passing an object or string of type `RouteLocationRaw` and use the `path` returned by `router.resolve()` as the preloaded record?

The `preloadRouteComponents()` will like bellow:

```ts
function preloadRouteComponents(to: RouteLocationRaw, router?: Router): Promise<void>
```

I don't know if it fits the core team's thinking.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

manniL:
Sounds reasonable to me 👍🏻

PR #19120, opened 2023-02-17
feat(nuxt): enable preloading object-syntax routes

Mini-ghost:
### 🔗 Linked issue

- #18845 

### ❓ Type of change

### 📚 Description

After this PR, the path parsed by the following code all will be `/home`. As a result, the preload function will not retry and will instead return early.

```html
<!-- Passing a string -->
<NuxtLink to="/home"> Home </NuxtLink>

<!-- Passing a string with query string -->
<!-- `router.resolve('/home?year=2023').path` is `/home` -->
<NuxtLink to="/home?year=2023"> Home </NuxtLink>

<!-- Passing a object -->
<!-- `router.resolve({ name: 'home', query: { year: '2023' } }).path` is `/home` -->
<NuxtLink :to="{ name: 'home', query: { year: '2023' } }"> Home </NuxtLink>
```

In addition, when users use `preloadRouteComponents`, they are not limited to type string.

```ts
const route: RouteLocationRaw = {
  name: 'dashboard'
}

preloadRouteComponents(dashboard)
const submit = async () => {
  const results = await $fetch('/api/authentication')
  if (results.token) {
    await navigateTo(dashboard)
  }
}
```
### 📝 Checklist


packages/nuxt/src/app/composables/preload.ts
```diff
@@ -34,11 +34,13 @@ function _loadAsyncComponent (component: Component) {
...
 
-export async function preloadRouteComponents (to: string, router: Router & { _routePreloaded?: Set<string>; _preloadPromises?: Array<Promise<any>> } = useRouter()): Promise<void> {
+export async function preloadRouteComponents (to: RouteLocationRaw, router: Router & { _routePreloaded?: Set<string>; _preloadPromises?: Array<Promise<any>> } = useRouter()): Promise<void> {
   if (process.server) { return }
 
+  const { path, matched } = router.resolve(to)
+
   if (!router._routePreloaded) { router._routePreloaded = new Set() }
-  if (router._routePreloaded.has(to)) { return }
+  if (router._routePreloaded.has(path)) { return }
```
danielroe, line 43: maybe we can also early-return if there are no matched components?

Mini-ghost: Like this?

```ts
if (!matched.length) { return }
if (!router._routePreloaded) { router._routePreloaded = new Set() }
if (router._routePreloaded.has(path)) { return }
```

I think it looks better this way, but I'm not sure when there is no matching component?

danielroe: Looks perfect!

danielroe:
Would you update this following the merge of your other PR? 🙏

Mini-ghost:
I resolved the conflict caused by #19144.

Mini-ghost:
I'm unsure which option is better between using `name` or `path` for dynamic route, Even though they resolve to the same page components, their paths may still differ. While `name` may not have this issue, it is possible for it to be `undefined` or `null`.

Does the `name` of the route always exist in Nuxt? If so, would it be better to use it?

Overall, this issue is not significant because although dynamic routing repeats preloading, it doesn't require sending additional network requests.

danielroe:
`name` does not always exist, so I think better to use `path`.

PR #19131, opened 2023-02-17
fix(nuxi): restart nuxt when `distDir` is unlinked

### LINKED ISSUE #15069: cannot find client.manifest.mjs when adding or removing package while dev server is up
### Environment

- Node Version:     `v16.17.0`
- Nitro Version:    `0.5.4`
- Package Manager:  `yarn@3.2.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

- start nuxt dev 
- on another terminal , install any package using yarn (not tested with npm or pnpm)
- try to go on any page 
- error 500 

### Describe the bug

Hi :wave: ,

adding a package while the dev server is on cause `.nuxt/dev` and `.nuxt/dist` being deleted.

stopping and reloading the server solves the issue since it regenerate `.nuxt` directory.

I reproduced this on window and WSL(ubuntu)

### Additional context

_No response_

### Logs

```shell
Cannot find module '/mnt/d/GIT/test/.nuxt/dist/server/client.manifest.mjs' imported from /mnt/d/GIT/test/.nuxt/dev/index.mjs
```

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/15069

### ❓ Type of change

### 📚 Description
Hi :wave: 
this adds a watcher on `.nuxt/dist`. The dev server would reload nuxt by itself if the `dist` directory has been removed. 
This would avoid to manually restart the server.

Use case: 
- Start dev server 
- "Oh i forgot to add another package"
- `npm install another-package` in another terminal 
- `postinstall` is run with `nuxi prepare`, clear buildDir
- dev server break :cry:                                         -- dev server restart 👍 
- `ctrl` + C, ⬆️ , `enter`, wait for server to restart --- nuxt ready

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -158,6 +158,11 @@ export default defineNuxtCommand({
 
     await load(false)
 
+    const distWatcher = chokidar.watch(resolve(rootDir, '.nuxt/dist'), { ignoreInitial: true })
+    distWatcher.on('unlinkDir', () => {
+      dLoad(true, '.nuxt/dist directory has been removed')
+    })
+
     return 'wait' as const
   }
 })
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -59,6 +59,8 @@ export default defineNuxtCommand({
     const { loadNuxt, buildNuxt } = await loadKit(rootDir)
 
     let currentNuxt: Nuxt
+    let distWatcher: chokidar.FSWatcher
+
     const showURL = () => {
       listener.showURL({
         // TODO: Normalize URL with trailing slash within schema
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/dev.ts
```diff
@@ -158,6 +158,11 @@ export default defineNuxtCommand({
 
     await load(false)
 
+    const distWatcher = chokidar.watch(resolve(rootDir, '.nuxt/dist'), { ignoreInitial: true })
```
danielroe, line 161: I think we can use nuxt.options.buildDir + dist

danielroe (commented):
What a great idea!

packages/nuxi/src/commands/dev.ts
huang-julien: yes :+1:

packages/nuxi/src/commands/dev.ts
```diff
@@ -75,11 +77,20 @@ export default defineNuxtCommand({
...
+        if (distWatcher) {
+          distWatcher.close()
+        }
+
         currentNuxt = await loadNuxt({ rootDir, dev: true, ready: false })
         if (!isRestart) {
           showURL()
         }
 
+        distWatcher = chokidar.watch(resolve(currentNuxt.options.buildDir, 'dist'), { ignoreInitial: true })
```
danielroe, line 89: shall we set `depth: 0` to only watch the directory itself?

danielroe (approved):
This is a very nice fix.

pi0 (approved):
Good idea! We might extend/change it later to other dirs such as `.cache` since we not always remove top level dist

Issue #19136, opened 2023-02-18
Also allow `<NuxtLink>` to execute prefetch when `to` type props are `RouteLocationRaw`


Mini-ghost:
### Describe the feature

In earlier issue, I mention a feature request that `preloadRouteComponents` allows passing values of type `RouteLocationRaw`👉 #18845 .

But I fund that the `<NuxtLink>` prefetch (or preload) just only execute when the props `to` type is string,

```html
<!-- Passing a string -->
<!-- It will prefetch -->
<NuxtLink to="/home"> Home </NuxtLink>

<!-- Passing a string with query string -->
<!-- It will prefetch -->
<NuxtLink to="/home?year=2023"> Home </NuxtLink>

<!-- Passing a object -->
<!-- It will no prefetch -->
<NuxtLink :to="{ name: 'home', query: { year: '2023' } }"> Home </NuxtLink>
```

That's because of the following code:
https://github.com/nuxt/nuxt/blob/524426ed64087d750fd6f9ee8c160d2e39a8a428/packages/nuxt/src/app/components/nuxt-link.ts#L179-L180

One of the conditions for `shouldPrefetch` is `typeof to.value === 'string'`. 

Is this expected? Or can this limit be remove?

If allow prefetch when `<NuxtLink>` to passing a value type of `RouteLocationRaw`, the  app runtime lifecycle hook `link:prefetch` type must also be updated together

https://github.com/nuxt/nuxt/blob/524426ed64087d750fd6f9ee8c160d2e39a8a428/packages/nuxt/src/app/nuxt.ts#L38

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/19144

PR #19144, opened 2023-02-18
fix(nuxt): prefetch object-syntax routes with `<NuxtLink>`

Mini-ghost:
### 🔗 Linked issue

- #19136

### ❓ Type of change

### 📚 Description

Based on PR #19120, the `preloadRouteComponents` function now accepts a `RouteLocationRaw` type as a value. This update has also been made to `<NuxtLink>` to allow `RouteLocationRaw` as a valid value type.

This PR does not modify the type of hook `link:prefetch`, but obtains `router.resolve(to.value).fullPath` before calling the hook, because others still need string path, rash modification may cause Breaking change.

https://github.com/nuxt/nuxt/blob/7f67fcec73016dcc31b232c92d4a50db4aaef9fc/packages/nuxt/src/pages/runtime/plugins/prefetch.client.ts#L23-L40

https://github.com/nuxt/nuxt/blob/7f67fcec73016dcc31b232c92d4a50db4aaef9fc/packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts#L26-L34

https://github.com/nuxt/nuxt/blob/7f67fcec73016dcc31b232c92d4a50db4aaef9fc/packages/nuxt/src/app/plugins/payload.client.ts#L14-L18

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -176,7 +176,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
       const el = process.server ? undefined : ref<HTMLElement | null>(null)
       if (process.client) {
         checkPropConflicts(props, 'prefetch', 'noPrefetch')
-        const shouldPrefetch = props.prefetch !== false && props.noPrefetch !== true && typeof to.value === 'string' && props.target !== '_blank' && !isSlowConnection()
+        const shouldPrefetch = props.prefetch !== false && props.noPrefetch !== true && props.target !== '_blank' && !isSlowConnection()
         if (shouldPrefetch) {
           const nuxtApp = useNuxtApp()
           let idleId: number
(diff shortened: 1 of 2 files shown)


danielroe (commented):
1. we currently support prefetching _external_ links if `experimental.crossOriginPrefetch` is enabled, which the router won't be able to resolve. Maybe worth adding a check for whether a link is external?
2. if we resolve in the `link:prefetch` hook we'll end up double-resolving
3. possible solution to 1 and reduction in resolutions - we can skip if `to.value` is already a string?

Mini-ghost:
If it is an external link, `to.value` is a string, so I only judge whether `to.value` is a string, if yes, skip it, if not, execute `router.resolve(to.value)` .

atinux (approved):

danielroe (approved):

nathanchase:
FYI: Something in this PR broke NuxtLink, and am getting flooded with
`[Vue warn]: Extraneous non-props attributes (rel) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.` anywher that uses a custom NuxtLink template.

Example:
```
<NuxtLink
   v-slot="{ href }"
  :custom="true"
  :to="`/user/${id}`"
>
    <a :href="href">
      <img src="avatar.png">
    </a>
</NuxtLink>
```

This didn't occur in 3.2.2, but does in 3.2.3. - I will try to create a repro of the error in a nuxt.new StackBlitz.

nathanchase:
StackBlitz with the error:
https://stackblitz.com/edit/github-hmvahh?file=app.vue

danielroe:
That should be resolved in the [edge channel](https://nuxt.com/docs/guide/going-further/edge-channel/), or in the next release with https://github.com/nuxt/nuxt/pull/19379

PR #19148, opened 2023-02-18
fix(nuxt): expose internal nitro type alias

danielroe:
### ❓ Type of change

### 📚 Description

`useStorage` is currently not typed in a Nuxt project as internally the `#internal/nitro` type is used in the Nitro auto-imported `useStorage`. This PR provides a missing alias that allows that to be typed properly.

### 📝 Checklist


pi0:
I'm worried by this, we also expose internal types from subpaths that can be changes in non semver-major. Is there anything better we can do to expose utils like `useStorage` from a subpath like `nitro/runtime` maybe.

danielroe:
The issue is the Nitro type for `useStorage` references the internal alias. Maybe it can use a relative path instead?

danielroe:
@pi0 Any further thoughts on this? I would like to merge this or resolve the bug upstream.

pi0:
We shouldn’t expose internal types of nitro. We should probably fix in upstream exporting public import types :)

danielroe:
The upstream problem to be solved is that the upstream [generated `.d.ts`](https://unpkg.com/browse/nitropack@2.2.3/dist/runtime/index.d.ts) relies on the internal alias:

```ts
export { useStorage } from "#internal/nitro/virtual/storage";
```

pi0:
Indeed. And it shouldn't (not are virtuals are to be exposed but some like `useStorage` are and should have been outside).

danielroe:
~> https://github.com/unjs/nitro/pull/1026 🙌

Issue #15592, opened 2023-01-13
Custom config schema


pi0:
In Nuxt 3, we have introduced a new package `@nuxt/schema` that holds types, defaults, and documentation for all build-in configurations for Nuxt and from one source using [untyped](https://github.com/unjs/untyped). In comparation with Nuxt 2, these were separately maintained, often with inconsistencies. This new system is also designed to be future proof to be able to **manage configuration via UI** using generated (JSON)Schema files and also **runtime validation** to notice any usage issues.

For Nuxt modules, it is already possible to extend a typescript interface `ModuleOptions` to provide types, however it has same limitations of Nuxt 2 era. We have introduced new `schema` in new `@nuxt/kit` module definition system however it became unused.

With the introduction of Nuxt Layers, themes can also now require some specific configuration. Most importantly for configuring `runtimeConfig` and app config (either via `appConfig` or `app.config`)

For this, we can use the same tooling (untyped) as a built-in feature to allow providing custom configuration schemas from Nuxt layers and modules (also directly from a Nuxt project). Benefiting from most possibilities (note 1) we have already for core config.

We have made an experiment [nuxt-experiments/nuxt-config-schema](https://github.com/nuxt-experiments/nuxt-config-schema) that is now almost ready to be moved into a core experiment, allowing to make it better usable. It is possible to provide custom config schema for nuxt projects in different ways:

- Using `nuxt.schema.ts` in main Nuxt project or a layer
- Using `$schema` key in `nuxt.config`
- Extend `nuxt.options.$schema` for modules
- Extend schema using `schema:extend`,  `schema:resolved` and `schema:beforeWrite` for modules (note 2)
- Use `schema` in `defineNuxtModule` (recommended way for modules. also supports defaults)

The output will be in `.nuxt` directory and initially mainly usable for the Augmentation of types and custom purposes such as UI.

**Notes:**

note 1: Since development and trial of the experimental module, one particularly important point we found is that, unlike the core usage, mixing schema with defaults is not a good practice. We can only merge schema after modules are setup and it is too late for them to **re**apply new defaults and do their config handling logic and also it makes it much harder to merge runtime configuration such as `app.config` since custom merger logic resides in runtime while schema is in the build-time namespace and merging strategy is different.

[...]

PR #19172, opened 2023-02-20
feat(nuxt): enable config schema by default

pi0:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/15592

### ❓ Type of change

### 📚 Description

Changes: 
- Enable schema without experimental flag to advocate using by module authors
- Generate schema with `nuxt prepare`
- Remove unnecessary markdown schema generation (integrations can still do the same!)
- Always write `.nuxt/schema` directory to avoid confusion and removal
- Extend AppConfig and AppConfigInput from internal `_AppConfig` to allow merging

### 📝 Checklist


atinux (approved):
Looks good to me at first sight, will have to try with nuxt-edge to see if it works properly on nuxt-icon for instance.

packages/schema/src/types/config.ts
```diff
@@ -135,7 +135,10 @@ export interface RuntimeConfig extends RuntimeConfigNamespace {
 }
 
 // -- App Config --
-export interface AppConfigInput extends Record<string, any> {
+
+export interface _AppConfig extends Record<string, any> { }
```
pi0, line 139: I'm thinking to rename it back to `CustomAppConfig` (we might use `_AppConfig` for merged type of `useAppConfig` later similar to plugins). Wyt @danielroe ?

danielroe:
Apologies for delay in reviewing properly. I'm planning to merge this in for v3.3 release (so ideally 1 Mar merge directly after patch release, then a week of testing on edge). Does that work for you @Atinux?

pi0:
@danielroe I think we should start thinking of semver/patch release branches this way we have more time for edge testing not blocked by patches :)

danielroe:
If you need more time to test, we can release the changes in next patch as long as we preserve default flag value as `false`.

pi0:
No i need to test the default behavior via edge channel on main branch :) Patch shouldn't be blocker will continue this on discord later today.

packages/schema/src/types/config.ts
danielroe: `CustomAppConfig` sounds better, and more understandable, for sure.

pi0:
Disabled flag for now to land into edge/patch without being blocked.

Checking (more generic) non blocker edge testing next week 👍🏼

test/fixtures/basic/types.ts
```diff
@@ -246,7 +246,7 @@ describe('app config', () => {
       nested: {
         val: number
       },
-      userConfig: number
+      userConfig?: number
```
danielroe, line 249: Is it correct that this is possibly undefined? We have both a default value in schema and also (when merged) also a value in user-provided config that we know about.

pi0: Schema is partial, so yes this is _expected_ (modules might or might not finally provide the schema values. and we do not load defaults from schema)

 But if you know a way to make it required when user provides value, feel free to push it :)

packages/nuxt/src/core/schema.ts
```diff
@@ -39,9 +38,14 @@ export default defineNuxtModule({
...
 
     // Register module types
-    nuxt.hook('prepare:types', (ctx) => {
+    nuxt.hook('prepare:types', async (ctx) => {
       ctx.references.push({ path: 'nuxt-config-schema' })
       ctx.references.push({ path: 'schema/nuxt.schema.d.ts' })
+      if (nuxt.options._prepare) {
+        await nuxt.hooks.callHook('schema:beforeWrite', schema)
+        await writeSchema(schema)
+        await nuxt.hooks.callHook('schema:written')
```
danielroe, line 47: would it make sense to move the hooks into `writeSchema` itself?

danielroe (approved):

PR #19173, opened 2023-02-20
feat(schema): allow adding page routes without a matching file

### LINKED ISSUE #19115: Defining a route (NuxtPage) requires a `file`, but a redirect does not need a file
### Describe the feature

The `pages:extend` hook gets an array of `NuxtPage` items passed. When adding another item, the `file` property is mandatory. That makes sense for routes that actually render something, but doesn't make sense for routes that are just meant to redirect to another route.

So currently, one needs to provide a file even though it's not used

A full example:
```
'pages:extend': async (routes) => {
  routes.push({
    name: 'index',
    path: '/',
    redirect: '/home',
    file: '~/pages/home.vue' // Needs to be provided here
  })
  routes.push({
    name: 'index',
    path: '/home',
    file: '~/pages/home.vue'
  })
}
```

The type definition of `NuxtPage`:
```
type NuxtPage = {
    name?: string;
    path: string;
    file: string;  // <-- This should be optional (in case of a redirect)
    meta?: Record<string, any>;
    alias?: string[] | string;
    redirect?: string;
    children?: NuxtPage[];
};
```

### Additional information

### Final checks

rebyatyonok:
### 🔗 Linked issue

#19115 

### ❓ Type of change

### 📚 Description

Resolves #19115 

Makes NuxtPage.file optional. 

### 📝 Checklist

- [ ] I have updated the documentation accordingly (there is no need?).


danielroe (approved):

danielroe (dismissed):
Type check is (rightly) picking up some issues where we need to confirm that pages _do_ have a file, for example in `getSources` and in generating route metadata.

Would you be able to update those places as well to filter out routes without associated files?

rebyatyonok:
Yeah, sure

packages/nuxt/src/pages/utils.ts
```diff
@@ -249,7 +249,7 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> =
   return {
     imports: metaImports,
     routes: genArrayFromRaw(routes.map((page) => {
-      const file = normalize(page.file)
+      const file = normalize(page.file || '')
```
rebyatyonok, line 252: I'm still not sure about this line. Should some random name be generated instead? Or pages without actual file should be filtered out?

danielroe: No, they should not be filtered out here. Nor should a random path be generated. This gets fed back into the source for `#build/routes.mjs`. (If we were to filter out these files then no route would be output without a filename, which I think is the opposite result you would like.)

https://github.com/nuxt/nuxt/blob/6016aef859efe485871fe007d9b572304916ebba/packages/nuxt/src/pages/module.ts#L168-L176

I think we can early return in this case without some of the properties on the object we normally return.

rebyatyonok: Probably fixed that. There are no errors with redirects in the playground as far as I can tell.

danielroe (approved):
Thank you! ❤️

I pushed a few refactors to the route generation utility that were unrelated to your work (your implementation was good 👌) which should hopefully improve the quality of the generated routes.

rebyatyonok:
@danielroe you're welcome, was glad to help ❤️

PR #19187, opened 2023-02-21
fix(nuxt): log errors thrown when booting the nuxt app

### LINKED ISSUE #19162: Onclick will not work if certain files are in the plugin folder
### Environment

I am running windows 10 and using firefox

Nuxt project info:                                                                                          

- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.2`
- Package Manager:  `npm@9.3.1`
- Builder:          `vite`
- User Config:      `css`, `build`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Create a vue page with a `@click` function on it that does something

Install `@urql/vue` using:
```

yarn add @urql/vue graphql
# or
npm install --save @urql/vue graphql
```

Next make a urql.js in plugin with this:

```
import { createClient, ssrExchange, dedupExchange, fetchExchange, Client } from '@urql/core';
import { cacheExchange } from '@urql/exchange-graphcache'


export default defineNuxtPlugin(nuxtApp => {

const ssrKey = '__URQL_DATA__'

const ssr = ssrExchange({
    isClient: process.client
  })

 // when app has rendered in server, send SSR state to client
if (process.server) {
    nuxtApp.hooks.hook('app:rendered', () => {
        nuxtApp.payload[ssrKey] = ssr.extractData() 
    })
}

// when app is created in browser, restore SSR state from nuxt payload
if (process.client) {
nuxtApp.hooks.hook('app:created', () => {
  ssr.restoreData(nuxt.payload[ssrKey])
})
}

//GraphCache setup
// const cacheConfig ={
//     schema,
//     keys:{},
//     resolvers:{
//         Query:{
//         //country: (_, args) => ({__typename: "Country", id: args.id})
//         },
//     },
// }

const cache = cacheExchange()

const client = createClient({
  url: 'http://localhost:8000/graphql',
  exchanges: [
            dedupExchange,
            cache,
            ssr, 
            fetchExchange,
        ],
});

nuxtApp.provide("urql", client);
nuxtApp.provide("$urql", client);

})
```

Run `npm run dev`

### Describe the bug

The click function will not fire. Not sure why.  If you have other javascript libraries like primevue or bootstrapvue the dropdown does not work too

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19162

### ❓ Type of change

### 📚 Description

This ensures we don't swallow plugin errors, in development mode at least. We could also consider adding a warning if user doesn't register their own handler?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -194,6 +194,11 @@ export function createNuxtApp (options: CreateOptions) {
     })
   }
 
+  // Log all errors in development
+  if (process.dev) {
+    nuxtApp.hook('app:error', (...args) => console.error('[nuxt] [dev]', ...args))
+  }
+
   // Expose runtime config
   const runtimeConfig = process.server
     ? options.ssrContext!.runtimeConfig
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -196,7 +196,7 @@ export function createNuxtApp (options: CreateOptions) {
 
   // Log errors captured when running plugins, in the `app:created` and `app:beforeMount` hooks
   // as well as when mounting the app and in the `app:mounted` hook
-  nuxtApp.hook('app:error', console.error)
+  nuxtApp.hook('app:error', (...args) => { console.error('[nuxt] error caught during app initialization', ...args) })
 
   // Expose runtime config
   const runtimeConfig = process.server
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -194,10 +194,9 @@ export function createNuxtApp (options: CreateOptions) {
     })
   }
 
-  // Log all errors in development
-  if (process.dev) {
-    nuxtApp.hook('app:error', (...args) => console.error('[nuxt] [dev]', ...args))
-  }
+  // Log errors captured when running plugins, in the `app:created` and `app:beforeMount` hooks
+  // as well as when mounting the app and in the `app:mounted` hook
+  nuxtApp.hook('app:error', console.error)
 
   // Expose runtime config
   const runtimeConfig = process.server


danielroe:
We do suppress errors in other places (in that we require users to handle them), specifically when it comes to universal router errors and in useAsyncData/useFetch. But in this case I am very persuadable.

pi0 (approved):

packages/nuxt/src/app/nuxt.ts
```diff
@@ -194,6 +194,10 @@ export function createNuxtApp (options: CreateOptions) {
     })
   }
 
+  // Log errors captured when running plugins, in the `app:created` and `app:beforeMount` hooks
+  // as well as when mounting the app and in the `app:mounted` hook
+  nuxtApp.hook('app:error', console.error)
```
pi0, line 199: Tag was also nice idea btw before. Not sure if also makes negative DX, in this case feel free to ignore this 

```suggestion
  nuxtApp.hook('app:error', error => console.error('[nuxt] [app]', error))
```

danielroe: The reason I'm not including it is (1) I don't think we should persist `[nuxt]` logging tags into production, and (2) this isn't really an error that Nuxt is throwing unlike other errors tagged `[nuxt]`.

I did initially think it might improve the stacktrace but hookable gets in the way, so it doesn't make much difference for DX - we either get a breakpoint in hookable's `index.mjs` or here in `nuxt.mjs`.

pi0: Mainly, this is to find the source of console errors (otherwise we are not sure if `console.error` is being called by user, nuxt or a library). And since this is a generic handler, i was thinking makes sense. I use this tags as standard method in unjs libraries too.

But also see that it might make another confusion that _nuxt_ is the source of the bug. What if make it more explicit: `Nuxt app caught error:`

danielroe: good idea!

pi0 (approved):

devilmark84:
Hello, I've just installed the new 3.2.3 version of Nuxt and now this happens when a page is not found.

```
[nuxt] error caught during app initialization H3Error: Page not found: /robots.txt                                                                                          11:13:52
    at Module.createError (file:///Users/marco/Desktop/nuxt-bootstrap/node_modules/h3/dist/index.mjs:128:15)
    at /Users/marco/Desktop/nuxt-bootstrap/node_modules/nuxt/dist/pages/runtime/plugins/router.mjs:141:113
    at triggerAfterEach (file:///Users/marco/Desktop/nuxt-bootstrap/node_modules/vue-router/dist/vue-router.mjs:3306:13)
    at file:///Users/marco/Desktop/nuxt-bootstrap/node_modules/vue-router/dist/vue-router.mjs:3209:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/marco/Desktop/nuxt-bootstrap/node_modules/nuxt/dist/pages/runtime/plugins/router.mjs:157:7
    at async createNuxtAppServer (/Users/marco/Desktop/nuxt-bootstrap/node_modules/nuxt/dist/app/entry.mjs:30:7)
    at async default (file:///Users/marco/Desktop/nuxt-bootstrap/node_modules/@nuxt/vite-builder/dist/runtime/vite-node.mjs:27:18)
    at async Object.renderToString (file:///Users/marco/Desktop/nuxt-bootstrap/node_modules/vue-bundle-renderer/dist/runtime.mjs:172:19)
    at async file:///Users/marco/Desktop/nuxt-bootstrap/.nuxt/dev/index.mjs:791:21 {
  statusCode: 404,
  fatal: false,
  unhandled: false,
  statusMessage: 'Page not found: /robots.txt',
  __nuxt_error: true
}
```

Is this behavior what you wanted? 🧐 How can I suppress these errors or catch them?

Thank you

pi0:
I also spotted same issue. I guess we might apply a filter to hide internal type of errors (i guess we added a field for it /cc @danielroe )

PR #19203, opened 2023-02-21
chore: remove docs for reactivity transform

milos018:

### 📚 Description

In a recent event, Vue founder announced giving up on reactivity transform.
It should be removed from Nuxt docs / Experimental page.

### 📝 Checklist


danielroe (changes requested):
We'll discuss this first as a team. One thing we'll consider is that reactivity transform is not yet gone and it is likely not going entirely, but moving into the capable maintenance of sxzz: https://github.com/sxzz/unplugin-vue-macros#readme
3.

**Note**, if we do deprecate and later remove, we should probably update Nuxt code as well to emit a warning (and once removed) tell people how to use the transform without this experimental option, and this could likely live in the docs page on the reactivity transform rather than removing it entirely, so we'll probably need to circle back to this PR later on.

For now, if you wanted to update the PR so we could merge it, you could amend the PR to add a warning that support for reactivity transform in Nuxt may well change in the future given Evan's announcement (and link to it).

danielroe:
Closing for now but I would still welcome a PR to add a warning that support for reactivity transform in Nuxt may well change in the future given Evan's announcement (and link to it). Feel free to update and ping me or open a new one. 🙏

Issue #14628, opened 2022-08-18
Define plugins with object syntax


pi0:
Nuxt plugins are async function hooks that run before anything else when initializing Nuxt app. They empower users and module authors to inject components, modify context and run any logic opon app initialization. However, there are limitations with function. A new additional object syntax, can increase possibilities:

```js
export default defineNuxtPlugin({
  name: 'my-plugin',
  enforce: 'after',
  hooks: {},
  islands: false,
  async setup() { }
})
```

`defineNuxtPlugin` wrapper (+ a runtime check) can convert plain fn functions into new object format.

Benefits of the new format:

- **Named plugins:** in logs and performance timings, we can properly show plugin name 
   - Previously depending on `Function.name` is also possible but tricky
- **Ordering plugins:** Plugins can request their expectation order by definition in runtime.
   - Previously was only possible by adjusting names or manually adding to config. Module additions were unshifted as workaround.
- **Hook-driven plugins:** Plugins can use hooks for implementation and be orderless
- **Consistency:** Format is closer to module and component defnitions with `async setup()`
- **Declarative behavior:** Plugins can declare their custom behavior (such as disabling for `islands` mode)

Issue #12406, opened 2023-01-06
AsyncData with a shallow ref


P4sca1:
### Describe the feature

Right now, `useAsyncData` uses a `ref` to store the data. This means that, when returning an object (or an array), the whole object will be made reactive. This can have unwanted performance penalties, depending on the size of the data that is fetched, and depending on the use-case, is often unneeded when returning static data from an API call that is not meant to be changed.
It would be great if you could use a `shallowRef` instead.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

Baroshem:
This could really make a significant difference IMO. I can work on this in the upcoming weeks once we decide that it should be implemented @danielroe 🙂

PR #19205, opened 2023-02-21
feat(nuxt): experimental option for rich json payloads

### LINKED ISSUE #19201: Switch back to devalue mainstream
Context: [Tweet](https://twitter.com/rich_harris/status/1628054361095000069?s=46&t=UOhaZgoecAWFwMnbm4WjDA) shared by @Atinux.

For serializing and restoring SSR state, we started using [devalue](https://github.com/Rich-Harris/devalue) in Nuxt 2. We made a fork [@nuxt/devalue](https://github.com/nuxt-contrib/devalue) in order to prevent hard errors about non-serializable objects plus few security fixes. Since this issue happens less often, we might switch back to upstream. (new [custom types](https://github.com/Rich-Harris/devalue#custom-types), could be also integrated via nuxt plugins)

**Note:** Newer version requires **runtime parser** function to be added to app if we want to switch to pure JSON.

pi0:
### 🔗 Linked issue

Resolves #19201

### ❓ Type of change

### 📚 Description

Started PR as a smaller refactor to switch to devalue v4. However other than the `uneval` function that works identically to the current version of devalue in Nuxt, devalue now supports native JSON and reduces/revives with it. This has many advantages over self-evaluating scripts we had been always using for Nuxt since devalue. 

Note: This PR is a work in progress and needs more testing, iterations, and reactors. Will be working with @danielroe on it. Meanwhile, feedback is more than welcome regarding the direction and API.

New features:

- Using [devalue v4](https://github.com/Rich-Harris/devalue)
- Using `<script type="application/json">` for payloads (instead of injecting a self-evaluating `window.__NUXT__` payload)
- Payload scripts are loaded using cross-origin fetch now (instead of inline `import` in HTML)
- Custom Types can be registered

TODO:

- [x] Discuss and validate the proposed custom types system
- [ ] Benchmark JSON performances over the current self-eval 
- [x] Verify XSS not happening 
- [x] Verify payload extraction mode behavior

### Custom Payload Types

Custom Payload types need to be globally registered by users so that we can reduce (serialize) them on rendering and revive (desrialize) them when initializing the Nuxt application. Previously things like custom classes were considered non-pojos and ignored.

```ts
// plugins/vector.ts
class Vector {
  x: number
  y: number
  constructor (x: number, y: number) {
    this.x = x
    this.y = y
  }

  magnitude () {
    return Math.sqrt(this.x * this.x + this.y * this.y)
  }
}

definePayloadType('Vector',
  value => value instanceof Vector && [value.x, value.y],
  ([x, y]) => new Vector(x, y)
)

  return {
    provide: {
      Vector
    }
  }
})
```

```vue
<script setup lang="ts">

const test = [1, 2, 3]
const { $Vector } = useNuxtApp()

const s = useState('s', () => ({
  vector: new $Vector(30, 40),
  test,
  nested: {
    test
  }
}))

</script>

<template>
  <div>
    {{ s }}
  </div>
</template>
```

HTML:

image

Output (hydrated!)

```
{ "vector": { "x": 30, "y": 40 }, "test": [ 1, 2, 3 ], "nested": { "test": [ 1, 2, 3 ] } }
```

### 📝 Checklist


danielroe:
Some key changes for consideration:

1. The feature is now opt-in, so we can merge and iterate over the 3.4 release cycle, and hopefully release stably in 3.5.
2. Registering custom types is now done in Nuxt plugins to avoid the risk of XRSP. Once we add [object syntax plugins](https://github.com/nuxt/nuxt/issues/14628), we can add an easy way for users to ensure their plugin runs at the right time. We can still consider adding global type support later (maybe for `$fetch` integration? 👀).
3. We support default Vue types of ref/reactive/shallowRef/shallowReactive as well as NuxtError, which will automatically be handled. This allows us to decrease memory usage on client by using `shallowReactive` for payload. I plan to add support shortly for using `shallowRef` for their useAsyncData/useFetch requests by passing `default: shallowRef`, resolving https://github.com/nuxt/nuxt/issues/12406.

Would very much appreciate any thoughts/review, particularly from @pi0 ❤️

pi0:
Thanks for the updates specially making it opt-in @danielroe 💯 Changes look good to me overall.

danielroe (approved):

Issue #19199, opened 2023-02-21
Issue with nuxt.socket.io and production server (works locally with npm run dev)


dataexcess:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v15.5.1`
- Nuxt Version:     `3.0.0-rc.13`
- Nitro Version:    `0.6.1`
- Package Manager:  `npm@7.3.0`
- Builder:          `vite`
- User Config:      `typescript`, `modules`, `css`, `components`, `runtimeConfig`, `app`, `io`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.1.3`, `@pinia/nuxt@0.4.6`, `nuxt-directus@3.2.2`, `@nuxt/image-edge@1.0.0-27821548.ab054e4`, `nuxt-socket-io@3.0.12`
- Build Modules:    `-`
------------------------------

### Reproduction

A nuxt 3 server instance with nuxt.socket.io served behind a nginx reverse proxy...

### Describe the bug

I am using nuxt.socket.io to let my client and server communicate with eachother... It works perfectly when building locally with 'npm run dev' and with this nginx.config.js file object for nuxt.socket.io:

```
    io: {
        sockets: [{
            name: 'mysite',
            url: 'http://localhost:3000',
            default: true
        }]
    }
```

Now I want to use it with my server. I build my nuxt server instance with 'npm run build' and then 'npm nuxi preview' or 'node  .output/server/index.mjs.

When I visit my website however I get the following 404 error in the client:

`Failed to load resource: the server responded with a status of 404 (Page not found: /socket.io/?EIO=4&transport=polling&t=OPqcaSF)`

Somehow the socket.io server seems not to be created in a production build? Why is this working when I am testing locally? Can it be a difference in the way the server is build after using 'npm run dev' versus 'node .output/server/index.mjs' ? Currently I can overcome this issue temporarily by using a dev build instead of a production build, but of course that build is not optimised. I really need a production build to be able to work with nuxt.socket.io.

Any help much appreciated :) 

### Logs

```shell-script
These are the logs from the server side:

[Vue Router warn]: No match found for location with path "/socket.io/?EIO=4&transport=polling&t=OPqvhgn"
[Vue Router warn]: No match found for location with path "/socket.io/?EIO=4&transport=polling&t=OPqvhgn"
[Vue Router warn]: No match found for location with path "/socket.io/?EIO=4&transport=polling&t=OPqvhgn"
[...]

PR #19230, opened 2023-02-22
docs: add information on running websocket on same port as nuxt

bf:
I have spent all day trying to figure out why proxy'ing socket.io through nuxt did not work. All the modules like `nuxt-socket-io` could not help me to achieve this. After deep diving into h3 and nitro code I noticed that h3 does not support `upgrade` requests (which makes sense from nuxt3) design perspective.

So I think we should openly document this and advise people to use a proxy in front of nuxt3 if they want to put everything on the same url / port.

### ❓ Type of change

### 📚 Description

Nuxt3 and socket.io cannot be on same port, even when using proxy or various npm modules.

https://github.com/nuxt/nuxt/issues/19199
https://github.com/nuxt/nuxt/issues/14786

### 📝 Checklist

- [ x] I have linked an issue or discussion.
- [x ] I have updated the documentation accordingly.

### DIFF
--- a/docs/2.guide/3.going-further/12.websocket-servers.md
+++ b/docs/2.guide/3.going-further/12.websocket-servers.md
@@ -47,7 +47,14 @@ const proxyServer = http.createServer((req, res) => {
 
 // upgrade client connection to websocket if requested
 proxyServer.on('upgrade', (req, socket, head) => {
-  proxySocketIO.ws(req, socket, head)
+  const isSocketIo = req.url.startsWith("/socket.io")
+
+  // choose backend depending on url
+  if (isSocketIo) {
+    proxySocketIO.ws(req, socket, head)
+  } else {
+    proxyNuxt.ws(req, socket, head)
+  }
 })
 
 proxyServer.listen(4000, () => console.log("proxy is listening"))
--- a/docs/2.guide/3.going-further/12.websocket-servers.md
+++ b/docs/2.guide/3.going-further/12.websocket-servers.md
@@ -34,9 +34,13 @@ import http from 'node:http'
 const proxySocketIO = httpProxy.createProxyServer({ target: 'http://localhost:3001' });
 const proxyNuxt = httpProxy.createProxyServer({ target: 'http://localhost:3000' });
 
-const proxyServer = http.createServer((req, res) => {
+// error handlers needed, otherwise it would `throw err` on client connection resets
+proxySocketIO.on("error", (err) => console.error("proxySocketIO error:", err))
+proxyNuxt.on("error", (err) => console.error("proxyNuxt error:", err))
+
+// create http server to merge both proxies on same port
+const httpServer = http.createServer((req, res) => {
   const isSocketIo = req.url.startsWith("/socket.io")
-  
   // choose backend depending on url
   if (isSocketIo) {
     proxySocketIO.web(req, res)
@@ -46,10 +50,8 @@ const proxyServer = http.createServer((req, res) => {
 })
 
 // upgrade client connection to websocket if requested
-proxyServer.on('upgrade', (req, socket, head) => {
+httpServer.on('upgrade', (req, socket, head) => {
   const isSocketIo = req.url.startsWith("/socket.io")
-
-  // choose backend depending on url
   if (isSocketIo) {
     proxySocketIO.ws(req, socket, head)
   } else {
@@ -57,7 +59,8 @@ proxyServer.on('upgrade', (req, socket, head) => {
   }
 })
 
-proxyServer.listen(4000, () => console.log("proxy is listening"))
+// listen for incoming data
+httpServer.listen(3000, () => console.log("proxy is listening"))
 ```
 
 The socket.io server can very easily be created to listen on port `3001`.
--- a/docs/2.guide/3.going-further/12.websocket-servers.md
+++ b/docs/2.guide/3.going-further/12.websocket-servers.md
@@ -1,0 +1,108 @@
+---
+title: "Websocket servers on same port as Nuxt3"
+description: "Nuxt3 websocket servers need a workaround."
+---
+
+
+# Websocket server support in Nuxt3
+
+Nuxt3 is built upon the powerful [nitro server engine](https://nitro.unjs.io/). Which in turn utilizes the [h3 HTTP framework](https://github.com/unjs/h3).
+
+Unfortunately, due to design choices of the underlying `h3` framework, true websocket connections cannot be natively etsablished over the Nuxt3 built-in web server.
+This means that the Nuxt3 server and your websocket server need to run on different ports.
+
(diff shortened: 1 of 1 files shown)


dataexcess:
Thank you @bf for mentioning this. Given thart you tried to make production server work with nuxt.socket.io, what would be a solution to this problem? Is it possible without changing nuxt.socket.io's code? 

Currently I am forced to run a nuxt dev build on my server which makes it work, but of course ideally I would like a production build on my server. 

Thank you this is of great help to me!

bf:
@dataexcess yesterday I tried everything. 

Problem with the nuxt-socket-io module is:
- it does not work with nuxt3 yet
- the maintainer has closed all issues because he was upset with people asking for help
- the maintainer is inactive since january

So yesterday I sat down and did a deep dive into nuxt, nitro and h3. I tried nitro poxy feature in various ways, I tried creating own server-middleware, creating proper nitro plugin, creating server-only routes. The only solution that ended up working for me was runnig a proxy in front of nuxt which checks if the incoming request is to `/socket.io/`  or not. In the end I figured out it was because h3 does not support the upgrade request.

*HOWEVER*, I just noticed that nuxt-devtools seems to run a websocket on the same port as nuxt at `ws://127.0.0.1:3000/__nuxt_devtools__/entry`. This is unlike vite which runs a websocket on port `24678` to circumvent the whole proxying of `upgrade: websocekt` requests problem.

We should check nuxt-devtools source code to see how they do it.

bf:
Relevant sourcecode from nuxt-devtools is https://github.com/nuxt/devtools/blob/3b33a602b34fe7a48dd4bb7f747941d1ebed4448/packages/devtools/src/module-main.ts  and https://github.com/nuxt/devtools/blob/3b33a602b34fe7a48dd4bb7f747941d1ebed4448/packages/devtools/src/rpc.ts

The problem with the nuxt-devtools websocket approach is that it is not working in production, see https://github.com/nuxt/devtools/blob/3b33a602b34fe7a48dd4bb7f747941d1ebed4448/packages/devtools/src/module-main.ts#L25

So my goal was to find something where I have exactly the same websocket setup in development and in production.

I know that you can always run the websocket on a different port and merge both via nginx reverse proxy in deployment. However, I wanted to remove this logic from nginx and put it into the javascript part of my app.

dataexcess:
Ok interesting!

Yes I am quite annoyed and disappointed with the nuxt.socket.io maintainer as he describes in his 'contributing' page: "Since the nuxt-socket-io has reached a stable state I'm currently disabling the "issues" feature, since it encourages more whining than actual problem solving" 

I am not sure how opening an issue saying that the project does not work on nuxt3 as is claimed on the read me of his project would be sonsidered 'whining' sigh.

That aside, as I understand from your thorough research (thank you for this!) it would be possible simply by running the nuxt.socket.io websocket server on a differenet port right? I am just unsure how to do this exactly. I am using nginx already on my server so I thought about this as a nginx config setup:

```
server {
       #name / ssl / etc....

        location / {
                add_header 'Access-Control-Allow-Origin' '*' always;
                proxy_pass https://localhost:3001;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection 'upgrade';
                proxy_set_header Host $host;
                proxy_cache_bypass $http_upgrade;

                client_max_body_size 100M;
        }

       location /socket.io/ {
               add_header 'Access-Control-Allow-Origin' '*' always;
                proxy_pass https://localhost:3002;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection 'upgrade';
                proxy_set_header Host $host;
       }
}
```

In this case all traffic to websockets would be routed to the sockets server right? The only problem I am having is that I don't know how to actually start this websocket server on port 3002 (using nuxt.socket.io) There is a 'url' field in the io object in the nuxt.config.js file but that is the one that is used from the client side if I understand it correctly... so that should be set to https://mysite.com and not https://localhost:3002. 

Do you have any idea? Or am I doomed to ditch nuxt.socket.io and go for your other solution.. :( hoping to still make it work with my current code. 

Any help much appreciated!

bf:
Just ditch nuxt-socket-io. Please don't blame the maintainer, he is obviously a bit overwhelmed with the situation. Open-source is a horrific thing because most of the time nobody pays you for anything and then people come to you for support and complain.

If you want to use nginx to merge both ports into one, then just create a nuxt server plugin (nitro plugin) like this:

```
// plugins/socket.io.ts
import { Server } from 'socket.io';

export default defineNitroPlugin((nitroApp) => {


const io = new Server(3000, {
  serveClient: false,
  cors: {
      origin: '*',
  }
});

io.engine.on("connection_error", (err) => {
  console.log("socket.io connection error", err)
})

io.on('connection', (socket) => {
  console.log('Connection', socket.id)
})

io.on('connect', (socket) => {
  socket.on("error", (msg) => {
    console.log("socket error", msg)
  })

  socket.emit('message', `welcome ${socket.id}`)
  socket.broadcast.emit('message', `${socket.id} joined`)

  socket.on('message', (data) => {
    console.log('message received: %s', data)
    socket.emit('message', { data })
  })

  socket.on('disconnecting', () => {
    console.log('disconnected', socket.id)
    socket.broadcast.emit('message', `${socket.id} left`)
  })
});
})
```

dataexcess:
Thank you I will try this code!

I am not blaming the maintainer I am really grateful for his work and I understand it is not easy. But to get rid of the issues section all-together seems a bit drastic to me personally. At least people can just mention what is not working and make other people aware before using this in development mode expecting for it to work in production (especially when it is stated in the read me file) I am not even saying he should fix the issue immediately but at least he could update the read me file so people in the future will not run into the same issue as I have.

Actually now that I think of it I could/should make a pull request editing the read me file mentioning just this.

dataexcess:
@bf Thank you your code did the trick 🙏

danielroe:
@pi0 @Atinux What do you think - would this be good to add to the Nuxt docs? Or maybe as an example in https://github.com/nuxt/examples?

pi0:
Please hold on this till adding ws support to to nitro. Custom preset like this can be broken in the future if we recommend.

timstrasser:
@bf Is there a working example project for this by any chance? I'm not sure where exactly to put the code snippets.

tobiasdiez:
> Please hold on this till adding ws support to to nitro. Custom preset like this can be broken in the future if we recommend.

Is there any timeline for ws support in nitro? Its needed if you want to use subscriptions in graphql (https://github.com/tobiasdiez/nuxt-graphql-server/issues/23).

danielroe:
linking issue this is waiting on: https://github.com/nuxt/cli/issues/108

danielroe:
This is now outdated as Nitro comes with first-class websocket support (on the same port) and I think we can close now. Let me know if you think there is anything more to update here and I'll happily reopen. 🙏

PR #19231, opened 2023-02-22
fix(nuxt): skip hydration mismatches with client components

huang-julien:
### 🔗 Linked issue

https://nuxt.com/docs/guide/directory-structure/components#paired-with-a-client-component

### ❓ Type of change

### 📚 Description

Hi 👋 this PR solves the hydration issues between `.client` and `.server` components.

This allow to have different html on initial loads, even if both does not have the same number of root nodes.

The "workaround" is to render a `staticVNode` before hydration and then render the `client` component when mounted.

Directives and attributes are removed before hydration (and set back on mounted) since `staticVNodes` canno't inherit attributes or receive directives. 

About the component `template` property, I didn't found a working solution for the runtime compiler, but since this isn't supported yet, I think this maybe removed or left as it is or even implemented with the support vue full build PR. 

### 📝 Checklist


packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -73,3 +94,50 @@ export function createClientOnly (component) {
...
 }
+
+function getStaticVNode (vnode) {
+  const fragment = getFragmentHTML(vnode.el)
+
+  if (fragment.length === 0) {
+    return null
+  }
+  return createStaticVNode(fragment.join(''), fragment.length)
+}
```
danielroe, line 105: Is there any way you think we can cut down on the extra code here and reduce bundle size when using client components? Is any of this shared with our server-component implementation?

huang-julien: sure we can :+1: ! Server components are just a copy/paste of theses lines of code 😃

packages/nuxt/src/app/components/client-only.ts
```diff
@@ -51,8 +53,23 @@ export function createClientOnly<T extends ComponentOptions> (component: T) {
...
   clone.setup = (props, ctx) => {
+    const instance = getCurrentInstance()!
+
+    // @ts-expect-error internal
+    const inheritAttrs = instance.inheritAttrs
+    // remove exsting directives during hydration
+    const directives = extractDirectives(instance)
+    // prevent attrs inheritance since a staticVNode is rendered before hydration
+    // @ts-expect-error internal
+    instance.inheritAttrs = false
```
huang-julien, line 64: not sure if it's a good idea to rely on internals

danielroe: yes, I'm not sure I would like to do that either. 🤔

antfu (approved):
Smart :) LGTM. Now I feel at some point this should be a built-in component of Vue.

danielroe (approved):

PR #19243, opened 2023-02-23
feat(vite): add `node_modules` and buildDir to `x_google_ignoreList`

danielroe:
### ❓ Type of change

### 📚 Description

Rollup now supports (https://github.com/rollup/rollup/pull/4848) adding certain files to the [`x_google_ignoreList` source map extension](https://developer.chrome.com/blog/devtools-better-angular-debugging/#the-x_google_ignorelist-source-map-extension). We can opt-in to this to provide a better default experience on Chrome.

| Before | After |
| - | - |
| CleanShot 2023-02-23 at 15 17 10@2x | CleanShot 2023-02-23 at 15 16 18@2x |

**Limitations**: This is a specific enhancement only for vite. More significantly, at the moment this is only processed in the rollup/vite production build rather than in development, which is a bit of a pain. ~~I've marked this as draft to investigate further. It might require an upstream vite enhancement.~~ This will need to be enabled upstream in vite first before being configurable here.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -63,6 +63,9 @@ export async function bundle (nuxt: Nuxt) {
           copyPublicDir: false,
           rollupOptions: {
             output: {
+              sourcemapIgnoreList: (relativeSourcePath) => {
+                return relativeSourcePath.includes('/node_modules/') || relativeSourcePath.includes(ctx.nuxt.options.buildDir)
+              },
               sanitizeFileName: sanitizeFilePath,
               // https://github.com/vitejs/vite/tree/main/packages/vite/src/node/build.ts#L464-L478
               assetFileNames: nuxt.options.dev


bmeurer:
Nice! I just captured this in our [tracking doc](https://goo.gle/devtools-ignoreList-adoption#bookmark=id.qucvob6jw8nu). 👍

packages/vite/src/vite.ts
```diff
@@ -63,6 +63,9 @@ export async function bundle (nuxt: Nuxt) {
           copyPublicDir: false,
           rollupOptions: {
             output: {
+              sourcemapIgnoreList: (relativeSourcePath) => {
+                return relativeSourcePath.includes('/node_modules/') || relativeSourcePath.includes(ctx.nuxt.options.buildDir)
```
bmeurer, line 67: Does this work on Windows? For `rollup` I had to change it to `relativeSourcePath.includes('node_modules')` to make the test pass on the Windows bot.

danielroe: updated in https://github.com/nuxt/nuxt/commit/73ade185b2e5d5fa1b2d659cb5c8f7d8d63a8ac5

PR #19252, opened 2023-02-23
feat(kit,nuxt): add component `priority` to allow overriding

### LINKED ISSUE #19221: Can't create a <Link> component.
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.14.0`
- Nuxt Version:     `3.2.0`
- Nitro Version:    `2.2.1`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `css`, `postcss`, `modules`
- Runtime Modules:  `@pinia/nuxt@0.4.6`, `@vueuse/nuxt@9.12.0`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-obfcsk?file=pages%2Findex.vue,pages%2Ffoo.vue,components%2FLink.vue,components%2FMyLink.vue

### Describe the bug

I get a next warning:
`WARN  Overriding Link component.`
and generated html is `<!-- -->`

### Additional context

I need to add extra prefix to all Links.
It works fine with Nuxt 2, but on migration process on Nuxt 3 it caught an issue

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19221

### ❓ Type of change

### 📚 Description

In line with the `priority` option support by auto-imported composables/utils, this PR adds a `priority` to components to allow them to override other components. By default, user components have a higher priority than components injected by modules without an explicit priority, but not higher than any Nuxt built-in components.

### 📝 Checklist


antfu (approved):

packages/schema/src/types/components.ts
```diff
@@ -10,6 +10,12 @@ export interface Component {
   global?: boolean
   island?: boolean
   mode?: 'client' | 'server' | 'all'
+  /**
+   * This number allows configuring the behaviour of overriding Nuxt components.
```
DamianGlowala, line 14:
```suggestion
   * This number allows configuring the behavior of overriding Nuxt components.
```

pi0:
I have initially removed `periority` (from Nuxt 2 => 3) in favor of respecting the order in the `components: []` array. Supporting for kit utils, seems a good idea 👍🏼  because it cannot control over the array push order. (But maybe user components should have something higher than modules by default?)

Regarding built-ins, I agree with your comment in https://github.com/nuxt/nuxt/issues/19221#issuecomment-1440259243. Same for built-in composables, fetch, etc, With auto imports and implicit sources, allowing a built-in to be overridden by user, can cause unwanted behavior by any module, theme layer, and code that was depending on Nuxt behavior and now is changed! (a layer can break other layer too). So i belive we might give Nuxt Built-ins priority too and document only as advanced _escape hatch_ to override them if really needed. 

Default priority: Built-ins > User Components > Layer 1 > Layer 2 > Module Aded [can be overridden with `peirority`]

Atinux:
What is the main advantage of overwriting built-ins components?

It can also create confusion for team members working on a Nuxt project with overwritten built-in components.

danielroe:
The Nuxt component built-ins are:

- NuxtWelcome
- NuxtErrorBoundary
- ClientOnly
- DevOnly
- ServerPlaceholder
- NuxtLink
- NuxtLoadingIndicator
- NuxtIsland
- NuxtLayout
- NuxtPage
- NoScript, Link, Base, Title, Meta, Style, Head, Html and Body

Of those, I think that it would be safe (and good DX) if users could provide their own implementations for `<NuxtLink>` and `<NuxtLoadingIndicator>`. Part of the point would be that by overriding their custom behaviour would then be picked up by any modules or layers (for example) that _did_ use those components.

For the other built-in components, their priority when registering them that ensures that it would require extra effort to override them, but still allows an escape hatch for advanced use cases.

wdyt?

> maybe user components should have something higher than modules by default

That is the [implementation here](https://github.com/nuxt/nuxt/pull/19252/files#diff-f05530fa474de8422c52d6d037f43d3a7f67d96c4f737f3f655304eb40386209R122-R123) - user components are given a default priority of 1, which means they will override module components.

PR #19256, opened 2023-02-24
fix(vite): handle non-iterable `noExternal` option

### LINKED ISSUE #19254: Vite Config ssr.noExternal erroring when true
### Environment

Breaks in stackblitz

### Reproduction

https://stackblitz.com/edit/github-sooc2d?file=vite.config.js

### Describe the bug

I'm trying to disable SSR on a particular module using `https://vitejs.dev/config/ssr-options.html#ssr-noexternal`. Found it wasn't doing what was expected when my vite config looked like:

```js
export default {
  ssr: {
    noExternal: ['stripe'],
  },
};
```

So I tried to disable all with `noExternal: true` and saw it broke my Nuxt project.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19254

### ❓ Type of change

### 📚 Description

This handles the case where `vite.ssr.noExternal` is not iterable (boolean, string or single string/RegExp)

### 📝 Checklist

### DIFF
--- a/packages/vite/src/utils/external.ts
+++ b/packages/vite/src/utils/external.ts
@@ -8,7 +8,7 @@ export function createIsExternal (viteServer: ViteDevServer, rootDir: string) {
       /virtual:/,
       /\.ts$/,
       ...ExternalsDefaults.inline || [],
-      ...viteServer.config.ssr.noExternal as string[]
+      ...Array.isArray(viteServer.config.ssr.noExternal) ? viteServer.config.ssr.noExternal : []
     ],
     external: [
       ...viteServer.config.ssr.external || [],


pi0:
In case of boolean (false) we are still doing kind of externals! In that case does it makes sense to also update schema to disallow it from vite config in nuxt config?

antfu:
Maybe we could throw an error to say it's not supported instead of silently ignoring it?

pi0:
That's what i was thinking to. Disallow both from types and value.

danielroe:
If the user sets `vite.ssr.noExternal` to `true`, this is actually supported by nuxt/vite and `isExternal` is never actually called - the setting is not ignored.

danielroe:
@antfu @pi0 Do you have any concerns with this PR or alternative suggestions?

pi0 (approved):
LGTM (considering `noExternal` as stated both true and false values are also respected)

danielroe:
Great. Yes, it works fine with `true` but `false` is not an allowable option for `noExternal`. The valid input options are `string | true | RegExp | (string | RegExp)[] | undefined`. `string` and `RegExp` are both normalized to arrays by `vite.mergeConfig`.

PR #19261, opened 2023-02-24
fix(nuxt): compatible route object for custom external routes

### LINKED ISSUE #19259: "[Vue Router warn]: No match found for location with path" on ANY NuxtLink with an external url
### Describe the feature

right now in Nuxt 3.2.2, any usage of NuxtLink to an external url gives a warning "[Vue Router warn]: No match found for location with path <path>".  

Even if you look at the example provided in the documentation and you look at the developer console, you'll see the warnings once the page loads up
https://nuxt.com/docs/examples/routing/nuxt-link#nuxtlink

This is unintuitive, because obviously external links will not be available for the underlying vue-router and it won't know about them.  The project I am working on is a conversion of a legacy site, and so because of that we have to link to external pages ALOT within the Nuxt part of the application to jump back and forth with the legacy site.  That means every SPA page has tons of warnings in the console.  

The documentation shows examples of using NuxtLink for both internal and external links, and it even has an "external" property for edge cases.  It would be really great if we could somehow disable those warnings when using external links or manually setting the external property on the NuxtLink component.

Right now the alternative for me is to render two completely different components, one that is a NuxtLink, and one that is a simple anchor tag.  This works, but isn't as neat as just using NuxtLink as it is currently documented.

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19259

### ❓ Type of change

### 📚 Description

We were calling `router.resolve()` on known external routes. We can either completely omit this route or we can parse it as we are doing in the universal router plugin. (I've chosen the latter as it's a non-breaking change if someone is already using the universal router + custom external links, but other thoughts welcome.)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -250,25 +250,17 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
             return null
           }
 
-          const props = {
+          return slots.default({
             href,
             navigate,
-            rel,
-            target,
-            isExternal: isExternal.value,
-            isActive: false,
-            isExactActive: false
-          }
-
-          Object.defineProperty(props, 'route', {
-            get () {
+            get route () {
               if (!href) { return undefined }
 
               const url = parseURL(href)
               return {
                 path: url.pathname,
                 fullPath: url.pathname,
-                query: parseQuery(url.search),
+                get query () { return parseQuery(url.search) },
                 hash: url.hash,
                 // stub properties for compat with vue-router
                 params: {},
@@ -278,10 +270,13 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
                 meta: {},
                 href
               }
-            }
+            },
+            rel,
+            target,
+            isExternal: isExternal.value,
+            isActive: false,
+            isExactActive: false
           })
-
-          return slots.default(props)
         }
 
         return h('a', { ref: el, href, rel, target }, slots.default?.())
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -1,7 +1,7 @@
 import type { PropType, DefineComponent, ComputedRef } from 'vue'
 import { defineComponent, h, ref, resolveComponent, computed, onMounted, onBeforeUnmount } from 'vue'
 import type { RouteLocationRaw } from 'vue-router'
-import { hasProtocol } from 'ufo'
+import { hasProtocol, parseQuery, parseURL } from 'ufo'
 
 import { preloadRouteComponents } from '../composables/preload'
 import { onNuxtReady } from '../composables/ready'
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -249,10 +249,25 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+          const url = href && parseURL(href)
           return slots.default({
             href,
             navigate,
-            route: router.resolve(href!),
+            route: url
+              ? {
+                  path: url.pathname,
+                  fullPath: url.pathname,
+                  query: parseQuery(url.search),
```
pi0, line 260: Small perf: this could be a getter

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -249,10 +249,25 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
           }
+          const url = href && parseURL(href)
           return slots.default({
             href,
             navigate,
-            route: router.resolve(href!),
+            route: url
+              ? {
+                  path: url.pathname,
+                  fullPath: url.pathname,
```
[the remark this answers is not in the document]
danielroe: that would be a nice perf. do you think we should make `route` itself a getter that calls `parseURL`?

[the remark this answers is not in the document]
pi0: Also better!

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -249,16 +249,39 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+
+          Object.defineProperty(props, 'route', {
+            get () {
+              if (!href) { return undefined }
+
+              const url = parseURL(href)
+              return {
+                path: url.pathname,
+                fullPath: url.pathname,
+                query: parseQuery(url.search),
```
pi0, line 271: Even if `.route` accessor is lazy, sub propers (especially this one could be). It might not have significant performance overhead but can also have when there are dozens of components rendering concurrently so probably micro-optimization worth :)

danielroe: The conditions required for this to be hit are that we are 1) using pages module, 2) using a custom link, 3) using an external link, 4) accessing `route` in this slot. I don't see there being dozens of components matching those conditions rendering concurrently; I think in most cases an external link will simply handle the case of the `href` rather than the route object at all.

My thought is that adding the code to handle the slight perf opt of accessing route.query lazily will add more kb to the bundle than is really worth it. But you think this is a much more likely scenario?

pi0: I agree. But this special kind of component can be used by a CMS / Project and repeated.  Code diff is alow low btw `get query: () => parseQuery`. This kind of optimization is not odd (`error.stack` for instance is also lazy even tough it is not always accessed by users)... So feel free to judge either make change or ignore my comment :p

Issue #15668, opened 2022-12-15
`useHeadSafe` composable to sanitise head metadata


belomx:
### Environment

https://codesandbox.io/s/new-haze-rmtleo?file=/app.vue

### Reproduction

Open the page and the alert will pop-up.

### Describe the bug

The useHead can be used for XSS Attack.
In my application, the user's inputs are part of the 'application/ld+json' script for SEO purposes. And a users can insert a malicious script.

### Additional context

Good to notice, the xss attack will work only if the script passed as a prop to a component.

danielroe:
This is now implemented.

PR #19302, opened 2023-02-27
feat(head): improved Unhead integration

harlan-zw:
### 🔗 Linked issue

https://github.com/unjs/unhead/issues/116, https://github.com/nuxt/nuxt/issues/15668

### ❓ Type of change

### 📚 Description

Unhead [v1.1.0](https://github.com/unjs/unhead/releases/tag/v1.1.0) was recently released.

With the dep update, I figure it's a good time to improve the integration:

## PR To Do

- [ ] benchmark v1.1  
- [ ] test treeshaking

### Features

**Dropping @vueuse/head dependency**

To avoid breaking a bunch of third-party packages, the dependency was kept on @vueuse/head, which acted as a simple compatibility wrapper for Unhead. While we still need some support for this, Unhead now exports the compatibility layer directly. 

This gives finer control over what's imported, allowing for better tree-shaking. For example, `@unhead/dom` should be treeshaken from the server build now.

**Support app.head.options**

While there are minimal build-time options available for Unhead, this is needed to enable `experimentalHashHydration` which I'm going to enable for `nuxt-unhead`. It will be enabled in the future by default but want to test it out a bit first.

**SSR: Skip app.head entry client-side**

This provides a performance improvement for SSR apps, fixes the bug https://github.com/unjs/unhead/issues/116 and allows for `experimentalHashHydration`.

```ts
  app: {
    head: {
      options: {
        experimentalHashHydration: true
      }
    }
  }
})
```

**New composables**

useServerSeoMeta - Basically what was already available but it provides a `{ mode: 'server' }` which allows for the experimental hash hydration

useHeadSafe / useServerHeadSafe - Bit verbose having both but we need the server function for the above and to treeshake it from the client build (once the vite plugin is implemented)

**Unhead Unplugin**

- Treeshakes server components
- Transforms useSeoMeta (saves ~2kb gzipd)

### Todo (subsequent PRs?)

The following items are things I'd like to solve but I don't think can exist within this PR.

**Expose Unhead Hooks**

These are pretty easy to implement but not sure if this PR is the best spot for them. They're currently not documented within the Unhead docs so should probably sort that out first.

**Document Template Params**

These are mainly used for lower-level integrations at this point.

### 📝 Checklist


packages/nuxt/src/head/runtime/plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  let head: VueHeadClient<HeadAugmentations>
+  if (process.server) {
+    head = createServerHead<HeadAugmentations>(headOptions)
+    // when SSR it's safe to only render the appHead server side
+    head.push(appHead, { mode: 'server' })
+  } else {
+    head = createHead<HeadAugmentations>(headOptions)
+    // only in SPA mode do we need to push the appHead client side
+    if (nuxtApp.ssrContext?.noSSR) {
```
harlan-zw, line 24: Is this the correct way to check for SPA?

packages/nuxt/src/imports/presets.ts
```diff
@@ -57,7 +54,14 @@ const appPreset = defineUnimportPreset({
     'prefetchComponents',
     'loadPayload',
     'preloadPayload',
-    'isPrerendered'
+    'isPrerendered',
+    // head
```
harlan-zw, line 58: Wasn't sure why we were importing from `#app`, was there a specific reason?

packages/nuxt/src/head/runtime/composables.ts
```diff
@@ -1,39 +1,57 @@
...
  */
-export const useServerSeoMeta: typeof _useSeoMeta = (meta) => {
+export const useServerSeoMeta: typeof _useServerSeoMeta = (meta) => {
   if (process.server) {
-    return _useSeoMeta(meta)
+    return _useServerSeoMeta(meta)
   }
 }
+
+export {
```
harlan-zw, line 53: We could wrap these but we don't need to augment the types so not much point.

packages/nuxt/src/head/runtime/composables.ts
```diff
@@ -1,39 +1,57 @@
...
+ * The `useServerHeadSafe` composable is identical to `useHeadSafe` except that it will
+ * have no effect (and will return nothing) if called on the client.
  *
- * It is advised to use `useServerSeoMeta` unless you _need_ client-side
- * rendering of your SEO meta tags.
+ * @see useHeadSafe
  */
-export const useSeoMeta: typeof _useSeoMeta = (meta) => {
-  return _useSeoMeta(meta)
+export const useServerHeadSafe: typeof _useServerHeadSafe = (input, options) => {
```
harlan-zw, line 34: Both `useServerHeadSafe` and `useServerSeoMeta` can be imported from @unhead/vue when the Vite plugin is implemented

danielroe:
@harlan-zw I think a fair amount of this has been merged in other PRs. But I think there's still some stuff here we would want to merge in, right?

harlan-zw:
Thanks for the ping. I'll close for now and create separate PRs where appropriate.

Issue #14131, opened 2022-06-08
add asset import typings for webpack and vite builders


cawa-93:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.3-27571095.9379606`
- Package Manager:  `npm@8.5.5`
- Builder:          `vite`
- User Config:      `telemetry`, `modules`, `trpc`, `typescript`, `runtimeConfig`, `css`
- Runtime Modules:  `trpc-nuxt@0.1.24`
- Build Modules:    `-`
------------------------------

### Reproduction

1. Create an `asset` folder in project root
2. Add any image to it.
3. Import this image from a TypeScript
	```ts
	import MyImage from 'assets/my-image.png'
	```

### Describe the bug

TS got type error:
```ts
import MyImage from 'assets/my-image.png'
// TS2307: Cannot find module 'assets/my-image.png' or its corresponding type declarations.
```

### Additional context

In the case of Vite, this error is easily corrected by adding "client" to the [`types`](https://github.com/nuxt/framework/blob/603f07c3b8f93feefb8c59c5ff3a970a60e38041/packages/nuxi/src/utils/prepare.ts#L23) in generated `.nuxt/tsconfig.json`. Something like: 
```diff
// nuxi/src/utils/prepare.ts
export const writeTypes = async (nuxt: Nuxt) => {
  const modulePaths = getModulePaths(nuxt.options.modulesDir)

  const tsConfig: TSConfig = defu(nuxt.options.typescript?.tsConfig, {
    compilerOptions: {
      jsx: 'preserve',
      target: 'ESNext',
      module: 'ESNext',
      moduleResolution: 'Node',
      skipLibCheck: true,
      strict: nuxt.options.typescript?.strict ?? false,
      allowJs: true,
      noEmit: true,
      resolveJsonModule: true,
      allowSyntheticDefaultImports: true,
-     types: ['node'],
+     types: nuxt.options.builder === '@nuxt/vite-builder' ? ["node", 'vite/client'] : ['node'],
      baseUrl: relative(nuxt.options.buildDir, nuxt.options.rootDir),
      paths: {}
    },
    include: [
      './nuxt.d.ts',
      join(relative(nuxt.options.buildDir, nuxt.options.rootDir), '**/*'),
[...]

PR #19338, opened 2023-02-28
fix(schema): update vite `import.meta` types

### LINKED ISSUE #15645: ViteImportMeta `import.meta.glob` does not change types for the `eager: true` flag
### Environment

Nuxt project info:                                                                                                                                          04:18:43

- Node Version:     `v16.12.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `yarn@1.22.10`
- Builder:          `vite`
- User Config:      `ssr`, `pages`, `app`, `css`, `plugins`, `modules`, `runtimeConfig`, `devServer`, `tailwindcss`, `postcss`, `vite`
- Runtime Modules:  `@pinia/nuxt@0.4.6`, `@nuxtjs/tailwindcss@6.1.3`, `@funken-studio/sitemap-nuxt-3@4.0.4`, `nuxt-swiper@0.1.6`
- Build Modules:    `-`

### Reproduction

I don't think a reproducer is needed here.

### Describe the bug

The `ViteImportMeta` type in the following file appears to be derived from an older version of Vite.

https://github.com/nuxt/framework/blob/a9df4b75d622b21c1c0cc6f3e601722e42714b90/packages/schema/src/types/global/vite.ts#L34

Current versions of vite have a generic that differs the return type based on whether eager is true or false.

https://github.com/vitejs/vite/blob/main/packages/vite/types/importMeta.d.ts

### Additional context

_No response_

### Logs

_No response_

clarkdo:
### 🔗 Linked issue

Fix https://github.com/nuxt/nuxt/issues/15645

### ❓ Type of change

### 📚 Description

Sync ImportMeta from Vite and try to import from Vite exported types.


packages/schema/src/types/global/vite.ts
```diff
@@ -1,35 +1,15 @@
...
-  send (event: any, data?: any): void
-}
-
-export interface ViteGlobOptions {
-  as?: string
-  eager?: boolean
-  import?: string
-  query?: string | Record<string, string | number | boolean>
-  exhaustive?: boolean
-}
```
danielroe, line 27: For now, could we keep these inlined, but updated to match the latest upstream vite types?

I would be very willing to consider directly referencing or adding vite types to the `nuxt.d.ts` build conditionally, in line with your previous PR - let's discuss at https://github.com/nuxt/nuxt/issues/14131.

clarkdo: > updated to match the latest upstream vite types?

Sorry, can you please clarify the benefits ?  This also already has vite as dev dep, so it doesn't introduce any structural change, and I think it doesn't have conflict with `directly referencing or adding vite types to the nuxt.d.ts`, we can always remove this file when it's resolved.

danielroe: The reason these are currently here and not already directly depending on vite's types is that @pi0 was reluctant to depend on vite's types because we may not support all of them, and this way we curate the types that are made available. You can see the context here: https://github.com/nuxt/framework/pull/1439.

Let's discuss a different approach in the linked issue, but for now (in this PR) we should keep the current approach but update the types.

clarkdo: Makes sense, I've updated the pr accordingly, thanks!

danielroe (approved):
Thank you. ❤️

clarkdo:
@danielroe What formatter rules are we following in nuxt now ? I don't see any prettier config, is it only editorconfig ?

PR #19339, opened 2023-02-28
Open external links in a new tab on the `2.installation.md` page

rafael-abuawad:
### ❓ Type of change

### 📚 Description

It is recommended to consider modifying the default behavior of external links in Markdown, as opening links in the same tab can be disruptive for users, especially beginners. Opening links in a new window may enhance user experience and prevent users from losing their place within the original content.

### 📝 Checklist

- [x] VS Code to _blank target
- [x] Node.js to _blank target
- [x] Volar Extension to _blank target


AloisSeckar:
I would like to [object against this](https://css-tricks.com/use-target_blank/)

danielroe:
I think the linked article makes a good case not to default to opening in a new tab.

PR #19356, opened 2023-03-01
docs: update $fetch documentation

ozum:
It is confusing why data is fetched twice when using `$fetch` on the client side without wrapping it with `useAsyncData`. It is explained [here](https://github.com/nuxt/nuxt/discussions/18731#discussioncomment-4860920) clearly. This PR updates docs to add clarification.

### 🔗 Linked issue

#18731

### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

It is confusing why `$fetch` fetches data twice when used on the client side with SSR. This PR adds details about this behaviour.

### 📝 Checklist

- [X] I have linked an issue or discussion.
- [X] I have updated the documentation accordingly.

PR #19357, opened 2023-03-01
docs: avoid `$fetch` in top-level `<script setup>`

ozum:
It is confusing why data is fetched twice when using `$fetch` on the client side without wrapping it with `useAsyncData`. It is explained [here](https://github.com/nuxt/nuxt/discussions/18731#discussioncomment-4860920) clearly. This PR updates docs to add clarification.

### 🔗 Linked issue

#18731

### ❓ Type of change

### 📚 Description

Complements PR #19356. It is confusing why $fetch fetches data twice when used on the client side with SSR. This PR adds details about this behavior.

### 📝 Checklist


docs/1.getting-started/6.data-fetching.md
```diff
@@ -352,3 +352,7 @@ Using `$fetch` has a number of benefits, including:
 It will handle 'smartly' making direct API calls if it's running on the server, or making a client-side call to your API if it's running on the client. (It can also handle calling third-party APIs.)
 
 Plus, it comes with convenience features including automatically parsing responses and stringifying data.
+
+When using server-side rendering, calling `$fetch` *on the client side* without wrapping it with `useAsyncData` causes fetching the data twice initially on the server, then again on the client.
```
danielroe, line 356:
```suggestion
When using server-side rendering, calling `$fetch`without wrapping it with `useAsyncData` will fetch the data twice - initially on the server, then again on the client.
```

ozum: Hi @danielroe,

**on the client side** part should be added again because it **only** fetches twice if it is used on the client-side. If it is used on the server side it fetches once. It is so important, I spent hours just trying to understand why `$fetch` in my code executes twice until I find the discussion.

Below is a contrived example which fetches only once using `$fetch`.
```ts
return process.server ? $fetch(...) : useLazyFetch(...);
```

Atinux: Actually the sentence correct by Daniel is correct. Also the example you describe should never exists 😬

ozum: > example you describe should never exists

@Atinux, I couldn't find a good example at the time. How about this one?

**/server/api/item.ts**
```ts
export default eventHandler(async (event: H3Event) => {
  cont item = $fetch(...);  
  return { item };
});
```

The above example does not cause double fetching. I want to separate the above case from the below case.

**/pages/hello.vue**
```ts
<script setup lang="ts">
  const data = $fetch(...); // double fetching
</script>
```

Could you please suggest a sentence which emphasize the separation between those cases?

manniL: The data will be fetched at the server side, will *not* be serialized and *not be transferred to the client side and fetched again on the client side.

ozum: To make it clearer, I rephrased the description and added examples to the docs in a new commit.

@manniL, @Atinux, or @danielroe, could you review whether the examples and descriptions are technically correct? If there are mistakes, please kindly correct them.

If they are technically correct, remember while reviewing the description, that you are part of the Nuxt team and what is very obvious to you may not be so obvious -even confusing- to the developers outside the Nuxt ecosystem.

I try to avoid ambiguity as much as possible. I spent hours why `$fetch` is fetching twice. I saw #18731, just before opening a bug issue. So I think this description is important for others.

docs/1.getting-started/6.data-fetching.md
```diff
@@ -352,3 +352,36 @@ Using `$fetch` has a number of benefits, including:
 It will handle 'smartly' making direct API calls if it's running on the server, or making a client-side call to your API if it's running on the client. (It can also handle calling third-party APIs.)
 
 Plus, it comes with convenience features including automatically parsing responses and stringifying data.
+
+When using server-side rendering, calling `$fetch` in an isomorphic code (which is executed both on the server side and on the client side) will fetch the data twice (initially on the server, then again on the client). `$fetch` will *not serialize* and *not transfer* the data to the client side and the data will be fetched again on the client side.
```
Atinux, line 356:
```suggestion
When using server-side rendering, calling `$fetch` in an isomorphic code (which is executed on both server and client side) will fetch the data twice (initially on the server and again on the client during the hydration phase). `$fetch` will *not serialize* and *not transfer* the data to the client side (via the payload) and the data will be fetched again on client side.
```

Atinux:
Much better, shall we move this into the **best-practices section**?

We should recommend everyone to **not use** `$fetch` in any `.vue` file if not wrapped with `useAsyncData` (or using `useFetch`) to avoid this double data fetching.

docs/1.getting-started/6.data-fetching.md
ozum: Excellent. Thank you.

ozum:
> We should recommend everyone to **not use** `$fetch` in any `.vue` file if not wrapped with `useAsyncData` (or using `useFetch`) to avoid this double data fetching.

Yes, that would be much better.

manniL:
> We should recommend everyone to **not use** `$fetch` in any `.vue` file if not wrapped with `useAsyncData` (or using `useFetch`) to avoid this double data fetching.

Well, there are a few exceptions, e.g. when `$fetch` is in a function to, let's say send a form.

Atinux:
Indeed, so I guess it's top-level await inside a server-rendered component?

danielroe:
I think @manniL is suggesting something like:

```vue
<script setup>
async function submitForm() {
  const data = /* */
  await $fetch('/api/form', { body: data })
}
</script>

**Comment by manniL** [comment]
> Indeed, so I guess it's top-level await inside a server-rendered component?

I think mainly top-level `$fetch` is relevant for this error, yes.

File: docs/1.getting-started/6.data-fetching.md [hunk 3]
```diff
@@ -352,3 +352,36 @@ Using `$fetch` has a number of benefits, including:
 It will handle 'smartly' making direct API calls if it's running on the server, or making a client-side call to your API if it's running on the client. (It can also handle calling third-party APIs.)
 
 Plus, it comes with convenience features including automatically parsing responses and stringifying data.
+
+When using server-side rendering, calling `$fetch` in an isomorphic code (which is executed on both server and client side) will fetch the data twice (initially on the server and again on the client during the hydration phase). `$fetch` will *not serialize* and *not transfer* the data to the client side (via the payload) and the data will be fetched again on client side.
```
danielroe, line 356:
```suggestion
When using server-side rendering, calling `$fetch` in code that is executed on both server and client side will fetch the data twice (initially on the server and again on the client during the hydration phase). `$fetch` will not serialize or transfer the data to the client side, so the data will be fetched again on client side.
```

docs/1.getting-started/6.data-fetching.md
```diff
@@ -352,3 +352,36 @@ Using `$fetch` has a number of benefits, including:
...
+```ts
+<script setup lang="ts">
+  const price = $fetch('/api/price');
+</script>
+```
+
+**/server/api/product.get.ts**: Server only code below executes `$fetch` only once at the server side.
+```ts
+export default eventHandler(async (event: H3Event) => {
+  const price = $fetch('/api/price');
```
pi0, line 371: I think we can move toward and recommend `event.$fetch` which has already some benefits such as preserving context, and also future possibilities like same request loop detects.

```suggestion
  const price = event.$fetch('/api/price');
```

ozum: @pi0, please forgive my lack of info. This is the first time I ever heard of `event.$fetch`. Even googling returns only one page about it with a single line of info [on release notes (3.2.0)](https://nuxtjs.org/releases/)

Referencing `event.$fetch` in an example without ever giving details could be confusing. Could you please add enough details about it?

pi0: You are right. It would be a nice idea to also add a section for the `event.$fetch` in server section.

ozum: Thanks @pi0. I'm also looking forward to learning what `event.$fetch` is.

pi0: You can read more in this PR: https://github.com/unjs/h3/pull/323

`event.$fetch` is bound to user request. Because of this, it can automatically proxy the context and incoming headers from SSR without need to manually pass them.

danielroe (approved):
This looks good to me. I agree we need to document `event.$fetch` but think it can be deferred to another PR.

atinux (approved):

PR #19369, opened 2023-03-01
feat(nuxi,schema): add support for setting nuxt `logLevel`

### LINKED ISSUE #18497: Quiet mode

### LINKED ISSUE #18656: CLI commands quiet mode

Baroshem:
resolves https://github.com/nuxt/nuxt/issues/18656
resolves https://github.com/nuxt/nuxt/issues/18497

This PR adds a `logLevel` option to basic nuxi commands and build that is then used to set certain log visibility for underlying builder (either `vite` or `webpack`).

Three main log levels have been defined:

1. `silent` - converted to `none` in webpack and `silent` in vite
2. `info` - converted to `normal` in webpack and `info` in vite
3. `verbose` - converted to `verbose` in webpack and `info` in vite

### ❓ Type of change

### 📝 Checklist


packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,23 @@ export default defineUntypedSchema({
...
+   * This option is then used as 'silent' in Vite and 'none' in Webpack
+   *
+   * @type {'quiet' | typeof import('../src/types/global/vite').ViteLogLevel | typeof import('../src/types/global/webpack').WebpackStats}
+   */
+  logLevel: {
+    $resolve: (val) => {
+      if (isMinimal) {
+        return 'quiet'
+      }
+      return val
```
danielroe, line 55: Probably we always want to respect an explicit value

```suggestion
      return val ?? (isMinimal ? 'quiet' : undefined)
```

packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,20 @@ export default defineUntypedSchema({
     },
   },
 
+  /**
+   * Log level when building logs.
+   *
+   * Defaults to 'quiet' when running in CI or when a TTY is not available.
+   * This option is then used as 'silent' in Vite and 'none' in Webpack
+   *
+   * @type {'quiet' | typeof import('../src/types/global/vite').ViteLogLevel | typeof import('../src/types/global/webpack').WebpackStats}
```
danielroe, line 48: I think we should create our own list of possible values rather than allowing _any_ possible vite/webpack value. We could support a subset of the vite values as that's our default setup, and map them to webpack values:

So, to start with:

- `silent` -> silent/none
- `info` -> info/normal

Baroshem: So basically to create an enum of NuxtLogs values like silent|info for now and do not accept any other?

If we do this, users would have to access vite/webpack configs directly from nuxt.config.ts. to change it to any other vite or webpack log option.

Is that correct?

danielroe: I wouldn't create an enum; just use string literals. But yes, that's right.

pi0: I love the idea for having built-in nuxt levels and then map them (otherwise we can never find single convention 😆). If i may suggest, we shall add at least a 3rd `verbose` variant for CLI compat (silent: nothing, info: default normal, vebose: all logs)

packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,20 @@ export default defineUntypedSchema({
...
+   * Log level when building logs.
+   *
+   * Defaults to 'silent' when running in CI or when a TTY is not available.
+   * This option is then used as 'silent' in Vite and 'none' in Webpack
+   *
+   * @type {'silent' | 'info'}
+   */
+  logLevel: {
+    $resolve: (val) => {
+      return val ?? (isMinimal ? 'silent' : null)
```
pi0, line 52: `isMinimal` includes when there is no TTY (basic CLIs) we still can log in there. We should use `isTest` only.

packages/schema/src/config/build.ts
danielroe: Agreed 👍

Baroshem: Not sure about verbose value 
https://vitejs.dev/config/shared-options.html#loglevel
The info is a default one
'info' | 'warn' | 'error' | 'silent'
So Nuxt log level info | verbose | silent would only work for Vite info and silent

packages/schema/src/config/build.ts
Baroshem: Changed in [a1f4f5c](https://github.com/nuxt/nuxt/pull/19369/commits/a1f4f5c24d5d6b6d9c04b258fcbd7018fad9e67c)

packages/schema/src/config/build.ts
pi0: Yes, so we basically **map** this values. In this case for vite, both info and verbose map to `info` for vite as it has nothing more.

Baroshem: I missed your message somehow @pi0. Sorry for that. Adding them right now so that we can proceed :)

packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,18 @@ export default defineUntypedSchema({
...
+  /**
+   * Log level when building logs.
+   *
+   * Defaults to 'silent' when running in CI or when a TTY is not available.
+   * This option is then used as 'silent' in Vite and 'none' in Webpack
+   *
+   * @type {'silent' | 'info' | 'verbose'}
+   */
+  logLevel: {
+    $resolve: (val) => val ?? (isTest ? 'silent' : 'info')
```
pi0, line 51: We probably need to validate val being in allowed list [silent,info, verbose] otherwise mapping logic in other places can be unexpected

danielroe (approved):
This looks solid to me! ❤️

packages/webpack/src/presets/base.ts
```diff
@@ -40,7 +42,7 @@ function baseConfig (ctx: WebpackConfigContext) {
     mode: ctx.isDev ? 'development' : 'production',
     cache: getCache(ctx),
     output: getOutput(ctx),
-    stats: 'none',
+    stats: statsMap[ctx.nuxt.options.logLevel],
```
pi0, line 45: As best practice, we might map to a default level `statMaps[logLevel] || statMaps[info]` in case of receiving an unexpected level.

packages/vite/src/vite.ts
```diff
@@ -41,6 +41,7 @@ export async function bundle (nuxt: Nuxt) {
     entry,
     config: vite.mergeConfig(
       {
+        logLevel: logLevelMap[nuxt.options.logLevel],
```
pi0, line 44: Same for here to have a default logLevel as fallback

packages/schema/src/config/build.ts
danielroe: I think with default value provided at location of map, we can probably omit the validation here and leave it to the schema to type-check for us.

pi0: It does for config i agree. But CLI is passing user args as is `--logLevel foobar` won't be validated. And it can go unnoticed.

pi0: I think both worth to be in place and making sure nuxt.options is predictable. A new log level can be introduced and easily forgotten to be added to the maps. Furthor in any external module that consumes this new option.

danielroe: The way I've typed the maps, if a new log level is added without a map entry, it will cause a nuxt build failure.

pi0: Types are not enough as external modules can depend on different schema. We need validation here. What is the downside of doing it?

danielroe: My point is not that we shouldn't validate, but that if we do, we do not need to provide a default value for the maps as we _know_ that the value validated by the schema exists in the map (because it will be a TS if it does not).

Baroshem:
Wow, @danielroe is so fast with adding improvements to my code that I just see that there were conversations resolved and commits added 😄

packages/schema/src/config/build.ts
pi0: I get your point. First, for schema validation, we need it because:

- A user can pass a wrongly/unsupported type via `nuxi dev --logLevel xyz`
- We keep it invalidated in `nuxt.options.logLevel` as `xyz`
- Nuxt internals silently handle a fallback
- Nuxt modules (3rd party) depending on `logLevel` can do unexpected behavior, trusting the type is within 3 possible values

Second, as **best practice** (as mentioned in other comment), we use a fallback. If an adhoc-module let's say supports 3 current types, and Nuxt version by user comes with a new supported type `warn`, the module fails. Again, i understand this won't happen considering we have both runtime and static type validation in place but it gives implementation consistency.

Feel free to revert first commit if you feel it would be extra even with this explanation :) Schema validation is probably most important part.

danielroe: I'm not disagreeing on validation, though I would note that your point about modules depending on the schema is really rather broader than this PR. The rest of the schema is not validated for the most part and all the comments you've just made about this _particular_ option are also relevant there too. Maybe worth adding some schema validation primitives upstream to `c12`?

pi0: It is special because is one of the first options coming directly from CLI args and they cannot be type-checked. And we do have normalization for some of other configs.

We definitely have to practice validation checks better (i guess `untyped` for auto validation).

packages/schema/src/config/build.ts
```diff
@@ -38,6 +39,23 @@ export default defineUntypedSchema({
...
+   *
+   * Defaults to 'silent' when running in CI or when a TTY is not available.
+   * This option is then used as 'silent' in Vite and 'none' in Webpack
+   *
+   * @type {'silent' | 'info' | 'verbose'}
+   */
+  logLevel: {
+    $resolve: (val) => {
+      if (val && !['silent', 'info', 'verbose'].includes(val)) {
+        console.error('Invalid `logLevel` option. Must be one of: `silent`, `info`, `verbose`.')
```
pi0, line 53: (we could also include `val` in message for better DX)

```suggestion
        console.warn('Invalid `logLevel` option. Must be one of: `silent`, `info`, `verbose`.')
```

packages/nuxi/src/commands/analyze.ts
```diff
@@ -11,7 +11,7 @@ import { defineNuxtCommand } from './index'
 export default defineNuxtCommand({
   meta: {
     name: 'analyze',
-    usage: 'npx nuxi analyze [rootDir]',
+    usage: 'npx nuxi analyze [--log-level, -l] [rootDir]',
```
pi0, line 14: On casing, we probably support case translation from citty, but i probably choose normalization to camel case matching JS programmatic API when calling it. Also on shortcut it could be useful to reserve it for future purposes such as `--listen, -l` in some subdommands.

danielroe: Existing options are all lower case and in some cases kebab case (e.g. `ssl-key`, `ssl-cert`). I think we should keep this kebab-case for consistency until `citty` supports camel-case normalisation.

Do you think we should require long form `--log-level` to avoid shadowing potential `--listen` shortcut? Very happy to remove `args.l` support...

pi0: For mapping, i meant mapping to nuxt config btw. But still should be supported with both conventions so that's okay.

Would be nice to remove shortcut indeed 👍🏼 (we could introduce `--verbose` / `--silent` later since are kinda de-facto convention for CLI logging controls)

pi0 (approved):
lgtm

PR #19371, opened 2023-03-01
feat(cli): programmatically pass nuxt config overrides (to dev)

### LINKED ISSUE #19347: Run dev server programmatically
### Describe the feature

It currently doesn't seem to be possible to run the dev server programmatically since it is tightly coupled to the console command. Or did I miss something?

If so, imho this is a step back from Nuxt 2 where you could run everything programmatically.

### Additional information

### Final checks

dword-design:
### 🔗 Linked issue

#19347

### ❓ Type of change

### 📚 Description

Allows to pass additional options to `runCommand`, so a Nuxt config can be passed programmatically. Resolves #19347. This isn't really a clean solution. Clean would be to have a function with the logic and the command consumes this function. The logic is also exported and someone can use it instead of the command.

### 📝 Checklist

- [x] I have updated the documentation accordingly. (there is no doc)


pi0:
Generally love the idea (tbh always been thinking to have it). Only there is a concern about `--config <path to custom config file>` (mapping to `configFile` from c12) vs `--config <config_overrides_object>`. 

I think we should use `--config` for passing path to config file.

For object passing, we might use `NUXT_CUSTOM_CONFIG` env variable revived by `destr` as it also allows passing custom config from different programmatic conditions with more flexibility.

dword-design:
@pi0 I'm not sure if passing the config via environment variables alone will be sufficient since a config can contain functions (e.g. a module imported and passed directly). I think there is no way around passing the config as a JS object directly.

pi0:
Fair enough. Only since `--config` should be reserved for config path, i think we can workaround it slightly differently by passing custom options as second argument to the sub-commands ([here](https://github.com/nuxt/nuxt/blob/6265012033bf38c107ecbdffcc8031c8b76d2bf9/packages/nuxi/src/run.ts#L13)) unblocking programmatic usage.

packages/nuxi/src/run.ts
```diff
@@ -2,9 +2,10 @@ import mri from 'mri'
 import type { Command, NuxtCommand } from './commands'
 import { commands } from './commands'
 
-export async function runCommand (command: string, argv = process.argv.slice(2)) {
+export async function runCommand (command: string, argv = process.argv.slice(2), options: any = {}) {
   const args = mri(argv)
   args.clear = false // used by dev
+  args.config = options.config
```
pi0, line 8: This should not be passed from to the `args.config` but instead as second argument of `invoke()`

packages/nuxi/src/commands/dev.ts
```diff
@@ -96,7 +97,8 @@ export default defineNuxtCommand({
           dev: true,
           ready: false,
           overrides: {
-            logLevel: args['log-level']
+            logLevel: args['log-level'],
+            ...(options.config || {})
```
pi0, line 101: `config` looks good to me but also wondering if `overrides` would give more consistency for options naming (and also more clarification this is strictly overriding any level of config)

```suggestion
            ...(options.overrides || {})
```

pi0 (approved):

PR #19309, opened 2023-02-27
fix(nuxt): preserve explicit `rel` attribute on internal link

igayamaguchi:
### 🔗 Linked issue

fix #19299

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Add `rel` attribute even if internal links is set.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #14897, opened 2022-09-14
NuxtLink with custom attribute causes vue warning in RC10


Andersso:
### Environment

Nuxi 3.0.0-rc.10
RootDir: /sandbox
Nuxt project info: 

------------------------------
- Operating System: `Linux`
- Node Version:     `v14.19.3`
- Nuxt Version:     `3.0.0-rc.10`
- Nitro Version:    `0.5.3`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://codesandbox.io/s/jolly-sky-uz9ovm?file=/pages/index.vue

Here's a snippet of the juicy bit:
```
<NuxtLink v-slot="{ href }" to="/foo" custom>
  <a :href="href">Here's a link</a>
</NuxtLink>
```

### Describe the bug

When using a NuxtLink component with custom attribute, a warning is logged.

### Additional context

In RC 8 it was working fine, my suspicion is that the new link prefetching feature is to blame.

Suspicious line of code: https://github.com/nuxt/framework/blob/v3.0.0-rc.10/packages/nuxt/src/app/components/nuxt-link.ts#L229

Maybe worth to mention, a workaround is to use vue router's RouterLink component directly instead of NuxtLink

### Logs

```shell
[Vue warn]: Extraneous non-props attributes (class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.
```

PR #19379, opened 2023-03-01
fix(nuxt): prevent fallthrough attributes on custom `NuxtLink`

### LINKED ISSUE #19375: <NuxtLink custom ...> throws a warning in Nuxt 3.2.3 due to explicit rel attribute
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.7.0`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Package Manager:  `pnpm@7.28.0`
- Builder:          `vite`
- User Config:      `telemetry`, `modules`, `vite`, `runtimeConfig`, `css`, `app`, `content`, `i18n`
- Runtime Modules:  `@nuxtjs/i18n@8.0.0-beta.10`, `@nuxt/content@2.4.3`
- Build Modules:    `-`

### Reproduction

https://github.com/nuxt/nuxt/pull/19309#issuecomment-1448885808

> ```
> 	<NuxtLink custom v-slot="{ href, navigate }" :to="targetRoute">
> 		<component :is="tag" :href="href" @click="navigate" role="link" class="domain"
> 			><slot>{{ displayDomain(domain, status) }}</slot></component>
> 	</NuxtLink>
> ```

### Describe the bug

The explicit `rel` attr/prop introduced in Nuxt 3.2.3 (https://github.com/nuxt/nuxt/pull/19309) is incompatible with `NuxtLink custom` (to use a tag other than `<a>`).

### Additional context

_No response_

### Logs

_No response_

lihbr:
### 🔗 Linked issue

Resolves: #19375

### ❓ Type of change

### 📚 Description

In #19309 we enabled the forwarding of the `rel` prop of `NuxtLink` as an attribute to internal links.

This resurfaced an issue we faced in the past (#14897) where the `custom` API of `NuxtLink` (from `RouterLink`) cannot handle such fallthrough attributes[^1]. This seems to happen because Vue slot API cannot assume that the slot content strictly is a VNode (as it could also be a fragment or text node), so it has to consider all cases.[^2]

I think this is not really an upstream issue (`vue-router` actually tries to let those attribute fallthrough as it's Vue default behavior, causing the warning) but a limitation of the slot API (again, it cannot just assume it will be a VNode).

Given those considerations we (mainly) have two choices:
1. Forcing users to forward fallthrough attributes manually using `v-bind="$attrs"` systematically when using the `custom` API
2. Not forwarding any additional attribute to the component when the `custom` API is used

I think n°2 makes more sense. Should someone use the `custom` API, it makes more sense for them to just apply those extra attributes on their slot content themselves rather than providing them to the link component which might eventually apply them to their slotted template. This comes with the tradeoff of not being able to programmatically define attributes to be applied on our end (like the prefetching class)[^3]

---

This PR also refactors the way `RouterLink` props are defined to be more readable (otherwise we would have multiple inline ternaries when defining the object).

The type of `routerLinkProps` is `Record<string, any>`, defined after Vue's internal `RawProps` type which is not exported by Vue (with reasons I assume)

### 📝 Checklist

[^1]: Vue documentation on fallthrough attributes behavior: https://vuejs.org/guide/components/attrs.html
[^2]: For reference, this also happens with `RouterLink` directly: https://stackblitz.com/edit/github-stiici-ukvew6
[^3]: `vue-router` seems to be doing the same tradeoff, albeit booleans used to define active classes and such are exposed to the slot: https://github.com/vuejs/router/blob/6b7b25a74e89f1dde534bf86d19cfcdbce25092b/packages/router/src/RouterLink.ts#L214-L230

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -211,20 +211,29 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
       return () => {
         if (!isExternal.value) {
+          const routerLinkProps: Record<string, any> = {
+            ref: process.server ? undefined : (ref: any) => { el!.value = ref?.$el },
+            to: to.value,
+            activeClass: props.activeClass || options.activeClass,
+            exactActiveClass: props.exactActiveClass || options.exactActiveClass,
+            replace: props.replace,
+            ariaCurrentValue: props.ariaCurrentValue,
+            custom: props.custom
+          }
+
+          // `custom` API cannot support fallthrough attributes as the slot
+          // may render fragment or text root nodes (#14897, #19375)
+          if (!props.custom) {
+            if (prefetched.value) {
+              routerLinkProps.class = props.prefetchedClass || options.prefetchedClass
+            }
+            routerLinkProps.rel = props.rel
+          }
+
           // Internal link
           return h(
             resolveComponent('RouterLink'),
-            {
-              ref: process.server ? undefined : (ref: any) => { el!.value = ref?.$el },
-              to: to.value,
-              ...((prefetched.value && !props.custom) ? { class: props.prefetchedClass || options.prefetchedClass } : {}),
-              activeClass: props.activeClass || options.activeClass,
-              exactActiveClass: props.exactActiveClass || options.exactActiveClass,
-              replace: props.replace,
-              ariaCurrentValue: props.ariaCurrentValue,
-              custom: props.custom,
-              rel: props.rel
-            },
+            routerLinkProps,
             slots.default
           )
         }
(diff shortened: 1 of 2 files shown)


TheAlexLichter (approved):

TheAlexLichter (approved):
I agree, option 2 seems the most reasonable here giving the user full control

danielroe (approved):

nathanchase:
The workaround of adding `v-bind="$attrs" mentioned [here](https://github.com/nuxt/nuxt/issues/19375#issuecomment-1450727999) is the only thing currently preventing the `Extraneous non-props attributes (rel) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.` errors in Nuxt 3.2.3.

manniL:
@nathanchase This should be resolved on the edge channel

Issue #12369, opened 2023-01-04
Ability to add your own keyed functions 


cawa-93:
### Describe the feature

Several composable functions ( `useState`, `useFetch`, `useAsyncData`, `useLazyAsyncData`, `useLazyFetch`) provide auto-key generation that is unique to the file name and line number of the instance of function call. However, this makes it impossible to create your own wrappers around these functions, because then the auto-generated key will always be the same for any wrapper-function call.

```ts
// composables/useAsyncSpecifiedData.ts
export function useAsyncSpecifiedData(params) {
	// ... do some work
	const handler = createHandler( params )
	const options = createOptions( params )
	return useAsyncData(handler, options) // <- the unique key always generated exactly for this file+line
}
```
I would like to have oportynity to somehow expand keyed functions with myown
https://github.com/nuxt/framework/blob/51bc6e44fef6e4a249946ba910a497089b089a64/packages/vite/src/plugins/composable-keys.ts#L15-L17

So that nuxt will generates a unique key directly for the wrapper function call, which will be passed down in `useAsyncData`. Since it is work for `useLazyAsyncData`
https://github.com/nuxt/framework/blob/51bc6e44fef6e4a249946ba910a497089b089a64/packages/nuxt/src/app/composables/asyncData.ts#L263-L274

### Additional information

- [x] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

Baroshem:
I think wrapping built in functionalities is not a good practise to be honest. 

I am closing this issue as there was no further attention in over a month. 

Feel free to reopen it though if needed :)

pi0:
Related: https://github.com/nuxt/nuxt/issues/14736

danielroe:
I think this would be a nice feature, and we'd welcome contribution on this. Maybe we could also consider configuration for other magic names, as in unctx transform or pure function annotation plugins.

Thoughts welcome on where the configuration should live.

PR #19383, opened 2023-03-02
feat(nuxt): allow configuring treeshakeable composables

harlan-zw:
### 🔗 Linked issue

Not an issue but to unblock me on https://github.com/nuxt/nuxt/pull/19302

### ❓ Type of change

### Background

Isolating composables to a specific environment is a useful way to optimise bundles. There are many instances of code that only need to run on a single environment that are introduced through modules.

However, there isn't an easy way for modules to opt-in to this treeshaking config.

For example:
- Unhead - `useServerHead`, `useServerSeoMeta`, `useServerHeadSafe`
- nuxt-schema-org - `useSchemaOrg`
- nuxt-og-image - `defineOgImageStatic`, `defineOgImageDynamic`, `defineOgImageScreenshot`
- nuxt-simple-robots - `defineRobotMeta`
- ...etc

There is currently a way to handle this by wrapping code in env if blocks. This works okay, but it means the function arguments won't be tree-shaken (afaik). 

You can wrap the composable call itself but this gets pretty verbose quickly and is not something module authors have control over.

```ts
export function useServerLog(input: any) {
  if (process.server) {
    console.log(input)
  }
}
```

```vue
<template lang="ts" setup>
// this is kept in the client build?
useServerLog({
  foo: 'bar'
  bar: 'foo',
  baz: 'bing',
  bong: 'bar',
})
// verbose!
if (process.server) {
  useServerLog({ /* */ })
} 
</template>
````

### Solution

Make the treeShake config extendable by moving it to nuxt config.

### 📝 Checklist

### CODE REVIEW

**Comment by antfu** [comment]
Interesting idea! However, I have a few questions:

- Should we allow both `.server` and `.client` functions to explore the same name to have different functions in different env? If so, it might need some changes on the unimport side.
- `Note: This will only work with composables when you aren't using their return value.` - This pitfall feels a bit too much and may lead to confusion. And honestly might not be that useful as in many cases the return value is important. It was working for Vue because we are tree-shaking mostly the hooks.
- The side-effects could also lead to confusion, as from a higher view, users would see the imports in `.server` being executed on the client side. It can be counterintuitive.

In that sense, I think I would push back for the current implementation as I see the trade-off is not worth it. People better use `onMounted` or `if (process.server)` or dynamic import to explicitly control the boundary at this moment.

**Comment by harlan-zw** [comment]
> Interesting idea! However, I have a few questions:
> 
> * Should we allow both `.server` and `.client` functions to explore the same name to have different functions in different env? If so, it might need some changes on the unimport side. (could also lead to problem of function signatures not matched)
> * `Note: This will only work with composables when you aren't using their return value.` - This pitfall feels a bit too much and may lead to confusion. And honestly might not be that useful as in many cases the return value is important. It was working for Vue because we are tree-shaking mostly the hooks.
> * The side-effects could also lead to confusion, as from a higher view, users would see the imports in `.server` being executed on the client side. It can be counterintuitive.
> 
> In that sense, I think I would push back for the current implementation as I see the trade-off is not worth it. People better use `onMounted` or `if (process.server)` or dynamic import to explicitly control the boundary at this moment.

Thanks for the detailed review! Good points

I agree that it could be a useful feature but the limitations open the scope for too many issues. I'd like to see it in Nuxt one day but sounds like some work needs to go into Unimport first.

I've minimised the scope of this PR to only make the treeShake config extendable which I think is a bit more reasonable.

File: test/fixtures/basic/composables/tree-shake.ts [hunk 1]
```diff
@@ -0,0 +1,18 @@
+export function useServerOnlyComposable () {
+  // @ts-ignore
+  if (process.client) {
+    throw new Error('this should not be called in the browser')
+  }
+}
+
+export function useClientOnlyComposable () {
+  // need to do some code that fails in node but not in the browser
+  // @ts-ignore
```
danielroe, line 10:
```suggestion
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -85,16 +86,27 @@ async function initNuxt (nuxt: Nuxt) {
...
 
-    // Add tree-shaking optimisations for SSR - build time only
-    addVitePlugin(TreeShakePlugin.vite({ sourcemap: nuxt.options.sourcemap.server, treeShake: removeFromServer }), { client: false })
-    addVitePlugin(TreeShakePlugin.vite({ sourcemap: nuxt.options.sourcemap.client, treeShake: removeFromClient }), { server: false })
-    addWebpackPlugin(TreeShakePlugin.webpack({ sourcemap: nuxt.options.sourcemap.server, treeShake: removeFromServer }), { client: false })
-    addWebpackPlugin(TreeShakePlugin.webpack({ sourcemap: nuxt.options.sourcemap.client, treeShake: removeFromClient }), { server: false })
+  addVitePlugin(TreeShakePlugin.vite(treeShakeOptions.server), { client: false })
+  addVitePlugin(TreeShakePlugin.vite(treeShakeOptions.client), { server: false })
+  addWebpackPlugin(TreeShakePlugin.webpack(treeShakeOptions.server), { client: false })
+  addWebpackPlugin(TreeShakePlugin.webpack(treeShakeOptions.client), { server: false })
```
danielroe, line 107: I think we could wait to add these plugins until `modules:done` if that would be helpful. And then the plugin could generate the matcher RegExp when initialised rather than needing a `generateTreeShakeConfig` function.

packages/schema/src/config/build.ts
```diff
@@ -59,6 +59,27 @@ export default defineUntypedSchema({
...
+        $resolve: async (val, get) => (await get('dev') ? [] : [
+          'onBeforeMount', 'onMounted', 'onBeforeUpdate', 'onRenderTracked', 'onRenderTriggered', 'onActivated', 'onDeactivated', 'onBeforeUnmount'
+        ]).concat(val).filter(Boolean)
+      },
+      client: {
+        $resolve: async (val, get) => (await get('dev') ? [] : [
+          'onServerPrefetch', 'onRenderTracked', 'onRenderTriggered'
+        ]).concat(val).filter(Boolean)
+      }
+    },
```
danielroe, line 81: I wonder if it's possible to hard code the unhead composables we need without exposing this in the build config, at least as-of yet....

harlan-zw: This is definitely possible for the Unhead composables as they are already core to nuxt, the issue is that I need this for some other modules:
- nuxt-schema-org - `useSchemaOrg`
- nuxt-og-image - `defineOgImageStatic`, `defineOgImageDynamic`, `defineOgImageScreenshot`
- nuxt-simple-robots - `defineRobotMeta`
- ...probably some others

If you have any ideas on how to solve this without exposing config then happy to explore it

test/fixtures/basic/pages/tree-shake.vue
```diff
@@ -0,0 +1,13 @@
+<script lang="ts" setup>
+// server only
+useServerOnlyComposable()
+// client only
+useClientOnlyComposable()
+// can only run client side, should be tree shaken from server build
+setTitleToPink()
```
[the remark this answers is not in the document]
harlan-zw: This composable is actually validating that the code _is_ running inside the composable, whereas the other two are just ensuring that it's _not_ running.

Minor distinctions but thought it was worth covering both

packages/nuxt/src/core/nuxt.ts
harlan-zw: Agreed, that's updated

packages/schema/src/config/build.ts
danielroe: We thought a new top level `optimization` option would be good. (We can include it in the same `build.ts` schema file.) It can include both this config (`optimization.treeshake.composables.client` & `.server`) and also future config like https://github.com/nuxt/nuxt/issues/12369.

packages/schema/src/config/build.ts
```diff
@@ -108,5 +108,37 @@ export default defineUntypedSchema({
...
+        server: {
+          $resolve: async (val, get) => (await get('dev') ? [] : [
+            'onBeforeMount', 'onMounted', 'onBeforeUpdate', 'onRenderTracked', 'onRenderTriggered', 'onActivated', 'onDeactivated', 'onBeforeUnmount'
+          ]).concat(val).filter(Boolean)
+        },
+        client: {
+          $resolve: async (val, get) => (await get('dev') ? [] : [
+            'onServerPrefetch', 'onRenderTracked', 'onRenderTriggered'
+          ]).concat(val).filter(Boolean)
+        }
```
danielroe, line 139: while we're defining the API, would it make sense to have an array of objects with name + exporter (e.g. these would all be from 'vue')? That way if we later improve the tree shaking algorithm to remove only ones from a certain import-path (or even integrate with unimport), then we can use a finer-grained hammer.

harlan-zw: Another great idea! Have implemented, let me know if it needs any further changes

danielroe (approved):
Looks great! ❤️

danielroe:
Merging so @harlan-zw can go ahead with linked PR. But would particularly appreciate feedback @antfu on the schema change in this PR. If we do end up supporting something like this in unimport down the line, does it give you enough info? (it's in the preset format, so `Record<packageName: string, imports: string[]>`).

(Happy to change the schema before we release v3.3 if you have any suggestions for improvement.)

Issue #12350, opened 2023-01-17
Restart required to get new Nuxt Module component into application's components.d.ts


mklueh:

madebyfabian:
Hey there! Please add a proper [minimal reproduction](https://nuxt.com/docs/community/reporting-bugs) so we can look into that properly, thank you 🙏🏻💚.

Since there isn't any activity here for a longer time, I'd suggest closing this, but feel free to reopen. Thank you 🙏🏻💚

Issue #12339, opened 2023-01-15
export `@nuxt/kit` utils from `nuxt/kit` for direct use


HigherOrderLogic:
### Describe the feature

When implementing modules locally within a project, developers would have to explicitly install `@nuxt/kit` package for things like `defineNuxtModule`, `createResolver`,... (of course you can use it without installing because it's `nuxt`'s direct dependency, but using something not listed in `package.json` probably isn't a good idea). So my proposal is to export it in `nuxt` package,  ex: `nuxt/kit`.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

Baroshem:
I think having it exported is a good idea as well :) 

I can definitely find use cases for local modules like that.

Strift:
I agree with @HigherOrderLogic and @Baroshem.

In Nuxt 3, modules seems like the better fitted solution for encapsulating logic to handle external services. 

It feels weird to me that I need to include a package that seems designed for publishing modules (`@nuxt/kit`), just to be able to create local modules. 

Plugins utilities like `defineNuxtPlugin` are currently exported, and it's useful for everyone, not just package maintainers. I think we could have a similar approach for modules.

Now, it's totally possible that I'm misusing modules. If that's the case I'd love to know alternative solutions. :)

PR #19394, opened 2023-03-02
feat(nuxt): auto-register modules in `~/modules`

danielroe:
### ❓ Type of change

### 📚 Description

This adds support for scanning `~/modules` and automatically registering any local modules there. The directory is configurable via `dir.modules` and supports local modules in layers.

It also adds support for automatically restarting Nuxt when these local module files are updated, resolving https://github.com/nuxt/nuxt/issues/12350. Note that at the moment, they are cached and a soft restart does not load an updated module - we should likely investigate the cause of this so we do not need a hard restart in this case.

### 📝 Checklist


docs/2.guide/2.directory-structure/1.modules.md
```diff
@@ -0,0 +1,18 @@
+---
+navigation.icon: IconDirectory
+title: 'modules'
+head.title: 'modules/'
+description: Use the modules/ directory to automatically register local modules within your application.
+---
+
+# modules Directory
```
Atinux, line 8:
```suggestion
# Modules Directory
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -127,6 +127,12 @@ async function initNuxt (nuxt: Nuxt) {
     ...nuxt.options._modules
   ]
 
+  // Automatically register user modules
+  for (const config of nuxt.options._layers.map(layer => layer.config)) {
+    const userModules = await resolveFiles(config.srcDir, `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`)
+    modulesToInstall.push(...userModules)
```
pi0, line 133: - We need to run these modules **before** ad-hoc modules (_modules). Probably after user modules. 
- Can you check if there are sorted a-z? A explicit sort by name would be nice. 
- We might need to duplicate them too if explicitly added by user for custom order

pi0 (commented):
Nice idea! As a complement we might also introduce [`nuxt/kit`](https://github.com/nuxt/nuxt/issues/12339) since users have only nuxt dependency.

packages/nuxt/src/core/nuxt.ts
danielroe: yes, they are sorted.
and they will be de-duplicated on run

pi0 (approved):

TheAlexLichter (approved):

Baroshem (approved):
Awesome feature idea!

I can easily see using it in my projects :)

antfu (approved):

Atinux:
I would like to update the behaviour to scanning `~/modules/<module-name>/module.ts` so user can have the same structure as https://github.com/nuxt/starter/tree/module/src 

This way it's easier to extract when then want to publish their package to the Nuxt community

Atinux:
Well actually I would prefer`~/modules/<module-name>/index.ts` which makes the most sense.

Why did we decide to go with `module.ts` on https://github.com/nuxt/starter/tree/module/src ?

pi0:
For module builder package template, top-level `index` is reserved for util exports.

BTW for local scanning, `{name}/index.ts` seems a good idea.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -126,7 +126,10 @@ async function initNuxt (nuxt: Nuxt) {
 
   // Automatically register user modules
   for (const config of nuxt.options._layers.map(layer => layer.config)) {
-    const userModules = await resolveFiles(config.srcDir, `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`)
+    const userModules = await resolveFiles(config.srcDir, [
+      `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`,
+      `${config.dir?.modules || 'modules'}/*/{index,module}{${nuxt.options.extensions.join(',')}}`
```
pi0, line 131: Thinking more, one other possible issue would be when there are some utils in modules with index, they might be wrongly scanned. 

@Atinux Are you good if we support `modules/*` + `modules/[name]/module.ts` for starting? This at least is compatible with current module structure recommendations.

pi0:
```suggestion
      `${config.dir?.modules || 'modules'}/*/module.{${nuxt.options.extensions.join(',')}}`
```

danielroe: I don't think it has the same constraints because you won't be importing from the module directly, I imagine?

Atinux:
I will try to keep it simple to `modules/{name}/index.ts` then 😊 

Explaining to rename to `module.ts` when packaging in order to let them export utils makes perfect sense.

I guess we could also support top-level file for very simple modules?

packages/nuxt/src/core/nuxt.ts
pi0: Indeed, only making it more future proof in case that we came with something (also if plan is to market it as build-your-module-then-extract, this has less change steps for converting it into a package)

pi0:
I'm okay with either convention in local dir. Each has cons and pros.

packages/nuxt/src/core/nuxt.ts
Atinux: It is a hard decision, but I will avoid having both `module` + `index` and decide with one.

For local module, I like the idea of `index.ts` because it implies that you are loading the directory entry.

And anyway we have the module starter to better explain why we choose `module.ts`. I think we should add an `index.ts` there to explain that we export utils.

Pushing an update to current PR for docs, `modules/*/index.ts` makes the more sense when explaining

packages/nuxt/src/core/nuxt.ts
```diff
@@ -124,18 +124,28 @@ async function initNuxt (nuxt: Nuxt) {
...
+      modulesToInstall.push(mod)
+      watchedPaths.add(relative(config.srcDir, mod))
+    }
   }
 
   // Register ad-hoc modules
   modulesToInstall.push(...nuxt.options._modules)
 
+  nuxt.hook('builder:watch', (event, path) => {
+    // TODO: fix cache issue so we do not need a hard restart
```
pi0, line 145: This is ultimately not doable with ESM native cache. You can remove todo :D

packages/nuxt/src/core/nuxt.ts
```diff
@@ -124,18 +124,28 @@ async function initNuxt (nuxt: Nuxt) {
...
+    for (const mod of userModules) {
+      modulesToInstall.push(mod)
+      watchedPaths.add(relative(config.srcDir, mod))
+    }
   }
 
   // Register ad-hoc modules
   modulesToInstall.push(...nuxt.options._modules)
 
+  nuxt.hook('builder:watch', (event, path) => {
```
pi0, line 144: Since this is a hard restart request, we might use `hookOnce` to avoid doing multiple requests:

```suggestion
  nuxt.hookOnce('builder:watch', (event, path) => {
```

packages/nuxt/src/core/nuxt.ts
pi0:
```suggestion
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -124,18 +124,28 @@ async function initNuxt (nuxt: Nuxt) {
   await nuxt.callHook('modules:before')
   const modulesToInstall = [...nuxt.options.modules]
 
+  const watchedPaths = new Set<string>()
+
   // Automatically register user modules
   for (const config of nuxt.options._layers.map(layer => layer.config)) {
```
pi0, line 130: We probably want to reverse the other of layers. A module in base module should be ready by the time user (higher order module) is executed

danielroe: If that's the case, wouldn't we also expect that those modules have been run before explicitly specified modules in nuxt.config (e.g. unshift these modules)?

docs/2.guide/2.directory-structure/1.modules.md
```diff
@@ -0,0 +1,46 @@
...
+
+The auto-registered files patterns are:
+- `modules/*/index.ts`
+- `modules/*.ts`
+
+You don't need to add those local modules to your [`nuxt.config.ts`](/docs/guide/directory-structure/nuxt.config) separately.
+
+::code-group
+```ts [modules/hello/index.ts]
+import { createResolver, defineNuxtModule, addServerHandler } from '@nuxt/kit'
```
danielroe, line 20:
```suggestion
import { createResolver, defineNuxtModule, addServerHandler } from 'nuxt/kit'
```

test/fixtures/basic/modules/auto-registered/index.ts
```diff
@@ -0,0 +1,15 @@
+import { createResolver, defineNuxtModule, addServerHandler } from '@nuxt/kit'
```
danielroe, line 1:
```suggestion
import { createResolver, defineNuxtModule, addServerHandler } from 'nuxt/kit'
```

test/fixtures/basic/modules/example.ts
```diff
@@ -1,5 +1,4 @@
-import { fileURLToPath } from 'node:url'
-import { defineNuxtModule, addPlugin, useNuxt } from '@nuxt/kit'
+import { defineNuxtModule, createResolver, addPlugin, useNuxt } from '@nuxt/kit'
```
danielroe, line 1:
```suggestion
import { defineNuxtModule, createResolver, addPlugin, useNuxt } from 'nuxt/kit'
```

packages/nuxt/src/core/nuxt.ts
pi0: I think that also make sense! It should be last layer > ... > first layer > user > ad-hoc

danielroe: sounds good

packages/nuxt/src/core/nuxt.ts
```diff
@@ -122,10 +122,28 @@ async function initNuxt (nuxt: Nuxt) {
...
-    ...nuxt.options.modules,
-    ...nuxt.options._modules
-  ]
+  const modulesToInstall = []
+
+  const watchedPaths = new Set<string>()
+
+  // Automatically register user modules
+  for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+    const userModules = await resolveFiles(config.srcDir, [
```
pi0, line 131: refactor We could rename this to `localModules` to avoid confusing with user specified modules from config

```suggestion
    const localModules = await resolveFiles(config.srcDir, [
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -122,10 +122,28 @@ async function initNuxt (nuxt: Nuxt) {
...
+      `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`,
+      `${config.dir?.modules || 'modules'}/*/index{${nuxt.options.extensions.join(',')}}`
+    ])
+    for (const mod of userModules) {
```
pi0, line 135:
```suggestion
    for (const mod of localModules) {
```

packages/nuxt/src/core/nuxt.ts
danielroe: renamed to `layerModules` 👍

HigherOrderLogic:
I'm against this idea. Although auto registering modules inside `~/modules` folder is convenient, not everytime the developer would want to load the modules in alphabetical order.

pi0:
>  not everytime the developer would want to load the modules in alphabetical order.

I fully agree, and because of that, when you manually specify them in modules[] in configuration, they will be kept in your custom order.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -122,10 +122,28 @@ async function initNuxt (nuxt: Nuxt) {
...
+  const watchedPaths = new Set<string>()
+
+  // Automatically register user modules
+  for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+    const layerModules = await resolveFiles(config.srcDir, [
+      `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`,
+      `${config.dir?.modules || 'modules'}/*/index{${nuxt.options.extensions.join(',')}}`
+    ])
+    for (const mod of layerModules) {
+      modulesToInstall.push(mod)
```
pi0, line 136: I belive we still need to deduplicate and avoid pushing if already exist in either `modulesToInstall` or `nuxt.options.modules` them here before pushing this to the module installer to try and deduplicate.

danielroe: yes, we _now_ do, because we are adding the explicit modules last.

danielroe: pushed a fix - wdyt?

pi0: Looks good!

pi0 (approved):

Atinux:
@HigherOrderLogic well you can do:
```bash
modules/
  1.c-module/
  2.d-module/
```

Similar to what we have in [Nuxt Content](https://content.nuxtjs.org/guide/writing/content-directory#ordering) for ordering links in the generated navigation.

Atinux:
Might be worth adding to the docs?

jawomg:
> ### 🔗ปัญหาที่เชื่อมโยง
> ### ❓ประเภทของการเปลี่ยนแปลง
> * [ ]  📖เอกสาร (อัปเดตเอกสารหรือ readme)
> * [ ]  🐞แก้ไขจุดบกพร่อง (การเปลี่ยนแปลงที่ไม่สิ้นสุดซึ่งช่วยแก้ไขปัญหา)
> * [x]  👌การปรับปรุง (ปรับปรุงการทำงานที่มีอยู่เช่นประสิทธิภาพ)
> * [ ]  ✨คุณลักษณะใหม่ (การเปลี่ยนแปลงที่ไม่สิ้นสุดที่เพิ่มฟังก์ชันการทำงาน)
> * [ ]  ⚠️การเปลี่ยนแปลงที่ไม่สมบูรณ์ (การแก้ไขหรือฟีเจอร์ที่อาจทำให้การทำงานที่มีอยู่เปลี่ยนไป)
> 
> ### 📚คำอธิบาย
> ซึ่งจะเพิ่มการรองรับการสแกน`~/modules`และลงทะเบียนโมดูลภายในโดยอัตโนมัติที่นั่น ไดเร็กทอรีสามารถกำหนดค่าผ่าน`dir.modules`และรองรับโมดูลโลคัลในเลเยอร์
> 
> นอกจากนี้ยังเพิ่มการสนับสนุนสำหรับการรีสตาร์ท Nuxt โดยอัตโนมัติเมื่อไฟล์โมดูลในเครื่องเหล่านี้ได้รับการอัปเดต แก้ไข[#](https://github.com/nuxt/nuxt/issues/12350) 12350 โปรดทราบว่าในขณะนี้ โมดูลเหล่านี้จะถูกแคชและการรีสตาร์ทแบบซอฟต์ไม่โหลดโมดูลที่อัปเดต เราควรตรวจสอบสาเหตุของสิ่งนี้ ดังนั้นเราจึงไม่จำเป็นต้องรีสตาร์ทฮาร์ดในกรณีนี้
> 
> ### 📝รายการตรวจสอบ
> * [ ] ฉันได้เชื่อมโยงปัญหาหรือการสนทนา
> * [x] ฉันได้อัปเดตเอกสารตามนั้น

summer-0609:
How to ignore local modules that I don't want to be automatically registered

ineshbose:
Regarding ordering, from the PR I understand that local modules load after `nuxt.options.modules` (maybe need to mention that in the docs) - is there a possibility to have one before (like the first loaded module possibly)? Not sure if adding the local module path to `nuxt.options.modules` could double-load it (if in `dir.modules`).

> How to ignore local modules that I don't want to be automatically registered

We can also add functionality for prepending filename with `.` to ignore through a PR you think? 🙂

> actually I would prefer~/modules/<module-name>/index.ts which makes the most sense.

The thing (for me at least) was to have consistent file-scanning in Nuxt (like #18418) because when the behaviour is different, it tends to confuse (me at least); though this PR has been long-released now. 😅 

(Sorry for continuing the thread!)

danielroe:
I think we already respect current `ignore` option when scanning modules.

And if local modules are added to `nuxt.options.modules` then they will run at that point and not be rerun later on.

PR #19403, opened 2023-03-02
feat(nuxt): allow accessing `NuxtPage` ref via `pageRef`

### LINKED ISSUE #19268: DefineExpose not working for page components. Transition property causing issues with console log
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.14.1`
- Nuxt Version:     `3.2.0`
- Nitro Version:    `2.2.1`
- Package Manager:  `npm@9.3.1`
- Builder:          `vite`
- User Config:     `app`, `modules`, `runtimeConfig`,`vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Page Transition with console.log in route navigation guard causing an error
https://stackblitz.com/edit/github-wrkvqn

Page Transition with defineExpose and setup functions
https://stackblitz.com/edit/github-ams2ou

### Describe the bug

The transitions property in the app.vue component causes an error when using a console log within a navigation guard. For example, within onEnter. It seems that it's also affected when using the names of navigation guards within the console log string

There seems to be an issue with the defineExpose function when using it on a page component instead of a “regular” component. On a regular component it works fine, but for a page component that’s being accessed in app.vue in the transitions property via a ref it fails.  I’ve gone back and forth with @daniele in direct messages to investigate the issue and find a solution for this, but this has been unsuccessful. Some of the things that were tried are: disabling the layout transitions, checking root nodes.

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/19268

### ❓ Type of change

### 📚 Description 
Hi :wave: this PR makes using `ref` on `<NuxtPage>` possible. 

The idea behind this PR is to set a `ref` on the page component called within `RouteProvider` and expose it. Then `NuxtPage` component will also expose what has been exposed through the `RouteProvider`'s ref.

I did some manuel tests, and didn't find any memory leaks.

### 📝 Checklist


antfu:
I am more curious about what would be the use cases behind it. This to me feels like quite some code added for features that most of ppl don't need. And also be able to work around with a custom inject/provide etc.

Other than that, I also feel the current design is a bit implicit, where the ref gets the child, not the component feels a bit counter-intuitive. This also makes it harder to keep type safety.

huang-julien:
I was planning to try another approach. 
Maybe should we do something done by vue core [here](https://github.com/vuejs/core/blob/650f5c26f464505d9e865bdb0eafb24350859528/packages/runtime-core/src/apiAsyncComponent.ts#L218)  ? Do you think This can be possible ?
This feature will indeed not be used by most of ppl 🤷‍♂️  , but I think those who wants to use it are expecting to be able to have an acess to refs.

antfu:
I would first challenge the real-world use case of the original request, and see if there are better solutions to achieve what they need (maybe no need to mess with the ref) before discussing the design of API etc.

JasonGotGithub:
Hey @danielroe @antfu, what information do you need specifically to progress this?

Does an example in Nuxt 2 help? This way you can see the use case in a demo instead of text

danielroe:
> Does an example in Nuxt 2 help? This way you can see the use case in a demo instead of text

That would be great ❤️

JasonGotGithub:
Cool! I'll make a demo asap just to be sure the use case is clear

JasonGotGithub:
Hey @danielroe and @antfu, I've made a quick [example](https://stackblitz.com/edit/github-c1mfzg). It basically contains two concepts:

1. Accessing a component which is (almost) always visible on a page. In this case it's the Navigation component. In the example the timeline tween gets triggered exactly when leaving a page. Reason being that visually it looks cohesive. It's possible to trigger the animation with a $route watcher in this case. However, it visually will happen after the page is shown. 
2. Looking at some data from the page component or its child component. In this case I've kept in simple. Based on if the home page timeline tween has progressed past a certain point you either reverse the timeline tween or you pause it and make a simpler tween. In more complex situations you might have a staggered animation of letters from -100% to 0%. If the animation has progress to something like .95, it might be more visually appealing to create a new staggered animation which goes to 100% then to reverse the existing/original staggered animation


Even though I've added two concepts, it's not exhaustive. Another concept/example might be accessing data from an 3DPool (when working with WebGL) and manipulating/animating those before arriving at a new route or finishing the route navigation

JasonGotGithub:
Hey @danielroe and @antfu, I want check in to see if you have everything you need from me for this and if I can help out in some way

JasonGotGithub:
Hi @antfu & @danielroe, is there an update on this?

danielroe (dismissed):
I'm very much up for the concept behind this. If we can expose the exposed properties of the page component 'upwards' that seems good to me. But I am wary of the particular implementation here which is clever, but very involved and I fear might be breakable.

How about exposing a proxy that accesses whatever the page exposes instead? That way we wouldn't need to manipulate vnodes. It would add an extra property name but otherwise pretty simple. Something like:

```ts
const pageRef = ref()

// later
pageRef.value.page.someMethod()
```

Mind you, I haven't tried this. What do you think?

danielroe:
Pushed an implementation. Let me know what you think and feel free to update/amend.

huang-julien:
The implementation is working well and way simpler than what i did. I like it :+1:

danielroe:
Can you think a better API/name for the exposed property? Maybe `pageRef` (similar to `pageKey`) and `layoutRef` directly?

huang-julien:
Sure ! This would be less confusing (`pageRef` and `pageLayout`)

danielroe (approved):

danielroe:
Just one more idea - might we pass `page-ref` as a _prop_? So then it removes double-level access?

huang-julien:
I think the tempalte transformation would transform `:page-ref="page"` in to `:page-ref="page.value"` when a ref need the full `ref()` object

PR #19458, opened 2023-03-05
feat(nuxt): support `trailingSlashBehavior` in `defineNuxtLink`

### LINKED ISSUE #18806: Support `trailingSlash` behavior for `createNuxtLink`
### Describe the feature

Related: https://github.com/nuxt/nuxt/issues/15462

To improve SEO for internal linking, `createNuxtLink` should get a new option `trailingSlashBehavior` that should be `APPEND`, `REMOVE` or `DO_NOT_TOUCH`, which is the default.

Then, people can create their own `<NuxtLink>` component with the respective behavior.

### Additional information

### Final checks

alex-key:
### 🔗 Linked issue

resolves #18806

### ❓ Type of change

### 📚 Description
Add a new option `trailingSlashBehavior` to options in `defineNuxtLink`. Using it user can define his own component like `<MyNuxtLink />` and either force always trailing slash or force no-slash for all internal urls.
Resolves #18806

### 📝 Checklist


alex-key:
@danielroe Daniel, could you confirm those failures are related to PR changes? It seems unrelated to me.

manniL:
> AssertionError: expected 93412 to be less than 93000

@alex-key you might have to update the test as the default size of the nuxt application increased.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -213,7 +223,9 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
         if (!isExternal.value) {
           const routerLinkProps: Record<string, any> = {
             ref: process.server ? undefined : (ref: any) => { el!.value = ref?.$el },
-            to: to.value,
+            to: options?.trailingSlashBehavior === 'append' || options.trailingSlashBehavior === 'remove'
+              ? resolveTrailingSlashBehavior(router.resolve(to.value))
+              : to.value,
```
manniL, line 228: idea: Could we reduce that down to a truth check?
```suggestion
            to: options?.trailingSlashBehavior
              ? resolveTrailingSlashBehavior(router.resolve(to.value))
              : to.value,
```

TheAlexLichter (approved):
LGTM 👍

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: In this case we'll have to send `router` as the second argument to the resolver function, add typings for `router` etc. To me it looks more complicated, so I decided to keep this check here and just send a resolved `Route` object.
Also current version helps to quickly recognize that `to` is not mutated if trailing slash strategy is not defined.

alex-key: To clarify what I mean: if we send resolved `route` and in the resolver based on no strategy defined we should just return original `to` value - we cannot do that.

docs/3.api/2.components/4.nuxt-link.md
```diff
@@ -115,6 +115,7 @@ defineNuxtLink({
   activeClass?: string;
   exactActiveClass?: string;
   prefetchedClass?: string;
+  trailingSlashBehavior?: 'append' | 'remove'
```
danielroe, line 118: @manniL what do you think about making the option name `trailingSlash` - as we don't accept boolean (wise!), I think append/remove make clear what will happen.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -53,6 +54,15 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
   }
+  const resolveTrailingSlashBehavior = (to: RouteLocation): RouteLocationPathRaw => {
+    const trailingSlash = options.trailingSlashBehavior === 'append' ? '/' : ''
+    // modify trailing slash in "path", exclude "name" to apply trailing slash behavior
+    return {
+      path: to.path.replace(/\/$/, '') + trailingSlash,
+      query: to.query,
+      hash: to.hash
+    }
+  }
```
danielroe, line 65: I think we can improve this a little:

1. if route passed is a string, we can use `withTrailingSlash` from `ufo` to add the slash and return the modified string
2. if route passed is an object without path, we can call `router.resolve` on the object to fully resolve it.
3. once we have path, we return an object with `.path` and `.fullPath` values with trailing slashes added.

docs/3.api/2.components/4.nuxt-link.md
manniL: Yes, fine for me 👍

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: Ok to all, except that `fullPath` is set by `vue-router`, when it resolves privided `to/href`, so I think only `path` should be modified.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -53,6 +54,32 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
   }
+  const resolveTrailingSlashBehavior = (
+    to: RouteLocationRaw,
+    resolve: (to: RouteLocationRaw) => RouteLocation & { href?: string }
+  ): RouteLocationRaw | RouteLocation => {
+    if (options.trailingSlash !== 'append' && options.trailingSlash !== 'remove') {
+      return to
+    }
+
+    const forceAppend = options.trailingSlash === 'append'
```
danielroe, line 65: you can simplify the logic below like this:

```suggestion
    const method = options.trailingSlash === 'append' ? withTrailingSlash : withoutTrailingSlash
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -53,6 +54,32 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
       console.warn(`[${componentName}] \`${main}\` and \`${sub}\` cannot be used together. \`${sub}\` will be ignored.`)
     }
   }
+  const resolveTrailingSlashBehavior = (
+    to: RouteLocationRaw,
+    resolve: (to: RouteLocationRaw) => RouteLocation & { href?: string }
+  ): RouteLocationRaw | RouteLocation => {
+    if (options.trailingSlash !== 'append' && options.trailingSlash !== 'remove') {
```
danielroe, line 61: would be worth early returning too if `to` is null or undefined

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -53,6 +54,32 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+      return to
+    }
+
+    const forceAppend = options.trailingSlash === 'append'
+    if (typeof to === 'string') {
+      console.log('withTrailingSlash', to, withTrailingSlash(to, true))
```
danielroe, line 67:
```suggestion
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -53,6 +54,32 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+        path: forceAppend ? withTrailingSlash(to.path, true) : withoutTrailingSlash(to.path, true)
+      }
+    }
+
+    const route: RouteLocation = resolve(to)
+    return {
+      path: forceAppend ? withTrailingSlash(route.path, true) : withoutTrailingSlash(route.path, true),
+      query: route.query,
+      hash: route.hash
+    }
```
danielroe, line 81: this ends up limiting the resolved route to just those three properties - is there a reason?

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: `if (!to || ...)`?

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: It will be later resolved in `RouterLink`. We should not provide `name` otherwise `path` will be overwritten. Others are not needed. These 3 are fully describing route AFAIK

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: btw, just asking: why did you disable eslint rule `no-console`?

danielroe: we emit console logs deliberately in lots of places

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -213,7 +240,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
         if (!isExternal.value) {
           const routerLinkProps: Record<string, any> = {
             ref: process.server ? undefined : (ref: any) => { el!.value = ref?.$el },
-            to: to.value,
+            to: resolveTrailingSlashBehavior(to.value, router.resolve),
```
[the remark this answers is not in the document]
alex-key: Yes, it was an initial intention to make a check for `trailingSlash` before resolver is called. Though in current implementation the logic is cleaner as all related checks are inclapsulated in `resolveTrailingSlashBehavior`

packages/nuxt/src/app/components/nuxt-link.ts
alex-key: done

alex-key:
Pushed suggested fixes so far

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -19,6 +19,7 @@ export type NuxtLinkOptions = {
   activeClass?: string
   exactActiveClass?: string
   prefetchedClass?: string
+  trailingSlash?: 'append' | 'remove'
```
harlan-zw, line 22: Any reason not to just use a `boolean`? it seems pretty explicit already

danielroe: We experienced some issues in Nuxt 2 where there was a subtle but important difference between `trailingSlash: false` and `trailingSlash: undefined`. Being more explicit here I think will avoid some of the problems. @manniL may also have some insight.

danielroe:
@alex-key have you pushed all of your changes so far?

alex-key:
Yes, done

danielroe (approved):

packages/nuxt/src/app/components/nuxt-link.ts
danielroe: I think just route state/history data. I've pushed a fix but let me know if you think of a better approach.

alex-key: Looks good. I've just never seen a `to.state` usage in projects

test/fixtures/basic/pages/nuxt-link/trailing-slash.vue
```diff
@@ -0,0 +1,73 @@
...
+  '/nuxt-link/trailing-slash?test=true&thing=other/thing#thing-other',
+  '/nuxt-link/trailing-slash/?test=true&thing=other/thing#thing-other',
+  { name: 'nuxt-link-trailing-slash' },
+  { query: { 'with-state': 'true' }, state: { foo: 'bar' } },
+  { query: { 'without-state': 'true' } }
+]
+
+const route = useRoute()
+const windowState = computed(() => {
+  console.log(route.fullPath)
```
alex-key, line 21: `console.log` intended?

danielroe: Fair play 😆

packages/nuxt/src/app/components/nuxt-link.ts
manniL: Strings are a bit more DX-friendly so you avoid false-positive checks like `if(!trailingSlash)`. ☺️

docs/3.api/2.components/4.nuxt-link.md
```diff
@@ -123,6 +124,7 @@ defineNuxtLink({
 - **activeClass**: A default class to apply on active links. Works the same as [Vue Router's `linkActiveClass` option](https://router.vuejs.org/api/#linkactiveclass). Defaults to Vue Router's default (`"router-link-active"`)
 - **exactActiveClass**: A default class to apply on exact active links. Works the same as [Vue Router's `linkExactActiveClass` option](https://router.vuejs.org/api/#linkexactactiveclass). Defaults to Vue Router's default (`"router-link-exact-active"`)
 - **prefetchedClass**: A default class to apply to links that have been prefetched.
+- **trailingSlash**: An option to force trailing slash in all urls or remove by default. If option is not defined or different from `"append"`/`"remove"` it will be ignored.
```
manniL, line 127:
```suggestion
- **trailingSlash**: An option to either enforce or remove trailing slashes URLs . If  unset or not matching the valid values `"append"` or `"remove"`, it will be ignored.
```

TheAlexLichter (approved):

test/fixtures/basic/pages/nuxt-link/trailing-slash.vue
danielroe:
```suggestion
```

danielroe: oops - I forgot why it was there - to make this reactive to the route - otherwise the window history won't be updated when the route changes, IIRC.

docs/3.api/2.components/4.nuxt-link.md
```diff
@@ -123,6 +124,7 @@ defineNuxtLink({
 - **activeClass**: A default class to apply on active links. Works the same as [Vue Router's `linkActiveClass` option](https://router.vuejs.org/api/#linkactiveclass). Defaults to Vue Router's default (`"router-link-active"`)
 - **exactActiveClass**: A default class to apply on exact active links. Works the same as [Vue Router's `linkExactActiveClass` option](https://router.vuejs.org/api/#linkexactactiveclass). Defaults to Vue Router's default (`"router-link-exact-active"`)
 - **prefetchedClass**: A default class to apply to links that have been prefetched.
+- **trailingSlash**: An option to either add or remove trailing slashes in the `href`. If unset or not matching the valid values `append` or `remove`, it will be ignored.
```
danielroe, line 127:
```suggestion
- **trailingSlash**: An option to either add or remove trailing slashes in the `href`. If unset or not matching the valid values `append` or `remove`, it will be ignored. **This option is currently only available on the [Edge Channel](/docs/guide/going-further/edge-channel/).** <!-- stabilityedge -->
```

harlan-zw (approved):

PR #19465, opened 2023-03-05
feat(nuxt): allow accessing `NuxtLayout` ref via `layoutRef`

huang-julien:
### ❓ Type of change

### 📚 Description

This PR allow the same thing as https://github.com/nuxt/nuxt/pull/19403 but for `NuxtLayout` : allow using `ref` .
However this ones use a prop `exposed` to tranfer data since the LayoutLoader's setup is async.
todo: 
- [x] manual test for memory leaks
- [x] fix tests

### 📝 Checklist


huang-julien:
@danielroe It seems that this also solves another issue. On the current main branch, [this line](https://github.com/nuxt/nuxt/blob/a7d75a84659b9c9cd93e099df4a7bf66409771b1/packages/nuxt/src/app/components/layout.ts#L28) is not triggered
And [this test](https://github.com/nuxt/nuxt/blob/a7d75a84659b9c9cd93e099df4a7bf66409771b1/test/basic.test.ts#L479) should be throwing since `with-props` layout does render a Fragment or multiple root nodes.
I fixed the CI by setting a single root node within `with-props.vue`, should I add another test to make sure we get a warning if a layout has multiple root nodes  ?

danielroe:
Added a comment here that is also relevant, I think: https://github.com/nuxt/nuxt/pull/19403#pullrequestreview-1473366308

packages/nuxt/src/app/components/layout.ts
```diff
@@ -34,14 +36,12 @@ const LayoutLoader = defineComponent({
     const LayoutComponent = await layouts[props.name]().then((r: any) => r.default || r)
 
     return () => {
-      if (process.dev && process.client && props.hasTransition) {
-        vnode = h(LayoutComponent, context.attrs, context.slots)
-        return vnode
-      }
-      return h(LayoutComponent, context.attrs, context.slots)
+      vnode = h(LayoutComponent, mergeProps(context.attrs, { ref: props.layoutRef }), context.slots)
+      return vnode
```
danielroe, line 40:
```suggestion
      if (process.dev && process.client && props.hasTransition) {
        vnode = h(LayoutComponent, mergeProps(context.attrs, { ref: props.layoutRef }), context.slots)
        return vnode
      }
      return h(LayoutComponent, mergeProps(context.attrs, { ref: props.layoutRef }), context.slots)
```

packages/nuxt/src/app/components/layout.ts
```diff
@@ -16,6 +16,8 @@ const LayoutLoader = defineComponent({
   inheritAttrs: false,
   props: {
     name: String,
+    exposed: Object as Record<string, any>,
```
danielroe, line 19:
```suggestion
```

danielroe (approved):

PR #19473, opened 2023-03-06
fix(test-utils): allow `dev` nuxtConfig overrides

harlan-zw:
### ❓ Type of change

### 📚 Description

Currently, it's not possible to provide nuxtConfig overrides for the `dev` environment testing. This is because the dev server is started with the equivalent of `nuxi dev` and doesn't have any context of the test being run.

There are certain ways around this limitation that are already being used in the test suite, namely just setting the individual environment variables for key config (i.e `process.env.NUXT_APP_BASE_URL`). This is a bit too limiting for modules though, instead, we provide a simple env variable that can be used to parse in a JSON-encoded nuxt config. 

Additionally, it's not even possible to just run a dev-test without creating a fixture and doing a build, which slows down the tests. I've reworked the logic to follow:
- either it's a dev test and we just need to `nuxi dev` the provided fixture
- OR it's not a dev test and we need to either build a server, build a fixture and / or run the node server.

### 📝 Checklist


harlan-zw:
Hey just FYI I'm exploring some ideas here. I originally made the PR just to allow custom config for the dev environment but I got blocked because you can't actually run tests that _only_ do the dev server testing. So a simple fix can't be done without also refactoring out other parts of the test suite.

I think overall the DX of targeting a specific environment to test (i.e `dev server`, `node server`) should be improved first.

Atinux:
Would the c12 update on overwriting configuration on environments will fix your issue @harlan-zw ?

https://github.com/unjs/c12#environment-specific-configuration

PR #19487, opened 2023-03-06
fix(nuxt): improve types for data fetching with `transform`

### LINKED ISSUE #14437: Type wrong: `default` option of `useFetch`
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-rc.6`
- Builder:          `vite`
- User Config:      `typescript`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-z9sira?file=app.vue

### Describe the bug

When the type has be changed by useFetch's `transform`, `default`'s return type should be its after transformed.

https://github.com/nuxt/framework/blob/v3.0.0-rc.6/packages/nuxt/src/app/composables/fetch.ts#L26

```
app.vue:29:20 - error TS2322: Type 'string[]' is not assignable to type 'Sample | Ref<Sample>'.

29     default: () => ['Unknown'],
                      ~~~~~~~~~~~

  node_modules/nuxt/dist/app/composables/asyncData.d.ts:11:15
    11     default?: () => DataT | Ref<DataT>;
                     ~~~~~~~~~~~~~~~~~~~~~~~~
    The expected type comes from the return type of this signature.


Found 1 error in app.vue:29


 ERROR  Command failed with exit code 2: vue-tsc --noEmit               00:51:48

  at makeError (node_modules/nuxi/dist/chunks/index5.mjs:1257:11)
  at handlePromise (node_modules/nuxi/dist/chunks/index5.mjs:2086:35)
  at <anonymous> (<anonymous>)
```

### Additional context

_No response_

### Logs

_No response_

clarkdo:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14437

### ❓ Type of change

### 📚 Description

Current `AsyncDataOptions` is not able to infer correct type on `default` option when `transform` exists, this pr fixes the issue and also improve type inference on composables `useAsyncData/useLazyAsyncData/useFetch/useLazyFetch` by inferring transformed data instead of transform function. 

### 👉 Migration

⚠️ This is a potential **breaking change** if the user is using generics explicitly. ⚠️

No migration is needed if you are only using the first generic of `useFetch` or `useAsyncData`. But if you were typing more fields, or trying to clone the Nuxt types, you will need to update your types to use (instead of a type for the transform function, you will provide a type for the return value of the transform function).

This will probably not affect you as the types were buggy in this case (which was what this PR was solving).

### 📝 Checklist


danielroe (commented):
This looks really nice, thank you! ❤️ 

Do you think there is a way we can keep this from being a breaking change, e.g. by using a function signature overload?

danielroe (approved):

Issue #19482, opened 2023-03-06
Impossible to configure `postcss-preset-env` in Nuxt 2.16


obulat:
### Environment

System:
    OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (12) x64 AMD EPYC
    Memory: 14.14 GB / 15.63 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 16.17.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.15.0 - /usr/local/bin/npm
    Watchman: 20220528.183901.0 - /usr/local/bin/watchman
  npmPackages:
    @nuxt/vue-app: ^2.16.2 => 2.16.2 
    nuxt: ^2.16.2 => 2.16.2 

### Reproduction

https://codesandbox.io/p/github/obulat/starter/draft/admiring-goldwasser?workspace=%257B%2522activeFileId%2522%253Anull%252C%2522openFiles%2522%253A%255B%255D%252C%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522gitSidebarPanel%2522%253A%2522COMMIT%2522%252C%2522spaces%2522%253A%257B%257D%252C%2522currentSpace%2522%253Anull%252C%2522spacesOrder%2522%253A%255B%255D%252C%2522hideCodeEditor%2522%253Afalse%257D

### Describe the bug

It is impossible to configure `postcss-preset-env` in Nuxt 2.16. There is a `preset` property in `nuxt.config`'s `build.postcss.postcssOptions`, but Nuxt does not use it. Instead, it tries to use the old `preset` option in `build.postcss` object.

It seems that #8899 might be touching this functionality, but I'm not sure if it's fixing this issue.

We have tried to update Nuxt from 2.15.8 to 2.16.x in [Openverse](https://github.com/WordPress/openverse), but it breaks the RTL layout because the `preset-env` by default is set not to support CSS logical properties like `margin-inline-start` (they are stage 2 of browser support).

The `postcss.preset` settings from `nuxt.config` are not passed to the `postcss-preset-env` plugin.

Currently, if postcss settings are not set using a file (the deprecated way), `webpack/src/utils` gets 2 objects from `nuxt.config`: `postcssLoaderOptions` (the value of `build.postcss`) and `postcssOptions` (the value of `build.postcss.postcssOptions`). 

[...]

PR #19495, opened 2023-03-07
feat(webpack): support passing function as `postcssOptions`

### LINKED ISSUE #7015: using function for build.postcss config option

obulat:
### 🔗 Linked issue

resolves #7015 
#19482
closes #8899

### ❓ Type of change

### 📚 Description

This is an updated version of #8899 that applies the changes to the now-default PostCSS 8 config.

If postcss config is a function, we execute the custom function, but we also apply the default config of Nuxt. Therefore we can ensure that the config has a minimum default.

Postcssconfig can either be an object (already supported) or a function which gets passed in the loadercontext. But nuxt does not support passing in a function currently.

#### Why

Passing in a function allows more complex uses cases, e.g. by applying different postcss steps for different ressources. (see also https://github.com/webpack-contrib/postcss-loader#function)

#### Solution

This approach adds a new "postcssConfig" option to the "postcss" Nuxt config. Either this is the regular postcssConfig, or it is a function. see https://webpack.js.org/loaders/postcss-loader/#postcssoptions
(only works with postcss@8, because postcss-loader >= 4.0 is required, and postcss7 nuxt setup conflicts with this version of postcss-loader)

e.g.
nuxt.config.js.
```
module.exports = {
  build: {
    postcss: {
      postcssOptions: loaderContext => {
        // e.g. do something else with ".sss" files
        if (/\.sss$/.test(loaderContext.resourcePath)) {
          return {
            parser: 'sugarss',
            plugins: [['postcss-short', { prefix: 'x' }], 'postcss-preset-env']
          };
        }

        return {
          plugins: [['postcss-short', { prefix: 'x' }], 'postcss-preset-env']
        };
      }
    }
  }
};
```
### 📝 Checklist


obulat:
I am not sure that this change correctly applies the `preset` properties to the `postcss-preset-env` configuration.
I also don't know how to add tests. Is there an existing test that I could adjust for this change, @danielroe?

I would also need to open a separate PR for updating the documentation. The current section on PostCSS config sessions has not yet been updated to Nuxt 2.16 changes (https://nuxtjs.org/docs/configuration-glossary/configuration-build#postcss)

obulat:
> I am not sure that this change correctly applies the preset properties to the postcss-preset-env configuration.

I have extracted the changes to `preset` to a separate PR (#19518) because it makes the review faster, and I could test that it works.

I'm not sure how to test this PR, though.

danielroe:
@obulat would you be happy to resolve conflicts in this PR? I can help with adding a test fixture.

danielroe (dismissed):

danielroe (approved):

danielroe:
Merging to include in 2.17. Feel free to come back to me if you think there's anything missing here.

obulat:
> Merging to include in 2.17. Feel free to come back to me if you think there's anything missing here.

Thank you for getting this PR finished, @danielroe!

PR #19518, opened 2023-03-08
fix(config): move `preset` to inner `postcssOptions`

### LINKED ISSUE #19482: Impossible to configure `postcss-preset-env` in Nuxt 2.16

obulat:
### 🔗 Linked issue

#19482 

### ❓ Type of change

### 📚 Description

Resolves #19482. Nuxt 2.16 has updated PostCSS to version 8, and introduced a change into the `nuxt.config` `build.postcss` settings. It did not, however, update the way the defaults are set, which made it impossible to set `preset`.

This PR moves the default `preset` from `build.postcss` to `build.postcss.postcssOptions`. It also adds a test for the default setting, and updates the snapshots.

This PR was extracted from #19495 to make it easier to review.

### 📝 Checklist


packages/webpack/src/utils/postcss.js
```diff
@@ -210,6 +210,14 @@ export default class PostcssConfig {
       if (Array.isArray(postcssOptions.plugins)) {
         defaults(postcssOptions, this.defaultPostcssOptions)
       } else {
+        if (postcssOptions.preset || postcssLoaderOptions.preset) {
+          postcssOptions.plugins = {
+            ...postcssOptions.plugins || {},
+            'postcss-preset-env': merge({}, postcssOptions.preset || postcssLoaderOptions.preset)
```
danielroe, line 216: it might be worth merging _all_ of them to support legacy modules (e.g. tailwind module) that are still interacting with nuxt config based on the old patterns.

obulat: Do you mean merging _all_ plugins, @danielroe? The `postcss.plugins` with `postcss.postcssOptions.plugins`? I think that that is done on lines 199...202.

Currently, the type for postcss is set to `postcss?: string[] | boolean | { postcssOptions: PostcssConfiguration | (() => PostcssConfiguration) }`, so it actually disallows the old way of setting the options. Should this type be ignored or updated to include the values that will be deprecated but are still in use?

I tried refactoring a `normalize` function (https://gist.github.com/obulat/4fd4cf28ec816c20e5818a053b15f3c5) to move all the top-level properties (plugins, order, preset) inside the inner `postcssOptions`, prioritizing the values that are already within `postcssOptions`. Would this be useful?

danielroe: The normalize function sounds great! My primary concern is not to support users explicitly passing in an old format but rather to handle the case where _modules_ set `nuxt.options.build.postcss.plugins` (or similar). This happens after we initially normalise the config.

obulat: I didn't know about modules setting plugins :) Do they set it _after_ getting the config using `postcss.config()`? Then this solution will not work. But if they update the build config, and then call `postcss.config` with it, then this `normalize` function should work:

The config gets the outer `postcss` object, and then converts it by moving any top-level `plugins`, `preset` and `order` values to the inner `postcssOptions` object. Thus, the _config object_ that we use to load plugins is `{postcssOptions: {plugins: {}|[], preset: {}, order: {}}`, with _no other top-level properties_ but `postcssOptions`.

I spent a lot of time trying to debug the plugins that `postcss-preset-env` loads. I incorrectly understood the documentation to say, "the higher you set the `stage`, the more plugins will be added", and in fact, it is the opposite. While reading the documentation, I also saw that the default `stage` is already set in `postcss-preset-env` at `2`, so **I removed the default stage option** of `{preset: { stage: 2 }}`  from the settings here (replaced with `{preset: {} }`. Please, let me know if it's better to add it back.

packages/webpack/src/utils/postcss.js
```diff
@@ -79,11 +83,11 @@ export default class PostcssConfig {
           })
         },
 
-        // https://github.com/postcss/postcss-url
+        // https://github.com/csstools/postcss-plugins/plugin-packs/postcss-url
         'postcss-url': {},
 
-        // https://github.com/csstools/postcss-preset-env
-        'postcss-preset-env': this.postcssLoaderOptions.preset || {},
+        // https://github.com/csstools/postcss-plugins/plugin-packs/postcss-preset-env
```
danielroe, line 89:
```suggestion
        // https://github.com/csstools/postcss-plugins/tree/main/plugin-packs/postcss-preset-env
```

packages/webpack/src/utils/postcss.js
```diff
@@ -79,11 +83,11 @@ export default class PostcssConfig {
           })
         },
 
-        // https://github.com/postcss/postcss-url
+        // https://github.com/csstools/postcss-plugins/plugin-packs/postcss-url
```
danielroe, line 86: I think the old url is still correct:

```suggestion
        // https://github.com/postcss/postcss-url
```

packages/webpack/src/utils/postcss.js
```diff
@@ -141,12 +145,58 @@ export default class PostcssConfig {
...
+    if (postcssOptions.postcssOptions && typeof postcssOptions.postcssOptions === 'function') {
+      consola.warn('Using a Function as `build.postcss.postcssOptions` is not yet supported in Nuxt 2.16.2')
+      return { postcssOptions: {} }
+    }
+    if (!('postcssOptions' in postcssOptions)) {
+      if (Object.keys(postcssOptions).length > 0) {
+        consola.warn('Using the top-level properties in `build.postcss` will be deprecated in Nuxt 3. Please move' +
+          'the settings to `postcss.postcssOptions`')
+      }
+      postcssOptions = { postcssOptions: {} }
```
danielroe, line 179: should this be:
```suggestion
      postcssOptions = { postcssOptions }
```

danielroe (commented):
This looks great! ❤️

danielroe (approved):

danielroe:
Thank you for all your work on this ❤️

PR #19519, opened 2023-03-08
perf(head): drop `@vueuse/head` dependency

harlan-zw:
### ❓ Type of change

### 📚 Description

Related: https://github.com/nuxt/nuxt/pull/19302

Switching to using Unhead directly allows for better tree-shaking (i.e @unhead/dom from server build), simplifies the code, and minimizes dependencies.

The issue with dropping the dependency is any modules or third-party code which used the pre-v1 API @vueuse/head and/or directly imported from it may break. The scope of that is quite hard to determine.

We can solve issue using a polyfill and an alias. If we're okay with potentially breaking code for some codebases relying on old low level @vueuse/head code, then we could ship without the polyfill enabled by default.

Without the polyfill, there's a 0.47kB client bundle improvement, otherwise, it's only 0.1kB. This unlocks further optimizations.

### 📝 Checklist


test/bundle.test.ts
```diff
@@ -40,10 +40,10 @@ describe.skipIf(isWindows)('minimal nuxt application', () => {
 
   it('default server bundle size', async () => {
     stats.server = await analyzeSizes(['**/*.mjs', '!node_modules'], serverDir)
-    expect(stats.server.totalBytes).toBeLessThan(93000)
+    expect(stats.server.totalBytes).toBeLessThan(94400)
```
harlan-zw, line 43: the @unhead/vue dependency is being transpiled now, previously it wasn't (it's slightly bigger than @vueuse/head was)

danielroe (commented):
Nice!

packages/nuxt/src/head/runtime/lib/unhead.plugin.ts
```diff
@@ -16,25 +16,25 @@ export default defineNuxtPlugin((nuxtApp) => {
...
+    head.hooks.hook('dom:beforeRender', (context) => { context.shouldRender = !pauseDOMUpdates })
     nuxtApp.hooks.hook('page:start', () => { pauseDOMUpdates = true })
     // wait for new page before unpausing dom updates (triggered after suspense resolved)
     nuxtApp.hooks.hook('page:finish', unpauseDom)
     nuxtApp.hooks.hook('app:mounted', unpauseDom)
   }
 
-  // useHead does not depend on a vue component context, we keep it on the nuxtApp for backwards compatibility
+  // support backwards compatibility, remove at some point
   nuxtApp._useHead = useHead
```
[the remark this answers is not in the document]
harlan-zw: Good to know. I was planning another PR to remove #head and clean up the composables, might be better to remove it in that.

docs/1.getting-started/5.seo-meta.md
```diff
@@ -228,7 +230,7 @@ Now, if you set the title to `My Page` with `useHead` on another page of your si
 
 ### Body Tags
 
-You can use the `body: true` option on the `link` and `script` meta tags to append them to the end of the `<body>` tag.
+You can use the `tagPosition: 'bodyClose''` option on applicable tags to append them to the end of the `<body>` tag.
```
danielroe, line 233:
```suggestion
You can use the `tagPosition: 'bodyClose'` option on applicable tags to append them to the end of the `<body>` tag.
```

antfu:
👍 Nice!

Unrelated @harlan-zw: Do you think it makes sense to deprecate `@vueuse/head` in favor of `@unhead/vue`?

packages/schema/src/config/experimental.ts
```diff
@@ -119,6 +119,14 @@ export default defineUntypedSchema({
...
      * @see https://github.com/nuxt/nuxt/issues/15592
      */
-    configSchema: true
+    configSchema: true,
+
+    /**
+     * The head module uses this polyfill as a compatible layer for modules, plugins or user code relying on the old
+     * `@vueuse/head` API.
+     *
+     * This can be enabled for most Nuxt sites to reduce the client-side bundle by ~0.5kb.
```
danielroe, line 128:
```suggestion
     * This can be disabled for most Nuxt sites to reduce the client-side bundle by ~0.5kb.
```

packages/schema/src/config/experimental.ts
```diff
@@ -119,6 +119,14 @@ export default defineUntypedSchema({
      *
      * @see https://github.com/nuxt/nuxt/issues/15592
      */
-    configSchema: true
+    configSchema: true,
+
+    /**
+     * The head module uses this polyfill as a compatible layer for modules, plugins or user code relying on the old
+     * `@vueuse/head` API.
```
danielroe, line 126:
```suggestion
     * Whether or not to add a compatibility layer for modules, plugins or user code relying on the old
     * `@vueuse/head` API.
```

harlan-zw:
> +1 Nice!
> 
> Unrelated @harlan-zw: Do you think it makes sense to deprecate `@vueuse/head` in favor of `@unhead/vue`?

Yup! This is planned once I sort out a couple more outstanding issues.

danielroe (approved):

PR #19523, opened 2023-03-08
feat(vite): use custom logger to show vite logs

### LINKED ISSUE #13914: Clear console on HMR

antfu:
### 🔗 Linked issue

Fix #13914

### ❓ Type of change

### 📚 Description

By providing our own custom logger, we are able to integrate Vite's log in a more organic way (using consola). Making it supports `clearScreen` and we no longer need to hide Vite's log with `logLevel: 'warn'` (I see this as one of the sources of confusion as sometimes it's not clear to the user what is going on when we hide log from Vite)

https://user-images.githubusercontent.com/11247099/223693799-a1f920b0-ad3b-4bab-b7a5-94b3d21bcafa.mp4

### 📝 Checklist


packages/schema/src/config/vite.ts
```diff
@@ -18,9 +18,7 @@ export default defineUntypedSchema({
     mode: {
       $resolve: async (val, get) => val ?? (await get('dev') ? 'development' : 'production')
     },
-    logLevel:{
-      $resolve: async (val, get) => val ?? (await get('dev') ? 'warn' : 'info')
-    },
+    logLevel: 'info',
```
danielroe, line 21: should we drop this entirely so we respect top-level `logLevel` instead?
```suggestion
```

antfu: Good point!

danielroe (approved):
Lovely.

Issue #9230, opened 2021-05-02
Nuxt getting stuck while Client build


luckymichaln:
### Versions

- nuxt: `2.14.2`
- node: `15.0.1`

### Reproduction

Just install dependencies and run `yarn dev` to build the project.

### Steps to reproduce

See above

### What is Expected?

Both Client and Server bundle being build properly, app runs without any dependency errors.

### What is actually happening?

Locally Client bundle is getting stuck at 68-69% while building . After a while it passes to the end, but there's a lot of warnings from `./node_modules` and 154 dependencies are missing.

While build on Netlify, logs errors starts from

```
[fatal] Nuxt build error
6:51:52 PM:   ERROR in ./node_modules/sirv/build.mjs 42:18-22
6:51:52 PM:   Can't import the named export 'join' from non EcmaScript module (only default export is available)
@ (webpack)-bundle-analyzer/lib/viewer.js
```

Don't exactly know when it started to happen. I'm almost certain that I didn't added any new dependency whatsoever. I've already searched for some solution, but nothing really helped with it. I'm completely stuck with this, can anyone help me? :( 

Here's my package.json file:

```
{
  "name": "olszowski-nuxt-prismic",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "dev": "nuxt",
    "build": "nuxt build",
    "start": "nuxt start",
    "generate": "nuxt generate"
  },
  "dependencies": {
    "@babel/preset-env": "~7.12.0",
    "@nuxtjs/axios": "^5.13.1",
    "@nuxtjs/prismic": "^1.2.6",
    "@prismicio/client": "^4.0.0",
    "core-js": "^3.8.3",
    "node-sass": "^5.0.0",
    "nuxt": "^2.14.2",
    "nuxt-lazy-load": "^1.2.6",
    "prismic-javascript": "^3.0.2",
    "sass-loader": "^10.1.1",
    "vuex": "^3.6.2"
  },
  "devDependencies": {},
  "resolutions": {
    "@babel/preset-env": "7.12.17"
  }
}

```

PR #19526, opened 2023-03-08
fix(types): sync vue type augmentations with Vue 2.7

rchl:
### ❓ Type of change

### 📚 Description

Fixing an error:

```
../../node_modules/@nuxt/types/app/vue.d.ts:21:13 - error TS2428: All declarations of 'ComponentOptions' must have identical type parameters.

21   interface ComponentOptions<

../../node_modules/vue/types/options.d.ts:168:18 - error TS2428: All declarations of 'ComponentOptions' must have identical type parameters.

168 export interface ComponentOptions<
                     ~~~~~~~~~~~~~~~~
```

Note that I haven't tested this probably and I admit I don't **fully** understand the intentions here so this needs some testing.

### 📝 Checklist


rchl:
As mentioned in #9239, I've verified that this issue happens in Vue 2.7 but not in Vue 2.6.

Are we fine with going forward with types being only Vue 2.7-compatible. It might be the case already as Nuxt ships with VueRouter version that relies on Vue 2.7 already and there are also issues with importing `Vuex` in this package when using Vue 2.6.

```
node_modules/@nuxt/types/app/index.d.ts:5:28 - error TS2307: Cannot find module 'vuex' or its corresponding type declarations.

5 import type { Store } from 'vuex'
```

So, either we just revert #9230 or use take this PR which reverts the breaking change at least or we could go with something like https://github.com/rchl/nuxt/commit/0feb9fe1fa4420630958edd91800734671ee975a which adapts the changes to make them Vue 2.7 compatible.

danielroe (dismissed):

rchl:
Just for the record (since we're discussing this in private anyway), with the current state of the code the inferring of types of properties returned from `asyncData` (the intent of the original https://github.com/nuxt/nuxt/pull/9239 fix) still doesn't work.

danielroe:
That is a different issue, and not caused by the fix I pushed here, which is correct as far as I can tell, but rather by the fact that there are other parts of these declarations that also need to be updated to work with Vue 2.7.

rchl:
I know but we have two ways we can go here. Either just fix the crash or fix the crash and support inferring of asyncData return values.

The latter seems complicated and potentially risky (Vue types are a little brittle).

So IMO it's an option to just fix the crash without the support for inferring and I feel like this is much safer.

Not duplicating the `asyncData` properties in `data` has potential for breakage and in some cases different runtime behavior anyway. While in theory `asyncData` should always run first and set the properties on `data`, in practice we've seen crashes reported to Sentry when we forgot to duplicate those properties in `data`. This is potentially only an issue in very broken environments but still it can make a difference.

danielroe (approved):

PR #19530, opened 2023-03-08
feat(nuxt): add `watch` option and refactor dev server restarting

danielroe:
### ❓ Type of change

### 📚 Description

This provides a number of improvements to the dev server restarting behaviour:

1. it allows configuring the files that cause a server restart (with either string/RegExp - more complex patterns can be handled with `builder:watch` hook
2. it fixes implementation detail where we were only listening in dev server to rootDir changes, but we were trying to match against srcDir - this would have worked unless there was a custom srcDir (an undetected bug?)
3. it now offers more precise matching in pages module to detect changes and trigger restart
4. we now support restarting on actual component/composables/utils directories that are used rather than hard-coded values

What's more, because we're moving the logic into the ad-hoc modules that actually care about this, it should mean we integrate better with upcoming nuxi-ng, but definitely cc: @pi0 in case you foresee any issues here.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -1,6 +1,6 @@
 import type { AddressInfo } from 'node:net'
 import type { RequestListener } from 'node:http'
-import { resolve, relative, normalize } from 'pathe'
+import { resolve, relative } from 'pathe'
 import chokidar from 'chokidar'
 import { debounce } from 'perfect-debounce'
 import type { Nuxt } from '@nuxt/schema'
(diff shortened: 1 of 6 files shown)


antfu (approved):

atinux (approved):
Looks much better now and also bringing back a needed feature from Nuxt 2 to Nuxt 3 (ie: Tailwind module for instance)

KnifeFed:
How do you use this? I tried all imaginable combinations of `watch: ['tailwind.config.ts']` but it's not restarting the server when changing the file.

danielroe:
@KnifeFed Those paths need to be absolute. I've merged a fix to allow you to use paths relative to your `srcDir`, which will be included in the next patch release.

danielroe:
@FreekVR Yes, this was merged. If you are experiencing a regression, please do open a new issue with a reproduction.

plcdnl:
I'm so sorry but i'm trying to do it with nuxt 3.6.5 and it doesn't restart the app. Both with the absolute url and with the relative one

PR #19531, opened 2023-03-08
fix(vite): allow disabling clear screen behaviour

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/13914

### ❓ Type of change

### 📚 Description

This allows respecting `vite.clearScreen` option (which is set to `false` by default). If we want to enable by default we can do this here:

https://github.com/nuxt/nuxt/blob/624314600de6ceea7834e9afe14429bb4146c822/packages/schema/src/config/vite.ts#L65

### 📝 Checklist


antfu:
Oh I missed it. I think we could enable it by default

PR #19537, opened 2023-03-08
refactor(kit,nuxi): resolve module paths using node algorithm

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/18431#issuecomment-1460176412

### ❓ Type of change

### 📚 Description

The linked issue is caused by us trying to resolve an ESM module's _path_ by CJS utilities, and then import it with native ESM.

This refactors our internal CJS utils in kit + nuxi to use `resolvePath` from `mlly` which uses `import-meta-resolve` under the hood (a [polyfill for the Node built-in](https://github.com/wooorm/import-meta-resolve)). To decrease the change surface, I've restricted changes to places where we have a single search path (e.g. rootDir or cwd) and have kept the existing behaviour where we have multiple paths (e.g. where `nuxt.options.modulesDir` is used).

I would be up for refactoring the remaining usage of CJS utils as well, but am also happy to defer that to a subsequent PR.

As part of this, the behaviour of `tryResolveModule` (exposed by kit) has changed (it is now async, it now returns a file URL, and it takes a single string as a second argument rather than an object). However, I think this is safe as it was a deprecated internal module and **[not used outside of nuxt itself](https://github.com/search?q=tryResolveModule+kit+-path%3Apackages&type=code)**.

I've also decreased the number of internal utilities we are exposing from the CJS part of kit as a number of them are not being used elsewhere in the repo.

### 📝 Checklist


pi0:
This is a really lovely initiative! Only I think we should use [mlly.resolvePath](https://github.com/unjs/mlly#resolvepath) that wraps same package internally for less deps, more consistency and path normalizations.

pi0 (approved):
Looks ~good~ must have to me 🚀

PR #19545, opened 2023-03-09
feat: allow `pick` in fetch API to be string

### LINKED ISSUE #15603: Pick as root `data` when using fetch API
### Describe the feature

Currently, the `useFetch` and `useAsyncData` options we have are:

```ts
type UseFetchOptions = {
  // other keys...
  pick?: string[]
```

It would be nice to have the option to pick a single key (so `pick?: string | string[]`), where the only key is automatically unwrapped as the data key of the fetch return object. Example:

```ts
const { data } = await useFetch('https://reqres.in/api/users/2', {
  pick: 'data'
}

// Then I can access directly data.value.id instead of data.value.data.id
```
The response sample can be accessed here: [https://reqres.in/api/users/2](https://reqres.in/api/users/2).

### Additional information

### Final checks

smilosevic:
### 🔗 Linked issue

[#15603](https://github.com/nuxt/nuxt/issues/15603)

### ❓ Type of change

### 📚 Description

If `pick` is used as a string, the value of that key in the data structure will be returned. Resolves [#15603](https://github.com/nuxt/nuxt/issues/15603)

### 📝 Checklist


danielroe (changes requested):
This is a nice idea and solution for the linked issue.

However, we need to update the types for asyncData so that the return type is correct when using this. Would you be able to do that?

Atinux:
Actually, why not using `transform` for this instead?

pi0:
I also fully agree that this is a nice feature but also transform seems safer. Keeping type complexity low and easily possible / readable to do. We might document this solution.

smilosevic:
Hi everyone. This is my first pull request, and I'm not sure what to do next. Should I close it if we are not going with this?

danielroe:
I think let's close it for now. We're thinking it would be simpler to advise users to use the `transform` option instead rather than increase type complexity, e.g.

```ts
useFetch('/api/test', {
  transform: data => data.thing,
  // rather than pick: 'thing'
})
```

The PR was otherwise good - thank you ❤️ I'll look forward to seeing your next one!

PR #19578, opened 2023-03-10
fix(nuxi): Improve server listening when 3000 already used

### LINKED ISSUE #19560: Improve server listening when 3000 already used
I believe we had it before but know the experience is quite frustrating if you already have a Nuxt server running.

It takes a random port and on top of it it is not reusing the last port when restarting the server:

https://volta.s3.fr-par.scw.cloud/Clean_Shot_2023_03_09_at_14_58_35_b46c3b8bd2.mp4

yangxiuxiu1115:
### ❓ Type of change

### 📚 Description

fixed #19560 ,Improve server listening when 3000 already used

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -58,7 +58,10 @@ export default defineNuxtCommand({
       showURL: false,
       clipboard: args.clipboard,
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+      port: {
+        port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+        portRange: [3000, 3100]
+      },
       hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
       https: (args.https !== false && (args.https || config.devServer.https))
         ? {


packages/nuxi/src/commands/dev.ts
```diff
@@ -58,7 +58,10 @@ export default defineNuxtCommand({
       showURL: false,
       clipboard: args.clipboard,
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+      port: {
+        port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
+        portRange: [3000, 3100]
```
pi0, line 63: I believe we could use it as default in listhen directly to use other fallbacks (actually this was behavior for trying 3000 + a range but behavior regressed as result of another fix)

danielroe: Okay, so let's resolve upstream in `listhen` rather than here. 👍 

whack-a-mole

Atinux: Agree, it seems to be a regression in `listhen` but thank you very much for the PR @RicardoErii ❤️

RicardoErii: It's okay, so I close this PR?

Atinux: Let's keep it open until https://github.com/unjs/listhen/issues/62 is fixed

pi0:
The issue should be fixed with [`listhen@1.0.4`](https://github.com/unjs/listhen/releases/tag/v1.0.4)

PR #19605, opened 2023-03-11
fix(nuxt): directly render server components

### LINKED ISSUE #18500: async server components are not rendered in initial HTML
### Environment

RootDir: /home/projects/github-aegvia        
Nuxt project info:                                    

- Operating System: `Linux`
- Nuxt Version:     `3.1.0`
- Nitro Version:    `2.0.0`
- Builder:          `vite`
- User Config:      `experimental`, `modules`
- Runtime Modules:  `nuxt-graphql-client@0.2.24`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-aegvia

### Describe the bug

When using the new server islands, as soon as you use a method from a module, it results in a hydration mismatch, even if the html template is not using the data. Check the reproduction for a working and not-working example.

Bildschirm­foto 2023-01-25 um 08 07 36

Works fine:
```vue
<template>
	<div>This component will only be rendered on the server.</div>
</template>

<script setup lang="ts">
	console.log('Server only!')
</script>
```

Also works fine:
```ts
// utils/example.ts
export const example = () => { return true }
```
```vue
<template>
	<div>This component will only be rendered on the server.</div>
</template>

<script setup lang="ts">
	const exampleRes = example()
	console.log('Server only!', { exampleRes })
</script>

```

Hydration mismatch (e.g. using another packages method, e.g. `nuxt-graphql-client`'s `useAsyncGql` method).
```vue
<template>
	<div>This component will only be rendered on the server.</div>
</template>

<script setup lang="ts">
	const { data, error } = await useAsyncGql({
		operation: 'ListPosts',
	})
	console.log('Server only!', { data: data.value, error: error.value })
</script>
```

So it seems that there are issues in server-components when using composables from other packages. The `useAsyncGql` composable should work here because it is one that runs on the server + client normally. So it should not be the cause.

Let me know what to add to the reproduction to test when a hydration error happens and when not.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #19604: NuxtIsland fails to render when its props change
### Environment

Nuxt 3.2.3 and edge

### Reproduction

https://stackblitz.com/edit/nuxt-starter-aq2t65?file=app.vue

Click on the first `add +1 to count`

watch the component disapear

### Describe the bug

Hi :wave:
this was found when working on a PR for server components. 
`NuxtIsland` seems to have an issue when it triggers its first re-render due to props change.

This is probably due to `html.value` being an empty string in `NuxtIsland`'s setup. But strangely `.server` components doesn't seems to have this issue

PR Incoming :) 

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/19604
resolve https://github.com/nuxt/nuxt/issues/18500

### ❓ Type of change

### 📚 Description

Hi :wave: 
This PR aims to allow using async server components by removing the usage of `Teleport`
It also fix a bug on `NuxtIsland` -> https://github.com/nuxt/nuxt/issues/19604

It remove `Teleport` and use what has been rendered instead. We just have to remove the root node to get the server components result.

### 📝 Checklist and to do

- [x] wait for https://github.com/nuxt/nuxt/issues/19623 to be resolved
- [x] finish writing tests
- [x] test .server components async
- [x] I have updated the documentation accordingly. (remove async warning)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -159,6 +159,7 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
 
 const PAYLOAD_CACHE = (process.env.NUXT_PAYLOAD_EXTRACTION && process.env.prerender) ? new Map() : null // TODO: Use LRU cache
 const PAYLOAD_URL_RE = /\/_payload(\.[a-zA-Z0-9]+)?.js(\?.*)?$/
+const ROOT_NODE_REGEX = /^<div id="__nuxt">([\s\S]*)<\/div>$/
```
danielroe, line 162: Unfortunately, we allow the id + tag to be customised (`app.rootId` and `app.rootTag`) so this regexp will not always work, but we do have access above on:

https://github.com/nuxt/nuxt/blob/859fbd0add7e97633c3350b9baf3aeb271930081/packages/nuxt/src/core/runtime/nitro/renderer.ts#L17

huang-julien: thanks !

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -159,6 +159,7 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
 
 const PAYLOAD_CACHE = (process.env.NUXT_PAYLOAD_EXTRACTION && process.env.prerender) ? new Map() : null // TODO: Use LRU cache
 const PAYLOAD_URL_RE = /\/_payload(\.[a-zA-Z0-9]+)?.js(\?.*)?$/
+const ROOT_NODE_REGEX = new RegExp(`<${appRootTag} id="${appRootId}">([\\s\\S]*)</${appRootTag}>`)
```
danielroe, line 162: should we keep the `^`/`$` chars still?

huang-julien: yes we'll keep it just in case :+1:

test/fixtures/basic/server/api/very-long-request.ts
```diff
@@ -0,0 +1,7 @@
+export default defineEventHandler(async () => {
+  await timeout(1000)
```
danielroe, line 2: this probably could be much shorter for the sake of our CI

danielroe:
```suggestion
  await timeout(20)
```

huang-julien: what do you think of 60ms ? (i forgot to set this pr to draft 😅 )

danielroe: perfect. oops, sorry for the quick review

danielroe (approved):

PR #19608, opened 2023-03-11
fix(nuxt): resolve builder using esm syntax

danielroe:
### ❓ Type of change

### 📚 Description

Similar to https://github.com/nuxt/nuxt/pull/19601, we also need to resolve `@nuxt/vite-builder` via rootDir/nuxt.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/builder.ts
+++ b/packages/nuxt/src/core/builder.ts
@@ -1,5 +1,5 @@
 import chokidar from 'chokidar'
-import { importModule, isIgnored } from '@nuxt/kit'
+import { isIgnored, tryResolveModule } from '@nuxt/kit'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/builder.ts
```diff
@@ -78,3 +78,12 @@ async function bundle (nuxt: Nuxt) {
     throw error
   }
 }
+
+async function loadBuilder (nuxt: Nuxt, builder: string) {
+  for (const root of [nuxt.options.rootDir, import.meta.url]) {
+    const builderPath = await tryResolveModule(builder, root)
```
pi0, line 84: I think we can pass search path directly to the kit util (at least mlly reslver supports it)

PR #19684, opened 2023-03-14
fix(nuxt): ignore falsy modules

antfu:
### ❓ Type of change

### 📚 Description

This allows user to enable some modules conditionally, say:

```ts
  modules: [
    '@vueuse/nuxt'
    process.env.CI ? undefined : '@nuxt/devtools',
  ],
})
```

Currently it will error out with:

```
[error] Nuxt module should be a function: undefined
```

The current workaround is:

```ts
  modules: [
    '@vueuse/nuxt'
    ...(process.env.CI ? [] : ['@nuxt/devtools']),
  ],
})
```

Which I think it's not idle.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -194,7 +194,7 @@ export default defineUntypedSchema({
    *   function () {}
    * ]
    * ```
-   * @type {(typeof import('../src/types/module').NuxtModule | string | [typeof import('../src/types/module').NuxtModule | string, Record<string, any>])[]}
+   * @type {(typeof import('../src/types/module').NuxtModule | string | [typeof import('../src/types/module').NuxtModule | string, Record<string, any>] | undefined | null | false)[]}
    */
   modules: [],
 
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -196,7 +196,9 @@ export default defineUntypedSchema({
    * ```
    * @type {(typeof import('../src/types/module').NuxtModule | string | [typeof import('../src/types/module').NuxtModule | string, Record<string, any>] | undefined | null | false)[]}
    */
-  modules: [],
+  modules: {
+    $resolve: val => [].concat(val).filter(Boolean)
+  },
 
   /**
    * Customize default directory structure used by Nuxt.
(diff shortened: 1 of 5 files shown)


test/fixtures/basic/nuxt.config.ts
```diff
@@ -135,7 +135,9 @@ export default defineNuxtConfig({
     function (_, nuxt) {
       nuxt.options.optimization.treeShake.composables.server[nuxt.options.rootDir] = ['useClientOnlyComposable', 'setTitleToPink']
       nuxt.options.optimization.treeShake.composables.client[nuxt.options.rootDir] = ['useServerOnlyComposable']
-    }
+    },
+    // To test falsy module values
+    undefined,
```
danielroe, line 140:
```suggestion
    undefined
```

packages/schema/src/config/common.ts
```diff
@@ -194,9 +194,11 @@ export default defineUntypedSchema({
    *   function () {}
    * ]
    * ```
-   * @type {(typeof import('../src/types/module').NuxtModule | string | [typeof import('../src/types/module').NuxtModule | string, Record<string, any>])[]}
+   * @type {(typeof import('../src/types/module').NuxtModule | string | [typeof import('../src/types/module').NuxtModule | string, Record<string, any>] | undefined | null | false)[]}
    */
-  modules: [],
+  modules: {
+    $resolve: val => [].concat(val).filter(Boolean)
```
antfu, line 200: It can also come from layers etc, right?

danielroe: I think this will handle layers too.

danielroe:
It might also be worth considering https://github.com/unjs/c12/pull/61 as an alternative approach to this.

Atinux:
They could actually do:

```ts
defineNuxtConfig({
  $development: {
     modules: ['@nuxt/devtools']
  },
  modules: ['@vueuse/nuxt']
})
```

In case `process.env.CI` is used during `nuxt build`

Also, you can also have:

```ts
defineNuxtConfig({
  modules: [
    '@vueuse/nuxt'
    '@nuxt/devtools',
  ],
  devtools: {
    enabled: !process.env.CI
  }
})
```

pi0:
Today, modules can self-disable based on environment that also introduced environment specific `$development` / `$production`.

Still i see value in supporting such filtering because it covers custom logic such as CI+Target or so.

On related topic, we might thinking about `disabledModules: []`. Today it is impossible to easily disable a module provided by a nuxt layer.

danielroe (approved):

PR #19697, opened 2023-03-15
docs: update seo and meta page

dosstx:
Add the `await` keyword before `useFetch`. useFetch is an asynchronous function that returns a Promise, so you need to use the await keyword to wait for the Promise to resolve before assigning its value to the post variable. This ensures that the data is available before it is used in the useServerSeoMeta function.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/1.getting-started/5.seo-meta.md
+++ b/docs/1.getting-started/5.seo-meta.md
@@ -82,22 +82,20 @@ useServerSeoMeta({
 
 **Reactive example:**
 
-When inserting tags that are reactive, for example, from an API request, you should use the computed getter syntax, the same as `useHead`.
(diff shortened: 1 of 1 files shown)


dosstx:
On second thought probably don't need to use AWAIT here since it's a computed getter.

Atinux:
Thank you @dosstx 

Actually I thought about it as well, but for the `useSeoMeta` example, I believe `useFetch` is not the right example.

Will update your PR to improve the example.

See my demo on https://stackblitz.com/edit/github-vlzrtd?file=app.vue

Atinux:
Preview link: https://nuxt.com/docs/getting-started/seo-meta?preview=83ceb1ab7b8ae2614a7d3b70bc1eea07

danielroe (approved):

PR #19738, opened 2023-03-16
fix(types): bring types from `less` into namespace

rchl:
### ❓ Type of change

### 📚 Description

The `Less` namespace is referenced in this file and the `@types/less` is one of the dependencies of `@nuxt/types` but `less` is not imported anywhere within the `@nuxt/types` package which causes the following error when running `tsc` in projects using `@nuxt/types`:

```
../../packages/types/config/build.d.ts:68:10 - error TS2503: Cannot find namespace 'Less'.

68   less?: Less.Options
            ~~~~
```

Projects can work around it by including `less` in `types` in `tsconfig.json` but they shouldn't need to.

### 📝 Checklist

### DIFF
--- a/packages/types/config/build.d.ts
+++ b/packages/types/config/build.d.ts
@@ -3,6 +3,7 @@
  * Documentation: https://nuxtjs.org/api/configuration-build
  */
 
+import 'less'
 import type { IncomingMessage, ServerResponse } from 'http'
 import type { TransformOptions, PluginItem } from '@babel/core'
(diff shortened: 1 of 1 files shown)


packages/types/config/build.d.ts
```diff
@@ -3,6 +3,7 @@
  * Documentation: https://nuxtjs.org/api/configuration-build
  */
 
+import 'less'
```
danielroe, line 6:
```suggestion
/// <reference types="less" />
```

danielroe (approved):

Issue #18615, opened 2023-01-30
Can't persist data between routes with useNuxtData() when using npm nuxi generate


idflood:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.13.0`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.1.0`
- Package Manager:  `npm@8.19.3`
- Builder:          `vite`
- User Config:      `ssr`, `experimental`, `modules`, `runtimeConfig`
- Runtime Modules:  `nuxt-graphql-client@0.2.25`
- Build Modules:    `-`

### Reproduction

1. Clone https://github.com/idflood/nuxt3-static-test
2. run `npm i`
3. run `npx nuxi generate`

For each routes it will console.log `no data found for key, loading again featuredpokemon`.

### Describe the bug

When generating a static website I would like to be able to load some static data only once so that it can be shared across pages. This can be useful for global settings, menu, … And loading these once could considerably decrease build time.

In the [example provided](https://github.com/idflood/nuxt3-static-test), I've added a <[FeaturedPokemon](https://github.com/idflood/nuxt3-static-test/blob/main/components/FeaturedPokemon.vue)> component in the default layout. In that component I use a [useCachedData](https://github.com/idflood/nuxt3-static-test/blob/main/composables/useCachedAsyncData.ts) composable which make use of `useNuxtData`.

Then, running `npx nuxi generate` I see that it fetch this static data for every routes instead of reusing the data.

Note that in the useCachedAsyncData composable if I try to add a `useNuxtApp()` and log the values of `nuxt.payload` and `nuxt.static` they both always return empty content. I've also tried to set directly the `nuxt.static.value` with some data but it seems to have the same behavior as payload.

### Additional context

For reference, the global data storage is also being discussed here: https://github.com/nuxt/nuxt/discussions/16479

danielroe:
`useNuxtData` isn't meant for persisting data between requests, but only within a single request.

I do think it's worth improving the generation experience, particularly when there's data you fetch once but need many times. There are a number of different ways of doing this already within Nuxt 3, so I'll have a think and get back to you with an example.

idflood:
I think I've found one way to pass the shared payload between routes. 

@danielroe Was this what you had in mind?

```javascript
// plugins/cache-static-data.ts
const cachedData = {}
const dataToCache = ['featuredpokemon']

export default defineNuxtPlugin((nuxtApp) => {
[...]

PR #19753, opened 2023-03-16
fix(nuxt): unset context after app is created

### LINKED ISSUE #14719: automatically call nuxt hooks with implicit nuxt context
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.16.0`
- Nuxt Version:     `3.0.0-rc.8`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-t5twtq-b3tflp?file=plugin%2Fserver.server.ts,nuxt.config.ts

500
nuxt instance unavailable

### Describe the bug

useNuxtApp doesn't work in lifecycle hooks on server side

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #19623: cross-request state pollution when `useAsyncData` is used in a plugin
### Environment

- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Package Manager:  `pnpm@7.29.1`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
-----------------------------

### Reproduction

https://github.com/huang-julien/nuxt/tree/test/xrps-server-component 
This repo contains the content of https://github.com/nuxt/nuxt/pull/19605
Run the fixtures in dev and go to `/islands`
Refresh 5-6 time, you should see that `PureComponent` get replaced by the result of `AsyncServerComponent`

You can also do it at `/` but you'd need to leave your finger on F5 for 2-3sec.

Check the console. There's 2 value, they should always be the same. The first one is an array length, the second one a Set size.

### Describe the bug

When there's a very  high rate of request with sync component server side (just leave you're finger on F5), sometimes the app instance retrieved from unctx with `useNuxtApp` does not belong to the current instance of vue App.

It seems that this does not apply to pages rendering but only when requesting for server components.

### Additional context

Found when working on https://github.com/nuxt/nuxt/pull/19605. It is way easier to reproduce it with async components

discussed with @danielroe 

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19623
resolves https://github.com/nuxt/nuxt/issues/14719

### ❓ Type of change

### 📚 Description

We can unset the nuxt app context after initialising it and calling each hook. It should be safe, and appears to resolve the linked reproduction.

This calls `app:created` and `app:error` (the only two server Nuxt hooks) with implicit Nuxt context, resolving https://github.com/nuxt/nuxt/issues/14719, though the normal caveats apply (the context will be lost within a hook after the first await).

### 📝 Checklist


packages/nuxt/src/app/entry.ts
```diff
@@ -30,11 +31,19 @@ if (process.server) {
 
     const nuxt = createNuxtApp({ vueApp, ssrContext })
 
+    async function contextCaller (hooks: HookCallback[], args: any[]) {
+      for (const hook of hooks) {
+        nuxtAppCtx.set(nuxt)
+        await hook(...args)
```
pi0, line 37: I think we might need to use `ctx.call` (otherwise same concurrent ctx sharing problem can happen during promise resolution)

danielroe: okay, using `.call`.

packages/nuxt/src/app/entry.ts
```diff
@@ -30,11 +31,17 @@ if (process.server) {
...
+    async function contextCaller (hooks: HookCallback[], args: any[]) {
+      for (const hook of hooks) {
+        await nuxtAppCtx.call(nuxt, () => hook(...args))
+      }
+    }
+
     try {
       await applyPlugins(nuxt, plugins)
-      await nuxt.hooks.callHook('app:created', vueApp)
+      await nuxt.hooks.callHookWith(contextCaller, 'app:created', vueApp)
```
pi0, line 42: Something else i'm wondering is that we probably only need this for server-side only. It can be a breaking change for client-side as there, we have context always available assuming there is one nuxt instance. 

Also we could directly patch `nuxt.callHook` with contextCaller wdyt?

danielroe: Yes, this change only affects the server side.

As for patching, that's a nice idea. Will update.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -187,6 +187,17 @@ export function createNuxtApp (options: CreateOptions) {
...
   nuxtApp.hook = nuxtApp.hooks.hook
+
+  if (process.server) {
+    async function contextCaller (hooks: HookCallback[], args: any[]) {
+      for (const hook of hooks) {
+        await nuxtAppCtx.call(nuxtApp, () => hook(...args))
+      }
+    }
+    // Patch callHook to preserve NuxtApp context on server
+    nuxtApp.hooks.callHook = (name: any, ...args: any[]) => nuxtApp.hooks.callHookWith(contextCaller, name, ...args)
```
pi0, line 198: For followup, we could support `setCaller` for hookable library https://github.com/unjs/hookable/issues/74

pi0 (approved):

packages/nuxt/src/app/nuxt.ts
pi0:
```suggestion
    // TODO: Refactor after https://github.com/unjs/hookable/issues/74
    nuxtApp.hooks.callHook = (name: any, ...args: any[]) => nuxtApp.hooks.callHookWith(contextCaller, name, ...args)
```

idflood:
@danielroe the changes with the new `if (process.server)` condition here "broke" my current [data persisting solution](https://github.com/nuxt/nuxt/issues/18615#issuecomment-1413693370) given in #18615 .

I don't know if there is something to be done here, or if there is now a better way to persist data. For now I will revert to 3.3.1 but I'm curious to know if you have any suggestion to implement this kind of global shared data for nuxt generate.

danielroe:
@idflood `app:created` should still be called. The issue you're experiencing is a regression which should be resolved in https://github.com/nuxt/nuxt/pull/19969 or possibly sooner in https://github.com/unjs/hookable/pull/79.

idflood:
@danielroe Thanks a lot, you are awesome 👍 I confirm #19969 fixes this issue : )

danielroe:
@idflood This should now be resolved with the latest version of `hookable` via https://github.com/unjs/hookable/pull/79. You should be able to get it by explicitly installing the latest version or by refreshing your lockfile.

Let me know if not.

Issue #10751, opened 2022-10-09
Build fails with `Unexpected character #` in `node-fetch-native`


jmsfilipe:
### Versions

- nuxt-edge: 2.16.0-27720022.54e852f
- node: v16.17.0

### Reproduction

<details open>
<summary>Additional Details</summary>
<br>

</details>

### Steps to reproduce
- Use latest nuxt-edge version
- `nuxt.config.js` with `target: 'static'` and `ssr: true`
- Running nuxt build fails with
```
ERROR in ./node_modules/node-fetch-native/dist/index.cjs 69:2
Module parse failed: Unexpected character '#' (69:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|  */
| class BlobDataItem {
>   #path
|   #start
| 
 @ ./node_modules/node-fetch-native/lib/index.cjs 1:18-46
 @ ./.nuxt/server.js
 @ multi ./node_modules/@nuxt/components/lib/installComponents.js ./.nuxt/server.js
```

### What is Expected?
- Running nuxt build should be successful

### What is actually happening?
- Seems like the `node-fetch-native` dependency used to be just `node-fetch`
- This new dependency includes [private methods syntax](https://babeljs.io/docs/en/babel-plugin-proposal-private-methods) and webpack/eslint don't know how to deal with that? Maybe it needs to be transpiled?

Billybobbonnet:
I have the exact same issue with regular Nuxt v2 (without the nuxt-edge package), same config.  I can't generate anymore.

sanderjn:
I'm running into the same problem. Though I'm not using `target: 'static'` and `ssr: true`.

Local `dev` and `build` commands work just fine, but I'm getting this error just at Vercel.

danielroe:
This seems an issue with webpack's support for class properties in node-fetch-native. We likely need to configure this in Nuxt 2.

cc: @pi0

sanderjn:
Any workaround for now?

Billybobbonnet:
> Any workaround for now?

Pin your Nuxt version to `"nuxt": "2.15.7"`

jbmarflo:
I cant solve this issue yet, Can anybody solve this error? I'm tried with babel

pimlie:
You could also add a alias from node-fetch-native back to node-fetch when using `v2.16.x`:
```js
// nuxt.config.js

  build: {
    extend(config) {
[...]

PR #19755, opened 2023-03-16
fix(webpack): add `node-fetch-native` to externals list

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/10751

### ❓ Type of change

### 📚 Description

This adds `node-fetch-native` to the externals list when building. It also marks cjs and mjs as allowable native file extensions to allow avoiding bundling these files.

### 📝 Checklist

### DIFF
--- a/packages/webpack/src/config/server.js
+++ b/packages/webpack/src/config/server.js
@@ -9,6 +9,8 @@ import WebpackBaseConfig from './base'
 
 const nativeFileExtensions = [
   '.json',
+  '.cjs',
+  '.mjs',
   '.js'
 ]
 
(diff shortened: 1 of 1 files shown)


packages/webpack/src/config/server.js
```diff
@@ -9,6 +9,8 @@ import WebpackBaseConfig from './base'
 
 const nativeFileExtensions = [
   '.json',
+  '.cjs',
+  '.mjs',
```
pi0, line 13: perfect!

pi0 (approved):
LGTM if already tested

packages/webpack/src/config/server.js
```diff
@@ -123,7 +125,7 @@ export default class WebpackServerConfig extends WebpackBaseConfig {
         maxEntrypointSize: Infinity,
         maxAssetSize: Infinity
       },
-      externals: [].concat(config.externals || [])
+      externals: ['node-fetch-native'].concat(config.externals || [])
```
pi0, line 128: Is it still essential with mjs fix?

danielroe: This is still required, yes, as it is not called with the full path here but just the specifier (`node-fetch-native`).

PR #19805, opened 2023-03-19
feat(nuxt): add `experimentalNoScripts` route rule

huang-julien:
### 🔗 Linked issue
Hi :wave: this PR adds `noScripts` option to `routeRules`

### ❓ Type of change

### 📝 Checklist and todo

- [x] I have updated the documentation accordingly -> udpate `routeRules` section of the docs


danielroe (approved):

packages/schema/src/config/nitro.ts
```diff
@@ -19,6 +19,10 @@ export default defineUntypedSchema({
 
   /**
    * Global route options applied to matching server routes.
+   * 
+   * It extends Nitro's `routeRules` with additionnal properties:
+   * - `routeRules`: `**boolean**`
+   * - `ssr`: `**boolean**`
```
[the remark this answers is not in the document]
huang-julien: Should i revert this ? 🤔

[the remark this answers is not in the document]
danielroe: Maybe. But it would be good to document this somewhere, maybe in https://nuxt.com/docs/guide/concepts/rendering#route-rules.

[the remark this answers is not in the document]
huang-julien: Indeed, i completly forgot this part

packages/nuxt/types.d.ts
```diff
@@ -11,8 +11,10 @@ declare global {
 declare module 'nitropack' {
   interface NitroRouteConfig {
     ssr?: boolean
+    noScripts?: boolean
```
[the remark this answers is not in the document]
huang-julien: I guess this would make sense :+1:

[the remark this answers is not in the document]
danielroe: I think the experimental part of `experimental.noScripts` is turning them off for the entire app. But I think this route rule makes sense to be non-experimental, as it is scoped by definition.

danielroe (changes requested):
I think we also would need to prevent rendering resource hints/body scripts/etc. With `experimental.noScripts` we disable these by means of modifying the build manifest. But we would need a different strategy to disable them per-route.

(My apologies for not spotting this before.)

packages/nuxt/src/pages/module.ts
```diff
@@ -3,8 +3,10 @@ import { addComponent, addPlugin, addTemplate, addVitePlugin, addWebpackPlugin,
 import { join, relative, resolve } from 'pathe'
 import { genImport, genObjectFromRawEntries, genString } from 'knitwork'
 import escapeRE from 'escape-string-regexp'
-import { joinURL } from 'ufo'
+import { joinURL, withoutLeadingSlash } from 'ufo'
 import type { NuxtApp, NuxtPage } from 'nuxt/schema'
+import micromatch from 'micromatch'
```
danielroe, line 8: what do you think about using the nitro route matcher, for 1:1 parity?

https://github.com/unjs/nitro/blob/8e06f2e4a65c6de0d2f2f3a5ef156ae3821ed60a/src/prerender.ts#L70-L75

huang-julien: sure 🙂 :+1:

huang-julien: like this ? 🤔

packages/nuxt/src/pages/module.ts
```diff
@@ -186,6 +190,32 @@ export default defineNuxtModule({
...
+      for (const page of pages) {
+        const rules = getRouteRules(page.path)
+        if (rules.noScripts && page.file && !pages.some(p => p.file === page.file && p !== page)) {
+          routes.push(relative(nuxt.options.srcDir, page.file))
+        }
+      }
+
+      for (const file of routes) {
+        if (routes.some(fileName => manifest[fileName] && manifest[file].resourceType === 'script')) {
+          await fse.rm(resolve(nuxt.options.buildDir, 'dist/client', withoutLeadingSlash(nuxt.options.app.buildAssetsDir), manifest[file].file), { force: true })
```
danielroe, line 213: We probably don't want to remove script entirely (or we will break client-side navigation).

```suggestion
```

packages/nuxt/src/pages/module.ts
```diff
@@ -186,6 +190,32 @@ export default defineNuxtModule({
...
+
+      for (const page of pages) {
+        const rules = getRouteRules(page.path)
+        if (rules.noScripts && page.file && !pages.some(p => p.file === page.file && p !== page)) {
+          routes.push(relative(nuxt.options.srcDir, page.file))
+        }
+      }
+
+      for (const file of routes) {
+        if (routes.some(fileName => manifest[fileName] && manifest[file].resourceType === 'script')) {
```
danielroe, line 212: I think we can simplify to:
```suggestion
        if (manifest[file].resourceType !== 'script') { continue }
        if (routes.some(fileName => fileName in manifest)) {
```

danielroe:
In your testing, does this remove scripts for designated routes? It seems still to be injecting entry + error scripts for me, even if not the page script.

huang-julien:
ah yes my bad, i'll add some tests to make sure this is correctly working.

packages/nuxt/src/pages/module.ts
```diff
@@ -186,6 +189,32 @@ export default defineNuxtModule({
...
+        }
+      }
+
+      for (const file of routes) {
+        if (!manifest[file] || manifest[file].resourceType !== 'script') { continue }
+        if (routes.some(fileName => fileName in manifest)) {
+          manifest[file].file = ''
+        }
+      }
+    })
```
danielroe, line 216: What are you thinking about this section - do we still need it? I think we have two options:

1. `noScripts` means there are no scripts corresponding to given routes _at all_ - so navigating to them is a MPA navigation (hard reload). This is tricky as vite may well try to preload these (triggering chunk errors if we remove the files), and we probably have to write both a runtime manifest + vue router integration to achieve this. (Or maybe we can use a nitro transform plugin for the final build, which would also mean we can use rollup to tree-shake out other dependencies from the build.)
2. `noScripts` simply refers to the _initial_ load from the server

I think the simplest solution for now is to choose the latter option, at least to start with. But we should likely make this behaviour clear in the documentation, and maybe also mark this as experimental. (Maybe by naming it `experimentalNoScripts: true`?) What do you think @pi0 - is there a better way to mark a route rule experimental?

huang-julien: Yes I think starting with the latter is a good start. Maybe should I have open a discussion about it before doing this 😅 ...
So client side navigation would still trigger a SPA behavior.
I guess we can remove this code ?

pi0: `experimentalNoScripts` seems a nice idea!

danielroe (approved):

nikodunk:
We just started using this on fedoraproject.org and all its sub-pages, thank you for this feature @danielroe! Most of the pages are full-static, but one or two require actual client-side hydration, so this is the perfect feature for us. Thank you! https://gitlab.com/fedora/websites-apps/fedora-websites/fedora-websites-3.0/-/blob/develop/nuxt.config.ts#L87-92

Issue #14669, opened 2022-08-20
allow skip building client and only emit nitro server


josvanos:
### Describe the feature

This feature is about deploying a nuxt application. In my case I use Cloudflare Pages for the application side, while the api (server) is placed at standalone node js server.

Every time I run `nuxi build`, i don't need the public directory, however that is the part that takes the longest. 

I would like to request a flag as `--skip=public` or a command that does the exact opposite from 'nuxi generate`, that just creates the server side only. 

 

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

Atinux:
What about:

```ts
export default defineNuxtConfig({
  build: {
    client: true
  }
})
```

But the thing is that we may want to also update the `defineRenderHandler`.

We could also thing of a `nuxi build --server-only` command 🤔

sroynano05:
Hi! I'd love to work on this since it's marked as a 'good first issue'. Before I dive into the codebase, has a final decision been made on whether this should be implemented as a config rule (like `build.client`) or a new CLI flag (like --server-only)? Thanks!

PR #19825, opened 2023-03-20
perf(nuxt): experimentally disable vue server renderer nitro endpoint

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/14669

### ❓ Type of change

### 📚 Description

This adds an experimental option to disable the vue server renderer endpoint.

The effect of this is substantial; it takes us from 2.79 Mb of server (uncompressed) down to 506 Kb in our minimal test fixture. For a site where all the content is prerendered, this is a significant improvement and should make deploying pure API + static HTML much more performant (e.g. using edge functions).

We might consider enabling this in certain conditions:

- no `router.options.ts` files (which might add dynamic routes)
- no custom `error.vue` (or support prerendering these?)
- no dynamic pages
- all fixed pages were prerendered

~~As written, this PR has preliminary support (not switched on by default) for automatically disabling this endpoint when all vue pages have been prerendered, but it does not fully cover the requirements above.~~

### 📝 Checklist


packages/nuxt/src/core/nitro.ts
```diff
@@ -232,6 +232,16 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
     handler: resolve(distDir, 'core/runtime/nitro/renderer')
   })
 
+  if (nuxt.options.experimental.noVueServer) {
+    nitro.hooks.hook('rollup:before', (nitro) => {
+      if (nitro.options.preset === 'nitro-prerender') { return }
+      nitro.options.handlers.shift()
```
pi0, line 238: This seems little bit unsafe workaround.

I think we might support `experimental.noRenderer` directly to nitro handling this case. (and disable the flag with `prerender`

danielroe: As this is an experimental option, what do you think about merging this as is, and then switch over to `noRenderer` once nitro supports it?

pi0: Looks good to me.

pi0 (approved):

atinux (approved):

PR #19830, opened 2023-03-20
fix(kit): check if nuxt is restarting before updating templates

danielroe:
### ❓ Type of change

### 📚 Description

I've noticed a couple of times when restarting Nuxt, the templates are still being updated when the server restarts, and I get an error associated with the nuxt instance not being defined.

### 📝 Checklist


packages/kit/src/template.ts
```diff
@@ -78,5 +78,5 @@ export function normalizeTemplate (template: NuxtTemplate<any> | string): Resolv
  * You can pass a filter within the options to selectively regenerate a subset of templates.
  */
 export function updateTemplates (options?: { filter?: (template: ResolvedNuxtTemplate<any>) => boolean }) {
-  return useNuxt().hooks.callHook('builder:generateApp', options)
+  return tryUseNuxt()?.hooks.callHook('builder:generateApp', options)
```
pi0, line 81: Previously `callHook` always returned a Promise. While this is small, i think would be nice if we use `async/await` to make sure return type is promise (`?.` leads to `undefined`)

pi0 (approved):
Other than small comment LGTM 💯

packages/kit/src/template.ts
```diff
@@ -78,5 +78,5 @@ export function normalizeTemplate (template: NuxtTemplate<any> | string): Resolv
  * You can pass a filter within the options to selectively regenerate a subset of templates.
  */
 export function updateTemplates (options?: { filter?: (template: ResolvedNuxtTemplate<any>) => boolean }) {
```
danielroe, line 80:
```suggestion
export async function updateTemplates (options?: { filter?: (template: ResolvedNuxtTemplate<any>) => boolean }) {
```

packages/kit/src/template.ts
danielroe, line 81:
```suggestion
  return await tryUseNuxt()?.hooks.callHook('builder:generateApp', options)
```

PR #19851, opened 2023-03-21
feat(nuxt): support basic slots on server only components

### LINKED ISSUE #19765: support client interactivity within server components
### Describe the feature

Right now, you can't have client components as children of server components, but enabling this would enable a bunch of usful usecases.

for example I have a big website using nuxt, with mostly static content but also bit of interactive content (like accordions or toggles to hide and show content).
Server components would allow me to not send the JS for most of my site to the client, greatly reducing time spend on script parsing and execution (hydration), but right now that would disallow me from using any interactive elements.
(for reference, next.js supports this: https://beta.nextjs.org/docs/rendering/server-and-client-components#moving-client-components-to-the-leaves)

I would imagine that the easiest way would be to enable server components to have slots and in the client somehow check if there is a client component slot present in the server component html.
Maybe NuxtServerComponent needs to parse the html to find out where to inject the slots or we can store that info in the island response.

I have a bunch of work time next week to dedicate to working on this, but I dont really know where to start, if anybody has any pointers where I can look or what the best approach would be, that would be appreciated, I am MiniDigger#3086 on discord.

### Additional information

### Final checks

### LINKED ISSUE #19884: Islands components not updating in production
### Environment

All 

### Reproduction

npm run build 
node .output/server/index.mjs

Click on  `add + 1`

the count is not updated

### Describe the bug

### Additional context

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/19765

### ❓ Type of change

### 📚 Description
Hi :wave: 

This PR is currently in progress (draft to keep tracking my progress on it)
- [x] PR done and ready

This PR allow to have slots with server only components.

Slots also works in SSR for pages !

Currently the only way to simulate a slot is to render a container tag, however it needs a `display: contents` which seems just a bit buggy atm in some cases but we still need it to ignore the container tag box, maybe will it be fixed one day. 
see: https://caniuse.com/css-display-contents

### How does this work ?
This introduce changes to `<NuxtIsland>`, `<IslandRenderer>`, server components and to the nitro renderer./

- server only components:
  - it now becomes a wrapper to call `NuxtIsland`
- `<NuxtIsland>`
  - each calls to the `islandRenderer` are now unique
  - it bypass its hydration
  - teleport all its slots to the slot containers rendered by `IslandRenderer`
- `IslandRenderer`
  - render the component with an UID in its attributes
  - now receive a list of slots (only names) to render a container tag for `NuxtIsland` to teleport its slots

### 📝 Checklist and TODO

related to the branch
- [x] clean the messy code

related to PR

- [x] fix prod/build
- [x] use uncrypto
- [x] improve perf for SSR renderer (don't try to replace if there's no v-ssr-component-uid or v-ssr-slot-name)
- [x] I have updated the documentation 
- [x] update docs about limitations (container tag for now, slots can't be conditional, scoped slots not available yet)
- [x] complete fixtures tests

fixtures 
- [x] test nuxt islands
- [x] test .server
- [x] test slot interactivity
- [x] test slot with server components mounted Post SSR


huang-julien:
ATM this PR is only working in dev mode.

However https://github.com/nuxt/nuxt/pull/19911 which fixes the server only components re-rendering (by using a key to force the static vnode to re-render) introduce another issue.
Since this PR use `Teleport`, re-rendering the static slot makes all slots disapear, most probably because `Teleport` are still being attached to the old static node. if we try to force re-render the Teleport, this won't work either because at the moment vue tries to `Teleport` the target node is not available yet in the DOM

[The current Implementation](https://beta.nextjs.org/docs/rendering/server-and-client-components#when-to-use-server-vs-client-components) of NextJS (did not dig too much into it) does not allow client-side interactivity. 
Maybe should we do the same thing ? This would allow to have a mix of non interactive static nodes AND fully insteractive slots.

Or maybe using a `allow-slot` prop that would enable the usage of slots (but disabling re-rendering the server only component) ?

MiniDigger:
this is absolutely awesome! 🎉

played around with the branch a bit today and it works as I would expect it to work.
one current limitation (that I dont care about rn but should prolly be documented, or maybe its just a bug?) is that slot props dont seem to work for me (the slot prop is just undefined)
similarly, conditional slots (v-if on the slot) dont seem to work (something breaks with the teleport)
for reference, here is the code: https://paste.gg/p/anonymous/c516b3a47b564a088c1f3e9911e193b8

but since server components are experimental anyways, this is a major improvement and such limitations can be addressed incrementally in the future

huang-julien:
Thank you ! 
In the first commits, slots containers were generated by using the `__asyncResolved` property which is a vue internal API (and it seems to be only working in dev), since we should avoid using vue internals the solution was to send the declared slots when updating the component. 

About scoped slots, i think this should be doable. I'll do something about it in the incoming week (won't have a lot of time) with another PR 

I'll update the docs about theses limitations.

huang-julien:
I feel like this PR can be closed since https://github.com/nuxt/nuxt/pull/20284 includes and goes beyond this PR for slot support (since it is forked from this one)

Issue #19852, opened 2023-03-22
Nuxt Layers: Cannot find module '@nuxt/kit'


warflash:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.3.1`
- Nitro Version:    `2.3.1`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `extends`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/warflash/nuxt-layers-base

https://github.com/warflash/nuxt-layers-parent

### Describe the bug

I'm trying to follow the docs here: https://nuxt.com/docs/guide/going-further/layers#relative-paths-and-aliases on how to import css in the base nuxt layer. When deploying the app that extends that layer to production it fails throwing `[error] Cannot find module '@nuxt/kit'`. 

### Additional context

I tried specifically installing nuxtkit as a dependency in both the base and parent repo but that didn't seem to help unfortunately

### Logs

```shell-script
[03:56:22.663] Running build in Washington, D.C., USA (East) – iad1
[03:56:22.715] Cloning github.com/warflash/nuxt-layers-parent (Branch: main, Commit: de720c0)
[03:56:22.917] Cloning completed: 202.097ms
[03:56:23.013] Previous build cache not available
[03:56:23.064] Running "vercel build"
[03:56:23.506] Vercel CLI 28.17.0
[03:56:24.074] Installing dependencies...
[03:56:27.455] npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
[03:56:27.502] npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
[03:56:31.215] 
[03:56:31.215] > postinstall
[03:56:31.215] > nuxt prepare
[03:56:31.216] 
[03:56:31.304] [log] Nuxi 3.3.1
[03:56:32.443] [error] Cannot find module '@nuxt/kit'
[03:56:32.443] Require stack:
[03:56:32.443] - /vercel/.cache/c12/github_warflash_nuxt-layers-base/nuxt.config.ts
[03:56:32.443]   Require stack:
[...]

PR #19873, opened 2023-03-23
docs: replace `@nuxt/kit` example with node built-ins

warflash:
### 🔗 Linked issue

#19852 

### ❓ Type of change

### 📚 Description

Currently the docs recommned using @nuxt/kit for path resolution which does not work.
To prevent confusion this PR adds a working path resolution example using node built-ins.

### 📝 Checklist

### DIFF
--- a/docs/2.guide/3.going-further/7.layers.md
+++ b/docs/2.guide/3.going-further/7.layers.md
@@ -163,11 +163,11 @@ Also when using relative paths in `nuxt.config` file of a layer, (with exception
 import { fileURLToPath } from 'url'
 import { dirname, join } from 'path'
 
-const __dirname = dirname(fileURLToPath(import.meta.url))
+const currentDir = dirname(fileURLToPath(import.meta.url))
 
 export default defineNuxtConfig({
   css: [
-    join(__dirname, './assets/main.css')
+    join(currentDir, './assets/main.css')
   ]
 })
 ```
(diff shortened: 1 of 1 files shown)


docs/2.guide/3.going-further/7.layers.md
```diff
@@ -160,13 +160,14 @@ When importing using aliases (such as `~/` and `@/`) in a layer components and c
 Also when using relative paths in `nuxt.config` file of a layer, (with exception of nested `extends`) they are resolved relative to user's project instead of the layer. As a workaround, use full resolved paths in `nuxt.config`:
 
 ```js [nuxt.config.ts]
-import { createResolver } from '@nuxt/kit'
+import { fileURLToPath } from 'url'
+import { dirname, join } from 'path'
 
-const { resolve } = createResolver(import.meta.url)
+const __dirname = dirname(fileURLToPath(import.meta.url))
```
danielroe, line 166: Might not matter here, but I would advise avoiding `__dirname` as it's sometimes used to detect a CJS environment.

pi0: I agree that we might encourage all ESM syntax with `impoort.meta.url`

BTW jiti probably will always support CJS (`require.resolve` is even available in `nuxt.config` and works perfectly fine...) it is only and only matter of consistency to not use this simpler syntax..

danielroe: I mean, avoid the variable name `__dirname` as a matter of good practice.

pi0: I understand and it is a nice practice to stick with ESM syntax.

Only mentioning it _is_ a cjs environment by jiti for loading config. there should be no issues with it.

danielroe (approved):

Issue #19717, opened 2023-03-15
v3.3.1 breaks v2.16


MooseSaeed:
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.13.1`
- Nuxt Version:     `2.16.2`
- Package Manager:  `npm@9.6.1`
- Builder:          `webpack`

-----

After the recent updates to `@nuxt/kit` and `@nuxt/schema` in nuxt v3.31, any nuxt app running on v2.16.* will have a problem if we run `npm install` without having package-lock.json. 

When running `npm run dev` or `npm run build`:

```
 ✖ Nuxt Fatal Error 
   │   Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in                             │
   │   /Users/said/Desktop/Mostafa/Gfinity/repos/stealthoptional/node_modules/@nuxt/kit/package.json 
```

This was solved when I replaced `@nuxt/kit@3.3.1` in package-lock.json with `@nuxt/kit@3.2.3` and `@nuxt/schema@3.3.1` with `@nuxt/schema@3.2.3`

### Reproduction

Reproducing the issue is easy:
1- Nuxt v2.16.2 project without package-lock.json file.
2- npm install
3- npm run dev or build

### Describe the bug

After the recent updates to `@nuxt/kit` and `@nuxt/schema` in nuxt v3.31, any nuxt app running on v2.16.* will have a problem if we run `npm install` without having package-lock.json. 

When running `npm run dev` or `npm run build`:

```
 ✖ Nuxt Fatal Error 
   │   Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in                             │
   │   /Users/said/Desktop/Mostafa/Gfinity/repos/stealthoptional/node_modules/@nuxt/kit/package.json 
```

danielroe:
Would you provide a [reproduction](https://nuxt.com/docs/community/reporting-bugs/#create-a-minimal-reproduction)? 🙏

MartinX3:
@danielroe 
```
"overrides": {
    "unplugin": "1.1.0"
}
```
fixes it

MooseSaeed:
> reproduction

I'm still getting the error even after adding the mentioned code to package.json

```
 FATAL  No "exports" main defined in /Users/said/Desktop/Mostafa/site/repos/site/node_modules/@nuxt/kit/package.json                  13:44:12

  at new NodeError (node:internal/errors:371:5)
[...]

PR #19874, opened 2023-03-23
fix(kit): add legacy entrypoints for pre v3.3 usage

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/19717

### ❓ Type of change

### 📚 Description

In various cases users on projects pre-3.3 who are consuming a 3.3 kit may receive an error message (because previously we were using cjs resolve utilities to resolve kit.

For example: https://github.com/nuxt/nuxt.new/pull/55.

This adds backward compatibility for those previous projects.

### 📝 Checklist

### DIFF
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -4,13 +4,12 @@
   "repository": "nuxt/nuxt",
   "license": "MIT",
   "type": "module",
-  "main": "./dist/index.mjs",
-  "module": "./dist/index.mjs",
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.mjs"
     },
     "./package.json": "./package.json"
   },
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -4,6 +4,8 @@
   "repository": "nuxt/nuxt",
   "license": "MIT",
   "type": "module",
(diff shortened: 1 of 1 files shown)


packages/kit/package.json
```diff
@@ -4,6 +4,8 @@
   "repository": "nuxt/nuxt",
   "license": "MIT",
   "type": "module",
+  "main": "./dist/index.mjs",
+  "module": "./dist/index.mjs",
```
pi0, line 8: Module field is only picked by bundlers are you sure we need it?

danielroe: We might get away with just main 👍

danielroe: Oh dear. It's worse than that. We actually need a `require` export condition.

packages/kit/package.json
```diff
@@ -8,7 +8,8 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.mjs"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.mjs"
```
pi0, line 12: This is unsafe especially that i constantly see modules wrongly build with `.cjs` in dist and actually do `require('@nuxt/kit'`) but seems we don't have any choice at least until patched versions of nuxt 3 are widely adapted...

pi0 (approved):

packages/kit/package.json
danielroe: oh, this is terrible, I agree. But I think `require` shouldn't work in any case, as we have `type: "module"`

PR #19911, opened 2023-03-24
fix(nuxt): use key to force server component re-rendering

### LINKED ISSUE #19884: Islands components not updating in production

huang-julien:
### 🔗 Linked issue

### ❓ Type of change

### 📚 Description
hi :wave: 

This PR fix https://github.com/nuxt/nuxt/issues/19884 by wrapping the staticNode within a functionnal component and adding a key to it to force re-render the staticVNode
It also use `NuxtIsland` for `.server` component

### 📝 Checklist and TODO

- [x] Implement Tests fixtures for server components re-rendering


huang-julien:
Before merging this please check https://github.com/nuxt/nuxt/pull/19851#issuecomment-1484244890. This PR introduce another issue on #19851 :/

huang-julien:
https://github.com/nuxt/nuxt/pull/19851 includes this PR

huang-julien:
reopen in case of a minor fix patch

packages/nuxt/src/components/runtime/server-component.ts
```diff
@@ -1,102 +1,15 @@
...
-
-    const res = useAsyncData(
-      `${props.name}:${hashId.value}`,
-      async () => {
-        nuxtApp[pKey] = nuxtApp[pKey] || {}
-        if (!nuxtApp[pKey][hashId.value]) {
-          nuxtApp[pKey][hashId.value] = _fetchComponent().finally(() => {
-            delete nuxtApp[pKey]![hashId.value]
-          })
-        }
```
danielroe, line 72: The main reason there was a separate `NuxtServerComponent` component is to allow us to include its content in the payload so we can prefetch it. This isn't ideal as it inflates the payload, but maybe we could instead add the uid into the payload and also add a plugin to prefetch any prerendered server components?

huang-julien: I don't really get why do we need to prefetch it since the HTML part wil be already loaded 🤔 . Did i miss something about it ?

danielroe: I mean, for subsequent navigation. We ideally don't want to click a link and _then_ download html for a server component on another page. Just as we have prefetched the payload in advance, we can prefetch used server components in advance.

huang-julien: i see i'll set back NuxtServerComponent with the fix

danielroe (approved):

PR #19934, opened 2023-03-27
fix(nuxt): don't override options signature with schema

danielroe:
### ❓ Type of change

### 📚 Description

A regression from introducing schema was overriding the basic signature of the `NuxtOptions` interface when there was no defined schema.

However, I would also consider fixing upstream in untyped - or maybe defaulting to `[key: string]: unknown`? (cc: @pi0)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/schema.ts
+++ b/packages/nuxt/src/core/schema.ts
@@ -129,7 +129,8 @@ export default defineNuxtModule({
       const _types = generateTypes(schema, {
         addExport: true,
         interfaceName: 'NuxtCustomSchema',
-        partial: true
+        partial: true,
+        allowExtraKeys: false
       })
       const types =
         _types +


pi0:
Defaulting to `: unkown` seems a nice idea for me 👍🏼

manniL:
> Defaulting to `: unkown` seems a nice idea for me 👍🏼

Was thinking about the same 👍🏻

danielroe:
Great - let's default to `[key: string]: unknown` upstream.

But I think we also do need to merge this for consistency.

1. Without this option, a single schema property generates an interface _without_ a signature, but if there are no defined properties there there _is_ a signature. That's inconsistent no matter what the default signature is, I think.

2. Moreover, I don't think we want to overriding NuxtOptions/NuxtConfig key signatures, or run the risk of incompatible ones.

So I would feel much safer omitting the signature entirely in Nuxt itself (unless there's something I'm missing that we need it for - @pi0?)

pi0 (approved):

PR #19969, opened 2023-03-28
fix(nuxt): avoid hooks being mutated whilst called

### LINKED ISSUE #19967: Second variable of `useCookie` doesn't get set on the browser when used inside of a plugin
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.3.2`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

[StackBlitz Repro](https://stackblitz.com/edit/github-bzt66v?file=plugins%2Fcookies.ts)

### Describe the bug

When trying to set 2 or more unique variables of `useCookie`, the second one doesn't get set on the client.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19967

### ❓ Type of change

### 📚 Description

With a custom context caller, we were in danger of the hooks array being mutated (by single-use hooks registered with `hookOnce`) and thereby accidentally skipping some hooks.

We might consider also iterating with a different array method, changing behaviour upstream in hookable or just documentating that the hook array is mutable - wdyt @pi0?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -190,7 +190,7 @@ export function createNuxtApp (options: CreateOptions) {
 
   if (process.server) {
     async function contextCaller (hooks: HookCallback[], args: any[]) {
-      for (const hook of hooks) {
+      for (const hook of [...hooks]) {
         await nuxtAppCtx.call(nuxtApp, () => hook(...args))
       }
     }


pi0:
Nice find!

We might indeed directly patch [callEachWith](https://github.com/unjs/hookable/blob/aff2fa515bd0a7520ac0e547652c577a85eb1a4f/src/utils.ts#L100) to copy the array before starting the call.

Can you also directly reproduce it within hookable tests?

danielroe:
This should now be resolved with the latest version of `hookable` via https://github.com/unjs/hookable/pull/79. You should be able to get it by explicitly installing the latest version or by refreshing your lockfile.

PR #19986, opened 2023-03-29
feat(nuxt): support `~`/`~~`/`@`/`@@` aliases within layers

danielroe:
### ❓ Type of change

### 📚 Description

This PR aims to fix a pain point with layers - that if you are extending a  layer it cannot use aliases like `~`, `~~`, `@` and `@@` as these will be resolved relative to the _root_ layer. This plugin uses a `resolveId` hook for vite and a slower transform for webpack as `resolveId` doesn't seem to be working (am I missing something @antfu?)

It is possible to disable this behaviour with the following config:

```ts
  experimental: {
    localLayerAliases: false
  }
})
```

### 📝 Checklist


packages/nuxt/src/core/plugins/layer-aliasing.ts
```diff
@@ -0,0 +1,69 @@
...
+const ALIAS_RE = /(?<=['"])[~@]{1,2}(?=\/)/g
+
+export const LayerAliasingPlugin = createUnplugin((options: LayerAliasingOptions) => {
+  const aliases = Object.fromEntries(options.layers.map(l => [l.config.srcDir || l.cwd, {
+    '~': l.config.srcDir || l.cwd,
+    '@': l.config.srcDir || l.cwd,
+    '~~': l.config.rootDir || l.cwd,
+    '@@': l.config.rootDir || l.cwd,
+    ...l.config?.alias || {}
+  }]))
```
pi0, line 22: Passing aliases to `resolveAlias`, removes all user (top level) aliases. We might merge it with user aliases on `nuxt.options.alias` (maybe lower priority but also for me makes sense that it has higher priority than layer aliases to give users an escape hatch)

packages/nuxt/src/core/plugins/layer-aliasing.ts
```diff
@@ -0,0 +1,69 @@
...
+  const layers = Object.keys(aliases)
```
pi0, line 23: We need to sort layers with length i think because how layer matching works.

packages/nuxt/src/core/plugins/layer-aliasing.ts
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    name: 'nuxt:layer-aliasing',
+    enforce: 'pre',
+    vite: {
+      resolveId: {
+        order: 'pre',
+        async handler (id, importer) {
+          if (!importer) { return }
+
+          const layer = layers.find(l => importer.startsWith(l))
```
pi0, line 34: This could be refactored out to be shared between vite/webpack

pi0 (approved):
This is really nice and a most-have feature! 💯 

I haven't tried locally but looks good overall other than mentioned comments specially regarding user aliases.

Feel free to merge as is or via an experimental flag to iterate faster and trying on end-user projects.

packages/nuxt/src/core/plugins/layer-aliasing.ts
pi0: Then there is no way to for example force resolve a library to specific version. But lets try and iterate.

packages/nuxt/src/core/plugins/layer-aliasing.ts
danielroe: suggestions welcome, but I think the shared code is only this single line and refactoring would be more complex

packages/nuxt/src/core/plugins/layer-aliasing.ts
danielroe: it would still work. the layer probably is not providing a resolution for that library, right?

packages/nuxt/src/core/plugins/layer-aliasing.ts
pi0: Logic can be bigger we don't want to match a node_modules path nested in a layer for example. I suggested this to make later works easier but we can do it later too :)

packages/nuxt/src/core/plugins/layer-aliasing.ts
pi0: Layer might also use aliases for resolution.

danielroe: I think we can actually just list it inline to allow layer to configure it, and that way we avoid the issue.

Rigo-m:
Can we document this change somehow? It's not clear if we have to explicitly define aliases for layers or not

danielroe:
Yes, of course - a PR would be great. You do not; those aliases should just work.

PR #19992, opened 2023-03-30
fix(nuxt): store payloads in cache without trailing slash

### LINKED ISSUE #19989: Payload generation makes loading data twice if links have a trailing slash
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.3.1`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `ssr`, `experimental`, `modules`, `runtimeConfig`
- Runtime Modules:  `nuxt-graphql-client@0.2.27`
- Build Modules:    `-`

### Reproduction

With the [following repo](https://github.com/idflood/nuxt3-static-test/tree/catchall-page-payload-issue) on the catchall-page-payload-issue branch, if you run `npx nuxi generate` it will log this.

Screenshot 2023-03-30 at 09 19 21

Notice the long time generation for each payloads.

If you remove the trailing slash from the links (added [in the last commit](https://github.com/idflood/nuxt3-static-test/commit/1b6372f91ec17682a96ecfcc1a20ca719aeb67fe)) the payload generation is instant.

Screenshot 2023-03-30 at 09 18 21

### Describe the bug

On a client website I noticed that when generating pages it was loading the data twice, once to generate the page and a second time to generate the _payload.js.

The issue appear when links url have a trailing slash. 

### Additional context

I tried to normalise the url and add a trailing slash just after https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/runtime/nitro/renderer.ts#L199 . It fixed the issue for me but I don't know if such fix could create other edge cases.

If this looks good I can make a pull request for that. For information, here is the one line fix I implemented.

    url += !url.endsWith("/") ? "/" : "";

### Logs

_No response_

idflood:
### 🔗 Linked issue

#19989

### ❓ Type of change

### 📚 Description

This normalise url by ensuring there is a trailing slash before adding the data to the PAYLOAD_CACHE.

This resolve an issue where the content was loaded twice, once for page and a second time for payload extraction. This happened when links had trailing slashes in content.

Resolves #19989

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -197,6 +197,7 @@ export default defineRenderHandler(async (event) => {
   const isRenderingPayload = PAYLOAD_URL_RE.test(url)
   if (isRenderingPayload) {
     url = url.substring(0, url.lastIndexOf('/')) || '/'
+    url += !url.endsWith("/") ? "/" : "";
     event.node.req.url = url
     if (process.env.prerender && PAYLOAD_CACHE!.has(url)) {
       return PAYLOAD_CACHE!.get(url)
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -5,7 +5,7 @@ import type { H3Event } from 'h3'
 import { appendHeader, getQuery, writeEarlyHints, readBody, createError } from 'h3'
 import devalue from '@nuxt/devalue'
 import destr from 'destr'
-import { joinURL } from 'ufo'
+import { joinURL, withoutTrailingSlash } from 'ufo'
 import { renderToString as _renderToString } from 'vue/server-renderer'
 import { hash } from 'ohash'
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -197,6 +197,7 @@ export default defineRenderHandler(async (event) => {
   const isRenderingPayload = PAYLOAD_URL_RE.test(url)
   if (isRenderingPayload) {
     url = url.substring(0, url.lastIndexOf('/')) || '/'
+    url += !url.endsWith("/") ? "/" : "";
```
danielroe, line 200: How about instead we strip the trailing slash at L257?

idflood: @danielroe it works if I do it on L266, but we might need to do it on L257 too. If this is needed on both lines, I'm considering adding a small function like this to remove trailing slash, but not for the main "/" path.

    const removeTrailingSlash = (url2) => url2.endsWith('/') && url2 != '/' ? url2.substring(0, url2.length - 1) : url2

Should I change this? Otherwise I will just replace these double quotes and remove the semicolon to make linting happy.

danielroe: No L266 is the right place. We can use `withoutTrailingSlash` from `ufo` in this place, because it will only be used in the prerenderer, not added to the runtime server bundle.

idflood: @danielroe looks clean and works nicely, thanks for the suggestion 👍

danielroe (approved):

PR #20003, opened 2023-03-30
feat(nuxt): support object-syntax plugins

### LINKED ISSUE #14628: Define plugins with object syntax

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14628

### ❓ Type of change

### 📚 Description

This adds support for new object-syntax plugins, which can be used to directly register hooks or enforce rough ordering within a list of plugins.

### 📝 Checklist


packages/nuxt/src/app/nuxt.ts
```diff
@@ -142,7 +142,7 @@ export const NuxtPluginIndicator = '__nuxt_plugin'
 
 export interface PluginMeta {
   name?: string
-  enforce?: 'pre' | 'post'
+  enforce?: 'payload' | 'pre' | 'post'
```
pi0, line 145: `payload` as `enforce` can be kinda confusing because we might as well also have a post-post enforce that is to modify final `payload` while this (i suppose) is for payload reducers.

Would be nice to make it more clear

danielroe: Can you think of a better name? Maybe `payload-type` or `custom-type`?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -336,12 +358,37 @@ export function normalizePlugins (_plugins: Plugin[]) {
...
+  const wrapper: Plugin<T> = (nuxtApp) => {
+    if (plugin.hooks) {
+      nuxtApp.hooks.addHooks(plugin.hooks)
+    }
+    if (plugin.setup) {
+      return plugin.setup(nuxtApp)
+    }
+  }
+
+  wrapper.meta = {
```
pi0, line 384: Assigning meta-objects to functions is not a good practice (i couldn't find a reference but there was also performance implications with v8).

We might better always normalize to object format internally...

packages/nuxt/src/app/nuxt.ts
```diff
@@ -336,12 +358,37 @@ export function normalizePlugins (_plugins: Plugin[]) {
...
+  ]
 }
 
-export function defineNuxtPlugin<T extends Record<string, unknown>> (plugin: Plugin<T>) {
-  plugin[NuxtPluginIndicator] = true
-  return plugin
+export function defineNuxtPlugin<T extends Record<string, unknown>> (plugin: Plugin<T> | ObjectPluginInput<T>) {
+  if (typeof plugin === 'function') {
+    plugin[NuxtPluginIndicator] = true
+    return plugin
```
pi0, line 372: Maybe add a basic meta here too? (name could be infered but better, enforce explicitly added, etc)

packages/nuxt/src/app/nuxt.ts
pi0: What if we switch to a numeric order? We can map `pre` and `post` for convenience to user values but this way we have full control and can add nuxt internal pre(s) before/after user pre(s). Something like this

- -10: nuxt-pre-all
- -5: user-pre
- -1: nuxt-default
- 0: default
- 1: nuxt-after
- +5: user-after
- +10: nuxt-after-all

danielroe: Perfect. 👌

pi0: Re payloads, we could also make a special `definePayloadPlugin` wrapper what do you think?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -336,12 +363,62 @@ export function normalizePlugins (_plugins: Plugin[]) {
...
-  plugin[NuxtPluginIndicator] = true
-  return plugin
+// -50: pre-all (nuxt)
+// -40: custom payload revivers (user)
+// -30: payload reviving (nuxt)
+// -20: pre (user) <-- pre mapped to this
+// -10: default (nuxt)
+// 0: default (user) <-- default behavior
+// +10: post (nuxt)
+// +20: post (user) <-- post mapped to this
```
pi0, line 376: (`this` should be `pre`/`post`)

packages/nuxt/src/app/nuxt.ts
```diff
@@ -139,9 +139,31 @@ interface _NuxtApp {
...
 export const NuxtPluginIndicator = '__nuxt_plugin'
+
+export interface PluginMeta {
+  name?: string
+  enforce?: 'pre' | 'default' | 'post'
+  /**
+   * This allows more granular control over plugin order and should only be used by advanced users.
+   * It overrides the value of `enforce` and is used to sort plugins.
+   */
+  priority?: number
```
pi0, line 150: Shall we call this `order`? Considering lower number comes first.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -336,12 +363,62 @@ export function normalizePlugins (_plugins: Plugin[]) {
...
+    })
+  }
+
+  return defineNuxtPlugin({
+    setup: plugin,
+    priority: -40
+  })
+}
+
+export function defineNuxtPlugin<T extends Record<string, unknown>> (plugin: Plugin<T> | ObjectPluginInput<T>): Plugin<T> {
```
pi0, line 399: suggestion: Considering the usage in `definePayloadPlugin`, we could accept a `meta` object as second parameter too. Also allows wrapping an existing plugin with additional meta like order.

packages/nuxt/src/app/nuxt.ts
pi0: Maybe also expose them as an (internal) map for nuxt hook numbers as constants/in a enum?

danielroe: I've deliberately not exposed them (for now) so we have freedom to change them internally.

pi0: I'm not suggesting to expose either. Just internally make a variable for levels.

danielroe: Ah, understood. I think the only place where we use them outside this file is the payload reviver plugin. I think it's simpler not to export it as it will be exposed via `#app/nuxt` if we do.

pi0: okay lets do it later

PR #20024, opened 2023-04-01
fix(nuxt): disable `x-nuxt-no-ssr` header by default

### LINKED ISSUE #15667: Opt-in (or out) of "x-nuxt-no-ssr" HTTP header feature
### Describe the feature

I came across the undocumented `x-nuxt-no-ssr` header, which makes it possible to disable SSR per request. While certainly useful in some cases, I think it's a bit problematic:

- If you build your app around the assumption that every request will for sure always be rendered server-side, you may not even test that a SPA response from the server actually works in the client. E.g. the page might be broken because you rely on data only fetched SSR which is passed to the client via payload.
- If you use an HTTP cache in front of your Nuxt app it would technically be possible for someone to put a SPA version of an otherwise SSR page into the cache, unless this header is taken into account for the cache key. This might "only" be bad for SEO or in the worst case serve broken pages to all users.

Of course this is all highly theoretical, as you'd also need to set the correct response headers yourself that make the response cacheable in the first place and have a few other things "go wrong", like passing any request headers through the cache layer to your Nuxt app (but I think this is default behavior for e.g. Cloudflare).
In addition, once you're aware that this header exists, you can easily remove it already on cache level before passing the request to Nuxt or also remove it in a Nuxt event handler. But nonetheless it's a bit of a hidden feature that can lead to unexpected behavior.

One solution would be to offer a way to opt-out of this feature, or even better, make it opt-in, as it does feel like something that is only used for very specific edge cases and shouldn't be enabled by default.

danielroe:
### 🔗 Linked issue

resolves #15667

### ❓ Type of change

### 📚 Description

We have an undocumented header for disabling SSR responses, `x-nuxt-no-ssr`. I think it would be better not to have this enabled by default as it allows third-parties to control the behaviour of the Nuxt renderer.

I think we can consider this a non-breaking change if clearly signposted in release notes.

### 👉 Migration

If you were using the `x-nuxt-no-ssr` flag, you can turn on the previous behaviour with:

```ts
  experimental: {
    respectNoSSRHeader: true
  }
})
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -189,6 +189,15 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
     }
   }
 
+  // Add backward-compatible middleware to respect `x-nuxt-no-ssr` header
+  if (nuxt.options.experimental.respectNoSSRHeader) {
+    nitroConfig.handlers = nitroConfig.handlers || []
+    nitroConfig.handlers.push({
+      handler: resolve(distDir, 'core/runtime/nitro/no-ssr'),
+      middleware: true
+    })
+  }
+
   // Register nuxt protection patterns
   nitroConfig.rollupConfig!.plugins = await nitroConfig.rollupConfig!.plugins || []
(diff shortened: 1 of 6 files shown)


TheAlexLichter (approved):

pi0 (approved):
LGTM in general to make ssr behavior more predictable

Issue #19626, opened 2023-03-13
navigateTo in defineNuxtRouteMiddleware done not work


Mr-prudent:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v18.12.0`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Package Manager:  `npm@8.19.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-3axmwt?file=middleware/mobile.global.js
Simply add navigateTo in `middleware/mobile.global.js`

### Describe the bug

Middleware use : /middleware/mobile.global.js
```javascript
// middleware/mobile.global.js
export default defineNuxtRouteMiddleware((to, from) => {
  const route = useRoute();
  if (process.server) {
    const isMobile = true;
    if (isMobile && route.path !== '/mobile') {
      return navigateTo('/mobile', {
        replace: true,
      });
    }
  }
});
```
I try to use route middleware in `middleware/mobile.global.js`  and nuxt@3.2.3, but it done not work.
it works in nuxt@3.0.0.

danielroe:
The issue here is that you are calling `useRoute()`. In middleware, the route hasn't yet changed, so you need to use `to`, not the current route.

Here's an updated sandbox: https://stackblitz.com/edit/github-3axmwt-cfhkss.

dargmuesli:
Maybe there could be a warning when use of `useRoute` inside a middleware is detected?

danielroe:
Implemented ✅

PR #20050, opened 2023-04-03
feat(nuxt): warn in dev when `useRoute` is used in middleware

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/19626#issuecomment-1493375833

### ❓ Type of change

### 📚 Description

This adds a warning in dev mode when using `useRoute` within nuxt middleware, which will almost certainly be the wrong thing to do.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -17,6 +17,9 @@ export const useRouter = () => {
 }
 
 export const useRoute = (): RouteLocationNormalizedLoaded => {
+  if (process.dev && isProcessingMiddleware()) {
+    console.warn('[nuxt] Calling `useRoute` within middleware may lead to misleading results. Instead, use the (to, from) arguments passed to the middleware to access the new and old routes.')
+  }
   if (getCurrentInstance()) {
     return inject('_route', useNuxtApp()._route)
   }


packages/nuxt/src/app/composables/router.ts
```diff
@@ -17,6 +17,9 @@ export const useRouter = () => {
 }
 
 export const useRoute = (): RouteLocationNormalizedLoaded => {
+  if (process.dev && isProcessingMiddleware()) {
+    console.warn('[nuxt] Calling `useRoute` within middleware may lead to misleading results. Instead, use the (to, from) arguments passed to the middleware to access the new and old routes.')
```
pi0, line 21: Should we continue on getting results? (not throwing an error, might be harder to make developers see/condider the warning)

pi0 (approved):

packages/nuxt/src/app/composables/router.ts
danielroe: Tricky. I am assuming that there might _possibly_ be a reason someone might want to use it, but happy to make a fatal error if you think that is a better DX.

pi0: Lets do it in next releases then

mercs600: @pi0 @danielroe let's imagine we have composable which can be used in middleware or components. 

```js
export const useApi = (
    path?: string,
    options?: FetchOptions<'json'>
  ) => {
    let defaultPath = path
  
    if (!defaultPath) {
      defaultPath = useRoute().fullPath
    }
  })
  ```
  
In the middleware I can provide path from `to` object 
```js
useApi(to.fullPath)
```
But in the vue components I don't have to. 

But with this implementation we won't be able to use it in this way and we always have to pass path to composable which could be use thanks to `useRoute`. Why ?

mercs600: Unfortunately we got warnings in our library even If I add this condition

danielroe: Then it's not related, or you are still using it somewhere it shouldn't be used. Would you create a new issue with a reproduction?

mercs600: sorry guys, I had some issues on my side, everything is ok.

PR #20052, opened 2023-04-03
fix(nuxt): throw error on protocol relative path in `useFetch`

### LINKED ISSUE #19930: Wrong domain for fetch call if protocol not defined
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.3.2`
- Nitro Version:    `2.3.2`
- Package Manager:  `yarn@1.22.10`
- Builder:          `vite`
- User Config:      `ssr`, `vite`, `modules`, `runtimeConfig`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.6.4`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/p/sandbox/gallant-wilbur-15ut2n?file=%2Fapp.vue

### Describe the bug

If the protocol is not defined and route domain start with `//`, the domain are ignored

### Additional context

_No response_

### Logs

_No response_

Suv4o:
### 🔗 Linked issue

resolves #19930

### ❓ Type of change

### 📚 Description

If the protocol is not defined in the fetch call, and the url starts with `//`, prefix the url with the `https` protocol.

### 📝 Checklist


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -105,14 +105,20 @@ export function useFetch<
     controller?.abort?.()
     controller = typeof AbortController !== 'undefined' ? new AbortController() : {} as AbortController
 
-    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/')
```
danielroe, line 108: Maybe we could use `hasProtocol` from `ufo` to test `_request.value` instead?

Suv4o: Good idea! I didn't think of that. Let me update it using `hasProtocol`. Thanks for suggesting that, @danielroe

pi0:
Universally supporting `//` is tricky. On server-side it has no semantical meaning. I would suggest to only normalize for client-side based on actual request and throw an error in server-side (ofetch and nitro will probably never support relative protocol)

Suv4o:
> Universally supporting `//` is tricky. On server-side it has no semantical meaning. I would suggest to only normalize for client-side based on actual request and throw an error in server-side or always throw an error for consistency. Adding explicit protocol is not that hard.
> 
> (ofetch and nitro will probably never support relative protocol)

Not sure if I've done it correctly, but I updated the code to always throw an error for consistency. Please review the changes when you have a chance. @pi0 @danielroe

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -59,6 +59,14 @@ export function useFetch<
   if (!request) {
     throw new Error('[nuxt] [useFetch] request is missing.')
   }
+
+  if (typeof request === 'function') {
+    const requestUrl = request() as string
+    if (requestUrl.startsWith('//')) {
+      throw new Error('[nuxt] [useFetch] the request URL must not start with "//".')
+    }
+  }
```
danielroe, line 68: maybe move this to L123 and test `_request.value` instead?

Suv4o: @danielroe I moved the logic to line 123. The reason I had it at line 63 earlier was because I wanted to keep all the "throw errors" together.

Now it works okay, but I cannot figure out why the error does not appear nicely on the screen when thrown. It kind of throws an error silently. This is not the case if I move the logic back to line 63.

danielroe: Good point - it's being captured and included in `error.value` but that's a little too hidden. We could move it up then, just after the creation of `_request` so we don't have to normalise the request twice.

pi0 (approved):
Seems good to me. Alternatively we could add same error directly to the `$fetch` client exposed by nitro for more universal fix.

Suv4o:
Thanks for moving the logic, @danielroe . Is there anything else I can do to improve this issue?

danielroe:
Looks good to me!

PR #20082, opened 2023-04-04
fix(nuxt): deprecate old (pre-rc) runtimeConfig

danielroe:
### ❓ Type of change

### 📚 Description

We've supported accessing non-existent properties on runtimeConfig (e.g. `config.testConfig` would not fail even if the actual property was `config.public.testConfig`). This was introduced in https://github.com/nuxt/framework/pull/4254 in the first Nuxt 3 RC but has not been documented and was intended to help migration between Nuxt 2 -> Nuxt 3. I think we can now remove it safely.

In addition, we do not need to wrap config in `reactive` a second time as it is already reactive.

### 📝 Checklist


pi0:
This seems a good idea to get rid of backward compatibility transform. However since it is unclear how many modules / direct project might still implicitly are relying on this proxy, it would be probably nicer that we soft-deprecate with a warning message and drop feature in next 1-2 minor versions. Wdyt?

danielroe:
Deprecation is a good idea in case users are using old Nuxt 2 modules, although it was neither documented nor typed so I think it's unlikely to be an issue. We can remove in v3.5.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -244,17 +244,19 @@ export function createNuxtApp (options: CreateOptions) {
...
-    get (target, prop) {
-      if (prop === 'public') {
-        return target.public
+    get (target, prop: string) {
+      if (prop in target) {
+        return target[prop]
       }
-      return target[prop] ?? target.public[prop]
+      if (process.dev && prop in target.public) {
+        console.warn(`[nuxt] [runtimeConfig] You are trying to access a public prop (\`${prop}\`) directly from runtime config. This currently works (for backward compatibility with Nuxt 2) but this compatibility layer will be removed in a future minor release.`)
```
danielroe, line 257:
```suggestion
        console.warn(`[nuxt] [runtimeConfig] You are trying to access a public runtime config value (\`${prop}\`) directly from the top level. This currently works (for backward compatibility with Nuxt 2) but this compatibility layer will be removed in a future minor release. Instead, you can update \`config['${prop}']\` to \`config.public['${prop}']\`.`)
```

pi0 (approved):

PR #20093, opened 2023-04-04
fix(nuxt): add key to `.client` component placeholders

### LINKED ISSUE #20061: `.client.vue` scoped styles are missing on ssr
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.3.3`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `experimental`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-a1d8zg?file=pages%2Findex.vue

### Describe the bug

Similar to what https://github.com/nuxt/framework/pull/8713 resolved a while back but it seems like the issue persists for `.client.vue` files?

Setting `experimental: {
    treeshakeClientOnly: false,
  },` which was the workaround for the mentioned pr does however not fix the issue

### Additional context

`.client` component on ssr:
image

`.client` component on client side navigation
image

The `<client-only>` wrapped component on the other hand works for both ssr and csr: 
image

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/20061

### ❓ Type of change

Hi :wave: this PR fix `.client` components root nodes missing scopeId by simply adding a `key`

### 📝 Checklist


antfu:
@huang-julien can you explain more about why adding a boolean key can help the scopeId?

huang-julien:
it is just to completly reload the render. 
My guess was that since the placeholder is a `div` and the components root node is also a `div`, vue does not set the `v-data` scope Id. We don't have the issue if the `.client` component renders a `Fragment` or any other tags.

antfu:
Maybe we could assign a random(or fixed) key to the placeholder but not the rendered component?
So placeholder with key should already not be treat the same as the rendered component without key. There shouldn't be much difference in practice but I am a bit worried overriding the `key` user might define, or causing the component to mount multiple times etc.

huang-julien:
ah yes it should be obvious :sweat_smile: ....

packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -37,7 +38,7 @@ export function createClientOnly (component) {
           ? createElementVNode(res.type, res.props, res.children, res.patchFlag, res.dynamicProps, res.shapeFlag)
           : h(res)
       } else {
-        return h('div', ctx.$attrs ?? ctx._.attrs)
+        return h('div', mergeProps(ctx.$attrs ?? ctx._.attrs, { key: randomUUID() }))
```
danielroe, line 41: do we need this to be random, or is assigning any key good enough?

huang-julien: I think it should work too, i'll try

danielroe (approved):

PR #20112, opened 2023-04-06
refactor(schema): clean up `experimental` options

danielroe:
### ❓ Type of change

### 📚 Description

For the upcoming minor release (v3.4), this PR makes three changes to `experimental` options:

1. remove `experimental.viteNode` which has already been deprecated for some time (https://github.com/nuxt/nuxt/commit/1b2304b632ed8d44da17cc767d213e8dfc10ff7c).
2. remove a completely unused option (`viteServerDynamicImports`).
3. ~~move `experimental.externalVue` to `vue.externalLib` and deprecate/warn if this option is being used.~~

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/app.ts
+++ b/packages/schema/src/config/app.ts
@@ -8,14 +8,6 @@ export default defineUntypedSchema({
    */
   vue: {
     /**
-     * Externalize `vue`, `@vue/*` and `vue-router` when building your server.
-     *
-     * @see https://github.com/nuxt/nuxt/issues/13632
-     */
-    externalLib: {
-      $resolve: async (val, get) => val ?? (await get('experimental.externalVue')) ?? true
-    },
-    /**
      * Options for the Vue compiler that will be passed at build time.
      * @see [documentation](https://vuejs.org/api/application.html#app-config-compileroptions)
      * @type {typeof import('@vue/compiler-core').CompilerOptions}
--- a/packages/schema/src/config/app.ts
+++ b/packages/schema/src/config/app.ts
@@ -8,6 +8,14 @@ export default defineUntypedSchema({
    */
   vue: {
     /**
+     * Externalize `vue`, `@vue/*` and `vue-router` when building your server.
+     *
+     * @see https://github.com/nuxt/nuxt/issues/13632
+     */
+    externalLib: {
+      $resolve: async (val, get) => val ?? (await get('experimental.externalVue')) ?? true
+    },
+    /**
      * Options for the Vue compiler that will be passed at build time.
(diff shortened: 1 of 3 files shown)


packages/schema/src/config/app.ts
```diff
@@ -7,6 +7,14 @@ export default defineUntypedSchema({
    * Vue.js config
    */
   vue: {
+    /**
+     * Externalize `vue`, `@vue/*` and `vue-router` when building your server.
+     *
+     * @see https://github.com/nuxt/nuxt/issues/13632
+     */
+    externalLib: {
+      $resolve: async (val, get) => val ?? (await get('experimental.externalVue')) ?? true
```
pi0, line 16: Are we sure this option has valid usecases? If it is only for edge-cases, i like to keep it under experimental and remove with a default behavior finally. Because if we support an option like this within schema we need to always keep it and this seems only an option for edge cases...

danielroe: You mean, are there valid use cases for create a server bundle that inlines vue? I think at the moment, the value is  that we can skip tracing babel, estree, etc. which are dependencies of the runtime vue compiler and otherwise get traced.

Having said that, we can always remove experimental option along with the necessary config if users were doing this manually, and ask anyone who was doing this to let us know the blocking issues for them.

pi0: From what I understand this change is moving the experimental option to a non-experimental one that is harder to deprecate. Why do we need to move it out of experimental as a permanent option for nuxt 3? 

Regarding tracing, i think nitro preset already is configured to mock those libs once making final bundle. Still we can internally improve things like this as long as there is no option that makes different conditions.

danielroe: For the schema, I agree with you and would be very happy to remove the option entirely, in favour of keeping Vue as an external, properly traced.

But as far as tracing is concerned, we include `@babel/parser`, `estree-walker` and even `source-map` in the minimal server node_modules. Nitro isn't doing anything to mock them.

https://github.com/nuxt/nuxt/blob/main/test/bundle.test.ts#L54

I would love to exclude those unless runtime compiler is enabled.

pi0: Keeping in externals would be probably better idea.

Made an upstream feature for nitro https://github.com/unjs/nitro/issues/1118

danielroe: Okay, let's keep it in `experimental` for now and aim to remove option entirely in v3.5 when upstream support has landed. ❤️

pi0 (approved):

PR #20115, opened 2023-04-06
docs: add custom fetch composable example

adrienZ:
### 🔗 Linked issue

#14736 

### ❓ Type of change

### 📚 Description

Adds an example in the `other` folder of custom use fetch composable. It handles default values and auth but can be overriden.

### 📝 Checklist


examples/other/use-custom-fetch-composable/composables/useCustomFetch.ts
```diff
@@ -0,0 +1,34 @@
...
+      // return new myBusinessResponse(response._data)
+    },
+
+    onResponseError (__ctx) {
+      // return new myBusinessError(error)
+    }
+  }
+
+  // for nice deep defaults, please use unjs/defu
+  const params = Object.assign(defaults, options)
```
danielroe, line 31: nice idea! let's go ahead and use defu then

examples/other/use-custom-fetch-composable/composables/useCustomFetch.ts
```diff
@@ -0,0 +1,34 @@
...
+    // cache request
+    key: url,
+
+    onRequest ({ options }) {
+      options.headers = options.headers ?? new Headers()
+
+      // set user token if connected
+      if (userAuth.value) {
+        options.headers.Authorization = `Bearer ${userAuth.value}`
+      }
```
danielroe, line 18: we might also need to handle the case where this is a `Headers` object and we need to do `headers.set()`. For simplicity, is there any reason not to use the `headers: {}` option directly instead of within `onRequest`?

examples/other/use-custom-fetch-composable/composables/useCustomFetch.ts
adrienZ: updated :)

examples/other/use-custom-fetch-composable/composables/useCustomFetch.ts
adrienZ: tried something simpler :)

examples/other/use-custom-fetch-composable/composables/useCustomFetch.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    onResponseError (__ctx) {
+      // return new myBusinessError(error)
+    }
+  }
+
+  // for nice deep defaults, please use unjs/defu
+  const params = defu(defaults, options)
+
+  return await useFetch(url, params)
```
danielroe, line 30: I think we can directly return `useFetch`:

```suggestion
  return useFetch(url, params)
```

danielroe (approved):
Thanks!

adrienZ:
should we close #14736 ?

danielroe:
We also need to add some composables for modifying the inner behaviour of the data fetching composables, much as we have defineNuxtLink.

PR #20126, opened 2023-04-07
feat(nuxt): prompt to install `devtools` when it's enabled

antfu:
### ❓ Type of change

### 📚 Description

This is for v3.4. We will ship `@nuxt/devtools` with `nuxt` and make it easier to opt-in by add `devtools: true` in the config. DevTools is still an experimental feature, and the changes will not be reflected on Nuxt's version.

### 📝 Checklist


antfu:
Mark it as a draft for now. I will need to work on the devtools for the install script and bump the version

pi0:
Currently, `@nuxt/devtools` has 12MB download and 76.2MB install size addition (https://packagephobia.com/result?p=%40nuxt%2Fdevtools).

Also as discussed in team meeting, adding direct dependency makes it harder for users to choose a major version of ad-hoc modules like devtools and image.

I love the idea of providing devtools out of the box but it can be installed on first use automatically as a user project dependency using nypm.

danielroe:
What about when using `nuxi upgrade`, if new version is v3.4+ and no devtools installed, we offer to install it instead and add it to the upgrade command? Then only thing required to enable is `devtools: true` in nuxt.config, as in this PR?

pi0:
Seems good idea with `upgrade` integration 💯 

Since schema resolvers can be async, we can check if flag is enabled and dependency not installed, to ask for installing it (if user says no, we simply disable flag)

antfu:
I pushed the change to make it prompt for auto-install. Which inevitably introduced three new dependencies (`local-pkg`, `nypm`, and `prompts`). Later we could move some functionality from `local-pkg` to `nypm` but I think it's non-blocking in this case.

For that is worth it, I managed to reduce devtools's package from 13MB to 3MB. And the install size doesn't present much as most are shared with Nuxt. So the total additional size would be around 5MB (where `nuxt` itself is 123MB, so I consider it quite acceptable).

danielroe:
I'm not worried about the extra size of the devtools dependency; I think it's worth it.

packages/nuxt/src/core/features.ts
```diff
@@ -0,0 +1,35 @@
...
+  })
+
+  if (!confirm) {
+    return false
+  }
+
+  logger.info(`Installing ${name}...`)
+  try {
+    await addDependency(name, {
+      cwd: rootDir
```
pi0, line 27: I think we should enable `dev: true` to add as dev dependency similar to `nuxt` one in starter

packages/nuxt/src/core/features.ts
```diff
@@ -0,0 +1,35 @@
+import { addDependency } from 'nypm'
+import { isPackageExists } from 'local-pkg'
+import { logger } from '@nuxt/kit'
+import prompts from 'prompts'
+
+export async function ensurePackageInstalled (rootDir: string, name: string) {
```
pi0, line 6: PR welcome to move all of logic from this util directly to nypm. We could early integrate with consola v3 there too also removing `prompts` install overhead

packages/nuxt/src/core/nuxt.ts
```diff
@@ -366,6 +366,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
     options._modules.push('@nuxt/telemetry')
   }
 
+  // Nuxt DevTools is currently opt-in
+  if (options.devtools === true || (options.devtools && options.devtools.enabled === true)) {
```
pi0, line 370: If provided an empty object, maybe a falsy check is safer?

```suggestion
  if (options.devtools === true || (options.devtools && options.devtools.enabled !== false)) {
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -366,6 +366,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
...
+    if (await import('./features').then(r => r.ensurePackageInstalled(process.cwd(), '@nuxt/devtools'))) {
+      options._modules.push('@nuxt/devtools')
+    } else {
+      logger.warn('Failed to install @nuxt/devtools, please install it manually, or disable devtools in nuxt.config')
```
pi0, line 374:
```suggestion
      logger.warn('Failed to install `@nuxt/devtools`, please install it manually, or disable `devtools` in `nuxt.config`')
```

antfu:
For another aspect of thinking:

- Currently, DevTools is an experimental feature, so we clearly say we expect changes without following semver
- Later, once we go stable, we could froze the interface with Nuxt (meanwhile we might consider enabling by default), and do majors along with Nuxt 4 if needed later.

Giving DevTools is mostly a UI, breaking changes usually won't affect much (only the modules API, which is not Nuxt core's concern either). So I feel it's ok for devtools to ship directly with Nuxt to make it easy to use and "zero-config".

pi0:
> I guess the main question is whether it will be harder to upgrade the devtools dependency in users' projects.

Upgrade prompts for nuxt itself and image are probably sharing same requirement and definitely something i have in vision for nuxi-ng 🤞 With manual prompts with can also explicitly ask them about major bumps and linking to changelog/release notes too.

pi0:
@antfu I agree with most of your points about the current status of devtools. Only it would be more consistent for me that we explicitly install ad-hoc modules similarly. It is not a concern about install size anymore at least with the current dependencies of devtools you optimized 💯 

If it is okay with you I would suggest using opt-in with manual install wizard while devtools is still in _experimental_ state and rethink once both dev tools marked as stable and nuxi-ng landed to check DX trade-offs.

danielroe:
Given that Nuxt will look for a module first in user-specified dependencies, even if we ship with a dependency on `@nuxt/devtools`, I believe the user will still be able to specify the version they want manually in `package.json`.

There are trade-offs, either way, of course, but I don't see any significant reason not to ship with a dependency on it, given that it can be disabled and that we are specifying it as experimental. Plus, the DX is 👌

Having said all of this, I'm very happy to leave this decision up to @antfu as he's taking responsibility for devtools. ❤️

pi0:
> Given that Nuxt will look for a module first in user-specified dependencies, even if we ship with a dependency on @nuxt/devtools, I believe the user will still be able to specify the version they want manually in package.json.

It would cause hoisting issues at least for having latest or different version of devtools on a a specific version of nuxt that pins to a 0.x major version by adding to `dependencies`. And currently, devtools has many of overlapping nuxt/unjs dependencies they can all be hoisted.

> we are specifying it as experimental.

Devtools option is not in experimental namespace too btw. Currently the only indication is that version is below zero... But i think should be okay as long as we ask to install and mention this.


All this said, i don't think we should add dependency regardless until devtools is marked as stable and we migrate to nuxi-ng with update banner/prompts. This makes it much easier to communicate users about upgrading individual implicit dependencies of nuxt such as nitro, image, devtools, content, etc with a parallel release cycle.

Adding a dependency to `nuxt` package that we _might_ change later is also technically a breaking change. I don't want to make this because of an early decision. And opt-in DX is already really nice 👌🏼

pi0:
I agree that we should move the toolkit parts toward the core `@nuxt/kit` once stable to encourage module authors to adopt integration with devtools once stable 👍🏼 

And on dependency, again, parallel versioning and releases are my main concern. Bumping from 0.3.x to 0.4.x fragments by nuxt version which of these (major) dependencies has, can potentially cause hoisting in `node_modules`, etc. 

I am even fine to add devtools to starter templates btw. Only not a dependency of nuxt package today.

antfu:
Also to note, we could specify the semver as `>=0.3.0` (which is my init commit doing), where the package manager will auto-dedupe when the user explicitly specifies the version when needed. After all, we are already asking for `devtools: true` opt-in to give enough awareness (also it prints an "experimental" message to the console and also in the UI)

antfu:
I am not saying no, in fact the PR is already doing prompt for installation but not the other way around. It's not about "trust" (ofc I trust you) we are just discussing trade-offs.

I am definitely not rushing on it. Just saying eventually we will ship it as dep to enable by default, but I am totally fine with the option + prompt in 3.4 as it's already a very good DX improvement.

In that sense, any other suggestions for the current implementation?

pi0:
I also trust you on your vision for devtools and it's integration to the core ❤️ (mentioned trust because there are points, i couldn't show today without shipping nuxi-ing first 😅)

pi0 (approved):
Overall looks good to me. The `ensurePackageInstalled` could be moved to nypm but i think we can merge this to try experience on edge and do refactor in last minute before 3.4 release. 

Thanks for going along with conv.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -366,6 +366,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
     options._modules.push('@nuxt/telemetry')
   }
 
+  // Nuxt DevTools is currently opt-in
+  if (options.devtools === true || (options.devtools && options.devtools.enabled !== false)) {
+    if (await import('./features').then(r => r.ensurePackageInstalled(process.cwd(), '@nuxt/devtools'))) {
```
danielroe, line 371:
```suggestion
if (await import('./features').then(r => r.ensurePackageInstalled(options.rootDir, '@nuxt/devtools', options.modulesDir))) {
```

packages/nuxt/src/core/features.ts
```diff
@@ -0,0 +1,36 @@
...
```
danielroe, line 6:
```suggestion
export async function ensurePackageInstalled (rootDir: string, name: string, searchPaths?: string[]) {
```

packages/nuxt/src/core/features.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (isPackageExists(name)) {
```
danielroe, line 7:
```suggestion
  if (isPackageExists(name, { paths: searchPaths })) {
```

danielroe (commented):
This looks beautiful. Nice work. Just a small suggestion for better monorepo support.

Atinux:
Happy to go that way for 3.4

danielroe (approved):

PR #20172, opened 2023-04-10
chore: remove `JITI_ESM_RESOLVE`

danielroe:
### ❓ Type of change

### 📚 Description

This removes using `JITI_ESM_RESOLVE` in CI, introduced in https://github.com/nuxt/framework/pull/3308. Would you remind me of the particular issue it was meant to solve @pi0?

### 📝 Checklist


pi0:
Mirrring from discord for record:

> It was estree-walker dependency. You might try removing it because another fix (prefer require over module condition) should have covered it not sure. But anyway this flag will be eventually on by default and is already on for internal jiti require calls. There was some early issues with mlly i had to disable not sure exactly what was that.

danielroe:
Great. I think let's keep behaviour as close to default as possible.

PR #20180, opened 2023-04-10
fix(imports): toggle nitro autoImports with `imports.autoImport`

harlan-zw:
### ❓ Type of change

### 📚 Description

When developing a Nuxt module using the provided starter, the auto imports are disabled with the `.nuxtrc` [file](https://github.com/nuxt/starter/blob/module/.nuxtrc). This is helpful because when a module is published, the auto imports won't work, allowing module authors to find import issues before publishing.

However, this config would seem to imply that all auto imports are disabled. This isn't the case as nitro auto imports still work. 

```ts
export default defineEventHandler((_event) => {
  return 'Hello!'
})
```

The above would work while in development but fail once deployed.

I've tested the fix in this PR and it seems to work correctly, however, one issue is that the type doesn't seem to be available, I'm not sure why that is. It seems purposefully excluded from the config types. Though within Nitro this config is parsed directly to the rollup plugin which uses this. Maybe a Nitro or an Unplugin PR is also needed?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -37,6 +37,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
     dev: nuxt.options.dev,
     buildDir: nuxt.options.buildDir,
     imports: {
+      autoImports: nuxt.options.imports.autoImport,
       imports: [
         {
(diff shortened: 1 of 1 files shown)


pi0 (approved):
LGTM. Good idea!

PR #20202, opened 2023-04-11
fix(schema): allow `ignorePrefix` to be changed

harlan-zw:
### ❓ Type of change

### 📚 Description

Modifying the `ignorePrefix` option does not work. It will always remain `-`.

```ts
  ignorePrefix: 'ignore-',
});
```

See here for the reproduction: https://stackblitz.com/edit/nuxt-starter-nqccka?file=pages/-is-ignored.vue (notice the server console.log output vs the config)

I had an issue that maybe is related to (https://github.com/harlan-zw/nuxt-simple-sitemap/issues/30) and figured we should fix it.

Sidenote: Personally I think this config should be deprecated in its current form, it should be an array of strings to be useful, in which case `ignore` would be better suited regardless.

### 📝 Checklist


danielroe (approved):
Thank you! ❤️

PR #20247, opened 2023-04-12
fix(nuxt): do not redirect when `vue-router` normalises url

### LINKED ISSUE #14316: Infinite redirect when the url contains ANY "+" character breaking the application
Hi guys, first of all thanks for the amazing job in this awesome tool!

I'm facing an several issue when the user tries to access the Nuxt application (SSR mode) using an url that contains any "+" character, the application apply an endless 302 redirect breaking all the application.

### Versions

- nuxt: [3.0.0-rc.4, 3.0.0-rc.5]
- node: [14.17.1, 16.14.2, 16x (Vercel)]

### Reproduction

Url to test:
https://test-nuxt-error.vercel.app/?asd=1231+

Example repo with the error:
https://github.com/igortrinidad/test-nuxt-error

### Steps to reproduce
Deploy an application on Vercel that contains vue-router (using Nuxt pages), if the application does not contains vue-router it won't happens...

### What is Expected?
The application been served normally

Infinite redirects 302 breaking the application

### I tried
Update the vue-router to the latest version
Create a new installation of Nuxt and deploy it
If the installations doesn't have pages (router...) the error doesn't happens
If the application has pages (what is pretty normal to an web application) the error starts to happens..

### Why

Why i need to access an url that contains the"+" character in my application, im implementing Google auth and the redirect url from the provider come with the "+" character in it which i can't control or sanitize.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14316

### ❓ Type of change

### 📚 Description

vue router normalises `/?asd=1231%20` to `/?asd=1231+` but this isn't actually a redirect performed by middleware. We can respect `to.redirectedFrom` to tell us whether there's been an _explicit_ redirection rather than just vue-router normalising things for us.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/plugins/router.ts
+++ b/packages/nuxt/src/pages/runtime/plugins/router.ts
@@ -8,7 +8,7 @@ import {
   createWebHistory
 } from 'vue-router'
 import { createError } from 'h3'
-import { isEqual, withoutBase } from 'ufo'
+import { withoutBase } from 'ufo'
 
 import type { PageMeta, Plugin, RouteMiddleware } from '../../../app/index'
 import { callWithNuxt, defineNuxtPlugin, useRuntimeConfig } from '#app/nuxt'
@@ -178,11 +178,10 @@ export default defineNuxtPlugin({
           statusMessage: `Page not found: ${to.fullPath}`
         })])
       } else if (process.server) {
-        const currentURL = to.fullPath || '/'
-        if (!isEqual(currentURL, initialURL, { trailingSlash: true })) {
+        if (to.redirectedFrom) {
           const event = await callWithNuxt(nuxtApp, useRequestEvent)
           const options = { redirectCode: event.node.res.statusCode !== 200 ? event.node.res.statusCode || 302 : 302 }
-          await callWithNuxt(nuxtApp, navigateTo, [currentURL, options])
+          await callWithNuxt(nuxtApp, navigateTo, [to.fullPath, options])
         }
       }
     })
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -130,6 +130,11 @@ describe('pages', () => {
     await page.close()
   })
 
+  it('returns 500 when there is an infinite redirect', async () => {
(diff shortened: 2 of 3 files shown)


smaeda-ks (approved):
Confirmed this is working as expected on Vercel deployment 👍 Thank you! 💯

packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -178,11 +178,10 @@ export default defineNuxtPlugin({
           statusMessage: `Page not found: ${to.fullPath}`
         })])
       } else if (process.server) {
-        const currentURL = to.fullPath || '/'
-        if (!isEqual(currentURL, initialURL, { trailingSlash: true })) {
+        if (to.redirectedFrom) {
```
pi0, line 181: Shouldn't we still check if `redirectedFrom` is not duplicate of current route (with/without trailing slash) to avoid loops?

danielroe: Vue router handles that for us. If users do redirect to the same route then we never get to this point. A 500 error is returned from vue-router and this warning is printed in the console:

```
[10:31:52 AM] [Vue Router warn]: Detected an infinite redirection in a navigation guard when going from "/" to "/?test". Aborting to avoid a Stack Overflow. This will break in production if not fixed.
```

danielroe: (I'll add a test.)

pi0: Good to know. Considering there are developers that simply ignore such warnings or situations that ONLY this trailing slash is reproducable in production, does it makes sense that we still either explicitly handle it (also for production) or make an improvement to vue-router to do this for production?

danielroe: It also does it in production.

pi0 (approved):
LGTM 👍🏼

pi0 (approved):

Issue #20056, opened 2023-04-03
Server Memory leaks, high CPU usage - appConfig


burda-osowiecki:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.15.0`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.3.2`
- Package Manager:  `npm@9.6.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

I cannot reproduce it in small scale and ofc also cannot share our project code...
I hope someone other will have similar problem and be able to reproduce in from scratch.

### Describe the bug

![image](https://user-images.githubusercontent.com/118734217/229534317-182a8443-7a4a-49f8-8ea6-20ef2ff8373e.png)
![image](https://user-images.githubusercontent.com/118734217/229534514-e5e3ccda-4c9c-4ef5-8e83-70897beb0ab4.png)
![image](https://user-images.githubusercontent.com/118734217/229535443-09882080-5583-4ef7-8725-6caaf448c5fe.png)

### Additional context

We disabled all plugins and unnecessary third party libraries.
I was also trying the solution from this issue:
https://github.com/unjs/destr/issues/9
I passed a parseResponse option to useFetch but it didn't change anything.

_No response_

danielroe:
Would you provide a [reproduction](https://nuxt.com/docs/community/reporting-bugs/#create-a-minimal-reproduction)? 🙏

You don't need to share exactly your code, but sharing the _kind_ of code you are running is a helpful start.

burda-osowiecki:
Looks like the problem causes appConfig.
I moved the appConfig (which is constant object in our case) to key in runtimeConfig, then created a composable to return this object from runtimeConfig and replaced all usage of `useAppConfig`.
Thats the results. It also effects the CPU usage.
**I'll try to create reproduction.**
![image](https://user-images.githubusercontent.com/118734217/231611160-0b70879e-0df1-434d-83dc-75b94f73bec2.png)
![image](https://user-images.githubusercontent.com/118734217/231611182-05b41413-7149-441c-a15d-883b3af399ac.png)

burda-osowiecki:
@danielroe finally I prepared the reproduction.
https://github.com/burda-osowiecki/nuxt-memory-leak

`npm run build`
`npm run start:debug`
`gtimeout 10 siege -c 10 http://localhost:3000`

[...]

PR #20278, opened 2023-04-14
fix(nuxt): clone app config on server

danielroe:
### ❓ Type of change

### 📚 Description

Calling `reactive` on an object mutates the object, so previously we were causing a memory leak by mutating app config (as the underlying object was shared between requests). Instead, this PR clones the object on server side - and does not make it reactive. This might help with https://github.com/nuxt/nuxt/issues/20056 (although I can't verify for sure as there is no reproduction).

~~Additionally, as app config is shared across requests _anyway_, I don't believe we need to set it on `NuxtApp` as well. This frees us up to use app config ambiently, not necessarily in a plugin/setup context. We can also somewhat reduce the implementation size by getting rid of a workaround for Vite HMR.~~

~~I would welcome context/discussion if you think differently - an alternative solution would be to use structuredClone or equivalent + not calling reactive on the app config.~~

### 📝 Checklist


Atinux:
I like it. We don’t need to make it reactive on SSR either to attach to nuxtApp instance

antfu (approved):

pi0 (dismissed):
Nice find but `appConfig` is not supposed to be shared across requests same. One of the main fundamental goals is that it can be modified per-request basis so it needs to be bound to nuxt app instance. Making things like [remote-config](https://github.com/nuxt-experiments/nuxt-remote-config/tree/main) possible but not limited to this only.

We might try structured cloning or simply a deep clone on first edit (via a proxy)

danielroe:
Since originally suggesting `structuredClone` I've changed my mind a little. Here's why:

1. We currently expose `useAppConfig` as usable in Nitro routes [_outside_ the event/request context](https://github.com/unjs/nitro/blob/d2b34e02131de9435bc01443e8a5a00ee2cc434f/src/runtime/config.ts#L16-L18). I think keeping DX consistent between Nitro + Nuxt means that we need to allow this kind usage.

1. Apart from that, I think it would be strange if we cannot access this 'build-time' config outside of a request - it means it can't be used for setting properties in Vue component defaults, or initialising global defaults. That seriously limits its usability, for example in themes.

Having said that, I think we have two options. (More ideas welcome.)

1. **Single composable**. We keep `useAppConfig` compatible with Nitro-style usage (be usable without referencing `event`) and equalise Nuxt behaviour so that it is the same - this config can be called and used in ambient JS. When this is called in ambient JS or outside of a request context, it is deep-frozen (on the server) just like Nitro. When it is called within a Vue/Nuxt request context, then it is automatically cloned, attached to Nuxt instance and made reactive.

1. **New composable**. We introduce a new function `useBaseAppConfig` which allows accessing (but not setting) the 'raw' app config. This is always deep-frozen on the server and can be used 'ambiently' outside of the request context. Alternatively, we add a new composable `useRequestAppConfig` which gives access to a reactive app config that is bound to the request.

Thoughts welcome.

pi0:
Regarding nitro, I'm thinking to also make it also stricter to require composable utils like `useStorage` and `useAppConfig` (also `useRuntimeConfig`) to be used within event handlers. not outside them because platforms such as cloudflare modules only provide environment within the context and that's why there are a callable composable not a direct export like `{ nuxtAppConfig }`. The current code you shared is going to be changed (very soon!)

We could fall back to the frozen version of the app config-if nuxt app context is not available to avoid introducing another utility but i would even make a warning for that. App config logic should be context aware. Wdyt?

pi0 (dismissed):
(discussion to be resolved it is not stale)

danielroe:
Updated (and following https://github.com/unjs/nitro/pull/1154 to use `klona` on server). Once that's merged, we can align the implementations further...

pi0 (approved):
LGM. Nice solution 💯

Issue #19772, opened 2023-03-17
server component roadmap


danielroe:
We have experimental support in Nuxt 3 for both island components and server-only components (which are implemented using islands) - see https://github.com/nuxt/framework/pull/5689 and https://github.com/nuxt/framework/pull/9972 for initial implementations and https://github.com/nuxt/nuxt/discussions/17956 for initial partial RFC.

There are a number of future linked improvements. This issue is to track those and provide some place to document the linking story, purpose and usage of server components within Nuxt.

## Purpose and usage

Currently we are collecting use patterns to ensure that the final API and implementation for server components is useful. But the following are some benefits we are exploring:

1. security and direct access to server APIs within certain components
2. separate caching strategies for individual chunks of HTML
3. leaner bundle size (keeping some dependencies solely on the server, e.g. syntax highlighting)
4. extensibility
5. rendering individual components within other modules or server routes (e.g. for component previews, OG images, PDF downloads, etc.)
6. ...

Comments and suggestions are very welcome.

```[tasklist]
### Roadmap
- [ ] Allow async components - https://github.com/nuxt/nuxt/issues/18500
- [ ] Client-side interactivity - https://github.com/nuxt/nuxt/issues/19765
- [ ] Hybrid/remote/streaming sources for server components - https://github.com/nuxt/nuxt/issues/12343
```

Twitch0125:
I think it'd make sense to have all the available auto imports for server route handlers be available for server components.

Hebilicious:
This looks promising. Here's my personal 2 cents for "server components".

_Disclaimer: I think this name in general is extremely confusing, due to the fact that it's already complicated to understand what's going on. This is not really a problem related to Nuxt, but React has the same problem with RSC, where most people don't really know how they differ from SSR components... Anyways here is some [reading](https://www.plasmic.app/blog/how-react-server-components-work) to get up to speed ... ._

## 1) More Unified Rendering API

Right now here's a recap of what is possible (correct me if I'm wrong) :

- Nitro [route rules](https://nitro.unjs.io/guide/routing) : It is possible to decide where and how a *page* will be rendered. This also allows the handling of more things, like caching rules, redirect, headers ... 
  - This works at the page level through a config file
  - This allows patterns like SPA/SSG/SSR/SSR+SWR
  
[...]

PR #20284, opened 2023-04-14
feat(nuxt): full scoped slots support for server components

### LINKED ISSUE #19765: support client interactivity within server components

### LINKED ISSUE #19884: Islands components not updating in production

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/19765
#19772 

### ❓ Type of change

### 📚 Description

hey there :wave: !

This PR is forked from my previous PR https://github.com/nuxt/nuxt/pull/19851 since it is ready for basic slots. 

This changes some things from #19851 to transform island components slots at build time instead and allow to have:
- conditionnal slots with v-if
- scoped slots
- slots with v-for

Let me know if you think that a case is missing about server-only components slots. 

Of course, https://github.com/nuxt/nuxt/pull/19851 will be closed if this PR is ready before it's merged

### How does it work

please check https://github.com/nuxt/nuxt/pull/19851 for previous implementation.

- kept the same idea of using `Teleport` in `NuxtIsland`
- transform all `<slot>` at build time, replacing it by a container div.
- transform all bindings and `v-for` directive to generate an array in its attribute so the slots within `NuxtIsland` can map over it to render
- we now use `nuxt-*` attributes for ids slots names and scoped data

### 📝 Checklist

- [ ] I have updated the documentation accordingly (todo).


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -28,15 +32,37 @@ export default defineComponent({
...
+    const ssrHTML = ref<string>(process.client ? getFragmentHTML(instance.vnode?.el ?? null).join('') ?? '<div></div>' : '<div></div>')
+    const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? randomUUID())
+    const availableSlots = computed(() => {
+      return [...ssrHTML.value.matchAll(SLOTNAME_RE)].map(m => m[1])
+    })
 
-    const html = ref<string>(process.client ? getFragmentHTML(instance?.vnode?.el).join('') ?? '<div></div>' : '<div></div>')
+    const html = computed(() => {
+      const currentSlots = Object.keys(slots)
+      return ssrHTML.value.replaceAll(SLOT_FALLBACK_RE, (full, slotName, content) => {
```
danielroe, line 50:
```suggestion
      return ssrHTML.value.replace(SLOT_FALLBACK_RE, (full, slotName, content) => {
```

packages/nuxt/src/components/runtime/server-component.ts
```diff
@@ -96,11 +107,57 @@ const NuxtServerComponent = defineComponent({
...
+    })
+    const availableSlots = computed(() => {
+      return [...res.data.value!.html.matchAll(SLOTNAME_RE)].map(m => m[1])
+    })
+
+    const html = computed(() => {
+      const currentSlots = Object.keys(slots)
+      return res.data.value!.html.replace(UID_ATTR, () => {
+        return `nuxt-ssr-component-uid="${randomUUID()}"`
+      }).replaceAll(SLOT_FALLBACK_RE, (full, slotName, content) => {
```
danielroe, line 129:
```suggestion
      }).replace(SLOT_FALLBACK_RE, (full, slotName, content) => {
```

danielroe, line 129:
```suggestion
      return res.data.value!.html
        .replace(UID_ATTR, () => `nuxt-ssr-component-uid="${randomUUID()}"`)
        .replace(SLOT_FALLBACK_RE, (full, slotName, content) => {
```

danielroe (approved):

danielroe:
This is great work. Let's iterate on it once merged, as I suspect there is some scope here for reduction in complexity, particularly once we move server components out of the payload.

rczdgbg:
When will v3.6.0 be released?

danielroe:
If you are looking for this PR, it is released in v3.5.0. But v3.6.0 should be released on or around June 20: https://github.com/nuxt/nuxt/milestone/5.

bernhardberger:
This doesn't allow us to have interactive client components in slots though, correct? Anything inside an island component is still rendered server side only without hydration?

Usually it's the wrapper components (components for Layouts and stuff) that have no need to be hydrated on the client side and cause JS size to explode. Is that something that will be possible one day? That would be huge.

danielroe:
It is indeed. An experimental version of it will be in v3.9.

PR #20329, opened 2023-04-17
feat(nuxt): add layer meta and env overrides to config types

danielroe:
### ❓ Type of change

### 📚 Description

This adds route metadata (via `$meta`) as well as per-environment overrides to Nuxt config typings, all implemented upstream in c12:

- https://github.com/unjs/c12/pull/64
- https://github.com/unjs/c12/pull/62

### 📝 Checklist


pi0 (approved):
💯

tonioriol:
Hey, not sure if this is the correct place to ask, but I'm not sure how the per-env configs takes precedence over non env specific configs. Could you give me some hints on how does that actually woks, or at least where can I check the implementation of the config to see it? 

```
$production: {
    routeRules: {
      '/**': process.env.CTF_PREVIEW === 'true' ? { ssr: false } : { isr: 86400 },
    },
  },
  $development: {
    routeRules: {
      '/**': { ssr: true },
    },
  },
  routeRules: {
    '/api/**': { cache: false, swr: false },
  },
```

For example, in this case, I'm not sure if the non env routeRules will be overridden by the catch-all rules in each env.

danielroe:
Yes, you can override the top level rule (`/**`) within a subpath (`/api/**`). Your example will end up merged into something like:

```js
  routeRules: {
    '/**': { ssr: true },
    '/api/**': { cache: false, swr: false },
  },
})
```

In future, it's best to ask a question on Discord or GitHub Discussions rather than commenting on a released PR.

tonioriol:
> Yes, you can override the top level rule (`/**`) within a subpath (`/api/**`). Your example will end up merged into something like:
> 
> ```js
> export default defineNuxtConfig({
>   routeRules: {
>     '/**': { ssr: true },
>     '/api/**': { cache: false, swr: false },
>   },
> })
> ```
> 
> In future, it's best to ask a question on Discord or GitHub Discussions rather than commenting on a released PR.

It's the only place on the internet I could find that remotely mentioned/documented. I'm actually looking for the implementation, to understand it better. But sorry about that, and thanks for answering.

PR #20345, opened 2023-04-18
fix(vite): avoid serving arbitrary file in vite-node middleware

antfu:
### 🔗 Linked issue

Code ported from Vite, await https://github.com/vitejs/vite/pull/12894 to refactor.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -143,7 +143,7 @@ function createViteNodeApp (ctx: ViteBuildContext, invalidates: Set<string> = ne
         throw createError({ statusCode: 400 })
       }
       if (isAbsolute(moduleId) && !isFileServingAllowed(moduleId, viteServer)) {
-        throw createError({ statusCode: 404 })
+        throw createError({ statusCode: 403 /* Restricted */ })
       }
       const module = await node.fetchModule(moduleId).catch((err) => {
         const errorData = {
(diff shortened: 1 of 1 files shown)


packages/vite/src/vite-node.ts
```diff
@@ -179,3 +183,63 @@ export async function initViteNodeServer (ctx: ViteBuildContext) {
...
+  if (server._fsDenyGlob(file)) { return false }
+
+  if (server.moduleGraph.safeModulesPath.has(file)) { return true }
+
+  if (server.config.server.fs.allow.some(dir => isParentDirectory(dir, file))) { return true }
+
+  return false
+}
+
+function isParentDirectory (dir: string, file: string): boolean {
```
pi0, line 237: Would be happy to see such utils going to `pathe/utils` would be much easier to share and also test...

packages/vite/src/vite-node.ts
```diff
@@ -179,3 +183,63 @@ export async function initViteNodeServer (ctx: ViteBuildContext) {
     `export { default } from ${JSON.stringify(pathToFileURL(manifestResolvedPath).href)}`
   )
 }
+
+/**
+ * The following code is ported from vite
+ * Awaits https://github.com/vitejs/vite/pull/12894
+ */
```
antfu, line 190: @pi0 I'd suggest we do nothing about the code below and wait for the Vite PR to be shipped (so we could remove them). Unless you are saying it could be a nice addition to pathe but not related to this PR.

pi0: Not specific for this PR of course. And only if finally vite can leverage from same util. Otherwise closest impl is best.

danielroe (approved):

PR #20360, opened 2023-04-18
fix(test-utils): import jest.fn from `@jest/globals`

danielroe:
### ❓ Type of change

### 📚 Description

This uses `@jest/globals` to import `jest.fn` from: https://jestjs.io/docs/api.

We might consider not adding as a dep given how heavy it is. wdyt @antfu?

### 📝 Checklist


packages/test-utils/src/setup/jest.ts
```diff
@@ -1,8 +1,7 @@
 import type { TestHooks } from '../types'
 
 export default async function setupJest (hooks: TestHooks) {
-  // @ts-expect-error jest is not a dependency
-  const jest = await import('jest')
+  const jest = await import('@jest/globals')
```
danielroe, line 4:
```suggestion
  const { jest } = await import('@jest/globals')
```

antfu:
Looks ok. Do we need to explicitly external it?

danielroe:
I've added to (optional) peerDeps - but again, happy if you have a different suggestion!

PR #20367, opened 2023-04-18
feat(nuxt): add experimental `typedPages` option

posva:
### ❓ Type of change

### 📝 Checklist


packages/kit/src/build.ts
```diff
@@ -32,7 +32,14 @@ export interface ExtendConfigOptions {
 export interface ExtendWebpackConfigOptions extends ExtendConfigOptions {
 }
 
-export interface ExtendViteConfigOptions extends ExtendConfigOptions {}
+export interface ExtendViteConfigOptions extends ExtendConfigOptions {
+  /**
+   * Prepends the plugin to the array with `unshit()` instead of `push()`. In practice this shouldn't be necessary, but
```
xiBread, line 37:
```suggestion
   * Prepends the plugin to the array with `unshift()` instead of `push()`. In practice this shouldn't be necessary, but
```
😄

olivermrose (changes requested):

packages/kit/src/build.ts
posva: oops! Thanks!

danielroe (approved):

antfu (commented):
Looks awesome 🔥! Would love to play with it in real apps!

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -81,12 +81,12 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
     props: {
       // Routing
       to: {
-        type: [String, Object] as PropType<string | RouteLocationRaw>,
+        type: [String, Object] as PropType<RouteLocationRaw>,
         default: undefined,
         required: false
```
posva, line 86: BTW are these intentional? It should be the exact same without them

packages/nuxt/src/pages/module.ts
```diff
@@ -67,6 +73,70 @@ export default defineNuxtModule({
...
+            if (page.meta) {
+              route.addToMeta(page.meta)
+            }
+            if (page.alias) {
+              route.addAlias(page.alias)
+            }
+            if (page.name) {
+              route.name = page.name
+            }
+            // TODO: implement redirect support
```
posva, line 100: I think I will need to add this before it gets merged

packages/nuxt/src/pages/module.ts
```diff
@@ -67,6 +73,70 @@ export default defineNuxtModule({
...
+          }
+        }
+      }
+
+      nuxt.hook('prepare:types', ({ references }) => {
+        // This file will be generated by unplugin-vue-router
+        references.push({ path: declarationFile })
+      })
+
+      const context = createRoutesContext(resolveOptions(options))
```
posva, line 118: I need to take a look, I think the options should be resolved internally anyway 🤔

packages/nuxt/src/pages/module.ts
```diff
@@ -67,6 +73,70 @@ export default defineNuxtModule({
...
+      const dtsFile = resolve(nuxt.options.buildDir, declarationFile)
+      await mkdir(dirname(dtsFile), { recursive: true })
+      await context.scanPages(false)
+
+      if (nuxt.options._prepare) {
+        // TODO: could we generate this from context instead?
```
posva, line 124: I could add a method to expose the generation of the dts content. I imagine that having an option not to write to the filesystem could also be helpful in this case. Let me know what you think

packages/nuxt/src/pages/module.ts
danielroe: Will it make a difference to the types? Because at the moment we are not using the unplugin for actual routes generation (until it can support route metadata).

packages/nuxt/src/pages/module.ts
danielroe: That would be perfect 👌

packages/nuxt/src/app/components/nuxt-link.ts
danielroe: Let's get rid of them then 😆

packages/nuxt/src/pages/module.ts
posva: you are right, it doesn't make any difference in terms of types

PR #20387, opened 2023-04-19
feat(nuxi): allow greater control of `nuxi analyze` from cli

antfu:
### ❓ Type of change

### 📚 Description

To provide better integrations with DevTools, the follow changes are made:

- Introduce `analyzeDir` option for modules to write additional content
- Introduce `build:analyze:done` hook for modules to write/read the final report
- Introduce `--no-serve` cli arg for `nuxi analyze`
- Introduce `--name` cli arg for `nuxi analyze` to allow multiple analysis report to be present (for comparison in the future)
- `nuxi analyze` will override `buildDir` `outDir` to the `analyzeDir`

### 📝 Checklist


packages/schema/src/config/common.ts
```diff
@@ -139,6 +139,15 @@ export default defineUntypedSchema({
     ]
   },
 
+  /**
+   * The directory where Nuxt will store the generated files when running `nuxt analyze`.
+   *
+   * If a relative path is specified, it will be relative to your `rootDir`.
+   */
+  analyzeDir: {
+    $resolve: async (val, get) => resolve(await get('rootDir'), val || '.nuxt-analyze')
```
pi0, line 148: This dir is not gitignored by default in any starter template. Cannot we use buildDir `.nuxt/analyze`?

antfu: I was going `.nuxt/analyze`, but changed it to this as `.nuxt` will be cleaned out on every start, while analyzes should better be persisted so we could compare with different builds.

antfu: We could add it to gitignore after is get shipped.

antfu: This is how DevTools likes like atm:
image

pi0: `.nuxt` (buildDir) should finally not be cleaned up entirely. Because this way we are also removing `.nitro/cache` and other files that can be particularly useful between builds to be kept. I think would be worth fixing this and not introducing another top-level directory.

pi0: We can also use `node_modules/.nuxt/analzye` as another workaround solution in meanwhile. Would that work?

antfu: I will explore on it :)

packages/nuxi/src/utils/fs.ts
```diff
@@ -12,11 +12,24 @@ export async function exists (path: string) {
...
+      if (!exclude.includes(name)) {
+        await fsp.rm(join(path, name), { recursive: true, force: true })
+      }
+    }))
+  }
   await fsp.mkdir(path, { recursive: true })
 }
 
+export function clearBuildDir (path: string) {
+  return clearDir(path, ['cache', 'analyze'])
```
pi0, line 30: LOVE IT!

packages/nuxi/src/commands/analyze.ts
```diff
@@ -11,62 +11,99 @@ import { defineNuxtCommand } from './index'
 export default defineNuxtCommand({
   meta: {
     name: 'analyze',
-    usage: 'npx nuxi analyze [--log-level] [rootDir]',
+    usage: 'npx nuxi analyze [--log-level] [--name] [--no-serve] [rootDir]',
```
danielroe, line 14: would it make sense to accept these as programmatic options rather than CLI options (for `analyzeCommand.invoke(args, options)`)? We do that for `dev` and could also do that here. Not sure if that would work for devtools. If so, it might avoid confusion when seeing help printed for the analyze command. (But if you think either option might be used by end users, then we can of course keep it here...)

https://github.com/nuxt/nuxt/blob/4c200e4a2a7c3472153b6fac3e81d4e911c9a50a/packages/nuxi/src/commands/dev.ts#L24

antfu: That would also work for me. The problem is that we need to call it from a subprocess since we are starting a new build. But it could be done from DevTools by providing a CLI wrapper to invoke it - or maybe that would be better for devtools to iterate.

danielroe: What do you think would be better?

antfu: I would still need this PR in order to support DevTools usage. Meanwhile, I will try to move some logic to DevTools.

danielroe: Oh, absolutely. I'm just wondering whether we keep these as CLI arguments or programmatic options.

packages/schema/src/config/vite.ts
```diff
@@ -70,6 +70,14 @@ export default defineUntypedSchema({
...
     },
     server: {
+      watch: {
+        ignored: {
+          $resolve: async (val, get) => [
+            await get('analyzeDir'),
+            ...val ?? []
+          ]
+        }
+      },
```
danielroe, line 80: I don't think we need to add this here, as it should be present from:

https://github.com/nuxt/nuxt/blob/9878c5b6e4b3d7241910a347412a0936dbc6a3cf/packages/vite/src/vite.ts#L88-L96

danielroe:
```suggestion
```

antfu: Yeah, good point, I forgot to remove it, Thanks!

danielroe (approved):
Looks good to me! 👌

Let's merge and test as we have a bit of time to test on edge in line with devtools feature iteration.

PR #20391, opened 2023-04-19
feat(nuxt): add support for `routeRules` defined within pages

danielroe:
### ❓ Type of change

### 📚 Description

This adds the ability to define `routeRules` inline, within the pages directory.

#### Example usage:

```vue
<script setup lang="ts">
defineRouteRules({
  ssr: false
  prerender: true,
  headers: {
    'X-Test': 'defined within a page'
  },
})
</script>
```

**TODO**:
- [ ] feedback on API
- [x] implement HMR support in dev mode (currently nitro's runtimeConfig is frozen so we can't update it there). I might consider overriding `event.context._nitro.routeRules` but this would also miss out `getRouteRulesForPath` being called directly 🤔 
- [ ] documentation updated

### 📝 Checklist


Atinux:
I love it!

packages/nuxt/src/pages/module.ts
```diff
@@ -6,10 +6,13 @@ import escapeRE from 'escape-string-regexp'
 import { joinURL } from 'ufo'
 import type { NuxtApp, NuxtPage } from 'nuxt/schema'
 
+import type { NitroDynamicConfig, NitroRouteConfig } from 'nitropack'
```
danielroe, line 9:
```suggestion
import type { NitroRouteConfig } from 'nitropack'
```

packages/nuxt/src/pages/plugins/route-rules.ts
```diff
@@ -0,0 +1,56 @@
...
+
+interface RouteRuleExtractorPluginOptions {
+  routeRules: Record<string, any>
+  pageMap: Record<string, string>
+  updateConfig: (() => void | Promise<void>)
+}
+
+export const routeRuleExtractorPlugin = createUnplugin((options: RouteRuleExtractorPluginOptions) => {
+  return {
+    name: 'test',
```
antfu, line 20: 😜

packages/nuxt/src/pages/plugins/route-rules.ts
```diff
@@ -0,0 +1,56 @@
...
+    async transform (code, id) {
+      if (!ROUTE_RULE_RE.test(code)) { return null }
+      const { pathname } = parseURL(decodeURIComponent(pathToFileURL(normalize(id)).href))
+
+      let rule = {}
+
+      walk(this.parse(code, {
+        sourceType: 'module',
+        ecmaVersion: 'latest'
+      }) as Node, {
```
antfu, line 34: Is it possible to share one plugin for pageMeta and avoid parsing twice?

danielroe: I've recently updated the page meta plugin so it only processes files with the query string, which shouldn't be processed with this plugin.

But the content will be nearly identical.

I wonder if there's a way to cache the AST?

packages/nuxt/src/pages/plugins/route-rules.ts
Atinux, line 20:
```suggestion
    name: 'routeRules',
```

packages/nuxt/src/pages/module.ts
```diff
@@ -144,10 +147,57 @@ export default defineNuxtModule({
...
+      pageMap: {} as Record<string, string>,
+      updateConfig: (() => {}) as (() => void | Promise<void>)
+    }
+
+    function pagesToMap (pages: NuxtPage[], prefix = ''): Array<[id: string, route: string]> {
+      const mappedPages = [] as Array<[id: string, route: string]>
+      for (const page of pages) {
+        if (!page.file) { continue }
+        const glob = pathToNitroGlob(prefix + page.path)
+        if (!glob) { continue }
```
Atinux, line 166: Shall we log that the routeRule could not be generated if glob == null?

Hebilicious:
I was just thinking about this, would it be possible that we generate some kind of artifact file that reflects the route rules state ? 
Something like a `routeRules.json`, maybe in the `app` or `.nuxt` directory. I think it could be helpful to know what's going on as the configuration can come from many different places with this change (ie defineRouteRules can be used through modules/layers). It will be useful to debug what happens, and know how the rules are generated.

Perhaps also add a tab to the `devtools` ?

danielroe:
@pi0 Would love your thoughts here. In dev mode, the _first_ request to any page with inline route rules (that is, the one that triggers a nitro reload) causes a 404 response.

This is also the same issue that https://github.com/nuxt/nuxt/pull/22494 works around - when nitro reloads its dev server it seems to return a 404 to any simultaneous request, rather than 'waiting' for the reload and then returning the response.

tsconfig.json
```diff
@@ -26,11 +26,8 @@
       "#internal/nitro": [
         "./node_modules/nitropack/dist/runtime"
       ],
-      "#internal/nitro/app": [
-        "./node_modules/nitropack/dist/runtime/app"
-      ],
-      "#internal/nitro/utils": [
-        "./node_modules/nitropack/dist/runtime/utils"
+      "#internal/nitro/*": [
```
pi0, line 29: Currently, (sadly) I cannot make any refactors to `/app` and `/utils` (which are explicit **internals**) because Nuxt is depending on them. This is dangerous and should be avoided. Please keep them explicit to these two and let's not extend to use anymore.

tsconfig.json
```diff
@@ -26,11 +26,8 @@
...
+        "./node_modules/nitropack/dist/runtime/*"
```
pi0, line 30: Same for here about assumptions of the internal runtime directory structure. It should be avoided when possible.

test/bundle.test.ts
```diff
@@ -32,7 +32,7 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
     const serverDir = join(rootDir, '.output/server')
 
     const serverStats = await analyzeSizes(['**/*.mjs', '!node_modules'], serverDir)
-    expect.soft(roundToKilobytes(serverStats.totalBytes)).toMatchInlineSnapshot('"64.6k"')
+    expect.soft(roundToKilobytes(serverStats.totalBytes)).toMatchInlineSnapshot('"64.7k"')
```
danielroe, line 35: @pi0 We can directly use `#internal/nitro` but it does have a bundle size cost.

pi0: No worries i think we can check possible tree-shaking issues later. Since we were already using `/app`, it should be fine to keep because anyway i cannot break in a non major version of nitro 🙈 (https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/runtime/nitro/renderer.ts)

packages/nuxt/src/pages/module.ts
```diff
@@ -237,10 +240,70 @@ export default defineNuxtModule({
...
+    // Extracted route rules defined inline in pages
+    const inlineRules = {} as { [glob: string]: NitroRouteConfig }
+    // User-provided route rules from `nuxt.config`
+    let baseRules: { [glob: string]: NitroRouteConfig } | undefined
+    nuxt.hook('nitro:config', (config) => { baseRules = config.routeRules })
+
+    // Allow telling Nitro to reload route rules
+    let updateRouteConfig: () => void | Promise<void>
+    nuxt.hook('nitro:init', (nitro) => {
+      updateRouteConfig = () => nitro.updateConfig({ routeRules: defu(inlineRules, baseRules) })
```
harlan-zw, line 261: One concern would be if other modules are also updating the route rules at runtime, it _seems_ like they would get overridden?

I think the scope for this happening, especially just in development, is pretty low.

Maybe worth merging in `nitro.options.routeRules` though?

harlan-zw:
Nice :+1:

packages/nuxt/src/pages/module.ts
danielroe: I think we can use `nitro.options._config.routeRules`.

cc: @pi0 - thoughts on how to avoid clobbering pre-existing route rules?

pi0: I guess merging with `nuxt.options.routeRules` would be safer as it contains the latest user (nuxt module) modified rules. In context of Nuxt with top level `routeRules`, it should be main source of trust.

danielroe: nitro:config hook is last place for modules to modify nitro config, so that might be safer. (Hence current and previous implementations.)

Otherwise when we update the rules we overwrite intervening changes.

Or maybe use proxy for runtimeConfig/routeRules in resolved options that automatically reloads both within nitro?

pi0: Modules might need to also apply route rules afterward as well (similar to runtimeConfig). We can either sync references so that nitro and nuxt modules update same object or proxy is also good idea for auto reload 👍🏼 (a possible downside is that sometimes we need to batch changes, advising to prefer update config hook is safer)

pi0:
This is a really nice initiative! I would recommend to enable it behind an experimental flag until next minor release if want to move forward faster. Coupling with nuxi to support HMR on `nuxt.config` and also route rule / event handler meta inlined support from Nitro, might affect this implementation and possibly introduce regressions affecting user experience.

packages/nuxt/src/pages/module.ts
danielroe: How about instead letting nitro keep track of current state and only merge in/reload the _updated_ rules? That way we wouldn't overwrite anything.

Atinux:
It is looking great 😍 

Agree to test with an experimental flag until Nuxt 3.8

PR #20396, opened 2023-04-19
fix(core,kit): ensure module transpilation paths are dirs

harlan-zw:
### 🔗 Linked issue

- https://stackblitz.com/edit/nuxt-starter-nb7fac?file=nuxt.config.ts
- https://stackblitz.com/edit/nuxt-starter-f7ehnh?file=nuxt.config.ts

### ❓ Type of change

### 📚 Description

When Nuxt installs modules it adds the modules entire directory to be transpiled. 

When using an absolute path to a module such as `/home/user/my-project/node_modules/my-module/dist/module.mjs` though, this transpiling won't take effect, as we're only targeting the module file itself. (see stackblitz demos)

To get around this we check if the path is to a file, if so we use the directory it's in. 

This is fairly edge-case as you can for most modules you would just use the module name itself. ]

Normally I would just say that this should be fixed in userland, but in this instance, it creates quite confusing bugs that aren't easy to debug, so I think it's worth fixing in Nuxt.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/install.ts
+++ b/packages/kit/src/module/install.ts
@@ -1,6 +1,6 @@
 import { lstatSync } from 'node:fs'
 import type { Nuxt, NuxtModule } from '@nuxt/schema'
-import { dirname } from 'pathe'
+import { dirname, isAbsolute } from 'pathe'
 import { isNuxt2 } from '../compatibility'
 import { useNuxt } from '../context'
 import { requireModule, resolveModule } from '../internal/cjs'
@@ -24,7 +24,7 @@ export async function installModule (moduleToInstall: string | NuxtModule, _inli
   }
 
   if (typeof moduleToInstall === 'string') {
-    nuxt.options.build.transpile.push(lstatSync(moduleToInstall).isFile() ? dirname(moduleToInstall) : moduleToInstall)
+    nuxt.options.build.transpile.push(normaliseModuleTranspilePath(moduleToInstall))
   }
 
   nuxt.options._installedModules = nuxt.options._installedModules || []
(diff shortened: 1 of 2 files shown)


packages/kit/src/module/install.ts
```diff
@@ -22,7 +23,7 @@ export async function installModule (moduleToInstall: string | NuxtModule, _inli
   }
 
   if (typeof moduleToInstall === 'string') {
-    nuxt.options.build.transpile.push(moduleToInstall)
+    nuxt.options.build.transpile.push(extname(moduleToInstall) ? dirname(moduleToInstall) : moduleToInstall)
```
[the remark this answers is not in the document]
harlan-zw: Much better idea, thanks. Is it safe to use `node:fs` within the `installModule` @nuxt/kit util :thinking:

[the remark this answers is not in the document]
danielroe: yes - we can rely on this being run in node 👍

packages/kit/src/module/install.ts
```diff
@@ -35,6 +37,17 @@ export async function installModule (moduleToInstall: string | NuxtModule, _inli
 
 // --- Internal ---
 
+export const normaliseModuleTranspilePath = (p: string) => {
```
danielroe, line 40: probably we want this to be US English. 🥲

```suggestion
export const normalizeModuleTranspilePath = (p: string) => {
```

harlan-zw: haha yup, always catches me out

danielroe: (it'll have to be the other files too)

danielroe (approved):

PR #20408, opened 2023-04-20
docs: add cdn proxy section

clemcode:
### ❓ Type of change

### 📚 Description
Some Cloudflare proxy options can lead to unexpected behavior for Nuxt. This PR integrates in the 3.x documentation the instructions documented on the [2.x docs](https://nuxtjs.org/deployments/cloudflare#deploy-nuxt-on-cloudflare)


docs/1.getting-started/10.deployment.md
```diff
@@ -149,3 +149,17 @@ Nuxt 3 can be deployed to several cloud providers with a minimal amount of confi
 - :icon{name="simple-icons:render" class="h-5 w-4 inline mb-2"} [Render](https://nitro.unjs.io/deploy/providers/render)
 - :icon{name="ph:cloud-duotone" class="h-5 w-4 inline mb-2"} [Stormkit](https://nitro.unjs.io/deploy/providers/stormkit)
 - :icon{name="simple-icons:vercel" class="h-5 w-4 inline mb-2 text-black dark:text-white"} [Vercel](https://nitro.unjs.io/deploy/providers/vercel)
+
+## Cloudflare Proxy
+
+In most cases, Nuxt can work with third party content that is not generated or created by Nuxt itself. But sometimes such content can cause problems, especially Cloudflare's "Minification and Security Options".
```
DamianGlowala, line 155: When _third party_ is used as a modifier, it should be hyphenated.
```suggestion
In most cases, Nuxt can work with third-party content that is not generated or created by Nuxt itself. But sometimes such content can cause problems, especially Cloudflare's "Minification and Security Options".
```

danielroe:
It feels like maybe this should be on a provider-specific page, maybe in the Nitro docs? Or, if there is more content like this in the docs, perhaps we could assemble it into a page on 'deployment caveats' or similar?

clemcode:
I agree this is not an ideal place. Is it possible it also affects Nitro independently of Nuxt? If so, happy to make the PR on the Nitro repository. 
For now, it is the only caveat of this type that I witnessed, but if more of these should arise, a dedicated page on the Nuxt docs seems the way to go.

docs/1.getting-started/10.deployment.md
```diff
@@ -149,3 +149,17 @@ Nuxt 3 can be deployed to several cloud providers with a minimal amount of confi
 - :icon{name="simple-icons:render" class="h-5 w-4 inline mb-2"} [Render](https://nitro.unjs.io/deploy/providers/render)
 - :icon{name="ph:cloud-duotone" class="h-5 w-4 inline mb-2"} [Stormkit](https://nitro.unjs.io/deploy/providers/stormkit)
 - :icon{name="simple-icons:vercel" class="h-5 w-4 inline mb-2 text-black dark:text-white"} [Vercel](https://nitro.unjs.io/deploy/providers/vercel)
+
+## Cloudflare Proxy
```
pi0, line 153: I think we could make it more _CDN Proxy_ section to open for more future advice about CDN deployments. Also other CDNs have similar features that need to be disabled.

clemcode, line 153:
```suggestion
## CDN Proxy
```

Atinux:
I could move this later in a deployment best practices / caveats but I think right now it is already good to have this in docs

atinux (approved):

PR #20440, opened 2023-04-21
fix(types): app config types extending from schema

atinux:
### ❓ Type of change

### 📚 Description

Related also to https://github.com/nuxt-modules/icon/issues/62

If we use the schema to specify the app config types, I expect the types to be in the app config types (not only input).

Though, we may want to use DeepPartial since those values may not exists in the `app.config` object.

Happy to get your feedback @danielroe and @pi0 

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -158,4 +158,4 @@ export interface NuxtAppConfig {
   keepalive: boolean | KeepAliveProps
 }
 
-export interface AppConfig { }
+export interface AppConfig extends CustomAppConfig { }


danielroe:
We're currently doing the merging here:

https://github.com/nuxt/nuxt/blob/f366ab4eb89c6f270ee1a0ae6e3f64d203a146c4/packages/nuxt/src/core/templates.ts#L213-L245

We also have a type test for it:

https://github.com/nuxt/nuxt/blob/574a5616b43d0caeddd97ddea3abf7703efec295/test/fixtures/basic/types.ts#L297

I believe the issue is with `nuxt-config-schema` - when disabled in `nuxt-icon` the types appear to work as expected.

PR #20472, opened 2023-04-24
fix(deps): relax upper node version constraint

### LINKED ISSUE #20445: Nuxt is not compatible with Node.js 20
### Environment

Nuxt project info:

- Node Version:     v20.0.0
- Nuxt Version:     3.4.2
- Nitro Version:    2.3.3
- Package Manager:  pnpm@8.3.1
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

1. Install Node.js v20.0.0
2. run `pnpx dlx nuxi init project`

### Describe the bug

```
Nuxi 3.4.2

WARN Current version of Node.js (20.0.0) is unsupported and might cause issues.
     Please upgrade to a compatible version (^14.18.0 || ^16.10.0 || ^17.0.0 || ^18.0.0 || ^19.0.0).
```

### Additional context

_No response_

### Logs

_No response_

dargmuesli:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/20445

### ❓ Type of change

### 📚 Description

Resolves https://github.com/nuxt/nuxt/issues/20445

### 📝 Checklist


.github/workflows/ci.yml
dargmuesli, line 1: Should `test-fixtures` run on Node 18 or 20 as well? Currently they run using Node 16, which is in maintenance state and EOL [2023-09-11](https://nodejs.org/en/blog/announcements/nodejs16-eol/).

danielroe: we deliberately run them on the lowest supported version of Node to ensure there are no regressions for users on that version.

package.json
```diff
@@ -81,6 +81,6 @@
   },
   "packageManager": "pnpm@8.3.1",
   "engines": {
-    "node": "^14.18.0 || ^16.10.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
+    "node": "^14.18.0 || ^16.10.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0"
```
danielroe, line 84: What about adding this as `>=20.0.0` for forward compatibility?

.github/workflows/ci.yml
dargmuesli: Fair!

package.json
dargmuesli: Hmm, wouldn't that mean that e.g. Node 21 would count as supported as well once it is released?
I remember https://github.com/nuxt/nuxt/pull/19687 and that you said only LTS versions should be supported officially if I remember correctly. That's why I chose to continue the caret notation.
Using `>=` would give the "benefit" of not having to create PRs like this for every new Node version, do I understand that correctly? Would that be desirable?

danielroe: It's more that I wouldn't want - in some future world where Node 36 has been released - for Nuxt 3.5 to be wholly unusable on it. I don't think it would count as an endorsement or 'supported' usage. I'm more interested in using the `engines.node` field to disallow _lower_ versions that have known issues with Nuxt.

dargmuesli: I see! So could we then shorten your suggestion to `"^14.18.0 || >=16.10.0"` or would I then be doing the semver we're speaking about wrong?

danielroe: Yep! Perfect. We can update the lower bound as needed.

danielroe (approved):

dargmuesli:
Always a pleasure working with you, @danielroe! :heart:

PR #20496, opened 2023-04-25
fix(nuxt): return `RenderResponse` for redirects

### LINKED ISSUE #20294: Page that gets redirected via navigateTo in a middleware still gets executed
### Environment

| Key              | Value       |
|------------------|-------------|
| Operating System | Linux       |
| Node Version     | v16.14.2    |
| Nuxt Version     | 3.4.1       |
| Nitro Version    | 2.3.3       |
| Package Manager  | npm@7.17.0  |
| Builder          | vite        |
| User Config      | -           |
| Runtime Modules  | -           |
| Build Modules    | -           |

### Reproduction

https://stackblitz.com/edit/nuxt-starter-7ebm3j

Go to the route `/i-get-redirected` 

### Describe the bug

There are two pages. 

1. i-get-redirected.vue
2. i-get-displayed.vue

`i-get-redirected` gets redirected to `i-get-displayed` via a global route middleware `redirect.global.ts`.  So I expect the code from the page `i-get-redirected` not to be executed. But it does.

image

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #20432: Middleware, useCookie & trailing slash => Cannot set headers after they are sent to the client
### Environment

```
- Operating System: Darwin
- Node Version:     v18.16.0
- Nuxt Version:     3.4.2
- Nitro Version:    2.3.3
- Package Manager:  pnpm@8.3.0
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

StackBlitz: https://stackblitz.com/edit/nuxt-starter-kiqw1y
Demo on Vercel: https://nuxt-cookie.vercel.app

Compare:
1. https://nuxt-cookie.vercel.app/dashboard
2. https://nuxt-cookie.vercel.app/dashboard/

### Describe the bug

If the `useCookie` composable is used on a page or component, and middleware is added to protect that page, an error message of `Cannot set headers after they are sent to the client` may occur. The use of a trailing slash typically causes this error, but probably it can also happen without it in some cases.

The issue becomes more serious on Vercel, as users may encounter a 500 error in their browsers instead of just a silent error in the console during development.

One more thing to consider is whether it's normal for the setup function to run even when we have middleware that blocks access to a page.

### Logs

Zrzut ekranu 2023-04-21 o 08 02 47

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/20244#discussion_r1165271853
resolves https://github.com/nuxt/nuxt/issues/20294
resolves nuxt/nuxt#20432

### ❓ Type of change

### 📚 Description

This adds an internal way for Nuxt composables to predetermine the type of response from the render handler, rather than directly interacting with `event.node.res`. It then updates `navigateTo` handling to use this internal `._renderResponse` property for indicating status code, etc. There is space here for improving this, including maybe using h3's `sendRedirect` method behind the scenes in nitro so we don't have to generate the HTML body, for example.

The hook `app:redirected` was also only added for the purpose of setting cookies before immediately returning a response. But now `app:rendered` is just as good a response, and we might consider deprecating `app:redirected` in favour of `app:rendered`. Thoughts welcome.

Finally, we resolve an issue with setup functions loading even after a redirect by switching from `beforeEach` to `afterEach`, and add a test fixture.

### 📝 Checklist


packages/nuxt/src/app/composables/router.ts
```diff
@@ -111,17 +111,26 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
...
-        .then(() => sendRedirect(nuxtApp.ssrContext!.event, redirectLocation, options?.redirectCode || 302))
-        .then(() => inMiddleware ? /* abort route navigation */ false : undefined)
+      const location = isExternal ? toPath : joinURL(useRuntimeConfig().app.baseURL, fullPath)
+
+      async function redirect () {
+        // TODO: consider deprecating in favour of `app:rendered` and removing
+        await nuxtApp.callHook('app:redirected')
+        const encodedLoc = location.replace(/"/g, '%22')
+        nuxtApp.ssrContext!._renderResponse = {
+          statusCode: sanitizeStatusCode(options?.redirectCode || 302, 302),
```
pi0, line 123: Finally nitro renderer should sanitize, wouldn't it?

pi0 (approved):
Smart solution love it 💯

packages/nuxt/src/app/composables/router.ts
danielroe: if that's possible, I'd love it. but I don't know how we can communicate that it's a redirect so default should be 302?

PR #20497, opened 2023-04-25
fix(nuxt): throw 404 when accessing `/__nuxt_error` directly

### LINKED ISSUE #20490: Remove "/__nuxt_error" hardcoded route (make it 404)
### Environment

Node.js 18.15.0
Nuxt 3.3.1

### Reproduction

Fresh Nuxt 3 install, go to **http://localhost:3000/__nuxt_error** and see Nuxt server error page.

### Describe the bug

I dont want users to see that page, becouse they can just go to mywebsite.com/__nuxt_error and instead of 404 see error messages.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #20490

### ❓ Type of change

### 📚 Description

Rather than throw a more specific error, I think it is fair that `/__nuxt_error` should return a 404.

The error is constructed in the same way as upstream 404s (but I would suggest updating both to use `message` rather than `statusMessage`).

https://github.com/nuxt/nuxt/blob/f366ab4eb89c6f270ee1a0ae6e3f64d203a146c4/packages/nuxt/src/app/plugins/router.ts#L245-L248

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -308,13 +308,13 @@ export default defineRenderHandler(async (event) => {
       NO_SCRIPTS
         ? undefined
         : (_PAYLOAD_EXTRACTION
-          ? process.env.NUXT_JSON_PAYLOADS
-            ? renderPayloadJsonScript({ id: '__NUXT_DATA__', ssrContext, data: splitPayload(ssrContext).initial, src: payloadURL })
-            : renderPayloadScript({ ssrContext, data: splitPayload(ssrContext).initial, src: payloadURL })
-          : process.env.NUXT_JSON_PAYLOADS
-            ? renderPayloadJsonScript({ id: '__NUXT_DATA__', ssrContext, data: ssrContext.payload })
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -180,7 +180,11 @@ export default defineRenderHandler(async (event) => {
   }
 
   if (ssrError && event.node.req.socket.readyState !== 'readOnly' /* direct request */) {
-    throw createError('Cannot directly render error page!')
+    throw createError({
+      statusCode: 404,
+      message: process.dev ? '[nuxt] Cannot directly render error page.' : undefined,
+      statusMessage: 'Page Not Found: /__nuxt_error'
```
pi0, line 186: I think we can remove it too (at least for prod) http clients have default message

pi0 (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
danielroe: I'll follow up with another PR to do that, for clearer signage in changelog 👍

Issue #20435, opened 2023-04-21
Nuxt dev server with https is broken on Node 18+


Atinux:
Reproduction:

```bash
npx nuxi init@latest my-app
cd my-app
pnpm i
npx nuxi dev --https
```

![CleanShot 2023-04-21 at 10.47.33@2x.png](https://volta.s3.fr-par.scw.cloud/Clean_Shot_2023_04_21_at_10_47_33_2x_83c54a9907.png)

I also notice that we don't support the `devServer.https: true` option, any reason for that?

Atinux:
Actually the `devServer.https: true` option works but display a TS error:
![CleanShot 2023-04-21 at 10.54.39@2x.png](https://volta.s3.fr-par.scw.cloud/Clean_Shot_2023_04_21_at_10_54_39_2x_53b0a27386.png)

sshwy:
facing the same bug

danielroe:
This is now working for me.

pi0:
Worth to mention, the final fix for this is that Nuxi CLI uses an internal `http` server + public `https` server which proxies to the internal one and vite dev server directly uses internal server.

In case of using `--https --no-fork`, we would need some Node.js flags to allow unauthorized keys (written in CLI output log)

PR #20498, opened 2023-04-25
fix(nuxi,schema): support `devServer.https: true`

danielroe:
### 🔗 Linked issue

nuxt/nuxt#20435

### ❓ Type of change

### 📚 Description

This adds support for setting `devServer.https` to `true` (rather than just an empty object) to enable SSL from within the `nuxt.config`.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/dev.ts
+++ b/packages/schema/src/config/dev.ts
@@ -18,7 +18,7 @@ export default defineUntypedSchema({
      * ```
      *
      *
-     * @type {false | { key: string; cert: string }}
+     * @type {boolean | { key: string; cert: string }}
      *
      */
     https: false,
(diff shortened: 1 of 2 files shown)


pi0 (approved):

packages/schema/src/config/dev.ts
```diff
@@ -18,7 +18,7 @@ export default defineUntypedSchema({
      * ```
      *
      *
-     * @type {false | { key: string; cert: string }}
+     * @type {boolean | { key: string; cert: string }}
      *
      */
     https: false,
```
pi0, line 24: This couldn’t be true after normalization. Please double check. Self generated cert is available for nuxt.options https when enabled

danielroe: Would you explain more? What do you mean?

PR #20511, opened 2023-04-25
fix(nuxt): call `app:error` in SSR before rendering error page

### LINKED ISSUE #20113: Error hooks don't catch html render error on SSR
### Environment

```
- Operating System: `Linux`
- Nuxt Version:     `3.3.3`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

### Reproduction

I noticed this issue only on ssr navigation, I didn't tested in depth on client navigation. Tried with both app:error and vue:error hooks.

Navigate to /script-error and check the console, app:error and vue:error hooks are firing.
Navigate to /html-error and see neither of the hooks firing.

Code: https://stackblitz.com/edit/github-wq5s8k

### Describe the bug

 While writing a custom error reporter I noticed that I could not catch when the SSR fails to render a route because of the template parsing (in my case a bad v-if). I expected at least app:error hook to fire in this situation.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #20183: New cookie value lost if template has error
### Environment

Nuxt project info:               

- Nuxt Version:     `3.3.3`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-3fmhy3 (you need to download it in order for cookies to work)

### Describe the bug

Saving a new cookie value does not work when the template has an error (for example, getting an undefined nested field). And it doesn't matter in which of the files we use cookieref (in my case, this is the pinia store) - cookies are not saved :(

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves nuxt/nuxt#20183
resolves https://github.com/nuxt/nuxt/issues/20113

### ❓ Type of change

### 📚 Description

This PR calls `app:error` when there's an error rendering to string and we'll directly return an error page. (Previously, it _only_ caught fatal errors in Nuxt plugins.) That allows us to hook into this for setting cookies, for example, as well as to provide some consistency when handling errors with hooks registered in nuxt plugins, and so on.

The one downside is that fatal errors in Nuxt plugins will now result in _two_ calls to `app:error`.

Alternatively, we could call `app:rendered` here - or add another hook (`app:render:error`?).

### 📝 Checklist

### DIFF
--- a/docs/1.getting-started/8.error-handling.md
+++ b/docs/1.getting-started/8.error-handling.md
@@ -41,6 +41,7 @@ This includes:
 
 * running Nuxt plugins
 * processing `app:created` and `app:beforeMount` hooks
+* rendering your Vue app to HTML (on the server)
 * mounting the app (on client-side), though you should handle this case with `onErrorCaptured` or with `vue:error`
 * processing the `app:mounted` hook
 
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -48,11 +48,12 @@ export function useCookie <T = string | null | undefined> (name: string, _opts?:
       }
     }
     const unhook = nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
-    nuxtApp.hooks.hookOnce('app:redirected', () => {
-      // don't write cookie subsequently when app:rendered is called
-      unhook()
+    const writeAndUnhook = () => {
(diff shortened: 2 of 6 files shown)


pi0 (approved):
💯

PR #20525, opened 2023-04-26
feat(kit): support passing getter to `addVitePlugin` and `addWebpackPlugin`

antfu:
### ❓ Type of change

### 📚 Description

This should save a bit by avoiding creating plugins that didn't end up using, like webpack plugins when running Vite and vice versa. 

### 📝 Checklist


pi0:
I love it in general. Only a question: Can't unplugin be lazy by default?

antfu:
When we pass the option, it returns a plugin object - even if the hooks are not in use, it still creates an extra object and the compatible layer. What else can be lazy? (`unplugin.webpack` is lazy, `unplugin.webpack({})` is not)

pi0:
> When we pass the option, it returns a plugin object - even if the hooks are not in use, it still creates an extra object and the compatible layer. What else can be lazy?

I think we can make hooks themselves lazy initialized (on first rollup call via a proxy or wrapper). This way it is automagically lazy initialized.

antfu:
I don't honestly think it needs to be more lazy, proxy on hooks that are called a lot of times might introduce extra overhead. And this feature is not only for unplugins but also for other modules to install different plugins.

danielroe (approved):

danielroe:
I think this PR will pair well with https://github.com/nuxt/nuxt/pull/20587 - we can remove some of the extra steps it takes to lazily initialise the plugin, and use the getter that this PR provides.

PR #20559, opened 2023-04-28
feat(nuxt): support separate server tsconfig

### LINKED ISSUE #13920: DX for what can be imported in nitro vs vue context
In a Nuxt app, there are two different contexts:

1. The Nitro context, in which event handlers run.
2. The Vue context (client and server) in which Nuxt plugins, route middleware, components, pages, etc. run

What is available in these two different contexts is dramatically different, although in some cases utilities can be imported in both places, or shared between them:

1. **Same utility**. In some cases the same utility can be used in each context. We are currently discussing whether to support h3's `createError` in the Vue portion of the app in https://github.com/nuxt/framework/pull/4539, for example.
1. **Different utilities with same name and signature**. `useRuntimeConfig()` has a same name and DX but is a different utility in Nitro and Vue.
1. **Utilities that are only available in one context**. For example, `defineEventHandler` or `defineComponent`.
1. **Different utilities with same name but different signature**. For example, `useCookie`.

For 1 and 2, we don't have an issue. But for 3 and 4, we have the problem that users may use the 'wrong' import, either because it isn't provided via auto-imports in the context in which they are using it, or because it won't _work_ in that context. Even worse, they may import directly from `#app` in a server context, or try to import from nitro in a Vue context.

There are a number of potential ways to resolve this and this issue exists for discussion of these and other potential solutions. They are not mutually exclusive.

1. Provide clearer hints in the naming of composables (e.g. `defineServerCookie`).
2. Ensure there is either a common signatures or different names for functions.
3. Provide runtime checks (which we currently do through [import protection](https://github.com/nuxt/framework/blob/4712e994112a97feba447b3edb4afb65e968ed63/packages/nuxt/src/core/plugins/import-protection.ts#L15-L24)).
4. Provide separate TypeScript environments so IDEs can recognise differences.

Other ideas are welcome!

Related: https://github.com/nuxt/nuxt.js/issues/13713

cc: @pi0, @Atinux, @antfu

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/13920

### ❓ Type of change

### 📚 Description

Based on the following upstream work in Nitro, https://github.com/unjs/nitro/pull/1021, https://github.com/unjs/nitro/pull/1041 and https://github.com/unjs/nitro/pull/1146, This adds the option of adding an additional `~/server/tsconfig.json` with the following content:

```json
{
  "extends": "../.nuxt/tsconfig.server.json"
}
```

This will prevent Vue-only auto-imports from being visible within Nitro server routes (so, for example, `useNuxtApp` or `useCookie` will be marked as undefined).

There is a current limitation in that _Nitro_ auto-imports are still globally visible within the Vue part of the app, which currently seems to be necessary in order for `$fetch` type inference to work. (Ideas welcome on this!)

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -32,6 +32,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
       ...nuxt.options.typescript.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
     ],
     exclude: [
+      relative(nuxt.options.buildDir, nuxt.options.serverDir),
       // nitro generate output: https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/nitro.ts#L186
       relative(nuxt.options.buildDir, resolve(nuxt.options.rootDir, 'dist'))
     ]
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -32,7 +32,6 @@ export const writeTypes = async (nuxt: Nuxt) => {
       ...nuxt.options.typescript.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
     ],
     exclude: [
-      relative(nuxt.options.buildDir, nuxt.options.serverDir),
       // nitro generate output: https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/nitro.ts#L186
       relative(nuxt.options.buildDir, resolve(nuxt.options.rootDir, 'dist'))
     ]
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -92,7 +92,9 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
       layer => resolve(layer.config.srcDir, 'app.config')
     ),
     typescript: {
-      generateTsConfig: false
+      strict: true,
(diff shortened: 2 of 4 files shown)


Atinux:
Would it be easier if we have a separation with `app/` and `server/` directory or it won't change anything?

packages/nuxi/src/utils/prepare.ts
```diff
@@ -32,6 +32,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
       ...nuxt.options.typescript.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
     ],
     exclude: [
+      relative(nuxt.options.buildDir, nuxt.options.serverDir),
```
pi0, line 35: To avoid making it a breaking change for type support shall we try to detect `server/tsconfig.json` and optionally ignore (also make a warning to enable file?)

pi0:
> Would it be easier if we have a separation with app/ and server/ directory or it won't change anything?

Finally with new structure exactly it would probably be much better to also have app specific config IMO that isolates nuxt auto imports within the dir (local modules for example should not see them in auto complete list...)

danielroe:
Separating `app/` and `server/` with the current implementation wouldn't make any difference as the issue is that we _need_ the nitro auto-imports within scope of the Vue app in order for type inference to work. My current best hope is that I can set up a TS partition setup (which _would_ be aided by app/server separation).

danielroe:
The current limitation is that the _IDE_ will correctly show errors but running `vue-tsc` doesn't test the separate server directory - its tests are being run solely in the root directory.

I still think this is worth merging, and we can iterate for a better way for users to typecheck their server code separately.

atinux (approved):

pi0 (approved):
Nice. LGTM. I have some ideas about internal renamings (specifically to isolate nitro's build dir) but i think we can make it done with backward compatibility and iterate.

PR #20587, opened 2023-04-29
feat(kit): `addBuildPlugin` for  builder-agnostic implementation

danielroe:
### ❓ Type of change

### 📚 Description

this adds a provider agnostic utility that can pass an object with keys (vite/webpack/rspack), each of which is a factory function for a plugin. This could be used to directly pass an `unplugin`, for example, and means that we can also 'upgrade' a module or integration if we + unplugin later support more builders (`rspack` being a case-in-point). All we would do is call the (new) factory function, if it exists.

### 📝 Checklist


packages/kit/src/build.ts
```diff
@@ -136,3 +136,18 @@ export function addVitePlugin (pluginOrGetter: VitePlugin | VitePlugin[] | (() =
...
+
+export function addBuildPlugin (pluginFactory: AddBuildPluginFactory, options?: ExtendConfigOptions) {
+  if (pluginFactory.vite) {
+    addVitePlugin(pluginFactory.vite, options)
+  }
+
+  if (pluginFactory.webpack) {
+    addWebpackPlugin(pluginFactory.webpack, options)
+  }
+}
```
pi0, line 153: Just an idea: This function would also check if _current builder_ is supported or not and make a warning a build plugin is added without vite/webpack support

pi0 (approved):

packages/kit/src/build.ts
danielroe: I love the idea. Also maybe provide verbose warning and suggestion in module building mode?

pi0: For module building (i guess testing is only possible) we could do if partial in either case too :)

PR #20599, opened 2023-04-30
fix(nuxt): always inline styles for server/island components

### LINKED ISSUE #20582: No styles for server component if `experimental.inlineSSRStyles` is turned off
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.3.3
- Builder:          vite
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-5hj7mv?file=nuxt.config.ts

1. `npm run build; npm run preview`. Styles from `test` component are present in rendered html
2. Go to `nuxt.config.ts` and set `experimental.inlineSSRStyles` to false.
3. `npm run build; npm run preview`. Now button hasn't got its fancy styles

### Describe the bug

Sometimes inline styles are bad, because they can be huge and as a result slow down the download of the main content. So, I decided to turn it off, but my page was somewhat broken because a lot of styles were missing (•_•)

### Additional context

if you visit `[repro host]/__nuxt_island/Test` with `inlineSSRStyles` set to `true`, there is `head.style` item. But if `inlineSSRStyles` set to `false`, there are nothing except for html content itself.

Guessing from [this line](https://github.com/nuxt/nuxt/blob/1464294d6abe13b54fdb8d204732345551c8e647/packages/nuxt/src/core/runtime/nitro/renderer.ts#L337) scoped style tag should help, but in reality styles are still not there. So link to styles are not even in [rendered html](https://github.com/nuxt/nuxt/blob/1464294d6abe13b54fdb8d204732345551c8e647/packages/nuxt/src/core/runtime/nitro/renderer.ts#L335) ?

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/20582

### ❓ Type of change

### 📚 Description
Hi :wave: currently disabling `experimental.inlineSSRStyles` leads to missing styles for server/islands components.
This PR keep it enabled for server/islands components even if it is disabled (or if there's a function)

### 📝 Checklist


packages/vite/src/plugins/ssr-styles.ts
```diff
@@ -93,10 +93,13 @@ export function ssrStylesPlugin (options: SSRStylePluginOptions): Plugin {
     async transform (code, id) {
       const { pathname, search } = parseURL(decodeURIComponent(pathToFileURL(id).href))
       const query = parseQuery(search)
-      if (!pathname.match(/\.(vue|((c|m)?j|t)sx?)$/g) || query.macro) { return }
-      if (options.shouldInline && !options.shouldInline(id)) { return }
 
       const relativeId = relativeToSrcDir(id)
+      if ((!pathname.endsWith('.server.vue') && !relativeId.startsWith('components/islands/'))) {
```
danielroe, line 98: can we have a better way of detecting whether these are island components, do you think? is this something the components module can handle for us?

huang-julien: indeed, I was wondering if i should move it and use `getComponents` from `src/app/components/module`

danielroe: Or have some kind of way to detect it - maybe through listening to or calling a components hook?

huang-julien: I think both are possible, within components module with `getComponents` and within vite with a `components:extend` 🤔 

do you prefer to keep it within `vite` or to move it to components  🙂 ?

huang-julien: it looks nice ! i'll be waiting for it :+1:

danielroe: Done.

huang-julien: https://github.com/nuxt/nuxt/blob/8b86d39e30ca970fb0adf2b5665a77f8dbbfb439/packages/nuxt/src/core/builder.ts#LL15C5-L15C5

Due to `app:templates` being called in `generateApp`, hooking into it won't work 🤔. to keep it within `nuxt/vite`, what do you think about passing down the `app` to `bundle()` ?
So we would have `bundle(nuxt, app)`  [here](https://github.com/nuxt/nuxt/blob/8b86d39e30ca970fb0adf2b5665a77f8dbbfb439/packages/nuxt/src/core/builder.ts#L39)

danielroe: Hm. Maybe instead we could set the resolved app on `nuxt.app` when we are building?

danielroe: ok - merged ✅

danielroe (approved):

PR #20600, opened 2023-05-01
Docs: Update docs/2.guide/2.directory-structure/1.plugins.md

sharkfin009:
### ❓ Type of change

### 📚 Description

Needs updating, as using Vue plugins is now way simpler than as described currently.

### DIFF
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -164,9 +164,11 @@ If you are using WebStorm, you may need to augment `@vue/runtime-core` until [th
 
 ## Vue Plugins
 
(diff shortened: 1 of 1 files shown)


sharkfin009:
There may well be something I've missed here . Just installing primevue and it seems creating a plugin file inside 'plugins'  is indeed needed in this case. Some plugins seem to not need that though. Need an expert here ;) But I do think the docs do need updating anyhow

danielroe:
A Vue plugin needs to be installed on the Vue app (https://vuejs.org/guide/reusability/plugins.html#plugins). So we need to communicate how to do that. What do you think is wrong about the docs here? (Is it specific to the vue-gtag example?)

danielroe (changes requested):
I believe the current text is correct. See example documentation from `vue-gtag`

CleanShot 2023-05-08 at 21 55 10@2x

I may have misunderstand what you're proposing though - please do let me know or open a new PR to improve things. Maybe a quick link out to the documentation on Vue plugins: https://vuejs.org/guide/reusability/plugins.html#plugins...

PR #20605, opened 2023-05-01
fix(webpack): warn when using deprecated `build.extend` hook

danielroe:
### ❓ Type of change

### 📚 Description

We haven't supported the `build.extend` option for some time. This removes the partial support for it within the webpack builder, in favour of the `webpack:config` hook and `extendWebpackConfig` helper function.

### 📝 Checklist


packages/webpack/src/utils/config.ts
```diff
@@ -57,31 +57,6 @@ export function fileName (ctx: WebpackConfigContext, key: string) {
...
 
 export function getWebpackConfig (ctx: WebpackConfigContext): Configuration {
-  const { options, config } = ctx
-
-  // TODO
-  const builder = {}
-  const loaders: any[] = []
-
-  // @ts-expect-error TODO: remove support for `build.extend` in v3.5
-  const { extend } = options.build
```
pi0, line 67: Shall we consider making a warning placeholder at least? This will be a silent deprecation with unnoticed regressions if a legacy module exists in projects...

danielroe: good idea. do you think it's worth going further and keeping the behaviour for another minor?

pi0: Up to you. I am pretty sure there are used modules depending on this and big projects depending on them. Warn + wait for another minor would be good idea too :)

pi0 (approved):

PR #20608, opened 2023-05-01
feat(nuxt): add `nuxtApp.runWithContext`

### LINKED ISSUE #14805: `nuxtApp.call`
### Describe the feature

Context: https://github.com/nuxt/nuxt.js/issues/14269 and https://github.com/nuxt/nuxt.js/issues/14723

While we try to restore context on best efforts, It is not always possible or convenient. There is an (intended to be internal) util `callWithNuxt` from `#app` can be used to use to restore context for async composables:

```ts
import { callWithNuxt } from '#app'

// unsupported transform context
const nuxtApp = useNuxtApp()
await something()
await callWithNuxt(navigateTo, ['/route', { replace: true }])
```

It is not convenient for DX because for performance reasons, we chose to accept second arg as arg array and that also it is not auto-imported. This utility is fundamentally bound to `nuxtApp` instance and can be exposed to public API:

```ts
const nuxtApp = useNuxtApp()
await something()
await nuxtApp.call(navigateTo, '/route', { replace: true })
```

Alternatives:
- We also have `defineAsync` to resolve issue but still there is no guarantee transform method works everyever.
- We have `callWithNuxt` which could be exposed as an auto import but it is not user friendly nor makes sense to use another auto import when nuxtApp is main context

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14805

### ❓ Type of change

### 📚 Description

This adds a new utility, aligned with https://github.com/vuejs/core/pull/7451, that allows easier calling of functions with the Nuxt context.

### 📝 Checklist


antfu (approved):
Make sense to me!

TheAlexLichter (approved):

danielroe:
Nice idea!

ColinRosati:
I would love to see this in nuxt bridge!!

Happy to get some support. Im not sure how the nuxt bridge is built from this lib

ColinRosati:
Great. I opened an issue [757](https://github.com/nuxt/bridge/discussions/757) to start before pushing a PR

PR #19966, opened 2023-03-28
feat(nuxt): definePageMeta in "pages:extend"

huang-julien:
### 🔗 Linked issue

 https://github.com/nuxt/nuxt/issues/19955

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Hi :wave: , this PR add the result of the `page:extend` hook so pages added within modules can be also transformed.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #20628, opened 2023-05-02
fix(nuxt): run page meta plugin on all pages (and only pages)

### LINKED ISSUE #19955: `definePageMeta` support for files added in `pages:extend`
### Environment

clean setup

Nuxi 3.3.2

RootDir: /home/projects/github-cvxlgw
Nuxt project info: 

- Operating System: `Linux`
- Nuxt Version:     `3.3.2`
- Nitro Version:    `2.3.2`
- Builder:          `vite`
- User Config:      `pages`, `modules`
- Runtime Modules:  `~/modules/pages/index`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-cvxlgw?file=modules/pages/index.ts

### Describe the bug

We have a module that:
1. Adds middleware through plugin
2. Adds page component by `extendPages`
3. Added page use added middleware by `definePageMeta`

Result: warning
`definePageMeta() is a compiler-hint helper that is only usable inside the script block of a single file component which is also a page. Its arguments should be compiled away and passing it at runtime has no effect.`

This only happens when the middleware is added by the module.

### Additional context

_No response_

### Logs

```shell-script
Nuxi 3.3.2                                                                                                        13:32:01
Nuxt 3.3.2 with Nitro 2.3.2                                                                                       13:32:02
                                                                                                                  13:32:02
  > Local:    http://localhost:3000/ 
  > Network:  http://192.168.1.104:3000/

ℹ Vite client warmed up in 611ms                                                                                  13:32:04
✔ Nitro built in 200 ms                                                                                     nitro 13:32:04
definePageMeta() is a compiler-hint helper that is only usable inside the script block of a single file component which is also a page. Its arguments should be compiled away and passing it at runtime has no effect.
definePageMeta() is a compiler-hint helper that is only usable inside the script block of a s
```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19955
alternate implementation of https://github.com/nuxt/nuxt/pull/19966

### ❓ Type of change

### 📚 Description

Huge credit to @huang-julien - while reviewing his PR (which was great) it occurred to me that we could take the opportunity to rework how the page meta plugin works, which should also have the effect of improving performance. So this PR:

1. _only_ runs page-meta plugin (which parses AST, and therefore is costly) on imports with `?macro=true` query, which means it also will run for all pages, however, they are added, including via `pages:extend` hook
2. for perf, in dev mode we can keep `definePageMeta` in the component and only warn based on _usage_ (namely, whether the component rendered is a route component). This will need testing to make sure we won't give users false positives here. I'd be up for fully removing the warning if needed.
3. in build we can fully tree-shake calls to `definePageMeta`

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -151,10 +151,7 @@ export default defineNuxtModule({
     // Extract macros from pages
     const pageMetaOptions: PageMetaPluginOptions = {
       dev: nuxt.options.dev,
-      sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client,
-      dirs: nuxt.options._layers.map(
-        layer => resolve(layer.config.srcDir, layer.config.dir?.pages || 'pages')
-      )
+      sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client
     }
     nuxt.hook('modules:done', () => {
       addVitePlugin(() => PageMetaPlugin.vite(pageMetaOptions))
--- a/packages/nuxt/src/pages/page-meta.ts
+++ b/packages/nuxt/src/pages/page-meta.ts
@@ -10,7 +10,6 @@ import MagicString from 'magic-string'
 import { isAbsolute, normalize } from 'pathe'
 
 export interface PageMetaPluginOptions {
-  dirs: Array<string | RegExp>
   dev?: boolean
   sourcemap?: boolean
 }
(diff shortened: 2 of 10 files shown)


huang-julien:
Looks wonderful ! ❤️ can't wait for it 🙌

antfu (approved):

ineshbose:
Awesome! I was thinking of suggesting a similar macro for components like `defineComponentMeta` that could take [component schema](https://github.com/nuxt/nuxt/blob/566fa85fc854bec7deb7013691946bbd86653b07/packages/schema/src/types/components.ts#LL1C22-L1C22) arguments?

danielroe:
Nice idea 👌

PR #20629, opened 2023-05-02
test: test with bundler module resolution

danielroe:
### ❓ Type of change

### 📚 Description

This tests `bundler` TS module resolution mode, which is what is recommended for frameworks like Nuxt (better than `Node16` as it's less restrictive but matches how bundlers should be working).

There are a number of unjs/nuxt libraries that need fixes:

**PR needs to land and be released**

**Package needs to be released**
- [x] pkg-types
- [x] perfect-debounce: https://github.com/unjs/perfect-debounce/pull/14
- [x] @nuxt/devalue: https://github.com/nuxt-contrib/devalue/pull/21
- [x] std-env: https://github.com/unjs/std-env/pull/51
- [x] cookie-es: https://github.com/unjs/cookie-es/pull/4
- [x] ufo

cc: @pi0 

### 📝 Checklist


pi0:
cookie-es and perfect-debounce updated in major version bump

PR #20637, opened 2023-05-02
feat(nuxt): add apps to nuxt build-time instance

huang-julien:
### 🔗 Linked issue
helps #20599 

### ❓ Type of change

### 📚 Description
Hi :wave:
This allow us to have access to `NuxtApp` within `Nuxt` to use some of its fields such as `components` later (in builders for example).

### 📝 Checklist


packages/nuxt/src/core/nuxt.ts
```diff
@@ -37,7 +37,17 @@ export function createNuxt (options: NuxtOptions): Nuxt {
     hook: hooks.hook,
     ready: () => initNuxt(nuxt),
     close: () => Promise.resolve(hooks.callHook('close', nuxt)),
-    vfs: {}
+    vfs: {},
+    app: {
```
pi0, line 41: For future compatibility with multi-layer/apps, i would suggest to expose it as `nuxt.apps.default`

huang-julien: :+1:

danielroe (approved):
❤️

PR #20639, opened 2023-05-02
fix(nuxt): allow resolving client nuxt app to singleton

danielroe:
### 🔗 Linked issue

https://github.com/danielroe/nuxt-vitest/issues/133

### ❓ Type of change

### 📚 Description

It can be handy to resolve a client-side nuxt app to a singleton. (It might also be worth, in dev mode, considering making this a global promise to avoid context conflicts when the entry is imported more than once from different contexts.)

### 📝 Checklist


packages/nuxt/src/app/entry.ts
```diff
@@ -53,7 +53,11 @@ if (process.client) {
     import.meta.webpackHot.accept()
   }
 
+  // eslint-disable-next-line
+  let app: Promise<any>
```
pi0, line 57: Shall we explicitly call it `nuxtApp`/`vueApp` to avoid future conflicts and naming consistency?

PR #20669, opened 2023-05-04
fix(utils): respect patterns within paths when sorting routes

### LINKED ISSUE #10630: sortRoutes doesn't handle * path correctly
### Versions

- nuxt: v2.15.8
- node: 14

### Reproduction

Non-simplified reproduction is in the issue reported in `@nuxtjs/i18n` - https://github.com/nuxt-community/i18n-module/issues/1480

It boils down to `sortRoutes` utility function not sorting the `*` correctly. It's not placed at the end thus it catches requests that are meant for later routes in the order.

It works correctly with `/*` path.

marcelobotega:
### 🔗 Linked issue

Fixes #10630 

### ❓ Type of change

### 📚 Description

Update sortRoutes to handle routes with letters and * (test*)
Resolves #10630 

### 📝 Checklist


package.json
```diff
@@ -38,6 +38,7 @@
     "@ls-lint/ls-lint": "^1.11.2",
     "@nuxtjs/eslint-config": "^12.0.0",
     "@nuxtjs/eslint-config-typescript": "^12.0.0",
+    "@nuxtjs/i18n": "^7.3.1",
```
danielroe, line 41: do you think we could adequately test this change without adding the module as a dependency?

marcelobotega: Without the module we can't generate the fixtures to test the "real" problem 😞 
However, I could try something, but not sure if it is the best solution. 
We can utilize both the routes.json files from the fixtures, unsort them, and apply them to new unit tests to evaluate the sortRoute method directly.
Would this be better @danielroe ?

danielroe: Yes, I think that would be fine 👍

marcelobotega:
@danielroe can we run the actions on this PR ?

marcelobotega:
@danielroe we got an error on the windows audit, how can we fix that ? It's a timeout error.

packages/utils/test/route.test.js
```diff
@@ -237,4 +237,109 @@ describe('util: route', () => {
...
+          { path: '/reports' },
+          { path: '/tech' },
+          { path: '/de/about' },
+          { path: '/de/poetry' },
+          { path: '/de/reports' },
+          { path: '/de/tech' },
+          { path: '/' },
+          { path: '/de/:post' },
+          { path: '/:post' },
+          { path: '/de*' },
```
rchl, line 299: Looks like `/de*` would be inaccessible this way? Should be before `/:post`?

marcelobotega: Nice catch, yeah, the /de* won't work on this scenario. Not sure how relevant is this to the use case, but I'll try to fix it and have the same output as the `Should sortRoutes with extendRoutes using /*` test case.

rchl (dismissed):
I've tested this on a couple of projects that (admittedly) only have a single `*` route but I haven't noticed any issues.

One thing that happens consistently is that the root route (`/`) changes position to after all other dynamic routes (ones that use `:`). So in theory that could result in slightly slower resolving of that particular route but not sure if it's something to be concerned about.

So thumbs up from me.

rchl (dismissed):

packages/utils/test/route.test.js
```diff
@@ -238,4 +238,109 @@ describe('util: route', () => {
       expect(routesResult).toMatchSnapshot()
     })
   })
+
+  describe('util: route sortRoutes', () => {
+    const files = [
+      'pages/_param.vue',
+      'pages/de/index.vue',
+      'pages/index.vue',
+      'pages/*.vue',
```
danielroe, line 247: shouldn't this (and other occurrences) be `_.vue` instead?

marcelobotega: If I'm not mistaken, nuxt uses the `_.vue` as a catch-all, right ?
I changed here and when using `_` it goes back to the initial problem.
I found out the [getRoutePathExtension](https://github.com/nuxt/nuxt/blob/696dbc68447c95527c0d5dc2c324d5f93e4dc0cf/packages/utils/src/route.js#LL243C7-L243C28) method, and I am adding a snippet there to check the `_` in the end of the key. This should fix the problem for this use case.

rchl: Actually I think you should not be doing anything with `_` that is neither the only character nor followed by some text.

You can test it yourself by creating `pages/test_.vue` file. Nuxt doesn't convert the `_` to `*`. So it's supposed to be handled as literal `_`.

marcelobotega: Hey @rchl, 
I tested and if I add the route `de_.vue` we go back to the initial problem. The `de_` route goes to the top of the list and catches all `de` routes. 
I changed the [getRoutePathExtension](https://github.com/nuxt/nuxt/blob/d24275a1dc85b5be29a49cc23b66a1eab2f94552/packages/utils/src/route.js#L243) to handle the `_` in the end of the string (now it converts to *) and with that change everything seems to be right.

rchl: Just test manually by creating a `de_.vue` file. Nuxt generates this route:
```
{
    path: "/de_",
    component: _0b861482,
    name: "de_"
  }
```

where `_` is treated literally and not converted to `*`. So it's fine if it's at the beginning and it shouldn't cause any issues because it's a literal match.

marcelobotega: Ok, I understood now. 
My only question now is: Should the `_` in the end of the value be treated as a literal or not ?
For me, it makes sense to be treated as a special character, but I'm not sure if Nuxt should handle it like that, either way seems fine to me.
@danielroe can you give your insights on it ?

rchl: [File-based routing in Nuxt](https://v2.nuxt.com/docs/features/file-system-routing) only supports the case when key is equal `_` or is followed by some characters. But modules can add routes with `*` anywhere in fact and the user can also do it using the [extendRoutes](https://v2.nuxt.com/docs/features/file-system-routing#extendroutes) option. So sorting of those should IMO still be supported but the changes should not affect the capabilities of the file-based routing (and I think your change in `getRoutePathExtension` would affect those).

marcelobotega: Yes, it affects those. I'll roll back the commit with changes on the `getRoutePathExtension`

rchl (approved):

danielroe (approved):

PR #20679, opened 2023-05-04
fix(nuxt, vite): ignore `nuxt_component` ssr style and `isVue`

### LINKED ISSUE #93: Broken with latest version of Nuxt
image

Reported by @Atinux ;

Can be worked around using `experimental.inlineSSRStyles: false` in the `nuxt.config` file while it gets fixed.

antfu:
### 🔗 Linked issue

fix https://github.com/Tahul/pinceau/issues/93

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/utils/plugins.ts
+++ b/packages/nuxt/src/core/utils/plugins.ts
@@ -14,8 +14,13 @@ export function isVue (id: string, opts: { type?: Array<'template' | 'script' | 
 
   const query = parseQuery(search)
 
-  // Macro and component wrapper
-  if ((query.macro || query.nuxt_component) && (!opts.type || opts.type.includes('script'))) {
+  // Component async/lazy wrapper
+  if (query.nuxt_component) {
+    return false
+  }
+
+  // Macro
+  if (query.macro && (!opts.type || opts.type.includes('script'))) {
     return true
   }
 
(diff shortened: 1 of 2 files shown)


danielroe (approved):

danielroe:
I think we might also want to add something like this:

```ts


  // wrapper import which is generated by importing from #components
  if (query.nuxt_component && opts.type && !opts.type.includes('script')) {
    return false
  }
```

here:

https://github.com/nuxt/nuxt/blob/c7be5b4ec6a3b129c68e892c9f3dad816446bb10/packages/nuxt/src/core/utils/plugins.ts#L17-L20

What do you think?

danielroe:
I think we might also want to add something like this:

```ts


  // wrapper import which is generated by importing from #components
  if (query.nuxt_component && opts.type && !opts.type.includes('script')) {
    return false
  }
```

here:

https://github.com/nuxt/nuxt/blob/c7be5b4ec6a3b129c68e892c9f3dad816446bb10/packages/nuxt/src/core/utils/plugins.ts#L17-L20

What do you think?

danielroe (commented):
I think we might also want to add something like this:

```ts


  // wrapper import which is generated by importing from #components
  if (query.nuxt_component && opts.type && !opts.type.includes('script')) {
    return false
  }
```

here:

https://github.com/nuxt/nuxt/blob/c7be5b4ec6a3b129c68e892c9f3dad816446bb10/packages/nuxt/src/core/utils/plugins.ts#L17-L20

What do you think?

antfu:
Do you think we should rename `?macro` to `?nuxt_macro` (or `?nuxt_page_meta`) to make them next-specific?

danielroe:
Yes, I would agree - `nuxt_page_meta` makes sense.

PR #20680, opened 2023-05-04
docs: add warning about redirect check

GalacticHypernova:
### 🔗 Linked issue

Fixes #20678

### ❓ Type of change

### 📚 Description

This PR fixes multiple issues with the newly merged docs, such as to.path usage in route guards (in stead of from.path, because we don't want to force people onto a specific route, the goal is to avoid infinite redirects). Also, this fixes an undefined variable 'user' in the "err as a String" section of the api/utils/abort-navigation page.
### 📝 Checklist


docs/2.guide/2.directory-structure/1.middleware.md
```diff
@@ -30,7 +30,7 @@ export default defineNuxtRouteMiddleware((to, from) => {
   if (to.params.id === '1') {
     return abortNavigation()
   }
-  if (to.path !== '/') {
+  if (from.path !== '/') {
```
danielroe, line 33: would you explain what the issue is here, from your point of view?

GalacticHypernova: The previous version with to.path forced the user onto the "/" route no matter where the navigation actually went, the new version makes sure the user isn't already on the "/" route before navigating

danielroe: Don't they both have the same effect?

GalacticHypernova: Maybe they do, but the to.path check will make it impossible for users to navigate, as it will force redirect to the "/" route

danielroe: indeed. you would never write this as-is but it's just providing an example of a redirection

GalacticHypernova: Even though it's only an example of a redirection, won't it be better to have that example be functional and implementable in real-life applications? It seems a little irrational to provide an unrealistic example. This maintains the 'exampleness' of it while also making it realistic and ready for use in actual websites.

GalacticHypernova: Not to mention the docs also had it wrong in the abort-navigation, which I also covered in this PR

GalacticHypernova:
Oops, didn't know what re-requesting review button would do

docs/2.guide/2.directory-structure/1.middleware.md
danielroe: I'm very up for making this more realistic, but `from` is no more realistic here than `to`, and in most cases will be inaccurate because it isn't guarding the route you are going *to*. We could also just add a comment to explain more.

GalacticHypernova: How does it not guard it? The middleware runs with a redirect, so if we see the user is on the same route he's trying to go to, it won't run as it won't redirect. Alternatively, if from is still problematic, we can change it to "if (to.path !== from.path){
    navigateTo(to.path)
}" (if needed can be changed to fullPath)

danielroe: I still don't think that will be right. 

* Current behaviour - correctly redirects any non-`/` routes to `/`: https://stackblitz.com/edit/github-3spxnz
* Proposed behaviour - results in an infinite redirect: https://stackblitz.com/edit/github-3spxnz-qztmiu

danielroe (changes requested):
Let's keep the current `to.path` rather than `from.path` but we can add a comment along the lines of:

```ts
// in a real app you would probably not redirect every route to `/`
// but it is important to check `to.path` before redirecting or you
// may cause an infinite redirect loop
```

GalacticHypernova:
Sure thing!

docs/3.api/3.utils/abort-navigation.md
```diff
@@ -36,7 +36,7 @@ export default defineNuxtRouteMiddleware((to, from) => {
     return abortNavigation()
   }
  
-  if (to.path !== '/edit-post') {
+  if (from.path !== '/edit-post') {
```
danielroe, line 39:
```suggestion
  if (to.path !== '/edit-post') {
```

docs/3.api/3.utils/define-nuxt-route-middleware.md
```diff
@@ -56,7 +56,7 @@ export default defineNuxtRouteMiddleware((to, from) => {
     return navigateTo('/login')
   }
 
-  if (to.path !== '/dashboard') {
+  if (from.path !== '/dashboard') {
```
danielroe, line 59:
```suggestion
  if (to.path !== '/dashboard') {
```

docs/3.api/3.utils/define-page-meta.md
```diff
@@ -144,7 +144,7 @@ The example below shows how the middleware can be defined using a `function` dir
             return navigateTo('/login')
         }
 
-        if (to.path !== '/checkout') {
+        if (from.path !== '/checkout') {
```
danielroe, line 147:
```suggestion
        if (to.path !== '/checkout') {
```

docs/3.api/3.utils/navigate-to.md
```diff
@@ -96,7 +96,7 @@ await navigateTo({
 
 ```ts
-  if (to.path !== '/search') {
+  if (from.path !== '/search') {
```
danielroe, line 99:
```suggestion
  if (to.path !== '/search') {
```

danielroe (approved):
Thanks!

Issue #20691, opened 2023-05-05
[Nuxt 2] Executes asyncData after page component navigated away from


rchl:
### Environment

```
  System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1
    Memory: 216.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.19.0 - ~/.nvm/versions/node/v16.19.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.19.3 - ~/.nvm/versions/node/v16.19.0/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Firefox: 112.0.2
    Safari: 16.4
    Safari Technology Preview: 16.4
```

### Reproduction

https://stackblitz.com/edit/github-6mgkvf?file=pages/index.vue

### Describe the bug

In case there is a global middleware (or a page has a middleware) that runs for a longer time, it's possible for the user to navigate to another page (either using browser back button or clicking another link) while the middleware is running. In that case the `asyncData` of the target page will still execute after the long-running middleware is finished but at that point the current route is no longer what that page expects so `context.route` will be unexpectedly pointing to another route.

That can lead to crashes due to unexpected input or if, for example, the page triggers and `error` function then it will actually apply to the page that is currently loaded rather than the one whose `asyncData` is running.

rchl:
Note that I've identified a couple of cases where things could go wrong:
 - a long-running middleware, then asyncData that calls either `error(...)` or throws an exception
 - a long-running asyncData that calls either `error(...)` or throws an exception

If the user navigates away from the page after the long-running operation starts then things will go wrong (the error page will replace the current page).

It fairly easy to handle the first case with long-running middlewares since we can fairly easily interrupt page loading after middlewares have finished. The case with long running asyncData is more tricky since it's not possible to interrupt asyncData and if we'd want to suppress the effects of calling `error(...)` then it could also be tricky to do (the exception case should be easy to handle).

But of course those would be a bit of band-aids for a problem that probably is not fully fixable (at least with current architecture). There is no way to just interrupt execution of a middleware or asyncData at given moment and those handlers could do other things that would affect the state after navigating away.

That said, handling all or most of the cases that I've listed at the beginning of this comment would be beneficial in many if not most cases, so I could prepare some PR to that effect. Basically the change would be to explicitly interrupt `render` (in `client.js`) after certain async operations.

[...]

PR #20719, opened 2023-05-08
fix(vue-app): skip page render early on error or navigation

rchl:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/20691

### ❓ Type of change

### 📚 Description

This change attempts to interrupt page "rendering" early if the route is navigated away from while middleware or asyncData is in progress. Additionally, also interrupts as early as possible when `error()` is called so that the error page is not shown after user has navigated to another page already.

Fixed by creating a plain `renderState` object that has an `aborted` property set to `false` initially, and passing that object to the internal `render` function and then also down to function that executes middlewares. If a new `render` is triggered (through the `router.beforeEach` hook) while the original one is in progress, the `aborted` property is set to `true` so that the original `render` pass can abort as soon as possible (once currently executing async function is done).

I realize that this is not 100% proof solution as the user could do many weird things that would affect the state of the app from the middlewares or the `asyncData` handlers, but I think this should cover most common cases.

Added tests that verify that the error page is not shown after navigating away from the page when the page triggers `error()` call or throws an exception. Unfortunately those have to rely on arbitrary 2s delay but I haven't found a better way to test that.

### 📝 Checklist


packages/vue-app/template/client.js
```diff
@@ -995,5 +995,6 @@ async function mountApp (__app) {
         errorHandler(err)
       }
     })
-  })
+  },
+  { aborted: false })
```
rchl, line 999: I have not put too much thought into that part of the code (the SPA case). I could look more if needed but it could be just left like that, without supporting early abort.

rchl:
@manniL I see you've been assigned. Could you have a look at this?

TheAlexLichter (approved):

manniL:
Looks like a good solution to me for covering basic use cases 👍

PR #20726, opened 2023-05-08
feat(nuxt): environment-specific plugin execution for islands

### LINKED ISSUE #20708: Provide an `islands` boolean to plugins
### Describe the feature

Would be nice to be able to disable some plugins for server and islands components (default to `true` to avoid any breaking changes).

### Additional information

### Final checks

huang-julien:
### 🔗 Linked issue
resolve #20708 

### ❓ Type of change

### 📚 Description
hi :wave: 
This PR adds an `islands` boolean (default `true`) to the `defineNuxtPlugin` Object input. This way we can disable some specific plugins for server-only and islands components

### 📝 Checklist


danielroe (commented):
I love this. Only reason I've delayed merging is a question of API. For example, I thinking instead about supporting `context` - then it could have both `island` and `error` properties, perhaps?

What do you think @antfu @pi0 @harlan-zw @lihbr?

(We could even also consider plugins being opted-out by default in a future version of both islands + error contexts - we could go ahead and add an opt-in to this behaviour via an experimental option.)

harlan-zw:
Thanks for the ping, my thoughts;

**Re default value:**

It's a bit of a tricky one as I could see both sides of it. With `nuxt-og-image` I'd personally like _no_ plugins to run as it makes sure there are no unnecessary scripts running (esp. trackers).

However from a user point of view, if they're trying to access something added through a plugin, it may be confusing and hard to debug why it doesn't work.

Ideally, I'd like to see it as undefined. Undefined would mean it's not used by default but we could add an escape hatch within the island component to use all plugins that haven't explicitly opted out (`islands: false`)

If we needed to go with binary, I would say `false`.

**Re context:**

Yes, I like the idea of this. Should probably include client/server as well?

pi0:
I love this. Also the idea of moving the flag into a namespace. Only thinking we might have other uses for `context` in plugins. What about `env.islands`?

danielroe:
I can't find it now, but I think @pi0's suggestion is a good one - maybe `env` key rather than `context` key?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -390,7 +403,7 @@ export function definePayloadPlugin<T extends Record<string, unknown>> (plugin:
 
 /*! @__NO_SIDE_EFFECTS__ */
 export function defineNuxtPlugin<T extends Record<string, unknown>> (plugin: Plugin<T> | ObjectPluginInput<T>, meta?: PluginMeta): Plugin<T> {
-  if (typeof plugin === 'function') { return defineNuxtPlugin({ setup: plugin }, meta) }
+  if (typeof plugin === 'function') { return defineNuxtPlugin({ setup: plugin, env: { islands: true } }, meta) }
```
danielroe, line 406:
```suggestion
  if (typeof plugin === 'function') { return defineNuxtPlugin({ setup: plugin, env: {} }, meta) }
```

danielroe (approved):

PR #20747, opened 2023-05-09
feat(nuxt): custom cache support for data fetching composables

### LINKED ISSUE #15445: cache support for useFetch/useAsyncData
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-rc.14-27802701.2f53495`
- Nitro Version:    `0.6.2-27796963.837f894`
- Package Manager:  `npm@8.8.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ccqtvp?file=pages%2Fabout.vue,pages%2Findex.vue,package.json

As you can see, on `rc.14 edge` if you click "Go to about page" then "Go to home", you will see that the data will be fetched again.
 If you change to `rc.13`,  the data won't be fetched again, loading from cache.

### Describe the bug

Everytime `useFetch` is being called, it ma dke a new request, it is not checking for the fetched cached data, as it used to be working by `rc.13`

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #20476: useFetch triggers a request on the client side in nested components (using resolveComponent)
### Environment

- Operating System: Linux
- Node Version:  v18.12.1
- Nuxt Version: 3.4.2
- Nitro Version: 2.3.3
- Builder:  vite
- User Config:  components
- Runtime Modules:  -
- Build Modules: -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-6ejkfx?file=pages/variable.vue

### Describe the bug

This issue happens when you use nested ```<component :is="component" />```

If you resolve the parent component with the string - useFetch triggers a request only on the server side and the data is transferred to the client, which is fine.
```ts
const component = resolveComponent('Main');
```

If you resolve the parent component with the variable, everything renders well, but useFech triggers a request both on the server side and the client side
```ts
const name = 'Main';
const component = resolveComponent(name);
 ```
 
Components are registered as global components as it was mentioned [here](https://github.com/nuxt/nuxt/issues/14036#issuecomment-1397340507).

### Additional context

There is an **index.vue** page in the [stackblitz](https://stackblitz.com/edit/nuxt-starter-38tkoc) example where everything works fine, and **variable.vue** which is "broken". 

### Logs

_No response_

### PR DESCRIPTION
**Pull request by darioferderber**
### 🔗 Linked issue
https://github.com/nuxt/nuxt/issues/20476
resolves https://github.com/nuxt/nuxt/issues/15445

### ❓ Type of change

### 📚 Description
Resolves https://github.com/nuxt/nuxt/issues/20476

This adds support for custom caching within `useAsyncData` and `useFetch`. It works by providing a `getCachedData` function via the options which receives a key and can return cached data if it exists. For example, to make all requests cached by default, the following options could be used:

```ts
const nuxtApp = useNuxtApp()
const { data } = await useAsyncData(() => { /* fetcher */ }, {
  // this will not refetch if the key exists in the payload
  getCachedData: key => nuxtApp.payload.static[key] || nuxtApp.payload.data[key]
})
```

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -109,10 +110,12 @@ export function useAsyncData<
   options.lazy = options.lazy ?? false
   options.immediate = options.immediate ?? true
 
+  options.forcePayloadExtraction = options.forcePayloadExtraction ?? false
+
   // Setup nuxt instance payload
   const nuxt = useNuxtApp()
 
-  const getCachedData = () => nuxt.isHydrating ? nuxt.payload.data[key] : nuxt.static.data[key]
+  const getCachedData = () => (nuxt.isHydrating || options.forcePayloadExtraction) ? nuxt.payload.data[key] : nuxt.static.data[key]
```
danielroe, line 118: I'm reluctant to merge this PR as it's really a feature to work around a bug.

What about instead supporting a `cache`/`store` option that defaults to `nuxt.isHydrating ? nuxt.payload.data : nuxt.static.data`? Then this could be set to a custom location, or forced to `nuxt.payload.data`, which would have the same effect as forcing payload extraction.

It could be part of https://github.com/nuxt/nuxt/issues/15445.

Thoughts @pi0?

danielroe (changes requested):
Would it be possible to update this instead to take the `cache`/`store` option? 🙏

Thoughts welcome on best API.

packages/nuxt/src/app/composables/asyncData.ts
pi0: More generic `cache` option with better defaults sound reasonably better alternative approach.

danielroe (approved):

docs/3.api/1.composables/use-async-data.md
```diff
@@ -31,6 +31,7 @@ type AsyncDataOptions<DataT> = {
   transform?: (input: DataT) => DataT
   pick?: string[]
   watch?: WatchSource[]
+  getCachedData?: (key: string) => any
```
manniL, line 34: Why is it `any` here but `getCachedData?: (key: string) => DataT` in the TS files 🤔

danielroe: It should be the same 👍

manniL: Will send a PR ☺️

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -235,7 +237,7 @@ export function useAsyncData<
     if (getCurrentInstance()) {
       onServerPrefetch(() => promise)
     } else {
-      nuxt.hook('app:created', () => promise)
+      nuxt.hook('app:created', async () => { await promise })
```
Mini-ghost, line 240: @darioferderber I'm curious about the reason behind the modification made to this line of code. Could you please explain it?

Mini-ghost: Oh I got it! because type defined in here `Promise<void>`

https://github.com/nuxt/nuxt/blob/5bb6a13eaeb5a283f0021b2144824ab1459ab82e/packages/nuxt/src/app/nuxt.ts#L25-L29

jswhisperer:
Sorry to message on an old PR, but I can't find any docs anywhere that actually say what storage mechanism client side is used with `getCachedData` or if it's customisable.

The API reminds me of CacheStorage with ServiceWorker, is that the case?

PR #20748, opened 2023-05-09
feat(nuxt): add `moduleBuilderMode`

danielroe:
### ❓ Type of change

### 📚 Description

This adds a new `moduleBuilderMode` toggle that can be switched on specifically when generating types (`nuxi prepare --module`). When executed this way it will:

1. by default look for and enable `src/module.ts`
2. turn off global auto-import types
3. write all templates to the `.nuxt` directory for auto-completing `#build` paths in development

This could completely replace `nuxi prepare playground` in many cases, meaning no need for `.nuxtrc` and also meaning playgrounds would be able to retain their own types and auto-imports.

We can also add more specific features later on that are particularly relevant for module authors when this is enabled.

**Note**: this could also be added directly to nuxt/module-builder as a special command e.g. `module-builder --prepare` to try it out first...

### 📝 Checklist


harlan-zw (approved):
Amazing

pi0:
Love this idea! But i guess would be much better if we first try to implement with `nuxt-module-builder` command for stub support. It can make behavior aligned with the version+configuration of module builder and using nuxi programmatic API to override required options.

Since we already integrated `nuxt builld-module`, this feature can be seamless integrated with nuxi as well.

PR #20749, opened 2023-05-09
fix(nuxt): add types for webpack/vite environments

### LINKED ISSUE #14131: add asset import typings for webpack and vite builders

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14131

### ❓ Type of change

### 📚 Description

This adds asset typings and other environment types that are more precisely tailored to the builder the end user is using.

~~With https://github.com/nuxt/nuxt/pull/20748,~~ When `@nuxt/module-builder` is detected, we can disable this so that our crafted types in `@nuxt/schema/env` can be used to prevent wrongly relying on features specific to just one of the builders (https://github.com/nuxt/framework/pull/5376#issuecomment-1177938331).

We also avoid polluting the global context when importing from `@nuxt/schema` directly.

### 📝 Checklist


packages/schema/src/config/typescript.ts
```diff
@@ -13,6 +14,14 @@ export default defineUntypedSchema({
      */
     strict: true,
 
+    /**
+     * whether or not to write declaration files to add specific types based on your builder environment. It is
+     * advised to turn this off for module authors, who should support multiple possible builders.
+     */
+    builderEnv: {
+      $resolve: async (val, get) => val ?? await resolvePath('@nuxt/module-builder', { url: await get('modulesDir') }).then(path => !path).catch(() => null) ?? true,
```
pi0, line 22: If both are installed in a mono-repo, this can make issues. Can we simply support `typescript.builder` as a way to override default (top level) `builder` only for types and also use `vite | webpack` instead of package names?

danielroe: I'm also very happy to add the override directly from nuxt/module-builder. 👍

pi0 (approved):
Small nit pick: It is `build(er)` vs `build-env`. Othwrwise lgtm...

PR #20755, opened 2023-05-10
fix(vite): expand `fs.allow` dirs to include app files

### LINKED ISSUE #14295: Vite unable to load modules when components are outside project directory
### Environment

Stackblitz, using components config:
```
- Operating System: `Linux`
- Nuxt Version:     `3.0.0-rc.4`
- Builder:          `vite`
- User Config:      `components`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

Local, using component:dirs hook:
```
- Operating System: `Darwin`
- Node Version:     `v14.19.3`
- Nuxt Version:     `3.0.0-rc.4`
- Package Manager:  `npm@6.14.17`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `hooks`, `modules`, `app`, `css`
- Runtime Modules:  `@storyblok/nuxt@4.0.1`
- Build Modules:    `-`
```

### Reproduction

Repro: https://stackblitz.com/edit/github-tncnqa

What I did (things to note):
1. moved the nuxt project into a `nuxt` directory
2. configured stackblitz to be able to work with this structure by adding a `.stackblitzrc` file
3. imported the markdown dependency as a test case
4. configured nuxt/vue to globally load components from the parent (`/components`) directory (using `components` config, but same problem happens if I used `components:dirs` hook)
5. added an `Inner.vue` and `Outer.vue` components - they basically convert and display some markdown to html

### Describe the bug

When adding components from outside the project directory, loading modules in those components messes up. 
For example, if the `makrdown` module was supposed to be imported, this error pops up during build:
> Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/projects/github-tncnqa/nuxt/markdown' imported from /home/projects/github-tncnqa/nuxt/.nuxt/dist/server/server.mjs

### Additional context

I'm not sure if this relates to nuxt or is purely a vite problem, unfortunately I don't know enough to find the problem and I have no idea how to make a reproduction with vite only.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14295

### ❓ Type of change

### 📚 Description

This adds the directories from app dependencies (like component folders) directly into the vite allow list to avoid potential issues. It's of course totally possible for a user to do this themselves but it can add friction. wdyt @antfu?

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -33,12 +33,14 @@ export async function bundle (nuxt: Nuxt) {
     nuxt.options.appDir,
     nuxt.options.workspaceDir,
     ...nuxt.options._layers.map(l => l.config.rootDir),
-    ...nuxt.apps.default?.components.map(c => dirname(c.filePath)),
-    ...nuxt.apps.default?.plugins.map(p => dirname(p.src)),
-    ...nuxt.apps.default?.middleware.map(m => dirname(m.path)),
-    ...Object.values(nuxt.apps.default?.layouts || {}).map(l => dirname(l.file)),
-    dirname(nuxt.apps.default.rootComponent!),
-    dirname(nuxt.apps.default.errorComponent!)
+    ...Object.values(nuxt.apps).flatMap(app => [
+      ...app.components.map(c => dirname(c.filePath)),
+      ...app.plugins.map(p => dirname(p.src)),
+      ...app.middleware.map(m => dirname(m.path)),
+      ...Object.values(app.layouts || {}).map(l => dirname(l.file)),
+      dirname(nuxt.apps.default.rootComponent!),
+      dirname(nuxt.apps.default.errorComponent!)
+    ])
   ].filter(d => d && existsSync(d))
 
   for (const dir of allowDirs) {
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -1,5 +1,6 @@
+import { existsSync } from 'node:fs'
 import * as vite from 'vite'
-import { join, resolve } from 'pathe'
+import { dirname, join, resolve } from 'pathe'
 import type { Nuxt, ViteConfig } from '@nuxt/schema'
 import { addVitePlugin, isIgnored, logger, resolvePath } from '@nuxt/kit'
 import replace from '@rollup/plugin-replace'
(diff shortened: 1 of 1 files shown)


packages/vite/src/vite.ts
```diff
@@ -27,6 +28,23 @@ export async function bundle (nuxt: Nuxt) {
...
     (nuxt.options.vite.devBundler === 'vite-node' && nuxt.options.dev)
   const entry = await resolvePath(resolve(nuxt.options.appDir, useAsyncEntry ? 'entry.async' : 'entry'))
+
+  let allowDirs = [
+    nuxt.options.appDir,
+    nuxt.options.workspaceDir,
+    ...nuxt.options._layers.map(l => l.config.rootDir),
+    ...nuxt.apps.default?.components.map(c => dirname(c.filePath)),
+    ...nuxt.apps.default?.plugins.map(p => dirname(p.src)),
+    ...nuxt.apps.default?.middleware.map(m => dirname(m.path)),
```
pi0, line 38: To avoid future refactor, shall we iterate over all possible `apps` keys?

packages/vite/src/vite.ts
```diff
@@ -27,6 +28,23 @@ export async function bundle (nuxt: Nuxt) {
...
+    ...Object.values(nuxt.apps.default?.layouts || {}).map(l => dirname(l.file)),
+    dirname(nuxt.apps.default.rootComponent!),
+    dirname(nuxt.apps.default.errorComponent!)
+  ].filter(d => d && existsSync(d))
```
pi0, line 42: I guess this list makes some duplicates (unnecessary nested paths `/foo` covers `/foo/bar`) but i guess should be fine for performance /cc @antfu ? (or a sort might help at least)

danielroe: This PR does handle nested paths - see:

https://github.com/nuxt/nuxt/blob/49bfa2d37984473727ad573b50ffe079558c6457/packages/vite/src/vite.ts#L44-L46

Deduplicating is certainly worth it though.

packages/vite/src/vite.ts
pi0: We might try multi-app via an experimental module as well. If core (or any modules might start depending on `nuxt.apps` can be ready for it, it could be a bonus  :)

danielroe: sure, makes sense - but I think it's worth noting (for the benefit of any modules) that we definitely do not support multi-app yet. we still need a team discussion first.

pi0: For sure :) It is only placeholder for the future.

pi0 (approved):

PR #20763, opened 2023-05-10
feat(kit): allow direct function in `defineNuxtModule`

ineshbose:
### ❓ Type of change

### 📚 Description

#20003 sought to allow plugins as objects and this is very helpful - especially for `meta` property.

This PR looks to allow some consistent definition behaviour between `defineNuxtModule` and `defineNuxtPlugin` as, say, a developer authors plugins as functions, switching to object-syntax causes a bit friction (still emphasising that object-syntax is preferred).

This reduces the need for minor syntactic sugar when defining internal modules where developers would have:

```ts
defineNuxtModule({
  setup (_, nuxt) {
    // ...
  }
})

// instead..
defineNuxtModule((_ nuxt) => {
  // ..
})
```

This should be pretty helpful since `modules/` directory is now scanned, so devs are also authoring these internal modules.

### 📝 Checklist


docs/2.guide/3.going-further/3.modules.md
```diff
@@ -127,7 +127,19 @@ export default function (inlineOptions, nuxt) {
...
+import { defineNuxtModule } from '@nuxt/kit'
+
+export default defineNuxtModule((options, nuxt) => {
+  nuxt.hook('pages:extend', pages => {
+    console.log(`Discovered ${pages.length} pages`)
+  })
+})
+```
+
+However, outside of short inline modules defined in `nuxt.config.ts`, **we do not recommend** using this low-level function definition. Instead, to define a module, **we recommend** using the object-syntax with `meta` property to identify your module.
```
danielroe, line 142: I think we don't need a helper for inline modules in `nuxt.config` because the types should be automatically inferred when used within the configuration definition.

But you raise a good point here. My main reservation is that by allowing this functional syntax we can miss out on the opportunity to de-duplicate modules (as there is no metadata).

Thoughts welcome from others.

ineshbose: Maybe mentioning `nuxt.config.ts` could be a bit confusing as this is the module author guide, while inline-function usage is shown in configuration - https://nuxt.com/docs/api/configuration/nuxt-config#modules

Suggested change:
```suggestion
However, **we do not recommend** using this low-level function definition. Instead, to define a module, **we recommend** using the object-syntax with `meta` property to identify your module.
```

ineshbose: Maybe mentioning `nuxt.config.ts` could be a bit confusing as this is the module author guide, while inline-function usage is shown in configuration - https://nuxt.com/docs/api/configuration/nuxt-config#modules

Suggested change:
```suggestion
However, **we do not recommend** using this low-level function definition. Instead, to define a module, **we recommend** using the object-syntax with `meta` property to identify your module.
```

danielroe:
@pi0 - would love your thoughts on this.

pi0 (approved):

PR #20770, opened 2023-05-10
feat(schema): enable rich json payloads by default

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/19205

### ❓ Type of change

### 📚 Description

This is a follow-up from https://github.com/nuxt/nuxt/pull/19205. We now enable rich JSON payload rendering by default. If this causes any issues in your projects, please do raise an issue to let us know (and, as a temporary workaround, you can set `experimental.renderJsonPayloads` to false):

```ts
  experimental: {
    renderJsonPayloads: false
  }
})
```

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/experimental.ts
+++ b/packages/schema/src/config/experimental.ts
@@ -93,9 +93,8 @@ export default defineUntypedSchema({
      */
     noScripts: false,
 
-    // TODO: enable by default in v3.5
     /** Render JSON payloads with support for revivifying complex types. */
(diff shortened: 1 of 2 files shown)


sharifzadesina:
I don't know about "issues" but it broke the whole application, previously I was able to use simple classes in `useAsyncData` but now I am not.
I need to simply disable this feature, otherwise, I get `Cannot stringify arbitrary non-POJOs` error.

PR #20779, opened 2023-05-11
fix(nuxt): ensure all dir parts are present in component name

### LINKED ISSUE #20613: Component name resolution is broken in some cases
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.3.3
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-nbj7gc?file=components/ItemHolder/ItemHolder.vue

### Describe the bug

components/ItemHolder/Item/Item.vue gets incorrectly resolved to Item
while
components/ItemHolder/Item1/Item1.vue correctly resolves to ItemHolderItem1

The problem is that here https://github.com/nuxt/nuxt/blob/8b86d39e30ca970fb0adf2b5665a77f8dbbfb439/packages/nuxt/src/components/scan.ts#L85 adding segments prematurely terminates because
components/ItemHolder
gets treated like
components/Item/Holder
would. 

### Additional context

_No response_

### Logs

_No response_

baiwusanyu-c:
### ❓ Type of change

### 📚 Description
close: #20613 

### 📝 Checklist


baiwusanyu-c:
close: #20613

danielroe (commented):
This looks great - thank you! ❤️ 

I've extracted your code into a separate helper function and wrote some unit tests to describe the behaviour I _think_ we are expecting, and which isn't entirely working yet... What do you think?

baiwusanyu-c:
@danielroe Thank you very much for your help. I think I misunderstood his rules and logic. Now I think we need to find the judgment condition and complete the prefixPart.

packages/nuxt/test/scan-components.test.ts
```diff
@@ -230,3 +243,21 @@ it('components:scanComponents', async () => {
   }
   expect(scannedComponents).deep.eq(expectedComponents)
 })
+
+const tests: Array<[string, string[], string]> = [
+  ['WithClientOnlySetup', ['Client'], 'ClientWithClientOnlySetup'],
+  ['Item', ['Item', 'Holder', 'Item'], 'ItemHolderItem'],
+  ['ItemHolder', ['Item', 'Holder', 'Item'], 'ItemHolderItem'],
```
danielroe, line 250: Oh, I think I might have added a wrong test  case here... 🤔

baiwusanyu-c: > Oh, I think I might have added a wrong test case here... 🤔

oh? 🤣, I was really a little confused when I modified it, I thought there were rules I didn't know. So what are the rules now?
These three are the output results of the use cases that failed to run the test before, I think only the second one is wrong
```
1.['ItemHolder', ['Item', 'Holder', 'Item'], 'ItemHolderItemHolder'],
2.['ItemHolderItem', ['Item', 'Holder', 'Item'], 'ItemHolderItemHolderItem'], // should be ItemHolderItem
3.['Item', ['Item', 'Item'], 'ItemItem'],
```

danielroe, line 250:
```suggestion
  ['ItemHolder', ['Item', 'Holder', 'Item'], 'ItemHolderItemHolderItem'],
```

danielroe:
```suggestion
  ['ItemHolder', ['Item', 'Holder', 'Item'], 'ItemHolderItemHolder'],
```

packages/nuxt/src/components/scan.ts
```diff
@@ -137,3 +121,38 @@ export async function scanComponents (dirs: ComponentsDir[], srcDir: string): Pr
...
+  // e.g Item/Holder/Item/ItemHolderItem.vue -> ItemHolderItem
+  if (fileNameParts.join('').toLowerCase() === prefixParts.join('').toLowerCase()) {
+    return pascalCase(fileNameParts)
+  }
+
+  // e.g Item/Item/Item.vue -> Item
+  if ([...new Set(prefixParts)].length === 1 &&
+    (fileNameParts[0] || '').toLowerCase() === (prefixParts[0] || '').toLowerCase()) {
+    return pascalCase(fileNameParts)
+  }
```
danielroe, line 143: I'm worried these two sections are too 'specific' to the test cases I provided rather than reproducible as an algorithm.

I think:
* let's revert back to your original proposed change without this kind of 'special-case' handling
* update the test cases to accurately describe the previous behaviour with the exception of the bug reported initially (so we don't end up _breaking_ anyone's expected component resolution)

baiwusanyu-c:
I added some tests based on the bug(` ['Item', ['Item', 'Holder', 'Item'], 'ItemHolderItem']`) and the test case you provided, and ran the original code and the original commit (3d6af90) separately, and found that they didn't pass the tests. 
| fileName | prefixParts     | expected | original code | 3d6af90|
|-----|----------|---------|---------|---------|
| WithClientOnlySetup | ['Client'] | ClientWithClientOnlySetup | ClientWithClientOnlySetup✅ | ClientWithClientOnlySetup ✅ |
| ItemHolderItem | ['Item', 'Holder', 'Item'] | ItemHolderItem| ItemHolderItem✅ | ItemHolderItemHolderItem❌ |
| Item | ['Item'] | Item| Item✅ | Item ✅ |
| Item | ['Item','Item'] | Item| Item✅ | ItemItem❌ |
| ItemTest | ['Item', 'Test'] | ItemTest | ItemTest✅ | ItemTest ✅ |
| ThingItemTest| ['Item', 'Thing'] | ThingItemTest| ThingItemTest✅ | ThingItemTest✅ |
| Item | ['Thing', 'Item']| ThingItem| ThingItem✅ | ThingItem✅ |
| Item| ['Item', 'Holder', 'Item'] | ItemHolderItem| Item❌ | ItemHolderItem✅ |
| ItemHolder| ['Item', 'Holder', 'Item'] | ItemHolderItemHolder | ItemHolder❌ | ItemHolderItemHolder✅ |
| ThingItemTest| ['Item', 'Thing', 'Foo'] | ItemThingFooThingItemTest| ItemThingItemTest❌ | ItemThingItemTest❌ |
| ItemIn| ['Item', 'Holder', 'Item', 'In'] | ItemHolderItemIn | ItemIn✅ | ItemHolderItemIn✅ |
| Item| ['Item', 'Holder', 'Test'] | ItemHolderTestItem| Item❌ | Item❌ |
| ItemHolderItem| ['Item', 'Holder', 'Item', 'Holder']| ItemHolderItemHolderItem| ItemHolderItem❌ | ItemHolderItemHolderItem✅ |

Especially `['Item', ['Item', 'Holder', 'Test'], 'ItemHolderTestItem']` and `['ThingItemTest', ['Item', 'Thing', 'Foo'], 'ItemThingFooThingItemTest ']`These two test cases.

So I used `startsWith` to modify the matching implementation, but he still failed the test of `['Item', ['Item', 'Item'], 'Item']`, so I added some judgments to ensure It passes the test and behaves the same as the original code

danielroe (approved):

PR #20795, opened 2023-05-11
fix(nuxt): avoid destructuring error prop (works around upstream bug)

### LINKED ISSUE #20789: Default error page does not work on Windows
### Environment

- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.4.0
- Package Manager:  npm@8.5.5
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Initial project with "nuxi init". Cannot reproduce on stackblitz, it seems this is a Windows-related issue.

### Describe the bug

For some reason [this](https://github.com/nuxt/nuxt/blob/0258acdc8a6f02bd5de5e31611b4c821481bb519/packages/nuxt/src/app/components/nuxt-error-page.vue#L10)  template compiles incorrectly. The code after "defineProps" tries to access "error" from global scope instead of "props" object, which leads to the "ReferenceError: error is not defined".

image

### Additional context

_No response_

### Logs

_No response_

Lehoczky:
### 🔗 Linked issue

Fixes #20789

### ❓ Type of change

### 📚 Description

This regression is probably caused by the changes regarding the experimental prop destructuring in Vue 3.3, which might not cover the edge case where someone is using destructuring without enabling this new experimental feature, although I haven't look really deep into this.

Fortunately it seems like we can avoid this bug by using a props variable and destructuring it later. This way the error object should still lose it's reactivity but now the compiled output is valid.

How I tested:

In the playground I changed `app.vue` to:
```vue
<script setup lang="ts"></script>

<template>
  <div>
    <NuxtLayout>
      <NuxtPage />
    </NuxtLayout>
  </div>
</template>

<style scoped></style>
```

Added an index page in `pages/index.vue`:
```vue
<template>
  <div>asd</div>
</template>
```

And navigated to a random url, like `/this-does-not-exist`, to get to the error page. Without changes I was able to reproduce the error in the linked ticket, but after the changes it works fine.

### 📝 Checklist


Lehoczky:
Also feel free to close this, if we want to wait for an upstream fix, this is really just a fast band aid solution for nuxt.

danielroe (approved):
Perfect - thanks!

Lehoczky:
@danielroe I created an upsteam issue: https://github.com/vuejs/core/issues/8289

PR #20806, opened 2023-05-12
fix(nuxt): allow `pages:extend` to enable pages module

### LINKED ISSUE #20801: <NuxtPage> info message does not look up for module pages
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.4.0
- Builder:          vite
- User Config:      modules
- Runtime Modules:  ./modules/test-module
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-hcx5za

### Describe the bug

If `pages/` directory is not created or is empty and all of the pages are defined using `extendPages` in module, it throws warning:

`Create a Vue component in the pages/ directory to enable <NuxtPage>`

and shows blank page.

If you create e.g. `pages/test.vue`, warning goes away and displays the intended page.

### Additional context

Also this warning goes away, if I create `app/router.options.ts` file.

### Logs

_No response_

darioferderber:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20801

### ❓ Type of change

### 📚 Description

Sets `pages` option to `true` if `extendPages` is called.
If `extendPages` gets called, it means that there are actually some pages, so we can set te `pages` option to `true`?

\+ updates typo in documentation

### 📝 Checklist


edgarsn:
Thank you for your fast work, but I think there's one edge case left behind which is not my case, but may happen to others later.

You can also extend pages by using hook - `pages:extend` instead of `extendPages()` function.

Shouldn't it be better to set it after `await nuxt.callHook('pages:extend', pages)` [here](https://github.com/nuxt/nuxt/blob/f7f4a7643c352ddd9e8daa597d9b61717c5b593d/packages/nuxt/src/pages/module.ts#LL247C5-L247C5) ? If pages array after hook call is not empty, then surely we can set `nuxt.options.pages = true;` right?

danielroe:
The safest thing is probably to do this within `isPagesEnabled`, directly before returning false:

```ts
const pages = await resolvePagesRoutes()
await nuxt.callHook('pages:extend', pages)
if (pages.length) {
  return true
}
```

We can keep the previous early returns for `true` as a more performant fast-path.

darioferderber:
I agree, but that means that the hook will be called multiple times if there are no pages, so I wasn't sure if that's a good idea.

danielroe (approved):

PR #20835, opened 2023-05-13
fix(nuxt): transform client fallbacks directly on SFCs

### LINKED ISSUE #20833: NuxtClientFallback hydration broken in webpack build
### Environment

- Node Version:     v18.15.0
- Nuxt Version:     3.4.1
- Nitro Version:    2.3.3
- Package Manager:  pnpm@8.1.1
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

- run `npx cross-env TEST_BUILDER=webpack pnpm nuxi build .\test\fixtures\basic\` on nuxt repo
- run it there will be a hydration issue. VNode will not be mounted correctly

### Describe the bug

A silent bug (no warning or error in the console). When using wbpack as a builder, `NuxtClientFallback` has a strange behavior, mounting its DOM node without discarding the old nodes

### Additional context

found with https://github.com/nuxt/nuxt/pull/20336

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix #20833 

### ❓ Type of change

### 📚 Description
Hi :wave: this PR fix an issue on `NuxtClientFallback` with webpack only. 
The fallback dom node are not removed client-side. However i have no clue on why does `{ type: ['template'] }` cause this issue

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/client-fallback-auto-id.ts
+++ b/packages/nuxt/src/components/client-fallback-auto-id.ts
@@ -25,7 +25,7 @@ export const clientFallbackAutoIdPlugin = createUnplugin((options: LoaderOptions
       if (include.some(pattern => id.match(pattern))) {
         return true
       }
-      return isVue(id, { type: ['template'] })
+      return isVue(id)
     },
     transform (code, id) {
       if (!CLIENT_FALLBACK_RE.test(code)) { return }
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -343,7 +343,8 @@ describe('pages', () => {
     // ensure components reactivity once mounted
     await page.locator('#increment-count').click()
     expect(await page.locator('#sugar-counter').innerHTML()).toContain('Sugar Counter 12 x 1 = 12')
-
+    // #20833
+    expect(await page.locator('body').innerHTML()).not.toContain('Hello world !')
     await page.close()
   })
 


packages/nuxt/src/components/client-fallback-auto-id.ts
```diff
@@ -25,7 +25,7 @@ export const clientFallbackAutoIdPlugin = createUnplugin((options: LoaderOptions
       if (include.some(pattern => id.match(pattern))) {
         return true
       }
-      return isVue(id, { type: ['template'] })
+      return isVue(id)
```
danielroe, line 28: it seems likely we need to update the `isVue` helper util to correctly identify template Vue imports within webpack, so we can preserve the performance improvement of restricting this transform to just vue templates.

danielroe: actually, I think your solution is right here, as we're directly replacing text in the SFC...

danielroe (approved):

packages/nuxt/src/components/client-fallback-auto-id.ts
huang-julien: i just begin to understand how it work, 
image
here's a `console.log(id, isVue(id), isVue(id, { type: ['template'] })` 🙂

PR #20836, opened 2023-05-13
perf(nuxt): use granular watcher to avoid crawling ignored dirs

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/12376

### ❓ Type of change

### 📚 Description

This follows up previous focus on watcher improvements (https://github.com/nuxt/nuxt/pull/20179) as a hybrid solution using `chokidar` - it filters out/ignores child directories within `srcDir` directly. In tests in the monorepo, it reduces chokdar startup on a M1 Mac by 50%. I would expect even more significant performance improvements on a windows machine.

To enable this you'll need to set `experimental.watcher` to `granular`. (Otherwise behaviour will stay the same as currently.)

I'd appreciate review - my main concern is whether we ought to make provision for users _deliberately_ seeking to watch changes in `.nuxt` or `node_modules` (which they have enabled with a negative glob pattern in `ignore`). 

cc: @stafyniaksacha

### 📝 Checklist


packages/nuxt/src/core/builder.ts
```diff
@@ -92,23 +93,41 @@ async function watch (nuxt: Nuxt) {
...
+    watcher.on('all', (event, path) => {
+      if (!pending) {
+        nuxt.callHook('builder:watch', event, normalize(path))
+      }
+      if (event === 'unlinkDir' && path in watchers) {
+        watchers[path].close()
+        delete watchers[path]
+      }
+      if (event === 'addDir' && path !== dir && !ignoredDirs.has(path) && !(path in watchers) && !isIgnored(path)) {
+        watchers[path] = chokidar.watch(path, { ...nuxt.options.watchers.chokidar, ignored: [isIgnored] })
```
stafyniaksacha, line 119: Are the options `ignoreInitial: false, depth: 0` missed voluntary there?

packages/schema/src/config/experimental.ts
```diff
@@ -165,10 +165,13 @@ export default defineUntypedSchema({
...
      * on Windows platforms.
      *
+     * You can also try setting this to `granular` to use an experimental granular
+     * watcher, which ignores top-level directories (like `node_modules` and `.git`).
+     *
      * @see https://github.com/paulmillr/chokidar
      * @see https://github.com/parcel-bundler/watcher
      * @default chokidar
-     * @type {'chokidar' | 'parcel'}
+     * @type {'chokidar' | 'parcel' | 'granular'}
```
antfu, line 174:
```suggestion
     * @type {'chokidar' | 'parcel' | 'chokidar-granular'}
```

Do you think it makes sense to do so, indicating it's still chokidar?

antfu (approved):
Amazing idea! Love the approach!

atinux (approved):

PR #20840, opened 2023-05-14
fix(nuxt): get fallback for `<DevOnly>` from parsed html

ineshbose:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/20817#discussion_r1192966653

### ❓ Type of change

### 📚 Description

It should now allow the documented example as well:

```html
    <DevOnly>
      <!-- arrows before fallback or... -->
       <ClientOnly>
          <template #fallback>
            <div>test</div>
          </template>
      </ClientOnly>
      <template #fallback>
        <div></div>
      </template>
    </DevOnly>
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/plugins/dev-only.ts
+++ b/packages/nuxt/src/core/plugins/dev-only.ts
@@ -3,6 +3,7 @@ import { stripLiteral } from 'strip-literal'
 import { parseQuery, parseURL } from 'ufo'
 import MagicString from 'magic-string'
 import { createUnplugin } from 'unplugin'
+import { type Node, parse } from 'ultrahtml'
 
 interface DevOnlyPluginOptions {
   sourcemap?: boolean
@@ -10,7 +11,6 @@ interface DevOnlyPluginOptions {
 
 export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions) => {
   const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/g
-  const FALLBACK_SLOT_RE = /<template\s+#fallback>(?<fallback>[\s\S]*?)<\/template>/
 
   return {
     name: 'nuxt:server-devonly:transform',
@@ -30,7 +30,10 @@ export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions) => {
       const s = new MagicString(code)
       const strippedCode = stripLiteral(code)
       for (const match of strippedCode.matchAll(DEVONLY_COMP_RE) || []) {
-        const replacement = match[0].match(FALLBACK_SLOT_RE)?.groups?.fallback || ''
+        const ast: Node = parse(match[0]).children[0]
+        const fallback: Node | undefined = ast.children?.find((n: Node) => n.name === 'template' && Object.values(n.attributes).includes('#fallback'))
+        const replacement = fallback ? match[0].slice(fallback.loc[0].end, fallback.loc[fallback.loc.length - 1].start) : ''
+
         s.overwrite(match.index!, match.index! + match[0].length, replacement)
       }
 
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/plugins/dev-only.ts
```diff
@@ -9,7 +9,8 @@ interface DevOnlyPluginOptions {
 }
 
 export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions) => {
-  const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly)>[^<]*(?:<template\s+#fallback>(?<fallback>[\s\S]*?)<\/template>)?[\s\S]*?<\/(?:dev-only|DevOnly)>/g
+  const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/g
+  const FALLBACK_SLOT_RE = /<template\s+#fallback>(?<fallback>[\s\S]*?)<\/template>/
```
huang-julien, line 13: This regex is working but only if there's no multiple `<template #fallback>` within the string. 
For example: 
````html
 <DevOnly>
    <template #fallback>
      <div><!-- empty div for flex.justify-between --></div>
      <ClientOnly>
        <template #fallback>
          <div>test</div>
        </template>
      </ClientOnly>
    </template>
</DevOnly>
````
will get you
````html
<div><!-- empty div for flex.justify-between --></div>
      <ClientOnly>
        <template #fallback>
          <div>test</div>
````
Also if you set a fallback somewhere before the `<template #fallback>` of devonly, it will catch what's inside the `<template #fallback>.

````html
 <DevOnly>
    <ClientOnly>
      <template #fallback>
        <div>test</div>
      </template>
    </ClientOnly>
    <template #fallback>
      <div><!-- empty div for flex.justify-between --></div>
    </template>
</DevOnly>
````
will get you
````html
      <template #fallback>
        <div>test</div>
      </template>
````

but i think the risk of someone using another `<template #fallback>` should be very low 😅. Yet, there's still a risk of someone doing it (is this murphy's law ?).

Comment (huang-julien) [reply packages/nuxt/src/core/plugins/dev-only.ts:13]: i didn't forgot to submit the review this time :joy:

Comment (ineshbose) [reply packages/nuxt/src/core/plugins/dev-only.ts:13]: Ooh that's a very good point. Let me brainstorm this

The thing with `ultrahtml` is that I'd think of using the sanitizer utils it provides (may increase dependency bundle) but that would also involve regex at a point while this is fast and simple so currently progressing on this

**Comment by ineshbose** [comment]
An app.vue that should cover the bases to test:

```html
<script setup lang="ts">
</script>

<template>
  <!-- Edit this file to play around with Nuxt but never commit changes! -->
  <div>
    Nuxt 3 Playground
    <DevOnly>standard</DevOnly>
    <DevOnly>
      standard
      <template #fallback>
        <span>Fallback content</span>
      </template>
    </DevOnly>

    <DevOnly>
      <div>Element</div>
    </DevOnly>
    <DevOnly>
      <div>Element</div>
      <!-- comment -->
      <template #fallback>Fallback content2</template>
    </DevOnly>
    <DevOnly>
      Empty fallback
      <!-- empty fallback -->
      <template #fallback></template>
    </DevOnly>

    <DevOnly>
      <div>dev only</div>
      <ClientOnly>
        client only
        <div>client div</div>
        <template #fallback>
          <div>client fallback</div>
        </template>
      </ClientOnly>
      <template #fallback>
        <div>dev fallback</div>
      </template>
    </DevOnly>
  </div>
</template>

<style scoped>
</style>
```

packages/nuxt/src/core/plugins/dev-only.ts
ineshbose: Gone for a hybrid approach that allowed this to be done without a walker and needing to be async 😄

huang-julien (approved):
:+1: working wonderfully on my side

danielroe (approved):

PR #20852, opened 2023-05-15
feat: form actions for progressive enhancement

Hebilicious:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/20649

### ❓ Type of change

### 📚 Description

Here's a standalone module with the most up-to-date version : https://github.com/Hebilicious/form-actions-nuxt

This PR is a POC that attempts to implements #20649. Almost everything is implemented as a Nuxt Module.

It adds the following helper features to Nuxt or Nitro : 

- `defineServerLoader`
- `defineFormActions`
- `actionResponse`

It adds the following helpers to H3 : 

- `getRequestFromEvent`
- `respondWithResponse`
- `respondWithRedirect`
- `getFormData`

All of these should only be available on the server side (ie `/server` directory)
*It could be argued that some of these helpers might be overkill and should be implemented by users, but for the sake of nice DX I added them anyways*

It adds the `useFormAction` composable to Nuxt.
And the `v-enhance` directive.

Finally it adds the possibility to Nitro to specify fallback routes (`nitro.config.routesWithFallback`) to an array of route objects through the configuration (`{ route: "api/todos", fallbackTo: '/**' }`). The fallback route will be matched with a GET method. 
Effectively This allows creating POST endpoints on the same route than pages.

There's a classic todo  example with optimistic updates for the delete method.

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "cheerio": "1.0.0-rc.12",
     "consola": "3.1.0",
     "devalue": "4.3.2",
-    "eslint": "8.42.0",
+    "eslint": "8.43.0",
     "eslint-import-resolver-typescript": "3.5.5",
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-jsdoc": "41.1.2",
@@ -67,9 +67,9 @@
     "nuxt": "workspace:*",
     "ofetch": "1.1.0",
     "pathe": "1.1.1",
-    "playwright": "1.35.0",
+    "playwright": "1.35.1",
     "rimraf": "5.0.1",
-    "semver": "7.5.1",
+    "semver": "7.5.2",
     "std-env": "3.3.3",
     "typescript": "5.0.4",
     "ufo": "1.1.2",
@@ -77,7 +77,7 @@
     "vitest": "0.32.2",
     "vue": "3.3.4",
     "vue-eslint-parser": "9.3.1",
-    "vue-tsc": "1.6.5"
+    "vue-tsc": "1.8.0"
   },
   "packageManager": "pnpm@8.6.2",
   "engines": {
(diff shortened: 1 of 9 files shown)


Hebilicious:
@pi0 I would like to know your thoughts about the proposed changes to Nitro, and if you think the extra H3 helpers should go into H3 too.
If you're happy with the way it looks, I will open PR in nitro and h3, and the rest of the changes in this PR can be distributed as a module so people can start playing around with it.

@danielroe There's a few things that I think could increase things nicely, but I'm not sure how to approach them. 

- HMR for the loaders : 
I'm thinking with a custom hook that triggers the logic that extract the loaders and add them to `public.__serverLoaders__` this could be implemented. You also mentioned using virtual files, however I think this would requires more changes to nitro (ie to import them)
- Type extraction for the loaders : 
We could have some Typescript magic going on to have the loaders data typed in the template, which isn't currently the case. `useFormAction` call `useFetch` under the hood (`const { data: result, refresh, pending } = useFetch(url, { watch, immediate: true, server: true })`) so maybe this isn't too hard to achieve ? 
- Automatically bind v-enhance to the first form.
Instead of this ...
```html
<script setup>
const { enhance } = useFormAction()  //This would automatically bind to the form.
</script>
<template>
  <form method="POST" v-enhance="enhance">
	 ...
  </form>
</template>
``` 
... we could have this.
```html
<script setup>
useFormAction() 
</script>
<template>
  <form method="POST">
     ...
  </form>
</template>
``` 
imo this isn't super important but it makes things neater in the simple use cases.
- DevTools integration
I'm actually not sure what that should look like, but I'm open to suggestions.

danielroe:
Marking as draft until we have time to test the amazing implementation work done at https://github.com/Hebilicious/form-actions-nuxt. ❤️

PR #20859, opened 2023-05-15
fix(nuxt): pass from + savedPosition to first `scrollBehavior`

### LINKED ISSUE #15568: scrollBehavior gets called two times on initial page load
### Environment

Nuxi 3.0.0                                                                                                                                                                             

- Operating System: `Darwin`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@8.5.5`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-drvjtv?file=app%2Frouter.options.ts,app.vue

### Describe the bug

When using `app/router.options.ts` to set a `scrollBehavior` callback for example to restore the old scroll position or scroll to an anchor it gets called two times on initial page load. 

Check the console output for `from` and `to` route information. It seems that the layout system is somehow messing with the routing as you can see in `to.meta` for the first and second call. Also `from.name`, `from.path` and `from.fullPath` are different.

Let me know if I can be of further help.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15568

### ❓ Type of change

### 📚 Description

This PR ensures we only call `scrollBehavior` once on loading a new page, and we pass the saved position correctly through to it.

### 📝 Checklist


packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -61,9 +62,17 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
...
+    let startPosition: Parameters<RouterScrollBehavior>[2] | null
     const initialURL = process.server ? nuxtApp.ssrContext!.url : createCurrentLocation(routerBase, window.location)
     const router = createRouter({
       ...routerOptions,
+      scrollBehavior: (to, from, savedPosition) => {
+        if (from === START_LOCATION) {
+          startPosition = savedPosition
+          return
+        }
+        return routerOptions.scrollBehavior?.(to, from, nuxtApp.isHydrating ? startPosition : savedPosition)
```
posva, line 74: I don't know which one is better: using `nuxtApp.isHydrating` or using `startPosition` only once and then using saved position. It seems like you reset it anyway in line 206. Otherwise this looks good

PR #20894, opened 2023-05-16
fix(nuxt): allow island renders without `/` route

danielroe:
### ❓ Type of change

### 📚 Description

Previously we were throwing a 404 if the root `~/pages/index.vue` path didn't exist - server components would silently fail to render. 😬

### 📝 Checklist


huang-julien (approved):
indeed 😮  thanks 🙌 !

PR #19624, opened 2023-03-12
fix(nuxt): prefer using currentInstance if available in `useNuxtApp`

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/19623

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
hi :wave: this is a workaround for https://github.com/nuxt/nuxt/issues/19623

when using `useNuxtApp` there is a small risk of CRSP with sync server components. However this risk gets higher with async components.
This keeps the current implementation,_ it just prefer using the current instance of the component if it is available.

Needed for https://github.com/nuxt/nuxt/pull/19605

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #15164, opened 2022-10-14
Server built on Windows not handling simultaneous requests


JeffreyTan1:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v16.16.0`
- Nuxt Version:     `3.0.0-rc.11`
- Nitro Version:    `0.5.4`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `modules`, `privateRuntimeConfig`, `publicRuntimeConfig`, `build`, `vite`, `typescript`, `dev`
- Runtime Modules:  `@pinia/nuxt@0.4.2`, `@nuxtjs/tailwindcss@5.3.3`
- Build Modules:    `-`

### Reproduction

- Clone https://github.com/JeffreyTan1/nuxt-bug-reproduction

I scaffolded the project like so https://v3.nuxtjs.org/getting-started/installation/ and the api from here https://v3.nuxtjs.org/guide/directory-structure/server/

- I use JMeter to simulate multiple users hitting the server at once. You can get JMeter from here: https://archive.apache.org/dist/jmeter/binaries/ I have the distribution found at apache-jmeter-5.4.3.zip.

As a quick tutorial of how to setup a JMeter stress test on Windows:

- Unzip the download
- Go into the bin folder
- Open the ApacheJMeter JAR file (opens the GUI)
- [Create thread group](https://i.stack.imgur.com/hjXG6.png) and set the following [properties](https://i.stack.imgur.com/i1KFf.png).
- [Create a HTTP Request sampler](https://i.stack.imgur.com/e3AVh.png) and set the following [properties](https://i.stack.imgur.com/9kZ66.png)
- [Create a View Results Tree listener](https://i.stack.imgur.com/rkCKt.png) and stay in this window
- Click the green play button to run the test and the gear+broomstick button to clear the output

You should see 200 responses interspersed by 503 responses. 
![image](https://user-images.githubusercontent.com/67099352/195734695-813f7e47-22ae-4664-a58c-881b58342eb6.png)

### Describe the bug

The Nuxt server is not handling simultaneous requests and is only happening on Windows.

Sometimes, when two or more requests hit the server at the same time, only one of them gets served, and the other requests return with a 503 Server Unavailable response instead of being added to the queue.

I have tested on ARM and Intel MacOS systems as well as WSL - all of which do not have the same problem. I have also tried Node v14.16 and LTS (v16.17.1). 

[...]

PR #20910, opened 2023-05-17
fix(nuxt): prioritise vue app context when available

### LINKED ISSUE #20741: Cross-site request pollution with useAsyncData
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.4.3`
- Nitro Version:    `2.4.0`
- Package Manager:  `npm@9.5.0`
- Builder:          `vite`
- User Config:      `extends`, `shopware`, `runtimeConfig`, `modules`, `i18n`, `css`, `vite`, `router`, `nuxtShopwareCaching`, `prismic`, `app`
- Runtime Modules:  `modules/domainRuntimeInfo`, `@nuxtjs/prismic@3.0.0-rc.5`, `@mothership-gmbh/nuxt-shopware-caching@2.0.0`, `nuxt-schema-org@2.1.3`
- Build Modules:    `-`

### Reproduction

I'm working on a minimal reproduction.

### Describe the bug

When I request my app nearly at the same time with more than one request, there seems to be some cross-request pollution. See the screenshot. 
Bildschirmfoto 2023-05-09 um 12 32 25

The markup rendered on the server side is actually correct for each request and logging on the server side also shows that the values are present and correct for each request.
But if I inspect `window.__NUXT__` on the client side, the second request has the data of the other request in there, too, while the first request has it set to null.

I simplified the API-request that is normally done inside this useAsyncData.
```
const route = useRoute()

const returnValue = await useAsyncData('cmsResponse' + route.path, async () => {
  return {
    routeName: 'frontend.detail.page',
    foreignKey: '9855e4444b20459bbf9b94cc4f1c0c4c',
  }
})
```

Does anyone have an idea, why the second request can have the data of the first one?

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20741
related https://github.com/nuxt/nuxt/pull/19624#issuecomment-1465838519
proper fix will need to come https://github.com/vuejs/core/pull/7743

### ❓ Type of change

### 📚 Description

Now that we have access to `hasInjectionContext` we can ensure we only check for current vue instance when we are within a component (or `getCurrentInstance` should fail).

TODO:
- [x] confirm that https://github.com/vuejs/core/issues/7733 doesn't cause request to be shared when using this strategy

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -435,7 +435,7 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp |
 /**
  * Returns the current Nuxt instance.
  */
-export function useNuxtApp () {
+export function useNuxtApp (): NuxtApp {
   let nuxtAppInstance
   if (hasInjectionContext()) {
     nuxtAppInstance = getCurrentInstance()?.appContext.app.$nuxt
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-use-before-define */
-import { getCurrentInstance, reactive } from 'vue'
+import { getCurrentInstance, hasInjectionContext, reactive } from 'vue'
 import type { App, Ref, VNode, onErrorCaptured } from 'vue'
 import type { RouteLocationNormalizedLoaded } from '#vue-router'
 import type { HookCallback, Hookable } from 'hookable'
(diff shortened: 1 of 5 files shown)


danielroe:
The test failure we're experiencing is a validation of https://github.com/nuxt/nuxt/issues/15164, meaning that the Nuxt test fixture is a reproduction (cc: @pi0).

pi0 (approved):
LGTM 💯

pi0:
> The test failure we're experiencing is a validation of https://github.com/nuxt/nuxt/issues/15164, meaning that the Nuxt test fixture is a reproduction (cc: @pi0).

I will test this afterwards if you can confirm only dev tests are failing.

danielroe:
@pi0 - Yes, that's the error code: https://github.com/nuxt/nuxt/actions/runs/5002108510/jobs/8961711745.

iijaachok:
Had to rewrite the implementation of a test because of SSR payload structure changes, however after running a few tests I couldn't reproduce the issue anymore even with higher volume of requests - about 100 requests/sec.

Thanks for your help! 👍

PR #20916, opened 2023-05-17
fix(nuxt): don't `refresh` when hydrating when data is present

### LINKED ISSUE #20900: useAsyncData with immediate false and call execute after makes fetching the data twice
### Environment

- Node Version:     v16.14.0
- Nuxt Version:     3.5.0
- Nitro Version:    2.4.0
- Package Manager: pnpm@8.3.1
- Builder: vite
- User Config: vite
- Runtime Modules: -
- Build Modules: -

### Reproduction

https://stackblitz.com/edit/github-btqy9q?file=server%2Fapi%2Ftemp.ts,app.vue

### Describe the bug

When useAsyncData with `immediate` option `false` and call later by `execute` or `refresh` method, it makes fetching data twice and `Hydration mismatch` if result of api call is different.  I chekced `NuxtApp payload`, it has cached data, but it seems it doesn't use it. Is this intended behavior?

image

I also have other [questions in discussions](https://github.com/nuxt/nuxt/discussions/20885). if possible, plz answer those too. Thanks!

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20900

### ❓ Type of change

### 📚 Description

It's an edge case, but when calling `execute` or `refresh` we will force refresh data on initial load of a component (ie. when hydrating), because we don't pass the `{ _initial: true }` option by default.

This PR updates behaviour to _default_ to skipping refresh of data if it exists in payload and we're hydrating. It can still be overridden by passing `{ _initial: false }`. I don't regard this as a breaking change as changing data in the payload will almost always be incorrect and cause a hydration failure.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -135,7 +135,7 @@ export function useAsyncData<
       (nuxt._asyncDataPromises[key] as any).cancelled = true
     }
     // Avoid fetching same key that is already fetched
-    if (opts._initial && hasCachedData()) {
+    if ((opts._initial || nuxt.isHydrating) && hasCachedData()) {
       return getCachedData()
     }
     asyncData.pending.value = true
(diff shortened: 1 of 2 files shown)


pi0 (approved):
makes sense

Issue #14269, opened 2022-07-05
using try/catch with an async method inside a middleware causes a `nuxt instance unavailable` error.


amrnn90:

radudalbea:
Hi,

I am having the same issue but with a request from a created method. Basically I am trying to do a simple fetch outside the setup function.

async created() {
      await this.fetchData();
}

In the fetch method I'm trying to call nuxt auth composable to retrieve the status of the user.

const { status, data } = useAuth();

The error is from useAuth(); I tried using callWithNuxt(who by the way is nowhere mentioned in the docs) and it doesn't work. If I do a entire refactoring and put the call in the setup script or method it works and I understand that I have an option to fix it but the most common use case is to do fetch from create or any other functions. I don't understand how an issue with some many complaints it's closed. Not every component can be written with the setup method in the best way and the main thing of Vue is that it let's you use both composition and options api. Here is still has an advantage over react.

Could you give us another solution to avoid this error?

Thanks.

embarq:
> Hi,
> 
> I am having the same issue but with a request from a created method. Basically I am trying to do a simple fetch outside the setup function.
> 
> async created() { await this.fetchData(); }
> 
> In the fetch method I'm trying to call nuxt auth composable to retrieve the status of the user.
> 
> const { status, data } = useAuth();
> 
> The error is from useAuth(); I tried using callWithNuxt(who by the way is nowhere mentioned in the docs) and it doesn't work. If I do a entire refactoring and put the call in the setup script or method it works and I understand that I have an option to fix it but the most common use case is to do fetch from create or any other functions. I don't understand how an issue with some many complaints it's closed. Not every component can be written with the setup method in the best way and the main thing of Vue is that it let's you use both composition and options api. Here is still has an advantage over react.
> 
> Could you give us another solution to avoid this error?
> 
> Thanks.

I ran into this issue too(it was way too dank to debug and fix btw). 

I feel like I understand better now thanks to an expanded explanation in the https://github.com/nuxt/nuxt/issues/14269#issuecomment-1397352832.

I was able to fix my stuff by following how it's done in the next-auth: https://github.com/sidebase/nuxt-auth/blob/main/src/runtime/composables/authjs/useAuth.ts#L60.

I have a custom page middleware with some async stuff going on and which had a weird behavior, but after wrapping the middleware function with `defineNuxtRouteMiddleware` as well as refactoring it, everything worked as expected!

Fix that worked for me:

[...]

PR #20918, opened 2023-05-17
feat(nuxt): experimental native async context support

pi0:
### 🔗 Linked issue

Related to #14269

### ❓ Type of change

### 📚 Description

See https://github.com/nuxt/nuxt/issues/14269#issuecomment-1397352832 and [unjs/unctx#async-context](https://github.com/unjs/unctx#async-context) for some information.

When using async composables, Nuxt app context will be only available before the first `await` statement, and for nested composables, it should be manually passed which is a negative DX especially as composables are designed to have access to the context without manually passing it!

Previously we had done some efforts in order to mitigate this DX issue:

- Vue uses transform magic in `<script setup>` to support async/await syntax
- Nuxt uses [asyncTransform](https://github.com/unjs/unctx#async-transform) for `defineNuxtPlugin`, `defineNuxtRouteMiddleware`, `defineNuxtComponent` (`setup` and `asyncData` keys), `defineNuxtPlugin` (`setup` key) and `definePageMeta` (`middleware` and `validate` keys) for **one level async support** via transforms similar to Vue (configurable/extendable via `optimization.asyncTransforms` option)
- Manual `nuxtApp.runWithContext(() => ...)` helper to make async context available (still one level only)

Developing more complex logic and nested async composables, makes it harder and harder to use previous solutions. A simple and problematic example:

```ts
const delay = () => new Promise(resolve => setTimeout(resolve, 10))

export async function nestedAsyncComposable () {
  await delay()
  return await fn1()
}

async function fn1 () {
  await delay()
  return await fn2()
}

async function fn2 () {
  await delay()
  // This fails with nuxt context unavailable error!
  const app = useNuxtApp()
  // Some logic here
}
```

This PR enables an experimental feature using [Node.js `AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage) and new unctx support to make async context available **natively** to **any nested async composable** without needing a transform or manual passing/calling with context.

Because `AsyncLocalStorage` is required for this feature and it is a Node.js only feature, we have already made a mock available via [unjs/unenv](https://github.com/unjs/unenv) for non-Node.js deployment presets (https://github.com/unjs/unenv/pull/98).

Also since this is a critical and new feature, I have made it available under an experimental flag to be enabled:

```ts
  experimental: {
    asyncContext: true
  }
})
```

TODO: 
- [x] Release unenv for (https://github.com/unjs/unenv/pull/98)
- [x] Try to find a way to prevent first `nuxtApp.runWithContext` requirement

Notes:

- This feature is limited to Node.js for now but there is an [TC39 proposal](https://github.com/tc39/proposal-async-context) ongoing to support native context in other platforms. Also Cloudflare workers recently announced [Async Context Tracking](https://blog.cloudflare.com/workers-node-js-asynclocalstorage/)! (upstream tracker: https://github.com/unjs/nitro/issues/1171)
- This is only first step in adopting native async context. Using this feature, we can extend composables to all server routes and nitro/h3 utils soon 🚀 
- (investigating yet to workaround) When using from `<script setup>`, we need to call first level composable with `nuxtApp.runWithContext` since Vue itself does not make async context available or has async context support.

### 📝 Checklist


danielroe:
@pi0 Reverted [`b1cae6f` (#20918)](https://github.com/nuxt/nuxt/pull/20918/commits/b1cae6fb8ad955390aeda40d7fac1ca90d1cde7f) just in case it was because of the test failure - feel free to reapply it.

danielroe:
@pi0 What do you mean by:

> Try to find a way to prevent first nuxtApp.runWithContext requirement

Is there something I can help with on this?

pi0:
@danielroe I have tried to explain it in the foot notes. ALS only works when we use `nuxtApp.runWithContext` once. We need to find a way to call this on the whole renderer (in nitro renderer maybe or app init) to track the async context. If you could find a working solution it is more than welcome!

danielroe (approved):

danielroe (approved):

sholohov:
Hello. Is it possible to use the experimental async context flag and not be afraid that it will be removed in some new version of Nuxt? This is very important for me.

pi0:
> Hello. Is it possible to use the experimental async context flag and not be afraid that it will be removed in some new version of Nuxt? This is very important for me.

Yes. Actually the only reason it is not universally enabled is that because some platforms don't support it yet. As soon as it is universal we move it out of the experimental state.

moshetanzer:
yip, still doesnt work in vercel! but looking forward...

Zachiah:
> > Hello. Is it possible to use the experimental async context flag and not be afraid that it will be removed in some new version of Nuxt? This is very important for me.
> 
> 
> 
> Yes. Actually the only reason it is not universally enabled is that because some platforms don't support it yet. As soon as it is universal we move it out of the experimental state.

Is there anywhere that has the list of platforms where this is and is not supported?

neo-reply-lukas:
Hello!
Would the nuxt AsyncContext also help with Pinia Cross Request Pollution when using useStore after an await in SSR? 

Last example of this Doc Page
https://pinia.vuejs.org/cookbook/composing-stores.html#Composing-Stores

PR #20955, opened 2023-05-19
fix(vite): skip generating keys for locally scoped functions

### LINKED ISSUE #20954: Should skip unique keys generation for local functions having same names with keyed composables
### Environment

Any

### Reproduction

https://stackblitz.com/edit/github-x5babp?file=app.vue

### Describe the bug

The composable-keys plugin only detects imports, we should detect local variables too. 
For example,
```js
function useState() {}
useState()
```

This may not that useful if the function is defined in user's code. But when some libraries has such functions and got bundled, there's chances those functions become local and break something. (e.g. `DatePicker` component of `@arco-design/web-vue`, )

### Additional context

I am willing to send a pr

### Logs

_No response_

ah-dc:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/20954

### ❓ Type of change

### 📚 Description

Resolves #20954

### 📝 Checklist


ah-dc:
How should I write tests for this pr?

danielroe:
I'll happily push a test if you would you be able to give me permission to push to your PR? (Normally this is an option when you create a PR - you should always enable it.)

ah-dc:
I split the tests into another file because scoped functions will be renamed to something like `useState2` after some transformation if there's already a variable with the same name.

And I am not sure how to properly handle the new property mentioned [here](https://github.com/nuxt/nuxt/issues/20954#issuecomment-1554421951) for specifying the import source of keyed composables. Currently in my implementation it behaves like:
1. when source is not specified, it is like the official keyed composables, if there is a function with that name, it won't be automatically keyed
1. when source is specified, it will be automatically keyed if it was called in the specified location or imported from specified location

packages/vite/src/utils/index.ts
```diff
@@ -43,3 +43,13 @@ export async function isDirectory (path: string) {
     return false
   }
 }
+
+export function matchWithStringOrRegex (value: string, matcher: string | RegExp) {
+  if (typeof matcher === 'string') {
+    return value === matcher
+  } else if (matcher instanceof RegExp) {
+    return !!value.match(matcher)
```
danielroe, line 51:
```suggestion
    return matcher.test(value)
```

PR #20970, opened 2023-05-20
feat(nuxt): sync `useCookie` state (including between tabs)

### LINKED ISSUE #13020: `useCookie` state is stored independently
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.17.3`
- Nuxt Version:     `3.0.0`
- Package Manager:  `yarn@3.1.1`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

```vue
<template>
  <div>
    <h1>
      {{ cookie2 }}
    </h1>
    <button @click="cookie1 = (Math.random() * 100).toFixed()">
      Change cookie
    </button>
  </div>
</template>

<script setup>
const cookie1 = useCookie('test')
const cookie2 = useCookie('test')
</script>
```

### Describe the bug

Multiple invocations of `useCookie` are not synced, and so setting one will not trigger a change to others. This is not a bug per-se, but it's inconsistent compared to `useState`, for example...

### Additional context

_No response_

### Logs

_No response_

aleksdotbar:
### 🔗 Linked issue
resolves #13020

### ❓ Type of change

### 📚 Description
Use BroadcastChannel API to sync state across `useCookie` invocations

### 📝 Checklist


xanderbarkhatov:
Maybe I should put it inside the `if (opts.watch)` condition?

xanderbarkhatov:
Updated it to use BroadcastChannel API. Now it is synced across tabs and we don't need new Nuxt events.

xanderbarkhatov:
Is `watchPaused` trick acceptable for those kind of scenarios or is there a better way? @danielroe

c-schwan:
@danielroe can you please take a look and review, thx

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -35,8 +35,24 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
...
     const callback = () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) }
+
     if (opts.watch) {
-      watch(cookie, callback, { deep: opts.watch !== 'shallow' })
+      const channel = new BroadcastChannel(`nuxt:cookies:${name}`)
+      onUnmounted(() => { channel.close() })
+      let watchPaused = false
+
+      watch(cookie, (newValue) => {
+        if (watchPaused) { return }
```
danielroe, line 45: shall we also skip if the new value is the same as old, to avoid infinite ping-pong?

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -35,8 +35,24 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
...
+        callback()
+        channel.postMessage(newValue)
```
danielroe, line 47: I think we probably want to post the message within callback rather than in the watcher. I think we still want to broadcast cookie changes even if we don't have `opts.watch` enabled.

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -35,8 +35,24 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
 
   if (process.client) {
     const callback = () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) }
+
     if (opts.watch) {
-      watch(cookie, callback, { deep: opts.watch !== 'shallow' })
+      const channel = new BroadcastChannel(`nuxt:cookies:${name}`)
+      onUnmounted(() => { channel.close() })
```
danielroe, line 41: The test failure is that this needs to run outside of component instance as well...

You could check `getCurrentInstance` to determine if we're in a component.

packages/nuxt/src/app/composables/cookie.ts
xanderbarkhatov: But the watch is not triggered if value is the same, as far as I know.

xanderbarkhatov:
```js
    channel.onmessage = (event) => {
      if (isEqual(cookie.value, event.data)) { return }
      watchPaused = true
      cookie.value = event.data
      nextTick(() => { watchPaused = false })
    }
```

Something like this?

xanderbarkhatov: Or i could put it inside the watcher

c-schwan:
@xanderbarkhatov thx well done... @danielroe can you please review?

danielroe:
@c-schwan You do not need to ping me to review.

danielroe (approved):

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -34,9 +34,28 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
   const cookie = ref<T | undefined>(cookies[name] as any ?? opts.default?.())
 
   if (process.client) {
-    const callback = () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) }
+    const channel = new BroadcastChannel(`nuxt:cookies:${name}`)
```
pi0, line 37: However, there is [small percentage](https://caniuse.com/broadcastchannel) of browsers not supporting this API, i would add a check guard  if feature is not available.

pi0 (approved):
❤️

cawa-93:
Um, and how it resolved issues when two components in same tab use `useCookie` and got unsynchronized state?

danielroe:
They are synced even within the same tab.

xanderbarkhatov:
I guess your title is a little bit confusing)

PR #20974, opened 2023-05-20
fix(nuxt): exclude plugin declaration from non-existent files

danielroe:
### ❓ Type of change

### 📚 Description

This PR fixes an issue where non-existent (i.e. virtual) plugins would cause type inference for plugin injections to fail.

### 📝 Checklist


IlyaSemenov:
I was wondering if this could be somehow backported to 3.4.x with some kind of user-land code? Type inference breaks in <3.5.1 due to `InjectionType<typeof import("../components.plugin").default>` which is fixed here.

IlyaSemenov:
OK, it was actually simple:

```ts
// Fix useNuxtApp() return type missing data injected by plugins.
//
// The problem was fixed in Nuxt 3.5.1, please remove after upgrading.
// See https://github.com/nuxt/nuxt/pull/20974#issuecomment-1592724945

import { addTemplate, defineNuxtModule } from '@nuxt/kit'

  setup(_option, nuxt) {
    addTemplate({
      getContents: () =>
        'declare const _default: any; export default _default;',
      filename: 'components.plugin.d.ts',
      dst: nuxt.options.buildDir + '/components.plugin.d.ts',
      write: true,
    })
  },
})
```

Issue #13367, opened 2022-02-14
Layers/Extends Support Tracker


pi0:
Nuxt layers are a powerful feature that you can use to share and reuse partial Nuxt applications within a monorepo, or from a git repository or npm package.

Docs:
- Usage: https://nuxt.com/docs/getting-started/layers
- Authoring: https://nuxt.com/docs/guide/going-further/layers

This issue is used to track the progress of enhancements and known issues in the roadmap.

Features:
- [x] Initial support (#3008)
- [x] `components/` dir (#3108)
- [x] `composables/` dir (https://github.com/nuxt/framework/pull/3423)
- [x] `plugins/` dir (https://github.com/nuxt/framework/pull/3462)
- [x] `server/` dir (#3673) [need refactor for nitropack]
- [x] `pages/` dir (#3783)
- [x] `middleware/` dir (#3783)
- [x] `layouts/` dir (through https://github.com/nuxt/framework/pull/3610)
- [x] `public/` dir (nitropack pending but could be workaround to pick one layer)
- [x] `app/` dir (`app/router.options`) (https://github.com/nuxt/framework/pull/3939)
- [x] app.vue (https://github.com/nuxt/framework/pull/6228)
- [x] hooks for extending dirs
- [x] Usage and authoring docs nuxt/framework#9405
- [x] Merge `hook` config key: https://github.com/nuxt/nuxt/pull/24639
- [x] install cloned dependencies (https://github.com/unjs/c12/issues/51)
- [ ] #30941
- [ ] #21635

Bugs:
- [x] Auto imports for nitro layers (#9384)

nonlinearcom:
Does anyone know a workaround to make extends work with imported modules? or is this not available at the moment?
https://github.com/nuxt/nuxt/issues/18448

szulcus:
Hi, your work is awesome! ❤️
Can I use `vite.css.preprocessorOptions.scss.additionalData` option in nuxt layer?

adam-code-labx:
> Does anyone know a workaround to make extends work with imported modules? or is this not available at the moment? #18448

I'm facing the same problem here at the moment. The only workaround is to make sure that all layer dependencies have been correctly installed in the parent core application. It's certainly frustrating, especially when we're trying to keep the theme dependencies separate from the core app ones.

Are dynamic images a factor in this issue? Currently, images have to be manually transferred to the core application, as they can not be resolved in the theme layer.

For example `~/assets/` path in our composable does not resolve to the theme layer.

[...]

PR #20976, opened 2023-05-21
docs: add link to layers from `pages/` docs

graup:
### ❓ Type of change

### 📚 Description

Not sure this is the best way, but I think it would be great if Layers would be mentioned on Pages as a way to create multiple pages directories. As someone who recently tried Next.js App Dir, I was looking for ways to achieve a similar code colocation file structure, and Layers seems to be a good way to do it. However, it took me about an hour to find out about this feature, after scouring through Github issues such as #13367 and nuxt/rfcs/issues/30 .

### 📝 Checklist

### DIFF
--- a/docs/2.guide/2.directory-structure/1.pages.md
+++ b/docs/2.guide/2.directory-structure/1.pages.md
@@ -389,4 +389,3 @@ export default defineNuxtConfig({
 ```
 
 :ReadMore{link="/docs/guide/going-further/layers"}
-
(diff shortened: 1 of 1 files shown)


docs/2.guide/2.directory-structure/1.pages.md
```diff
@@ -361,3 +361,33 @@ function navigate(){
...
+```ts
+// some-app/nuxt.config.ts
+export default defineNuxtConfig({
+})
+
+// nuxt.config.ts
+export default defineNuxtConfig({
+  extends: ['./some-app'],
+})
+
```
danielroe, line 389:
```suggestion
```

danielroe (approved):
Nice idea!

PR #20980, opened 2023-05-21
fix(nuxt): recreate `asyncData` when `immediate` is disabled

### LINKED ISSUE #20979: Refresh result of `useAsyncData` with `immediate : false` when navigation occurs
### Describe the feature

I opened a discussion #20885 a few days ago, no one answered so I decided to dig into that. I was exploring `Nuxt` and found some clues of my questions. 
Since nuxt/framework#8885, when navigation occurs on client-side, `useAsyncData` ignores cached data and makes a new call (#15445). However, I found that when `options` gets `immediate : false`, It uses cached data first. I am not sure it is intended behavior but because of nuxt/framework#8885, the result has to be the same regardless of  `immediate`'s value. Which means, when `useAsyncData` with `immediate : false` option, call `refresh` method, move to other page and come back, the result of `useAsyncData` has to be `undefined` first, not cached data because it was `undefined` on the first page loaded.
Additionally, if `useAsyncData` with `lazy: true, default : () => value`, it shows default value for the first page loaded, but after moving to other page and coming back, it shows cached data first Instead of default value. I've considered that this is also related to above issue.

### Additional information

### Final checks

jongmin4943:
### 🔗 Linked issue

#20979

### ❓ Type of change

### 📚 Description

Resolves #20979

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -214,7 +214,12 @@ export function useAsyncData<
           cbs.forEach((cb) => { cb() })
           cbs.splice(0, cbs.length)
         })
-        onUnmounted(() => cbs.splice(0, cbs.length))
+        onUnmounted(() => {
+          if (!options.immediate || options.lazy) {
```
jongmin4943, line 218: I was not sure this condition is needed because it works the same without this.

jongmin4943:
How do I write a test for this?

danielroe (dismissed):
If multiple invocations of `useAsyncData` are used (for example, in components) and one is unmounted, this would unset data for the others. It will also lead to the wrong behaviour when we're using payload extraction (for example, on a statically generated site it will remove data from the payload).

How about, instead, re-initialising the value using `options.default` when `useAsyncData` is called with `immediate: false`? (We'll still need to think about whether this is desired behaviour.)

jongmin4943:
> If multiple invocations of `useAsyncData` are used (for example, in components) and one is unmounted, this would unset data for the others. It will also lead to the wrong behaviour when we're using payload extraction (for example, on a statically generated site it will remove data from the payload).
> 
> How about, instead, re-initialising the value using `options.default` when `useAsyncData` is called with `immediate: false`? (We'll still need to think about whether this is desired behaviour.)

I also thought about using `options.default` first actually. I'll try that. Thanks for the quick reply.

jongmin4943:
@danielroe I was considering where to put the re-installation codes carefully. Could you check that this is okay? Thanks!

danielroe (commented):
I'm not sure about the approach here, as if the state is shared with another composable it will clear it out.

What about instead updating L124 to:

```ts
if (!nuxt._asyncData[key] || !options.immediate) {
```

That way it would give a 'clean slate' for each new call of the composable but not interfere with other composable calls.

Maybe it would be valuable to add a test to the suite to cover these edge cases, as well as the original issue. Would you be able to add a test as part of the PR?

jongmin4943:
I tried adding some more conditions at L124 before. If `!options.immediate` added, other components that have `useAsyncData` with the same key will not use cached data. It always makes new value. Even though they have the same key, they do not hold equal value.
>That way it would give a 'clean slate' for each new call of the composable but not interfere with other composable calls.
If I understand correctly, it means `options.immediate : false` will always create new value although the same key is provided?

BTW, I am not familiar with writing a test. Could you please give me some advice for a test if possible?

danielroe:
I think at the moment the best way to write a test for this is to do an e2e test in `test/basic.test.ts`. You will probably need to create some pages in `test/fixtures/basic/pages` and `test/fixtures/basic/components`. You'll use `createPage` to go to a page with some data, then navigate to another page. And you can make some assertions about what should happen.

I would make sure to test that data isn't _removed_ from a component that persists (e.g. in a layout) but that 'old' data doesn't persist when re-initialising `useAsyncData` in the new page. And confirm that the test fails on `main` branch but passes in this one.

jongmin4943:
@danielroe I'm sorry for the late response. I was completely lost because I had a hard time writing a test haha.. I'm not sure this is the right approach. Could you please review?

pi0 (approved):

danielroe:
Thank you ❤️

PR #21013, opened 2023-05-22
fix(nuxt): treeshake `<DevOnly>` with webpack

### LINKED ISSUE #21012: `DevOnly` is not treeshaken with webpack
### Environment

Nuxi 3.5.0                                                               
                                                                           
RootDir: J:/GIT/nuxt   

### Reproduction

https://github.com/huang-julien/nuxt/tree/dev_only_treeshake_webpack

### Describe the bug

DevOnly is not being treeshaken with webpack. I think it has the same origin as https://github.com/nuxt/nuxt/issues/20833

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix https://github.com/nuxt/nuxt/issues/21012

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/plugins/dev-only.ts
+++ b/packages/nuxt/src/core/plugins/dev-only.ts
@@ -20,7 +20,7 @@ export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions, meta
       const { type } = parseQuery(search)
 
       // vue files
-      if (pathname.endsWith('.vue') && (meta.framework === 'webpack' ? true : type === 'template' || !search)) {
+      if (pathname.endsWith('.vue') && (meta.framework === 'webpack' || type === 'template' || !search)) {
         return true
       }
     },
--- a/packages/nuxt/src/core/plugins/dev-only.ts
+++ b/packages/nuxt/src/core/plugins/dev-only.ts
@@ -9,7 +9,7 @@ interface DevOnlyPluginOptions {
   sourcemap?: boolean
 }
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/plugins/dev-only.ts
```diff
@@ -20,7 +20,7 @@ export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions) => {
       const { type } = parseQuery(search)
 
       // vue files
-      if (pathname.endsWith('.vue') && (type === 'template' || !search)) {
+      if (pathname.endsWith('.vue') && (meta.framework === 'webpack' ? true : type === 'template' || !search)) {
```
ineshbose, line 23:
```suggestion
      if (pathname.endsWith('.vue') && (meta.framework === 'webpack' || type === 'template' || !search)) {
```

huang-julien: :+1:

ineshbose: Although my preference goes to this if we need to include more (unlikely though)

```suggestion
      if (pathname.endsWith('.vue') && (['webpack'].includes(meta.framework) || type === 'template' || !search)) {
```

ineshbose (approved):

packages/nuxt/src/core/plugins/dev-only.ts
huang-julien: we only support webpack and vite for now, i think a PR for rspack is on the way, maybe should we wait for it and change the implementation ?

danielroe (approved):

Issue #15714, opened 2022-12-03
`useAsyncData` return value that `pending` is `true` at server side when `server: false`


Mini-ghost:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-qorbrt?file=app.vue

Open DevTool, `<NuxtWelcome />` not render at server side.

### Describe the bug

When we use `useAsyncData()` and set `server: false`. the return value `pending` will get `true` at server side. It should be `false`? because this will not invoke at server side.

Quineone:
> My suggestion is a new `fetching` property. It should be `true` only when the handler's promise is pending.
> 
> I've been exploring this idea in a module I'm working on, you can check it out in [nuxt-query](https://github.com/mahdiboomeri/nuxt-query).

Yes, indeed, I agree.  `pending ` is more like whether the data is fetched complete or not. act as the 
vueuse `isReady` did,   `https://vueuse.org/core/useAsyncState/` , so we also need the `isLoading` property

manniL:
We decided to introduce a `status` property for more fine-grained control as `pending` is a mix-up of a few statuses here as discussed.

Closing this discussion here 👍🏻
Thanks everyone for your input

Follow-up on #18810

PR #21045, opened 2023-05-24
feat(nuxt): return `status` from `useAsyncData`

### LINKED ISSUE #18810: `status` property for `useAsyncData`/`useFetch` (+ lazy equivalents)
### Describe the feature

Follow-up of https://github.com/nuxt/nuxt/issues/15714

This feature request should provide a more fine-grained check of the status of `useFetch`/`useAsyncData` (and the lazy equivalents) by introducing a `status` ref that contains the status a a string.

## `Status` values

* `IDLE` -> Request did not start yet, e.g. because:
	* `{ immediate: false }` and `execute` was not called ye.
	* `{ server: false }` and status is logged during SSR
* `PENDING` -> Requesting data did start
* `SUCCESS` -> Requesting data succeeded
* `ERROR` -> Requesting data failed.

**Any other kinds of `status`es needed?**

### Additional information

### Final checks

nicolaspayot:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/18810

### ❓ Type of change

### 📚 Description

- This PR introduces a new `status` property in the object returned by the composables `useAsyncData` / `useFetch` and their `lazy` variations.
- This `status` property is a  `ref<string>` that can contain the following values:
    - `idle` -> when request did not start
        - (`{ immediate: false }` and `execute` not called yet
        - `{ server: false }` -> `status` should be `idle` on server side
    - `pending` -> when started
    - `success` -> when request succeeded
    - `error` -> when request failed
- It aims to bring a more powerful way to track the status of an async data request. Relying on `pending` flag only has a limitation detailed in [this issue](https://github.com/nuxt/nuxt/issues/15714).
- Resolves https://github.com/nuxt/nuxt/issues/18810

### 📝 Checklist


docs/3.api/1.composables/use-async-data.md
```diff
@@ -34,11 +34,14 @@ type AsyncData<DataT, ErrorT> = {
...
   execute: (opts?: AsyncDataExecuteOptions) => Promise<void>
   error: Ref<ErrorT | null>
+  status: Ref<AsyncDataRequestStatus>
 };
 
 interface AsyncDataExecuteOptions {
   dedupe?: boolean
 }
+
+type AsyncDataRequestStatus = 'IDLE' | 'PENDING' | 'SUCCESS' | 'ERROR'
```
danielroe, line 44: would you mind making these lowercase?

danielroe (commented):
This is beautifully done - thank you ❤️

docs/3.api/1.composables/use-async-data.md
danielroe: Just for consistency - our other string literal types are lowercase.

nicolaspayot: Makes sense, thanks.

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -222,6 +229,7 @@ export function useAsyncData<
     if (fetchOnServer && nuxt.isHydrating && hasCachedData()) {
       // 1. Hydration (server: true): no fetch
       asyncData.pending.value = false
+      asyncData.status.value = asyncData.error.value ? 'error' : 'success'
```
danielroe, line 232: would it make sense instead to initialise it to this value above on L132?

danielroe: Just to explain, I wasn't thinking we should reset to idle but rather could directly initialise it to 'error' or 'success' based on the value in the payload _rather_ than initialising it to 'idle'.

But on reflection, I think this is good too 👍

danielroe (approved):

packages/nuxt/src/app/composables/asyncData.ts
nicolaspayot: I misunderstood, gotcha! 👍

PR #21064, opened 2023-05-25
fix(nuxt): display errors in client overlay with `typeCheck`

### LINKED ISSUE #20851: `vite-plugin-checker` not currently working when `typeCheck: true`
issue documented and workaround provided at https://vite-plugin-checker.netlify.app/faq/integration.html#overlay-does-not-display-in-development-mode-1

see https://github.com/nuxt/nuxt/issues/14336#issuecomment-1547588854

aleksdotbar:
### 🔗 Linked issue
resolves #20851

### ❓ Type of change

### 📝 Checklist


xanderbarkhatov:
What do you think? @danielroe

packages/nuxt/src/core/nuxt.ts
```diff
@@ -308,6 +308,22 @@ async function initNuxt (nuxt: Nuxt) {
     addPlugin(resolve(nuxt.options.appDir, 'plugins/preload.server'))
   }
 
+  // Add type-checking
+  if (process.dev && nuxt.options.builder === '@nuxt/vite-builder' && nuxt.options.typescript.typeCheck === true) {
```
danielroe, line 312: I think this section should ideally be in the vite builder package but the plugin approach seems reasonable.

xanderbarkhatov: Yeah, I tried to put it right after `addVitePlugin` part, but couldn't get it to work.
`addTemplate` does nothing at that point.

danielroe: Yes, that does make sense. 👍 I wonder if instead we could add the import to the entry in the vite client config?

xanderbarkhatov: Updated the code. Hope I understood correctly.

danielroe: Thanks! Yes, perfect 👍

danielroe (approved):

PR #21081, opened 2023-05-26
fix(nuxi,nuxt): remove baseUrl + use relative paths in tsconfig

### LINKED ISSUE #21070: Type shadowing of global packages due to baseUrl
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.1
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-gjecyr?file=app.vue,graphql%2Findex.ts

Run `npm run dev`, you will see graphql version is displayed by `app.vue`, meaning code imports from the global `graphql` package.

Run `nuxi typecheck`, you will see that it doesn't type check, as types import from the local `graphql` directory:

```
❯ nuxi typecheck
Nuxi 3.5.1
app.vue:2:8 - error TS1192: Module '"/home/projects/nuxt-starter-gjecyr/graphql/index"' has no default export.

2 import graphql from 'graphql';
         ~~~~~~~


Found 1 error in app.vue:2
```

### Describe the bug

Adding a directory under project root shadows the global package with the same name **when type checking** (not in runtime).

In the reproduction, I added `graphql/index.ts` under Nuxt project root. That breaks `import ... from "graphql"` for type checking. What is worse, type imports break inside _other_ libraries such as `@urql/vue` or `villus` which is hard to diagnose.

### Additional context

This is caused by `.nuxt/tsconfig.json` having `baseUrl` pointing to project root. I believe that serves no real purpose and is a time bomb for types shadowing. As I gathered, it's only needed for setting up non-relative path aliases such as `"#imports": [".nuxt/imports"]` which is simple to avoid (the majority of the generated aliases are already relative).

I tried removing `baseUrl` and converting remaining non-relative path aliases to relative (`"@/*": ["../*"]`, `"#imports": ["./imports"]`, etc.) and everything seemed to work fine.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/21070

### ❓ Type of change

### 📚 Description

This fixes an issue where we were resolving paths in TS relative to `srcDir`. This might also have caused a few other issues that were reported.

### 📝 Checklist


packages/nuxi/src/utils/prepare.ts
```diff
@@ -45,15 +44,17 @@ export const writeTypes = async (nuxt: Nuxt) => {
   // Exclude bridge alias types to support Volar
   const excludedAlias = [/^@vue\/.*$/]
 
+  const basePath = tsConfig.compilerOptions!.baseUrl || nuxt.options.buildDir
```
danielroe, line 47:
```suggestion
  const basePath = tsConfig.compilerOptions!.baseUrl ? resolve(nuxt.options.buildDir, tsConfig.compilerOptions!.baseUrl) : nuxt.options.buildDir
```

PR #21246, opened 2023-05-30
feat(kit): module compatibility utils

harlan-zw:
### ❓ Type of change

### Background

As more modules are being developed, there is more opportunity for modules to integrate together, providing a better DX.

There are many areas that could be improved here. We start by allowing modules to answer the following:
- Is module "foo" being used?
- Does module "foo"'s version match my semver requirements?

### Implementation

To provide semver checks, we need to know what version of a module is being used. However, there is a core issue in that build-time metadata (version) isn't exposed within the module. 

To fix this, we merge in `module.json` data into the module's metadata (if it exists). This provides flexibility for exposing other build-time metadata that may be introduced in the future.

Other than that the changes are fairly straightforward and non-breaking.

### Limitations

- `hasNuxtModule` won't work if called before a module is added programmatically

### 📝 Checklist


packages/kit/src/module/install.ts
```diff
@@ -48,9 +48,8 @@ export const normalizeModuleTranspilePath = (p: string) => {
   return p.split('node_modules/').pop() as string
 }
 
-async function normalizeModule (nuxtModule: string | NuxtModule, inlineOptions?: any) {
-  const nuxt = useNuxt()
-
+export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, nuxt: Nuxt = useNuxt()) {
```
harlan-zw, line 51: I renamed this and changed the signature:
- the name didn't really describe what was happening
- the inlineOptions was not accessed at all
- nuxt instance was recreated

packages/kit/src/module/install.ts
```diff
@@ -1,16 +1,16 @@
...
+import { defu } from 'defu'
 import { isNuxt2 } from '../compatibility'
 import { useNuxt } from '../context'
 import { requireModule } from '../internal/cjs'
 import { importModule } from '../internal/esm'
 import { resolveAlias, resolvePath } from '../resolve'
 
 /** Installs a module on a Nuxt instance. */
-export async function installModule (moduleToInstall: string | NuxtModule, _inlineOptions?: any, _nuxt?: Nuxt) {
-  const nuxt = useNuxt()
```
harlan-zw, line 12: Previously this `_nuxt` is unused

packages/kit/src/compatibility.ts
```diff
@@ -2,6 +2,10 @@ import satisfies from 'semver/functions/satisfies.js' // npm/node-semver#381
 import type { Nuxt, NuxtCompatibility, NuxtCompatibilityIssues } from '@nuxt/schema'
 import { useNuxt } from './context'
 
+export function normaliseSemanticVersion (version: string) {
```
danielroe, line 5: It hurts me deeply but we should probably use American English here (and elsewhere where this function is used):

```suggestion
export function normalizeSemanticVersion (version: string) {
```

packages/kit/src/module/compatibility.ts
```diff
@@ -0,0 +1,53 @@
+import satisfies from 'semver/functions/satisfies.js' // npm/node-semver#381
+import type { Nuxt, NuxtModule } from '@nuxt/schema'
+import { useNuxt } from '../context'
+import { normaliseSemanticVersion } from '../compatibility'
+import { loadNuxtModuleInstance } from './install'
+
+/**
+ * CHeck if a Nuxt module is installed by name.
+ *
+ * This will check both the installed modules and the modules to be installed.
```
danielroe, line 10:
```suggestion
 * This will check both the installed modules and the modules to be installed. Note
 * that it cannot detect if a module is _going to be_ installed programmatically by another module.
```

packages/kit/src/module/install.ts
```diff
@@ -29,7 +29,7 @@ export async function installModule (moduleToInstall: string | NuxtModule, _inli
 
   nuxt.options._installedModules = nuxt.options._installedModules || []
   nuxt.options._installedModules.push({
-    meta: await nuxtModule.getMeta?.(),
+    meta: defu((await nuxtModule.getMeta?.()) || {}, buildTimeModuleMeta || {}),
```
danielroe, line 32: `defu` should handle null/undefined/falsy values:

```suggestion
    meta: defu(await nuxtModule.getMeta?.(), buildTimeModuleMeta),
```

danielroe (commented):
I _love_ this. Nice work. 👌

danielroe (approved):

PR #21263, opened 2023-05-30
fix(nuxt): key distinct pages differently for legacy asyncData

### LINKED ISSUE #20686: Parent page asyncData lost when child page has asyncData
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.3.3
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-a1n4jf

Open the page. The page will first show:

```
parent_data=parent
child_data=child
```

Then in a moment, it will be replaced with:

```
parent_data=
child_data=child
```

### Describe the bug

When `[id].vue` and `[id]/index.vue` both have `asyncData`, both run and the result of the latter leaks into the parent component. The parent component breaks.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20686

### ❓ Type of change

### 📚 Description

When working with legacy `asyncData`, we key asyncData in relation to the route path. But this results in conflicting keys between any components using this legacy asyncData. This PR instead generates a key whenever `defineNuxtComponent` is used, falling back to the previous behaviour.

We also need to document the behaviour of `fetchKey` for customising the key generated.

### 📝 Checklist


atinux (dismissed):
Good for me, I have no idea where we could document the legacy fetchKey though

PR #21285, opened 2023-05-31
docs: add hint for useAsyncData() in local environment

jannescb:
When working with APIs on `localhost` it seems that you have to use real IPs in order to work with `useAsyncData()` in `SSR` mode.
A small hint in the docs might help a lot as it's hard to find any information on this.

### DIFF
--- a/docs/3.api/1.composables/use-async-data.md
+++ b/docs/3.api/1.composables/use-async-data.md
@@ -9,6 +9,10 @@ Within your pages, components, and plugins you can use useAsyncData to get acces
 `useAsyncData` is a composable meant to be called directly in a setup function, plugin, or route middleware. It returns reactive composables and handles adding responses to the Nuxt payload so they can be passed from server to client without re-fetching the data on client side when the page hydrates.
 ::
 
+::alert{type=warning}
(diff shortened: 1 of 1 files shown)


oritwoen:
Thanks for updating the documentation @jannescb . However, for me, even using the IP does not help because it still treats it as part of the nuxt local instance and a internal loop is created.

I had to change /etc/hosts and create `IP -> domain` mask to test local CMS (Strapi or some own system).

It's very problematic. Not sure if it shouldn't be an issue @danielroe ?

danielroe:
Yes, this sounds like a bug; this hint shouldn't be necessary. I would appreciate an issue/reproduction if you can. 🙏

danielroe:
I'm closing this for now as I think this shouldn't be required, but please do create an issue with a reproduction if you can 🙏

Alternatively I think there are some factors developers on Windows might need to know and we could possibly hint about them. (cc: @pi0)

oritwoen:
@danielroe Actually, I haven't noticed this problem lately.

I use Linux every day and the problem probably solved itself when I updated Nuxt or packages and the `/etc/hosts` file.

I can't say now what caused the problem (at least in my case), but I suspect it could be due to a wrong system configuration somewhere.

PR #21333, opened 2023-06-02
feat(nuxt): add `open` option in `navigateTo` helper

### LINKED ISSUE #20625: Provide target option to navigateTo
### Describe the feature

We have a UI component library that should be completely decoupled from Nuxt. Therefore, we want to inject a callback function that is called if a anchor tag is clicked and which calls `navigateTo`. 

But `navigateTo` provides no possibility to add target options like `_blank` to open the to path in a new tab.

### Additional information

### Final checks

wilsonpinto:
### 🔗 Linked issue

#20625 

### ❓ Type of change

### 📚 Description

1. Extend the _NavigateToOptions_ interface to include a new optional object parameter, `open`. 
This parameter, when set, will instruct the navigateTo function to open the URL using window.open method.

2. Modify the _navigateTo_ function to handle the `open` option. This option is only applicable on the client side and will be ignored on the server side.

Resolves #20625 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -99,10 +99,10 @@ export type OpenOptions = {
 }
 
 export interface NavigateToOptions {
-  replace?: boolean;
-  redirectCode?: number;
-  external?: boolean;
-  open?: OpenOptions;
+  replace?: boolean
+  redirectCode?: number,
+  external?: boolean,
+  open?: OpenOptions
 }
 
 export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: NavigateToOptions): Promise<void | NavigationFailure | false> | false | void | RouteLocationRaw => {
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -80,10 +80,29 @@ const isProcessingMiddleware = () => {
   return false
 }
 
+export interface WindowOpenFeatures {
+  readonly width?: number;
+  readonly innerWidth?: number;
+  readonly height?: number;
+  readonly innerHeight?: number;
+  readonly left?: number;
+  readonly screenX?: number;
+  readonly top?: number;
+  readonly screenY?: number;
+  readonly noopener?: boolean;
+  readonly noreferrer?: boolean;
+}
+
+export interface WindowOpenOptions {
+  readonly target: '_self' | '_blank' | '_parent' | '_top';
+  readonly windowFeatures?: WindowOpenFeatures;
+}
+
 export interface NavigateToOptions {
-  replace?: boolean
-  redirectCode?: number,
-  external?: boolean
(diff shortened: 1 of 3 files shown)


wilsonpinto:
~> linking here: https://github.com/nuxt/nuxt/issues/20625#issuecomment-1574792626

danielroe (commented):
Thank you so much for opening this!

The feature, I think, is a good one to have. I think it would be good to align with `<NuxtLink>` options - so support a `target` option instead of a boolean flag. It might also be good to support window features as another option (maybe object-syntax?).

danielroe:
Yes, I agree that would be nice & intuitive 👍

packages/nuxt/src/app/composables/router.ts
```diff
@@ -92,6 +111,26 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
...
+  // Early open handler
+  if (options?.open) {
+    if (process.client) {
+      const { target = '_blank', windowFeatures = {} } = options.open
+
+      const features = Object.entries(windowFeatures)
+        .map(([feature, value]) => {
+          const formattedValue =
+              typeof value === 'boolean' ? (value ? 'true' : 'false') : value
+          return `${feature.toLowerCase()}=${formattedValue}`
```
danielroe, line 124: I think implicit conversion to string should work to convert boolean -> string.

```suggestion
          return `${feature.toLowerCase()}=${value}`
```

wilsonpinto: Sorry for overwrite your changes, forgot to put this on draft.

Checking this now

wilsonpinto:
Okay, ready to review now :)

packages/nuxt/src/app/composables/router.ts
```diff
@@ -92,6 +111,23 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
...
+
+      const features = Object.entries(windowFeatures)
+        .filter(([_, value]) => value !== undefined)
+        .map(([feature, value]) => `${feature.toLowerCase()}=${value}`)
+        .join(', ')
+
+      open(toPath, target, features)
+    }
+
+    return Promise.resolve()
```
danielroe, line 128: Currently this will do nothing at all on server but simply return. Maybe worth adding a warning in development mode in this case?

danielroe (commented):
Looking great!

packages/nuxt/src/app/composables/router.ts
wilsonpinto: I understand the rationale behind it, but it would be ideal if this behavior is automatically ignored on the server side, particularly if it resembles a window.open replication. 

If we use this in the middlewares or setup this will trigger few warnings

TheAlexLichter (approved):

danielroe (approved):
Nice work ❤️

Issue #21315, opened 2023-06-02
Access result of route handler on server side middleware


MSchutt:
### Describe the feature

From my stackoverflow question: https://stackoverflow.com/questions/76323883/nuxt-3-after-response-middleware

I have a use-case where I would like to call a middleware after the response went through the route handler. The docs describe that the standard server middleware only runs BEFORE the request is handled (https://nuxt.com/docs/guide/directory-structure/server).

What I'd like to accomplish is:

```
// file: server/api/test/index.ts
export default defineEventHandler(async (event) => {
    return { "test": true }
})
```
When I call the endpoint via GET /api/test I would like the response to be:

`{ "result": { "test": true } }`

So basically mapping all APIs response in a object with key "result". This is quite easy to do with express middleware and other frameworks as you can usually await the route handler's result and then just wrap the result in the object.

It seems that currently there is no way to access the route handler's response inside of a server side middleware. Within most frameworks (i.e. express) you can usually await the response within the middleware function.

Suggested syntax which would be desirable:
```
// file: server/middleware/test.ts
export default defineEventHandler(async (event) => {
  // Before
  console.log('New request: ' + getRequestURL(event))
  const result = await event.requestHandler // any function/property that can be await that contains the result
  // After (wrap in result object)
  return { result };
})
```

### Additional information

- [ ] Would you be willing to help implement this feature?
- [X] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
Would you raise this as a feature request on h3 instead?

https://github.com/unjs/h3

PR #21337, opened 2023-06-03
docs: include example for server utilities

MSchutt:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/21315

### ❓ Type of change

### 📚 Description

I have a use case where I need to encapsulate all responses from each endpoint within an object. This can be achieved by utilizing a server utility. However, it appears that the documentation currently lacks proper explanation of this feature, and it would be beneficial to include a concise example for clarification. (Credits to [@maou-shonen](https://github.com/maou-shonen) for bringing this possibility to my attention, as discussed in this GitHub issue: https://github.com/unjs/h3/issues/397)

### 📝 Checklist

### DIFF
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -102,6 +102,29 @@ Server routes are powered by [unjs/h3](https://github.com/unjs/h3) which comes w
 
 You can add more helpers yourself inside the `~/server/utils` directory.
 
+
+For example, you can define a custom handler utility that wraps the original handler and performs additional operations before returning the final response.
+
+**Example:**
+
+```ts [server/utils/handler.ts]
+import type { H3Event } from 'h3'
+
+export const defineMyHandler = (handler: (event: H3Event) => any) => {
+  const _handler = async (event: H3Event) => {
+    const response = await handler(event)
+
+    // mutate the response for all routes
+    const responseNew = doSome(response)
+
+    return responseNew
+  }
+
+  return defineEventHandler(_handler)
+}
+```
+
+
 ## Server Types
 
 ::alert{type="info"}
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -102,7 +102,6 @@ Server routes are powered by [unjs/h3](https://github.com/unjs/h3) which comes w
 
 You can add more helpers yourself inside the `~/server/utils` directory.
 
-
 For example, you can define a custom handler utility that wraps the original handler and performs additional operations before returning the final response.
 
 **Example:**
@@ -123,7 +122,6 @@ export const defineMyHandler = (handler: (event: H3Event) => any) => {
   return defineEventHandler(_handler)
 }
 ```
-
 
 ## Server Types
 
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -107,20 +107,20 @@ For example, you can define a custom handler utility that wraps the original han
 **Example:**
 
 ```ts [server/utils/handler.ts]
-import type { H3Event } from 'h3'
+import type { EventHandler } from 'h3'
 
-export const defineMyHandler = (handler: (event: H3Event) => any) => {
-  const _handler = async (event: H3Event) => {
-    const response = await handler(event)
-
-    // mutate the response for all routes
-    const responseNew = doSome(response)
-
-    return responseNew
-  }
-
-  return defineEventHandler(_handler)
-}
+export const defineWrappedResponseHandler = (handler: EventHandler) =>
+  defineEventHandler(async (event) => {
+    try {
+      // do something before the route handler
+      const response = await handler(event)
+      // do something after the route handler
+      return { response }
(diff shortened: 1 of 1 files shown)


danielroe (commented):
I think this is a useful thing to do in your case. Having said that, this is a very specific kind of utility that is probably not relevant for most people - and also potentially quite complex. For example, what does `doSome` look like? How does it handle errors? or null responses?

What do you think about providing a simpler example, such as this kind of thing:

```ts
import { H3Event } from 'h3'

export function getUserSession(event: H3Event) {
  return useSession(event, { password: config.sessionPassword })
}

export function setUserSession(event: H3Event, data: any) {
  return updateSession(event, { password: config.sessionPassword }, data)
}
```

MSchutt:
I would like to draw attention to a valuable and practical aspect: the ability to access the result both before and after the route handler by utilizing server utilities. This feature is commonly found in middleware frameworks like Express or Fastify, but it has not been explicitly documented in Nuxt yet. However, I completely understand and get any concerns you may have. What do you think about this code snippet that gracefully handles errors, encapsulating both the result and the error within objects. I would appreciate your thoughts on it, but maybe even such a use-case might be too specific as server utils are quite powerful. 
Thank you!

```ts [server/utils/handler.ts]
import type { H3Event } from 'h3'
export const defineMyHandler = (handler: (event: H3Event) => any) => {
  const _handler = async (event: H3Event) => {
    try {
      // do something before the route handler
      const response = await handler(event)
      // do something after the route handler
      return { response }
    } catch (err) {
      // Error handling
      return { err }
    }
  }
  return defineEventHandler(_handler)
}
```

pi0:
For having init/once logic for an event handler defenition, we can use `lazyEventHandler` wrapper. It supports both async init and types without need to manually import:


```ts
export default lazyEventHandler(() => {
  // Initial preparation logic
  // Can be async too
  return defineEventHandler((event) => {

  })
})
```

danielroe:
How about this?

```ts
import type { EventHandler } from 'h3'

export const defineWrappedResponseHandler = (handler: EventHandler) =>
  defineEventHandler(async (event) => {
    try {
      // do something before the route handler
      const response = await handler(event)
      // do something after the route handler
      return { response }
    } catch (err) {
      // Error handling
      return { err }
    }
  })
```

pi0 (approved):
👍🏼 

Worth to note that event handlers with object syntax are coming soon to h3 (with `pre`, `post`, `error` hooks). Once ready we shall update the docs to advice using object composable API over custom wrapper

OrlS15:
> How about this?
> 
> ```ts
> import type { EventHandler } from 'h3'
> 
> export const defineWrappedResponseHandler = (handler: EventHandler) =>
>   defineEventHandler(async (event) => {
>     try {
>       // do something before the route handler
>       const response = await handler(event)
>       // do something after the route handler
>       return { response }
>     } catch (err) {
>       // Error handling
>       return { err }
>     }
>   })
> ```

This isn't typed properly.
Fetching an endpoint with $fetch or useFetch returns wrong type
```ts
let a = $fetch("/api/wrappedEndpoint")
//  let a: Simplify<SerializeObject<{
//     response: any;
//     err?: undefined;
//} > | SerializeObject<{
//     err: unknown;
//     response?: undefined;
// }>>
```
So the editor doesn't give you suggestions when you write a.something

Maybe something like this fixes it?
```ts
export function defineWrappedResponseHandler<T>(handler: EventHandler<T>) : EventHandler<T>
```

sutoh:
I had the same problem, but I found a solution with the following code. I hope this helpful.

```typescript
import { type EventHandler, type H3Event, H3Error } from 'h3'
import { Prisma } from '@prisma/client'

export interface IErrorResponse {
  statusCode: number
  statusMessage?: string
  cause?: any
}

const errorResponse = (
  event: H3Event,
  response: IErrorResponse
): IErrorResponse => {
  event.node.res.statusCode = response.statusCode
  event.node.res.statusMessage = response.statusMessage as string
  return {
    statusCode: response.statusCode,
    statusMessage: response.statusMessage,
    cause: response.cause ? response.cause : null,
  }
}

export const defineWrappedResponseHandler = <T>(
  handler: EventHandler<T>
): EventHandler<T> =>
  defineEventHandler(async (event) => {
    try {
      // do something before the route handler
      const response = await handler(event)
      // do something after the route handler
      return response
    } catch (err) {
      console.error(err)

      // Error handling
      // catch the createError()
      if (err instanceof H3Error) {
        throw errorResponse(event, {
          statusCode: err.statusCode,
          statusMessage: err.statusMessage,
          cause: err.message,
        })
      }

      // Handling Prisma errors
      if (err instanceof Prisma.PrismaClientKnownRequestError) {
        throw errorResponse(event, {
          statusCode: 400,
          statusMessage: 'Bad Request',
          cause: 'Database access error',
        })
      }
      if (err instanceof Prisma.PrismaClientValidationError) {
        throw errorResponse(event, {
          statusCode: 400,
          statusMessage: 'Bad Request',
          cause: 'Database validation error',
        })
      }

      // Handling other errors
      throw errorResponse(event, {
        statusCode: 500,
        statusMessage: 'Internal Server Error',
        cause: 'Unknown error. Please check the server log.',
      })
    }
  })

```

PR #21344, opened 2023-06-04
Fix hmr server port

### LINKED ISSUE #21341: HMR in Gitpod + VS Code Browser
Hi,

I'm having a hard time fixing HMR while developing on Gitpod using VS Code Browser.
Apparently, it's working on Vite normally, but it doesn't with Nuxt.
I tried everything I found online, but it seems I'm stuck in a chicken-and-egg problem.

Please refer to this URL for the context:
https://www.gitpod.io/docs/introduction/languages/javascript#vite

Any ideas on what can be the problem and how to make it work in Nuxt?
I can let you know the specific behavior of specific settings if it helps.

Thanks!

ghostdevv:
~~This fixes #21341 by removing some code that set vite's hmr port to `24678` everytime. In Vite 3 [this pr](https://github.com/vitejs/vite/pull/8650) set the default behaviour to use the `clientPort` (defaults to 3000 in nuxt) for HMR rather than a `24678`. Removing the code in nuxt's vite builder that set this port makes vite's default behaviour kick in causing #21341 to be fixed~~


ghostdevv:
Trying to workout why I sometimes see this error
image

ghostdevv:
HMR in the playground also seems to not work at all, even when I try locally on main :thinking:

PR #21380, opened 2023-06-05
docs: remove irregular spaces

antfu:
for some reason I found we have many 0xA0 spaces instead of 0x20. I guess they are not intentional.


danielroe (approved):

danielroe:
I would love to know how you happened to find this!

PR #21409, opened 2023-06-06
feat(nuxt): provide `clearNuxtState` composable

### LINKED ISSUE #21293: Provide `clearNuxtState` function for `useState`
### Describe the feature

The `useState` uses `useNuxtApp().payload.state` to preserve the value after server-side rendering and share the value across all components through a unique key. If `useState` is previously used with a key and if the key is the same, it will use the cached value regardless of whether it has a different `init` method or not. It is useful to use cached data when needed. However, in some cases, the value of `useState` needs to be redeclared with a new `init` method. It seems like the only way is creating unique key every time.

It would be nice to have just like `clearNuxtData` function to clear cached state value, so we don't have to think how to make a unique key when we want to change the value of `useState`

### Additional information

### Final checks

jongmin4943:
### 🔗 Linked issue
#21293

### ❓ Type of change

### 📚 Description
Resolves #21293

### 📝 Checklist


packages/nuxt/src/app/composables/state.ts
```diff
@@ -35,3 +36,23 @@ export function useState <T> (...args: any): Ref<T> {
...
+      ? _allKeys.filter(keys)
+      : Array.isArray(keys)
+        ? keys.map(k => useStateKeyPrefix + k)
+        : [useStateKeyPrefix + keys]
+
+  for (const key of _keys) {
+    if (key in nuxtApp.payload.state) {
+      nuxtApp.payload.state[key] = undefined
+    }
+  }
```
danielroe, line 57: slightly more performant not to do another full iteration

```suggestion
  const _keys: string[] = !keys
    ? _allKeys
    : typeof keys === 'function'
      ? _allKeys.filter(keys)
      : Array.isArray(keys) ? keys : [keys]

  for (const key of _keys) {
    if (key in nuxtApp.payload.state) {
      nuxtApp.payload.state[useStateKeyPrefix + key] = undefined
    }
  }
```

danielroe (approved):

jongmin4943:
Thanks for the review! Should I write `docs` about this in this PR?

danielroe:
Oh, yes please!

packages/nuxt/src/app/composables/state.ts
```diff
@@ -35,3 +36,21 @@ export function useState <T> (...args: any): Ref<T> {
...
+  const nuxtApp = useNuxtApp()
+  const _allKeys = Object.keys(nuxtApp.payload.state)
+  const _keys: string[] = !keys
+    ? _allKeys
+    : typeof keys === 'function'
+      ? _allKeys.filter(keys)
+      : Array.isArray(keys) ? keys : [keys]
+
+  for (const key of _keys) {
+    if (key in nuxtApp.payload.state) {
```
jongmin4943, line 52: keys in `_keys` are not started with `useStateKeyPrefix("$s")`. Should the if statement be `("$s" + key in nuxtApp.payload.state)`?

danielroe: quite right!

packages/nuxt/src/app/composables/state.ts
```diff
@@ -35,3 +36,21 @@ export function useState <T> (...args: any): Ref<T> {
...
+      nuxtApp.payload.state[useStateKeyPrefix + key] = undefined
```
danielroe, line 53:
```suggestion
      nuxtApp.payload.state[key] = undefined
```

docs/3.api/3.utils/clear-nuxt-data.md
```diff
@@ -12,4 +12,4 @@ clearNuxtData (keys?: string | string[] | ((key: string) => boolean)): void
 
 ## Parameters
 
-* `keys`: On or an array of keys that are used in `useAsyncData` to delete their cached data. If no keys are provided, **every data** will be invalidated.
+* `keys`: One or an array of keys that are used in `useAsyncData` to delete their cached data. If no keys are provided, **every data** will be invalidated.
```
danielroe, line 15:
```suggestion
* `keys`: One or an array of keys that are used in `useAsyncData` to delete their cached data. If no keys are provided, **all data** will be invalidated.
```

docs/3.api/3.utils/clear-nuxt-state.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Type
+
+```ts
+clearNuxtState (keys?: string | string[] | ((key: string) => boolean)): void
+```
+
+## Parameters
+
+* `keys`: One or an array of keys that are used in `useState` to delete their cached state. If no keys are provided, **every state** will be invalidated.
```
danielroe, line 15:
```suggestion
* `keys`: One or an array of keys that are used in `useState` to delete their cached state. If no keys are provided, **all state** will be invalidated.
```

PR #21439, opened 2023-06-07
fix(nuxt): avoid early returning renderer with empty response

### LINKED ISSUE #21414: Caching is broken
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      routeRules
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-zfr2rj?file=nuxt.config.ts

Observe how the timestamp from app.vue never gets updated when reloading the page, even after 10 seconds.

### Describe the bug

With the following nuxt config:

```
{
  routeRules: {
    '**': {
      cache: {
        maxAge: 10,
      },
    },
  },
}
```

I would expect SSR rendered content to be cached for 10 seconds.
While the max-age header reflects that correctly, the same stale HTML gets served indefinitely.

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #21414

### ❓ Type of change

### 📚 Description

We are short-circuiting nitro renderer when headers or response is (presumably) already sent which should not be the case because this way nitro cannot handle caching properly.

Currently [nitro v2 allows this](https://github.com/unjs/nitro/blob/2ad43894a6039207fd535d7fdc30ba92d4847ee6/src/runtime/renderer.ts#L27) since Nuxt was depending on this behavior and we cannot remove it until Nitro v3 as a breaking change. This PR makes sure Nuxt is prepared for this.

**Note:** Effect of this PR is fixing the root cause of similar issues and this one but there is still an unknown issue with the nuxt and cache wrapper that sets `headersSent` and `writableEnded` even when the response is not sent and it needs to be investigated separately. (Another workaround only for this, https://github.com/unjs/nitro/pull/1289 <-- still _hides_ real issue)

### 📝 Checklist


danielroe:
Does it make sense to align merging this with https://github.com/unjs/nitro/pull/1289?

danielroe (approved):

PR #21446, opened 2023-06-07
feat(webpack, schema): add webpack experiments configuration

### LINKED ISSUE #20647: ability to setup webpack config.experiments in nuxt config
### Feature

right now it is not possible to configure experiments configuration with webpack builder.  for example I would need to setup: 

```js
  webpack: {
    experiments: {
      asyncWebAssembly: true,
    }
  },
```

there are many use cases for this particularly considering the growing number of applications using wasm modules.  

### Additional information

### Final checks

wilsonpinto:
### 🔗 Linked issue

#20647 

### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Add the ability to set up webpack experiments configuration in nuxt.config

```ts
 webpack: {
    experiments: {
        asyncWebAssembly: true,
        buildHttp: true,
        layers: true,
        lazyCompilation: true,
        outputModule: true,
        syncWebAssembly: true,
        topLevelAwait: true,
    }
  }
 ```

Resolves #20647 

### 📝 Checklist

### DIFF
--- a/packages/webpack/src/presets/base.ts
+++ b/packages/webpack/src/presets/base.ts
@@ -39,7 +39,9 @@ function baseConfig (ctx: WebpackConfigContext) {
       ...options.webpack.optimization,
       minimizer: []
     },
-    experiments: {},
+    experiments: {
+      ...options.webpack.experiments
+    },
     mode: ctx.isDev ? 'development' : 'production',
     cache: getCache(ctx),
     output: getOutput(ctx),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/webpack/src/presets/base.ts [hunk 1]
```diff
@@ -39,7 +39,9 @@ function baseConfig (ctx: WebpackConfigContext) {
       ...options.webpack.optimization,
       minimizer: []
     },
-    experiments: {},
+    experiments: {
+      ...options.webpack.experiments
```
danielroe, line 43: I think we have to handle the case where `false` is passed - I think this will throw an error in that case.

wilsonpinto: tested it :D weird but 

```
{
...false
}
```

==> {}

danielroe: You are quite right - my mistake.

danielroe (approved):

PR #21450, opened 2023-06-07
fix(nuxt): fix error on layout switching

### LINKED ISSUE #13309: TypeError: Cannot read properties of null (reading 'parentNode')
### Environment

- Operating System: `Linux`
- Node Version:     `v14.18.1`
- Nuxt Version:     `3.0.0-27389502.88da719`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Vite`
- User Config:      `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

use this [sandbox](https://codesandbox.io/s/tender-kare-e87h0), Visit `${url}/organization/365824000000000003/template/new` click on the "a" link then the "b" link and this error is thrown:

```
runtime-core.esm-bundler.js:38 [Vue warn]: Unhandled error during execution of scheduler flush. This is likely a Vue internals bug. Please open an issue at https://new-issue.vuejs.org/?repo=vuejs/core 
  at <Transition name="page" mode="out-in" > 
  at <Dashboard name="dashboard" > 
  at <AsyncComponentWrapper name="dashboard" > 
  at <NuxtLayout name="dashboard" > 
  at <BaseTransition mode="out-in" appear=false persisted=false  ... > 
  at <Transition name="layout" mode="out-in" > 
  at <RouterView > 
  at <NuxtPage> 
  at <App> 
  at <NuxtRoot>
```
```
runtime-dom.esm-bundler.js:35 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'parentNode')
    at parentNode (runtime-dom.esm-bundler.js:35:30)
    at ReactiveEffect.componentUpdateFn [as fn] (runtime-core.esm-bundler.js:5109:17)
    at ReactiveEffect.run (reactivity.esm-bundler.js:160:29)
    at updateComponent (runtime-core.esm-bundler.js:4968:26)
    at processComponent (runtime-core.esm-bundler.js:4901:13)
    at patch (runtime-core.esm-bundler.js:4489:21)
    at patchKeyedChildren (runtime-core.esm-bundler.js:5262:17)
    at patchChildren (runtime-core.esm-bundler.js:5205:21)
    at processFragment (runtime-core.esm-bundler.js:4886:17)
    at patch (runtime-core.esm-bundler.js:4482:17)
```

### Describe the bug

hooks `page:start` and `page:finish` don't seem to fire on pages  `${url}/organization/*` and an error is thrown when using `await` in the `<script setup>` and then navigating to a new page. 

I don't know if this is due to the page structure or the the way the pages are set up. It's taken me all day to get this error to reproduce and the only way I've gotten this to throw is with this folder structure and by 
* first visiting a page that doesn't use await in the `<script setup>` 
* then visiting a page that does uses await
* then by using vue-router to navigate to another page

### Additional context

_No response_

### Logs

_No response_
```[tasklist]
### Tasks
```

### LINKED ISSUE #21534: Router.replace in <setup> breaks when using <NuxtLayout> within pages
### Environment

- Node Version:     v18.16.0
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/github/itm-ape/plugin-types/

### Describe the bug

TypeError when using router.replace inside `<script setup>` in a page-component while using `<NuxtLayout>`

In the sandbox i provided above I have 2 different pages. One using NuxtLayout that breaks and one without NuxtLayout that works. This used to work in Nuxt version 3.4.2. 

This error is preventing us from updating Nuxt. 

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #21544: v-for doesn't work on change layout
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/v-for-layout?file=layouts%2Fdefault.vue

### Describe the bug

When navigate to a route with a different layout, **and there's a `v-for` with a `ref()`** list inside the page, then return a vue warn and a error.

### Additional context

_No response_

### Logs

```shell-script
[Vue warn]: Unhandled error during execution of scheduler flush. This is likely a Vue internals bug. Please open an issue at https://new-issue.vuejs.org/?repo=vuejs/core 
  at <NuxtLayout> 
  at <App key=3 > 
  at <NuxtRoot>


Uncaught (in promise) TypeError: Cannot read properties of null (reading 'parentNode')
    at parentNode (chunk-MNWBZQKJ.js:82:30)
    at ReactiveEffect.componentUpdateFn [as fn] (chunk-3Q27KRJ5.js:5823:11)
    at ReactiveEffect.run (chunk-BOVKDABU.js?v=1765f250:181:19)
    at instance.update (chunk-3Q27KRJ5.js:5860:52)
    at callWithErrorHandling (chunk-3Q27KRJ5.js:213:32)
    at flushJobs (chunk-3Q27KRJ5.js:411:9)
```

antfu:
### 🔗 Linked issue

fix https://github.com/nuxt/nuxt/issues/13309
resolves https://github.com/nuxt/nuxt/issues/21544
resolves https://github.com/nuxt/nuxt/issues/21534

### ❓ Type of change

### 📚 Description

I am not 100% sure, but it fixes the error in this reproduction: https://github.com/nuxt/nuxt/issues/13309#issuecomment-1578090631

From what I see, is that our `LayoutLoader` is an async component that has a unique key. When the layout gets changed, a new `LayoutLoader` is created. Because Vue's suspense can not catch async patches on deep branches (an known issue), causing it to have a DOM that is already dismissed.

The solution to solve that is to wrap it with a suspensible Suspense to capture the update properly.

### 📝 Checklist


danielroe:
I think we _should_ inject `<Suspense>` above the layout loader here (ie. it's the right thing to do). We should also investigate the regressions in the test suite; they need to be solved, but I don't think it indicates this is the wrong approach.

darthf1:
Is it possible to land this the [v3.6.0](https://github.com/nuxt/nuxt/pull/21424) as it resolves a lot of issues (at least for me)? @antfu mentioned this does not work with transitions enabled, but if it does not break anything its already an improvement? I have transitions disabled because https://github.com/nuxt/nuxt/issues/13471.

danielroe:
This is not merged because it does break some other things (not just transitions). But we'll merge ASAP as soon as we can, and release in a patch.

packages/nuxt/src/app/components/layout.ts
```diff
@@ -84,18 +46,53 @@ export default defineComponent({
...
+const LayoutProvider = defineComponent({
+  name: 'NuxtLayoutProvider',
+  inheritAttrs: false,
+  props: ['name', 'layoutProps', 'hasTransition', 'shouldProvide'],
+  setup (props, context) {
+    // Prevent reactivity when the page will be rerendered in a different suspense fork
+    if (props.shouldProvide) {
+      // eslint-disable-next-line vue/no-setup-props-destructure
+      const name = props.name
+      provide('_layout', {
```
antfu, line 72: I guess we could use a symbol as the injection key to avoid conflicts with user code?

packages/nuxt/src/app/components/layout.ts
```diff
@@ -84,18 +46,66 @@ export default defineComponent({
...
+      type: Boolean
+    }
+  },
+  setup (props, context) {
+    // Prevent reactivity when the page will be rerendered in a different suspense fork
+    if (props.shouldProvide) {
+      // eslint-disable-next-line vue/no-setup-props-destructure
+      const name = props.name
+      provide('_layout', {
+        isCurrent: (route: RouteLocationNormalizedLoaded) => name === (route.meta.layout ?? 'default')
```
antfu, line 86: Just out of curiosity, what would be the difference between providing the name in plain string over a function?

c-schwan:
@antfu is not ready for review, it does not fix the all the linked issues

danielroe:
@c-schwan Could you be more specific? A partial fix is still worth merging but sometimes error messages like #13309 stem from different causes and aren't fixed in a single change.

c-schwan:
> @c-schwan Could you be more specific? A partial fix is still worth merging but sometimes error messages like #13309 stem from different causes and aren't fixed in a single change.

#21701 should work on my opinion for major releases. if you have a guard like auth protecting page resources it should work as expected

danielroe:
@c-schwan This PR fixes that issue, in my testing.

c-schwan:
> @c-schwan This PR fixes that issue, in my testing.

Nuxt 3.6.0-28124217.43844474 (npm:nuxt3@pr-21450)

https://stackblitz.com/edit/middleware-nuxt

but still not working

danielroe:
Have you tried it locally?

c-schwan:
> Have you tried it locally?

i tried locally with the stackblitz project and locally in the our project, both no luck

test/fixtures/basic/components/ComponentWithRef.vue
```diff
@@ -0,0 +1,12 @@
+<script setup lang="ts">
+const list = ['a', 'b', 'c']
+const templateRef = ref<HTMLElement[] | null>(null)
+</script>
+
+<template>
+  <div>
+    <div v-for="item in list" ref="templateRef">
```
danielroe, line 8:
```suggestion
    <div v-for="item in list" ref="templateRef" :key="item">
```

c-schwan:
> Have you tried it locally?

it works well on the first time switching the layout but going back (Browser or via NuxtLink) it won't be called anymore

packages/nuxt/src/app/components/layout.ts
danielroe: it works just as well, but means we have to move the `?? 'default'` into page.ts, and somehow I felt it made sense for it to be co-located in layout.ts where the matching logic is. But happy to refactor this!

danielroe (approved):

PR #21363, opened 2023-06-04
fix(schema): update `vue-loader` options type

Lehoczky:
### 🔗 Linked issue

#21332

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The [schema for the vue-loader options](https://github.com/nuxt/nuxt/blob/main/packages/schema/src/config/webpack.ts#L154-L163) did not extend the [VueLoaderOptions ](https://github.com/vuejs/vue-loader/blob/main/src/index.ts#L32-L62) interface and therefore the generated type, which the `defineNuxtConfig` function used was incorrect. 

There are probably issues with other properties of the [webpack.loaders](https://github.com/nuxt/nuxt/blob/main/packages/schema/src/config/webpack.ts#LL136C7-L136C7) object as well. They most likely should extend the option interface of their webpack plugin too. I can try to address this in another PR if you want, Daniel.

Before:
image

After:
image

Resolves #21332

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #21475, opened 2023-06-08
fix(schema): add more explicit types for config schema

Lehoczky:
### 🔗 Linked issue

Follow-up for https://github.com/nuxt/nuxt/pull/21363

### ❓ Type of change

### 📚 Description

This PR adds or updates the type annotation for the config properties wherever it was achievable (some webpack loaders for example don't have their own type annotations).

For other details, see my comments below.

### 📝 Checklist


packages/schema/src/config/generate.ts
```diff
@@ -14,11 +14,15 @@ export default defineUntypedSchema({
...
      * ```
+     * 
+     * @type {string | string[]}
      */
     routes: [],
 
     /**
      * This option is no longer used. Instead, use `nitro.prerender.ignore`.
+     * 
+     * @deprecated
```
Lehoczky, line 25: Because we don't use this value, it felt appropriate to mark it as deprecated, but I'm happy to revert this change if we have plans for this prop later.

packages/vite/src/vite-node.ts
```diff
@@ -48,7 +48,7 @@ export function viteNodePlugin (ctx: ViteBuildContext): VitePlugin {
           markInvalidates(server.moduleGraph.getModulesByFile(typeof plugin === 'string' ? plugin : plugin.src))
         }
         for (const template of ctx.nuxt.options.build.templates) {
-          markInvalidates(server.moduleGraph.getModulesByFile(template?.src))
+          markInvalidates(server.moduleGraph.getModulesByFile(template?.src as string))
```
Lehoczky, line 51: because `ctx.nuxt.options.build.templates` is no longer `any`, TypeScript shows an error here, because the `src` in [NuxtTemplate](https://github.com/nuxt/nuxt/blob/main/packages/schema/src/types/nuxt.ts#L30) can be `undefined`. 

I'm guilty of not trying out what happens (or even can it happen) if it is `undefined` here, but I can investigate it further.

packages/schema/src/config/adhoc.ts
```diff
@@ -70,7 +70,7 @@ export default defineUntypedSchema({
    * 
    * @see  [Nuxt DevTools](https://devtools.nuxtjs.org/) for more information.
    * @experimental
-   * @type {boolean | { enabled: boolean }}
+   * @type {boolean | import('@nuxt/devtools').ModuleOptions}
```
danielroe, line 73: For this, the user might not have devtools installed yet (although we do default now to installing it with the starter)

Lehoczky: I've added the package to the dev dependencies, but you are right, this should either be something like

```ts
interface {
  enabled: boolean
  [key: string]: any
} 
```

or `@nuxt/devtools` should be moved to the dependencies

packages/vite/src/vite-node.ts
danielroe: Hm. We do need to investigate - is it possible that the module would be added by `template.dst` instead? 🤔  This may be the origin of a bug we're encountering elsewhere...

Lehoczky: By quickly looking at the code, I think we expect to use `template.getContents` when `src` is not present ([see](https://github.com/nuxt/nuxt/blob/main/packages/kit/src/template.ts#L54)), and we call that function [at one place in the codebase](https://github.com/nuxt/nuxt/blob/main/packages/kit/src/internal/template.ts#L21), but we never set `src` to anything if it is `undefined`, so we can potentially call `server.moduleGraph.getModulesByFile()` with `undefined`. 

Although yeah, I just glanced over the code, I'm gonna check this out more thoroughly tomorrow.

packages/schema/src/config/adhoc.ts
danielroe: We later augment the type in `.nuxt/types/schema.d.ts`:

https://github.com/nuxt/nuxt/blob/3e1ef92dc5ff808bf417ef3967365ab555036c6a/packages/nuxt/src/core/templates.ts#L124-L169

We need to make sure that whatever we set this to does not conflict with that.

I'd prefer not to install devtools as a dependency as we've kept it separate in order to iterate more quickly.

Lehoczky: I've removed `@nuxt/devtools` from the dev dependencies, and updated the type

danielroe (approved):
This is a great PR - many thanks for it ❤️

Lehoczky:
> This is a great PR - many thanks for it ❤️

My pleasure @danielroe  <3

PR #21488, opened 2023-06-09
fix(babel-preset-app): add explicit dep on used babel plugin

danielroe:
### 🔗 Linked issue

https://github.com/babel/babel-plugin-proposal-private-property-in-object/issues/1#issuecomment-1584287021
resolves nuxt/nuxt#21489
resolves https://github.com/nuxt/create-nuxt-app/issues/1030

### ❓ Type of change

### 📚 Description

This adds an explicit dependency on the version of `@babel/plugin-proposal-private-property-in-object` we need in the project.

### 📝 Checklist

### DIFF
--- a/packages/babel-preset-app/package.json
+++ b/packages/babel-preset-app/package.json
@@ -19,6 +19,7 @@
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.18.6",
     "@babel/plugin-proposal-optional-chaining": "^7.21.0",
     "@babel/plugin-proposal-private-methods": "^7.18.6",
+    "@babel/plugin-proposal-private-property-in-object": "^7.20.0",
     "@babel/plugin-transform-runtime": "^7.22.4",
     "@babel/preset-env": "^7.22.4",
     "@babel/runtime": "^7.22.3",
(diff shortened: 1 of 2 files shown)


Eugene313 (approved):

nicolo-ribaudo:
@danielroe If it takes too long to release this, I can temporarily add a workaround in Babel. Out of curiosity, did you get similar bug reports ~1 week ago?

In Babel 7.22.0 we removed the `@babel/plugin-proposal-private-property-in-object` because we didn't need anymore. That broke all create-react-app installs because it has a bug identical to the one fixed in this PR, and since create-react-app is not maintained we temporarily added the dependency back to Babel.

To avoid having the unnecessary dependency, we now replaced it with a placeholder whose unique goal is to fix CRA compatibility, and that otherwise behaves as a missing dependency when required.

I'm surprised that we didn't discover this problem in Nuxt when we released 7.22.0 🤔

nicolo-ribaudo:
Oh well :)

danielroe:
Sorry for the pain @nicolo-ribaudo!

For everyone else, until this is released in 2.17, you can opt in to use `nuxt-edge` in your projects, or add this plugin explicitly as a dev dependency. (I'll be triggering an edge release shortly.)

We expect to release 2.17 this weekend.

Eugene313:
Yes I had same problem 1 week ago

PR #21501, opened 2023-06-09
fix(vite): invalidate templates by `dst` not `src`

danielroe:
### ❓ Type of change

### 📚 Description

We were previously invalidating templates by their _source_ but in fact it's the `dst` that will actually be part of the build. In most cases, the file that needs to be valid is actually prefixed with `virtual:nuxt:` but we already cover that earlier:

https://github.com/nuxt/nuxt/blob/ff842347bcec23bbe7a50e25c4a8583d5565a91f/packages/vite/src/vite-node.ts#L42-L46

I'm not sure this has actually caused any issues, but 🤷‍♂️ 

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -48,7 +48,7 @@ export function viteNodePlugin (ctx: ViteBuildContext): VitePlugin {
           markInvalidates(server.moduleGraph.getModulesByFile(typeof plugin === 'string' ? plugin : plugin.src))
         }
         for (const template of ctx.nuxt.options.build.templates) {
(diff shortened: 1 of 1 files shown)


slavanossar:
Sorry for the ping @danielroe but would the following error be related to this?

```
Uncaught TypeError: Failed to resolve module specifier "vite/preload-helper". Relative references must start with either "/", "./", or "../".
```

It happens on page load in production build, very tricky to troubleshoot as console reports error coming from the HTML doc, been stuck on this for a few days now. If not, will open an issue describing it.

Thanks!

danielroe:
It's not likely to be related to this PR, unfortunately.

PR #21521, opened 2023-06-11
fix(nuxt): abort navigation when updating `window.location`

### LINKED ISSUE #21314: In CSR app, `navigateTo` (with `external:true`) redirection from route middleware triggers page script setup
### Environment

Nuxi 3.5.2
- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.2
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      ssr
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

From https://stackblitz.com/edit/nuxt-starter-77e2za?file=README.md, click on `Go to foo` link on home page.

You can see `Hello, foo!` blinking on the browser before it gets redirected. Also, you can see the log `onMounted: hello, foo!` from the console.

### Describe the bug

In a CSR application, when using a route middleware to redirect externally from a Nuxt page to another location (by using `navigateTo` with `external: true`), the setup script of the page where we apply the middleware is executed whereas it should not be (I believe?). As well as the lifecycle hooks such as `onMounted`. It's not the case when redirecting to an internal route.

### Additional context

_No response_

### Logs

_No response_

nicolaspayot:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/21314

### ❓ Type of change

### 📚 Description

On client side, when we use the `navigateTo` function to redirect to an external URL from a middleware (and when the Nuxt application is hydrated => while navigating from `NuxtLink` or `navigateTo`, ...), we abort the navigation so that the page component is not rendered before location changes.

Resolves #21314

### 📝 Checklist


packages/nuxt/src/app/composables/router.ts
```diff
@@ -183,6 +184,13 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
     } else {
       location.href = toPath
     }
+    // Within in a Nuxt route middleware handler
+    if (inMiddleware) {
+      // Abort navigation when app is hydrated
+      if (!nuxtApp.isHydrating) {
+        return false
+      }
+    }
```
nicolaspayot, line 193: There is a corner case that's not handled here: when the Nuxt app is loaded on a route that has a redirect middleware. The page gets rendered (and is visible during a very short time) before the location changes. If we also return `false` when `nuxtApp.isHydrating`, it renders a 404 error / page (that's visible during a very short time).

There's something we can do: delay the return statement, like so, for example:
```ts
if (!nuxtApp.isHydrating) {
  return false
}
return new Promise(resolve => setTimeout(resolve, 10))
```
But it looks a bit hacky 🤔

danielroe: I wonder what happens if you return a never-resolving promise 🤔

nicolaspayot: I actually think that's what's happening with `return new Promise(resolve => setTimeout(resolve, 10))` as the location changes before the `resolve` function gets called.

I like the idea of returning a never-resolving promise (`return new Promise(() => {})`) for this case 👍

danielroe (approved):

danielroe:
Thank you so much ❤️

PR #21525, opened 2023-06-11
docs: add styling section

Hebilicious:
- docs: add styling section

### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/21491

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds a styling section to the documentation. 
Please correct me if there's any mistakes/wrong information.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #21573, opened 2023-06-14
fix(nuxt): inline css directly in root component

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/21525

### ❓ Type of change

### 📚 Description

This removes the (unintended) difference between importing CSS in the `css` property of your `nuxt.config` vs importing directly in a vue component. It will be inlined as expected in your HTML if you have `experimental.inlineSSRStyles` enabled.

This PR also takes things further to remove duplicated CSS from the `css` property in nuxt config. If it is inlined, it won't be included in the emitted CSS files as well.

The test makes clear the opportunities for further improvement here.

### 📝 Checklist


RodrigoProjects:
Maybe not related but I found an issue with inlining CSS, when I use the same component in two different places and one is loaded in SSR and the other is wrapped inside a ClientOnly it seems it does not inline the CSS for that component. If the ClientOnly counter-part is removed it works fine. Any insights or relation with this issue?

danielroe:
Yes, that is a current limitation of style inlining - it happens on the server, and the server can't 'see' client-only components as they are not part of the server build.

RodrigoProjects:
Just a couple of important notions here: 
 - When I say it's a ClientOnly component I mean I have a component like: `<ClientOnly><DummyComponent /></ ClientOnly>`
 - I have two instances of the component, one just uses the `<DummyComponent>` without the `<ClientOnly>` component somewhere on the page, so it is present on the server build. The other instance is equal to the first notion and it is present in the page too.
 
 Removing the second notion makes the inline work again, it's kinda weird for me to understand this logical flow (thinking)

huang-julien (commented):
Are we supposed to have the global styles injected twice in the HTML due to server components ?
I don't think that was the case before.

danielroe:
Good point! Here are the options, I think:

1. Do not inline global CSS.
2. Inline global CSS for all routes via `nuxt-root` and include it in islands.
3. Inline global CSS for all routes via `nuxt-root` and filter it out of island response.
4. Inline global CSS via `app.vue` - it won't be inlined on error pages or islands.

I think we _do_ need to inline global CSS, including on error pages. So we probably need to filter it out of the island response and (possibly) need to revisit when we come to remote sources.

huang-julien:
If we want to include the global styles in the remote islands, we'd need to avoid filtering it only if the request comes from another origin i guess

packages/nuxt/src/core/plugins/css-import.ts
```diff
@@ -0,0 +1,32 @@
...
+export const CSSPlugin = (options: CSSPluginOptions) => createUnplugin(() => {
+  return {
+    name: 'nuxt:css-import:transform',
+    enforce: 'pre',
+    transformInclude (id) {
+      return id === options.rootId()
+    },
+    transform (code) {
+      const s = new MagicString(code)
+      s.replace(SCRIPT_RE, ['$0', ...options.css.map(i => genImport(i))].join('\n'))
```
antfu, line 22: We would better check `s.hasChanges()` and turn `null` to bypass transform in improve perf.

antfu (approved):

PR #21585, opened 2023-06-15
feat(nuxt): use build plugin to access nuxt route injection

### LINKED ISSUE #20674: $route updated too early when switching page
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.3
- Nitro Version:    2.3.3
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-wyu52r

1. Click "Open in new tab"
2. Open dev tools
3. Click "bar" in the top menu
4. See Vue warnings in console

### Describe the bug

When a page is switched with `<nuxt-link>`, `$route` updates too early (before the old page unmounts). This leads to a whole bunch of bugs unless specifically treated.

In this reproduction, it emits Vue warnings when switching from `foo` to `bar`:

```
[Vue Router warn]: No match found for location with path "/bar/1"
[Vue Router warn]: No match found for location with path "/bar/2"
[Vue Router warn]: No match found for location with path "/bar/3"
```

because `foo/index.vue` was generating links with `$route.path + "/${id}"` and somehow it re-renders with the new `$route` before the page fully switches to `bar`.

Another example could be page watching `$route.params` and updating itself, which also crashes unless specifically handled.

This was not happening in Nuxt 2, and makes it difficult to port old apps.

### Additional context

In this simple reproduction I could be generating static links with hardcoded prefix such as `/foo/${id}`, but that doesn't work for reusable components so it's not really an option.

The problem is broader than simply vue-router warnings. The old page should not re-render with the new `$route` at all!

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20674

### ❓ Type of change

### 📚 Description

This adds an experimental option to sync `$route` template object with the Nuxt-managed route.

### 📝 Checklist


packages/nuxt/src/pages/module.ts
```diff
@@ -252,6 +252,11 @@ export default defineNuxtModule({
     // Add prefetching support for middleware & layouts
     addPlugin(resolve(runtimeDir, 'plugins/prefetch.client'))
 
+    if (nuxt.options.experimental.templateRouteMixin) {
+      // Add global mixin to ensure template $route is kept in sync with `<NuxtPage>`
+      addPlugin(resolve(runtimeDir, 'plugins/prefetch.client'))
```
Atinux, line 257:
```suggestion
      addPlugin(resolve(runtimeDir, 'plugins/route-injection.client'))
```

danielroe: Oh goodness 🙈

Atinux:
Not sure that this should be as an experimental feature since it actually give the same behavior without and with vue-router.

Also, I also think that the `$route` and `useRoute()` should only changed once the next page is resolved.

I would consider this as a bug fix IMO. Happy to discuss about it if you see it with another POV.

danielroe:
Agreed. However, personally I would advise against using these kind of injections as much as possible in the template as it makes the 'dependencies' of the component less explicit.

Yet the main issue I have is that I'm not sure of the impact on performance. We might be able instead to transform and auto execute useRoute based on usage. What do you think?

danielroe:
@antfu Would love your thoughts on whether it's safe to access `_ctx._.provides` in this way.

danielroe:
@antfu Very up for a different implementation here, but wanted to get this in for 3.7 release.

PR #21611, opened 2023-06-17
perf(nuxt): extract and apply plugin order at build time

danielroe:
### ❓ Type of change

### 📚 Description

When we introduced object-syntax plugins, we highlighted that this might be extracted at build-time.

This plugin implements a build-time step to extract plugin metadata, and also normalise plugins at build-time.

This slightly reduces the work that needs to be done on Nuxt app hydration.

### 📝 Checklist


huang-julien:
BTW it seems that using the devtools completly breaks the dev server at this line 
image

`[nuxt] [request error] [unhandled] [500] plugin.apply is not a function `

huang-julien:
image

i have an issue with the route injection now (devtools enabled), route is being undefined. I can't reproduce it on the main branch but it is really strange that we have an injection issue

packages/nuxt/src/core/plugins/plugin-metadata.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+  walk(parse(code) as Node, {
+    enter (_node) {
+      if (_node.type !== 'CallExpression' || (_node as CallExpression).callee.type !== 'Identifier') { return }
+      const node = _node as CallExpression & { start: number, end: number }
+      const name = 'name' in node.callee && node.callee.name
+      if (name !== 'defineNuxtPlugin' && name !== 'definePayloadPlugin') { return }
+
+      if (name === 'definePayloadPlugin') {
+        meta.order = -40
```
antfu, line 45: Should we have them as a constant map as well to avoid the usage of magic numbers?

packages/nuxt/src/core/plugins/plugin-metadata.ts
```diff
@@ -0,0 +1,181 @@
...
+      meta[propertyKey] = property.value.value as any
+    }
+    if (property.value.type === 'UnaryExpression' && property.value.argument.type === 'Literal') {
+      meta[propertyKey] = JSON.parse(property.value.operator + property.value.argument.raw!)
+    }
+  }
+  return meta
+}
+
+export const RemovePluginMetadataPlugin = (nuxt: Nuxt) => createUnplugin(() => {
```
antfu, line 97: Maybe we could have some unit tests for transforms like this?

packages/schema/src/types/nuxt.ts
```diff
@@ -52,6 +52,7 @@ export interface NuxtPlugin {
   ssr?: boolean
   src: string
   mode?: 'all' | 'server' | 'client'
+  order?: number
```
antfu, line 55: Maybe we could have some jsdoc on this to note our predefined numbers in [plugin-metadata.ts](https://github.com/nuxt/nuxt/pull/21611/files#diff-cd49fde9efc76ee9562216ef2ae79223c8c134c9fdb1de5c68a03922d0d22a86)?

huang-julien (approved):
:+1:

PR #21636, opened 2023-06-19
feat(nuxt): expose `useNuxtApp` to window for convenience

### LINKED ISSUE #15587: Allow accessing to nuxt instance from window shortcut
> Originally asked by @Atinux to bring back similar feature (`window.$nuxt`) from Nuxt 2

**Security:** Nuxt instance can be currently accessed in several ways from the global `window` in client-side, used by Vue Devtools and other extensions. Exposing it via a shortcut, does not imply any security concerns. 

**History:** In Nuxt 2, it was being exposed as a single `window.$nuxt` both for debugging and also plugins that needed a nuxt instance outside of the plugin. It made some issues especially for when using more than once instance of Nuxt in one window (although never recommended). As a workaround we introduced a series of `globalName` options that allowed us to manually change it to something else and avoid conflicts. However it required heavily templating techniques that we no longer use for Nuxt 3. Also it was error prune for module authors to consider that this global name can be customized.

**New API:** For Nuxt 3, i propose we expose it as `$useNuxtApp()` with forward compatibility in mind if ever support back multiple instances per page, using `$useNuxtApp(name?)`. Possibly using bundler plugins to automatically inject the (custom) name.

**Current workarounds:**

- Using Unctx in Window: `__unctx__.get('nuxt-app').use()`
- Using Vue in DOM: `[...document.querySelectorAll('*')].find(el => el.__vue_app__?.$nuxt).__vue_app__.$nuxt`

**Remarks:** Accessing `nuxtApp` from global instance is still in all ways considered an anti-pattern outside of lifecycle (vue rendering and plugins) as we cannot use `useNuxtApp()` even client-side which is intentionally. This API should be documented strictly for debugging purposes. And we might also offer an option to disable it.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15587

### ❓ Type of change

### 📚 Description

This allows access to the nuxtApp on client side through a global `window.useNuxtApp` which doesn't rely on auto-imports.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -286,7 +286,7 @@ export function createNuxtApp (options: CreateOptions) {
       nuxtApp.callHook('app:chunkError', { error: (event as Event & { payload: Error }).payload })
     })
 
-    window.useNuxtApp = useNuxtApp
+    window.useNuxtApp = window.useNuxtApp || useNuxtApp
 
     // Log errors captured when running plugins, in the `app:created` and `app:beforeMount` hooks
     // as well as when mounting the app.
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -286,6 +286,8 @@ export function createNuxtApp (options: CreateOptions) {
       nuxtApp.callHook('app:chunkError', { error: (event as Event & { payload: Error }).payload })
     })
 
+    window.useNuxtApp = useNuxtApp
+
     // Log errors captured when running plugins, in the `app:created` and `app:beforeMount` hooks
     // as well as when mounting the app.
     const unreg = nuxtApp.hook('app:error', (...args) => { console.error('[nuxt] error caught during app initialization', ...args) })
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/nuxt.ts
```diff
@@ -286,6 +286,8 @@ export function createNuxtApp (options: CreateOptions) {
       nuxtApp.callHook('app:chunkError', { error: (event as Event & { payload: Error }).payload })
     })
 
+    window.useNuxtApp = useNuxtApp
```
pi0, line 289: (Maybe we avoid overriding existing util if exists w/without a warning?)

pi0 (approved):

packages/nuxt/src/app/nuxt.ts
pi0: The comment above is also little bit confusing https://github.com/nuxt/nuxt/blob/1f8b3fa5e8dc0347cde3b376a600aec2d5ecc985/packages/nuxt/src/app/nuxt.ts#L283

danielroe:
```suggestion
    window.useNuxtApp = window.useNuxtApp || useNuxtApp
```

atinux (approved):

Issue #14507, opened 2022-08-07
Payload extraction support


pi0:

codeflorist:
@pi0 @danielroe Is there any reason, why payload files are not hashed? This can be a rather annoying and dangerous trap, since normally one caches JS files indefinitely with Nuxt3, and it requires some special workarounds to exclude the payload-files from this. And once they are in the browser-cache of the users, they are there to stay, preventing these users to see updated content, even if the cache control headers were updated in the meantime.

If it has to be this way, i would definitely recommend adding a special notice in the docs (e.g. on https://nuxt.com/docs/api/commands/generate).

Smef:
I've run into a similar issue with the payload extraction being cached indefinitely. Here's an example of what is happening with the caching and how it's a problem.

We have a blog site with an index page showing the 10 most recent posts: post numbers 10-1 listed in that order

```
10
9
8
7
6
5
4
3
2
1
```

SSG creates static HTML which has these 10 posts in this order, and a matching `payload.js` which also has the same 10 posts in order. The `payload.js` is used for hydration after the HTML is loaded, and everything is good.

When a new post is created and the SSG is re-run, new HTML is created which lists the new 11th post, and post 1 is bumped off to the next page:

```
11
10
9
8
7
6
5
4
3
2
```

A matching` payload.js` is also created, which lists these 10 posts (11 - 2) in the correct order.

Unfortunately, this is where we run into problems.

When a user visits the site they get the static HTML of posts 11-2. If they had visited the page before, they still have the old `payload.js` cached in their browser, and the 10-1 payload gets used for hydration. This means that Vue hydrates and replaces the content on the page with the cached data. While the correct 11-2 content was on the page for a split second, it quickly gets replaced by the old 10-1 content. The user doesn't get to see the new content on the site due to the mismatch between the HTML content and the cached payload content. Refreshing the page leaves you in a weird state of mixed content, and you have to totally clear the browser cache to get the new `payload.js` properly loading.

unshame:
I have a project where I want to only make fetch requests at build time for a subset of pages. Right now, when navigating to a page, the fetch requests are made on the client even if  `experimental.payloadExtraction` is enabled. I'm assuming the bullet point `Payload prefetching with hybrid rendering` is talking about this.

[...]

PR #21641, opened 2023-06-19
fix(nuxt): experimental build manifest + client route rules

### LINKED ISSUE #15024: Page with redirect not generated, causing "Cannot load payload  ..._payload.js TypeError: Failed to fetch dynamically imported module"
### Environment

- Operating System: `Linux`
- Nitro Version:    `0.5.4`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-xyxale

### Describe the bug

The page `redirect.vue` just contains a redirect:
```html
<template>
  <div />
</template>

<script setup>
navigateTo('/target');
</script>
```

When running `nuxi generate` there is no corresponding static page generated.

But scripts try to fetch the corresponding `/redirect/_payload.js` causing warnings in the console:
```
entry.d2c112ae.js:5 [nuxt] Cannot load payload  /redirect/_payload.js TypeError: Failed to fetch dynamically imported module: https://nuxt-starter-xyxale--3000.local.webcontainer.io/redirect/_payload.js
```

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #19228: `routeRules` apply only to server responses
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.3`
- Builder:          `vite`
- User Config:      `routeRules`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-u1ewa9?file=pages/index.vue

### Describe the bug

Using the `routeRules` part of the configuration to specify redirects only works when changing the URL, but not when using a `NuxtLink` to navigate, and also does not work with trailing slashes, even though they should be allowed by default (and do work for normal navigation to existing pages).

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #21370: Provide an app manifest
### Describe the feature

Nuxt should provide an app manifest at build time which contains useful information about the application, such as a list of routes and their mapped payloads.

In the future, it could also be used to differentiate between deployed versions.

Related: https://github.com/nuxt/nuxt/pull/18419, https://github.com/nuxt/nuxt/issues/20972

### Additional information

### Final checks

### LINKED ISSUE #22720: [SSG] Prerendered ignored routes will still request payload(and get 404).
### Environment

```
- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      nitro
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/nuxt-starter-odxwrz?file=pages%2Findex.vue,pages%2Fmanage.vue,nuxt.config.ts

Run: `npm run generate && npm run preview`

### Describe the bug

When navigating to /manage, nuxt-link will attempt to fetch `/manage/_payload.json` and receive a 404 response. In my Vercel production environment, all 404 errors are handled with a default Nuxt content page. Given that I have numerous routes under /manage, these nonexistent _payload.json files can potentially consume a significant amount of bandwidth and make website slow. In my situation, the content size of the 404 page is 18kb.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #21370
closes https://github.com/nuxt/nuxt/pull/18419
resolves https://github.com/nuxt/nuxt/issues/15024
related: https://github.com/nuxt/nuxt/issues/14507
resolves https://github.com/nuxt/nuxt/issues/19228
resolves nuxt/nuxt#22720

### ❓ Type of change

### 📚 Description

This adds an app manifest (accessible at `/_nuxt/builds/meta.json`) and uses it to determine whether to load a payload. We might allow this to be toggleable as an experimental feature.

**TODO**:
- [x] respect redirect rules within router middleware
- [x] extract relevant parts to nitro and refactor as needed

### 📝 Checklist


Hebilicious:
Some things that I would like to see in the manifest : 

- development mode
- devtools support
- list of all server handlers / api routes 
- list of all pages

packages/nuxt/src/app/composables/manifest.ts
```diff
@@ -0,0 +1,18 @@
...
+  routeRules: Record<string, any>
+  prerendered: string[]
+}
+
+let manifest: Promise<NuxtAppManifest>
+
+export function getAppManifest (): Promise<NuxtAppManifest> {
+  const config = useRuntimeConfig()
+  // TODO: use build id injected
+  manifest = manifest || $fetch(joinURL(config.app.cdnURL || config.app.baseURL, '_builds/latest.json'))
```
pi0, line 16: It would be nice if we can handle promise rejections to clear value and improve error

pi0: PS: We need a ttl somehow. User opens page and fetches manifest > a deploy happens > client side navigation is again not aware of updates!

packages/nuxt/src/core/nitro.ts
```diff
@@ -203,6 +206,36 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+  if (nitroConfig.prerender?.routes) {
+    const manifestPrefix = '/_builds'
+    nitroConfig.prerender.routes.push(joinURL(manifestPrefix, 'latest.json'))
+    nitroConfig.prerender.routes.push(joinURL(manifestPrefix, `meta.${buildId}.json`))
+    nitroConfig.devHandlers!.push({
+      route: joinURL(manifestPrefix, '**'),
+      handler: defineEventHandler(() => ({
+        id: 'dev',
+        timestamp: Date.now(),
+        routeRules: {},
```
pi0, line 221: Maybe WIP? In dev mode, route rules such as redirect can be applied as well

pi0 (commented):
Nice work 💯

packages/nuxt/src/core/nitro.ts
```diff
@@ -369,8 +402,30 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+          const rules = defu({}, ...routeRulesMatcher.matchAll(url).reverse()) as Record<string, any>
+          if (!rules.prerender) {
+            prerenderedRoutes.add(url)
+          }
+        }
+      })
       await prerender(nitro)
 
+      for (const file of ['latest.json', `meta.${buildId}.json`]) {
+        const manifestFile = join(nitro.options.output.publicDir, '_builds', file)
```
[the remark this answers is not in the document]
danielroe: Initially I was thinking directly `_nuxt/builds` as it's a build asset - updated as there was an issue when prerendering - let me check again....

packages/nuxt/src/app/plugins/payload.client.ts
```diff
@@ -26,5 +18,15 @@ export default defineNuxtPlugin({
...
     })
+
+    onNuxtReady(() => {
+      // Load payload into cache
+      nuxtApp.hooks.hook('link:prefetch', async (url) => {
+        if (!parseURL(url).protocol) {
+          await loadPayload(url)
+        }
+      })
+      getAppManifest()
```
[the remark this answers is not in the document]
danielroe: we need it to prefetch payloads, and visible links will be prefetched as soon as nuxt is marked as ready, so adding a timeout might not make much difference

but happy to 👍

packages/nuxt/src/core/nitro.ts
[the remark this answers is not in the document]
danielroe: let's leave it as `_builds` for now but iterate (happy to rename when nitro feature comes out). Other options could be `_meta` or to keep it as `_nuxt/builds`

packages/nuxt/src/app/composables/payload.ts
```diff
@@ -5,27 +5,33 @@ import { getCurrentInstance } from 'vue'
...
-  if (cache[payloadURL]) {
+  if (cache[payloadURL] || cache[payloadURL] === null) {
     return cache[payloadURL]
   }
-  cache[payloadURL] = _importPayload(payloadURL).then((payload) => {
-    if (!payload) {
-      delete cache[payloadURL]
-      return null
+  cache[payloadURL] = getAppManifest().then((manifest) => {
+    if (manifest.prerendered.includes(url) || Object.keys(manifest.routeRules).some(key => key === url || (key.endsWith('/**') && (url + '/').startsWith(key.replace('/**', '/'))))) {
```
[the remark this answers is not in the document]
danielroe: note: it adds 4kb to the bundle

packages/nuxt/src/core/nitro.ts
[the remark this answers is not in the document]
pi0: `_nuxt/builds` can be tricky because `_nuxt` itself is part of a single build.

`_builds` could also be tricky to align next minor versions of nitro because it might override nuxt one. Why not using `_nuxtBuilds` for now?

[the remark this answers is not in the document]
danielroe: aesthetic reasons only - I don't think it looks or sounds nice. I want it to sound generic, without `nuxt` included in the name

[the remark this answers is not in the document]
pi0: 🤷 I think for the same reason we use `_nuxt` as the default build dir (however customizable). It is really important otherwise my hands will be tied to even try introducing the feature in nitro!! (same for users to leverage a directory called `_build` they won't be able to)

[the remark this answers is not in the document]
danielroe: previously you said:
> however i think for nitro we should also support custom name so it is nuxt compatible

so I thought it would be fine to use `_builds` for nuxt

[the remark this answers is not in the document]
pi0: By default, it will generate it when landed. PR thinking more on this topic here, i think even for nuxt is unsafe to use a generic name like `_builds`. Users might need a public directory like this and it will conflict without vendor prefix.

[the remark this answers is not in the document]
danielroe: What generic name will you use in Nitro? I can avoid it then.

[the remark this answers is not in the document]
pi0: I guess what I am trying to say is, for nitro, it should also be `_nitro/builds` or something with vendor prefix but nuxt probably shouldn't also use `_builds`. it was something unforeseen in our early discussions. (issue is not nitro anymore but also for nuxt to be non conflicting)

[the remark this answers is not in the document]
pi0: It can still be a required path for users. Same as `/.well-known/` which is however starting with dot, it is a valid path and might be needed.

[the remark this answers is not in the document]
danielroe: As with nitro, I think we can make it configurable if users need an escape hatch. 👍

packages/nuxt/src/app/composables/manifest.ts
danielroe: excellent point. on it 👍

danielroe: actually, seems I already implemented at https://github.com/nuxt/nuxt/commit/4f85cff8d94c575c5f4238a4a8f8472beb1bce5d. Review very appreciated 🙏

packages/nuxt/src/app/plugins/payload.client.ts
```diff
@@ -26,5 +18,15 @@ export default defineNuxtPlugin({
...
     })
+
+    onNuxtReady(() => {
+      // Load payload into cache
+      nuxtApp.hooks.hook('link:prefetch', async (url) => {
+        if (!parseURL(url).protocol) {
+          await loadPayload(url)
+        }
+      })
+      setTimeout(getAppManifest, 1000)
```
pi0, line 29: Would be nice if we also checked the connection speed as well. Performance "hopefully" is okay because hopefully, mobiles load the critical path in less than a second. But i am more worried about paid internet connections that waste bandwidth if they don't need the manifest.

packages/nuxt/src/app/plugins/check-outdated-build.client.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default defineNuxtPlugin((nuxtApp) => {
+  let timeout: NodeJS.Timeout
+  const config = useRuntimeConfig()
+
+  async function getLatestManifest () {
+    const currentManifest = await getAppManifest()
+    if (timeout) { clearTimeout(timeout) }
+    timeout = setTimeout(getLatestManifest, 1000 * 60 * 60)
+    const meta = await $fetch<NuxtAppManifestMeta>(joinURL(config.app.cdnURL || config.app.baseURL, '_builds/latest.json'))
```
pi0, line 13: Should we update here too to assetsDir/builds?

danielroe: quite right :+1:

packages/nuxt/src/app/plugins/payload.client.ts
danielroe: how do you think we should handle mobile or slow internet? increase the timeout?

pi0: longer timeout seems a fine or also maybe conditionally disable here to lazily fetch it on CSR navigations? I see we await on it in middleware

packages/nuxt/src/app/plugins/chunk-reload.client.ts
```diff
@@ -14,11 +15,19 @@ export default defineNuxtPlugin({
...
     nuxtApp.hook('app:chunkError', ({ error }) => { chunkErrors.add(error) })
 
+    function reloadAppAtPath (to: RouteLocationNormalized) {
+      const isHash = 'href' in to && (to.href as string).startsWith('#')
+      const path = isHash ? config.app.baseURL + (to as any).href : joinURL(config.app.baseURL, to.fullPath)
+      reloadNuxtApp({ path, persistState: true })
+    }
+
+    nuxtApp.hook('app:manifest:update', () => {
+      router.beforeResolve(reloadAppAtPath)
```
danielroe, line 25: comment from @atinux: consider checking for latest manifest when there are other router errors?

packages/nuxt/src/app/plugins/check-outdated-build.client.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { NuxtAppManifestMeta } from '#app'
+import { defineNuxtPlugin, getAppManifest, onNuxtReady, useRuntimeConfig } from '#app'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  if (import.meta.test) { return }
+
+  let timeout: NodeJS.Timeout
+  const config = useRuntimeConfig()
+
+  async function getLatestManifest () {
```
danielroe, line 11: might be worth exposing this or allow customising timeout?

packages/nuxt/src/core/nitro.ts
```diff
@@ -205,6 +208,80 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+            prerenderedRoutes.add(url)
+          }
+        }
+      }
+
+      const manifest = {
+        id: buildId,
+        timestamp: buildTimestamp,
+        matcher: exportMatcher(routeRulesMatcher),
+        prerendered: nuxt.options.dev ? [] : [...prerenderedRoutes]
```
danielroe, line 273: push `/**` as a default route rule when prerendering site

packages/nuxt/src/app/plugins/payload.client.ts
```diff
@@ -26,5 +18,17 @@ export default defineNuxtPlugin({
...
+
+    onNuxtReady(() => {
+      // Load payload into cache
+      nuxtApp.hooks.hook('link:prefetch', async (url) => {
+        if (!parseURL(url).protocol) {
+          await loadPayload(url)
+        }
+      })
+      if (navigator.connection?.effectiveType !== 'slow-2g') {
+        setTimeout(getAppManifest, 1000)
```
harlan-zw, line 30: Is this 1s arbitrary? 

If so, there is a specific threshold you need to exceed for the script not to be considered blocking (different for devices), if this isn't passing that, maybe it's worth just loading immediately on Nuxt ready?

harlan-zw:
Looking good :+1: Seems quite powerful, I like the hook for end users to implement how they like.

Is it worth solving the tree-shaking route rules for the server/client before merging this to minimise the manifest payload size?

danielroe:
Merging to test on next patch. Please do follow up on comments - I think they still need to be addressed.

lwpinion:
I am excited that you completed this feature!
However, I do have a question:
Is there a way to actually generate these routes so that Vue Router knows they exist? Or if not currently, do you have any plans to add this? As it stands, I have these redirects in place with Route Rules and they do work clientside, but I get warnings in the console when I have a NuxtLink pointing to one of them.

lwpinion:
> @lwpinion Could you open an issue with a reproduction? 🙏

Done!
https://github.com/nuxt/nuxt/issues/24112

blowsie:
`appManifest` is not working with `NUXT_APP_BASE_URL`

Reproduction:
```
NUXT_APP_BASE_URL=/Company/Project/ nuxi build --preset github_pages
```

js-writer:
Because of `<buildId>.json` there is 404 error on IIS_node server because genereted web.config doesn't have json mime type in it.
I already created issue in nitro repo: https://github.com/nitrojs/nitro/issues/2999

PR #21656, opened 2023-06-20
fix(nuxt): stop indicator when not changing page component

### LINKED ISSUE #19528: `<NuxtLoadingIndicator>` is not taking middleware into account
### Environment

Nuxi 3.2.3

RootDir: /home/projects/github-dsgmts
Nuxt project info: 

- Operating System: `Linux`
- Nuxt Version:     `3.2.3`
- Nitro Version:    `2.2.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Here's a minimal reproduction: https://stackblitz.com/edit/github-dsgmts

### Describe the bug

Clicking on the About links triggers the `about` middleware, which waits for 2 seconds before resolving. This 2 second-delay is not reflected on the `<NuxtLoadingIndicator>`, which only shows and hide after the middleware's delay.

### Additional context

# Current "fix"

I created a global middleware triggering a custom hook:

```ts
// middleware/before.global.ts
export default defineNuxtRouteMiddleware(async (to, from) => {
    await useNuxtApp().callHook('custom:before-middleware')
})

```

I then copied https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/components/nuxt-loading-indicator.ts and simply replaced `nuxtApp.hook('page:start', indicator.start)` with `nuxtApp.hook('custom:before-middleware', indicator.start)`.

TypeScript is not happy with this because `TS2345: Argument of type '"custom:before-middleware"' is not assignable to parameter of type 'HookKeys<RuntimeNuxtHooks>'.` but it works as expected (and it does not break pages without middleware).

# My suggestion

Maybe Nuxt could call a new hook before a page is loaded (and before a potentially middleware is invoked)?  
If so, this hook could be used by `nuxt-loading-indicator.ts` in place of `page:start`?

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19528#issuecomment-1597800300

### ❓ Type of change

### 📚 Description

When disabling loading indicator, we _also_ need to disable if after middleware there is no change in page components but there is a change in route.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-loading-indicator.ts
+++ b/packages/nuxt/src/app/components/nuxt-loading-indicator.ts
@@ -38,7 +38,7 @@ export default defineComponent({
 
     globalMiddleware.unshift(indicator.start)
     router.beforeResolve((to, from) => {
-      if (to === from) {
+      if (to === from || to.matched.every((comp, index) => comp.components && comp.components?.default === from.matched[index]?.components?.default)) {
         indicator.finish()
       }
     })


danielroe:
is the issue that the navigation resolves too quickly to display it? Try setting `:throttle="0"`.

c-schwan:
> is the issue that the navigation resolves too quickly to display it? Try setting `:throttle="0"`.

thx this resolve the issue, perfect

emaia:
> is the issue that the navigation resolves too quickly to display it? Try setting `:throttle="0"`.

@danielroe This seems not working properly. Check this example: https://stackblitz.com/edit/nuxt-starter-3ab9sw?file=app.vue

Whitout `:throttle="0"`, navigate to "blog" and then to page 2. Loading indication works. But now if you navigate to page 1, 3 or 4 it not works. This pages are in the same component `[page].vue` but with different "page" param. Slugs `/blog/page/1`, `/blog/page/2`...

If you add `:throttle="0"` the indicator appears but with a weird speed... starting/stopping quickly.

I think it's necessary to do an extra check... maybe something like:

```js
if (to === from || to.matched.every((comp, index) => {
  return comp.components && comp.components?.default === from.matched[index]?.components?.default && 
Object.values(to.params) === Object.values(from.params)
})) {
  indicator.finish();
}
```

This works for my example.

mercs600:
I totally agree with @emaia, it does not work for [...slug].vue pages where we fetch new content from API based on slug.

PR #21659, opened 2023-06-20
fix(nuxt): support custom route name meta with `typedPages`

### LINKED ISSUE #21637: [TYPED ROUTER] - Custom route name through definePageMeta are ignored on generated routes types
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-ufaubc?file=test.ts

### Describe the bug

Using the new experimental typed router,  when overriding a route name through `definePageMeta`, the new route name should be taken into account when generating routes types at build-time.

Example : 

```ts
// ROUTE: index.vue
definePageMeta({
  name: 'home'
})
```

Expected generated route name type : `home`
Currently generated : `index`

### Additional context

_No response_

### Logs

_No response_

ChronicStone:
### 🔗 Linked issue

#21637 

### ❓ Type of change

### 📚 Description

Resolves #21637 by parsing at build-time definePageMeta on routes & extracting custom route name

### 📝 Checklist


packages/nuxt/src/pages/utils.ts
```diff
@@ -63,6 +64,12 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
       children: []
     }
 
+    const fileContent = fs.readFileSync(resolve(pagesDir, file), 'utf-8')
```
danielroe, line 67: We should also support reading from virtual fs (accessible from `nuxt.vfs`).

packages/nuxt/src/pages/utils.ts
```diff
@@ -94,6 +101,12 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
   return prepareRoutes(routes)
 }
 
+function getRouteName (file: string) {
+  const regExp = /definePageMeta\(\s*{[^}]*name:\s*(["'`])((?:[^"'`\\]|\\.)+?)\1/g
+  const match = regExp.exec(file)
```
danielroe, line 106: slightly better perf to hoist regexp out of function, and I think also we don't want it to be global.

```suggestion
const PAGE_META_NAME_RE = /definePageMeta\(\s*{[^}]*name:\s*(["'`])((?:[^"'`\\]|\\.)+?)\1/
function getRouteName (file: string) {
  const match = PAGE_META_NAME_RE.exec(file)
```

danielroe: Having said that, I do have questions about whether this might yield false positives - I'll need to do some testing

packages/nuxt/src/pages/utils.ts
ChronicStone: Not 100% sure I got it right but this has been added as well

packages/nuxt/src/pages/utils.ts
```diff
@@ -94,6 +101,12 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
   return prepareRoutes(routes)
 }
 
+const PAGE_META_NAME_RE = /definePageMeta\(\s*{[^}]*name:\s*(["'`])((?:[^"'`\\]|\\.)+?)\1/
```
danielroe, line 104: I think this is too permissive.

For example, this matches:

```ts
definePageMeta({
 something: () => ({ name: 'test' })
})
```

I think we probably would need to parse the file.

ChronicStone:
Following @danielroe feedback, the route name resolver now uses estree parser to extract value from AST

ChronicStone:
Updated with your latest advice @danielroe. Now AST parsing happens only if definePageMeta is defined, and happen only on the definePageMeta portion of code to minimize AST parsing.

danielroe:
Would you mind taking a look at the failing CI?

ChronicStone:
@danielroe Yep, I planned to have a look tomorrow

ChronicStone:
@danielroe  Failing CI issue solved, unit tests now work properly with new implementation of `generateRoutesFromFiles`. 

Also new unit test implemented to cover route name override

ChronicStone:
@danielroe Sorry to mention, I know it isn't ideal, but the new error on CI seems to be related to commits from main being merged to local branch.

Considering the mature of the issue, I'm a bit confused on how I sould proceed to fix this.

`packages/schema/src/types/config.ts:141`

image

danielroe:
The CI error is because you haven't added all the dependencies you're using to the package.json dependencies, and they are being inlined in the package output.

danielroe (approved):

PR #21665, opened 2023-06-20
fix(nuxt): guard `res.end()` calls with check if event is handled

danielroe:
### 🔗 Linked issue

https://github.com/unjs/nitro/pull/1326#issuecomment-1599673341

### ❓ Type of change

### 📚 Description

Largely Nuxt doesn't directly call `res.end()`, preferring to use utilities from `h3` and `nitropack`, but we were doing so within the error handler. (Though likely we can also migrate from these...)

This PR checks after the final async call before any sync setting of headers/`res.end()` whether the response has already been made.

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -6,6 +6,7 @@ import { useNitroApp, useRuntimeConfig } from '#internal/nitro'
 import { isJsonRequest, normalizeError } from '#internal/nitro/utils'
 
 export default <NitroErrorHandler> async function errorhandler (error: H3Error, event) {
+  if (event.handled) { return }
```
pi0, line 9: Since we might have hooks and logs, this check around `res.end` is probably enough and can be removed from here

danielroe: happy to move it later when we have hooks - but I think we do need it currently as calling `setResponseHeader` will also throw if the event has already been sent back, right?

pi0: Hmm makes sense to also guard it but it only sets a map value (but not sends it without `res.end/send/sendHeaders)

PR #21705, opened 2023-06-22
feat(nuxt): `usePreviewMode` composable

### LINKED ISSUE #15639: Document successor of preview mode
### Describe the feature

I've not managed to find an equivalent for https://nuxtjs.org/docs/features/live-preview/ in the Nuxt3 codebase and am wondering if there's similar functionality/plans to add them/a successor (potentially in a module)?

### Additional information

### Final checks

logotip4ik:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15639

### ❓ Type of change

### 📚 Description

Adds `usePreviewMode` composable that will force `useAsyncData` and `useFetch` to make new requests even if payload was there

### 📝 Checklist


logotip4ik:
Hey @danielroe, my first time here! Would love any feedback 😀

pi0:
Highlighting notes by @danielroe: https://github.com/nuxt/nuxt/issues/15639#issuecomment-1596891939 (added myself as reviewer to check on nitro side)

logotip4ik:
I don't think that i fully understand what provider aware means and nitro job in preview mode. Could you provide some more details ?

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,65 @@
...
+let isPreviewEnabled: boolean
+let addedAfterNavigationCallback = false
+
+export function usePreviewMode (options?: PreviewOptions) {
+  options = { ...previewDefaultOptions, ...(options || {}) }
+
+  const preview = reactive<PreviewState>({
+    enabled: false,
+    token: null
+  })
```
danielroe, line 32: Is it worth using `useState` or otherwise making this shared between invocations of `usePreviewMode` in different components?

logotip4ik: I think `isPreviewEnabled` on `L23` should handle cross components invocations?, but `useState` would be beneficial for future server integrations.

Will soon refactor it 👍

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,65 @@
...
+  const previewState = useState('_preview-composable', () => reactive<PreviewState>({
+    enabled: false,
+    token: null,
+    addedAfterNavigationCallback: false
+  }))
+
+  if (!previewState.value.enabled) {
+    const previewParam = route.query[previewQueryName]
+
+    previewState.value.enabled = !!previewParam && previewParam === 'true'
```
danielroe, line 37:
```suggestion
    previewState.value.enabled = previewParam === 'true'
```

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,65 @@
...
+    tokenQueryName: 'token',
+    ...(options || {})
+  }
+
+  const router = useRouter()
+  const route = useRoute()
+
+  const previewState = useState('_preview-composable', () => reactive<PreviewState>({
+    enabled: false,
+    token: null,
```
danielroe, line 30: A few ideas:
* instead of defaulting to just `token` which I agree is the most common scenario, what do you think about setting arbitrary state?
* What about an option that ignores query param and allowing user to manually engage/disengage query mode instead (for more complex situations? 
* when `enabled` is manually set to false, we probably want to remove the refreshNuxtData router hook - maybe a computed ref with manual set/get values could be useful here?

logotip4ik: I have a little question, is there a point of something like `shouldEnable` if we give user ability to toggle `on` and `off` preview mode ? I think of flag `shouldAutoDetect` and if it is `true`, then detect `preview` in query (or we could say that `shouldAutoDetect` can be `boolean` or `string`, if it is a string we will match for that string instead), if `false`, let user manually do some detection (in for example `onMounted` hook) and set `enabled` to true ?

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,86 @@
...
+      Object.assign(preview.value.state, newState)
+    }
+  }
+
+  const enabled = previewEnabled
+  const state = preview.value.state as NonNullable<ReturnType<GetPreviewState>>
+
+  return (normalizedOptions.controls
+    ? { enabled, state }
+    : enabled) as Controls extends true ? { enabled: typeof enabled, state: typeof state } : typeof enabled
```
lihbr, line 85: Maybe for ease of refactoring/upgrade and for keeping a similar interface the composable could just return `{ enabled }` instead of `enabled` when controls are disabled, WDYT?

lihbr (approved):
Looking good on my end, will simplify a lot CMS integrations for preview~

Atinux:
Thank you for the PR @logotip4ik 

What is `controls` used for?

logotip4ik:
@Atinux @lihbr

Thank you for the review! I was using the VueUse library as an example and they use `controls` in some composables to expose more controls from the return value (for example [useNow](https://github.com/vueuse/vueuse/blob/main/packages/core/useNow/index.ts#L45)). If you think controls aren't a good fit here, I would happily remove them )

danielroe:
Would it be possible to update the docs? 🙏

danielroe (changes requested):
Would it be possible to update the docs? 🙏

Buffalom:
Hi, just wanted to knock and ask if this will be supported in ISR as well. We are very much looking forward to this.

logotip4ik:
Hi @Buffalom, yes it revalidates on the **client** all `useAsyncData` and `useFetch` even if payload was there.

To clarify what would happen, after the initial request to Vercel (for example), it will send the prerendered page with Nuxt's payload in it. After client-side javascript loads, it will check if preview mode is enabled, and if so, Nuxt will revalidate its payloads.

Buffalom:
Sounds awesome. Thanks for clarifying. Do you have an idea on when this might be ready to 🚢?

MuhammadM1998:
@logotip4ik Maybe you need to re-request review? It's in the 'request changes' now

H2XaO

logotip4ik:
Double checked, re-request review button is showing only for lihbr review, and Atinux, pi0, Daniel showing as pending

packages/nuxt/src/app/composables/preview.ts
danielroe: I agree, this makes sense to me.

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,86 @@
...
+let shouldEnablePreviewMode: ShouldEnableFunc = () => {
+  const route = useRoute()
+  const previewQueryName = 'preview'
+
+  return route.query[previewQueryName] === 'true'
+}
+
+let unregisterRefreshHook: () => any
+const _previewEnabled = ref(false)
+const previewEnabled = computed({
```
danielroe, line 22: We should handle the situation `import.meta.server`, or this will otherwise lead to a memory leak/XRSP.

logotip4ik: What do you mean by ?

> situation `import.meta.server`

Would `previewEnabled` be shared between requests ? Then we need to create little stub for server ?

danielroe: yes, it's very important that even if users call `usePreviewMode` on the server, that it should not be shared between requests

logotip4ik: Refactored composable to rely only on `useState` values, so it should not share data between requests, and it passes previous tests 🤗

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (newState !== preview.value.state) {
+        Object.assign(preview.value.state, newState)
+      }
+
+      if (import.meta.client && !unregisterRefreshHook) {
+        refreshNuxtData()
+
+        unregisterRefreshHook = useRouter().afterEach((() => refreshNuxtData()))
+      }
+    } else if (!value && unregisterRefreshHook) {
```
danielroe, line 46:
```suggestion
    } else if (unregisterRefreshHook) {
```

packages/nuxt/src/app/composables/preview.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface Preview {
+  enabled: boolean
+  state: Record<any, unknown>
+}
+
+type EnteredState = Record<any, unknown> | null | undefined | void
+
+let unregisterRefreshHook: (() => any) | undefined
+
```
luc122c, line 15: If this lands in 3.10.0, please add:
```suggestion

/** @since 3.10.0 */
```

luc122c (changes requested):

logotip4ik:
CI failed because of Lychee error on: https://stackoverflow.com/help/minimal-reproducible-example and other two tests failed with not related changes

danielroe (approved):
Thank you! ❤️ 

And apologies for how long it's taken to get this in 🙏

logotip4ik:
No worries. I am glad to participate in Nuxt community ❤️

algora-pbc:
@logotip4ik: Your claim has been rewarded! Please [complete your Algora onboarding](https://console.algora.io/onboarding/solver) to collect the bounty.

PR #21740, opened 2023-06-24
fix(nuxt): proxy headers to islands + returned prerender hints

danielroe:
### ❓ Type of change

### 📚 Description

When using server components, currently they are unable to hint to Nitro to prerender resources. For example, a `<nuxt-img>` rendered in a server component will not output a statically generated image at the end of the build.

I think it makes sense to proxy both event context (e.g. auth headers) _to_ islands as well as to set any headers on the response. (Honestly, I thought event fetch did the latter piece anyway, but it seems I am mistaken: cc: @pi0.)

### 📝 Checklist


huang-julien (approved):
I even thought that all headers were and should be proxied

PR #21747, opened 2023-06-24
feat(nuxt): auto-install `@nuxt/webpack-builder` when needed

### LINKED ISSUE #21746: Error running nuxt locally when the builder webpack is enabled in config
### Environment

- Operating System: Linux
- Node Version:     v16.19.1
- Nuxt Version:     3.6.0
- Nitro Version:    2.5.1
- Package Manager:  npm@8.19.3
- Builder:          webpack
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

To reproduce the error we need to add the following flag in nuxt.config:

`builder: "webpack"`

**StackBlitz:** [Click here](https://stackblitz.com/edit/github-ktzztp)

### Describe the bug

When the builder option it's changed in nuxt.config, the project is not running locally. 

`[1:55:07 PM]  ERROR  Cannot start nuxt:  Cannot destructure property 'bundle' of '(intermediate value)(intermediate value)(intermediate value)' as it is undefined.`

### Additional context

_No response_

### Logs

_No response_

jeanpierrecarvalho:
### 🔗 Linked issue

resolves #21746

### ❓ Type of change

### 📚 Description

This solves problems related to the webpack builder when the builder options its set to webpack but the module it's not installed. 

### 📝 Checklist


packages/webpack/package.json
```diff
@@ -8,6 +8,7 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
+      "require": "./module.cjs",
```
danielroe, line 11: I suspect this is only required within the monorepo in stub mode rather than when built. Would you try removing the change to the packaging and running `pnpm build` before testing it?

jeanpierrecarvalho: Hey @danielroe, I tried doing a pnpm build before adding the changes in webpack, but unfortunately, I'm getting the error message that says "No 'exports' main defined." after the build.

danielroe:
We'll aim to get this out for the 3.7 release, next month.

@antfu It looks like this is failing within `isPackageExists`. `resolvePackageJSON` from `pkg-json` seems to work fine. Maybe worth updating the logic or using a `mlly` utility there?

antfu:
I don't mind switching at all. But wonder what cause the failing. Vitest is using it heavily but does seems to have the issue

danielroe:
It is using a synchronous CJS-compatible utility only: https://github.com/antfu/local-pkg/blob/main/index.mjs#L70-L84

antfu:
Oh yeah, in that case, I guess I might need to rewrite it to make all functions async. For nuxt I think we could directly use unjs packages and not been blocked by it.

danielroe (approved):

danielroe:
Thank you - this is great.

PR #21815, opened 2023-06-27
fix(nuxt): only redirect if path is not the same as initial url

### LINKED ISSUE #20491: DynamicRoute from router.beforeEach in plugin:  changed behaviour
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.4.1`
- Nitro Version:    `2.3.3`
- Package Manager:  `npm@9.5.1`
- Builder:          `vite`
- User Config:      `modules`, `runtimeConfig`, `sourcemap`, `devtools`
- Runtime Modules:  `@storyblok/nuxt@5.4.0`, `@nuxtjs/tailwindcss@6.6.6`, `@nuxtjs/robots@3.0.0`
- Build Modules:    `-`

### Reproduction

Working in v3.4.0
https://stackblitz.com/edit/github-1cetba-gmp4ez?file=plugins%2FdynamicRouting.ts

Not working in v3.4.2
https://stackblitz.com/edit/github-1cetba-sn9iac?file=plugins%2FdynamicRouting.ts

### Describe the bug

I have created a plugin that on router.beforeEach checks if the current path matches a path in en external API, and if so, creates a dynamicRoute for that path with a component matching the API item type, in the example, a Product.

This works in Nuxt 3.4.0 but have stopped working in Nuxt >= 3.4.1. In the newer version, the vue router seems to forget the route...

### Additional context

This is the idea behind the functionality
https://router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes-inside-navigation-guards

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20491

### ❓ Type of change

### 📚 Description

A regression was introduced in https://github.com/nuxt/nuxt/pull/20247 - namely, that it became impossible to add a new route and then redirect to it within the same request - this would become a redirection.

This adds a check that enables adding dynamic routes in this kind of way - we won't send a redirect to the same URL as the initial request.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/plugins/router.ts
+++ b/packages/nuxt/src/pages/runtime/plugins/router.ts
@@ -198,7 +198,7 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
           fatal: false,
           statusMessage: `Page not found: ${to.fullPath}`
         })))
-      } else if (process.server && to.redirectedFrom) {
+      } else if (process.server && to.redirectedFrom && to.fullPath !== initialURL) {
         await nuxtApp.runWithContext(() => navigateTo(to.fullPath || '/'))
       }
     })
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -117,6 +117,11 @@ describe('pages', () => {
     expect(headers.get('location')).toEqual('/')
   })
 
+  it('allows routes to be added dynamically', async () => {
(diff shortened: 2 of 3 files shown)


TheAlexLichter (approved):

manniL:
Good catch!

CernyMatej:
Thanks!

PR #21817, opened 2023-06-27
fix(nuxt): improve default prerender behaviour

### LINKED ISSUE #14676: fail nitro build if there are errors prerendering pages
### Environment

- Operating System: `Linux`
- Node Version:     `v18.3.0`
- Nuxt Version:     `3.0.0-rc.8`
- Package Manager:  `npm@8.11.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-6hv1ba?file=package.json,app.vue

* take a look at `package.json` & `app.vue`
* execute `npm run generate`
* run `echo $?` to get the exit code (not working inside stackblitz)
  * this should return non zero but it returns 0

### Describe the bug

The `--fail-on-error` is ignored by the `nuxi generate` command so the generate command is not backwards compatible to nuxt2.
The lack of considering this feature can lead to broken website deployments because it is not possible to differentiate between a succeed and failed generation. 

### Additional context

* https://github.com/nuxt/nuxt.js/issues/4991
* https://nuxtjs.org/docs/get-started/commands/#fail-on-error

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14676

### ❓ Type of change

### 📚 Description

This adds default values for new nitro features (https://github.com/unjs/nitro/pull/1294 and https://github.com/unjs/nitro/pull/1223) - supporting default 10 page concurrency of prerendering and failing a build if any paths don't prerender successfully. (Users can add any paths they _intend_ to fail to `nitro.prerender.ignore`, or disable this behaviour if they want.)

### 📝 Checklist


pi0 (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -131,6 +131,8 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
         .map(dir => ({ dir }))
     ],
     prerender: {
+      failOnError: true,
+      concurrency: 10,
```
pi0, line 135: We can infer this from user CPU count (also using a power of 2 number makes sure less context switches)

PR #21823, opened 2023-06-27
fix(nuxt): preserve old vnode when leaving nested route

### LINKED ISSUE #20798: if leaving nested route, child page is unmounted before suspense resolves
[stackblitz.com/edit/github-f4gpfk-qhgrfy?file=pages/test.vue](https://stackblitz.com/edit/github-f4gpfk-qhgrfy?file=pages/test.vue)

> https://github.com/nuxt/nuxt/issues/13814#issuecomment-1544828371

### LINKED ISSUE #21417: Child page setup called without <nuxt-page> on parent page
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.3
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-9hmdww

### Describe the bug

During client-side navigation to `/page` when there are both `page.vue` and `page/index.vue`, the `setup()` function of `page/index.vue` is:
1) called when it shouldn't (if `page.vue` doesn't have `<nuxt-page>`);
2) called twice (if `page.vue` has `<nuxt-page>`).

* The problem does not repeat on hard refresh (SSR), it only happens during client-side navigation.
* To trigger the bug, the originating page must also be a nested page (`foo.vue` + `foo/whatever.vue`).
* To trigger the bug, `page.vue` should have `setup()` function, possibly doing nothing.

On the reproduction URL there are two simple demos:

- Demo 1 is the simplest minimal reproduction, but it needs you to open devtools console. You will see that `setup` of the child page (`1/index.vue`) is called when it should not.
- Demo 2 shows the bug even without devtools. It prints how many times the child `setup` was called, you can see that it is called twice.

### Additional context

This is a regression introduced in Nuxt 3.5. The same code works correctly in Nuxt 3.4.3.

_**UPDATE:** This is not entirely true. The reproduction above crashes in Nuxt 3.4, too. [See below](https://github.com/nuxt/nuxt/issues/21417#issuecomment-1580724743) for the actual regression reproduction._

### Logs

_No response_

### LINKED ISSUE #21798: useFetch called twice after layout switching
### Environment

Nuxt Version: 3.6.1

### Reproduction

https://stackblitz.com/edit/nuxt-j8hefc

### Describe the bug

useFetch is called twice after the layout changed.

pages/index.vue (called useFetch) -> pages/other (different layout) -> pages/index.vue (called twice useFetch)

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20798
resolves https://github.com/nuxt/nuxt/issues/21417
resolves https://github.com/nuxt/nuxt/issues/21798

### ❓ Type of change

### 📚 Description

Thjs handles a couple of extra suspense cases:

- navigating into a nested route
- navigating out of a nested route
- navigating between nested routes where the parent rerenders the child

I think some of these really should be handled by the router, but for now this should make things more synchronous and reduce flickers/flashes/out-of-date state between route/layout transitions.

### 📝 Checklist


danielroe:
@antfu Let me know if you can think of any more tests to add.

packages/nuxt/src/app/components/injections.ts
```diff
@@ -0,0 +1,11 @@
+import type { InjectionKey } from 'vue'
+import type { RouteLocationNormalizedLoaded } from 'vue-router'
+import { _wrapIf } from './utils'
```
HigherOrderLogic, line 3: This is unused in the file.

HigherOrderLogic, line 3:
```suggestion
```

This is unused in the file.

packages/nuxt/src/app/components/route-provider.ts
```diff
@@ -0,0 +1,57 @@
...
+    // eslint-disable-next-line vue/no-setup-props-destructure
+    const previousRoute = props.route
+
+    // Provide a reactive route within the page
+    const route = {} as RouteLocation
+    for (const key in props.route) {
+      (route as any)[key] = computed(() => previousKey === props.renderKey ? props.route[key as keyof RouteLocationNormalizedLoaded] : previousRoute[key as keyof RouteLocationNormalizedLoaded])
+    }
+
+    provide(PageRouteSymbol, reactive(route))
```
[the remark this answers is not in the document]
danielroe: Let's do it 👍

marr:
I'm not sure how to reproduce in a minimal way, but this PR appears to have caused my app to stop working properly.
My app has a login page that redirects to the /login page when not authorized. After logging in (form), I have a plugin that handles redirection. With https://github.com/nuxt/nuxt/commit/d0dde6426fe6ae58ec1ea94e47c66f6e0e2a4d63 in place, the redirect never happens. If I checkout the previous commit, things work fine.
Does that indicate that I might be doing something wrong? Any ideas would be helpful.

PR #21837, opened 2023-06-28
fix(nuxt, vite): do not use cjs utils to resolve/alias vue

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/14146

### ❓ Type of change

### 📚 Description

Vite and Nitro correctly resolve vue dependencies, and it is far better to defer resolution to the bundlers than override it ourselves, particularly with support for non-hoist:

You can test with this plugin:

<details>
<summary>Details</summary>

```ts
const logger = {
  name: 'log',
  writeBundle (_options, bundle) {
    for (const file in bundle) {
      const chunk = bundle[file]
      if (chunk.type === 'asset') { continue }
      const resolvedVueFiles = chunk?.moduleIds.filter(i => i.includes('vue'))
      if (resolvedVueFiles?.length) {
        console.log(resolvedVueFiles)
      }
    }
  }
}

  experimental: {
    // you can try with and without
    externalVue: false
  },
  nitro: {
    rollupConfig: {
      plugins: [logger]
    }
  },
  vite: {
    plugins: [logger]
  }
})
```

</details>

(In my testing, before this PR, with `externalVue: false` vue was actually being bundled twice in nitro bundle.)

### 📝 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ dist
 .vercel
 .netlify
 .output
+.output-*
 .gen
 
 # Junit reports
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -2,7 +2,7 @@ import { existsSync, promises as fsp, readFileSync } from 'node:fs'
 import { join, relative, resolve } from 'pathe'
 import { build, copyPublicAssets, createDevServer, createNitro, prepare, prerender, scanHandlers, writeTypes } from 'nitropack'
 import type { Nitro, NitroConfig } from 'nitropack'
-import { logger, resolvePath } from '@nuxt/kit'
+import { logger } from '@nuxt/kit'
 import escapeRE from 'escape-string-regexp'
 import { defu } from 'defu'
 import fsExtra from 'fs-extra'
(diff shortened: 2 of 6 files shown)


pi0 (approved):
LGTM if you have already tested it against latest nitro. 

We had to get rid of this workaround sooner or later.

Also linking to https://github.com/unjs/nitro/issues/1118 which might be needed to dedup external vue

danielroe:
I've tested it in the framework repo here. Are there module resolution changes coming upstream from nitro?

sei-jdshimkoski:
I think this might have broken this particular Nuxt Config setting (Nuxt 3.6.2 with Nitro 2.5.2):

```
vite: {
    build: {
      rollupOptions: {
        output: {
          manualChunks(_id) {
            return 'index'
          }
        },
      },
    },
  },
```

I now get a bunch of `<BLAH> cannot be included in manualChunks because it is resolved as an external module by the "external" option or plugins.`

danielroe:
@sei-jdshimkoski Maybe you you could open an issue with reproduction? I don't have an issue with that code in a minimal sandbox.

wtagain:
> I think this might have broken this particular Nuxt Config setting (Nuxt 3.6.2 with Nitro 2.5.2):
> 
> ```
> vite: {
>     build: {
>       rollupOptions: {
>         output: {
>           manualChunks(_id) {
>             return 'index'
>           }
>         },
>       },
>     },
>   },
> ```
> 
> I now get a bunch of `<BLAH> cannot be included in manualChunks because it is resolved as an external module by the "external" option or plugins.`

Same question. Did you fix it?

sei-jdshimkoski:
Sorry, was out of town and away from my computer. I'll try to open an issue with reproduction today.

sei-jdshimkoski:
Ticket is here: https://github.com/nuxt/nuxt/issues/22127

PR #21847, opened 2023-06-28
fix(schema): warn if user provides `vite.publicDir`

### LINKED ISSUE #21831: how to copy public folder when change dir.public
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.1
- Nitro Version:    2.5.1
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      vite, dir
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-kf4vv3?file=nuxt.config.ts,app.vue

### Describe the bug

When change `dir.public` and `vite.publicDir`，it can work at dev, 
But `public` folder not copy at production, maybe just a config error

### Additional context

Step

1. `npm run dev` (can see image)
2. `npm run build`
3. `node .output/server/index.mjs` (can not see image)

### Logs

_No response_

nozomuikuta:
### 🔗 Linked issue

Resolves #21831

### ❓ Type of change

### 📚 Description

This PR is based on https://github.com/nuxt/nuxt/issues/21831#issuecomment-1611180881

I hope I'm doing it in right way and as @danielroe intended. 🙏 

References:

- https://vitejs.dev/guide/assets.html#the-public-directory

### 📝 Checklist


packages/schema/src/config/vite.ts
```diff
@@ -30,6 +30,7 @@ export default defineUntypedSchema({
       extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue']
     },
     publicDir: {
+      $schema: { deprecated: 'use `dir.public` option instead' },
```
NozomuIkuta, line 33: Just to confirm, is this used for documentation auto-generation and no influence of coding actual application code?

NozomuIkuta: By the way, the message is quoted from `dir.static` option.

packages/vite/src/vite.ts
```diff
@@ -173,6 +173,12 @@ export async function bundle (nuxt: Nuxt) {
     })
   }
 
+  nuxt.hook('vite:extendConfig', (config) => {
+    if (config.publicDir) {
+      consola.warn('Using `options.vite.publicDir` option is not supported together with Nuxt. Use `options.dir.public` instead. You can read more in `https://nuxt.com/docs/api/configuration/nuxt-config#public`.')
+    }
+  })
```
NozomuIkuta, line 180: `if (config.publicDir) {...}` checks that `config.publicDir !== false && config.publicDir !== ''`, both of which prevent public assets from being copied.

This warning message is shown twice in Terminal, because it's run for client build and server build. Should I show it only in either of them by checking `isClient` or `isServer` props?

packages/schema/src/config/vite.ts
NozomuIkuta: Although I deprecate it at docs level, developers still can specify `vite.publicDir` option in Nuxt config file, because Nuxt accepts `ViteConfig` as `vite` option that  extends/inherits all the properties from Vite's `UserConfig`.

danielroe: yes, that's tricky. Setting `$schema` here won't have any effect, as you say, because it's being shadowed by `ViteConfig`. So instead we can set the deprecation on `ViteConfig` directly.

packages/vite/src/vite.ts
danielroe: I've moved the warning to schema so it will only warn once. what do you think?

danielroe (approved):

NozomuIkuta:
@danielroe 

Oh, sorry I couldn't reply to your comments.
Thank you for refactoring my PR and merging it!

Issue #21867, opened 2023-06-30
Nuxt documentation about what the current LTS version is


default-kaas:
### Describe the feature

It is unclear what the current LTS version is for Nuxt.
For example which Nuxt version is by the team currently the most recommended version to use?
Which version of Nuxt will have long time stable support?
And what is the current road map for LTS versions?

When creating a new Nuxt project you will get to use Nuxt V3.4.2,
But the latest version of Nuxt is V3.6.1.

Which one of the two would the Nuxt team recommend that you would use.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
When creating a new Nuxt project it should be 3.6.1: https://github.com/nuxt/starter/blob/v3/package.json#L14

default-kaas:
> When creating a new Nuxt project it should be 3.6.1: https://github.com/nuxt/starter/blob/v3/package.json#L14

Sorry that is correct, a new Project will use nuxt 3.6.1.

I am new to contributing to a major open source project.
Can I add a setup for the information as an new page about the current LTS version?
Or add it as additional information to an existing page?
And that some from the core team either fills the information in or checks if it is correct.

I would like to add the following information:

- [ ] Current LTS version
  - Also make it a clickable link to release on github such as this: https://github.com/nuxt/nuxt/releases/tag/v3.6.1
- [ ] Next LTS version
  - Also make it a clickable link to release if it is available on github such as this: https://github.com/nuxt/nuxt/releases/tag/v3.6.1 
- [ ] A table with all (Max 6) LTS versions
  - [ ] 3 LTS versions from the past (if available)
  - [ ] Current LTS version
  - [ ] 2 future LTS versions (if available)
  - [ ] The table will have the following columns
    - [ ] Version number (Number) 
      -  Also make it a clickable link to release if it is available on github such as this: https://github.com/nuxt/nuxt/releases/tag/v3.6.1 
    - [ ] Status (String)
    - [ ] Codename (String|optional)
    - [ ] Initial Relase (Date)
    - [ ] Active LTS Start (Date)
    - [ ] Maintenance LTS Start (Date)
    - [ ] End Of Life (Date)

[...]

PR #21893, opened 2023-07-02
docs: add support details and update roadmap (#21867)

default-kaas:
### 🔗 Linked issue

#21867 

### ❓ Type of change

#### Describe your changes in detail ####

I have added and LTS section to the roadmap page. Which containts information about the current and future LTS versions. It also includes a table with past, current and future LTS versions. The table has the following data points per version:

- version - number  && link to that version  release on github
- status - string
- initial  release - date
- active lts start - date
- maintenance lts start - date
- end of life - date
- vue version - number
- minimal node version - number

It also has an legend table that describes each status.

#### Why is this change required? What problem does it solve? 

An important part for adopting a new framework for your code base is to know how long a certain version of the framework will be maintained.  It is also valuable information for planning purposes so that the developed teams can plan accordingly. 

The LTS version will help when trying to adopt a technology over a wide range of applications within your company. So that you can use the accumulated migration knowledge for migrating the other applications within your company.

#### Call to action

I am not part of the Nuxt core team, so keep in mind that all the data in the tables are only place holder data to give you an impression on how it could look like. It would be nice if members from the Nuxt core team could look at this table an fill it with legitimate and accurate data.

#### Preview

[Preview](https://nuxt.com/docs/community/roadmap#long-time-support-lts-versions)

### 📝 Checklist


NozomuIkuta:
@default-kaas 

JFYI, when you create a PR, you can link it to issue(s) by using GitHub keywords.
By linking them, the issue will be closed once the PR is merged.

See [GitHub docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for more detail.

docs/5.community/6.roadmap.md
```diff
@@ -58,3 +58,32 @@ Auth           | Planned             | 3.x          | `nuxt/auth` to be announce
...
+---------------------------------------------------------|-------------|-----------------|------------------|-----------------------|--------------|-------------|---------------------
+[V3.2](https://github.com/nuxt/nuxt/releases/tag/v3.2.3) | Unsupported | 02-09-2022      | 02-28-2022       | 03-14-2022            | 04-02-2022   | V3.2        | V16
+[V3.3](https://github.com/nuxt/nuxt/releases/tag/v3.3.3) | Maintenance | 03-14-2022      | 04-02-2022       | 04-11-2022            | 04-28-2022   | V3.2        | V16
+[V3.4](https://github.com/nuxt/nuxt/releases/tag/v3.4.3) | Maintenance | 04-11-2022      | 04-28-2022       | 05-16-2022            | 06-06-2022   | V3.2        | V16
+[V3.5](https://github.com/nuxt/nuxt/releases/tag/v3.5.3) | Current     | 05-16-2022      | 06-06-2022       | 06-23-2022            | 07-14-2022   | V3.3        | V16
+[V3.6](https://github.com/nuxt/nuxt/releases/tag/v3.6.1) | Active      | 06-23-2022      | 07-14-2022       | 07-28-2022            | 08-09-2022   | V3.3        | V16
+V3.7                                                     | Scheduled   | 07-28-2022      | 08-07-2022       | 08-21-2022            | 09-04-2022   | ~           | V18
```
NozomuIkuta, line 79: As a non-native English speaker, I prefer European date format (i.e. DD-MM-YYYY), because American date format is confusing when DD part is less than 13:

I can determine `04-12-2023` is `April 12th 2023` or `4th December 2023` only by checking other date strings in the same page.

Using word instead of numbers for months would be an alternative: `4th Dec. 2023` or `Dec. 4th 2023`.

Please ignore this comment if Nuxt docs uses American style date format in other places already and you just followed it.

image
_Image by https://iso.mit.edu/americanisms/date-format-in-the-united-states/_

danielroe: What about using YYYY-MM-DD? - I think this should be equally understandable for both Americans and rest of world.

NozomuIkuta: That is also fine. 👍 

My points is that it is beneficial for more people to use more intuitive (less confusing) date format and to use it consistently in Nuxt documentation (i.e. using other than American style date format).

danielroe: Completely agree 👍

default-kaas: The next two days I will be a bit busy, but I will add the changes on Wednesday (Berlin time).

Jannchie, line 79: I noticed that there has been no update here for two weeks now, and I'm submitting a suggestion change to be able to close this pr.

```suggestion
Version number                                           | Status      | Initial release | Active LTS Start | Maintenance LTS Start | End Of Life  | Vue version | Minimal node version
---------------------------------------------------------|-------------|-----------------|------------------|-----------------------|--------------|-------------|---------------------
[V3.2](https://github.com/nuxt/nuxt/releases/tag/v3.2.3) | Unsupported | 2022-02-09      | 2022-02-28       | 2022-03-14            | 2022-04-02   | V3.2        | V16
[V3.3](https://github.com/nuxt/nuxt/releases/tag/v3.3.3) | Maintenance | 2022-03-14      | 2022-04-02       | 2022-04-11            | 2022-04-28   | V3.2        | V16
[V3.4](https://github.com/nuxt/nuxt/releases/tag/v3.4.3) | Maintenance | 2022-04-11      | 2022-04-28       | 2022-05-16            | 2022-06-06   | V3.2        | V16
[V3.5](https://github.com/nuxt/nuxt/releases/tag/v3.5.3) | Current     | 2022-05-16      | 2022-06-06       | 2022-06-23            | 2022-07-14   | V3.3        | V16
[V3.6](https://github.com/nuxt/nuxt/releases/tag/v3.6.1) | Active      | 2022-06-23      | 2022-07-14       | 2022-07-28            | 2022-08-09   | V3.3        | V16
V3.7                                                     | Scheduled   | 2022-07-28      | 2022-08-07       | 2022-08-21            | 2022-09-04   | ~           | V18
```

Jannchie: But I find it very easy to document out of date LTS versions here... Maybe there needs to be an automated tool to generate the tables?

Aareksio:
Isn't current _LTS_ nuxt **3**? 

The minor versions are scheduled to be released roughly every month, I do not think it makes sense to designate them as LTS.

default-kaas:
> Isn't current _LTS_ nuxt **3**?
> 
> The minor versions are scheduled to be released roughly every month, I do not think it makes sense to designate them as LTS.

I think a system like that is going to backfire.

I mean why would you otherwise explain why each major framework: node.js, vue.js and laravel make use of such a system?

docs/5.community/6.roadmap.md
default-kaas: I have been quite, bussy the last 3 weeks.
I have have updated the placeholder data.
Hopefully someone from the core team will give it the correct data.

default-kaas: > What about using YYYY-MM-DD? - I think this should be equally understandable for both Americans and rest of world.

Should this be implemented every where?

I try checking the docs repo any use of the following format `\d{2}-\d{2}-\d{4}` in VSCode,
But I did not find any other use of it, so could this be the only place this date format is used?

maximepvrt:
We are in 2023 not in 2022 😅

default-kaas:
> We are in 2023 not in 2022 😅

> Call to action
I am not part of the Nuxt core team, so keep in mind that all the data in the tables are only place holder data to give you an impression on how it could look like. It would be nice if members from the Nuxt core team could look at this table an fill it with legitimate and accurate data.

@maximepvrt  Maybe I am missing something this is what I wrote in the pull request. It is just place holder data, that someone from the core team should up date. I only changed the date format, because people request this.

@danielroe Do you still think this is a valid pull request, or is it nullified?
Because if you go down this route, it makes very little sense to keep this up.

> Isn't current LTS nuxt 3?
>The minor versions are scheduled to be released roughly every month, I do not think it makes sense to designate them as LTS.

But in that case maybe write this down in the docs, because it is not self evident. 

I suspect that there is a valid reason why other frameworks don't go down this route. So it would be nice to write down why this is the best course of action for Nuxt, because in that way you can later on change this policy if the team thinks this type of reasoning is not benefical any more.

Aareksio:
A big part of LTS is the existence of _old_, long time supported version which is expected to be maintained even after next release comes. And perhaps a future version, which is worked on concurrently to the current release. Nuxt has none of that, it's trunk based development with minor releases making the changes public.

Currently `3.7` is the `latest` release, which (as per lack of "Pre-release" tag) is also considered `stable` by the maintainers. There is no observable "long term support" plan for any minor release - previous one gets replaced with new one as soon as it gets released.

Having established that no minor version of `3.x` can attain title of LTS, only alternative is to call the major, `3`, LTS. Or more accurately "future LTS" as technically there is still `2` which is _sometimes supported_. But it doesn't make much sense to me - calling something "LTS" without actual long term support strategy is pointless and misleading.

Therefore current system of versioning, with "latest" being the version you are supposed to use, and keep reasonably up-to-date with, seems sensible. At least until the resources can be allocate to keep old versions maintained.

danielroe:
Apologies for the long time this PR has languished. I think it's a great idea to specify _what_ our LTS policy is and this PR is a  helpful prompt.

I'll edit it directly (so don't close it).

default-kaas:
> A big part of LTS is the existence of _old_, long time supported version which is expected to be maintained even after next release comes. And perhaps a future version, which is worked on concurrently to the current release. Nuxt has none of that, it's trunk based development with minor releases making the changes public.
> 
> Currently `3.7` is the `latest` release, which (as per lack of "Pre-release" tag) is also considered `stable` by the maintainers. There is no observable "long term support" plan for any minor release - previous one gets replaced with new one as soon as it gets released.
> 
> Having established that no minor version of `3.x` can attain title of LTS, only alternative is to call the major, `3`, LTS. Or more accurately "future LTS" as technically there is still `2` which is _sometimes supported_. But it doesn't make much sense to me - calling something "LTS" without actual long term support strategy is pointless and misleading.
> 
> Therefore current system of versioning, with "latest" being the version you are supposed to use, and keep reasonably up-to-date with, seems sensible. At least until the resources can be allocate to keep old versions maintained.

I fine what that, it seems like a sensible policy.

So if the documentation could include an explanation such as this, I personally think that would greatly help developers understanding the direction Nuxt is currently on. And would help with what I wrote down earlier.

>But in that case maybe write this down in the docs, because it is not self evident.
>I suspect that there is a valid reason why other frameworks don't go down this route. So it would be nice to write down why this is the best course of action for Nuxt, because in that way you can later on change this policy if the team thinks this type of reasoning is not benefical any more.



---

> I think it's a great idea to specify what our LTS policy is and this PR is a helpful prompt.

I am with you with that

pi0 (approved):

atinux (approved):

PR #21904, opened 2023-07-03
fix(nuxt): use `$fetch.raw` in dev client mode for islands

huang-julien:
### 🔗 Linked issue
fix #21740

### ❓ Type of change

### 📚 Description

Hi :wave: 

There's currently a bug that can happen on client-side nav in dev mode (did not test build mode and don't know why this hasn't been detected by the fixtures tests).

bug reproduction:
- run test/fixtures/basic in dev
- go to localhost:3000
- click on the `islands` ``NuxtLink

Sometimes when navigating to a page containing a server component/island, we can have a 503 response by the server. With the native fetch API, this response isn't handled and breaks completly the page.

### reproduction env
- Node Version:     v18.16.1
- Nuxt Version:     3.6.1
- Nitro Version:    2.5.2
- Package Manager:  pnpm@7.9.0
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -11,6 +11,7 @@ import type { NuxtIslandResponse } from '../../core/runtime/nitro/renderer'
 import { getFragmentHTML, getSlotProps } from './utils'
 import { useNuxtApp } from '#app/nuxt'
 import { useRequestEvent } from '#app/composables/ssr'
+import { FetchResponse } from 'ofetch'
 
 const pKey = '_islandPromises'
 const SSR_UID_RE = /nuxt-ssr-component-uid="([^"]*)"/
@@ -42,7 +43,7 @@ export default defineComponent({
     const hashId = computed(() => hash([props.name, props.props, props.context]))
     const instance = getCurrentInstance()!
     const event = useRequestEvent()
-    const eventFetch = process.server ? event.fetch : globalThis.fetch
+    const eventFetch = process.server ? event.fetch : $fetch.raw
     const mounted = ref(false)
     onMounted(() => { mounted.value = true })
 
@@ -81,7 +82,7 @@ export default defineComponent({
         ...props.context,
         props: props.props ? JSON.stringify(props.props) : undefined
       }))
-      const result = await r.json() as NuxtIslandResponse
+      const result = process.server ? await r.json() : (r as FetchResponse<NuxtIslandResponse>)._data
       // TODO: support passing on more headers
       if (process.server && process.env.prerender) {
         const hints = r.headers.get('x-nitro-prerender')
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -81,7 +82,7 @@ export default defineComponent({
         ...props.context,
         props: props.props ? JSON.stringify(props.props) : undefined
       }))
-      const result = await r.json() as NuxtIslandResponse
+      const result = process.server ? await r.json() : (r as FetchResponse<NuxtIslandResponse>)._data
```
danielroe, line 85: can we resolve this issue by just adding a `.catch` block to `r.json()`?

I also can't reproduce the issue on my end so my guess is that it's an issue with nitro's dev server socket + windows.

huang-julien: Yes i already suspected that this was a windows issue. $fetch was solving this by rerunning the request. 
Should we just allow an infinite refetch on 503 instead of a single catch ? It's quite rare but i can have up to 3 failed 503 request

danielroe: Hm. It feels like this is an issue that needs to be resolved in _Nitro_ rather than worked around in this kind of way. That said, I'm happy to merge this as a hotfix if essential.

Have you tried with the latest nitropack version? There was a fix merged that should have improved things on Windows.

huang-julien: you're probably referring to https://github.com/unjs/nitro/pull/1115 🤔  ? on the edge version, i still have the 503 issue (nitropack 2.5.2)

would love to have someone on windows to also confirm this

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -42,7 +43,7 @@ export default defineComponent({
     const hashId = computed(() => hash([props.name, props.props, props.context]))
     const instance = getCurrentInstance()!
     const event = useRequestEvent()
-    const eventFetch = process.server ? event.fetch : globalThis.fetch
+    const eventFetch = process.server ? event.fetch : $fetch.raw
```
danielroe, line 46:
```suggestion
    // TODO: remove use of `$fetch.raw` when nitro 503 issues on windows dev server are resolved
    const eventFetch = process.server ? event.fetch : process.dev ? $fetch.raw : globalThis.fetch
```

packages/nuxt/src/app/components/nuxt-island.ts
danielroe, line 85:
```suggestion
      const result = process.server || !process.dev ? await r.json() : (r as FetchResponse<NuxtIslandResponse>)._data
```

danielroe: yes, exactly. any chance you could see if you can provide a pure nitro reproduction for @pi0?

danielroe (approved):

packages/nuxt/src/app/components/nuxt-island.ts
huang-julien: Sure :+1:

PR #21912, opened 2023-07-03
fix(nuxi): respect nitro ssl cert/key variables

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/21044#issuecomment-1617623539

### ❓ Type of change

### 📚 Description

This respects `NITRO_SSL_CERT` and `NITRO_SSL_KEY` variables when passed together with `--https`.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -59,12 +59,12 @@ export default defineNuxtCommand({
       showURL: false,
       clipboard: args.clipboard,
       open: args.open || args.o,
-      port: args.port || args.p || process.env.NUXT_PORT || config.devServer.port,
-      hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
+      port: args.port || args.p || process.env.NUXT_PORT || process.env.NITRO_PORT || config.devServer.port,
+      hostname: args.host || args.h || process.env.NUXT_HOST || process.env.NITRO_HOST || config.devServer.host,
       https: (args.https !== false && (args.https || config.devServer.https))
         ? {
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/dev.ts
```diff
@@ -63,8 +63,8 @@ export default defineNuxtCommand({
       hostname: args.host || args.h || process.env.NUXT_HOST || config.devServer.host,
       https: (args.https !== false && (args.https || config.devServer.https))
         ? {
-            cert: args['ssl-cert'] || (typeof config.devServer.https !== 'boolean' && config.devServer.https.cert) || undefined,
-            key: args['ssl-key'] || (typeof config.devServer.https !== 'boolean' && config.devServer.https.key) || undefined
+            cert: args['ssl-cert'] || process.env.NITRO_SSL_CERT || (typeof config.devServer.https !== 'boolean' && config.devServer.https.cert) || undefined,
+            key: args['ssl-key'] || process.env.NITRO_SSL_KEY || (typeof config.devServer.https !== 'boolean' && config.devServer.https.key) || undefined
```
pi0, line 67: We might in addition, support `NUXT_` variables similar to `NUXT_PORT` and `NUXT_PORT` and also NITRO_ ones for upstream docs compat.

danielroe: Would make sense to support `NUXT_` as long as we also support those at runtime.

pi0: I guess for Node, yes (if not, we might implement alt env name support). I think regardless we can support it here and _fix_ in upstream.

pi0 (approved):

PR #21918, opened 2023-07-03
feat(nuxt): allow 'lazy' (non-blocking) server components

danielroe:
### ❓ Type of change

### 📚 Description

This adds support for 'lazy' server components that don't block navigation on the client-side, which is particularly useful when a server component might take a long time to render something and we want to show a loading state.

It adds support for a `fallback` slot (to display while a component is rendering) and a `lazy` prop. It would be nice to come up with a way for server components to have type support for slot/prop. 🤔 

**TODO**:
 - [ ] ensure `lazy` type works
 - [x] add support for fallback slot when component is loading

### 📝 Checklist


huang-julien (approved):
👍 ❤️

PR #21940, opened 2023-07-04
fix(nuxt): add cookies on server side fetching `useCookie`

### LINKED ISSUE #22631: accessing `useCookie` on server immediately after setting it doesn't return cookie
### Environment

Nuxi 3.6.5

- Operating System: Darwin
- Node Version:     v16.15.1
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Builder:          vite
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/sync42johnny/nuxt-app

### Description:
An issue has been identified with the authentication logic in the middleware of a Nuxt application. The application attempts to authenticate a user by checking for an `authToken` cookie. However, due to the `authToken` cookie not being set on the server-side before navigateTo, the middleware encounters an infinite redirection loop.

### Details:
**Affected Files**:
- `index.vue`
- `main.vue`
- `app.global.ts`
- `auth.ts`
- `confirm.ts`

**Workflow**:
1. On accessing the index page (`index.vue`), the `confirm` middleware is invoked. This middleware tries to set the `authToken` cookie on the server-side.
2. The middleware then redirects to `/main`.
3. The `/main` page is protected by the `auth` middleware, which checks for the presence of the `authToken` cookie.
4. Due to the cookie not being recognized (or initialized correctly on the server), the `auth` middleware redirects the user back to `/`, leading to a continuous loop of redirections.

### Additional context

The issue might be related to the incorrect functioning of the **`useCookie`** function when used within middleware before navigateTo.

### Logs

```shell-script
authToken.value in app.global undefined                                                                                                       2:10:58 PM
in confirm                                                                                                                                    2:10:58 PM
authToken.value in app.global undefined                                                                                                       2:10:58 PM
authToken in auth undefined
```

kyumoon:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22631

### ❓ Type of change

### 📚 Description
i set cookie using `useCookie` in plugins/{plugin}.ts ( update token to new value when token is expired )
and get cookie value using `useCookie` in pages/{page}.vue , it returns `undefined`
because `useCookie` get value from `event.node.req.headers.cookie`, this value not set until next request

i think, 'cookie' should be accessible before the next request.

reproduce: https://github.com/kyumoon/nuxt-cookie-issue/blob/main/plugins/auth.server.ts

### 📝 Checklist


pi0:
Thanks for this PR! 

I think it makes sens that `useCookie` can read both **request** cookies and also to **response** cookies since in Nuxt this is a two-way (read + write) utility.

Partially implementing cookie handling logic and serialization in Nuxt is probably not the best idea, however. Especially since h3 updates optimized internals for header handling.

I think it would be best if we introduce `parseResponseCookies` + `getResponseCookie` composables in `unjs/h3`. Then we can use `getResponseCookie` **first** and then fallback to `getCookie`.

An alternative method would be caching cookies state in memory of `nuxtApp` and check it first place.

kyumoon:
Thank you for your comments.

Since it is serialized at the time of 'app:rentered', I thought it could be used temporarily before.
it would be better if h3 supported it.

I'm trying to save token to cookie, so I'm not trying to save it in state.

danielroe:
marking this as draft until we implement via https://github.com/unjs/h3/issues/509

kyumoon:
It seems that the h3 modification is delayed. How about modifying it like this?

commit: https://github.com/kyumoon/nuxt/commit/ed4f21527558d0a7e2bfab959235a09be2994c48
example: https://github.com/kyumoon/nuxt/commits/example/use-cookie-on-server

@danielroe @pi0

awacode21:
As still still is draft, what is the current state?

PR #21980, opened 2023-07-06
refactor(nuxt): don't wrap server placeholders/client fallbacks

huang-julien:
### ❓ Type of change

### 📚 Description

Hi :wave: 

  This PR adds a small feature which i don't think will be used by many.  

  `noClientOnlyTransform` will prevent a client-side component to be wrapped within a `createClientOnly`.  

  As i'm currently working on a server components rework, this may be used (not a 100% sure) but it can be good to have this option as `NuxtClientFallback` will be using this instead of hard-coding the prevent behavior. 
This can also be needed for some very advanced usages from modules author to avoid having some server code within client bundle due to possible side-effects (`vue/server-renderer` with `NuxtClientFallback`) by having a .server and .client component.

### 📝 Checklist


packages/nuxt/src/core/nuxt.ts
```diff
@@ -251,7 +251,8 @@ async function initNuxt (nuxt: Nuxt) {
       name: 'NuxtClientFallback',
       priority: 10, // built-in that we do not expect the user to override
       filePath: resolve(nuxt.options.appDir, 'components/client-fallback.client'),
-      mode: 'client'
+      mode: 'client',
+      noClientOnlyTransform: true
```
huang-julien, line 255: Is there a better naming for this ?

danielroe: prewrapped? raw?

MiniDigger: raw might be confusing since vue has markRaw which you might use on a component https://vuejs.org/api/reactivity-advanced.html#markraw

huang-julien: oh i'd prefer raw over prewrapped 😅

packages/schema/src/types/components.ts
```diff
@@ -16,6 +16,11 @@ export interface Component {
    * components will be used instead of lower priority components.
    */
   priority?: number
+  /**
+   * This prevent the component to be transformed with a `createClientOnly()` client-side if the mode is `client`
+   * An example is the `<NuxtClientFallback>` component which handles it's hydration client-side
+   */
+  noClientOnlyTransform?: boolean
```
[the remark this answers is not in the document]
huang-julien: I'm not sure about that 🤔 I guess we'd need to know if some users need it. It's mainly a way to keep side-effects modules only on 1 side (server or client).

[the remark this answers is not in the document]
danielroe: I'd be happy to merge an fix needing this with an internal property that we set and consume for `<NuxtClientFallback>`, rather than exposing this as a public property for components which can be enabled by users ...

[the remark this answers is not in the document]
huang-julien: let's mark it private for now :+1:

danielroe (approved):

PR #21997, opened 2023-07-06
fix(cli): add aliases to monorepo paths to tsconfig includes

### LINKED ISSUE #21996: Importing components with alias in monorepo causes VSCode/Volar to fail finding modules/type declarations
### Environment

- Operating System: Darwin
- Node Version:     v18.16.0
- Nuxt Version:     3.6.1
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.6.6
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/github/samydoesit/nuxt-monorepo-import-reproduction
https://github.com/samydoesit/nuxt-monorepo-import-reproduction

### Describe the bug

Hi, i am experiencing issues while importing components in a monorepo using an alias.

The problem arises when using VSCode with the Volar extension. Since our ui directory is located outside of the Nuxt project directory, VSode (Volar) is unable to locate the module or type declaration.

However, if i manually add the path of the alias to the "includes" field in nuxt's tsconfig, the resolution in VSCode works.

### Additional context

_No response_

### Logs

```shell-script
Cannot find module '@ui/components/button/Button.vue' or its corresponding type declarations.
```

samydoesit:
### 🔗 Linked issue

resolves #21996

### ❓ Type of change

### 📚 Description

Add aliases to include property of tsconfig so that VSCode ts and intellisense works fine in a monorepo.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -52,6 +52,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
   const basePath = tsConfig.compilerOptions!.baseUrl ? resolve(nuxt.options.buildDir, tsConfig.compilerOptions!.baseUrl) : nuxt.options.buildDir
 
   tsConfig.compilerOptions = tsConfig.compilerOptions || {}
+  tsConfig.include = tsConfig.include || []
 
   for (const alias in aliases) {
     if (excludedAlias.some(re => re.test(alias))) {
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/utils/prepare.ts
```diff
@@ -68,6 +69,10 @@ export const writeTypes = async (nuxt: Nuxt) => {
     }
   }
 
+  for (const alias in nuxt.options.alias) {
+    tsConfig.include.push(aliases[alias])
+  }
```
danielroe, line 74: can we deduplicate this somehow? For example, we shouldn't add any aliases that are already covered by an existing glob.

packages/nuxi/src/utils/prepare.ts
```diff
@@ -68,6 +69,13 @@ export const writeTypes = async (nuxt: Nuxt) => {
     }
   }
 
+  for (const alias in nuxt.options.alias) {
+    if (aliases[alias].startsWith(nuxt.options.srcDir)) {
+      continue
+    }
+    tsConfig.include.push(aliases[alias])
+  }
```
danielroe, line 77: let's move this up into the previous for/in loop so we don't iterate twice over the aliases. We might also want to consider having different handling for directories (maybe need glob)?

samydoesit: moved it up and aligned with previous alias handling in the for loop.
If it's a directory is it necessary to also add the path with `/*`?

danielroe (approved):

danielroe:
oops, my mistake - didn't mean to merge this with co-authored-by 🤦

Issue #22036, opened 2023-07-07
Install upgraded nuxt version as main dependency


wokalek:
https://github.com/nuxt/nuxt/blob/2edd7a34419e042f6a694929d32178ecb379be51/packages/nuxi/src/commands/upgrade.ts#L54C29-L54C29

Why is its upgrade like devDependecy? It needs to be the main package if it's nuxt project, isn't it?

wokalek:
> Most things in nuxt3 are dev dependencies. The only thing that need to be regular dependencies are things that vue needs.

Why vue? I working with nuxt, not specifically with vue.

manniL:
Let's consolidate this in https://github.com/nuxt/nuxt/issues/22050

PR #22037, opened 2023-07-08
fix(nuxi): install nuxt as main dependency when doing nuxi upgrade

wokalek:
### 🔗 Linked issue

#22036 

### ❓ Type of change

### 📚 Description

Basically, projects based on the Nuxt framework depend on Nuxt directly and it is used as a core package. Accordingly, in cases when you work with Nuxt in production, you install dependencies via `pnpm install --prod`, while the Nuxt is not installed and you get an error.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/upgrade.ts
+++ b/packages/nuxi/src/commands/upgrade.ts
@@ -51,7 +51,7 @@ export default defineNuxtCommand({
 
     // Install latest version
     consola.info('Installing latest Nuxt 3 release...')
-    execSync(`${packageManager} ${packageManager === 'yarn' ? 'add' : 'install'} -D nuxt`, { stdio: 'inherit', cwd: rootDir })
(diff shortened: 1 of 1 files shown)


danielroe:
Currently we [deliberately add `nuxt` as a devDependency](https://github.com/nuxt/starter/blob/v3/package.json#L11-L15) but I think we could revisit this. Let's start with an issue to discuss: https://github.com/nuxt/nuxt/issues/22050

danielroe (changes requested):
For now, can we respect whatever option the user has selected? ie. if nuxt is in their devDeps it will use `-D` flag and otherwise not?

It may be worth making this updated PR directly to https://github.com/nuxt/cli though, rather than here.

pi0:
I think we might introduce this feature (and smart switch) for unjs/nypm util as well which is used in next CLI as new `upgradeDependency` utility.

PR #22049, opened 2023-07-09
perf(nuxt): prepopulate island payloads from rendered html

danielroe:
### ❓ Type of change

### 📚 Description

When using server components, when the rendered HTML is present in the DOM, we can populate the payload to avoid having to refetch the component if we later navigate away + back to the page.

### 📝 Checklist


huang-julien (commented):
Mounting another component in the page triggers a bug where a slots vnode isn't destroyed. 

-> test/fixtures/basic
- http://localhost:3000/islands
- click on `show` button from route island component
- there will be 2 `Interactive testing slot` with one uninteractive (previous vnode)

huang-julien (approved):
:+1:

PR #22074, opened 2023-07-10
fix(nuxt): skip scanning components that do not produce a name

YunYouJun:
### ❓ Type of change

### 📚 Description

An error component type was generated, like this (pascalName is null)

image

### 📝 Checklist


danielroe:
Thanks for spotting this. This might be an upstream issue in that library which should be fixed. Additionally, we should fix in Nuxt, but probably by filtering the component list earlier when it is generated rather than just here when the types are written.

YunYouJun:
I see. I add a filter for `scanComponent` function now.

danielroe (approved):

PR #22140, opened 2023-07-14
fix(nuxt): skip view transitions where there is no route change

### LINKED ISSUE #20977: experimental.viewTransition + Anchor point link will cause temporary page downtime
### Environment

- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.5.0
- Nitro Version:    2.4.1
- Builder:          vite
- User Config:      modules, experimental
- Runtime Modules:  @nuxt/content@2.6.0, @nuxt/ui@0.4.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-36uxp6?file=pages%2Ftest.vue

steps:
1. View the `/` or `/test` page
2. Click the heading link text

If you directly access a with a anchor link, the same situation will occur after loading is complete.

### Describe the bug

This problem occurs when you click on the link on the page to change the url. I observed through the performance panel in the chrome devtools tool that a script command that takes 4 seconds will be executed after clicking.

image

In the meantime, the following is the performance profile file I recorded
[Profile-20230521T104434.txt](https://github.com/nuxt/nuxt/files/11523639/Profile-20230521T104434.txt)

### Additional context

If you use `transition.ready.then()` in the `view-transition.client.ts` file, you will get an update timeout error message on the console.

```js [view-transition.client.ts]
transition.ready.then((v) => {
      console.log('ready', v)
})
```

> Uncaught (in promise) DOMException: Transition was aborted because of timeout in DOM update

By the way, I know very little about startViewTransition API, but I think it's more like a chromium problem. I added the same interceptor code to the vue-router project and got the same results.

### Logs

_No response_

Jannchie:
### 🔗 Linked issue

#20977

### ❓ Type of change

### 📚 Description

Resolves #20977

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -9,9 +9,8 @@ export default defineNuxtPlugin((nuxtApp) => {
 
   const router = useRouter()
 
-  router.beforeResolve((to) => {
-    if (to.meta.pageTransition === false) { return }
-
+  router.beforeResolve((to, from) => {
+    if (to.path === from.path || to.meta.pageTransition === false) { return }
     const promise = new Promise<void>((resolve, reject) => {
       finishTransition = resolve
       abortTransition = reject
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -10,7 +10,9 @@ export default defineNuxtPlugin((nuxtApp) => {
   const router = useRouter()
 
   router.beforeResolve((to, from) => {
-    if (to.path === from.path || to.meta.pageTransition === false) { return }
+    if (to === from || to.matched.every((comp, index) => comp.components && comp.components?.default === from.matched[index]?.components?.default)) {
+      return
+    }
     const promise = new Promise<void>((resolve, reject) => {
       finishTransition = resolve
       abortTransition = reject


packages/nuxt/src/app/plugins/view-transitions.client.ts
```diff
@@ -9,9 +9,8 @@ export default defineNuxtPlugin((nuxtApp) => {
 
   const router = useRouter()
 
-  router.beforeResolve((to) => {
-    if (to.meta.pageTransition === false) { return }
-
+  router.beforeResolve((to, from) => {
+    if (to.path === from.path || to.meta.pageTransition === false) { return }
```
danielroe, line 13:
```suggestion
    if (to.fullPath === from.fullPath || to.meta.pageTransition === false) { return }
```

danielroe (changes requested):
We also need to check whether the components will change by verifying whether components will change by looking into the `matched` property. (If they don't, then `page:finish` won't be called and the transition will abort.)

Alternatively, we can update view-transitions implementation so `finishTransition` is called after the route updates, even if there has been no call of `page:finish`.

packages/nuxt/src/app/plugins/view-transitions.client.ts
Jannchie: I tried using `fullPath`. But this causes a transition to be triggered when clicking on the anchor when navigating within the page. The page still gets stuck for a few seconds.

danielroe: Yes, that's a fair point. Can we instead use the matching algorithm e.g. 
https://github.com/nuxt/nuxt/blob/9f5130d06b3d449dd16d33ee5e41c820d037e2c4/packages/nuxt/src/app/components/nuxt-loading-indicator.ts#L43-L47

Jannchie: I tried it, and it does work.

Jannchie: I have recommit the code with reference to indicator's implementation, thanks for the tip!

danielroe (approved):

packages/nuxt/src/app/plugins/view-transitions.client.ts
danielroe: perfect 👌

PR #21860, opened 2023-06-29
fix(nuxt): ensure static presets equivalent to `nuxi generate`

danielroe:
### 🔗 Linked issue

resolves nuxt/nuxt#21254

### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This finishes the second task on nuxt/nuxt#21254. Previous we enabled static nitro preset in `nuxi generate` (https://github.com/nuxt/nuxt/pull/21655), but this follows up to ensure that running `nuxi build --prerender` (or `NITRO_PRESET=static nuxi build` have the same behaviour.

From this point, we no longer use `options._generate` flag but directly access `nitro.options.static`. We can remove the internal `._generate` flag in a future minor release.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #22162, opened 2023-07-15
fix(nuxt): work around payload extraction issue

### LINKED ISSUE #22068: [nuxt] Cannot load payload  /_payload.json
### Environment

- Node Version:     v18.16.0
- Nuxt Version:     3.6.2
- Nitro Version:    2.5.2
- Package Manager:  npm@9.8.0
- Builder:          vite
- User Config:      pages, debug, dev, modules, pwa, app, vue, ignore
- Runtime Modules:  @nuxt/devtools@0.2.5
- Build Modules:    -

### Reproduction

its a pretty big project, It always worked fine and all I did before this happened was running npm install

### Describe the bug

When using my nuxt site using a different domain navigating to other pages stops working

when I build my site and upload it to azure it comes available on multiple domains

1. the azure domain (Site works fine here)
2. Our domain with a special subdomain (usually its help.domain.app)

On our subdomain the website navigation stops working and spits out the following error:
image

### Additional context

_No response_

### Logs

```shell-script
build logs
https://hastebin.com/share/oyamipasuv.bash
```

TheAlexLichter:
### 🔗 Linked issue

Resolves #22068 

### ❓ Type of change

### 📚 Description

Due to changes from https://github.com/nuxt/nuxt/pull/21860, [`process.env.NUXT_PAYLOAD_EXTRACTION`](https://github.com/nuxt/nuxt/blob/fed5868184c6ec7256c114cb804253b8deacdf64/packages/nuxt/src/core/nitro.ts#L194) is set "too early" for mentioned changes to come into effect. To resolve it, I've used Nuxt's nitro config to check for the `static` override. This should be removed in the future (3.8 or so).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -378,7 +378,7 @@ async function initNuxt (nuxt: Nuxt) {
 
   // Init nitro
   await initNitro(nuxt)
-  
+
   if (!nuxt.options.dev && nuxt.options.experimental.payloadExtraction) {
     addPlugin(resolve(nuxt.options.appDir, 'plugins/payload.client'))
   }
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -1,7 +1,7 @@
 import { join, normalize, relative, resolve } from 'pathe'
 import { createDebugger, createHooks } from 'hookable'
 import type { LoadNuxtOptions } from '@nuxt/kit'
-import { addBuildPlugin, addComponent, addPlugin, addVitePlugin, addWebpackPlugin, installModule, loadNuxtConfig, logger, nuxtCtx, resolveAlias, resolveFiles, resolvePath, tryResolveModule, useNitro } from '@nuxt/kit'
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -369,15 +369,16 @@ async function initNuxt (nuxt: Nuxt) {
 
   addModuleTranspiles()
 
-  // Init nitro
-  await initNitro(nuxt)
-
   // TODO: remove when app manifest support is landed in https://github.com/nuxt/nuxt/pull/21641
   // Add prerender payload support
-  if (useNitro().options.static && nuxt.options.experimental.payloadExtraction === undefined) {
+  if (nuxt.options.nitro?.static && nuxt.options.experimental.payloadExtraction === undefined) {
```
danielroe, line 374: I think we need to find another way to resolve. Nitro static presets need static behaviour even though this won't be set in the Nuxt options.

manniL: Oh, I see. Then let's chat about that 👍🏻

danielroe (approved):

PR #21793, opened 2023-06-26
perf(nuxt): render css/inline css higher than resource hints

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/21784#issuecomment-1607199537

### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

As suggested by @prashantpalikhe, this reorders the head tags to prioritise styles over resource hints. I think there's likely some future space for optimisation here, both in unhead and (possibly) in nuxt/bundle-renderer, such as 'splitting up' head metadata to prioritise 'important' meta at top, and defer less important meta after resource hints. (cc: @harlan-zw) Would value thoughts.

Co-authored-by: Prashant Palikhe <prashpal09@gmail.com>

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #22179, opened 2023-07-17
feat(nuxt): render all head tags on server with `unhead`

### LINKED ISSUE #22082: Can't order head tags after Nuxt head tags with `tagPriority`
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.6.1`
- Nitro Version:    `2.5.2`
- Package Manager:  `pnpm@8.6.0`
- Builder:          `vite`
- User Config:      `devServer`, `app`, `routeRules`, `alias`, `modules`, `experimental`, `vite`, `css`, `typescript`, `devtools`
- Runtime Modules:  `@vueuse/nuxt@10.2.1`, `@unocss/nuxt@0.53.4`, `nuxt-beastcss@1.0.3`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-luygfj?file=nuxt.config.ts

(Inspect frame)

### Describe the bug

Hi :wave: 
I am trying to add this noscript tag to the bottom of `<head>`. By default, Nuxt adds it just below `<title>`. I tried using `tagPriority` and assumed the bigger the number, the later it would be registered.
```ts
  app: {
    head: {
      noscript: [{ innerHTML: '...css link', tagPriority: 20 }],
    },
  }});
```

`tagPriority` has the following jsdoc
```
(property) tagPriority?: number | "critical" | "high" | "low" | `before:${string}` | `after:${string}` | undefined

The priority for rendering the tag, without this all tags are rendered as they are registered (besides some special tags).

The following special tags have default priorities:

-2 <meta charset ...>
-1
0 <meta http-equiv="content-security-policy" ...>
All other tags have a default priority of 10
```

Negative values seem to work, a `-10` put it first in head. I have tried 10 - 200, but can't make it lower than just below `<title>`. I would've tried to put it in body, but I'm putting a css link inside, and [that's not allowed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noscript).

### Additional context

Not exactly related... https://github.com/nuxt/nuxt/issues/21570

### Logs

_No response_

harlan-zw:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22082
https://github.com/unjs/unhead/issues/116

### ❓ Type of change

### 📚 Description

Currently, Nuxt users will modify their pages <head> using `useHead` or nuxt.config appHead  This appears to be stable and working well.

The Nuxt core itself though does not rely on this same API, instead opting for building head output using strings and hardcoding output ordering.

By migrating all head output to go through Unhead we get the following:
- Full type safety
- Tags can be ordered for for better performance (see https://github.com/danielroe/nuxt-capo/issues/36)
- Gives end users and module authors more control over where tags end up and which they want to include (for example, pages with no scripts could easily be filtered out via a plugin hook)

```ts
  name: 'remove-scripts',
  enforce: 'pre',
  setup(nuxtApp) {
   // pretty much the same thing as NO_SCRIPTS logic in renderer.ts
    nuxtApp.ssrContext!.head.hooks.hook('tags:resolve', (ctx) => {
      ctx.tags = ctx.tags.filter(t => {
        if (t.tag === 'script')
          return false
        if (t.tag === 'link' && t.props.rel === 'modulepreload')
          return false
        return true
      })
    })
  }
})
```

This comes at a likely very small performance cost for SSR. A benchmark would be useful.

### Performance Improvement

- nuxt.config is only rendered server-side now, this should slightly improve the CSR time to interactive metric. This was already the behavior of SPAs.

### :warning: Slightly breaking changes

While these changes should be mostly safe, it will require a delicate eye over the migration to make sure I haven't omitted any previous tags or changed the order. I don't think it's safe to implement app-wide head re-ordering just yet.

- Overriding the meta using `ssrContext.renderMeta` won't work anymore, afaik this was internal and shouldn't really be used in any production apps? 
- Using `#head-static`, shouldn't be used

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -127,7 +132,7 @@ const getSPARenderer = lazyCachedFunction(async () => {
   const renderer = createRenderer(() => () => {}, options)
   const result = await renderer.renderToString({})
 
-  const renderToString = (ssrContext: NuxtSSRContext) => {
+  const renderToString = async (ssrContext: NuxtSSRContext) => {
```
danielroe, line 135:
```suggestion
  const renderToString = (ssrContext: NuxtSSRContext) => {
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -10,23 +10,14 @@ import type { H3Event } from 'h3'
 import type { AppConfig, AppConfigInput, RuntimeConfig } from 'nuxt/schema'
 import type { RenderResponse } from 'nitropack'
 
-// eslint-disable-next-line import/no-restricted-paths
+import type { MergeHead, VueHeadClient } from '@unhead/vue'
 import type { NuxtIslandContext } from '../core/runtime/nitro/renderer'
```
danielroe, line 14:
```suggestion
// eslint-disable-next-line import/no-restricted-paths
import type { NuxtIslandContext } from '../core/runtime/nitro/renderer'
```

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -9,14 +15,16 @@ import destr from 'destr'
...
 import { hash } from 'ohash'
+import { renderSSRHead } from '@unhead/ssr'
 
 import { defineRenderHandler, getRouteRules, useRuntimeConfig } from '#internal/nitro'
 import { useNitroApp } from '#internal/nitro/app'
 
-// eslint-disable-next-line import/no-restricted-paths
+import type { Link, Script } from '@unhead/vue'
+import { createServerHead } from '@unhead/vue'
 import type { NuxtPayload, NuxtSSRContext } from '#app/nuxt'
```
danielroe, line 25:
```suggestion
// eslint-disable-next-line import/no-restricted-paths
import type { NuxtPayload, NuxtSSRContext } from '#app/nuxt'
```

harlan-zw:
Related: https://github.com/nuxt/nuxt/pull/21793

danielroe:
I was thinking this PR would close that one. But maybe I misunderstood?

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -472,25 +499,41 @@ function renderPayloadResponse (ssrContext: NuxtSSRContext) {
...
-    'type="application/json"',
-    `id="${opts.id}"`,
-    `data-ssr="${!(process.env.NUXT_NO_SSR || opts.ssrContext.noSSR)}"`,
-    opts.src ? `data-src="${opts.src}"` : ''
-  ].filter(Boolean)
+function renderPayloadJsonScript (opts: { id: string, ssrContext: NuxtSSRContext, data?: any, src?: string }): Script[] {
   const contents = opts.data ? stringify(opts.data, opts.ssrContext._payloadReducers) : ''
-  return `<script ${attrs.join(' ')}>${contents}</script>` +
-    `<script>window.__NUXT__={};window.__NUXT__.config=${uneval(opts.ssrContext.config)}</script>`
+  const payload : Script = {
```
danielroe, line 504:
```suggestion
  const payload: Script = {
```

danielroe (approved):
Lovely.

danielroe:
Just to add that in my testing, this is slightly faster than our existing implementation.

<details>
<summary>New approach (basic fixture)</summary>

```
Concurrency Level:      10
Time taken for tests:   289.765 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      53800000 bytes
HTML transferred:       51840000 bytes
Requests per second:    34.51 [#/sec] (mean)
Time per request:       289.765 [ms] (mean)
Time per request:       28.976 [ms] (mean, across all concurrent requests)
Transfer rate:          181.32 [Kbytes/sec] received
```

</details>


<details>
<summary>Old approach (basic fixture)</summary>

```
Concurrency Level:      10
Time taken for tests:   293.881 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      52830000 bytes
HTML transferred:       50870000 bytes
Requests per second:    34.03 [#/sec] (mean)
Time per request:       293.881 [ms] (mean)
Time per request:       29.388 [ms] (mean, across all concurrent requests)
Transfer rate:          175.55 [Kbytes/sec] received
```

</details>


<details>
<summary>New approach (minimal fixture)</summary>

```
Concurrency Level:      10
Time taken for tests:   4.232 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      7900000 bytes
HTML transferred:       6560000 bytes
Requests per second:    2362.97 [#/sec] (mean)
Time per request:       4.232 [ms] (mean)
Time per request:       0.423 [ms] (mean, across all concurrent requests)
Transfer rate:          1822.99 [Kbytes/sec] received
```

</details>


<details>
<summary>Old approach (minimal fixture)</summary>

```
Concurrency Level:      10
Time taken for tests:   5.401 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      7810000 bytes
HTML transferred:       6470000 bytes
Requests per second:    1851.51 [#/sec] (mean)
Time per request:       5.401 [ms] (mean)
Time per request:       0.540 [ms] (mean, across all concurrent requests)
Transfer rate:          1412.13 [Kbytes/sec] received
```

</details>

ijkml:
Awesome! 🙏

PR #22198, opened 2023-07-18
fix(nuxt): avoid premature hydration when using async layouts

### LINKED ISSUE #21999: custom layout makes useAsyncData run on the client
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.2
- Nitro Version:    2.5.2
- Package Manager:  npm@9.4.2
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Source:
https://stackblitz.com/edit/nuxt-starter-ufrun1?file=pages%2Fadmin.vue

Preview:
https://nuxtstarterufrun1-yc3l--3000--e809191e.local-credentialless.webcontainer.io/admin

Repro:
1. Open dev tools network tab
2. Open preview link above
3. Note fetch call. It shouldnt be there
4. Click link to index, then link back to admin
5. Note double fetch calls

### Describe the bug

Since v3.6 I am getting client calls from useAsyncData when using custom layout.

Thank you

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #22091: Top-level await in layout breaks hydration in Nuxt 3.6
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.20.0`
- Nuxt Version:     `3.6.2`
- Nitro Version:    `2.5.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

[Here's a reproduction of the bug running on Nuxt 3.6.2](https://stackblitz.com/edit/nuxt-starter-4cbwhd?file=pages%2Findex.vue).
And for comparison [here's the same reproduction running on Nuxt 3.5.3](https://stackblitz.com/edit/nuxt-starter-l5ts6s?file=layouts%2Fdefault.vue) and the bug isn't there.

### Describe the bug

Starting in Nuxt 3.6, using top-level await in a layout causes the hydration to fail because `nuxt.isHydration` is set to `false` before the hydration has actually finished.

This leads to any `useAsyncData`s you have, checking for data in `nuxt.static.data[key]` where there is none, instead of `nuxt.payload.data[key]` where the data is ([see code](https://github.com/nuxt/nuxt/blob/08acfb749de2394e30e7c27d49f4226429743232/packages/nuxt/src/app/composables/asyncData.ts#L123)). This means all your `useAsyncData`s are suddenly set to `null` and the refetched, causing major hydration errors.

### Additional context

I suspect it's caused by PR #21450 reintroducing `defineAsyncComponent` that had previously been replaced by a work-around because of an upstream bug with `<Suspense>` in Vue (vuejs/core#6638)

It's been really difficult to debug this since all the hydration errors and problems happened after this bug silently caused `nuxt.isHydration` to be set to `false` too soon. Any tips on how to find what's causing bugs like this in a large codebase are very much appreciated.

This is blocking our upgrade to Nuxt 3.6 since we fetch some data to initialize a Pinia store in our layout. Don't know if that's bad practice though. Might move it to a plugin instead. At least until this is fixed.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22091
resolves https://github.com/nuxt/nuxt/issues/21999

### ❓ Type of change

### 📚 Description

In https://github.com/nuxt/nuxt/pull/21450, we wrongly removed a workaround for an upstream vue bug (https://github.com/vuejs/core/issues/6638), introducing a regression.

This PR reverts back to the workaround but ultimately we still need to resolve in vue core.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/layout.ts
+++ b/packages/nuxt/src/app/components/layout.ts
@@ -13,6 +13,21 @@ import layouts from '#build/layouts'
 import { appLayoutTransition as defaultLayoutTransition } from '#build/nuxt.config.mjs'
 import { useNuxtApp } from '#app'
 
+// TODO: revert back to defineAsyncComponent when https://github.com/vuejs/core/issues/6638 is resolved
+const LayoutLoader = defineComponent({
+  name: 'LayoutLoader',
+  inheritAttrs: false,
+  props: {
+    name: String,
+    layoutProps: Object
+  },
+  async setup (props, context) {
+    const LayoutComponent = await layouts[props.name]().then((r: any) => r.default || r)
+
+    return () => h(LayoutComponent, props.layoutProps, context.slots)
+  }
+})
+
 export default defineComponent({
   name: 'NuxtLayout',
   inheritAttrs: false,
@@ -45,14 +60,16 @@ export default defineComponent({
       // We avoid rendering layout transition if there is no layout to render
       return _wrapIf(Transition, hasLayout && transitionProps, {
         default: () => h(Suspense, { suspensible: true, onResolve: () => { nextTick(done) } }, {
-          // @ts-expect-error seems to be an issue in vue types
-          default: () => h(LayoutProvider, {
-            layoutProps: mergeProps(context.attrs, { ref: layoutRef }),
-            key: layout.value,
-            name: layout.value,
-            shouldProvide: !props.name,
-            hasTransition: !!transitionProps
-          }, context.slots)
+          default: () => h(
+            // @ts-expect-error seems to be an issue in vue types
+            LayoutProvider,
+            {
+              layoutProps: mergeProps(context.attrs, { ref: layoutRef }),
+              key: layout.value,
+              name: layout.value,
+              shouldProvide: !props.name,
+              hasTransition: !!transitionProps
+            }, context.slots)
         })
       }).default()
     }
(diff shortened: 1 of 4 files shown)


antfu (approved):

gerbenvandijk:
This issue still seems present when you are using a custom error layout (error.vue in the root of the application)

manniL:
@gerbenvandijk can you provide a reproduction + opening a new issue? Feel free to link this one in the new issue 👍

gerbenvandijk:
will sort it in the coming week somewhere :) thanks for the quick response!

gerbenvandijk:
> @gerbenvandijk can you provide a reproduction + opening a new issue? Feel free to link this one in the new issue 👍


will sort it in the coming week somewhere :) thanks for the quick response!

gerbenvandijk:
@manniL @danielroe 

After setting up a reproduction I was able to resolve it on my end, it had to do with a hydration that was introduced due to misunderstanding on my end. 

I was using the `useRequestHeaders()` composable in a plugin (to pass some headers to Apollo) not realising that `useRequestHeaders()` is not available on the client side. After storing the specific header using the `useState()` composable this problem went away. Thanks anyways for your quick response and apologies :)

danielroe:
No problem! Glad you resolved it 😊

PR #22214, opened 2023-07-19
perf(nuxt,schema): do not watch `buildDir` and `node_modules`

danielroe:
### 🔗 Linked issue

https://github.com/unjs/nitro/issues/1458, https://github.com/nuxt/nuxt/discussions/20213

### ❓ Type of change

### 📚 Description

This follows up some earlier fixes to the watching implementation for Nuxt, which will particularly affect Linux and Windows machines. There are two issues fixed here as well as a third potential issue:

1. we were adding `.nuxt` and `.nuxt/analyze` in their absolute form which were ignored, meaning we were watching the build directory 😱
2. we were not ignoring `node_modules` within layers, only at the top level 😱
3. we have also added additional output directories like `.vercel`, `.netlify` and `dist`, as well as excluding `.output`, `.git` and `.cache` even when these are within subdirectories. (There has been no explicit report that this fixes, but I think it should be safe and improve performance.)

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -355,9 +355,9 @@ export default defineUntypedSchema({
       '**/*.stories.{js,ts,jsx,tsx}', // ignore storybook files
       '**/*.{spec,test}.{js,ts,jsx,tsx}', // ignore tests
       '**/*.d.ts', // ignore type declarations
-      '.output',
-      '.git',
-      '.cache',
+      '**/.output',
+      '**/.git',
+      '**/.cache',
       relative(await get('rootDir'), await get('analyzeDir')),
       relative(await get('rootDir'), await get('buildDir')),
       await get('ignorePrefix') && `**/${await get('ignorePrefix')}*.*`
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -352,15 +352,11 @@ export default defineUntypedSchema({
    */
   ignore: {
     $resolve: async (val, get) => [
-      '**/*.stories.{js,ts,jsx,tsx}', // ignore storybook files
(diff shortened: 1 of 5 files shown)


antfu (approved):

packages/schema/src/config/common.ts
```diff
@@ -355,11 +355,14 @@ export default defineUntypedSchema({
       '**/*.stories.{js,ts,jsx,tsx}', // ignore storybook files
       '**/*.{spec,test}.{js,ts,jsx,tsx}', // ignore tests
       '**/*.d.ts', // ignore type declarations
```
antfu, line 357: Now we also have `.d.mts` and `.d.cts`

packages/schema/src/config/common.ts
```diff
@@ -355,11 +355,14 @@ export default defineUntypedSchema({
...
-      '.git',
-      '.cache',
-      await get('analyzeDir'),
-      await get('buildDir'),
+      '**/dist',
+      '**/.vercel',
+      '**/.netlify',
+      '**/.output',
+      '**/.git',
+      '**/.cache',
```
pi0, line 363: We can possibly read these from `.gitignore` to be extendable. `.data` also can/should be added to the list coming from next version of nitro for `data:` storage.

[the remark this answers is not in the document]
danielroe: nice 👌

danielroe: .gitignore seems a nice/better idea but I wonder if someone has (for example) a generated schema that they want to be in build but not committed to git...

Maybe worth including in 3.7 with explanation on how to 'unignore' it?

TheAlexLichter (approved):

pi0 (approved):

PR #22251, opened 2023-07-20
fix(nuxi): allow nuxt.config and module setup to use `process.dev`

harlan-zw:
### 🔗 Linked issue

-

### ❓ Type of change

### 📚 Description

To check whether Nuxt is running in development, you can use the `process.dev` flag. 

This works as expected at runtime. However, when using the same flag at build time, the result will be undefined (see https://stackblitz.com/edit/nuxt-starter-jev4hu?file=nuxt.config.ts).

The recommendation is to use `nuxt.options.dev`, but for consistency, we should override this environment variable when we set the `NODE_ENV` as well.

This isn't exactly a bug, just something that has just caught me out.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/env.ts
+++ b/packages/nuxi/src/utils/env.ts
@@ -5,4 +5,5 @@ export const overrideEnv = (targetEnv: string) => {
   }
 
   process.env.NODE_ENV = targetEnv
+  process.dev = targetEnv === 'development'
 }


pi0:
I would suggest using`isDevelopment` from `std-env` for a standard way to check this. `process.dev` shim was only introduced in runtime to allow tree-shaking.

harlan-zw:
I don't think this is a great recommendation for modules, modules would need to explicitly include std-env as a dependency just for this check which could result in version conflicts (unless marked as external).

We have nuxt.options.dev which is fine, but IMO a consistent API is favorable.

danielroe:
My worry would be that we don't want to obscure the differences between build-time and runtime code, which otherwise we're at great pains to establish (e.g. not importing `@nuxt/kit` in runtime code, etc.).

> A valid use case for a consistent API is when build-time and runtime code is shared. For example when emitting assets after prerendering is finished. (i.e nuxt-simple-sitemap, nuxt-og-image)

I think this is something we should think about - what about a utility that can run + require code 'within' a nuxt context? Or code registered in a nitro (runtime) plugin that runs after a build is complete. I think storage/bundled storage solves some of this but maybe worth brainstorming some ideas....

harlan-zw:
> My worry would be that we don't want to obscure the differences between build-time and runtime code, which otherwise we're at great pains to establish (e.g. not importing `@nuxt/kit` in runtime code, etc.).
> 
> > A valid use case for a consistent API is when build-time and runtime code is shared. For example when emitting assets after prerendering is finished. (i.e nuxt-simple-sitemap, nuxt-og-image)
> 
> I think this is something we should think about - what about a utility that can run + require code 'within' a nuxt context? Or code registered in a nitro (runtime) plugin that runs after a build is complete. I think storage/bundled storage solves some of this but maybe worth brainstorming some ideas....

This is better for targeting the root issue.

Have moved to a discussion here: https://github.com/nuxt/nuxt/issues/22255

Issue #21277, opened 2023-05-31
only null Type is allowed to useAsyncData's default Type


qoo1476:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v16.14.2
- Nuxt Version:     3.5.2
- Nitro Version:    2.4.1
- Package Manager:  yarn@1.22.19
- Builder:          vite
- User Config:      telemetry, ssr, experimental, app, modules, css, imports, build, nitro
- Runtime Modules:  @vueuse/nuxt@10.1.2, @unocss/nuxt@0.52.5, @pinia/nuxt@0.4.11
- Build Modules:    -
------------------------------

### Reproduction

i tried to reproduction but it doesn't show typescript error

### Describe the bug

sorry my english is bad...

when i use useAsyncData with default data

```
type User = {
  name: string
  auth: string
}

const { data, pending, refresh } = useAsyncData<User>(
  () => $fetch('/users/1'),
  {
    default: () => ({
      name: '',
      auth: '',
    }),
  },
)
```

it shows typescript error 

so i checked useAsyncData's Type
```
export declare function useAsyncData<ResT, DataE = Error, DataT = ResT, PickKeys extends KeysOf<DataT> = KeysOf<DataT>, DefaultT = null>(handler: (ctx?: NuxtApp) => Promise<ResT>, options?: AsyncDataOptions<ResT, DataT, PickKeys, DefaultT>): AsyncData<PickFrom<DataT, PickKeys> | DefaultT, DataE | null>;
```

there are two ways i found
1. don't use default option or (awalys return null value [X]mean nothing)
```
const { data, pending, refresh } = useAsyncData<User>(
  () => $fetch('/users/1'),
  {
    default: () => null
  },
)
```
2.  generate default Type
```
const { data, pending, refresh } = useAsyncData<
  User,
  Error,
  User,
  KeysOf<User>,
  User <= [ here is default option's type ]
>(() => $fetch('/users/1'), {
  default: () => ({
    name: '',
    auth: '',
  }),
})
```

so i wonder it's any reason to default option's default type is null ?
i think it could be better with
```
DefaultT = null    =>   DefaultT = DataT [or] DefaultT = null|DataT
```
or i awalys have to generate default type when use useAsyncData

and i also checked nuxt.com api document and it looks like what i said

![image](https://github.com/nuxt/nuxt/assets/68536972/f2ecacda-e9dc-43a5-a073-344ca066ed0c)

[...]

PR #22258, opened 2023-07-20
fix(nuxt): add overloads for `asyncData` generic + default

### LINKED ISSUE #22164: TypeScript type error in `useFetch` default option with `ResT` generic and `transform` returning `unknown` when making external server requests
### Environment

Node Version: v16.20.0
Nuxt Version: 3.6.2

### Reproduction

https://stackblitz.com/edit/nuxt-starter-xrqqhx?file=composables%2Ftest.ts

### Describe the bug

When using `useFetch` with the `ResT` generic, a TypeScript type error occurs with the `default` option. The error message is as follows:

```ts
useFetch<number[]>('/api/count', {
  default: () => [0]
}
```

```
Type '() => number[]' is not assignable to type '() => Ref<null> | null'.
```

However, the error can be resolved by using the following syntax:

```ts
useFetch<number[]>('/api/count', {
  default: (): number[] => [0]
}
```

But when making requests to an external server, using this syntax causes the `res` parameter in the `transform` function to be of type `unknown` instead of `number[]`. An example of this issue is shown below:

```ts
const { data: data2 } = useFetch('https://reqres.in/api/users?page=2', {
  default: (): number[] => [],
  transform(res){
    return res.map(i => i);
  }
});
```

```
'res' is of type 'unknown'.
```

### Additional context

_No response_

### Logs

_No response_

94726:
### 🔗 Linked issue

resolves #22164

### ❓ Type of change

### 📚 Description

Adds overloads to `useFetch` and `useAsyncData` (+Lazy variants) for handling the correct type of the default function while also changing the type of `data` from `DataT | null` to `DataT` when providing a default.

I tried implementing it without adding even more overloads, but I just couldn't figure it out :/

### 📝 Checklist

### DIFF
--- a/test/fixtures/basic-types/types.ts
+++ b/test/fixtures/basic-types/types.ts
@@ -323,6 +323,17 @@ describe('composables', () => {
     expectTypeOf(useLazyFetch<string>('/test', { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
     expectTypeOf(useAsyncData<string>(() => $fetch('/test'), { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
     expectTypeOf(useLazyAsyncData<string>(() => $fetch('/test'), { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
+
+    // transform must match the explicit generic because of typescript limiations microsoft/TypeScript#14400
+    expectTypeOf(useFetch<string>('/test', { transform: () => 'transformed' }).data).toEqualTypeOf<Ref<string | null>>()
+    expectTypeOf(useLazyFetch<string>('/test', { transform: () => 'transformed' }).data).toEqualTypeOf<Ref<string | null>>()
+    expectTypeOf(useAsyncData<string>(() => $fetch('/test'), { transform: () => 'transformed' }).data).toEqualTypeOf<Ref<string | null>>()
+    expectTypeOf(useLazyAsyncData<string>(() => $fetch('/test'), { transform: () => 'transformed' }).data).toEqualTypeOf<Ref<string | null>>()
+
+    expectTypeOf(useFetch<string>('/test', { default: () => 'test', transform: () => 'transformed' }).data).toEqualTypeOf<Ref<string>>()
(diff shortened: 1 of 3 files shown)


danielroe:
Thank you! Would you mind adding some tests to `test/fixtures/basic-types/types.ts` to cover these changes?

test/fixtures/basic-types/types.ts
```diff
@@ -309,6 +309,22 @@ describe('composables', () => {
...
+    useLazyFetch<string>('/test', { default: () => 0 })
+    // @ts-expect-error default type should match generic type
+    useAsyncData<string>(() => $fetch('/test'), { default: () => 0 })
+    // @ts-expect-error default type should match generic type
+    useLazyAsyncData<string>(() => $fetch('/test'), { default: () => 0 })
+
+    expectTypeOf(useFetch<string>('/test', { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
+    expectTypeOf(useLazyFetch<string>('/test', { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
+    expectTypeOf(useAsyncData<string>(() => $fetch('/test'), { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
+    expectTypeOf(useLazyAsyncData<string>(() => $fetch('/test'), { default: () => 'test' }).data).toEqualTypeOf<Ref<string>>()
```
danielroe, line 325: nice. could you also add tests with both default + transform so we verify those still work?

94726: I just noticed that providing a generic parameter results in the `transform` not being able to return a different type.
e. g. this is not possible because `number` isn't assignable to `string`
```ts
useFetch<string>('/test', {default: () => 'test', transform: (arg) => 0})
```
Unfortunately this seems to be a typescript limitation at the moment, see https://github.com/microsoft/TypeScript/issues/14400

---
Should I simply add tests for the case of returning the matching type?

danielroe: As long as it's not a regression in this PR, I'm happy just documenting current behaviour in the tests 👍

94726: Added some more tests and there shouldn't be any regressions :)

danielroe (approved):
Thank you ❤️

qoo1476:
resolves #21277 too, thanks

PR #22277, opened 2023-07-21
fix(nuxt): update `useNuxtData` to access shared asyncData state

### LINKED ISSUE #19704: useNuxtData setting overwriting data doesn't work
### Environment

- Operating System: `Linux`
- Node Version:     `v18.11.0`
- Nuxt Version:     `3.3.1`
- Nitro Version:    `2.3.1`
- Package Manager:  `pnpm@7.13.6`
- Builder:          `vite`
- User Config:      `modules`, `srcDir`, `ssr`, `nitro`, `devServer`, `typescript`, `devtools`, `runtimeConfig`
- Runtime Modules:  `@unocss/nuxt@0.50.4`, `@vueuse/nuxt@9.13.0`, `@nuxt/devtools@0.2.5`, `nuxt-typed-router@3.0.6`
- Build Modules:    `-`

### Reproduction

I've recreated the [optimistic update example from the docs](https://nuxt.com/docs/api/composables/use-nuxt-data#optimistic-updates):
https://stackblitz.com/edit/github-a2zxx3?file=requests/todos.ts
1. Click add button
2. Click add button again (every second add will fail)
3. See that the data doesn't get overwritten.

### Describe the bug

When using `useNuxtData`, it is not possible to overwrite the data directly:

```ts

const {data: todos} = useFetch('/api/todos', {key: 'todos'})

const {data: nuxtDataTodos} = useNuxtData<string[]>('todos') // array of todo
nuxtDataTodos.value = ["replace all todos"] // won't update todos from useFetch/useAsyncData
```

From looking at [useNuxtData docs](https://nuxt.com/docs/api/composables/use-nuxt-data#optimistic-updates), I'd expect this to work.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #22696: Unable to fetch data on client side if useNuxtData used before useFetch
### Environment

- Operating System: Darwin
- Node Version:     v20.5.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Builder:          vite
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

**Github**: https://github.com/RomanSkrypnik/nuxt-server-false-issue
**Videos**:
1. https://monosnap.com/direct/UYhZDjC21caXQAovqKsuAGyEMm4Pi0
2. https://monosnap.com/direct/lDKGf7wxSvv58r6TTCPuIk0ZNRk7KH

### Describe the bug

Using `useNuxtData` before `useFetch` on client side prevents request to API
If I try to comment/remove `useNuxtData` before `useFetch` - it will work just fine
Also, using `useNuxtData` with `useFetch` on server side works fine

### Additional context

_No response_

### Logs

_No response_

94726:
### 🔗 Linked issue

resolves #19704
resolves https://github.com/nuxt/nuxt/issues/22696

### ❓ Type of change

### 📚 Description

This change synchronizes the data value of `useNuxtData` and `useAsyncData`.

Currently, they aren't really in sync. This happens because the data in the payload (used by `useNuxtData`) and asyncData are not the same refs. They're only set to the same values on two occasions:
1. when a fetch resolved
2. when asyncData is first called per key and it gets hydrated by the payload

### 📝 Checklist


danielroe:
Would you be able  add a regression test?

94726:
Will do, but I noticed that my implementation only works after hydration and not on the server which may lead to issues.
Also the default may now get thrown into the payload which doesn't make sense, I guess?

Maybe it's better to change `useNuxtData` to use `nuxt._asyncData` instead of the payload. Though that would make `useNuxtData` return the default, at times :/
What do you think?

danielroe:
Sorry for the long delay in getting back to you. I don't there's an issue with `useNuxtData` coming up with the default value. (But I do think we should keep the default value out of the payload.)

I'll mark this as draft as it sounds like there's a little more to do here. (And a test would be appreciated.)

logotip4ik:
idk, but this does solve issue for me if page is initially loaded (when `isHydrating` is `true`). But it still breaks with client side navigation (`isHydrating` is `false`, because nuxt tries to get data from `nuxt.static.data` field, but there is none, for non **generated** websites). If you also add fallback for `.static.data` it will solve issue even for client side navigation.

in case any body wondering, here are changed lines in my patch (`asyncData.js` file):

```diff
-  const getCachedData = () => nuxt.isHydrating ? nuxt.payload.data[key] : nuxt.static.data[key];
-  const hasCachedData = () => getCachedData() !== void 0;
+  const getCachedData = () => nuxt.isHydrating ? toRef(nuxt.payload.data, key) : (nuxt.static.data[key] || toRef(nuxt.payload.data, key));
+  const hasCachedData = () => unref(getCachedData()) !== void 0;
```

94726:
> Sorry for the long delay in getting back to you. I don't there's an issue with `useNuxtData` coming up with the default value. (But I do think we should keep the default value out of the payload.)

I updated `useNuxtData` to simply use `_asyncData` instead of the payload. That seemed like the easiest way to ensure synchronization without polluting the payload. The only drawback is that calling `useNuxtData` before `useAsyncData` will lead to skipping the initial `default` as `_asyncData` already exists. Probably not to much of a problem, but maybe could be changed if necessary.

I also included some tests for the synchronization + additional ones for asyncData `default` because there were none and I didn't want to break anything :)

94726:
This should also resolve #22696

metkm:
Any idea when this pr will be merged? This would be a good fix right now!

metkm:
Any idea when this pr will be merged? This would be a good fix and needed I think

94726:
I just fixed the remaining issues, But I'm not sure whether this is the best approach, as it makes all the `_asyncData` fields optional only because `data` is also being controlled by `useNuxtData`.

I also had the idea of splitting `data` into a seperate object and using the fact that `toRef(reactiveObject, 'key')` handles creation/tracking of the key even if it doesn't exist yet. [See example](https://play.vuejs.org/#eNp9Uk1v2zAM/SuELukAL0axnQI32FcP22EbsmGXqQdVZlI1MmVIVJrC8H8vLaNpURS9SXyPT49PHNTnvl8eMqqVapKNrmdIyLkHb2h3oRUnrdaaXNeHyDBARGPZHbACDhvcwgjbGDpYiMRCkyYbKDFQPjJcwKAJwKR7st8Mm3/GZ0yrk0SzQRti2ySOjnYVZNpTuKP1+mwY32kan+Ra6Ra58uLZpL18IVrBYo/35wtpa+p5DDEtF8au94ZRbgDNzfl6GIq5lwIwjk0tcKE56jPD4X0XWvSSwfS8VlDP6HVmDgSfrHd2L+hrcv9nO1diOlOLW0fYSo7Wo4lNPSuIWlM/86cqCVvm3brd8jYFkh8p+WllQ9c7j/FXz07y0Go1Jzthxvtw96PUOGasHuv2Bu3+lfptOk41rX5HTBgPqNUJYxN3yDN8+ecnHuV8AiWM7IX9BrjBFHyePM60LzK72H7GK26/l12SL/+bLo+MlB6HmoxOzLHwtZKd+vrG6E92Pyw/lj5ZGjU+AFUd9iQ=).

Not sure what the better approach is here. Would love to get some feedback on it :)

metkm:
@danielroe I hate to do this but any chance we can get this released on the next version? I think this bug should be a higher priority because using useNuxtData without sync does not make sense

danielroe (approved):

PR #22292, opened 2023-07-23
fix(nuxt): disable View Transitions when `prefers-reduced-motion: reduce`

### LINKED ISSUE #22291: Disable View Transitions when prefers-reduced-motion is reduce
### Describe the feature

Currently the experimental view transition API doesn't honour the [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) user preference. A simple check should be added to the plugin to early return when that media query is set to reduce.

### Additional information

### Final checks

lukewarlow:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22291

### ❓ Type of change

### 📚 Description

View transitions now early return when the `prefers-reduced-motion` user preference is reduce. This improves the out of the box accessibility of the feature.

Resolves #22291 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
@@ -1,0 +1,8 @@
+import { defineNuxtPlugin } from '#app/nuxt'
+import plugin from '#app/plugins/view-transitions.client'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  if (!document.startViewTransition || window.matchMedia('(prefers-reduced-motion: reduce)').matches) { return }
+
+  plugin(nuxtApp)
+})
--- a/packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
@@ -1,8 +1,0 @@
-import { defineNuxtPlugin } from '#app/nuxt'
-import plugin from '#app/plugins/view-transitions.client'
-
-export default defineNuxtPlugin((nuxtApp) => {
-  if (!document.startViewTransition || window.matchMedia('(prefers-reduced-motion: reduce)').matches) { return }
-
-  plugin(nuxtApp)
-})
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -1,9 +1,11 @@
 import { isChangingPage } from '../components/utils'
 import { useRouter } from '../composables/router'
 import { defineNuxtPlugin } from '../nuxt'
+// @ts-expect-error virtual file
+import { viewTransition } from '#build/nuxt.config.mjs'
 
 export default defineNuxtPlugin((nuxtApp) => {
-  if (!document.startViewTransition) { return }
+  if (!document.startViewTransition || (viewTransition !== 'always' && window.matchMedia('(prefers-reduced-motion: reduce)').matches)) { return }
 
   let finishTransition: undefined | (() => void)
   let abortTransition: undefined | (() => void)
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -3,7 +3,7 @@ import { useRouter } from '../composables/router'
 import { defineNuxtPlugin } from '../nuxt'
 
 export default defineNuxtPlugin((nuxtApp) => {
-  if (!document.startViewTransition) { return }
+  if (!document.startViewTransition || window.matchMedia('(prefers-reduced-motion: reduce)').matches) { return }
 
   let finishTransition: undefined | (() => void)
   let abortTransition: undefined | (() => void)
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -3,7 +3,7 @@ import { useRouter } from '../composables/router'
 import { defineNuxtPlugin } from '../nuxt'
 
 export default defineNuxtPlugin((nuxtApp) => {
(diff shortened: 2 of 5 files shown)


packages/nuxt/src/app/plugins/view-transitions.client.ts
```diff
@@ -2,7 +2,7 @@ import { useRouter } from '#app/composables/router'
 import { defineNuxtPlugin } from '#app/nuxt'
 
 export default defineNuxtPlugin((nuxtApp) => {
-  if (!document.startViewTransition) { return }
+  if (!document.startViewTransition || window.matchMedia('(prefers-reduced-motion: reduce)').matches) { return }
```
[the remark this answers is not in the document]
lukewarlow: I couldn't get disabling them in CSS to work myself.

Also view transitions are designed to be a pure enhancement as mentioned in the spec. So I would consider them unnecessary motion when it comes to preferring reduced motion.

Lastly unfortunately a fact of life is that most developers simply won't account for this leading to less accessible sites so it'd be nice for the framework to handle it.

DamianGlowala:
This might help:

lukewarlow:
> This might help:
> 
> https://developer.chrome.com/docs/web-platform/view-transitions/#reacting-to-the-reduced-motion-preference
> 
> > However, a preference for 'reduced motion' doesn't mean the user wants no motion. Instead of the above, you could choose a more subtle animation, but one that still expresses the relationship between elements, and the flow of data.

Ah thanks god knows how I missed that when I read through that post. I was missing the -group pseudo. That appears to be working now.


I still feel it's valuable to have an option to skip the transition (it skips the steps of generating the old and new images of DOM) but I get if you would rather this not be the default. I'm not familar with the plugin API does nuxtApp expose an options object?

lukewarlow:
https://github.com/withastro/astro/pull/7757 fwiw astro appear to have gone the route of disabling view transitions by default (they used CSS instead)

packages/nuxt/src/app/plugins/view-transitions.client.ts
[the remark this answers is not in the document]
danielroe: We could update `experimental.viewTransition` to accept either `true | false | 'always'`. `true` will disable (as above) if `window.matchMedia('(prefers-reduced-motion: reduce)')`. But `always` will allow developers to handle the CSS for reduced motion themselves.

What do you think?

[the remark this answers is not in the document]
lukewarlow: That sounds good I'll make that change later today.

packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
```diff
@@ -0,0 +1,8 @@
+import { defineNuxtPlugin } from '#app/nuxt'
+import plugin from '#app/plugins/view-transitions.client'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  if (!document.startViewTransition || window.matchMedia('(prefers-reduced-motion: reduce)').matches) { return }
```
lukewarlow, line 5: Making a separate safe plugin seemed like the best way to do this? Plugins don't seem to have any access to the nuxt config option?

packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
```diff
@@ -0,0 +1,8 @@
...
+
+  plugin(nuxtApp)
```
lukewarlow, line 7: Should this return the plugin?

packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
danielroe: We can specifically expose it via this template:

https://github.com/nuxt/nuxt/blob/09161d005da48b0ba82308093e509a53d9a8b899/packages/nuxt/src/core/templates.ts#L367-L378

... and then import the value from '#build/nuxt.config.mjs' as we do here:

https://github.com/nuxt/nuxt/blob/d7b11164434e5ae318cd8cf85417362fddb99d71/packages/nuxt/src/app/plugins/payload.client.ts#L7-L8

danielroe: (Alternatively, the approach you've taken here would work too.)

lukewarlow: Is there a preference for one or the other I'm happy to change it?

danielroe: Either way works for me 👍

lukewarlow: Okay in that case this should be ready to go when the team are :)

danielroe: actually, let me backpedal on this.

the reason is that we're doing more build-time optimisations by parsing plugins, and importing and running a plugin might cause issues in future.

if we have one plugin and read the config option from `#build/nuxt.config.mjs` that would be better

lukewarlow: No worries I'll get that changed!

lukewarlow:
Apologies it's taken so long to get back to this. Hopefully the latest commit is correct

packages/nuxt/src/app/plugins/view-transitions-safe.client.ts
lukewarlow: I think this is done correctly now :)

danielroe (approved):

TheAlexLichter (approved):

lukewarlow:
Just double checking is this waiting on me for anything? No worries if it just needs time or other reviews, wanna make sure i'm not the blocker though.

danielroe:
Nope! It's good to go, and I plan to merge it in time for the next minor release.

PR #22301, opened 2023-07-24
feat(nuxt): add `strategy: parallel` option for asyncData

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/12391

### ❓ Type of change

### 📚 Description

This PR adds a new `strategy` option for useAsyncData + useFetch, which can be set to `blocking` (current default behaviour), `lazy` (same as `lazy: true` which I have deprecated in favour of explicit `strategy: lazy`) and a new `parallel` option, which will ensure the promise is resolved before changing the route, but allows it to be run in parallel with other data fetching requests.

In order for parallel strategy to work, the `useAsyncData` composable should not be awaited.

This PR also changes behaviour when `useAsyncData` with current default strategy is not awaited - it will ensure it is awaited before the route changes. I regard this as a bug fix.

**TODO**:

- [ ] add tests to cover new strategy/option

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/route-provider.ts
+++ b/packages/nuxt/src/app/components/route-provider.ts
@@ -62,7 +62,8 @@ export const RouteProvider = defineComponent({
 const ResolvePendingPromises = defineComponent({
   async setup () {
     const nuxtApp = useNuxtApp()
-    await Promise.all(Object.values(nuxtApp._asyncDataPromises).filter(p => p?.strategy !== 'lazy'))
+    const promises = Object.values(nuxtApp._asyncDataPromises).filter(p => p?.strategy !== 'lazy')
+    await (Promise.allSettled?.(promises) ?? Promise.all(promises).catch(() => {}))
   },
   render: () => null
 })
--- a/packages/nuxt/src/app/components/route-provider.ts
+++ b/packages/nuxt/src/app/components/route-provider.ts
@@ -2,6 +2,7 @@ import { defineComponent, h, nextTick, onMounted, provide, shallowReactive } fro
 import type { Ref, VNode } from 'vue'
 import type { RouteLocation, RouteLocationNormalizedLoaded } from '#vue-router'
 import { PageRouteSymbol } from '#app/components/injections'
+import { useNuxtApp } from '#app/nuxt'
 
 export const RouteProvider = defineComponent({
   name: 'RouteProvider',
@@ -50,10 +51,18 @@ export const RouteProvider = defineComponent({
     return () => {
       if (process.dev && process.client) {
         vnode = h(props.vnode, { ref: props.vnodeRef })
-        return vnode
+        return [vnode, h(ResolvePendingPromises)]
       }
 
-      return h(props.vnode, { ref: props.vnodeRef })
+      return [h(props.vnode, { ref: props.vnodeRef }), h(ResolvePendingPromises)]
     }
   }
+})
+
+const ResolvePendingPromises = defineComponent({
+  async setup () {
+    const nuxtApp = useNuxtApp()
+    await Promise.all(Object.values(nuxtApp._asyncDataPromises).filter(p => p?.strategy !== 'lazy'))
+  },
+  render: () => null
 })
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -215,7 +215,7 @@ export function useAsyncData<
         }
         delete nuxt._asyncDataPromises[key]
       })
-    nuxt._asyncDataPromises[key] = promise
+    nuxt._asyncDataPromises[key] = Object.assign(promise, { strategy: options.strategy })
     return nuxt._asyncDataPromises[key]
   }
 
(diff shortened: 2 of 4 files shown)


pi0:
LOVE IT 🔥  (and still didn't read code lol)

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -38,7 +38,9 @@ export interface AsyncDataOptions<
   DefaultT = null,
 > {
   server?: boolean
+  /** @deprecated Use strategy: 'lazy' */
   lazy?: boolean
+  strategy?: 'lazy' | 'parallel' | 'blocking'
```
pi0, line 43: Nice 👍🏼

packages/nuxt/src/app/nuxt.ts
```diff
@@ -103,7 +103,7 @@ interface _NuxtApp {
   [key: string]: unknown
 
   /** @internal */
-  _asyncDataPromises: Record<string, Promise<any> | undefined>
+  _asyncDataPromises: Record<string, Promise<any> & { strategy?: 'lazy' | 'blocking' | 'parallel' } | undefined>
```
QuiiBz, line 106: Small suggestion, this union is duplicated with `AsyncDataOptions` - might be worth to extract both to a type, or get it from the options interface:

```suggestion
  _asyncDataPromises: Record<string, Promise<any> & AsyncDataOptions['strategy']>
```

pi0: We might also make a type export for `AsyncDataStrategy` 👍🏼

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -38,7 +38,9 @@ export interface AsyncDataOptions<
   DefaultT = null,
 > {
   server?: boolean
+  /** @deprecated Use strategy: 'lazy' */
```
Hebilicious, line 41: Maybe we could deprecate the composable too ? 
`useLazyFetch`, `useLazyAsyncData`

packages/nuxt/src/app/components/route-provider.ts
```diff
@@ -50,10 +51,18 @@ export const RouteProvider = defineComponent({
...
-      return h(props.vnode, { ref: props.vnodeRef })
+      return [h(props.vnode, { ref: props.vnodeRef }), h(ResolvePendingPromises)]
     }
   }
 })
+
+const ResolvePendingPromises = defineComponent({
+  async setup () {
+    const nuxtApp = useNuxtApp()
+    await Promise.all(Object.values(nuxtApp._asyncDataPromises).filter(p => p?.strategy !== 'lazy'))
```
Hebilicious, line 65: Doesn't that mean `blocking` and `parallel` have the same behaviour ?

Hebilicious (commented):
Love this change.
However, I have a nitpick about the strategy name : 

Promise.all doesn't run promises in parallel, but concurrently, which is significantly different https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#promise_concurrency

However I could be missing something regarding the implementation ...

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -274,7 +278,7 @@ export function useAsyncData<
 
   // Allow directly awaiting on asyncData
   const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]).then(() => asyncData) as AsyncData<ResT, DataE>
```
DawidKopys, line 280: I believe it should be something like:
```suggestion
  const asyncDataPromise = options.strategy === 'blocking'
    ? Promise.resolve(nuxt._asyncDataPromises[key]).then(() => asyncData) as AsyncData<ResT, DataE>
    : Promise.resolve(asyncData)
```
right?
Then, below line could be restored to `Object.assign(asyncDataPromise, asyncData)`

The `Object.assign` in current form seems to be doing nothing for strategies different than `blocking`. It copies `asyncData` **into** `Promise.resolve(asynData)` object, which is not used anywhere later on.

Anoesj:
Wouldn't some helper utils to create parallel/serial pipelines be nicer? What if you want to do two parallel calls (A + B), but C can start right after A is done and doesn't have to wait for B to finish? Like:
```ts
const route = useRoute() 
const [
  [
    {
      data: aData,
    }, 
    {
      data: cData,
    },
  ], 
  {
    data: bData,
  }, 
] = useParallelFetchers([
  async () => {
    const a = await useFetch(() => `/api/a/${route.params.id}`)
    const c = await useFetch('/api/c', {
      method: 'post',
      body: computed(() => a.data.value.someValue)
    })

    return [a, c]
  }, 
  useFetch('/api/b'),
])
```

(just brainstorming a bit here)

packages/nuxt/src/app/components/route-provider.ts
```diff
@@ -1,70 +1,69 @@
...
-}) as unknown as RouteProviderComponent
+  }
+})
 
-export const RouteProvider: RouteProviderComponent = defineRouteProvider()
+const ResolvePendingPromises = defineComponent({
+  async setup () {
+    const nuxtApp = useNuxtApp()
+    const promises = Object.values(nuxtApp._asyncDataPromises).filter(p => p?.strategy !== 'lazy')
+    await (Promise.allSettled?.(promises) ?? Promise.all(promises).catch(() => {}))
```
cernymatej, line 66: could we await a new hook here instead? 🙏

Issue #13499, opened 2022-03-16
component name visible in chunk names


Aareksio:
### Environment

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.13.2`
- Nuxt Version:     `3.0.0-27456065.f38cace`
- Package Manager:  `npm@8.5.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

```bash
git clone git@github.com:Aareksio/nuxt3-chunk-naming-strategy.git
cd nuxt3-chunk-naming-strategy
npm ci
npm run build
ls -l .output/public/_nuxt/ | grep "my-component"
```

As you can see, a chunk was generated for `my-component`, `my-component-15e4fa4c.mjs` in my case.

### Describe the bug

The chunk name may be blocked by adblocker software, some keywords commonly blocked:

- banner -> hero-banner.vue
- tracking -> usage-tracking-widget.vue
- ad -> ad-space.vue

Failing to load **any** of those chunks results in complete app failure. Not only the blocked component, but whole app becomes unusable.

### Additional context

Escalating from [discussion](https://github.com/nuxt/framework/discussions/3378) as it did not get any traction and the issue is critical for any production deployment. Read the discussion post for additional context if necessary.

danielroe:
Yes, you can pick your chunk names:

```ts
export default defineNuxtConfig({
  vite: {
    build: {
      rollupOptions: {
        output: {
          chunkFileNames: '_nuxt/[hash].js',
          assetFileNames: '_nuxt/[hash][extname]',
          entryFileNames: '_nuxt/[hash].js'
        }
      }
    }
  }
})
```

bharatmajority:
The entryFileNames seems to break the application. 

`Error: Could not load /..../.nuxt/dist/server/server.mjs (imported by node_modules/nuxt/dist/core/runtime/nitro/renderer.js): ENOENT: no such file or directory, open '/..../.nuxt/dist/server/server.mjs'`

chunk and asset file names seem to work just fine. We don't really need to change the entryFileNames. Just bringing up the error for your reference @danielroe .

manniL:
@danielroe Should we close this issue or do you think there is more discussion needed?

danielroe:
I think no more discussion is needed. We could document the solution for those for whom it matters.

Aareksio:
[...]

PR #22302, opened 2023-07-24
feat(schema,vite): allow `$client` and `$server` vite env overrides

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/13499#issuecomment-1643476119

### ❓ Type of change

### 📚 Description

This allows configuring client/server configuration for vite directly in the Nuxt config, without needing to do so in a hook. This is pure convenience, but avoids issues where setting a vite configuration for the _client_ ends up having negative implications for the server configuration, as in the linked issue above.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -10,7 +10,7 @@ export default defineUntypedSchema({
    * See https://vitejs.dev/config for more information.
    * Please note that not all vite options are supported in Nuxt.
    *
-   * @type {typeof import('../src/types/config').ViteConfig}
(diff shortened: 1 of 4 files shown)


pi0:
This is a nice idea for env (client/server) specific configuration. However, I'm a little bit concerned about making it "vite" specific. It could be extended to other bundlers (webpack) but also possibly for edge cases such as app config which is client/server specific. I think we can move forward with this and always alias only to mention about possibility of a (more generic) feature.

danielroe:
Yes, I think we should support webpack out of the box in this PR 👍

danielroe:
@pi0 Any changes you would suggest? I don't think there's anything blocking us from refactoring to a native c12 feature if we end up building it.

pi0 (approved):

PR #22307, opened 2023-07-24
fix(nuxt): test `watch` paths against all layer `srcDir`s

### LINKED ISSUE #22272: `nuxt.options.watch` not restarting dev server
### Environment

```shell
npx nuxi info
Nuxi 3.6.2                                                                                                                                                                16:43:20
                                                                                                                                                                          16:43:20
RootDir: D:/work/projects/userquin/GitHub/userquin/vuetify-nuxt-module-main-2                                                                                             16:43:20
Nuxt project info:                                                                                                                                                        16:43:20

- Node Version:     v18.16.0
- Nuxt Version:     3.6.2
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.6.9
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://github.com/userquin/vuetify-nuxt-module/tree/userquin/feat-add-external-vuetify-config-file

Remove the following lines:
- https://github.com/userquin/vuetify-nuxt-module/blob/userquin/feat-add-external-vuetify-config-file/src/utils/layers.ts#L29-L30
- https://github.com/userquin/vuetify-nuxt-module/blob/userquin/feat-add-external-vuetify-config-file/src/utils/layers.ts#L53-L54

Remove the following hook:
- https://github.com/userquin/vuetify-nuxt-module/blob/userquin/feat-add-external-vuetify-config-file/src/module.ts#L67-L70

Register the paths in `nuxt.options.watch` (uncomment next line):
- https://github.com/userquin/vuetify-nuxt-module/blob/userquin/feat-add-external-vuetify-config-file/src/module.ts#L66

Install dependencies `pnpm install` and run dev server `nr dev`, open the app in the browser, once the app loaded, change any vuetify config file:
- `playground/vuetify.config.mts`: change `MyBadge: 'VBadge',` to `MyXXBadge: 'VBadge',`
- or `playground/layers/vuetify-layer/vuetify.config.ts`: change `MyAvatar: 'VAvatar',` to ` MyXXAvatar: 'VAvatar',`

Once changed, the dev server should restart, but it doesn't.

### Describe the bug

When adding files to watch in `nuxt.options.watch`, they should be relative to the `srcDir`, but the `path` received in the callback seems to be always absolute paths, and so there is no dev server restart.

### Additional context

image

### Logs

_No response_

userquin:
### 🔗 Linked issue

closes #22272

### ❓ Type of change

### 📚 Description

`watch` option will be matched against the file path relative to the project `srcDir`, this PR normalize the `path` received in `builder:watch` to be relative to project `srcDir` checking also against the normalized path.

### 📝 Checklist


packages/nuxt/src/core/nuxt.ts
```diff
@@ -347,13 +347,16 @@ async function initNuxt (nuxt: Nuxt) {
       return nuxt.callHook('restart', { hard: true })
     }
 
-    // User provided patterns
+    // User provided patterns:
+    // - nuxt.options.watch relative to srcDir
```
danielroe, line 351: we should also normalise the paths that are provided in the schema to absolute paths (`resolve(await get('srcDir'), val)`, which will make for a better solution for layers:

https://github.com/nuxt/nuxt/blob/03abb6aa4a879c2d3b97595cd3880c682c180dc2/packages/schema/src/config/common.ts#L373-L375

userquin: The normalized check is required for Regex, for example, my playground config is (I have one.ts, the-two.ts and two.ts in root with dymmy content):
```ts
  watch: ['one.ts', /^two/]
})
```

If we change `two.ts`, we must normalize it relative to `srcDir`, it will not match the regex.

userquin: builder module is also normalizing the layers watch option to be relative to project `srcDir`, may we should revert my last commit:

https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/builder.ts#L98-L103

userquin: I need to add a check, the path should be inside the layer `srcDir`

userquin:
So the problem is the watch is merged when using local layers, but the merge is using the relative path from project `srcDir` and so any watch configured inside the layer will not be matched, you need to add the watch entries from the layer in the project watch adding the relative path (for example `../layer-1/a.ts`  or `./layer-2/b.ts`: these 2 paths will be included as `a.ts` and `b.ts` and will never match the corresponding layer watch).

If we use the watch from the layers previous problem is solved, since the path will be relative to the layer `srcDir`.

packages/schema/src/config/common.ts
```diff
@@ -371,7 +371,9 @@ export default defineUntypedSchema({
    * @type {Array<string | RegExp>}
    */
   watch: {
-    $resolve: val => [].concat(val).filter((b: unknown) => typeof b === 'string' || b instanceof RegExp),
+    $resolve: async (val, get) => [].concat(val).filter((b: unknown) => typeof b === 'string' || b instanceof RegExp).map(
+      async (b: string | RegExp) => typeof b === 'string' ? resolve(await get('srcDir'), b) : b
+    )
```
danielroe, line 376: we should await the values _here_ rather than when `options.watch` is used.

```suggestion
    $resolve: async (val, get) => {
      return Promise.all([].concat(val).filter((b: unknown) => typeof b === 'string' || b instanceof RegExp).map(
        async (b: string | RegExp) => typeof b === 'string' ? resolve(await get('srcDir'), b) : b
      ))
	  }
```

danielroe:
On investigation, the change to absolute paths was a regression, now fixed in https://github.com/nuxt/nuxt/pull/22333. I think that means we should normalise watched paths to absolute in schema (as you're doing), and then _re-normalise_ them to absolute and compare them against absolute path in `nuxt.ts`.

I _think_ we shouldn't need to do this per layer, but directly using `nuxt.options.watch` because if we normalise in the schema, the layer watch paths should already be resolved in relation to the layer directory.

Would you be able to update the PR, or would you prefer me do it?

userquin:
> Would you be able to update the PR, or would you prefer me do it?

I'll try...

danielroe (approved):

PR #22316, opened 2023-07-25
fix(kit): handle tuple-format modules in `hasNuxtModule`

harlan-zw:
### ❓ Type of change

### 📚 Description

The `hasNuxtModule` function is incorrectly returning false for the following conditions:

- module with inline options
- module instance (with and without inline options)

```ts
  modules: [
    ['my-module', { foo: 'bar' }],
   defineNuxtModule({ name: 'my-module' }),
   [defineNuxtModule({ name: 'my-module' }), { foo: 'bar' }],
  ]
})
```

Additionally, `getNuxtModuleVersion` is failing when using any of the above.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/compatibility.ts
+++ b/packages/kit/src/module/compatibility.ts
@@ -24,7 +24,7 @@ export function hasNuxtModule (moduleName: string, nuxt: Nuxt = useNuxt()) : boo
   // check installed modules
   return nuxt.options._installedModules.some(({ meta }) => meta.name === moduleName) ||
     // check modules to be installed
-    nuxt.options.modules.some(m => resolveNuxtModuleEntryName(m))
+    nuxt.options.modules.some(m => moduleName === resolveNuxtModuleEntryName(m))
 }
 
 /**
--- a/packages/kit/src/module/compatibility.ts
+++ b/packages/kit/src/module/compatibility.ts
@@ -11,8 +11,20 @@ import { loadNuxtModuleInstance } from './install'
  * that it cannot detect if a module is _going to be_ installed programmatically by another module.
  */
 export function hasNuxtModule (moduleName: string, nuxt: Nuxt = useNuxt()) : boolean {
+  // check installed modules
   return nuxt.options._installedModules.some(({ meta }) => meta.name === moduleName) ||
-    nuxt.options.modules.includes(moduleName)
+        // check modules to be installed
+        Boolean(
+          nuxt.options.modules
+            .find((m) => {
+              // input may either a string, an array or a module instance
+              function resolveModuleEntry (input: typeof m): boolean {
+                if (typeof input === 'object') { return (input as any as NuxtModule).name === moduleName }
+                return Array.isArray(input) ? resolveModuleEntry(input[0]) : input === moduleName
+              }
+              return resolveModuleEntry(m)
+            })
+        )
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/kit/src/module/compatibility.ts
```diff
@@ -11,8 +11,22 @@ import { loadNuxtModuleInstance } from './install'
...
+              // input may either a string, an array or a module instance
+              function resolveModuleEntry (input: typeof m): boolean {
+                if (typeof input === 'object' && !Array.isArray(input)) {
+                  return (input as any as NuxtModule).name === moduleName
+                }
+                return Array.isArray(input) ? resolveModuleEntry(input[0]) : input === moduleName
+              }
+              return resolveModuleEntry(m)
+            })
+        )
```
danielroe, line 29: Maybe we could use `modules.some`?

harlan-zw: Always forget about some, thanks!

packages/kit/src/module/compatibility.ts
```diff
@@ -11,8 +11,20 @@ import { loadNuxtModuleInstance } from './install'
...
  */
 export function hasNuxtModule (moduleName: string, nuxt: Nuxt = useNuxt()) : boolean {
+  // check installed modules
   return nuxt.options._installedModules.some(({ meta }) => meta.name === moduleName) ||
-    nuxt.options.modules.includes(moduleName)
+    // check modules to be installed
+    nuxt.options.modules
+      .some((m) => {
+        // input may either a string, an array or a module instance
+        function resolveModuleEntry (input: typeof m): boolean {
```
danielroe, line 20: Do you think we can hoist this out of the inline function?

harlan-zw: Good idea :+1:

packages/kit/src/module/compatibility.ts
```diff
@@ -1,18 +1,30 @@
...
  *
  * This will check both the installed modules and the modules to be installed. Note
  * that it cannot detect if a module is _going to be_ installed programmatically by another module.
  */
 export function hasNuxtModule (moduleName: string, nuxt: Nuxt = useNuxt()) : boolean {
+  // check installed modules
   return nuxt.options._installedModules.some(({ meta }) => meta.name === moduleName) ||
-    nuxt.options.modules.includes(moduleName)
+    // check modules to be installed
+    nuxt.options.modules.some(m => resolveNuxtModuleEntryName(m))
```
danielroe, line 27:
```suggestion
    nuxt.options.modules.some(m => moduleName === resolveNuxtModuleEntryName(m))
```

packages/kit/src/module/compatibility.ts
```diff
@@ -1,18 +1,30 @@
 import satisfies from 'semver/functions/satisfies.js' // npm/node-semver#381
-import type { Nuxt, NuxtModule } from '@nuxt/schema'
+import type { Nuxt, NuxtModule, NuxtOptions } from '@nuxt/schema'
 import { useNuxt } from '../context'
 import { normalizeSemanticVersion } from '../compatibility'
 import { loadNuxtModuleInstance } from './install'
 
+export function resolveNuxtModuleEntryName (m: NuxtOptions['modules'][number]): string | false {
```
danielroe, line 7:
```suggestion
function resolveNuxtModuleEntryName (m: NuxtOptions['modules'][number]): string | false {
```

danielroe (approved):

PR #22317, opened 2023-07-25
feat(kit): get module options with `getNuxtModuleOptions`

harlan-zw:
### ❓ Type of change

### 📚 Description

When integrating with other modules, it's common to use some of their configuration. Relying on just the module config key isn't safe, instead, we introduce a utility to extract the options in a safer way.

For example, within Nuxt Simple Sitemap:

```ts
if (hasModule('@nuxtjs/i18n')) {
  const i18nVersion = await getNuxtModuleVersion('@nuxtjs/i18n')
  if (!await hasNuxtModuleCompatibility('@nuxtjs/i18n', '>=8'))
    logger.warn(`You are using @nuxtjs/i18n v${i18nVersion}. For the the best compatibility, please upgrade to @nuxtjs/i18n v8.0.0 or higher.`)
  const nuxtI18nConfig = await getNuxtModuleOptions('@nuxtjs/i18n') as NuxtI18nOptions
  if (nuxtI18nConfig?.pages) {
    // do logic
  }
}
```

### 📝 Checklist


harlan-zw:
This PR should also look to investigate if it's feasible to use this function internally to resolve the options for a module, so that we know the logic is consistent. This would also let us implement a caching layer to reduce the time needed to setup modules.

AndreyYolkin:
Nice! I'll track this PR to include it in new kit documentation after merge

danielroe:
@harlan-zw Is this still something you are planning on working on?

harlan-zw:
I don't have the capacity to see if we can use this function internally, so we can either close this, merge it as is, or someone else wants to take it on.

PR #22336, opened 2023-07-26
refactor(nuxi,schema): move loading template into schema

### LINKED ISSUE #14353: Allow for specification of UI templates module when using nuxi dev command
The `dev` command defined in the `nuxi` package currently imports a loading template from `@nuxt/ui-templates`. I wish to have this as a default but to have the ability to provide my own package for development purposes.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14353

### ❓ Type of change

### 📚 Description

This change is partly necessitated by https://github.com/nuxt/nuxt/issues/14146. Rather than loading the loading ui template dynamically within nuxi, which has no declared peer dep on it, I think we can move this into schema directly, which (as a side effect) allows customising the loading template by end users.

This also means we can decouple the template from nuxi itself.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -46,7 +46,7 @@ export default defineNuxtCommand({
     let currentHandler: RequestListener | undefined
     let loadingMessage = 'Nuxt is starting...'
     const loadingHandler: RequestListener = async (_req, res) => {
-      const { loading: loadingTemplate } = await importModule('@nuxt/ui-templates', config.modulesDir)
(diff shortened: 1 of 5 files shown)


pi0:
It is a good idea to remove the loading template as a bundled dependency of nuxt. But i am thinking why not make it a dependency of nuxt package? nuxi anyway requires to resolve `nuxt` as only guaranteed dependency from users rootDir. By exposing it from nuxt package, we are more sure to be pnpm compatible.

danielroe:
It is a dependency of Nuxt, as well.

pi0 (approved):
LGTM. However, it shouldn't be a required dependency of schema IMO. Like every peer dep, it should come from nuxt and CLI resolve it on _behalf_ of nuxt package but it is also a safe thing to do..

PR #22362, opened 2023-07-27
feat(nuxt): autocomplete layouts in `setPageLayout`/`<NuxtLayout>`

DamianGlowala:
### ❓ Type of change

### 📚 Description

Similarily to the `definePageMeta`'s layout property, `setPageLayout` could hint available layouts to the user, rather than accepting an arbitrary string.

### 📝 Checklist


danielroe (approved):

DamianGlowala:
WDYT about providing autocompletion for the `<NuxtLayout />` component as well? 😉

danielroe:
Oh, that would be great!

DamianGlowala:
Tried adding autocomplete for the `<NuxtLayout />`. I'm not sure why importing `PageMeta` from `#app` doesn't work in this case. Also, I left `MaybeRef<false>` instead of `Ref<false>` (even though `PageMeta` already contains `false` in its type, to make it `PageMeta`-independent.

I have also renamed the file from `layout.ts` to `nuxt-layout.ts` for consistency (and relevant import path). Let me know if there's any reason to keep it as it was, I'll revert it back.

danielroe:
This looks like it's working well to me:

CleanShot 2023-07-30 at 23 52 45@2x

For linting, note that it should be run after `pnpm build:stub` and not after building the project.

danielroe (approved):

PR #22384, opened 2023-07-29
perf(nuxt): write templates in single sync step + improve logs

danielroe:
### ❓ Type of change

### 📚 Description

This PR bundles three performance-focused improvements to template compilation and writing:

1. it writes all template files in synchronous step to avoid additional runtime overhead of cascading HMRs from vite/webpack
2. It logs errors granularly rather than hard erroring out, doesn't block the start of Nuxt, and makes it clearer what the issue was
   <table><tbody><tr><td>CleanShot 2023-07-29 at 05 14 47@2x</td><td>CleanShot 2023-07-29 at 05 15 07@2x</td></tbody></table>
3. It logs the time each template takes to compile in debug mode (or if any template takes more than 500ms - we could potentially lower this)

### 📝 Checklist


packages/nuxt/src/core/app.ts
```diff
@@ -49,12 +55,23 @@ export async function generateApp (nuxt: Nuxt, app: NuxtApp, options: { filter?:
...
-        await fsp.mkdir(dirname(fullPath), { recursive: true })
-        await fsp.writeFile(fullPath, contents, 'utf8')
+        writes.push(() => {
+          mkdirSync(dirname(fullPath), { recursive: true })
+          writeFileSync(fullPath, contents, 'utf8')
+        })
       }
     }))
 
+  for (const write of writes) { write() }
```
pi0, line 73: AFAIK, using sync methods, blocks (our) main js render loop but it won't speed up v8 i/o threads. Have you tried with this new batching technique ++ async `Promise.all` to see timing differences?

danielroe: I've done some tests but the performance benefit I'm looking for here is reducing cascading hmr.

pi0: Yes, I understand that main point 👍🏼 But i suppose (native) HMR watchers are not in the same event loop of main javascript process we are writing files from (unless our receiver is sync again). If either method is faster, it reduces final chance to have duplicate HMR events but if you already tried both, feel free to ignore this comment :)

danielroe: it may be an unnecessary optimisation. I tried both without appreciable time difference, and flagged up the fact that we are actually not watching the `buildDir` with vite. I will have a look to see if we actually _should_ be watching the directory. Not doing so might be causing some occasionally issues with vite-node cache, for example.

antfu (approved):
That makes sense to me, I think it's worth trying. Shall we better leave a comment inside to explain it's on purpose?

packages/nuxt/src/core/app.ts
danielroe: I think you have a good point here and we should definitely iterate on this. I'd be very open to rethinking this if you notice any issue..

PR #22389, opened 2023-07-29
docs: escape `]` in code-block filenames

nobkd:
### ❓ Type of change

### 📚 Description

Adds `\\` to every code-block filename with `]` to escape it. This prevents the filename from being cropped.

---

#### Related & Info

See this PR https://github.com/nuxt/content/pull/2169, released in `nuxt/content` since [`v2.7.2`](https://github.com/nuxt/content/releases/tag/v2.7.2), to fix this with an escape character

The fix was already shipped in [Docus `v1.14.4`](https://github.com/nuxt-themes/docus/releases/tag/v1.14.4) with https://github.com/nuxt-themes/docus/commit/44aefd545b170f105b525a619a3f3475ae25e1f0, **but it seems, that this version isn't used here currently.**
So if this is accepted, before the deps use the latest Docus, the filenames may look like in the following example with an extra `\` before cutting off the filename.

---

#### Current Example

Here is an example image of how it currently looks, without escaping:

image

The route, where it is from: https://nuxt.com/docs/guide/directory-structure/server#using-a-nested-router

### 📝 Checklist


danielroe:
Checking the live preview, this now doesn't look quite right:

CleanShot 2023-07-30 at 10 39 04@2x

The tricky thing is that the website isn't yet open sourced but atinux is on that.

nobkd:
As mentioned, this'll look wrong until Docus v1.14.4 is used.

danielroe:
Nuxt.com doesn't use docus, and has already updated to the latest Nuxt content version.

nobkd:
Thanks for clearing that up.
Seems, that there's a different issue

Barbapapazes:
> Thanks for clearing that up. Seems, that there's a different issue in Nuxt Content

Ah!

Barbapapazes:
In fact, there is the issue even in the content.nuxtjs.org website. 

image

nobkd:
Created an issue https://github.com/nuxt/content/issues/2201

danielroe:
Thank you ❤️

danielroe (approved):

Issue #22449, opened 2023-08-02
nuxt generate command takes lot of memory


parzychu:
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.17.0`
- Nuxt Version:     `3.6.5`
- Nitro Version:    `2.5.2`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `css`, `plugins`, `app`, `postcss`, `runtimeConfig`, `modules`, `graphqlMiddleware`, `vite`, `nitro`, `experimental`, `ssr`, `vue`, `schemaOrg`
- Runtime Modules:  `nuxt-graphql-middleware@3.0.2`, `@nuxt/content@2.5.2`, `@pinia/nuxt@0.4.9`, `nuxt-schema-org@2.2.0`
- Build Modules:    `-`

### Reproduction

I've created a small application that shows the problem with keeping memory for long time:
https://github.com/parzychu/nuxt-memory-reproduction

Please see it's README file where I put more description than in this issue and a way to run it.

To get out of memory error you need application with more components and about 2.5k statically generated pages. 

### Describe the bug

When using `nuxi generate` application holds memory for longer than I think it should and in application having more than 2k pages it can cause out of memory issue. 

See log from generation of application from [reproduction repo](https://github.com/parzychu/nuxt-memory-reproduction)  [demo.log](https://github.com/parzychu/nuxt-memory-reproduction/blob/master/demo.log).
If you search for `/page3652` you will see that setup of this page was run at `2:14`, page was generated at `2:16`, but payload was extracted at `9:00`. I guess that after payload extraction memory frees, but in between over 6600 pages were generated and their content was added to memory.

This delay is visible in memory usage even for demo application. For my production application difference is much more evident and 4GB RAM is used after payload extraction of 260 pages when 2500 were generated. I need about 11GB of RAM to successfully render about 5k pages.

Situation with memory is the same when I turn off payloadExtraction. I guess that payloadExtraction starts generating just before page generation process is finished, but in the meantime its data is stored in memory.

Situation with memory slightly improves (generates 40-50% more pages), but is dozen times slower when I set `"nitro.prerender.concurrency": 1`. But then in my application case pages that were rendered within first 15 min had payload extracted 4 hrs later.

[...]

PR #22465, opened 2023-08-03
perf(nuxt): use LRU prerender cache and limit cached items

danielroe:
### 🔗 Linked issue

related https://github.com/nuxt/nuxt/issues/22449

### ❓ Type of change

### 📚 Description

This uses nitro storage to cache island/payload responses for prerendering. To reduce memory usage, we use an LRU cache. (It would be possible for a user to customise the maxSize in a hook.) Because we have no guarantee responses are preserved, we also need to cache props (in query string) of island renders so we can get these later when the island renders with only its hash.

### 📝 Checklist


packages/nuxt/src/core/nitro.ts
```diff
@@ -273,6 +273,12 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Init nitro
   const nitro = await createNitro(nitroConfig)
 
+  // Set prerender-only options
+  nitro.options._config.storage ||= {}
+  nitro.options._config.storage['internal:nitro:prerender'] = { driver: 'memory' }
+  nitro.options._config.storage['internal:nitro:prerender:island'] = { driver: 'lruCache' }
+  nitro.options._config.storage['internal:nitro:prerender:payload'] = { driver: 'lruCache' }
```
pi0, line 280: Since we are mounting and consuming them in nuxt maybe use `internal:nuxt`? Also is default lru params enough? (an explicit option could be nice).

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -157,7 +157,12 @@ const getSPARenderer = lazyCachedFunction(async () => {
 
 async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
   // TODO: Strict validation for url
-  const url = event.node.req.url?.substring('/__nuxt_island'.length + 1) || ''
+  let url = event.node.req.url || ''
+  if (process.env.prerender && event.node.req.url && await useStorage('internal:nitro:prerender:island-props').hasItem(event.node.req.url)) {
+    // rehydrate props from cache so we can rerender island if cache does not have it any more
+    url = await useStorage('internal:nitro:prerender:island-props').getItem(event.node.req.url) as string
```
pi0, line 163: Maybe use prefixed storages to keep code readable?

pi0 (commented):
Love it!

packages/nuxt/src/core/runtime/nitro/renderer.ts
danielroe: ahh ... it does actually increase the size.

pi0 (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -273,6 +273,12 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Init nitro
   const nitro = await createNitro(nitroConfig)
 
+  // Set prerender-only options
+  nitro.options._config.storage ||= {}
+  nitro.options._config.storage['internal:nitro:prerender'] = { driver: 'memory' }
+  nitro.options._config.storage['internal:nitro:prerender:island'] = { driver: 'lruCache', max: 1000 }
+  nitro.options._config.storage['internal:nitro:prerender:payload'] = { driver: 'lruCache', max: 1000 }
```
pi0, line 280: Oh my bad approved too fast! This didn't updated

PR #22474, opened 2023-08-03
fix(nuxt): correctly update object cookies in `useCookie`

### LINKED ISSUE #22001: useCookie doesn't change value of an object stored in cookie
### Environment

- Node Version:     v18.16.0
- Nuxt Version:     3.6.2
- Nitro Version:    2.5.2
- Package Manager:  npm@9.5.1
- Builder:          vite
- User Config:      runtimeConfig, app, modules, bugsnag, pwa, cookieControl, pinia, components, css
- Runtime Modules:  @nuxtjs/tailwindcss@6.8.0, @pinia/nuxt@0.4.11, @vueuse/nuxt@10.2.1, @dargmuesli/nuxt-cookie-control@5.10.3, @vite-pwa/nuxt@0.0.9
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-hunkey

### Describe the bug

When I store an object in a cookie using `useCookie` composable and then want to change the value of a key of that object it doesn't effect the cookie.

### Additional context

I tried Nuxt 3.5.3 and it worked.

### Logs

_No response_

pgraug:
### 🔗 Linked issue

Resolves #22001

### ❓ Type of change

### 📚 Description

Fixes a bug where mutating a nested property on a `useCookie` object doesn't update the actual cookie. Only the `ref`.

The bug was introduced in PR #20970 where an `isEqual(newVal, oldVal)` eject was added inside the `useCookie` watcher.

Problem is that with nested property mutations, `newVal` and `oldVal` are always identical in `watch` callbacks, as mentioned in [Vue's docs](https://vuejs.org/guide/essentials/watchers.html#deep-watchers):

> Note: `newValue` will be equal to `oldValue` here because they both point to the same object!

Therefore I just removed that `isEqual` and now it works great!
I was worried it would result in `callback()` getting called multiple times when using multiple `useCookie`s with the same name but `watchPaused` seems to prevent that just fine on its own.

### 📝 Checklist


danielroe (approved):

AkhrorbekDev:
Hi there. In my use Cookies situation, don't stop writing the value in cookies
`watch(useCookie('cart-products-v3', { watch: false }), (newVal) => {
    ids.value = newVal
    if (newVal) {
      ...do something
    } else {
      ...do something
    }
  },
  {
    immediate: false,
  })`

PR #22478, opened 2023-08-03
feat(nuxt): resolve unresolved paths within `node_modules`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/14146

### ❓ Type of change

### 📚 Description

This adds a resolve step to resolve paths deeply within `node_modules` which cannot otherwise be resolved.

In my testing, this enables use with vite and without shamefully-hoist, as long as `vue` and `vue-router` are explicitly installed (as they are in vite's `optimizeDeps.include`).

### 📝 Checklist


packages/nuxt/src/core/plugins/resolve-deep-imports.ts
```diff
@@ -0,0 +1,25 @@
...
+    name: 'nuxt:resolve-bare-imports',
+    enforce: 'post',
+    async resolveId (id, importer, options) {
+      if (!importer || isAbsolute(id) || !isAbsolute(importer) || id.startsWith('virtual:')) {
+        return
+      }
+      const { dir } = parseNodeModulePath(importer)
+      return await this.resolve?.(id, dir || pkgDir, { skipSelf: true }) ?? await resolvePath(id, {
+        url: [dir || pkgDir, ...nuxt.options.modulesDir],
+        conditions: options.ssr ? ['node', 'import', 'require'] : ['import', 'require']
```
pi0, line 18: Note: With nitro, we use runtime environment condition sets. Not sure if we can easily get resolve conditions from nitro now but a TODO would be nice to consider

packages/nuxt/src/core/plugins/resolve-deep-imports.ts
```diff
@@ -0,0 +1,25 @@
...
+      }).catch(() => {
+        console.log('could not resolve id', id, importer)
```
pi0, line 20:
```suggestion
        console.log('[nuxt] Could not resolve id', id, importer)
```

packages/vite/src/server.ts
```diff
@@ -57,7 +57,13 @@ export async function buildServer (ctx: ViteBuildContext) {
       }
     },
     ssr: {
-      external: ['#internal/nitro', '#internal/nitro/utils', ...nitroDependencies],
+      external: [
+        '#internal/nitro', '#internal/nitro/utils',
+        // explicit dependencies we use in our ssr renderer - these can be inlined (if necessary) in the nitro build
+        'unhead', '@unhead/ssr', '@unhead/vue', 'unctx', 'h3', 'devalue', '@nuxt/devalue',
```
pi0, line 63: We also have radix3 and unstorage. Maybe we can include them here and also think of a export from nitro for them?

packages/nuxt/src/core/plugins/resolve-deep-imports.ts
danielroe, line 18:
```suggestion
        // TODO: respect nitro runtime conditions
        conditions: options.ssr ? ['node', 'import', 'require'] : ['import', 'require']
```

packages/nuxt/src/core/plugins/resolve-deep-imports.ts
```diff
@@ -0,0 +1,30 @@
...
+      }
+      id = normalize(id)
+      id = resolveAlias(id, nuxt.options.alias)
+      const { dir } = parseNodeModulePath(importer)
+      return await this.resolve?.(id, dir || pkgDir, { skipSelf: true }) ?? await resolvePath(id, {
+        url: [dir || pkgDir, ...nuxt.options.modulesDir],
+        // TODO: respect nitro runtime conditions
+        conditions: options.ssr ? ['node', 'import', 'require'] : ['import', 'require']
+      }).catch(() => {
+        console.log('[nuxt] Could not resolve id', id, importer)
```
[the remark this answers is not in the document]
danielroe: That makes sense to me.

PR #22521, opened 2023-08-07
feat(nuxt): warn on external `webpack`/`nitro`/`postcss` config

Lehoczky:
### 🔗 Linked issue

Well not an issue, but: https://youtu.be/a2r8Zr1hnEw?t=4166

I can create an issue too though:D

### ❓ Type of change

### 📚 Description

The main change is that with this PR, we check whether the user has a `webpack.config`, `postcss.config`, `nitro.config` file and not just a `vite.config` file.

image
image

For better performance, this check runs parallel to the bundling process, see [this line](https://github.com/nuxt/nuxt/compare/main...Lehoczky:nuxt:main?expand=1#diff-1383cf7dbb577617bc4b0fa583e7c0cb08b0ca403582069ffc6bf2dd5316ae31R47).

One caveat is that regardless of whether the user has chosen webpack or vite as a builder, we will always look for both of their config files (for example a warning will be made about an existing webpack config file, even if the user uses vite). I thought this is a pretty unlikely scenario, but if needed, I can adjust the code to take into consideration the used builder as well. It wouldn't be such a big change.

As a small additional improvement, the `loadBuilder()` function is now more type safe, and gracefully handles when a builder cannot be resolved:

Before:
image

After:
image

### 📝 Checklist


packages/nuxt/build.config.ts
```diff
@@ -27,6 +27,7 @@ export default defineBuildConfig({
     'ofetch'
   ],
   externals: [
+    'consola',
```
Lehoczky, line 30: The build failed with the error: 
> │ - Inlined implicit external consola
> │  ERROR  Exiting with code (1). You can change this behavior by setting failOnWarn: false .

I wasn't sure whether I should add consola here or set `failOnWarn: false` to the build config to solve this.

danielroe: instead we should add `consola` as a dependency

danielroe (approved):
This is nice 👌

PR #22528, opened 2023-08-07
feat(nuxt): make `app.rootId` optional

DamianGlowala:
### ❓ Type of change

### 📚 Description

This PR allows setting `app.rootId` to `false`, which mounts the Vue app to the `<body>`'s direct descendant `app.rootTag` without the ID (should users didn't have a need to inject one).

I am not sure whether this has any negative implications, but at least it's an idea worth considering.

### 📝 Checklist


packages/nuxt/src/app/entry.ts
```diff
@@ -75,7 +75,7 @@ if (process.client) {
     try {
       await nuxt.hooks.callHook('app:created', vueApp)
       await nuxt.hooks.callHook('app:beforeMount', vueApp)
-      vueApp.mount('#' + appRootId)
+      vueApp.mount(appRootId ? `#${appRootId}` : `body > ${appRootTag}`)
```
danielroe, line 78: it would be nice not to include this code in every app for this edge case. maybe we could use a flag to ensure no extra code is added in default case?

danielroe (approved):

PR #22531, opened 2023-08-08
docs: update api example for `pick`

iiio2:
### ❓ Type of change

### 📚 Description

when we use `pick` with `useFetch`, then it works only on single document. it does not work on array of objects. in the doc, there were some error in the api that i fixed by making it single document.

This is required because it fixes the functionality of `useFetch` while using `pick`.

### 📝 Checklist


danielroe (approved):

danielroe:
Thank you ❤️

PR #22546, opened 2023-08-08
chore(schema): enable linting

DamianGlowala:
### ❓ Type of change

### 📚 Description

This PR enables linting for `packages/schema` and applies suggestions, as well as addresses errors resulting from this change.

closes https://github.com/nuxt/nuxt/pull/22544

### 📝 Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,4 @@
 dist
 node_modules
-schema
 **/*.tmpl.*
 sw.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 dist
 node_modules
+packages/schema/schema
 **/*.tmpl.*
 sw.js
--- a/packages/schema/src/types/nuxt.ts
+++ b/packages/schema/src/types/nuxt.ts
@@ -5,54 +5,54 @@ import type { Component } from './components'
 import type { NuxtOptions } from './config'
 
 export interface NuxtPlugin {
-    /** @deprecated use mode */
-    ssr?: boolean
-    src: string
-    mode?: 'all' | 'server' | 'client'
-    /**
-     * This allows more granular control over plugin order and should only be used by advanced users.
-     * Lower numbers run first, and user plugins default to `0`.
-     *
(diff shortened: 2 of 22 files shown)


.eslintignore
```diff
@@ -1,5 +1,4 @@
 dist
 node_modules
-schema
 **/*.tmpl.*
```
danielroe, line 3: I think this was meant to ignore `packages/schema/schema` but of course we do want to include `packages/schema` - maybe refine it instead?

```suggestion
schema/schema
**/*.tmpl.*
```

DamianGlowala: Good point! I'll add `packages/schema/schema` for explicitness to the `.eslintignore` (even though it only contains .d.ts, .md and .json, which are not linted, but surely it's better not to rely on file extensions).

packages/schema/src/types/nuxt.ts
```diff
@@ -4,6 +4,57 @@ import type { NuxtHooks, NuxtLayout, NuxtMiddleware } from './hooks'
...
+    ssr?: boolean
+    src: string
+    mode?: 'all' | 'server' | 'client'
+    /**
+     * This allows more granular control over plugin order and should only be used by advanced users.
+     * Lower numbers run first, and user plugins default to `0`.
+     *
+     * Default Nuxt priorities can be seen at [here](https://github.com/nuxt/nuxt/blob/9904849bc87c53dfbd3ea3528140a5684c63c8d8/packages/nuxt/src/core/plugins/plugin-metadata.ts#L15-L34).
+     */
+    order?: number
```
danielroe, line 18:
```suggestion
  /** @deprecated use mode */
  ssr?: boolean
  src: string
  mode?: 'all' | 'server' | 'client'
  /**
   * This allows more granular control over plugin order and should only be used by advanced users.
   * Lower numbers run first, and user plugins default to `0`.
   *
   * Default Nuxt priorities can be seen at [here](https://github.com/nuxt/nuxt/blob/9904849bc87c53dfbd3ea3528140a5684c63c8d8/packages/nuxt/src/core/plugins/plugin-metadata.ts#L15-L34).
   */
  order?: number
```

DamianGlowala: Wondering the same, as I tried changing quotes to double quotes, and it correctly detected the issue. Looking at the `ssr`'s indent, for some reason it seems to be fine...

image

DamianGlowala: It appears to have something to do with the folder name `types` this file is in.

PR #22558, opened 2023-08-09
feat(nuxt): allow experimental `global: 'sync'` components

danielroe:
### ❓ Type of change

### 📚 Description

In general, we want to discourage global components because it will inflate bundle size. Libraries or users could go and register all components of a ui library as global as an easy choice, with the downside of hugely inflating bundle size.

So, we would prefer if people use auto import so only what they need is required (and can be bundle-split appropriate).

When global components _are_ used, we register them as async components and the chunks only get fetched when used.

However, there is a significant downside when these components are small and used frequently as this approach produces many small chunks. One option is to provide a custom chunk handler to rollup, though this can only really be done by an end user and not by an integration.

This PR provides an escape hatch for edge cases (like `@nuxt/content`) which want to register components globally and _synchronously_ in the main bundle.

**Note**: Within `nuxt/content`, I would advise using this this an experiment only. I would prefer to see compilation of markdown into vue (or server components, or partial compilation of just the tags) so auto-import/bundle-splitting can be used rather than increasing the _entry_ bundle size. For a large site using `nuxt/content`, if all components used in content across all pages are registered synchronously, I would expect to see very poor initial load performance scores.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/templates.ts
+++ b/packages/nuxt/src/components/templates.ts
@@ -49,7 +49,6 @@ export const componentsPluginTemplate: NuxtPluginTemplate<ComponentsTemplateCont
 
     const lazyComponents = [...lazyGlobalComponents]
     const syncComponents = [...syncGlobalComponents]
-    console.log({ lazyComponents, syncComponents })
 
     return `import { defineNuxtPlugin } from '#app/nuxt'
 import { ${[...lazyComponents.map(c => 'Lazy' + c), ...syncComponents].join(', ')} } from '#components'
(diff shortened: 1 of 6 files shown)


pi0 (approved):
LGTM as experiment and agree with points (both downside and requirement to reduce lazy chunks)

pi0:
> . One option is to provide a custom chunk handler to rollup, though this can only really be done by an end user and not by an integration.

I am wondering why is that? Can't module configure rollup bundler and also use magic comments to hint chunk groups?

danielroe:
Modules can configure rollup bundler but the function wouldn't be merged. So it would conflict with a custom chunk function provided by user. Or by another module. We could potentially provide a _hook_ that would allow modules hooking into a single function that nuxt provides, though.

pi0:
Supporting a hook to make chunk handling strategy could be nice! (webpack has a really good API too it could be unified)

antfu (approved):

PR #22593, opened 2023-08-11
chore: depend on `h3-nightly` on edge releases

danielroe:
### ❓ Type of change

### 📚 Description

This also pins h3 on edge releases.

### 📝 Checklist

### DIFF
--- a/scripts/bump-edge.ts
+++ b/scripts/bump-edge.ts
@@ -15,10 +15,10 @@ async function main () {
   const commit = execSync('git rev-parse --short HEAD').toString('utf-8').trim().slice(0, 8)
   const date = Math.round(Date.now() / (1000 * 60))
 
-  const nuxtPkg = workspace.find('nuxt')
+  const resolutions: Record<string, string> = {}
   for (const [pkg, name] of Object.entries(nightlyPackages)) {
     const { version: latest } = await $fetch<{ version: string }>(`https://registry.npmjs.org/${name}/latest`)
-    nuxtPkg.data.dependencies[pkg] = `npm:${name}@^${latest}`
+    resolutions[pkg] = `npm:${name}@^${latest}`
   }
 
   const bumpType = await determineBumpType()
@@ -28,6 +28,14 @@ async function main () {
     workspace.setVersion(pkg.data.name, `${newVersion}-${date}.${commit}`, {
       updateDeps: true
     })
+    for (const [name, version] of Object.entries(resolutions)) {
+      if (name in pkg.data.dependencies) {
+        pkg.data.dependencies[name] = version
+      }
+      if (name in pkg.data.devDependencies) {
+        pkg.data.dependencies[name] = version
+      }
+    }
     const newname = pkg.data.name === 'nuxt' ? 'nuxt3' : (pkg.data.name + '-edge')
     workspace.rename(pkg.data.name, newname)
   }
--- a/scripts/bump-edge.ts
+++ b/scripts/bump-edge.ts
@@ -1,5 +1,4 @@
 import { execSync } from 'node:child_process'
-import { $fetch } from 'ofetch'
 import { inc } from 'semver'
 import { determineBumpType, loadWorkspace } from './_utils'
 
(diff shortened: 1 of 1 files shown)


scripts/bump-edge.ts
```diff
@@ -3,18 +3,23 @@ import { $fetch } from 'ofetch'
...
-  const nuxtPkg = workspace.find('nuxt')
-  const { version: latestNitro } = await $fetch<{ version: string }>('https://registry.npmjs.org/nitropack-edge/latest')
-  nuxtPkg.data.dependencies.nitropack = `npm:nitropack-edge@^${latestNitro}`
-
-  const { version: latestNuxi } = await $fetch<{ version: string }>('https://registry.npmjs.org/nuxi-ng/latest')
-  nuxtPkg.data.dependencies.nuxi = `npm:nuxi-ng@^${latestNuxi}`
+  const resolutions: Record<string, string> = {}
+  for (const [pkg, name] of Object.entries(nightlyPackages)) {
+    const { version: latest } = await $fetch<{ version: string }>(`https://registry.npmjs.org/${name}/latest`)
+    resolutions[pkg] = `npm:${name}@^${latest}`
```
pi0, line 21: I am thinking to use `h3-nightly@latest` (same for nitropack) to remove requirement of two time releases nitro>nuxt (and if both are not released we might have conflicting h3 nightly versions by each)

danielroe: A couple of issues:

* see [pnpm resolution mode](https://pnpm.io/npmrc#resolution-mode) - default behaviour best case is that `latest` will be resolved to be latest version when package was published (same as hash here)
* in some cases it can be the latest version _already known about_ on user computer, ie. not actually latest. by using caret + hash we _force_ update to at least that version, but allow it to be newer

pi0: I am not sure if Caret is respected for edge versions with `-` suffix. Also main issue is that nitro and nuxt requesting different nightly version of h3 (this is issue as of today since nitro requires h3 with `^1.8.0-rc.3` but that is also not respected. Also notice that we should use `"h3": "npm:..."` directly in dependencies.

danielroe: > Also notice that we should use "h3": "npm:..." directly in dependencies.

What do you mean?

pi0: On other solution would be that nuxt pulls out the `h3` version from latest `nitropack-edge` version by the time of publishing nuxt-edge but it is really not good honestly because everytime i want to rest nitropack, i have to manually add nitropack resolution too.

(avoiding `latest` was my initial intention that wrote this script with fetch.. but it proven to be less effective)

danielroe: The caret is respected for edge versions: see https://stackblitz.com/edit/github-fwjf8z?file=package.json,pnpm-lock.yaml

pi0: I still think latest is the right choice because technically versioning (endge edge versioning) of nuxt, nitropack and h3 is decoupled. I hope you respect this decision from the parson that is mainly maintaining both of them. This helps me much better in triage and when asking people to use nuxt-edge instead of additional steps for h3, nitropack additional resolution.

danielroe: have you tested using `latest` in nuxt package? I don't think it will solve this issue.

pi0: I (re)tested with an older caret h3 in nitro. pnpm resolves to exactly the same nightly version (it is a valid behavior from pnpm. versions with `-` suffix are exact match) not latest (by the time f3d0bc9)

image

in order to avoid hosting, both nitro-edge and nuxt3 need same version range request. And (sadly) `^1` is not also possible with edge:

image

danielroe: Okay - let's give it a try. I think pnpm will resolve to latest version as at time of when the nightly version was published but hopefully if the constraints are the same that will avoid different edge versions 🤞

pi0: Sure! we might give also dashless naming (h3-nightly@1.8.1001) a try later on to keep on semver-major at least but appreciated moving this forward 🙏🏼

PR #22614, opened 2023-08-13
fix(nuxi): only watch `distDir` after first build

wattanx:
### 🔗 Linked issue
https://github.com/nuxt/bridge/issues/832

### ❓ Type of change

### 📚 Description
nuxt 2 seems to delete buildDir at build time.
https://github.com/nuxt/nuxt/blob/2.x/packages/builder/src/builder.js#L144

I think this problem is occurring because they are executed in the following order.
1. watch for distDir(`.nuxt/dist`)
2. delete buildDir(`.nuxt`)

reproduction
https://stackblitz.com/edit/github-3fgb91-mgegmc
It occurs when I stop the dev server once and run `npm run dev` again.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -134,11 +134,6 @@ export default defineNuxtCommand({
 
         await currentNuxt.ready()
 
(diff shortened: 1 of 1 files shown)


wattanx:
None of this appears to be due to this correction.

danielroe (approved):

PR #22649, opened 2023-08-15
feat(nuxt): allow client components within `NuxtIsland`

huang-julien:
### 🔗 Linked issue

### ❓ Type of change

### 📚 Description

Hi :wave: This PR finally introduce client components within island components.

## How does it works ?
This still use `Teleport` and the same trick of reactivity from client-side slots. 

- simply mark a component with a `nuxt-client` prop/attr -> `<SugarCounter :multiplier nuxt-client />`

## How to enable it

This feature is behind a flag: `experimental.componentIslands.selectiveClient` (i'm open for any other name like `nuxtClient`, `selectiveHydration` ?) which needs to be set to true

## About remote islands
- When using a `NuxtIsland` with remote islands (it's the `source` prop introduced in Nuxt 3.7), you need to set `dangerouslyAllowClientComponents` to `true` in order to load client components for security reasons (you should'nt trust a remote island)
`<NuxtIsland source="some-source.com" name="SomeComponent" dangerouslyAllowClientComponents />`.
- Remote islands can only provide Vue components as client components
- ~~Allow deeply selective hydration within server component (.server -> component -> component -> nuxt-client)~~ later

## About `nuxt-client`
Since we use `nuxt-client` to determine wich component should be loaded client-side, this means that this cannot be used as attribute by users (we remove it). 

## About payload
Since we cannot only rely on what is received by the HTML during hydration, this means we need to retrieve some additionnal payloads:
- props
- chunks
- teleports

## What's next

- A refactor of server components 
- A breaking change (if you're using remote islands) will probably come in a close future to settle a definitive API for islands components so other frameworks or servers (looking at you CMS 👀) can have a stable implementation to provide island components to Nuxt. 

## Todo (by priorities) - it's working but some small improvement can be done

- [x] remove `nuxt-client` attr from the client component
- [x] improve payload
- [ ] ~~filter teleports (in the island response) if needed~~
- [x] improve chunks (by using `input` or `manualChunk`)
  - [ ] ~~IF using `manualChunk` instead of `rollupOptions.build.input` to chunk components, make `manualChunk` hookable so users can extend it~~
  - [x] try to avoid making a single chunk for lazy components
  - [x] if going for `input`, fix the client/ssr styles
- [x] tests, unit and fixtures
  - [x] fixtures
  - [x] unit
- [x] update doc
- [ ] ~~Make the unplugin compatible with webpack (if possible :( ).~~ (later....probably)
- [x] some rename + clean code

### 📝 Checklist


huang-julien:
Feel free to ping me if you have any question about this 🙂. I think this is ready, i'll keep playing with it until it's being merged.

huang-julien (commented):
now thinking of it, should we make this configurable ? 

What about

````ts
type ComponentIslandConfig = {
   // setting to true doesn't enable remote not client components
    componentsIsland: {
       remote: boolean
       // deep for deep interactive component (in the future)
       // true doesn't enable it
       clientComponent: boolean | { deep: true }
    }
} | boolean
````

instead of `boolean | 'locale+remote'`

File: packages/nuxt/src/components/islandsTransform.ts [hunk 1]
```diff
@@ -1,86 +1,122 @@
...
+  const components = options.getComponents()
+  const isVite = meta.framework === 'vite'
+  const { isDev, rootDir } = options
   return {
     name: 'server-only-component-transform',
     enforce: 'pre',
     transformInclude (id) {
       if (!isVue(id)) { return false }
 
-      const components = options.getComponents()
```
danielroe, line 24: IIRC I think this is needed here if the the components _change_ after the start of Nuxt.

packages/schema/src/config/experimental.ts
```diff
@@ -142,13 +142,17 @@ export default defineUntypedSchema({
 
     /**
      * Experimental component islands support with <NuxtIsland> and .island.vue files.
-     * @type {true | 'local' | 'local+remote' | false}
+     * @type {true | Partial<{ remoteIsland: boolean, selectiveClient: boolean }> | false}
```
danielroe, line 145:
```suggestion
     * @type {true | 'local' | 'local+remote' | Partial<{ remoteIsland: boolean, selectiveClient: boolean }> | false}
```

test/basic.test.ts
```diff
@@ -1457,16 +1469,39 @@ describe('server components/islands', () => {
...
 
+  it('should not preload ComponentWithRef', async () => {
+    // should not add <ComponentWithRef> to the modulepreload list since it is used only server side
+    const { page } = await renderPage('/islands')
+    const links = await page.locator('link').all()
+    for (const link of links) {
+      if (await link.getAttribute('rel') === 'modulepreload') {
+        expect(await link.getAttribute('href')).not.toContain('ComponentWithRef')
+      }
+    }
```
danielroe, line 1490:
```suggestion
    }

    await page.close()
```

danielroe (approved):

PR #22685, opened 2023-08-17
docs: mention `NuxtIsland` in server only components docs

huang-julien:
### ❓ Type of change
Hi :wave: 

this is a small improvement of server components documentation so users understand that server only components are island components

### 📝 Checklist

### DIFF
--- a/docs/2.guide/2.directory-structure/1.components.md
+++ b/docs/2.guide/2.directory-structure/1.components.md
@@ -299,9 +299,9 @@ Now you can register server-only components with the `.server` suffix and use th
 </template>
 ```
 
-Server only components uses `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
+Server-only components use `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
 
-This also means that server only components are accessible as remote sources and that every features of `<NuxtIsland>` are available on server only components.
+This also means that server-only components are accessible as remote sources and have all features of the `<NuxtIsland>` component available.
 
 ::alert{type=info}
 Slots can be interactive and are wrapped within a `<div>` with `display: contents;`
--- a/docs/2.guide/2.directory-structure/1.components.md
+++ b/docs/2.guide/2.directory-structure/1.components.md
@@ -299,7 +299,7 @@ Now you can register server-only components with the `.server` suffix and use th
 </template>
 ```
 
-Server only components uses `<NuxtIsland>` under the hood, meaning `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
+Server only components uses `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
 
 This also means that server only components are accessible as remote sources and that every features of `<NuxtIsland>` are available on server only components.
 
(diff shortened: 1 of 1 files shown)


docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -299,6 +299,10 @@ Now you can register server-only components with the `.server` suffix and use th
 </template>
 ```
 
+Server only components uses `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
```
DamianGlowala, line 302:
```suggestion
Server-only components use `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
```

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -299,6 +299,10 @@ Now you can register server-only components with the `.server` suffix and use th
...
+
+This also means that server only components are accessible as remote sources and that every features of `<NuxtIsland>` are available on server only components.
```
DamianGlowala, line 304:
```suggestion
This also means that server-only components are accessible as remote sources and have all features of the `<NuxtIsland>` component available.
```

DamianGlowala (changes requested):

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -299,6 +299,10 @@ Now you can register server-only components with the `.server` suffix and use th
 </template>
 ```
 
+Server-only components use `<NuxtIsland>` under the hood, meaning that `lazy` prop and `#fallback` slot are both passed down to `<NuxtIsland>`.
+
+This also means that server-only components are accessible as remote sources and have all features of the `<NuxtIsland>` component available.
+
```
danielroe, line 305: I don't think we currently support specifying sources for server components directly.

```suggestion
```

huang-julien: It's technically possible, I think this is a bug.

I also think we should hide server components from being easily accessible by NutxIsland.

https://stackblitz.com/edit/nuxt-starter-pa96dq?file=nuxt.config.ts,components%2FHello.server.vue,components%2FWorld.server.vue,app.vue

There's also a conflict when naming both island and `.server` by the same name

danielroe: Oh, I mean the other way around. If you're using a server component - you can't specify the remote source via `source` prop. (Maybe we can reword this sentence instead?)

I think we need NuxtIsland to be able to render arbitrary server components.

Ultimately we'll need 'rules' for rendering islands/server components remotely (that is, which ones to do so, and setting CORS headers for them, where appropriate) and (for example) exposing/exporting component types for remotely rendered components.

But the name conflict sounds like something we should render better errors for.

huang-julien: I see :eyes: ...

We can't specify source on `<AsyncServerComponent>`.

So 
- about rules and CORS on server only components. What about a `defineIslandRules` macro ? with `remote` to `true` by default for `island` and `false` for `.server`. I was thinking about hiding `.server` by hashing their names but I don't know if users think that `.server` should be accessible remotely or not
- about the name conflict, a warning log should be enough 🤔 for now ?
- about types, that means we'd need to expose components types in build too ?

danielroe: We should discuss types/CORS more but I would disable remote source _consumption_ by default - I would say it's opt-in.

I think a warning log for name should be enough (if there's a conflict), as we currently do when there are multiple components with the same name.

danielroe (approved):

PR #22708, opened 2023-08-19
perf: Optimizing Performance and Resource Management with Concurrent Execution using `Promise.all` 🚀

sanjaiyan-dev:
I believed that executing async functions using `Promise.all` would lead to better performance and resource management, as it allows the code to run concurrently. For more information, please refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#description.

I apologize if I've made any mistakes :(

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -204,6 +204,8 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
     }
   })
 
-  await buildClient(ctx)
-  await buildServer(ctx)
+  await Promise.all([
+    buildClient(ctx),
+    buildServer(ctx),
+  ])
 }


Kasparas-G55:
Please try to understand the code before changing something like that, like why are the client and server are built sequentially. 😕

One example is in `buildServer(ctx)` it calls `writeManifest(ctx)`,

https://github.com/nuxt/nuxt/blob/09971c62a0d66e44f30329e1c581a7f271b00e62/packages/vite/src/server.ts#L128-L140

which reads the clients `manifest.json` file

https://github.com/nuxt/nuxt/blob/09971c62a0d66e44f30329e1c581a7f271b00e62/packages/vite/src/manifest.ts#L30-L32

which won't exist if they run in parallel because `buildClient(ctx)` most likely won't have enough time to build.

danielroe:
@Kasparas-G55 is right here.

I also see some context by reading https://github.com/vercel/next.js/pull/53617#issuecomment-1666610953 - that's good advice.

Having said that, I'd love to see future PRs from you. ❤️

sanjaiyan-dev:
> Please try to understand the code before changing something like that, like why are the client and server are built sequentially. 😕
> 
> One example is in `buildServer(ctx)` it calls `writeManifest(ctx)`,
> 
> https://github.com/nuxt/nuxt/blob/09971c62a0d66e44f30329e1c581a7f271b00e62/packages/vite/src/server.ts#L128-L140
> 
> which reads the clients `manifest.json` file
> 
> https://github.com/nuxt/nuxt/blob/09971c62a0d66e44f30329e1c581a7f271b00e62/packages/vite/src/manifest.ts#L30-L32
> 
> which won't exist if they run in parallel because `buildClient(ctx)` most likely won't have enough time to build.

I am extremely sorry 😞

sanjaiyan-dev:
> @Kasparas-G55 is right here.
> 
> I also see some context by reading [vercel/next.js#53617 (comment)](https://github.com/vercel/next.js/pull/53617#issuecomment-1666610953) - that's good advice.
> 
> Having said that, I'd love to see future PRs from you. ❤️

🤗 sorry for the inconvenience

PR #22725, opened 2023-08-21
fix: useRouter() fallback to useRouter() from 'vue-router'

chakAs3:
### ❓ Type of change

- [ x ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ x ] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

useRouter should fall-back to `useRouter` from `vue-router`, as that will be cases where it is not defined

Maybe this is not a usual case for normal Nuxt Project, but when i was implementing Nuxt framework project in Storybook this occurs i fixed by a workaround which set the $router manually before registering plugins

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -13,7 +13,7 @@ import type { PageMeta } from '#app'
 import { PageRouteSymbol } from '#app/components/injections'
 
 export const useRouter: typeof _useRouter = () => {
-  return useNuxtApp()?.$router as Router
+  return useNuxtApp()?.$router ?? _useRouter() as Router
 }
 
 export const useRoute: typeof _useRoute = () => {


packages/nuxt/src/app/composables/router.ts
```diff
@@ -13,7 +13,7 @@ import type { PageMeta } from '#app'
 import { PageRouteSymbol } from '#app/components/injections'
 
 export const useRouter: typeof _useRouter = () => {
-  return useNuxtApp()?.$router as Router
+  return useNuxtApp()?.$router ?? _useRouter() as Router
```
danielroe, line 16: The main issue here is that this will add `useRouter` to the build and not allow it being tree-shaken out if `vue-router` isn't otherwise used.

chakAs3: ah ok i did not pay attention that useRouter imported as type only. 
in this case we can lazy loaded like this ?
`const { useRouter } = await import("#vue-router") `

danielroe: unfortunately that would make it asynchronous ...

markpinero: I've run into this issue as well. In a Storybook story, `NuxtLink` will cause this error as it has no access to the Nuxt instance outside of the main application:

```
Error: [nuxt] instance unavailable
  at useRouter (/node_modules/nuxt/dist/app/composables/router.js?v=2fa1c944:8:10))
```

danielroe: I think this should be the responsibility of the storybook module to provide a mocked nuxtApp. It's expected that Nuxt composables expect that nuxtApp is defined and available in their context.

PR #22768, opened 2023-08-23
fix(nuxt): add `exclude` paths to nitro `tsconfig.server.json`

huseeiin:
Fixed `Cannot write file '/Users/{USERNAME}/nuxt-app/dist/_nuxt/A.vue.62ba23ef.js' because it would overwrite input file.`

### ❓ Type of change

### 📝 Checklist


packages/nuxt/src/core/nitro.ts
```diff
@@ -112,10 +112,14 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
       strict: true,
       generateTsConfig: true,
       tsconfigPath: 'tsconfig.server.json',
-      tsConfig: {
-        include: [
-          join(nuxt.options.buildDir, 'types/nitro-nuxt.d.ts')
-        ]
+      typescript: {
```
danielroe, line 115: I think you may accidentally have introduced an extra layer of nesting.

danielroe (changes requested):

packages/nuxt/src/core/nitro.ts
```diff
@@ -112,10 +112,14 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+        strict: true,
+        generateTsConfig: true,
+        tsconfigPath: "tsconfig.server.json",
+        tsConfig: {
+          include: [join(nuxt.options.buildDir, "types/nitro-nuxt.d.ts")],
+          exclude: ["../dist", "../.output"]
```
danielroe, line 121: We also will need to exclude node_modules if explicitly mentioning exclude paths

xxxhussein: I'm unfamiliar with the repo. I just saw this bug while developing a nuxt app. You probably can make those exclude paths better :)

danielroe (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -115,6 +115,11 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
       tsConfig: {
         include: [
           join(nuxt.options.buildDir, 'types/nitro-nuxt.d.ts')
+        ],
+        exclude: [
+          ...nuxt.options.modulesDir.map(m => relativeWithDot(nuxt.options.buildDir, m)),
```
pi0, line 120: Note: Consider `modulesDir` can technically be normal directories with parent/sibiling `node_modules` dir) as well. We might have filtered out the ones not ending with `node_modules` here for greater safety.

danielroe: That's a nice idea. I've just copied across our current default in the nuxt tsconfig so if we filter it further we should update them in parallel.

PR #22787, opened 2023-08-24
feat(schema): support setting `hidden` sourcemaps

### LINKED ISSUE #19253: No way to set sourcemap.client = hidden
### Environment

No need

### Reproduction

No need

### Describe the bug

`vite` allows to set sourcemap to `boolean | 'inline' | 'hidden'` https://vitejs.dev/config/build-options.html#build-sourcemap
But in `nuxt.config.ts` it is possible to set only to boolean. 
It would be great to implement at least `'hidden'`.
For webpack can be used `hidden-source-map` https://webpack.js.org/configuration/devtool/#production

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19253

### ❓ Type of change

### 📚 Description

We now support setting `hidden` as a possible option for sourcemaps since this is now supported in Nitro. Vite already supported it and we now also set [appropriate sourcemap settings](https://webpack.js.org/configuration/devtool/) in webpack.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/module.ts
+++ b/packages/nuxt/src/components/module.ts
@@ -208,16 +208,16 @@ export default defineNuxtModule<ComponentsOptions>({
       config.plugins = config.plugins || []
       if (nuxt.options.experimental.treeshakeClientOnly && isServer) {
         config.plugins.push(TreeShakeTemplatePlugin.vite({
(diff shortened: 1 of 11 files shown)


pi0 (approved):
👍🏼

PR #22798, opened 2023-08-24
fix(nuxt): default `spaLoadingTemplate` to `false`

danielroe:
### ❓ Type of change

### 📚 Description

This is a behaviour change from v3.5 - we now default the Nuxt logo SPA loading template to `false`. Despite being beautiful, probably users would want to remove this before deploying their Nuxt sites. If you want to keep the previous behaviour, you can instead set `spaLoadingTemplate` to `true.

### 👉 Migration

```ts
  ssr: false,
  spaLoadingTemplate: true
})
```

### 📝 Checklist


pi0:
Disabling by default is probably better than current state 👍 However it might affect users experience when there is a SPA error, they will only see a white screen (also when bundle is too big/connection is slow) We might also differ to later but any possible fallback to indicate long loadings and errors..

Issue #22801, opened 2023-08-24
Volar not able to infer referenced component types in `<template>`


rchl:
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.19.0`
- Nuxt Version:     `2.17.1`
- Nitro Version:    `-`
- Package Manager:  `npm@9.8.1`
- Builder:          `webpack`

### Reproduction

I haven't been able to make a minimal reproduction based on the proprietary project.

### Describe the bug

Volar can't recognize types of components defined in the project:

![Screenshot 2023-08-24 at 22 48 16](https://github.com/nuxt/nuxt/assets/153197/9f76ab01-df1c-462e-abd0-5ab3ff902146)

It seems like the issue is that `declare module 'vue/types'` overrides something in vue types. Which I guess makes sense because `defineComponent` functions are defined in [vue/types/v3-define-component](https://github.com/vuejs/vue/blob/main/types/v3-define-component.d.ts) and not `vue/types`.

I don't fully understand why the current code works for the `script` tag (it's tested by tests) but not for the `template` tag but I suppose it has to do with how Volar does things.

Changing to `declare module 'vue/types/v3-define-component'` makes things work:

![Screenshot 2023-08-24 at 22 53 52](https://github.com/nuxt/nuxt/assets/153197/5ffa4a70-4dba-46e9-9374-b887bf8d5173)

PR #22802, opened 2023-08-24
fix(types): prevent overwriting vue types in template

rchl:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22801

### ❓ Type of change

### 📚 Description

See the issue.

It looks like this code was just plain wrong from the beginning (I've introduced it) and it worked for `<script>` tags but not for `<template>`.

I don't fully understand why the current code works for the script tag (it's tested by tests) but not for the template tag but I suppose it has to do with how Volar does things.

### 📝 Checklist


danielroe (dismissed):

rchl:
Just to add a bit more information here - the original code seemed to break this interface extend:

```ts
import 'vue';

declare module 'vue' {
    export interface GlobalComponents {
        MyComponent: typeof import('./components/my-component.vue').default;
        // ...
    }
}
```

which makes volar just see `GlobalComponents` as `any`.
Not sure why it doesn't reproduce in my simple project but I suppose it depends on what types are evaluated and in what order.

danielroe:
Would be nice to add a `vue-tsc` typecheck to the 2.x branch to catch this.

rchl:
Would you like that to be a part of this PR?

I'm a bit skeptical about that because Volar/vue-tsc is not shy of breaking things (on purpose or not) so it could contribute to instability of the CI.

danielroe:
That would be great, if possible! I realise this is a straight bugfix but would be nice to catch any regression later.

I'm not worried about CI instability because that will reflect real user instability, and will give us warning so we can alert Johnson and hopefully resolve.

danielroe (approved):

danielroe:
Beautiful. Thank you ❤️

PR #22808, opened 2023-08-25
feat(kit,nuxt): respect nuxt ignore patterns in nitro

### LINKED ISSUE #13427: `.nuxtignore` support for server middleware & static assets
we still need to implement for Nitro scanned middleware & assets. It would make sense to wait until https://github.com/nuxt/nuxt.js/issues/13360 is closed.

_Originally posted by @danielroe in https://github.com/nuxt/framework/issues/3424#issuecomment-1054390636_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/13427

### ❓ Type of change

### 📚 Description

We now have support for ignore patterns in nitro (https://github.com/unjs/nitro/pull/945 and https://github.com/unjs/nitro/pull/1430).

This PR now passes on nuxt's ignore settings to nitro so public assets, api handlers, etc. can be ignored in the same configuration.

related: https://github.com/unjs/nitro/pull/1632

### 📝 Checklist


packages/kit/src/ignore.ts
```diff
@@ -35,6 +28,31 @@ export function isIgnored (pathname: string): boolean {
...
+
+    const nuxtignoreFile = join(nuxt.options.rootDir, '.nuxtignore')
+    if (existsSync(nuxtignoreFile)) {
+      const contents = readFileSync(nuxtignoreFile, 'utf-8')
+      nuxt._ignorePatterns.push(...contents.trim().split('\n'))
+    }
+  }
+
+  if (relativePath) {
+    return nuxt._ignorePatterns.map(p => p.startsWith('*') || p.startsWith('!*') ? p : relative(relativePath, resolve(nuxt.options.rootDir, p)))
```
pi0, line 50: I haven't test locally but since tests passing, i think rootDir (vs ./server) is working well with nitro ignore patterns 👍🏼

packages/nuxt/src/core/nitro.ts
```diff
@@ -207,6 +207,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
 
   // Resolve user-provided paths
   nitroConfig.srcDir = resolve(nuxt.options.rootDir, nuxt.options.srcDir, nitroConfig.srcDir!)
+  nitroConfig.ignore = resolveIgnorePatterns(nitroConfig.srcDir)
```
pi0, line 210: Shouldn't we extend/merge if user also provides `nitro.ignore`? (thinking there might be cases people only need nitro ignore patterns)

danielroe: oh, quite right 👍

packages/kit/src/ignore.ts
danielroe: what do you mean? are you suggesting a change?

pi0: i mean should be safe as long as tests are happy!

PR #22810, opened 2023-08-25
chore: add `.devcontainer`

pi0:
### ❓ Type of change

### 📚 Description

Dev containers allow to locally develop and test nitro in isolated (Linux) docker containers.

Made this while locally testing cli issue with linux env. Might be worth to keep this setup in repo (also helps for windows users having a linux env)

https://code.visualstudio.com/docs/devcontainers/containers

Related experiment for nitro: https://github.com/unjs/nitro/pull/1633

### 📝 Checklist


danielroe (approved):
lovely 👌

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:18
```
danielroe, line 1: what about an evergreen tag like:

```suggestion
FROM node:latest
```

pi0: We might also use `lts` might be safer

PR #22815, opened 2023-08-25
docs: update nuxt bridge migration guide

wattanx:
### ❓ Type of change

### 📚 Description
I think Nuxt Bridge to be a module that assists with the migration to Nuxt 3.
I would like to update the Nuxt Bridge Migration Guide to make it easier to understand how migration to Nuxt 3 can proceed using Nuxt Bridge.

### 📝 Checklist


docs/6.bridge/5.nuxt3-compatible-api.md
```diff
@@ -1,112 +1,10 @@
-# Composition API
+# Nuxt 3 compatible API
```
danielroe, line 1: What about something like this to make it clearer it's in opposition to 'Legacy Composition API'?

```suggestion
# New Composition API
```

danielroe (approved):
This is really good. Only one suggestion, but the rest looks 👌

docs/6.bridge/5.nuxt3-compatible-api.md
wattanx: I think it's good!

danielroe (approved):

Issue #22847, opened 2023-08-26
typed layout name isn't working


94726:
### Environment

- Operating System: Linux
- Node Version:     v18.16.0
- Nuxt Version:     3.7.0
- CLI Version:      3.7.0
- Nitro Version:    2.6.1
- Package Manager:  pnpm@8.6.8
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/sandbox/reproduction-nuxt-layout-name-types-w5gj9c?file=%2Fapp.vue%3A4%2C23-4%2C42

### Describe the bug

Nuxt 3.7 introduced a typed layout name, added in https://github.com/nuxt/nuxt/pull/22362.
Unfortunately, it doesn't seem to work :(.
`name` of layout is always typed as `string`, even though a correct `LayoutKey` is available.

### Additional context

Looks like the types aren't preserved after build.

**source types**: https://github.com/nuxt/nuxt/blob/424dc8d7c0448b4bdc3051791adbeed15907220f/packages/nuxt/src/app/components/nuxt-layout.ts#L83-L85

**bundled types**: https://www.runpkg.com/?nuxt@3.7.0/dist/app/components/nuxt-layout.d.ts

PR #22855, opened 2023-08-27
fix(nuxt): ensure typed layout prop persists through build

94726:
### 🔗 Linked issue

#22847

### ❓ Type of change

### 📚 Description

The issue started because of 0d9a0b75387ac376a2c77b2739415093027771fc. For some reason the `?` makes it so that the type is changed. I have no idea why but this seems to fix it. Would appreciate an explaination if someone knows why :)
To me, it looks like a bug of unbuild/dts generation.
Maybe not a bug per se, but it looks like it tries to evaluate the type at build time and in this case probably doesn't find PageMeta?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-layout.ts
+++ b/packages/nuxt/src/app/components/nuxt-layout.ts
@@ -81,7 +81,7 @@ export default defineComponent({
     }
   }
 }) as unknown as DefineComponent<{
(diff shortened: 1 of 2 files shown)


danielroe (approved):
This looks good. You're right that it's an upstream issue in `mkdist` that we likely need to resolve there as well. Likely will be solved in https://github.com/unjs/mkdist/pull/154.

PR #22863, opened 2023-08-28
feat(nuxt): add `prerenderRoutes` ssr composable

danielroe:
### ❓ Type of change

### 📚 Description

This adds a new composable that will hint to nitro to prerender a path. If we need to change the internal header or make changes, this should make it easier to do so.

Alternatively, we could locate this composable within Nitro auto-imports and re-export in Nuxt so there's at least a stub function on client. Thoughts welcome @pi0.

### 📝 Checklist


docs/3.api/3.utils/prerender-path.md
```diff
@@ -0,0 +1,18 @@
+---
+description: prerenderPath hints to Nitro to prerender an additional route.
+---
+# `prerenderPath`
```
Atinux, line 4:
```suggestion
---

# `prerenderPath`
```

Atinux:
For consistency with `nitro.prerender.routes` I would rename it to `prerenderRoute` or `addPrerenderRoute`.

The `add` prefix lets us know it's adding it to a list where I except `prerenderRoute()` to be async and returns me the HTML.

danielroe:
Good idea. Will also refactor to accept an array.

danielroe:
Maybe `queuePrerenderRoute`?

pi0:
I like it! Let's keep it close to Nuxt for now since you might want to make variations in API (but also fits for a Nitro built-in composable PR welcome 👍🏼 )

I personally like `prerenderRoute` since is a straight forward name and does not imply how we do prerender (enqueue or immediate) and is open to changes.

Arthur-Chevron:
> Good idea. Will also refactor to accept an array.

Atinux:
Alright, then let's go fo `prerenderRoute(string)`

PR #22889, opened 2023-08-29
fix(nuxt): load layer plugins before project plugins

### LINKED ISSUE #22414: Plugin Registration Order with layers
## The Environment
3 layers
1. @nuxt-themes/docus
2. @q-soft/qs-docu-tool (with i18n, vuetify, mermaid and stuff)
3. app layer > the app you are working on

image

Clarification about layers shown in the image
-  `@q-soft/qs-docu-tool` extends from `@nuxt-themes/docus`, 
- the app extends from `@q-soft/qs-docu-tool`

## the challenge
I want to access i18n in my plugin carousel of the app layer. I cant, since the plugins of @q-soft/qs-docu-tool are loaded afterward.

I tried both,
setup plugins just by name and with number prefix. The plugins of the base/extended layer got loaded AFTER the app layer.
(please help me with the naming convention of the layers)

## Expected

Load plugins of docus first, then @q-soft/qs-docu-tool and at last app layer plugins. In that case I could use i18n in my carousel plugin, since its loaded before.

### Additional information

### Final checks

userquin:
### 🔗 Linked issue

closes  #22414

### ❓ Type of change

### 📚 Description

Plugins are only sorted by `order`, should also be sorted by name.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -126,31 +126,21 @@ async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   }
 
   // Re-initialise plugins
-  app.plugins = nuxt.options.plugins.map(normalizePlugin)
-
-  // Track scanned plugins separately so we can apply a sorting mechanism to them
-  const scannedPlugins: [string, NuxtPlugin][] = []
-  for (const config of nuxt.options._layers.map(layer => layer.config)) {
-    for (const plugin of config.plugins || []) {
-      app.plugins.push(normalizePlugin(plugin as NuxtPlugin | string))
-    }
-
-    if (config.srcDir) {
-      const pluginDir = join(config.srcDir, config.dir?.plugins || 'plugins')
-      for (const file of await resolveFiles(pluginDir, [
-        '*.{ts,js,mjs,cjs,mts,cts}',
-        '*/index.*{ts,js,mjs,cjs,mts,cts}' // TODO: remove, only scan top-level plugins #18418
-      ])) {
-        scannedPlugins.push([file.replace(pluginDir, ''), normalizePlugin(file)])
-      }
-    }
+  app.plugins = []
+  let idx = 0
+  // Sort Plugins: layers plugins first then project plugins
+  for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+    app.plugins[idx++ === (nuxt.options._layers.length - 1) ? 'push' : 'unshift'](...[
+      ...(config.plugins || []),
+      ...config.srcDir
+        ? await resolveFiles(config.srcDir, [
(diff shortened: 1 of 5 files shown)


danielroe (commented):
This will result in reordering plugins that don't have an explicit order.

That's fine for plugins within a user's project, but for plugins added by modules, they might have been added with an explicit and deliberate order.

Maybe we can explicitly sort _only_ scanned plugins within layers?

packages/nuxt/src/core/app.ts
```diff
@@ -125,20 +125,31 @@ async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
...
+        '*/index.*{ts,js,mjs,cjs,mts,cts}' // TODO: remove, only scan top-level plugins #18418
+      ])) {
+        scannedPlugins.push([file.replace(pluginDir, ''), normalizePlugin(file)])
+      }
+    }
+  }
+
+  // Sort scanned plugins by their names so a scanned `01.plugin.ts` will always run before finished `02.plugin.ts`
+  // and append to the plugin array
+  for (const [_name, plugin] of scannedPlugins.sort(([a], [b]) => a.localeCompare(b))) {
```
userquin, line 151: uhmm, we should also apply order, you can have plugins there with object notation

userquin: ok, order applied for client/server in `annotatePlugins`

danielroe (approved):

packages/nuxt/src/core/app.ts
pi0, line 151: As per https://github.com/nuxt/nuxt/pull/22889#pullrequestreview-1600781599, shouldn't this be _within_ the for loop to sort each layer by itself?

Sorting all plugins might sound reasonable when authoring all layers but when extending with 3rd party layers, this order can be easily messed up with numbering. I think we shall at least have an opt-in (or opt-out if you see benefits of this as default) for layers meta config. (also the fact that, plugins shall not depend on ordering as much as possible and prefer hooks...)

danielroe: thanks - I appreciate the potential breaking nature of this pr so really appreciate feedback on it.

I think the main scenario is that if a user creates a plugin, they would expect to be able to access any injections from layers, and currently all layer plugins are loaded _after_ all user plugins. a lighter weight solution might be to simply reverse that order and load layer plugins before user plugins? I think this might also be good enough.

escape hatch for layer authors is mostly that plugins can be specified in `nuxt.config` and they will then be loaded before any scanned plugins.

pi0: Reversing order of layers seems a good idea! Because indeed by nature layers are supposed to be overridable and accessible by higher order

Issue #8937, opened 2021-03-06
'Generating pages with full static mode ' means that page js will not working?


chenPengXu:
### Versions

- nuxt:  v2.15.2
- node: v12.10.0

<details open>
<summary>Additional Details</summary>
<br>

</details>

### Steps to reproduce
1. creat app with cli
2. modify pages/index.vue   with
`
export default {
    name: 'TDemo',
    data() {
      return {
        a: 'hello a',
      }
    },
    mounted() {
      console.log('------mounted hello-----')
    },
    methods: {
      onTest() {
	      this.a = 'change a'
      }
    }
  }
`

3. run build and generate  (with target: 'static'  in config)
4. push the dist to my server

### What is Expected?
1. console should show the log '------mounted hello-----'
2. click button 'change', text will change

### What is actually happening?
1. log not show
2. click not working

So, in full static mode, pages generated will not working with life circle and methods ?

danielroe:
@chenPengXu Would you provide a reproduction please?

chenPengXu:
> @chenPengXu Would you provide a reproduction please?
index.vue as follow;
and nuxt.config.js : target is 'static' and mode is 'universal'

![1615027640564](https://user-images.githubusercontent.com/12268452/110204141-7a85dd00-7eac-11eb-822c-0d1ef1f26485.jpg)

danielroe:
@chenPengXu When I say a reproduction, I mean a codesandbox via https://template.nuxtjs.org or a repository with all (but only) the code needed to reproduce the issue.

chenPengXu:
> @chenPengXu When I say a reproduction, I mean a codesandbox via https://template.nuxtjs.org or a repository with all (but only) the code needed to reproduce the issue.

codesandbox code :  https://codesandbox.io/s/sleepy-spence-uqwq6?file=/pages/index.vue
and my code online : https://www.visgoods.com/dist/index.html

chenPengXu:
> The location you're hosting your files doesn't match the expectation of your Nuxt app. If you want to host them in `dist` then you can do so by setting [a custom `router.base`](https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-router/#base). And you shouldn't access them via `index.html` but just the directory - e.g. `/dist/`.
> 
> See [a working CodeSandbox](https://codesandbox.io/s/gallant-andras-dbojn?file=/package.json) with generate + start.
> 
[...]

PR #22904, opened 2023-08-30
feat(nuxt): add  option value 'routerPrefixForChangeableBase'

chenPengXu:
add  option value 'routerPrefixForChangeableBase' in router's options

### 🔗 Linked issue
https://github.com/nuxt/nuxt/issues/8937 ,  some like this case would be fixed(new nuxt user can not understand base usage, and deploy static app would be not work. )

### ❓ Type of change

### 📚 Description

add  option value 'routerPrefixForChangeableBase' in router's options for this case: The app deploy in the address: https://xxx/base1/routername/pagename.html， in this case, the base value may be '/base1/'， and i hope the address can also be visit at https://xxx/someSpePath/base1/routername/pagename.html，then routerPrefixForChangeableBase can be set '/routername'。the base value is too restrictive in some case. So, i hope fixed it.

### 📝 Checklist


chenPengXu:
add option value 'routerPrefixForChangeableBase' in router's options for dynamic base

danielroe:

PR #22905, opened 2023-08-30
fix(nuxt): ensure middleware is processed when returning `true`

### LINKED ISSUE #22898: The behavior of middleware is inconsistent with vue-router's `beforeEach`
### Environment

**Stackblitz**
- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.0
- CLI Version:      3.7.2
- Nitro Version:    2.6.2
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-owjclb

### Describe the bug

When switching routes, both global middleware should be executed, but currently only one is being executed and then it stops.

I believe this might be happening because it immediately exits the middleware loop when a middleware returns `true`. It should continue executing the subsequent middleware. Can I create a pull request to address this issue?

### Additional context

_No response_

### Logs

_No response_

Mini-ghost:
### 🔗 Linked issue

Resolve #22898

### ❓ Type of change

### 📚 Description

When switching routes, all middleware should run to completion unless there is a middleware redirection or an error occurs. However, currently, if we return `true` in one of the middleware, the loop will terminate without continuing.

As shown below, when `return true` is encountered, the loop exits at line 183, even though there are other unexecuted middlewares afterward:
https://github.com/nuxt/nuxt/blob/aa0ea8bf2d71b59344a773bcebd776f89f4cdafd/packages/nuxt/src/pages/runtime/plugins/router.ts#L182-L184

Therefore, I added a check in the previous code. If `result` is `true`, it allows the loop to continue, executing the next middleware.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -58,7 +58,7 @@ function getRouteFromPath (fullPath: string | Partial<Route>) {
   }
 }
 
-type RouteGuardReturn = void | Error | string | false
+type RouteGuardReturn = void | Error | string | boolean
 
 interface RouteGuard {
   (to: Route, from: Route): RouteGuardReturn | Promise<RouteGuardReturn>
@@ -130,7 +130,7 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>({
           // Cancel navigation
           if (result === false || result instanceof Error) { return }
           // Redirect
-          if (result) { return handleNavigation(result, true) }
+          if (typeof result === 'string' && result.length) { return handleNavigation(result, true) }
         }
 
         for (const handler of hooks['resolve:before']) {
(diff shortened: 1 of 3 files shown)


Mini-ghost:
@danielroe 

In the source code, there are two files that handle middleware: `packages/nuxt/src/pages/runtime/plugins/router.ts` and `packages/nuxt/src/app/plugins/router.ts`. I'm not sure what the difference is between these two, should I make modifications to both?

In `packages/nuxt/src/app/plugins/router.ts`, it seems that the middleware type is different from Vue Router. I'm not sure if this is intentional?

danielroe:
@Mini-ghost One of them applies when the `vue-router` integration is not enabled. (Basically when there is no `~/pages` directory.) The other one applies when it is.

Mini-ghost:
@danielroe

Thank you for your explanation. I have updated the relevant code. I appreciate your help!

danielroe (approved):

Issue #22900, opened 2023-08-30
useAsyncData break pages if multiple navigation during page change


huang-julien:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v18.17.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  yarn@1.22.19
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-suvj1g?file=pages%2Fasyncdata.vue,pages%2Findex.vue,app.vue

- click 10 time within a second on the link
- don't get navigated and the page breaks
- open your console
- see an error

### Describe the bug

when clicking a lot of time on a link with a navigation to a page which requires a long asyncData, nuxt breaks with an vue error when navigating and mounting the new page

### Additional context

I think this issue can happens a lot on slow pages with users double/triple clicking on a nav link but not on pages that has a fast useAsyncData. It can also depend on the user network speed

danielroe:
Let's track in https://github.com/nuxt/nuxt/issues/13309.

bombasticseahorse:
@danielroe This bug was introduced in **v3.6.2**. **v3.6.1** with the same Vue version (**v3.3.4**) doesn't have this bug. It looks like it's related to Nuxt.

danielroe:
Yes, it did start then, but that's likely because we fixed an upstream bug, and began awaiting suspense resolution properly.

The root cause is upstream.

agracia-foticos:
> Yes, it did start then, but that's likely because we fixed an upstream bug, and began awaiting suspense resolution properly.
> 
> The root cause is upstream, as far as I can tell.
> 
> We're currently investigating a possible solution in Vue itself.

Perfect!!!

MarioC3:
Has someone found a workaround for this issue?

agracia-foticos:
Same issue

Josepdal:
Same here

firsttarhan:
Same issue

hendrikheil:
Can you try disabling transitions?

You can see how to do that here:
https://nuxt.com/docs/getting-started/transitions#disable-transitions

This is possibly fixed by https://github.com/vuejs/core/pull/9309, but disabling the transitions should lead to the same result.

Josepdal:
> Can you try disabling transitions?
> 
[...]

PR #22947, opened 2023-09-01
test: add separate suspense test suite

antfu:
Add failing test for https://github.com/nuxt/nuxt/issues/13309

Based on repro by @huang-julien in https://github.com/nuxt/nuxt/issues/22900


patches/@vue__runtime-core@3.3.4.patch
```diff
@@ -0,0 +1,47 @@
...
+       if (suspense.deps <= 0) {
+         suspense.resolve();
+       } else if (isInFallback) {
+@@ -6119,6 +6123,9 @@ function baseCreateRenderer(options, createHydrationFns) {
+   const move = (vnode, container, anchor, moveType, parentSuspense = null) => {
+     const { el, type, transition, children, shapeFlag } = vnode;
+     if (shapeFlag & 6) {
++      // This I am not yet sure why, but it can be undefined in some conditions
++      if (!vnode.component)
++        return
```
antfu, line 44: I am able to fix the bug by patching Vue with these two changes. I am thinking of sending PRs to Vue core but can't find a good way to reproduce it with pure Vue, nor do I find a rational reason for them.

Also not sure how feasible it is for Nuxt to ship a patched Vue or so

hendrikheil: We were able to solve this without any patches by disabling transitions in nuxt.

If this applies to this exact issue as well, https://github.com/vuejs/core/pull/9309 should fix this upstream

danielroe (approved):

PR #22972, opened 2023-09-03
fix(nuxt): disallow `write: false` for type templates

AndreyYolkin:
### ❓ Type of change

### 📚 Description

As stated in https://github.com/nuxt/nuxt/blob/4e05650cde31ca73be4d14b1f0d23c7854008749/packages/kit/src/template.ts#L89-L92 `.d.ts` files always be written to the filesystem. This pull request introduces the NuxtTypeTemplate interface, where the write parameter for the addTypeTemplate function is omitted.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/nuxt.ts
+++ b/packages/schema/src/types/nuxt.ts
@@ -18,7 +18,11 @@ export interface NuxtPlugin {
   order?: number
 }
 
-export interface NuxtTemplate<Options = Record<string, any>> {
+// Internal type for simpler NuxtTemplate interface extension
+
+type TemplateDefaultOptions = Record<string, any>
+
+export interface NuxtTemplate<Options = TemplateDefaultOptions> {
   /** resolved output file path (generated) */
   dst?: string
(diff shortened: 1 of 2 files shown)


packages/schema/src/types/nuxt.ts
```diff
@@ -38,6 +38,8 @@ export interface ResolvedNuxtTemplate<Options = Record<string, any>> extends Nux
   dst: string
 }
 
+export type NuxtTypeTemplate<Options = Record<string, any>> = Omit<NuxtTemplate<Options>, 'write'>
```
danielroe, line 41: What do you think about instead having `write?: true` to preserve general compatibility?

AndreyYolkin: I was thinking about it too. In purpose of global compatibility we can, of course, keep the param, but set it to `true`

AndreyYolkin:
Something unexpected happened with tests 😟

danielroe (approved):

AndreyYolkin:
One day I'll write a PR title that won't have to change 🤞

Issue #22763, opened 2023-08-23
`useHead` does not inject styles in concurrent pre-rendering


Atinux:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v18.12.1
- Nuxt Version:     3.6.5
- Nitro Version:    2.6.0
- Package Manager:  pnpm@8.6.12
- Builder:          vite
- User Config:      devtools, nitro
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-4ah2ty?file=app.vue,nuxt.config.ts

### Describe the bug

We found an issue with `useHead` used in a plugin and using `nuxt generate`.

Download the Stackblitz because they don't have the `grep` command.

First, run `npm run generate`, it should generate 100 pages.

Running:
```bash
grep -RHl '__nuxt' dist/page/ | wc
```

Shows:
```
     100     100    2390
```

Then run:
```bash
grep -RHl 'header-height' dist/page/ | wc
```

Returns:

```bash
      69      69    1725
```

Happy to take a look at it @harlan-zw, seems related directly to `unhead`.

### Additional context

If I set the `concurrency` to `1`, it works perfectly:

```ts
export default defineNuxtConfig({
  nitro: {
    prerender: {
      concurrency: 1,
    }
  }
})
```

But as you know this is not the best case.

harlan-zw:
This is the valid work around working in Nuxt 3.7

```ts
export default defineNuxtPlugin((nuxtApp) => {
  const head = nuxtApp.ssrContext!.head
  head.push({
    style: [
      {
        innerHTML: ':root { --header-height: 5em; }',
        tagPriority: -2,
        id: 'nuxt-usehead-concurrency-bug',
      },
    ],
  })
});
```

harlan-zw:
Seems to be an issue with the useHead context being unsynced with the nuxt app head instance

harlan-zw:
Maybe related https://github.com/nuxt-modules/turnstile/pull/271?notification_referrer_id=NT_kwDOAFFGHbI3NTEzMzg3ODIwOjUzMjYzNjU#event-10201716443

harlan-zw:
This is now fixed in Unhead [1.4.0 ](https://github.com/unjs/unhead/releases/tag/v1.4.0) without any workarounds.

Ni55aN:
`useSeoMeta` doesn't work for me since 3.7.0, but worked correctly in 3.3.1

[...]

Issue #22712, opened 2023-08-20
`useSeoMeta` doesn't work properly in an async composable


andreww2012:
### Environment

```
- Operating System: Windows_NT
- Node Version:     v16.16.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.6.10
- Builder:          vite
- User Config:      devtools, imports
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://github.com/andreww2012/nuxt3-useSeoMeta-doesnt-work-properly-in-async-composable

### Describe the bug

- Let's say the script section of pages `a` and `b` is async and contains `useSeoMeta` call after awaiting the same async operation.
- We want to avoid code duplication and move this code to a shared function (composable). Essentially, we replace this code with `await thisNewComposable()` expression.
- After that, if we perform parallel requests to those pages, meta tags defined in `useSeoMeta` start randomly missing from the rendered HTML. This doesn't happen if we don't move that code in the composable.

danielroe:
You will need to do something like this:

```ts
export const useMeta = async (name: string) => {
  const nuxtApp = useNuxtApp();
  await new Promise((resolve) => setTimeout(resolve, 1000));

  nuxtApp.runWithContext(() => useSeoMeta(() => ({
    title: `${name} layout title`,
    description: `${name} layout description`,
  })));
};
```

See more here: https://nuxt.com/docs/guide/concepts/auto-imports#using-vue-and-nuxt-composables

andreww2012:
Thank you for pointing that out! I did manage to work this around using a different approach in the end.

Several factors lead me to not reading this piece of documentation:
- No error messages. It worked absolutely fine, and I only started to suspect that something's not right when I started receiving emails from search engines complaining that meta tags are randomly disappearing on the production site. It took me around 6 hours in total to debug this :(
- I don't use auto imports. Consequently, I didn't even read the "Auto-imports" page in the docs.
- If even I did read that, it would probably take me a while to think of a solution that you proposed. `useNuxtApp` is not mentioned on that page, and `runWithContext` cannot be found in the whole documentation:

![2023-08-30](https://github.com/nuxt/nuxt/assets/6554045/de7bbfb2-4ff8-4fa3-9af1-9e680802214c)

I would like to propose:
- Documentation improvements in this regard (move this to a different section, more bad code examples + workarounds)?
[...]

Issue #19244, opened 2023-02-23
Calling `useHead` after an asynData cause freezing head tags


farnabaz:
### Environment

```
------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.2.2`
- Nitro Version:    `2.2.3`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
---
```

### Reproduction

https://stackblitz.com/edit/nuxt-starter-ezanbf?file=components%2FTest.vue,pages%2Findex.vue,pages%2Finvalid.vue,app.vue

Originally reported in https://github.com/nuxt/content/issues/1919

### Describe the bug

- Open reproduction, click on `invalid` link
- Check `<title>` tag in head
- Click in index link
- Check title, still shows `invalid` page's title

harlan-zw:
Thanks, looks like a regression in the latest minor patch of Unhead. Have a release coming soon

harlan-zw:
On further investigation, it seems like this isn't fixed with the latest version.

This reproduction isn't exactly right, it's using the nuxt apps `useHead` which isn't going to have any sort of lifecycle events. We want the title to be removed so we need to use the `useHead` composable directly. Here's a more accurate one: https://stackblitz.com/edit/nuxt-starter-yqtvc1?file=app.vue

The specific issue is that the Vue `getCurrentInstance` isn't available after the async is resolved. It needs this scope to attach the `onBeforeUnmount` listener which disposes of the entry.

Are you aware of any ways around this @danielroe?

Getting around is easy enough, similar to the Nuxt app usage, just define the `useHead` upfront with the reactive data. But it may be confusing for end users.

```ts
const title = ref('')

useHead({
  title,
})

const { data } = await getTitle()
title.value = data
```

danielroe:
This is a limitation of Vue 3 script setup - the instance is only preserved in `<script setup>` blocks. In this case, you must use all composables that require access to the Vue or Nuxt contexts synchronously (or before the first await).

harlan-zw:
Thanks for the explanation! I think this can be closed once there's documentation available on the useHead or seo / meta page then

harlan-zw:
> @harlan-zw did you find a workaround for this?

See

>  If you are wanting to use a <script> block, you must use all composables that require access to the Vue or Nuxt contexts synchronously (or before the first await).

[...]

PR #22973, opened 2023-09-03
fix(nuxt): resolve head instance from Nuxt app

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22763
https://github.com/nuxt/nuxt/issues/22712
https://github.com/nuxt/nuxt/issues/19244

### ❓ Type of change

### 📚 Description

We want to be able to use `useHead` when within a Nuxt App context but not necessarily within a Vue app context.

To do this, we provide a static injection handler to resolve the head instance.

It would be good to sanity-check my approach here, we want to avoid a regression with the parallel request issue.

Also worth noting, using the `setHeadInjectionHandler` function will stop the default logic from running, that is using the shared instance of the head `getActiveHead`, this should solve some other weird quirks with parallel requests (at least within a nuxt app context).

### 📝 Checklist


docs/1.getting-started/5.seo-meta.md
```diff
@@ -128,6 +128,29 @@ interface MetaObject {
...
+- Use `runWithContext` to preserve to context:
+
+```ts
+const nuxtApp = useNuxtApp();
+const data = await $fetch(`/api/data/${key}`) // context is lost from the await
+nuxtApp.runWithContext(() => {
+  useHead({
+    title: `${data.title}`,
+  })
+});
```
danielroe, line 149: do you think we can create a wrapper around `useHead` and other head composables that does this automatically, instead of asking users to do this?

harlan-zw: How can it solve the lost context from the await though :thinking:

harlan-zw: To give some more context, if the vueApp is available then useHead will work. So the only time this error comes up is from edge cases like https://github.com/nuxt/nuxt/issues/19244

From my understanding, `asyncContext` is aiming to fix users having to do anything to solve it.

danielroe: asyncContext is great but currently only for node + bun.

In `<script setup>` vue transforms it and we preserve vueApp context. Elsewhere, there are places where we transform things to preserve _nuxt_ context but we can't preserve the vue context as we're not in control there. So we might have nuxtApp (from which we can access vue app) even where inject might not work.

harlan-zw: I might have solved this, can you review :pray:

packages/nuxt/src/head/runtime/plugins/unhead.ts
```diff
@@ -1,18 +1,25 @@
...
   setup (nuxtApp) {
     const head = import.meta.server
       ? nuxtApp.ssrContext!.head
       : createClientHead({
         plugins: unheadPlugins
       })
+    // allow useHead to be used outside a Vue context but within a Nuxt context
+    setHeadInjectionHandler(
+      // need a fresh instance of the nuxt app to avoid parallel requests interfering with each other
+      () => useNuxtApp().vueApp._context.provides.usehead
```
danielroe, line 21: could we just do:

```suggestion
      () => nuxtApp.vueApp._context.provides.usehead
```

docs/1.getting-started/5.seo-meta.md
danielroe: this looks _very_ promising. just one question for you but otherwise nice work

packages/nuxt/src/head/runtime/plugins/unhead.ts
harlan-zw: I think it may cause issues if the nuxtApp context is being stored globally.

Here's the code: https://github.com/unjs/unhead/blob/main/packages/vue/src/composables/injectHead.ts

Maybe you have a better understanding

danielroe: ohh, I see. 👍

danielroe (approved):

PR #22974, opened 2023-09-03
perf(nuxt): prevent head dom from rendering twice

harlan-zw:
### ❓ Type of change

### 📚 Description

With the move to immediately rendering the dom instead of debouncing it, we've introduced a regression in that the head will render to the DOM twice.

While the performance impact of this is reduced by quite a bit with Unhead 1.3, it's still not ideal and maybe adding ~5ms of extra load time.

The underlying issue is that we should wait for the suspense to resolve if we know the app is hydrating.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/head/runtime/plugins/unhead.ts
+++ b/packages/nuxt/src/head/runtime/plugins/unhead.ts
@@ -28,8 +28,7 @@ export default defineNuxtPlugin({
       // wait for new page before unpausing dom updates (triggered after suspense resolved)
       nuxtApp.hooks.hook('page:finish', () => {
         // app:suspense:resolve hook will unpause the DOM
-        if (!nuxtApp.isHydrating)
-          syncHead()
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/head/runtime/plugins/unhead.ts
```diff
@@ -26,7 +26,10 @@ export default defineNuxtPlugin({
       head.hooks.hook('dom:beforeRender', (context) => { context.shouldRender = !pauseDOMUpdates })
       nuxtApp.hooks.hook('page:start', () => { pauseDOMUpdates = true })
       // wait for new page before unpausing dom updates (triggered after suspense resolved)
-      nuxtApp.hooks.hook('page:finish', syncHead)
+      nuxtApp.hooks.hook('page:finish', () => {
```
danielroe, line 29: should we use `app:suspense:resolve` instead?

danielroe (approved):

Issue #20988, opened 2023-05-22
upgrade to 3.5.1 ,get an error on Browser console TypeError: webCrypto.randomUUID is not a function


GITleonine1989:
### Environment

- Operating System: Windows_NT
- Node Version:     v16.16.0
- Nuxt Version:     3.5.1
- Nitro Version:    2.4.1
- Package Manager:  npm@8.11.0
- Builder:          vite
- User Config:      experimental
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-qleyyo 
here is this code ,but can`t  reproduce this issue.the same code on my local machine get this error
![image](https://github.com/nuxt/nuxt/assets/7582490/98185022-e6e7-475d-9ba0-4203df4f27e2)

### Describe the bug

when  i use a server component,  and set componentIslands true, the browser console will get a error: `TypeError: webCrypto.randomUUID is not a function`.

```
components/Header.server.vue
```
```
 experimental: {
    componentIslands: true,
  },
```
app.vue: 
```
<template>
  <div>
    <Header />
  </div>
</template>
```

But i cant reproduce  this issue on the stackblitz, and this issue occur when i upgrade  from v.3.4.3 to v.3.5

maou-shonen:
WebCrypto is only enabled in HTTPS and localhost environments.
could this be the cause of your issue?

GITleonine1989:
@maou-shonen i tried,you are right ,  this issue  occur  when i through the ip+port visit my site.

PR #22975, opened 2023-09-03
fix(nuxt): always use increment for id with client side islands

bounoable:
### 🔗 Linked issue

Issue: #20988
PR: #20992

### ❓ Type of change

### 📚 Description

I still get the webCrypto.randomUUID() error when navigating to a page with a server component in dev mode. I think #20992 missed to replace a `randomUUID()` call because this fix seems to fix work.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -86,7 +86,7 @@ export default defineComponent({
       ssrHTML.value = renderedHTML
     }
     const slotProps = computed(() => getSlotProps(ssrHTML.value))
-    const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? randomUUID())
+    const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? getId())
     const availableSlots = computed(() => [...ssrHTML.value.matchAll(SLOTNAME_RE)].map(m => m[1]))
 
     const html = computed(() => {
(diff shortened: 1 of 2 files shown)


danielroe (approved):

danielroe:
Would you provide a reproduction of the bug this change solves? 🙏

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -86,7 +86,7 @@ export default defineComponent({
       ssrHTML.value = renderedHTML
     }
     const slotProps = computed(() => getSlotProps(ssrHTML.value))
-    const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? randomUUID())
+    const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? getId())
```
huang-julien, line 89: Not quite sure, but I did have a hydration mismatch. i'll investigate

bounoable:
I tried but this actually cannot reproduced on Stackblitz & co. because webCrypto is only non-available in non-https environments 🥲

This error is triggered within my application when "navigating to another page that needs to render a server component on the client". I'm using the `.server.vue` suffix, not the `<NuxtIsland>` component.

packages/nuxt/src/app/components/nuxt-island.ts
huang-julien: I'm resolving this as i can't reproduce it anymore...

huang-julien (approved):
I think this should be OK. There shouldn't be any risk of having a conflict between ids since we still use `randomUUID` server side.
The issue here is that `randomUUID` isn't available in non-secure context. 
I can't reproduce it locally even with ngrok, and it seems like `localhost` is a secure context by default ?

What do you think @danielroe ?

BTW you can update snapshots with `pnpm test:fixtures --u` 🙂

PR #23040, opened 2023-09-06
fix(nuxt): find parent routes by exact path match

### LINKED ISSUE #22977: Optional parameter route shows page of required parameter route.
### Environment

- Node Version:     v16.15.0                                                                                                                                                                                                                                                                                        
- Nuxt Version:     3.6.5                                                                                                                                                                                                                                                                                           
- Nitro Version:    2.5.2                                                                                                                                                                                                                                                                            22:56:42  
- Builder:          vite
- User Config:      nitro, css, build, modules, runtimeConfig, devtools
- Runtime Modules:  @pinia/nuxt@0.4.11
- Build Modules:    -

### Reproduction

Create a page named `[[foo]].vue` and a page named `[foo].vue` and add a NuxtPage within the template of `[[foo]].vue`.

### Describe the bug

What happens is that the page of `[foo].vue` gets rendered at the index route of `[[foo]].vue`. (So basically, you do not pass the parameter foo which is optional for `[[foo]].vue`) This is weird because the parameter `foo` of `[foo].vue` should be required but gets rendered anyway.

### Additional context

It is not my intention to write a PR for this issue. Maybe this is suppose to work this way, I am happy to know if it is or not.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves nuxt/nuxt#22977

### ❓ Type of change

### 📚 Description

It's possible to have paths that produce similarly _named_ paths but are not really the same, for example `/[[foo]].vue` and `/[foo].vue`. These should be siblings and not create a parent/child relationship in this case.

As this is a sensitive area, I would particularly value spotting any edge cases I may have overlooked.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -80,7 +80,7 @@ export async function generateRoutesFromFiles (files: string[], pagesDir: string
       route.name += (route.name && '/') + segmentName
 
       // ex: parent.vue + parent/child.vue
-      const path = getRoutePath(tokens)
+      const path = getRoutePath(tokens).replace(/\/index$/, '/')
       const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path)
 
       if (child && child.children) {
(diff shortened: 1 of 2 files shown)


pi0 (approved):
LGTM

atinux (approved):
Looks good to me

PR #23048, opened 2023-09-06
fix(nuxt): load `spaLoadingTemplate` if file exists

### LINKED ISSUE #23047: fix(nuxt): allow load default spa loading indicator
### Environment

- Operating System: Darwin
- Node Version:     v20.1.0
- Nuxt Version:     3.7.0
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  npm@9.6.4
- Builder:          -
- User Config:      ssr, build, modules, app, css, vite, runtimeConfig
- Runtime Modules:  @vueuse/nuxt@10.4.1
- Build Modules:    -

### Reproduction

add `~/app/spa-loading-template.html`

https://stackblitz.com/edit/github-6nhkcv

### Describe the bug

Loading `spaLoadingTemplate` by default from `~/app/spa-loading-template.html` no longer works as documented.

Implemented in https://github.com/nuxt/nuxt/pull/21640
- Default value: null, then `~/app/spa-loading-template.html` was loaded.

Changed default value to `false` in https://github.com/nuxt/nuxt/pull/22798
- `false` if undefined or null, then `~/app/spa-loading-template.html` is never loaded.

### Additional context

Since last version, `spaLoadingTemplate: true` only loads nuxt animation and does not load `~/app/spa-loading-template.html` if exist.

### Logs

_No response_

davidglezz:
### 🔗 Linked issue

resolves #23047

### ❓ Type of change

### 📚 Description

The behavior is not the same as documented because was changed in https://github.com/nuxt/nuxt/pull/22798

This PR restores the loading of ~/app/spa-loading-template.html when `spaLoadingTemplate` is `true`

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -421,5 +421,5 @@ function spaLoadingTemplate (options: Nuxt['options']) {
   }
 
   console.warn(`[nuxt] Could not load custom \`spaLoadingTemplate\` path as it does not exist: \`${options.spaLoadingTemplate}\`.`)
-  return '';
+  return ''
 }
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -30,11 +30,6 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
     ? [new RegExp(`node_modules\\/(?!${excludePaths.join('|')})`)]
     : [/node_modules/]
 
-  const spaLoadingTemplate = nuxt.options.spaLoadingTemplate ?? resolve(nuxt.options.srcDir, 'app/spa-loading-template.html')
-  if (spaLoadingTemplate && nuxt.options.spaLoadingTemplate && typeof spaLoadingTemplate !== 'boolean' && !existsSync(spaLoadingTemplate)) {
-    console.warn(`[nuxt] Could not load custom \`spaLoadingTemplate\` path as it does not exist: \`${spaLoadingTemplate}\`.`)
-  }
-
   const nitroConfig: NitroConfig = defu(_nitroConfig, {
     debug: nuxt.options.debug,
     rootDir: nuxt.options.rootDir,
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/app.ts
```diff
@@ -189,12 +189,13 @@ export default defineUntypedSchema({
...
   /**
-   * A path to an HTML file, the contents of which will be inserted into any HTML page
+   * Boolean or a path to an HTML file with the contents of which will be inserted into any HTML page
    * rendered with `ssr: false`.
    *
-   * By default Nuxt will look in `~/app/spa-loading-template.html` for this file.
+   * Default value: `false` (Disabled).
    *
-   * You can set this to `false` to disable any loading indicator.
+   * If true, Nuxt will look for `~/app/spa-loading-template.html` file, if it does not exist,
```
danielroe, line 197: I think the presence of `~/app/spa-loading-template.html` should enable the template. I think the default Nuxt image should only be used if this is set to `true` and if that file is not present:

- `true` +  `~/app/spa-loading-template.html` - template provided
- `true` - default nuxt icon
- `false` - no template
- unset +  `~/app/spa-loading-template.html` - template provided
- unset - no template

What do you think?

davidglezz: Thanks for your review.
I implemented in that way.

Since its my first contribution, I expect change request.

packages/nuxt/src/core/nitro.ts
```diff
@@ -415,3 +402,24 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
 function relativeWithDot (from: string, to: string) {
   return relative(from, to).replace(/^([^.])/, './$1') || '.'
 }
+
+function spaLoadingTemplate (options: Nuxt['options']) {
```
danielroe, line 406: Totally minor point, but let's take `nuxt` as the argument, for consistency with other code.

packages/schema/src/config/app.ts
danielroe: nice work

danielroe (approved):

danielroe:
Thank you for your great work on this PR ❤️

PR #23057, opened 2023-09-07
perf(nuxt): use module as key to deduplicate inline styles

danielroe:
### ❓ Type of change

### 📚 Description

This is a drive-by change that I hope might improve performance. Rather than needing to compare entire (potentially long) string added to a set, I believe unhead is already comparing keys when provided.

This also avoids an additional array creation.

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -497,15 +497,15 @@ function renderHTMLDocument (html: NuxtRenderHTMLContext) {
...
   for (const mod of usedModules) {
     if (mod in styleMap) {
       for (const style of await styleMap[mod]()) {
-        inlinedStyles.add(style)
+        inlinedStyles.push({ key: mod, innerHTML: style })
       }
     }
   }
-  return Array.from(inlinedStyles).map(style => ({ innerHTML: style }))
+  return inlinedStyles
```
harlan-zw, line 508: if performance is the main concern it's probably faster to dedupe in this function?

```ts
async function renderInlineStyles (usedModules: Set<string> | string[]): Promise<Style[]> {
  const styleMap = await getSSRStyles()
  const inlinedStyles: Record<string, string> = {}
  for (const mod of usedModules) {
    if (mod in styleMap) {
      for (const style of await styleMap[mod]()) {
        inlinedStyles[mod] = style
      }
    }
  }
  return Object.values(inlinedStyles).map(style => ({ innerHTML: style }))
}
```

danielroe: No, those styles are never hydrated 👍

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -497,15 +497,15 @@ function renderHTMLDocument (html: NuxtRenderHTMLContext) {
 
 async function renderInlineStyles (usedModules: Set<string> | string[]): Promise<Style[]> {
   const styleMap = await getSSRStyles()
-  const inlinedStyles = new Set<string>()
```
harlan-zw, line 500: So with this setup duplicates would only happen if the user is writing/importing the same CSS across modules?

danielroe: It depends on how rollup chunks them, but the same CSS could be added if multiple components have the same chunk of CSS.

Looking at this with fresh eyes, I think probably we can't dedupe with module key _anyway_ as it's the content of the CSS that matters.

PR #23086, opened 2023-09-08
fix(nuxt): include method when creating `useFetch` auto key

maximepvrt:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -85,7 +85,7 @@ export function useFetch<
     return unref(r)
   })
 
-  const _key = opts.key || hash([autoKey, unref(opts.method), unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
+  const _key = opts.key || hash([autoKey, opts.method ? unref(opts.method).toUpperCase() : '', unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
   if (!_key || typeof _key !== 'string') {
     throw new TypeError('[nuxt] [useFetch] key must be a string: ' + _key)
   }
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -85,7 +85,7 @@ export function useFetch<
     return unref(r)
   })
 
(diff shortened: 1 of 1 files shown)


danielroe (approved):

TheAlexLichter (approved):
Even though the use case for anything else than `GET` is rather rare, it seems like a good idea 👍

Hebilicious:
> Even though the use case for anything else than `GET` is rather rare, it seems like a good idea 👍

Fun fact, I recently learned that there's way more registered HTTP method than I realised https://www.iana.org/assignments/http-methods/http-methods.xhtml

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -85,7 +85,7 @@ export function useFetch<
     return unref(r)
   })
 
-  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
+  const _key = opts.key || hash([autoKey, unref(opts.method), unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
```
danielroe, line 88:
```suggestion
  const _key = opts.key || hash([autoKey, opts.method ? unref(opts.method).toUpperCase() : '', unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
```

manniL: Good catch! 👍

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -85,7 +85,7 @@ export function useFetch<
     return unref(r)
   })
 
-  const _key = opts.key || hash([autoKey, unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
+  const _key = opts.key || hash([autoKey, opts.method ? unref(opts.method as MaybeRef<string>).toUpperCase() : 'GET', unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
```
danielroe, line 88:
```suggestion
  const _key = opts.key || hash([autoKey, opts.method ? unref(opts.method as MaybeRef<string>)?.toUpperCase() || 'GET' : 'GET', unref(opts.baseURL), typeof _request.value === 'string' ? _request.value : '', unref(opts.params || opts.query)])
```

PR #23093, opened 2023-09-08
fix(nuxt): don't `joinURL` with remote sources on NuxtIsland

### LINKED ISSUE #23092: remote sources is not working due to a `joinURL`
### Environment

- Node Version:     v18.15.0
- Nuxt Version:     3.7.1
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  pnpm@8.6.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

try any remote source

### Describe the bug

````ts
      const r = await eventFetch(withQuery(import.meta.dev && import.meta.client ? url : joinURL(config.app.baseURL ?? '', url), {
        ...props.context,
        props: props.props ? JSON.stringify(props.props) : undefined
      }))
````

There's a missing `|| props.source` in the condition. `NuxtIsland` is trying to `joinURL` the baseURL or an empty string with a fullURL resulting in `/${fullURL}`

### Additional context

_No response_

### Logs

_No response_

### PR DESCRIPTION
**Pull request by huang-julien**
### 🔗 Linked issue
fix https://github.com/nuxt/nuxt/issues/23092

### ❓ Type of change

### 📚 Description
Hi :wave: 

When doing a previous fix on adding `baseURL` for nuxt islands, and merging the remote sources PR after it, a condition hasn't been updated causing remote islands to not work

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -117,7 +117,7 @@ export default defineComponent({
       }
       // TODO: Validate response
       // $fetch handles the app.baseURL in dev
-      const r = await eventFetch(withQuery(import.meta.dev && import.meta.client ? url : joinURL(config.app.baseURL ?? '', url), {
+      const r = await eventFetch(withQuery(((import.meta.dev && import.meta.client) || props.source) ? url : joinURL(config.app.baseURL ?? '', url), {
         ...props.context,
         props: props.props ? JSON.stringify(props.props) : undefined
       }))
@@ -174,7 +174,7 @@ export default defineComponent({
     } else if (import.meta.server || !nuxtApp.isHydrating || !nuxtApp.payload.serverRendered) {
       await fetchComponent()
     }
-
+    console.log(html.value, error.value)
     return () => {
       if ((!html.value || error.value) && slots.fallback) {
         return [slots.fallback({ error: error.value })]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/nuxt/src/app/components/nuxt-island.ts [hunk 1]
```diff
@@ -174,7 +174,7 @@ export default defineComponent({
     } else if (import.meta.server || !nuxtApp.isHydrating || !nuxtApp.payload.serverRendered) {
       await fetchComponent()
     }
-
+    console.log(html.value, error.value)
```
danielroe, line 177:
```suggestion
```

huang-julien: just saw it and pushed 😅 you're fast 🔥  !

huang-julien:
@danielroe do you know if there's a way to test remote island on the repo ? I don't think we can start two seperate servers on fixtures tests and this would cost a lot of time on the CI.
Maybe starting a small h3 server  which would listen to `3001` when running a specific test ?

danielroe:
I think it might be sufficient to provide a mock endpoint, which we could create in the test suite?

huang-julien:
Do you mean something like `<NuxtIsland source="/mock/island" >` ?

danielroe:
Or even fire up a mock server with listhen + h3 with a URL....

huang-julien:
A mock server may be the best solution atm, I don't think having `/mock/island` would work as a second argument of `new URL()`, it will probably throw

huang-julien:
Setting this to draft, i need to verify something on `URL`

danielroe (approved):

PR #23095, opened 2023-09-08
fix(nuxt): exclude `data-v` attrs from server component props

### LINKED ISSUE #23051: Nuxt3 .server (island) components fail to render on soft navigation if they have different props on pages
### Environment

Works in any environment, including StackBlitz:

```
- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.1
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  npm@9.4.2
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

Please refer to:

https://stackblitz.com/edit/github-ydxe3t?file=pages%2Findex.vue

1. Do `nuxt generate` and `npx serve .output/public`.
2. CLick the link 'Go to page with scoped styles'
3. **Note the 404 in console on click to the link and absence of the component**
4. Reload the page
5. Note the component was loaded normally
6. Remove scoped attribute from style tag in pageWithScopedStyles.vue and repeat the excercise

### Describe the bug

Apparently because scoped styles in Vue are implemented using attributes and the server components are rendered based on attribute values, server components **in some instances** will break (the client will request a non-existing URL of `__nuxt_island/whatever_id`, leading to a 404. This will lead to the component not being displayed. 

The sufficient conditions for the reproduction are following: 
- There is a server component used on two separate pages
- It's not used directly, but through a third component (I used a wrapper) which uses this component as root
- It has a prop that is different on these pages (I injected the route, I suppose it's also possible that a state variable may be passed as a prop

This is not reproduced in dev, only if the site is statically generated using `nuxt generate` and the client navigates to the page which uses scoped styles using soft navigation, e.g. `<NuxtLink />`

I'm not certain what causes the bug.

### Additional context

I know this might sound like an edge case but I arrived to this bug in a real-world scenario and I imagine this could be an issue for example  for component library creators who want to use .server components (llike myself). An easy enough workaround is to avoid using it as a root in the wrapper but I'm sure there are also other unforeseen scenarios that might trigger this problem.

The fact is, it works OK in dev, but breaks during SSG.

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix https://github.com/nuxt/nuxt/issues/23051

### ❓ Type of change

### 📚 Description
Hi :wave: this PR resolves #23051 . 

I'm not really a fan of having to make a  computed + a reduce, but i didn't find another solution atm. Let's admit that server components are not going to be updated every milliseconds 😅 .... If there's a better solution (like a config from vue), let me know.

This only affect `.server` components since `NuxtIsland` directly use `props` props attrbute and ignore any `data-v` attribute

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/runtime/server-component.ts
+++ b/packages/nuxt/src/components/runtime/server-component.ts
@@ -7,12 +7,10 @@ export const createServerComponent = (name: string) => {
     inheritAttrs: false,
     props: { lazy: Boolean },
     setup (props, { attrs, slots }) {
-      // #23051
-      const attrsWithoutVueDataAttr = Object.entries(attrs).reduce<Record<string, string>>((acc, [key, value]) => key.startsWith('data-v-') ? acc : Object.assign(acc, { [key]: value }), {})
       return () => h(NuxtIsland, {
         name,
         lazy: props.lazy,
-        props: attrsWithoutVueDataAttr
+        props: attrs
       }, slots)
     }
   })
--- a/packages/nuxt/src/components/runtime/server-component.ts
+++ b/packages/nuxt/src/components/runtime/server-component.ts
@@ -1,4 +1,4 @@
-import { computed, defineComponent, h } from 'vue'
+import { defineComponent, h } from 'vue'
 import NuxtIsland from '#app/components/nuxt-island'
 
 export const createServerComponent = (name: string) => {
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/components/runtime/server-component.ts
```diff
@@ -7,10 +7,12 @@ export const createServerComponent = (name: string) => {
     inheritAttrs: false,
     props: { lazy: Boolean },
     setup (props, { attrs, slots }) {
+      // #23051 - remove data-v attributes
+      const attrsWithoutVueDataAttr = computed(() => Object.entries(attrs).reduce<Record<string, string>>((acc, [key, value]) => key.startsWith('data-v-') ? acc : Object.assign(acc, { [key]: value }), {}))
```
danielroe, line 11: I think we should do this within the render function, and I think it would likely be simpler to do `Object.fromEntries` + `Object.entries().filter`.

arsengoian:
Apologies, but if it's fixed by not passing the data-v attribute, won't this break styles during SSG? I think that might be a bigger problem actually, since that means that scoped styles won't apply during SSG and if you want to style server components used inside your components with scoped styles you'll have to use `:deep()`

I mean, this can break styles in already existing applications, no?

huang-julien:
Oh the title of the PR may not be explicit enough.
The issue is that we sent some unwanted attribute as props to `NuxtIsland`, thus causing the hash to change with SSG.

arsengoian:
Thanks for the clarification!

danielroe (approved):

PR #23120, opened 2023-09-11
fix(nuxt): handle unset spa-loading fallback

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/23048

### ❓ Type of change

### 📚 Description

Unintended regression after merging https://github.com/nuxt/nuxt/pull/23048, 'undefined' result normalises through to `{}` in untyped. Also we were always warning even if it was unset.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/app.ts
+++ b/packages/schema/src/config/app.ts
@@ -243,7 +243,7 @@ export default defineUntypedSchema({
    * @type {string | boolean}
    */
   spaLoadingTemplate: {
-    $resolve: async (val, get) => typeof val === 'string' ? resolve(await get('srcDir'), val) : val
+    $resolve: async (val, get) => typeof val === 'string' ? resolve(await get('srcDir'), val) : val ?? null
   },
 
   /**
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/app.ts
```diff
@@ -243,7 +243,7 @@ export default defineUntypedSchema({
    * @type {string | boolean}
    */
   spaLoadingTemplate: {
-    $resolve: async (val, get) => typeof val === 'string' ? resolve(await get('srcDir'), val) : val
+    $resolve: async (val, get) => typeof val === 'string' ? resolve(await get('srcDir'), val) : val ?? null
```
pi0, line 246: Would `undefined` work equally same as `null` here? (if so, it might be better default. null is an evil to introduce bugs!)

danielroe: No, undefined is the issue. untyped resolves it to `{}` in that case.

pi0 (approved):

PR #23121, opened 2023-09-11
fix(kit): improve generated `tsconfig.json` defaults

danielroe:
### ❓ Type of change

### 📚 Description

Based on https://github.com/vuejs/tsconfig/blob/main/tsconfig.json, this adds a couple of improvements to the generated `tsconfig.json`, along with a TODO for some slightly more significant changes that we can defer to next minor.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/template.ts
+++ b/packages/kit/src/template.ts
@@ -117,6 +117,7 @@ export async function writeTypes (nuxt: Nuxt) {
     compilerOptions: {
       forceConsistentCasingInFileNames: true,
       jsx: 'preserve',
+      jsxImportSource: 'vue',
       target: 'ESNext',
       module: 'ESNext',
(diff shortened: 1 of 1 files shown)


packages/kit/src/template.ts
```diff
@@ -117,13 +117,19 @@ export async function writeTypes (nuxt: Nuxt) {
     compilerOptions: {
       forceConsistentCasingInFileNames: true,
       jsx: 'preserve',
+      jsxImportSource: 'vue',
```
pi0, line 120: Don't we also need `jsImportFactory`?

pi0 (approved):
Nice changes! (Haven't tried but hope it also updates nitro generated ts config to be consistent)

PR #23133, opened 2023-09-11
fix(nuxt): allow changing dirs within modules

### LINKED ISSUE #15686: changing the nuxt "dir" option through a module doesn't work
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Builder:          `vite`
- User Config:      `modules`
- Runtime Modules:  `()`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-m8hu6q?file=nuxt.config.ts

### Describe the bug

Customizing the dir options using a module doesn't work anymore.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15686

### ❓ Type of change

### 📚 Description

This allows changing the `dir` options by respecting `nuxt.options.dir` when used in the root/base layer.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -190,9 +190,10 @@ async function initNuxt (nuxt: Nuxt) {
 
   // Automatically register user modules
   for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+    const modulesDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.modules || 'modules'
     const layerModules = await resolveFiles(config.srcDir, [
-      `${config.dir?.modules || 'modules'}/*{${nuxt.options.extensions.join(',')}}`,
-      `${config.dir?.modules || 'modules'}/*/index{${nuxt.options.extensions.join(',')}}`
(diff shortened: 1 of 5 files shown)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -190,9 +190,10 @@ async function initNuxt (nuxt: Nuxt) {
 
   // Automatically register user modules
   for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+    const modulesDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.modules || 'modules'
```
pi0, line 193: Note: We shall find a better way to mark top level layer (in resolver maybe) to avoid rootDir === rootDir checks

pi0 (approved):
Nice fixes! I think as later improvement refactors we might:

- Mark top level layer explicitly when loading config
- Have kit util like `getDir` and possibly an ENUM for the defaults to avoid possible issues in different places.

packages/nuxt/src/core/nuxt.ts
danielroe: that would be great 👍

PR #23148, opened 2023-09-12
fix(nuxt): respect layer order for other layer plugins

### LINKED ISSUE #23123: [Bug] Plugin Registration Order is not deterministic when using layers
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.1
- CLI Version:      3.8.1
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      extends
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-g42wkd?file=nuxt.config.ts

### Describe the bug

The order of plugin execution is not correct when using layers. Plugins that are defined in lower layers execute after plugins defined in higher layers. This causes incorrect behavior when a plugin override is necessary. In the provided repo a use case of overriding the same entity from a higher level is not possible due to this issue.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23123

### ❓ Type of change

### 📚 Description

This builds on work done by @userquin in https://github.com/nuxt/nuxt/pull/22889 to ensure that other plugins specified directly in nuxt config by layers are loaded in sequence matching layer order.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -126,9 +126,7 @@ async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   }
 
   // Resolve plugins, first extended layers and then base
-  app.plugins = [
-    ...nuxt.options.plugins.map(normalizePlugin)
-  ]
+  app.plugins = []
   for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
     app.plugins.push(...[
       ...(config.plugins || []),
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Haven't tested but looks good 💯

packages/nuxt/src/core/app.ts
```diff
@@ -141,6 +139,14 @@ async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
     ].map(plugin => normalizePlugin(plugin as NuxtPlugin)))
   }
 
+  // Add back plugins not specified in layers or user config
+  for (const p of nuxt.options.plugins) {
+    const plugin = normalizePlugin(p)
+    if (!app.plugins.some(p => p.src === plugin.src)) {
+      app.plugins.unshift(plugin)
```
dargmuesli, line 146: Maybe this should be `push` instead of `unshift`? (https://github.com/nuxt/nuxt/issues/23163)

PR #23195, opened 2023-09-13
fix(vite): respect `devServer.host` for `hmr`

pi0:
### ❓ Type of change

### 📚 Description

When the browser points to `127.0.0.1` (the new default for nuxi in windows), vite HMR port should also listen to `127.0.0.1`. Defaulting to `""` leads to VSCode listening WSL2 on `localhost` which (surprisingly) mismatches in browser URL!

This PR fixes the issue by respecting the host that is set by nuxi. (which is currently always `127.0.01`)

image

### Alternatives

This is not the best solution since causes HMR only working on localhost. Something we were working around previously by (unsafely) listening to all interfaces.

I am thinking we might use main server HMR port for upgrade events on vite path (as a feature)

### 📝 Checklist


pi0:
Closing in flavor of https://github.com/nuxt/cli/pull/184

Issue #14723, opened 2022-08-27
Document composables concept


pi0:
Nuxt 3 uses Vue 3 with Composition API and [unjs/unctx](https://github.com/unjs/unctx) for supporting composables with the implicit context in nuxtApp (plugins, middleware) and nuxt kit too.

However, there are many cases in this needs a better explanation of how it works internally to understand why an issue like [nuxt instance unavailable](https://github.com/nuxt/framework/issues?q=nuxt+instance+unavailable) happens and how to fix it with proper usage.

I've made an outline story in https://github.com/nuxt/nuxt.js/issues/14269 which can probably be used as a basis to initiate this concept page.

floriankapaun:
> I was able to test callWithNuxt and it's working just fine. I've finally been able to beat multiple awaits and build Nuxt 2-like useAsyncData in my code, even though I had to use this function.

Was able to achieve a [workaround to my problem](https://stackoverflow.com/questions/76033123/nuxt-usefetch-useasyncdata-twice-in-method-throws-error-nuxt-instance-not-avai/76033170#76033170) with you help, but is that really the recommended approach to do things or is there a better way? 😬

daniluk4000:
@floriankapaun I've created discussion to improve DX of this behavior with more helper functions, and also to document this callWithNuxt approach: https://github.com/nuxt/nuxt/discussions/19567

Sadly discussion received no response :(

I've worked much with Vue and Nuxt internals recently and I can say that callWithNuxt now is the only way to not lose context. In my opinion, if there is no coming fix for context from Vue team this needs to be properly documented and maybe DX-improved. 

Speaking of myself, I didn't find any better way to not lose app context on SSR but that.

danielroe:
That's a nice discussion - sorry, I did not see it until now!

PR #23258, opened 2023-09-17
docs: add `nuxtApp.runWithContext`

fimion:
### 🔗 Linked issue

#14723

### ❓ Type of change

### 📚 Description

@danielroe suggested [I take a crack at this](https://twitter.com/danielcroe/status/1703433066247565323) as there was not any public documentation about `callWithNuxt`. 

I've added a header stating that this method should be used sparingly, and that ultimately this is something that would ideally be solved at the framework level.

I've also included the [detailed explanation](https://github.com/nuxt/nuxt/issues/14269#issuecomment-1397352832) with edits that make more sense for documentation.

Let me know if there is any further details I should add, or if the format needs to change at all.

### 📝 Checklist


danielroe (commented):
This is absolutely beautiful work.

The only thing I'd suggest is that we document `nuxtApp.runWithContext(() => fn())` instead, as it's a bit more intuitive how it works, and I think largely replaces `callWithNuxt`.

Very little would need to change for us to do this, just the function signature...

fimion:
Alright, I've moved it into the `useNuxtApp` docs and reworked the signature.

Are y'all using details/summary to collapse extra info? or would the detailed context documentation be better in a separate file under [Advanced](docs/3.api/4.advanced)?

danielroe (approved):

danielroe:
I think this is great, and as it's in the API section I think we can include all of this there. It's possible we could add something less/abbreviated in future in **Concepts**, maybe within https://nuxt.com/docs/guide/concepts/vuejs-development, or possibly a bit longer in a **Composables** section in  **Get Started**.

PR #23296, opened 2023-09-19
fix(nuxt): only add/remove trailing slash for http protocols

### LINKED ISSUE #23295: Trailing Slash behavior for fragment links or various protocols (e.g. mailto:)
### Reproduction

https://stackblitz.com/edit/nuxt-issue-23295?file=components%2FAppLink.vue,app.vue

Hover over links/check out hrefs. These should not have a trailing slash (for fragment, ideally *before* the #)

### Describe the bug

When using `trailingSlash: 'append'`, then the trailing slash will be appended to fragments (if existing) or also to `mailto:` or `tel:` links. This is not desired.

Not sure if this sould be resolved here *and*/*or* in ufo (see https://github.com/unjs/ufo/pull/175)

### Additional context

_No response_

### Logs

_No response_

TheAlexLichter:
### 🔗 Linked issue

resolves #23295 

Related: https://github.com/unjs/ufo/pull/175, https://github.com/unjs/ufo/issues/189

### ❓ Type of change

### 📚 Description

While https://github.com/unjs/ufo/issues/189 is still pending, Nuxt can have a solution on it's own to avoid adding/removing slashes from provided URLs unless the http/https protocol is set. 
This is important for e.g. `mailto:abc@example.com` links.

### 📝 Checklist


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -9,7 +9,7 @@ import { navigateTo, useRouter } from '../composables/router'
 import { useNuxtApp } from '../nuxt'
 import { cancelIdleCallback, requestIdleCallback } from '../compat/idle-callback'
 
-const firstNonUndefined = <T> (...args: (T | undefined)[]) => args.find(arg => arg !== undefined)
+const firstNonUndefined = <T>(...args: (T | undefined)[]) => args.find(arg => arg !== undefined)
```
danielroe, line 12:
```suggestion
const firstNonUndefined = <T> (...args: (T | undefined)[]) => args.find(arg => arg !== undefined)
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -219,8 +218,8 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
                     const path = typeof to.value === 'string' ? to.value : router.resolve(to.value).fullPath
                     await Promise.all([
-                      nuxtApp.hooks.callHook('link:prefetch', path).catch(() => {}),
-                      !isExternal.value && preloadRouteComponents(to.value as string, router).catch(() => {})
+                      nuxtApp.hooks.callHook('link:prefetch', path).catch(() => { }),
+                      !isExternal.value && preloadRouteComponents(to.value as string, router).catch(() => { })
```
danielroe, line 222:
```suggestion
                      nuxtApp.hooks.callHook('link:prefetch', path).catch(() => {}),
                      !isExternal.value && preloadRouteComponents(to.value as string, router).catch(() => {})
```

danielroe:
Looks like this is well on the way to being merged in https://github.com/unjs/ufo/pull/175, so we probably can avoid merging here.

manniL:
Will adapt this PR as soon as https://github.com/unjs/ufo/pull/175 lands 👍

danielroe (approved):

PR #23225, opened 2023-09-14
chore: try nuxi-edge

pi0:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Try stability with latest changes in nuxi cli

(currently based on https://github.com/nuxt/nuxt/pull/23302 to allow better testing)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #23302, opened 2023-09-19
refactor(kit,nuxt,schema): use `consola` and improve test dx

pi0:
### 🔗 Linked issue

Extracted from #23225

### ❓ Type of change

### 📚 Description

This PR refactors all packages to leverage consola logger exported from `@nuxt/kit`. 

We use a magical conosla wrapper in CLI that was automatically wrapping all console usages but it can be easily missed when running in tests or with programmatic usage.

Exceptions:
- For tests, we add a guard of `process.client` to avoid clattering output
- For schema package, we directly use consola to avoid kit dependency
- For nuxt/app and other runtime code, we use console for bundle size

Added back a (warning level) eslint rule also for `no-console`

**Note:** This PR also contains small other eslint tweaks to remove all current warnings.

### 📝 Checklist


pi0:
TLDR: Logs are still somehow leaking and spamming CI/Test output. Also, we have random failures with Ubuntu preset (most likely concurrency) but i gues it is worth to iterate over such big refactor into smaller steps...

test/utils.ts
```diff
@@ -30,7 +30,11 @@ export async function renderPage (path = '/') {
     pageErrors.push(err)
   })
   page.on('request', (req) => {
-    requests.push(req.url().replace(url('/'), '/'))
+    try {
+      requests.push(req.url().replace(url('/'), '/'))
+    } catch (err) {
+      // TODO
```
danielroe, line 36: what kind of error was being triggered here?

pi0: The lovely context is not available one! Turns out we somehow still keep loading the Browser pages after tests are being done and we don't have context anymore.

This was more visible in new nuxi-edge case because we early close server properly now and made it more visible.

However i think would be nice later investigate root causes. Seems we are wasting too much on browser allocation in our tests.

danielroe (approved):

test/utils.ts
danielroe: nice find!

PR #23345, opened 2023-09-21
docs: add explicit `await` to clarify `sendRedirect` is async

hecktarzuli:
### ❓ Type of change

### 📚 Description
made the redirect async/await

### DIFF
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -315,8 +315,8 @@ export default defineEventHandler((event) => {
 ### Sending Redirect
 
 ```ts [server/api/foo.get.ts]
-export default defineEventHandler((event) => {
-  return sendRedirect(event, '/path/redirect/to', 302)
+export default defineEventHandler(async (event) => {
+  return await sendRedirect(event, '/path/redirect/to', 302)
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/2.guide/2.directory-structure/1.server.md
```diff
@@ -315,8 +315,8 @@ export default defineEventHandler((event) => {
 ### Sending Redirect
 
 ```ts [server/api/foo.get.ts]
-export default defineEventHandler((event) => {
-  return sendRedirect(event, '/path/redirect/to', 302)
+export default defineEventHandler(async (event) => {
+  return await sendRedirect(event, '/path/redirect/to', 302)
```
danielroe, line 319: It's fine to return the `sendRedirect` because that means the function returns the promise, and it will be awaited.

The issue is only if it's not returned.

danielroe: We could do this so it's more obviously copy/pasteable.

```suggestion
  await sendRedirect(event, '/path/redirect/to', 302)
```

danielroe (approved):

danielroe:
Thank you ❤️

docs/2.guide/2.directory-structure/1.server.md
```diff
@@ -315,8 +315,8 @@ export default defineEventHandler((event) => {
 ### Sending Redirect
 
 ```ts [server/api/foo.get.ts]
-export default defineEventHandler((event) => {
-  return sendRedirect(event, '/path/redirect/to', 302)
+export default defineEventHandler(async (event) => {
+  await sendRedirect(event, '/path/redirect/to', 302)
```
pi0, line 319: Would be nice idea to return + await (only for readability). This way we can chain the result and be upward compatible.

PR #23351, opened 2023-09-22
fix(nuxt): refetch both undefined/null values in `useAsyncData`

### LINKED ISSUE #22696: Unable to fetch data on client side if useNuxtData used before useFetch

### LINKED ISSUE #23349: `useFetch server: false` breaks `useNuxtData` and vice versa
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.3
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-vryf1x-a7jgca?file=app.vue,pages%2Findex.vue,pages%2Ftest.vue

### Describe the bug

Reloading a page where we use `use(Lazy)Fetch()` and also access the key of that fetch call using `useNuxtData` seems to omit the response from useFetch instead return the inital `null` value that useFetch has while mounting. 

### Additional context

_No response_

### Logs

_No response_

warflash:
### 🔗 Linked issue

Resolves #23349
Resolves  https://github.com/nuxt/nuxt/issues/22696

### ❓ Type of change

### 📚 Description

This ensures nuxt payload data is checked against `null` and `undefined` values in the (hopefully) least breaking way possible.
When `useNuxtData` is called before a client side `useAsyncData` call is resolved, `useNuxtData`will set the value of the key to `null`. `hasCachedData` so far only checked for `undefined` values so `null` would pass and be returned, both for the `useAsyncData` and `useNuxtData`.
I feel like this is not the cleanest way to go about this, setting they key to undefined like `clearNuxtData` does might be more concise. However then the return signature of `useNuxtData => <DataT | undefined>` might break existing user logic.

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -142,7 +142,9 @@ export function useAsyncData<
   const nuxt = useNuxtApp()
 
   const getCachedData = () => nuxt.isHydrating ? nuxt.payload.data[key] : nuxt.static.data[key]
-  const hasCachedData = () => getCachedData() !== undefined
+  const hasCachedData = () => nuxt.isHydrating
+    ? (nuxt.payload.data[key] !== null && nuxt.payload.data[key] !== undefined)
+    : nuxt.static.data[key] !== undefined
```
danielroe, line 147: what about:
```suggestion
  const hasCachedData = () => nuxt.isHydrating
    ? key in nuxt.payload.data
    : key in nuxt.static.data
```

danielroe, line 147: The issue here seems to be that when there is an _error_ the key is set to `null`.

danielroe (changes requested):
Would you be able to add a regression test to cover the situation in the linked issue? 🙏 

~~We can then look at fixing the currently failing tests.~~

danielroe (approved):

Issue #23352, opened 2023-09-22
docs: `nuxt build` - Better document disabling GZIP report


polarathene:
### Environment

```console
$ npx nuxi info

Working directory: /reproduction
Nuxt project info: 

------------------------------
- Operating System: Linux
- Node Version:     v20.6.1
- Nuxt Version:     3.7.3
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.8.1
- Builder:          -
- User Config:      vite
- Runtime Modules:  -
- Build Modules:    -
------------------------------
```

### Reproduction

```console
# Setup reproduction environment + demo project:
$ docker run --rm -it node:20-bookworm-slim bash
$ npx -y nuxi init --packageManager npm --gitInit false .

# Add `vite.build.reportCompressedSize: false` to `nuxt.config.ts`:
$ sed -i 's/devtools.*/vite: { build: { reportCompressedSize: false } }/' nuxt.config.ts
$ cat nuxt.config.ts
// https://nuxt.com/docs/api/configuration/nuxt-config
export default defineNuxtConfig({
  vite: { build: { reportCompressedSize: false } }
})


$ time npm run build

Nuxt 3.7.3 with Nitro 2.6.3


ℹ Building client...
ℹ vite v4.4.9 building for production...
ℹ ✓ 112 modules transformed.
ℹ .nuxt/dist/client/manifest.json                   1.59 kB
ℹ .nuxt/dist/client/_nuxt/error-500.c5df6088.css    1.95 kB
ℹ .nuxt/dist/client/_nuxt/error-404.7fc72018.css    3.63 kB
ℹ .nuxt/dist/client/_nuxt/entry.3c4ebcd6.css       12.13 kB
ℹ .nuxt/dist/client/_nuxt/error-500.e5a1dd9d.js     1.88 kB
ℹ .nuxt/dist/client/_nuxt/error-404.d5f8e506.js     6.53 kB
ℹ .nuxt/dist/client/_nuxt/entry.8be7e620.js       197.90 kB
ℹ ✓ built in 1.07s
✔ Client built in 1083ms


ℹ Building server...
ℹ vite v4.4.9 building SSR bundle for production...
ℹ ✓ 61 modules transformed.
ℹ .nuxt/dist/server/_nuxt/entry-styles.8755339a.mjs             0.08 kB
ℹ .nuxt/dist/server/_nuxt/error-404-styles.1e0dde27.mjs         0.15 kB
[...]

PR #23359, opened 2023-09-22
perf(vite): don't print server compressed size after vite build

danielroe:
### 🔗 Linked issue

related to nuxt/nuxt#23352

### ❓ Type of change

### 📚 Description

This disables vite reporting compressed size for ~~client +~~ server assets, which should save some time when building. We will disable client compressed size once we can rely on nitro doing this for us (https://github.com/unjs/nitro/issues/130).

### 📝 Checklist


polarathene:
> resolves #23352

Not quite, this was already applied in the reproduction example.

It doesn't allow opt-out from the final "Nitro Server" build step, which I'm not sure what the config setting for that is.

pi0 (approved):

Issue #23255, opened 2023-09-17
Accessibility roadmap


danielroe:
We might build these directly into nuxt or possibly via a core or adhoc module or even decoupled module (nuxt/a11y).

```[tasklist]
### Tasks
- [ ] https://github.com/nuxt/nuxt/issues/14673
- [ ] built-in [axe core](https://github.com/dequelabs/axe-core) or https://github.com/vue-a11y/vue-axe?
- [ ] devtools integration? cc: @antfu
```

franklin-tina:
😍

chstappert:
This is an absolutely great idea.

mayank99:
I just wanted to chime in to say that I love this initiative. It would be great if Nuxt helped with accessible client-side routing, which is is often the biggest omission in full stack frameworks.

Route announcer is a good first step, but is generally not enough and also needs to be possible to disable because it could conflict with focus announcements. Some other things to consider adding:
- a convenient way to move focus on every navigation.
- skip links for traversing to/from the top-most layout that changed.
- documentation and examples on accessible routing patterns.

I would recommend reading Marcy Sutton's writeup on [accessible client-side routing](https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/).

Happy to discuss this further and help in any way I can :)

danielroe:
These are great ideas to add to the list; thank you.

I think route announcer should be a built-in component like `<NuxtLoadingIndicator>` that is included in the default `app.vue` but can be removed/customised.

Gonzo17:
Hey guys! I was thinking about writing a Nuxt module that brings axe-core to the Nuxt Devtools because I was looking for support in Vue/Nuxt and it seems that there is no actively maintained tool. Luckily I found this issue! Is someone already working on it? Can I help on this topic?

margaretjoanmiller:
is this still being worked on? It doesn't seem like a lot of attention has been given to this issue but maybe things have been changing outside the issue?

ultrasites:
I have written a simple a11y plugin for nuxt 3 / vue 3. Maybe it is useful for you!

[https://github.com/ultrasites/nuxt-a11y](https://github.com/ultrasites/nuxt-a11y)

maximepvrt:
🚀 https://github.com/nuxt/a11y

PR #23368, opened 2023-09-22
feat(nuxt): `useId` composable

TakNePoidet:
### ❓ Type of change

### 📚 Description
The `useId` composable creates a unique id.

### 📝 Checklist


packages/nuxt/src/app/composables/id.ts
```diff
@@ -0,0 +1,17 @@
...
+  const autoKey = typeof args[args.length - 1] === 'string' ? args.pop() : undefined
+  if (typeof args[0] !== 'string') {
+    args.unshift(autoKey)
+  }
+  const [_key] = args as [string]
+  if (!_key || typeof _key !== 'string') {
+    throw new TypeError('[nuxt] [useId] key must be a string: ' + _key)
+  }
+
+  return readonly(ref(useIdKeyPrefix + _key))
```
danielroe, line 16: Is there a reason we're using a ref rather than directly returning the string?

I think we could also simplify the check, e.g. a single `typeof`.

docs/3.api/1.composables/use-id.md
```diff
@@ -0,0 +1,25 @@
+---
+title: "useId"
+description: The useId composable creates a unique id.
+---
+
+# `useId`
```
danielroe, line 6: Might be worth naming this something like `getUniqueID` rather than `use*` as the latter is more often associated with composables, but this doesn't rely on Nuxt or Vue context.

danielroe (commented):
Nice idea! and relevant for https://github.com/nuxt/nuxt/issues/23255.

docs/3.api/1.composables/use-id.md
TakNePoidet: maybe it's better to use a `uniqueID`/ `uniqueId`?

Atinux: `getUniqueID()`

docs/3.api/3.utils/get-unique-id.md
```diff
@@ -0,0 +1,25 @@
...
+title: "getUniqueID"
+description: The getUniqueID utility returns a unique identifier.
+---
+
+# `getUniqueID`
+
+The `getUniqueID` utility returns a unique identifier.
+
+```ts
+getUniqueID<T>(): string
```
danielroe, line 11:
```suggestion
getUniqueID(): string
```

packages/vite/src/plugins/composable-keys.ts
```diff
@@ -104,6 +104,10 @@ export const composableKeysPlugin = createUnplugin((options: ComposableKeysOptio
               if (stringTypes.includes(node.arguments[0]?.type)) { return }
               break
 
+            case 'getUniqueID':
+              if (stringTypes.includes(node.arguments[0]?.type)) { return }
+              break
```
danielroe, line 109: I think we do not need to add this here as L100 should cover us.

packages/nuxt/src/app/composables/id.ts
```diff
@@ -0,0 +1,11 @@
+const getUniqueIDKeyPrefix = '$id'
+
+export function getUniqueID(): string
+export function getUniqueID (...args: any): string {
+  const key = args[0]
+  if (typeof key !== 'string') {
+    throw new TypeError(`[nuxt] [getUniqueID] key must be a string: ${key}}`)
+  }
+
+  return getUniqueIDKeyPrefix + key
```
pi0, line 10: Is this functional now?

danielroe: What do you mean - why do you ask?

(personally I think we should probably omit the prefix)

pi0:
This is a nice idea however I think there are easy pitfalls cases if we base such generically named composables on transform-based key generators for runtime extending limitations currently we have with key-based data fetching composables.

- Using `getUniqueID` in a component that is rendered multiple times (like a UI text input which i suppose would be common for accessibility), it always uses one ID which is not probably what we want

image

- Using `getUniqueID()` in reusable utilities or externalized dependencies won't be effective and can lead to runtime issues

- `getUniqueID()` composable naming sounds it generates an increamental/unique name by each call which it won't

- The `getUniqueID` can be possibly conflict with future nitro/uncrypto auto imported utils sharing common name (maybe also otherlibs or even vue-core sometime!)


---


Having these in mind, i think might makes sense to make sure we use a name that is clear it comes from nuxt and also it is an code-location-based static key. Something like `getNuxtStaticID()` (name is terrible just to give some idea)

packages/nuxt/src/app/composables/id.ts
pi0: Yeah without a prefix should be unique enough...  Added more to the context in https://github.com/nuxt/nuxt/pull/23368#issuecomment-1738799146

TakNePoidet:
You can try adding a local ID
```typescript
  const nuxt = useNuxtApp()
  const localId = (nuxt.payload?.localIds?.[key] ?? 0) + 1
  nuxt.payload.localIds = nuxt.payload.localIds ?? {}
  nuxt.payload.localIds[key] = localId
```

Then repeatedly calling the component will not cause conficts
```vue
<script setup lang="ts">
import {getUniqueID} from '#imports'

withDefaults(defineProps<{ id?: string }>(), {
  id: () => getUniqueID()
});
</script>

<template>
  <label :for="id">Label</label>
  <br>
  <input type="text" :id="id" :placeholder="id"/>
</template>

```
image

danielroe:
Those are both good suggestions. A hybrid approach that gets a unique constant and uses it as prefix + an additional per-instance value seems good to me.

mayank99:
> I would say it should be unique not only per component instance, but globally unique (per HTML page). Because different instances of the same component on a single page must not have collisisons of IDs of their inner HTML elements.

yes, i should have clarified: what i really mean is it should be unique for each component instance, regardless of whether that is two instances of the _same_ component or one instance each of two _different_ components.

```vue
<Comp1 /> // id1
<Comp1 /> // id2
<Comp2 /> // id3
```

danielroe:
I'm marking this as draft because I think it needs to produce globally unique (and stable) ids tied to component instances, as the last two comments have pointed out.

Atinux:
Added back to `useId` to be similar to [React useId](https://react.dev/reference/react/useId) to avoid more mental work and since it has to be used inside a component.

I updated it to generate unique ID per component instance as well as being able to generate multiple unique id in the same component instance.

docs/3.api/2.composables/use-id.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "useId"
+description: Generate a ssr-friendly unique Id that can be passed to accessibility attributes.
```
danielroe, line 3:
```suggestion
description: Generate a ssr-friendly unique identifier that can be passed to accessibility attributes.
```

docs/3.api/2.composables/use-id.md
```diff
@@ -0,0 +1,29 @@
...
+---
+
+`useId` generates a ssr-friendly unique Id that can be passed to accessibility attributes.
```
danielroe, line 6:
```suggestion
`useId` generates an SSR-friendly unique identifier that can be passed to accessibility attributes.
```

docs/3.api/2.composables/use-id.md
```diff
@@ -0,0 +1,29 @@
...
+
+Call `useId` at the top level of your component to generate a unique ID:
```
danielroe, line 8:
```suggestion
Call `useId` at the top level of your component to generate a unique string identifier:
```

docs/3.api/2.composables/use-id.md
```diff
@@ -0,0 +1,29 @@
...
+</template>
+```
+
+## Parameters
+
+`useId` does not take any parameters.
+
+## Returns
+
+`useId` returns a unique ID string associated with this particulat `useId` call in this particular component.
```
danielroe, line 29:
```suggestion
`useId` returns a unique string associated with this particular `useId` call in this particular component.
```

packages/nuxt/src/app/composables/id.ts
```diff
@@ -0,0 +1,40 @@
...
+  let id
+  if (nuxt.payload.serverRendered && nuxt.isHydrating) {
+    instance._nuxtIdIndex = instance._nuxtIdIndex || 0
+    const serverIds = instance.vnode.el?.getAttribute?.(ATTR_KEY)?.split?.(',') || []
+    if (serverIds.length) {
+      id = parseInt(serverIds[instance._nuxtIdIndex], 10) || 0
+      instance._nuxtIdIndex++
+    } else {
+      if (!instance.vnode.el?.getAttribute && (!instance.parent || instance.parent.type.name !== 'ClientOnly')) {
+        console.error('[nuxt] useId() needs to be used with a component having on single root element.')
```
danielroe, line 24: Note that this will not work if we have a component which calls `useId` being the single root element of another element which calls `useId`. That is not an inconceivable situation with (for example) form elements in a UI library.

test/basic.test.ts
```diff
@@ -874,6 +874,28 @@ describe('composables', () => {
...
+    // Wrapped in <ClientOnly>
+    expect(html).not.toContain('data-n-ids="6,7"')
+    expect(html).not.toContain('id="n:6"')
+    expect(html).not.toContain('id="n:7"')
+
+    const { page } = await renderPage('/use-id')
+    const clientHtml = await page.content()
+    expect(clientHtml).not.toContain('data-n-ids="6,7"')
+    expect(clientHtml).toContain('id="n:6"')
+    expect(clientHtml).toContain('id="n:7"')
```
danielroe, line 897:
```suggestion
    expect(clientHtml).toContain('id="n:7"')
    await page.close()
```

packages/nuxt/src/app/composables/id.ts
Atinux: What do you suggest?

danielroe: happy for me to push something?

Atinux: Always ❤️

manniL:
Linking https://github.com/vuejs/rfcs/discussions/557 here

alex-eliot:
> Given that the main use case for this utility is to generate unique `IDREF`s (for labeling inputs and creating other ARIA associations), I would expect that it is:
> 
> * unique per component instance
> * stable after hydration
> 
> Here's a blog post that has some useful ideas and links: https://www.jovidecroock.com/blog/preact-use-id

I have given it a try locally (while also fixing some edge cases that are present in its current state), but I was unable to find a way to make Preact's solution work in SSR.

On the client, it is possible to locate a certain vnode's path by traversing up the tree until you reach the root, storing each node's index on its parent's children. But on the server, the `instance.subTree` property is `null` and after debugging I didn't find any information about the current vnode that could either help specify its path, or at least be used as a unique identifier that remains stable after hydration (`instance.uid` is not stable for example).

If we are able to locate the vnode's path in the tree on the server, or have some unique identifier to each instance that remains stable after hydration, then this composable could be used in any component without having to store the ids in the root element's attributes (it could go directly to the payload as a `Record<string, number[]>`), which means that its template would be completely irrelevant to whether you can use `useId()`.

I'm not very well-versed in either Nuxt or Vue internals, so please tell me if I've missed something.

manniL:
As a note - this is planned for Vue 3.5 to be supported natively. [Source](https://x.com/youyuxi/status/1745379112456429688?s=20)

alex-eliot:
Oh that is very good news, thanks for sharing. Should be much more reliable this way as Vue handles the vnode tree.

CernyMatej:
Does that mean that this isn't coming to Nuxt and will be entirely implemented in Vue core?

danielroe (approved):
I think we can go ahead and merge this for now, and swap across to the Vue native implementation when it comes (and iterate to fix any bugs). For now, there's nothing Nuxt-specific about this implementation other than the automatic provision of a key that is unique per file + line number - and all data is transferred from server -> client via data attributes.

fabkho:
I get the following error when trying to use `useId()`:
`The requested module '*/node_modules/nuxt/dist/app/components/client-only.js?v=d16211f9' does not provide an export named 'clientOnlySymbol'`

danielroe:
Try removing `node_modules/.cache`. It sounds like vite may not have updated its cache with the new version.

fabkho:
@danielroe fixed it! Thanks a lot.

DJafari:
image
are you sure this is ssr friendly ?

fabkho:
@DJafari i am actually having the same problem.

danielroe:
If you have an issue, please open a new issue with a reproduction rather than commenting on this PR 🙏

DJafari:
@danielroe Because I cannot recreate this problem in the new project, I am commenting here

but it's has a bug, in some case `data-n-ids` attribute doesnt add to element

i use `useId` in two component, in one component `data-n-ids` added, but in another, not

image

manniL:
@DJafari no matter if you can reproduce it or not, commenting on a merged PR is **not the place** to report bugs. That should be an own issue. If you want reproduce the issue the problem might be part of your code then?

PR #23386, opened 2023-09-24
fix: prevented data-v filter added in #23095 from impacting styles

### LINKED ISSUE #23385: Fix for #23051 broke
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.1
- CLI Version:      3.7.3
- Nitro Version:    2.6.2
- Package Manager:  npm@9.4.2
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

You can use my old preproduction from #23051 https://stackblitz.com/edit/github-ydxe3t?file=pages%2Findex.vue

See that the scoped style (margin) applies to the component. However, if you update Nuxt to 3.7.3, the style no longer applies, since the data-v attribute was removed. 

### Describe the bug

The issue #23051 was fixed in #23095.

However, this prevented scoped style from being applied to the component. 

### Additional context

Sorry for the delayed feedback. As I noted in https://github.com/nuxt/nuxt/pull/23095#issuecomment-1712583276 the removal of data-v attribute entirely lead to styles not being reproduced. I already debugged the issue and I'm going to create a PR with a simple suggested fix.

This PR will revert the expected behavior of scoped styles as in 3.7.1 while keeping @huang-julien 's fix for 404 on soft navigation due to different component hash.

### Logs

_No response_

arsengoian:
### 🔗 Linked issue

Fix #23385

### ❓ Type of change

### 📚 Description

Fixes #23385
The data-v attribute is only removed for the hash function.
Please validate, this is my first PR 😉

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -50,7 +50,8 @@ export default defineComponent({
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
-    const hashId = computed(() => hash([props.name, props.props, props.context, props.source]))
+    const propsForHash = props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {}
+    const hashId = computed(() => hash([props.name, propsForHash, props.context, props.source]))
     const instance = getCurrentInstance()!
     const event = useRequestEvent()
     // TODO: remove use of `$fetch.raw` when nitro 503 issues on windows dev server are resolved
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -50,8 +50,8 @@ export default defineComponent({
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
-    const propsForHash = props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {}
-    const hashId = computed(() => hash([props.name, propsForHash, props.context, props.source]))
+    const propsForHash = computed(() => props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {})
+    const hashId = computed(() => hash([props.name, propsForHash.value, props.context, props.source]))
     const instance = getCurrentInstance()!
     const event = useRequestEvent()
     // TODO: remove use of `$fetch.raw` when nitro 503 issues on windows dev server are resolved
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -50,7 +50,8 @@ export default defineComponent({
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
-    const hashId = computed(() => hash([props.name, props.props, props.context, props.source]))
+    const propsForHash = props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {}
```
huang-julien, line 53: This won't be reactive, you need to move it into the computed or to make it computed

huang-julien (changes requested):

packages/nuxt/src/app/components/nuxt-island.ts
arsengoian: Good point

arsengoian:
Will this work?

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -50,7 +50,8 @@ export default defineComponent({
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
-    const hashId = computed(() => hash([props.name, props.props, props.context, props.source]))
+    const propsForHash = computed(() => props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {})
+    const hashId = computed(() => hash([props.name, propsForHash.value, props.context, props.source]))
```
danielroe, line 54:
```suggestion
    const filteredProps = computed(() => props.props ? Object.fromEntries(Object.entries(props.props).filter(([key]) => !key.startsWith('data-v-'))) : {})
    const hashId = computed(() => hash([props.name, filteredProps.value, props.context, props.source]))
```

packages/nuxt/src/components/runtime/server-component.ts
```diff
@@ -11,8 +11,7 @@ export const createServerComponent = (name: string) => {
         return h(NuxtIsland, {
           name,
           lazy: props.lazy,
-          // #23051 - remove data-v attributes
-          props: Object.fromEntries(Object.entries(attrs).filter(([key]) => !key.startsWith('data-v-')))
+          props: attrs
```
danielroe, line 14:
```suggestion
          props: filteredProps.value
```

huang-julien (commented):
I also think that you'll need to update the test in tests/nuxt/nuxt-island.test.ts, it will probably break

arsengoian:
Yeah, just saw that too. Will do

huang-julien (approved):
lgtm :+1: will probably try to implement nuxt-island runtime test later

PR #23389, opened 2023-09-24
docs: extend description of `handler` param for useAsyncData composable

### LINKED ISSUE #23388: Docs: explicitly describe that `useAsyncData` callback must return a value
### Describe the feature

Sometimes people are using `useAsyncData` for side-effects, like fetching the data and passing it to the store. Although this is not what was `useAsyncData` built for, but people still going to use it. And later they find that useAsyncData is called twice, because their side-effect callback returns nothing.

I see 2 ways of resolving this problem:
1. Create new hook in addition to `useAsyncData` and encourage people to use it
2. Update the docs to display some warnings about side-effect usage and about return value
3. Accept `undefined` as valid value (found: https://github.com/nuxt/nuxt/pull/23351)

I prefer 2nd option, but want to have some thoughts about it

### Additional information

### Final checks

MaxKostenko:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23388

### ❓ Type of change

### 📚 Description
Some developers use useAsyncData composable for fulfilling `pinia` store. As a result, the handler returns nothing (`undefined`) that leads to API request duplication on the client side. I extended the handler param's description to emphasize this feature.

### 📝 Checklist


Atinux:
Interesting, thanks for the feedback, why using `useAsyncData` in that case and not directly using an `await` to fill your store wrapped with an `if` statement to check if not already filled?

MaxKostenko:
First of all, it is not my personal feedback it is one of the questions that people asked at least twice this week in our community.
 If you open the official pinia [guide](https://pinia.vuejs.org/ssr/nuxt.html) you will see:
`
<script setup>
const store = useStore()
const { data } = await useAsyncData('user', () => store.fetchUser())
</script>
`
That is correct if fetchUser returns data, but not all developers understand that it must return something. 
A lot of people still look at the useAsyncData like full analog asyncData hook from Nuxt2

posva:
I think that example in the pinia docs is a bit misleading. I will have to think about a different example that makes sense and use that instead. Maybe, not use `useAsyncData()` at all in that example

docs/3.api/1.composables/use-async-data.md
```diff
@@ -51,7 +51,7 @@ type AsyncDataRequestStatus = 'idle' | 'pending' | 'success' | 'error'
 ## Params
 
 * **key**: a unique key to ensure that data fetching can be properly de-duplicated across requests. If you do not provide a key, then a key that is unique to the file name and line number of the instance of [`useAsyncData`](/docs/api/composables/use-async-data) will be generated for you.
-* **handler**: an asynchronous function that returns a value
+* **handler**: an asynchronous function that must return a value that should be different from `undefined` to avoid request duplication on the client side
```
danielroe, line 54:
```suggestion
* **handler**: an asynchronous function that must return a truthy value (for example, not `undefined` or `null`)
```

MaxKostenko: As I remember that ‘return null’ also works fine.

MaxKostenko: Duplication happens because keys with undefined values are automatically removed from hydrations payload
```
JSON.stringify({ a: 1, b: undefined })
'{"a":1}'
JSON.stringify({ a: 1, b: null })
'{"a":1,"b":null}'
```

MaxKostenko: https://github.com/nuxt/nuxt/blob/5abd8a231f26a2e08071a59caa87b80eaaa6cd91/packages/nuxt/src/app/composables/asyncData.ts#L145C3-L145C60

danielroe: Keys with undefined values should still be serialised, as we use `devalue`. And the reason I'm suggesting not returning `null` is for safety - see https://github.com/nuxt/nuxt/pull/23351.

danielroe (approved):

vushe:
It's quite tricky to understand, how the fetching data from store is supposed to work.

>  why using useAsyncData in that case and not directly using an await to fill your store wrapped with an if statement to check if not already filled?

Probably because you may want to refresh data and check loading state, for example. `useAsyncData()` is rather a compact and an elegant way of fetching data, isn't it? Another reason is that it [handles adding responses to the Nuxt payload so they can be passed from server to client without re-fetching the data on client side when the page hydrates](https://nuxt.com/docs/api/composables/use-async-data). 

So yes, not a surprise if people prefer fetching from store, having side effects and still utilize `useAsyncData()` for rendering component. Even more, if you load data directly to component, what's the point in store? 

Also when you're rushing to migrate your app from Nuxt 2 to Nuxt 3, you may want to copy approaches used previously (though `asyncData` hook in Nuxt 2 was supposed to return as well).

I can't say how many people need it, but probably documentation on both Nuxt and Pinia require more clarifications about data fetching from store. If fetching data from store isn't a recommended way, it could be explicitly explained.

MaxKostenko:
@vushe  Nuxt 2 was based on vuex which is always global, so any state set on the server side with vuex was transited to client. So it was ok to fill a store there.

I guess store usage should look like this:
```
export const useUserStore = defineStore('user', () => {
    const user = ref();
    
    async function getUser() {
         if(user.value) return user;
         
         return await fetchUser();
    }
    
    async function fetchUser() {
         user.value = await $fecth('/api/user');
         return user;
    }

    return {
        user,
        getUser,
        fetchUser
    }
})

<script setup>
const userStore = useUserStore();
await userStore.getUser();
</script>
```
But I agree this way causes some questions: 
Will it block routing like useAsyncData or not? 
Depending on the answer we have another question. How do we achieve the opposite behavior?
And the main question is: Why Nuxt3 has too many ways to do the same thing? (For example [useFetch](https://nuxt.com/docs/api/composables/use-fetch#usefetch) doesn't fit any project if it is bigger than a demo IMHO)

In the same time, people who just start with Nuxt very often do something like this:
```
<script setup>
const name = ref()
const email = ref()

async function saveCustomer() {
       await useFetch(() => `/updateCustomer?name=${unref(name)}&email=${email}`)
}
</script> 
```
I know that documentation recommends using `$fetch`

vushe:
Thanks @MaxKostenko! What you wrote is very close to how I understood this, including thoughts about `useFetch` (which is imho more relevant as an addition on top of the framework, e.g. in a [library](https://vueuse.org/core/useFetch/)). 

When we fetch from `script setup`, it is already on client, correct? Just one call then. What if I user data is something that should be loaded from [route middleware](https://nuxt.com/docs/guide/directory-structure/middleware)? Then we either need `useAsyncData()` or we have to manually block second request as mentioned [earlier](https://github.com/nuxt/nuxt/pull/23389#issuecomment-1733846449). Is there another way?

In addition, if we elaborate your first example more with `try-catch`, we can expect [known issues](https://nuxt.com/docs/api/composables/use-nuxt-app#runwithcontext). 

So this is what I mean when writing, it's a challenge to understand, how fetching is supposed to work from store. Not mentioning problems that e.g. i18n brings when it can't suddenly detect a context from `Promise.all()`.

Yes, if it's a demo, you don't care from where you load data or where you store them. However, a developer of a solid application would be definitely happy to have a clear example instead of guessing.

PR #23428, opened 2023-09-27
fix(nuxt): initialise `asyncData` errors with `null`

### LINKED ISSUE #23411: Mismatch between value and type definition of error returned by data fetching composables
### Environment

- Node Version:     v20.7.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  pnpm@8.7.6

### Reproduction

https://codesandbox.io/p/sandbox/fragrant-bush-m6xj7x?file=%2Fapp.vue%3A1%2C1

### Describe the bug

When conditionally checking `useAsyncData` for error, that is `error.value !== null`, it doesn't pass, even though (according to type definition) the expected value should be `null` when there is no error. Instead, `undefined` is returned.

`undefined` in this case is present before the function is run. Further attempts to check the error's value return `null`.

image

### Additional context

_No response_

### Logs

_No response_

DamianGlowala:
### 🔗 Linked issue

Closes #23411

### ❓ Type of change

### 📚 Description

This PR aligns the initial values of error(s) set in `asyncData` composable and Nuxt payload with their type definitions.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -373,11 +373,11 @@ export function clearNuxtData (keys?: string | string[] | ((key: string) => bool
       nuxtApp.payload.data[key] = undefined
     }
     if (key in nuxtApp.payload._errors) {
-      nuxtApp.payload._errors[key] = undefined
+      nuxtApp.payload._errors[key] = null
     }
     if (nuxtApp._asyncData[key]) {
       nuxtApp._asyncData[key]!.data.value = undefined
-      nuxtApp._asyncData[key]!.error.value = undefined
+      nuxtApp._asyncData[key]!.error.value = null
       nuxtApp._asyncData[key]!.pending.value = false
       nuxtApp._asyncData[key]!.status.value = 'idle'
     }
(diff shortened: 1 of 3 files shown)


danielroe:
Do you think we could add a test page in `test/fixtures/basic`?

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -149,7 +149,7 @@ export function useAsyncData<
     nuxt._asyncData[key] = {
       data: ref(getCachedData() ?? options.default!()),
       pending: ref(!hasCachedData()),
-      error: toRef(nuxt.payload._errors, key),
+      error: ref(nuxt.payload._errors[key] ?? null),
```
danielroe, line 152: this will lose reactivity/sync state. I think we likely need to keep using `toRef` but initialise it beforehand

test/nuxt/composables.test.ts
```diff
@@ -125,6 +125,12 @@ describe('useAsyncData', () => {
     expect(pending.value).toBe(false)
   })
 
+  // https://github.com/nuxt/nuxt/issues/23411
+  it('should initialize with error set to null when immediate: false', async () => {
+    const { error } = await useAsyncData(() => '', { immediate: false })
+    expect(error.value).toBe(null)
```
danielroe, line 131:
```suggestion
    const { error, execute } = useAsyncData(() => ({}), { immediate: false })
    expect(error.value).toBe(null)
    await execute()
    expect(error.value).toBe(null)
```

danielroe (approved):

Issue #22752, opened 2023-08-23
Automatic import of `useRoute` is not working properly on production build


trandaison:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.16.0`
- Nuxt Version:     `3.6.5`
- Nitro Version:    `2.5.2`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `telemetry`, `modules`, `pinia`
- Runtime Modules:  `@pinia/nuxt@0.4.11`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/p/sandbox/sad-lake-cjxzvt?file=/stores/test.ts:10,6

### Describe the bug

1. Visit the reproduction link
2. Click on any `/?id=1..n` button to change the id value on the route URL.
3. The value of `currentData` getter is updated correctly ✅
4. Click the `/home` link.
5. At the home page, click on any `/?id=1..n` button to navigate back to the index page.
6. Repeat step no. 2, but this time, the value of getter will not be updated ❌

![261888115-9201b18c-3654-4288-b614-9790d52000c3](https://github.com/nuxt/nuxt/assets/13447235/7b2500bc-06eb-4af9-a4aa-cd2424913997)

### Additional context

- In the sandbox above, this bug happen. In the localhost, you have to build production to reproduce the bug

```bash
npx nuxi generate
npx nuxi preview
```

- This happen when navigate to the page in client side. If you enter the page from URL, the getter still working.
- If I turn of the ssr from the nuxt config. And move the useRoute statement inside the computed/getter, it works.

```diff
- const route = useRoute();
  const currentData = computed(() => {
+   const route = useRoute();
    const index = Number(route.query.id || 1);
    return data.value[index];
  });
```
However this code never working on server side, cuz the `useRoute()` is call outside the setup context.

- I already open a issue on pinia repo (https://github.com/vuejs/pinia/issues/2361). However, adding a manual import from vue-router does fix it:

```ts
import { useRoute } from 'vue-router'
```

Heunsig:
I have the same problem. I am working on nuxt@^3.7.0

danielroe:
Let's track in https://github.com/nuxt/nuxt/issues/28804 - see https://github.com/nuxt/nuxt/issues/28804#issuecomment-2327483434 for an explanation of what is happening and why.

PR #23440, opened 2023-09-27
fix(nuxt): rename dts to avoid shadowing w/ baseUrl

### LINKED ISSUE #23435: Vue Router issue with upgrade of Nuxt 3.6.5 to 3.7.4 (tsconfig baseUrl)
### Environment

* Node Version: v18.17.1
* Nuxt Version: 3.7.4
* Package Manager: npm@9.6.7

### Reproduction

1. Create new nuxt app `npx nuxi@latest init my-app`
2. Add `tsconfig.json` at root level with the following content:
```json
{
  "extends": "./.nuxt/tsconfig.json",
  "compilerOptions": {
    "baseUrl": "./.nuxt"
  }
}
```
3. Import something from `#vue-router`:
```html
<script setup lang="ts">
import { createRouter } from "#vue-router";
</script>
```

The following error message appears:
```
Module '"vue-router"' has no exported member 'createRouter'.ts(2305)
```

See Example on StackBlitz: https://stackblitz.com/edit/github-pzljlg. However, the error does not appear on StackBlitz, but only on a local project.

### Describe the bug

Nuxt 3.7 removed the baseUrl in `.nuxt/tsconfig.json`. However, tools like Vitest and Storybook rely on the baseUrl in order to resolve path aliases.

* Adding the baseUrl back into the project's `tsconfig.json` at the root level breaks existing imports from `#vue-router`.
  * `./nuxt/vue-router.d.ts` contains the line: `export * from 'vue-router'`, but it seems something isn't working correctly. Removing this file seems to remove the TypeScript issue.

* The temporary solution for now is to create a new `tsconfig.json` file only for Vitest and Storybook. Only this config file contains the baseUrl.

Ideally, there would be a way to specify the baseUrl without breaking the import for `vue-router`. What would be the correct way to specify the baseUrl so that it works the same way as before the upgrade?

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23435

### ❓ Type of change

### 📚 Description

This resolves an issue with setting baseUrl to `.nuxt` (which probably shouldn't be done anyway), where this would end up shadowing `vue-router` with `.nuxt/vue-router.d.ts`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -96,7 +96,7 @@ export default defineNuxtModule({
     }
 
     addTemplate({
-      filename: 'vue-router.mjs',
+      filename: 'vue-router-stub.mjs',
       // TODO: use `vue-router/auto` when we have support for page metadata
       getContents: () => 'export * from \'vue-router\';'
     })
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -75,11 +75,11 @@ export default defineNuxtModule({
 
     // adds support for #vue-router alias (used for types) with and without pages integration
     addTemplate({
-      filename: 'vue-router.d.ts',
+      filename: 'vue-router-stub.d.ts',
       getContents: () => `export * from '${useExperimentalTypedPages ? 'vue-router/auto' : 'vue-router'}'`
     })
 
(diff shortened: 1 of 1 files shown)


trandaison:
Hi @danielroe,
Does this solve this issue? https://github.com/nuxt/nuxt/issues/22752

PR #23449, opened 2023-09-28
fix(nuxt): stringify cookie values before broadcasting them

### LINKED ISSUE #21651: BroadcastChannel.postMessage in useCookie has issue with ProxyObject
### Environment

Using Nuxt 3 with a forked [composables/cookie.ts from main](https://gist.github.com/tommie/23539b45f55547d54eac3fdcf8d14231).

```
Google Chrome | 113.0.5672.126 (Official Build) (64-bit)
-- | --
OS | Linux
JavaScript | V8 11.3.244.11
```

### Reproduction

https://stackblitz.com/edit/github-hjpstl?file=app.vue

### Describe the bug

Caused by https://github.com/nuxt/nuxt/pull/20970

```
cookie.ts?t=1687256275358:29 Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'BroadcastChannel': #<Object> could not be cloned.
    at callback (http://localhost:8080/_nuxt/composables/cookie.ts?t=1687256275358:29:16)
    at watch.deep (http://localhost:8080/_nuxt/composables/cookie.ts?t=1687256275358:48:11)
    at callWithErrorHandling (http://localhost:8080/_nuxt/node_modules/.cache/vite/client/deps/chunk-RABNIDPS.js?v=3f801703:213:18)
```

If I wrap [`cookie.value`](https://gist.github.com/tommie/23539b45f55547d54eac3fdcf8d14231#file-cookie-ts-L49) in Vue's `toRaw`, the error goes away. I can't find any information about ProxyObject not being allowed in `postMessage`...

### Additional context

_No response_

### Logs

_No response_

hendrikheil:
### 🔗 Linked issue
https://github.com/nuxt/nuxt/issues/21651

### ❓ Type of change

### 📚 Description
Resolves #21651 by preparing the cookie value for the [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) that's used by `BroadcastChannel.postMessage`.

It does this simply by cloning the value which might potentially hold nested Refs. `toRaw` unfortunately only applies to the root Ref, not to any nested Refs.

### 📝 Checklist


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -1,10 +1,9 @@
-import type { Ref } from 'vue'
+import type { MaybeRef, Ref } from 'vue'
 import { getCurrentInstance, nextTick, onUnmounted, ref, toRaw, watch } from 'vue'
 import type { CookieParseOptions, CookieSerializeOptions } from 'cookie-es'
 import { parse, serialize } from 'cookie-es'
 import { deleteCookie, getCookie, getRequestHeader, setCookie } from 'h3'
 import type { H3Event } from 'h3'
-import destr from 'destr'
 import { isEqual } from 'ohash'
```
danielroe, line 7:
```suggestion
import destr from 'destr'
import { isEqual } from 'ohash'
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -114,3 +114,7 @@ function writeServerCookie (event: H3Event, name: string, value: any, opts: Cook
     // else ignore if cookie doesn't exist in browser and value is null/undefined
   }
 }
+
+function prepareStructuredClone (value: MaybeRef<any>): any {
+  return JSON.parse(JSON.stringify(toRaw(value)))
+}
```
danielroe, line 120: what about instead literally using the decode/encode option of the cookie?

hendrikheil: We could do that. Though, I think since this is specifically for making the payload encodable for structuredClone, it could introduce unexpected problems when a user supplies an encoder that isn't compatible. 

We'd also have to throw an error if the encode or decode functions are undefined, which is theoretically possible.

I'd probably not open up that error source. Though we could also make `prepareStructuredClone` user definable in the CookieOptions.

hendrikheil:
This PR will be squashed right? I'd force push to avoid the little commits, if not ;-)

danielroe:
Yes, the pr will be squashed. I guess I'm thinking that the encoder has to produce a string to save the cookie. (It won't be undefined as we have defaults.)

packages/nuxt/src/app/composables/cookie.ts
danielroe: I don't think this will open the error source as the encoder/decoder are used anyway to save/read the cookie. We should use the defaults if the user doesn't supply a custom encoder/decoder.

hendrikheil: Makes sense. Made that change, though I had to move the default encoder and decoder to separate functions as typescript could not infer `CookieDefaults.encode` and `CookieDefaults.decode` to be always defined.

danielroe: I think we could use `!` or make an assertion on L34 instead.

danielroe: pushed a refactor to use `satisfies` to avoid need for `!` entirely.

danielroe (approved):

danielroe:
Thank you ❤️

hendrikheil:
@danielroe I noticed you removed the ternary. This could theoretically lead to `encode` or `decode` being undefined. `{...defaults, ..._opts }`, unlike `defu` overwrites when a key is set to `undefined`.

Not sure if thats something this should really worry about as it would clearly be the users fault, but it's theoretically possible, so I wanted to mention it :)

danielroe:
Ah, you're right. I initially thought it would error if the user did that, but I think implementation in `cookie-es` would actually just fallback to `encodeURIComponent` directly.

I think this is likely an edge case but if you want to, we could also cover this scenario. I just don't want to add too much runtime code to cover this scenario.

Aietes:
Has this actually been merged? I'm still facing the issue on the latest nuxt release: 
```
DOMException: Failed to execute 'postMessage' on 'BroadcastChannel': #<Object> could not be cloned.
    at SupabaseAuthClient._notifyAllSubscribers (http://localhost:3000/_nuxt/node_modules/.cache/vite/client/deps/chunk-A274CJAX.js?v=fac2f4d6:2156:31)
    at SupabaseAuthClient._recoverAndRefresh (http://localhost:3000/_nuxt/node_modules/.cache/vite/client/deps/chunk-A274CJAX.js?v=fac2f4d6:2101:20)
    at async SupabaseAuthClient._initialize (http://localhost:3000/_nuxt/node_modules/.cache/vite/client/deps/chunk-A274CJAX.js?v=fac2f4d6:987:7)
    at async http://localhost:3000/_nuxt/node_modules/.cache/vite/client/deps/chunk-A274CJAX.js?v=fac2f4d6:950:16
_recoverAndRefresh @ chunk-A274CJAX.js?v=fac2f4d6:2105
await in _recoverAndRefresh (async)
_initialize @ chunk-A274CJAX.js?v=fac2f4d6:987
await in _initialize (async)
```

danielroe:
It was released in Nuxt 3.8. Are you sure the log you linked is related to this code, and not another issue in the supabase module?

PR #23466, opened 2023-09-29
fix(nuxt): tree shake all occurrences of `<DevOnly>`

maritaria:
### 🔗 Linked issue

Various issues exist that may or may not be related to this bugfix.
The main problem solved is that the first occurence of `<DevOnly>` is not properly stripped from module sources, likely leading to every component within not to be tree-shaken. At runtime the behaviour is not noticable because of the fallback feature / rendering using `import.meta.dev`.

There does exist an issue which seems related to it: https://github.com/nuxt/nuxt/issues/21012.

### ❓ Type of change

### 📚 Description

Because RegExp is stateful, the `.test` call to bail early causes the `lastIndex` to be set after the first occurence of the DevOnly. This causes the following matchAll to exclude the first match of DevOnly tags.
Setting `lastIndex` to zero resets the state of the RegExp.

I came about this issue while trying to clone the devonly implementation for my own (a variation controllable with an env var, so that we can have some extra tools on an internal version of the system. 
This while production will not expose the tools.

As I was debugging my variation on the compiler plugin I noticed that the code sometimes still.

### 📝 Checklist


danielroe (commented):
Nice spot. I think we should just use a non-global regexp here though, as we do elsewhere:

https://github.com/nuxt/nuxt/blob/c37ebab1ef5eb163e88f551416d664efb9077c20/packages/nuxt/src/core/plugins/layer-aliasing.ts#L16-L17

maritaria:
> Nice spot. I think we should just use a non-global regexp here though, as we do elsewhere:

Ah yes, then you avoid the whole stateful ness alltogether.

huang-julien:
Seems to be ok for me locally through a unit test. Do you want me to merge it within this PR ? 
I'll take a look at what's going on with the fixtures tests

danielroe:
Thank you ❤️

huang-julien:
strange... can't reproduce the failing tests locally 🥲 ...

danielroe:
@huang-julien To reproduce, run `pnpm nuxi build test/fixtures/basic && node test/fixtures/basic/.output/server/index.mjs`. Then open `http://localhost:3000`. You should see there's been a hydration error and the content of the page (from the `DevOnly` component onward) is reproduced twice.

packages/nuxt/src/core/plugins/dev-only.ts
```diff
@@ -9,9 +9,10 @@ interface DevOnlyPluginOptions {
   sourcemap?: boolean
 }
 
-export const DevOnlyPlugin = createUnplugin((options: DevOnlyPluginOptions, meta) => {
-  const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/g
+const DEVONLY_COMP_SINGLE_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/
+const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/g
```
GalacticHypernova, line 13:
```ts
const DEVONLY_COMP_SINGLE_RE = /<(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>/
const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>/g
```
Ideally it should also catch the lazy variation of the dev only component.

packages/nuxt/test/devonly.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    <div>
+      <DevOnly>
+        <SomeDevOnlyComponent></SomeDevOnlyComponent>   
+      </DevOnly>
+    </div>
+    <SomeComponent>
+        <dev-only>
+          test
+        </dev-only>
+    </SomeComponent>
```
GalacticHypernova, line 24:
```suggestion
    <div>
      <LazyDevOnly>
         <SomeDevOnlyComponent></SomeDevOnlyComponent>   
      </LazyDevOnly>
      <DevOnly>
        <SomeDevOnlyComponent></SomeDevOnlyComponent>   
      </DevOnly>
    </div>
    <SomeComponent>
        <dev-only>
          test
        </dev-only>
        <lazy-dev-only>
          lazy test
        </lazy-dev-only>
    </SomeComponent>
```
This adds tests to the lazy variation as well.

packages/nuxt/test/devonly.test.ts
```diff
@@ -0,0 +1,41 @@
...
+          </div>
+          <SomeComponent>
+              
+          </SomeComponent>
+          </template>"
+    `)
+
+    expect(result).not.toContain('dev-only')
+    expect(result).not.toContain('DevOnly')
+  })
```
GalacticHypernova, line 40:
```suggestion
   expect(result).not.toContain('lazy-dev-only')
   expect(result).not.toContain('LazyDevOnly')
 })
```
This makes sure lazy variation isn't found either.

GalacticHypernova (commented):
Added lazy devonly component detection and testing.

packages/nuxt/src/core/plugins/dev-only.ts
danielroe, line 13:
```suggestion
const DEVONLY_COMP_SINGLE_RE = /<(?:dev-only|DevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly)>/
const DEVONLY_COMP_RE = /<(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>[\s\S]*?<\/(?:dev-only|DevOnly|lazy-dev-only|LazyDevOnly)>/g
```

danielroe (approved):

PR #23480, opened 2023-09-30
docs: setting middleware at build time

justinkekeocha:
Instead of using 'definePageMeta' on each page, it may be handy to just on certain routes that should have a certain middle ware

### ❓ Type of change

### 📝 Checklist


danielroe (commented):
I'm not sure I would recommend this as it will only run on the first load and not on subsequent navigations.

You could however add the named middleware in `~/app/router.options.ts` or even at build time in `pages:extend` hook, which would probably be the most performant option.

justinkekeocha:
I tried it and you are correct. could you please show how to do this from the pages:extend hook

justinkekeocha:
Thank you for your help Dan. While I was waiting for your answer yesterday, I went experimenting and discovered that i could achieve the same thing by creating a global middleware and calling other middleware like this:

`import auth from "~/middleware/auth";
import admin from "~/middleware/admin";

  const currentRoute = to.fullPath;

  if (currentRoute.includes("/dashboard")) {
    //The check for admin group should come first, else /dashboard will run and return
    //Then routes like /dashboard/admin will pass
    if (currentRoute.includes("/admin")) {
      return admin(to, from);
    }
    return auth(to, from);
  }
});
`

What do you think about this?

danielroe (approved):

PR #23483, opened 2023-09-30
fix(nuxt): ignore prefix if `clearNuxtState` called w/o keys

### LINKED ISSUE #23482: clearNuxtState composable does not work with function callback or clear all (no arguments)
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.5.1`
- Nuxt Version:     `3.7.4`
- CLI Version:      `3.9.0`
- Nitro Version:    `2.6.3`
- Package Manager:  `pnpm@8.7.1`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

These tests fail:
```ts
  it('clears state in payload for function', () => {
    const key = 'clearNuxtState-test'
    const state = useState(key, () => 'test')
    expect(state.value).toBe('test')
    clearNuxtState(() => false)
    expect(state.value).toBe('test')
    clearNuxtState(k => k === key)
    expect(state.value).toBeUndefined()
  })

  it('clears all state when no key is provided', () => {
    const state1 = useState('clearNuxtState-test', () => 'test')
    const state2 = useState('clearNuxtState-test2', () => 'test')
    expect(state1.value).toBe('test')
    expect(state2.value).toBe('test')
    clearNuxtState()
    expect(state1.value).toBeUndefined()
    expect(state2.value).toBeUndefined()
  })
```

### Describe the bug

`clearNuxtState` has a documented feature that lets you reset all state by providing no arguments, or filter on specific state using a function callback. Both are not working, because the `useStatePrefix` is not stripped from the existing keys when matching.

### Additional context

_No response_

### Logs

_No response_

AlbertBrand:
### 🔗 Linked issue

resolves #23482.

### ❓ Type of change

### 📚 Description

This fixes `clearNuxtState` behavior when called with a function or with no arguments. Internally, `useState` prefixes all keys. When `clearNuxtState` is called with a callback, it should not be needed to provide the prefix on keys from the calling side (this is an implementation detail). Also, when called without argument, all keys should be reset. The current behavior resets none, because all prefixed keys are prefixed again and then removed (which matches none).

I added mulitple tests as a separate `describe` to make sure all functionality is tested.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -42,6 +42,8 @@ export function clearNuxtState (
 ): void {
   const nuxtApp = useNuxtApp()
   const _allKeys = Object.keys(nuxtApp.payload.state)
+    .map(key => key.substring(useStateKeyPrefix.length))
+
   const _keys: string[] = !keys
     ? _allKeys
     : typeof keys === 'function'
--- a/test/nuxt/composables.test.ts
+++ b/test/nuxt/composables.test.ts
@@ -214,12 +214,50 @@ describe('useState', () => {
     useState('key', () => 'value')
     expect(Object.entries(useNuxtApp().payload.state)).toContainEqual(['$skey', 'value'])
   })
+})
 
-  it.todo('clearNuxtState', () => {
-    const state = useState(() => 'test')
+describe('clearNuxtState', () => {
+  it('clears state in payload for single key', () => {
+    const key = 'clearNuxtState-test'
+    const state = useState(key, () => 'test')
     expect(state.value).toBe('test')
+    clearNuxtState(key)
+    expect(state.value).toBeUndefined()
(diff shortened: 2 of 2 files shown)


danielroe (approved):
Thank you ❤️

packages/nuxt/src/app/composables/state.ts
```diff
@@ -42,6 +42,8 @@ export function clearNuxtState (
 ): void {
   const nuxtApp = useNuxtApp()
   const _allKeys = Object.keys(nuxtApp.payload.state)
+    .map(key => key.substring(useStateKeyPrefix.length))
```
danielroe, line 45: I would also welcome a future PR that improves perf by not enumerating state keys when an array of strings is provided as the keys arg.

PR #23512, opened 2023-10-03
fix(nuxt): use `test`/`dev` as manifest buildId when appropriate

danielroe:
### ❓ Type of change

### 📚 Description

With new app manifest enabled, we can potentially have a source of unreliable test data in that we have a constantly changing buildId which is difficult to mock (see https://github.com/danielroe/nuxt-vitest/pull/354).

This PR defaults the buildId (unless user overrides) to `test` when running in e2e/unit test environments.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -213,7 +213,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Add app manifest handler and prerender configuration
   if (nuxt.options.experimental.appManifest) {
     // @ts-expect-error untyped nuxt property
-    const buildId = nuxt.options.appConfig.nuxt!.buildId ||= randomUUID()
+    const buildId = nuxt.options.appConfig.nuxt!.buildId ||= (nuxt.options.test ? 'test' : randomUUID())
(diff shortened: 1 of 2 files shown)


pi0 (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -213,7 +213,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Add app manifest handler and prerender configuration
   if (nuxt.options.experimental.appManifest) {
     // @ts-expect-error untyped nuxt property
-    const buildId = nuxt.options.appConfig.nuxt!.buildId ||= randomUUID()
+    const buildId = nuxt.options.appConfig.nuxt!.buildId ||= (nuxt.options.test ? 'test' : randomUUID())
```
pi0, line 216: Haven't checked impl but is `dev` also considered?

danielroe: Do you mean in dev (non-test) we should default to `dev` or something similar?

pi0: Yes (either this field is optional or if we can serve it dev, i think a predictable value could be nice).

PR #23531, opened 2023-10-04
fix(types): don't pin webpack types to exact version

rchl:
### ❓ Type of change

### 📚 Description

Since the current convention in `@nuxt/types` is to pin all dependencies to exact versions, I'm pretty sure this is not an acceptable solution but I wanted to raise an issue with the current approach.

The `@types/webpack` is pinned to `4.41.33`. Since `4.41.34` is now the latest version, if I do `yarn upgrade` in my project, this will introduce both `4.41.33` and `4.41.34` and then the shim that extends `webpack` from `@nuxt/types/shims.d.ts` will stop working. A code like:

```ts
import '@nuxt/types'
import { MultiStats, AssetInfo } from 'webpack'
```

should show that `MultiStats` that the shim is supposed to introduce is not present.

It works if only `4.41.33` is installed in the project.

I'm not sure what to do with this problem. Clearly there was a reason for going with exact versions but as evident here, this is also not a good solution. Ideally a range should be allowed and as long as packages follow semver, it should not cause issues (but clearly did in the past).

@danielroe 

### 📝 Checklist

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -23,7 +23,7 @@
     "@types/pug": "2.0.6",
     "@types/serve-static": "1.15.2",
     "@types/terser-webpack-plugin": "4.2.1",
-    "@types/webpack": "4.41.33",
+    "@types/webpack": "^4.41.33",
     "@types/webpack-bundle-analyzer": "3.9.5",
     "@types/webpack-hot-middleware": "2.25.5"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -2834,10 +2834,10 @@
     "@types/source-list-map" "*"
     source-map "^0.7.3"
 
-"@types/webpack@4.41.33", "@types/webpack@^4", "@types/webpack@^4.41.8":
-  version "4.41.33"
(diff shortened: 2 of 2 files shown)


rchl:
CI here also started failing because of that issue.

packages/types/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/pug": "2.0.6",
     "@types/serve-static": "1.15.2",
     "@types/terser-webpack-plugin": "4.2.1",
-    "@types/webpack": "4.41.33",
+    "@types/webpack": "^4.41.33",
```
danielroe, line 26: Maybe?

```suggestion
    "@types/webpack": "~4.41.33",
```

rchl: Should be fine for now if you prefer that but could cause a similar issue in the future.

For reference, there seemed to be an issue with `@types/webpack-hot-middleware` before that broke types in a patch release (#18827) but I'm not aware of there being any issues with `@types/webpack`.

rchl: @danielroe so would you prefer to use `~`? I'm a bit skeptical because then it's guaranteed to break once webpack types update the minor version. I'd rather trust and hope and they don't break types in minor revision.

danielroe (approved):

PR #23546, opened 2023-10-05
docs: add the Nuxt context section for normalized context-only composable wording

### LINKED ISSUE #23542: useCookie documentation is self contradictory
### Describe the feature

[The very first sentence in useCookie docs](https://nuxt.com/docs/api/composables/use-cookie#:~:text=Within%20your%20pages%2C%20components%20and%20plugins%20you%20can%20use%20useCookie) says:

> Within your pages, components and plugins you can use useCookie, [...]

Then, a little below, there is [a box](https://nuxt.com/docs/api/composables/use-cookie#:~:text=useCookie%20only%20works%20during%20setup%20or%20Lifecycle%20Hooks.) saying:

> 👉 useCookie only works during setup or Lifecycle Hooks.

Because of that, its not clear if `useCookie` can be used in plugins or not.

### Additional information

### Final checks

GalacticHypernova:
### 🔗 Linked issue
fix #23542

### ❓ Type of change

### 📚 Description

Adds a standardized wording and dedicated section to the nuxt context from which certain composables need to be used.
### 📝 Checklist


Atinux:
Wondering if this alert is very useful and not instead misleading, we should mention where we could not mention the composable instead IMO

GalacticHypernova:
I mean, we could remove this alert entirely as we are basically already stating where it can and cannot be used in the first line of its description..

GalacticHypernova:
What do you think would be a more fitting approach?

danielroe (approved):

danielroe:
Ah, just catching up on conversation. I think we could have a standard wording or component that we can include on all components that require access to Nuxt context as it's a common issue - and linking to a common section where this is explained.

docs/3.api/1.composables/use-cookie.md
```diff
@@ -11,7 +11,7 @@ const cookie = useCookie(name, options)
 ```
 
 ::alert{icon=👉}
-`useCookie` only works during `setup` or `Lifecycle Hooks`.
+`useCookie` only works during `setup`, plugins, and `Lifecycle Hooks`.
```
huang-julien, line 14:
```suggestion
`useCookie` only works during `setup`, `plugins`, and `Lifecycle Hooks`.
```

should we remove \` on "`Lifecycle Hooks`" or add it to "plugins" ?

huang-julien (approved):

docs/3.api/1.composables/use-cookie.md
danielroe: the other issue here with 'lifecycle hooks' is that really there are some (vue) caveats here:

using these kind of composables after an await or outside certain lifecycle hooks can also cause memory leaks. (e.g. `watch` can't be auto-cleaned up).

GalacticHypernova: Maybe we can add a warning to life cycle hooks that it's generally a less optimal approach and that it's better to use in plugins and setup?

GalacticHypernova:
> Ah, just catching up on conversation. I think we could have a standard wording or component that we can include on all components that require access to Nuxt context as it's a common issue - and linking to a common section where this is explained.

That's a good idea. Should I open a new PR for that?

danielroe:
@GalacticHypernova Yes, that's a good idea.

GalacticHypernova:
@danielroe of course!
I'll rewrite this portion of the docs and update the title accordingly.

GalacticHypernova:
@danielroe I have updated the docs, I decided to add a section in /docs/2.guide/3.going-further/6.nuxt-app.md that explains about the Nuxt context and updated all the links accordingly. I think it is now ready for review.

docs/3.api/1.composables/use-async-data.md
```diff
@@ -6,7 +6,7 @@ description: useAsyncData provides access to data that resolves asynchronously.
 Within your pages, components, and plugins you can use useAsyncData to get access to data that resolves asynchronously.
 
 ::alert{type=warning}
-[`useAsyncData`](/docs/api/composables/use-async-data) is a composable meant to be called directly in a setup function, plugin, or route middleware. It returns reactive composables and handles adding responses to the Nuxt payload so they can be passed from server to client without re-fetching the data on client side when the page hydrates.
+[`useAsyncData`](/docs/api/composables/use-async-data) is a composable meant to be called directly in a [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context) scope. It returns reactive composables and handles adding responses to the Nuxt payload so they can be passed from server to client without re-fetching the data on client side when the page hydrates.
```
warflash, line 9: I would suggest `xxx is a composable meant to be called directly within the Nuxt context` eliminating the need for the term "scope" since it's not present in the explanation of `## The Nuxt context`. 
"a Nuxt context" implies the existance of multiple contexts so using "the" instead would make things even more clear I feel like.

GalacticHypernova: Hmm, it's possible, the question is won't it be too "restricting" if we change wording to `the` considering there are multiple available scopes (like plug-ins, middleware, setup functions, etc), and won't it be better to add the scope word in the nuxt context section (like `the possible Nuxt context scopes are ...`)?
Would love to hear your opinion about it!

warflash: I guess this comes down to technicalities.

From my understanding and mental model there is only one nuxt context at a time hence `the`.

Scopes with how the word is used currently to me implies that the nuxt context is scoped to different parts of the app and so differs between them. 
I think it'd make sense to be careful with the wording so that the programmers "scope" and physical location "scope" don't get mixed here and we don't introduce another technical term to the already existing "context" and "instance" that users need to understand

GalacticHypernova: I see what you mean. Do you think we could perhaps combine the 2 and say something like that the nuxt context is the same context, but is accessible from multiple parts of the app? We think of a better phrasing for that, because while it is the same context, it's noteworthy that you can reach it from multiple areas.

warflash: I think the introduction you've added "The nuxt context" covers that pretty well already.
"Currently, the Nuxt context exists only in plugins, Nuxt hooks, Nuxt middleware, and setup functions (in pages and components)."
To align it with your suggestion it could perhaps be 
"Currently, the Nuxt context is only accessible in plugins, Nuxt hooks, Nuxt middleware and setup functions (in pages and components)."

GalacticHypernova: Good idea! I'll add that, thanks!

GalacticHypernova: I have also rephrases it accordingly so users will know what "scopes" mean in the given context (no pun intended). That was I feel we could make it clear that while it's a shared context, the scopes refer to the areas in your app that can access it (so scopes in physical definition, rather than the functional definition). Let me know what you think of it!

warflash: Nice 👍
Personally I feel like I'd just avoid the introduction of "nuxt context scope" alltogether. That way there's less things a user needs to read and understand and less confusion about terms. Especially given that vue also uses scopes (explictly) for a similar ish concept. https://vuejs.org/api/reactivity-advanced.html#effectscope. 
I just think that throwing another term into an already good explanation has little upside but introduces more potential for confusion.
But that's super nitpicky and it's only one sentence so both ways work fine of course :)

GalacticHypernova: How would it make it clear where it's usable then? Saying it's only usable in the nuxt context?

warflash: Well currently that's explained already, right?
> These include plugins, Nuxt hooks, Nuxt middleware, and setup functions (in pages and components).

So I feel like

"The Nuxt context is only accessible in certain parts of your application, specifically from plugins, Nuxt hooks, middleware, and setup functions in pages and components."

stresses context is not accessible in all places, lists the areas it is accessible from and doesn't introduce any new words that would require second thought. 
Perhaps the listing of places with context could also link out to the documentation so plugins links to /plugins etc.

GalacticHypernova: Done! Added all the links, if there's any more feedback I'd love to hear it!

warflash: Nice 👍
Only thing I noticed is that useRouter https://github.com/nuxt/nuxt/pull/23546/files#diff-8637761ca22caaa4a8a782eef64f7da236d0b7c412ec64f308f604f6cbed4263R8 still uses the term scope, but LGTM

GalacticHypernova: Oh oops!

GalacticHypernova: Fixed it! Thanks for catching it!

GalacticHypernova:
Oh no, conflicts are too complex. This might delay the resolve process a bit as I don't have access to the client version currently.

docs/3.api/3.utils/set-response-status.md
```diff
@@ -12,6 +12,8 @@ Nuxt provides composables and utilities for first-class server-side-rendering su
 
 `setResponseStatus` sets the statusCode (and optionally the statusMessage) of the response.
 
+`setResponseStatus` can only be called in the [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context).
+
```
danielroe, line 16:
```suggestion
::callout
`setResponseStatus` can only be called in the [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context).
::

```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -10,6 +10,8 @@ links:
 
 When prerendering, you can hint to Nitro to prerender additional paths, even if their URLs do not show up in the HTML of the generated page.
 
+`prerenderRoutes` can only be called within the [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context).
```
danielroe, line 13:
```suggestion
::callout
`prerenderRoutes` can only be called within the [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context).
::
```

docs/2.guide/3.going-further/6.nuxt-app.md
```diff
@@ -19,6 +19,10 @@ In Nuxt 2, this was referred to as **Nuxt context**.
 Jump over the `NuxtApp` interface documentation.
 ::
 
+## The Nuxt context
+
+Many composables and utilities, both built-in and user-made, may require access to the Nuxt instance. You may get an error stating that "A composable that requires access to the Nuxt instance was called outside of a plugin, Nuxt hook, Nuxt middleware, or Vue setup function." when trying to use a certain composable (for example, [useState](/docs/api/composables/use-state)). That is because that specific composable needs to access the Nuxt context, which doesn't exist everywhere on your application. Currently, the Nuxt context is only accessible in [plugins](/docs/guide/directory-structure/plugins), [Nuxt hooks](/docs/guide/going-further/hooks), [Nuxt middleware](/docs/guide/directory-structure/middleware), and [setup functions](https://vuejs.org/api/composition-api-setup.html) (in pages and components).
```
danielroe, line 24:
```suggestion
Many composables and utilities, both built-in and user-made, may require access to the Nuxt instance.

You may get an error stating that 'A composable that requires access to the Nuxt instance was called outside of a plugin, Nuxt hook, Nuxt middleware, or Vue setup function.' when trying to use a composable (for example, [useState](/docs/api/composables/use-state)). That is because that specific composable needs to access the Nuxt context, which doesn't exist everywhere on your application.

Currently, the Nuxt context is only accessible in [plugins](/docs/guide/directory-structure/plugins), [Nuxt hooks](/docs/guide/going-further/hooks), [Nuxt middleware](/docs/guide/directory-structure/middleware), and [setup functions](https://vuejs.org/api/composition-api-setup.html) (in pages and components). You can also explicitly call functions within this context by using [`nuxtApp.runWithContext`](/docs/api/composables/use-nuxt-app#runwithcontext).
```

docs/2.guide/3.going-further/6.nuxt-app.md
```diff
@@ -19,6 +19,14 @@ In Nuxt 2, this was referred to as **Nuxt context**.
 Jump over the `NuxtApp` interface documentation.
 ::
 
+## The Nuxt context
```
danielroe, line 22:
```suggestion
## The Nuxt Context
```

danielroe (approved):
Nicely done - thank you ❤️

GalacticHypernova:
Glad to help!

PR #23549, opened 2023-10-06
fix(nuxt): unset `useCookie` value when it expires

### LINKED ISSUE #21709: `useCookie` value still remains after cookie expiration
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.5.3`
- Nitro Version:    `2.4.1`
- Package Manager:  `pnpm@8.6.0`
- Builder:          `vite`
- User Config:      `ssr`, `plugins`, `css`, `vite`, `build`, `devtools`, `modules`, `routeRules`, `runtimeConfig`, `elementPlus`, `i18n`, `auth`, `pinia`, `veeValidate`
- Runtime Modules:  `@vueuse/nuxt@10.1.2`, `@pinia/nuxt@0.4.11`, `./modules/auth`, `@nuxtjs/i18n-edge@8.0.0-beta.12-28113576.7bc33c3`, `@element-plus/nuxt@1.0.5`, `@unocss/nuxt@0.52.7`, `nuxt-lodash@2.4.1`, `@vee-validate/nuxt@4.9.6`
- Build Modules:    `-`

### Reproduction

```ts
const testCookie = useCookie('test', { maxAge: 10 });
testCookie.value = 'some-value';

onMounted(() => {
  setTimeout(() => {
    console.log('testCookie', testCookie.value);
  }, 13000);
});
```

https://stackblitz.com/edit/github-mkuone?file=app.vue

### Describe the bug

* Actual result: `'some-value'` on console.
* Expected result: `undefined`.

Since the cookie has expired, accessing the cookie value should return `undefined` instead of the original value.

### Additional context

_No response_

### Logs

_No response_

jaegar-wolf:
### 🔗 Linked issue

resolves #21709

### ❓ Type of change

### 📚 Description

(**FIRST PR EVER**)

I've changed the cookie ref to be a custom Ref. If someone passes expires or maxAge option the custom Ref will make a setTimeout that will update the cookie value to undefined when it will expire, if none of those options were passed no timeout will be set on the ref.

Resolves #21709

### 📝 Checklist


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -114,3 +123,27 @@ function writeServerCookie (event: H3Event, name: string, value: any, opts: Cook
...
+          timeout = setTimeout(() => {
+            value = undefined
+            trigger()
+          }, delay)
+        }
+        trigger()
+      }
+    }
+  })
+}
```
huang-julien, line 149: I think there's a risk of having a memory leak here server-side. This should be probably client only i think?

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -31,7 +31,16 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
...
+  let delay: number | undefined
+  if (opts.maxAge) {
+    delay = opts.maxAge * 1000 // convert to ms for setTimeout
+  } else if (opts.expires) {
+    delay = opts.expires.getTime() - new Date().getTime() // getTime() already return time in ms
+  }
+  const cookie = useCustomCookieRef<T | undefined>(
+    (cookies[name] as any) ?? opts.default?.(),
+    delay
+  )
```
huang-julien, line 43:
```suggestion
  const cookie = import.meta.server ? ref<T | undefined>(cookies[name] as any ?? opts.default?.()) : useCustomCookieRef<T | undefined>(
    (cookies[name] as any) ?? opts.default?.(),
    delay
  )
```

packages/nuxt/src/app/composables/cookie.ts
danielroe: Also, I think on client side we might also want to clear the timeout when the component containing `useCookie` is unmounted...

danielroe:
This is very nice - thank you! Could we add a test in `test/nuxt/composables.test.ts` to cover the behaviour here, perhaps? 🙏

packages/nuxt/src/app/composables/cookie.ts
jaegar-wolf: Thanks for your advices, I will work on it

danielroe:
marking as draft just to help my own triage process - please do unmark as draft when it's ready to look at again.

looking forward to it ❤️

jaegar-wolf:
Sorry, I didn't had the time to work on it the last week.

Now if you unmount the component that use "useCookie" the timeout will be cleared.
But I don't really understand how to only trigger the "server side" part of the code, so I don't know if it was necessary to clear the timeout on the server side.

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -27,11 +27,25 @@ const CookieDefaults = {
...
   const cookies = readRawCookies(opts) || {}
 
-  const cookie = ref<T | undefined>(cookies[name] as any ?? opts.default?.())
+  let delay: number | undefined
+  if (opts.maxAge) {
+    delay = opts.maxAge * 1000 // convert to ms for setTimeout
+  } else if (opts.expires) {
+    delay = opts.expires.getTime() - new Date().getTime() // getTime() already return time in ms
+  }
+  // use customRef if on client side overwize use basic ref
```
manniL, line 42:
```suggestion
  // use customRef if on client side otherwise use basic ref
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -27,11 +27,25 @@ const CookieDefaults = {
   encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val))
 } satisfies CookieOptions<any>
 
+let timeoutIDToClear: NodeJS.Timeout
```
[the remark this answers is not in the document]
jaegar-wolf: I didn't found another way to clear the timeout when **useCookie** is unmounted, because when you set the timeout inside customRef you can't have acces to his ID, I've tried to return the value of the cookie and the timeout from customRef but that was a bad idea.

I think creating a registry object is a good idea.
But in which file should I declare this variable ?

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -31,7 +31,16 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
   const opts = { ...CookieDefaults, ..._opts }
   const cookies = readRawCookies(opts) || {}
 
-  const cookie = ref<T | undefined>(cookies[name] as any ?? opts.default?.())
+  let delay: number | undefined
+  if (opts.maxAge) {
+    delay = opts.maxAge * 1000 // convert to ms for setTimeout
+  } else if (opts.expires) {
+    delay = opts.expires.getTime() - new Date().getTime() // getTime() already return time in ms
```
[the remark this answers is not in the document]
jaegar-wolf: I didn't know about this one, thank you

danielroe (approved):
This is a nice PR and great initiative - thank you! ❤️ 

I pushed a couple of changes - let me know if they look good to you 🙏

jaegar-wolf:
For me it's perfect, I just learn about onScopeDispose thanks to you ❤️

PR #23600, opened 2023-10-10
feat(nuxt): support `deep: false` for data composables

### LINKED ISSUE #12406: AsyncData with a shallow ref

stafyniaksacha:
resolves https://github.com/nuxt/nuxt/issues/12406

### ❓ Type of change

### 📚 Description

This adds new option to control output of `useAsyncData` and `useFetch` data in order to have `ShallowRef` instead of `Ref`

It also introduces a new experimental flag `asyncDataShallow` to change the default behavior to use `shallow` by default

### 📝 Checklist


danielroe:
What do you think about doing something like `{ default: shallowRef }`?

stafyniaksacha:
> What do you think about doing something like `{ default: shallowRef }`?

I think it's less clear about our intentions, default should be keep for default data, here it's more like `ref(shallowRef(...))`

packages/nuxt/src/core/nuxt.ts
```diff
@@ -329,6 +329,11 @@ async function initNuxt (nuxt: Nuxt) {
     addPlugin(resolve(nuxt.options.appDir, 'plugins/view-transitions.client'))
   }
 
+  // Add experimental automatic view transition api support
```
huang-julien, line 332:
```suggestion
```

docs/3.api/1.composables/use-state.md
```diff
@@ -24,3 +24,12 @@ Because the data inside [`useState`](/docs/api/composables/use-state) will be se
 
 ::ReadMore{link="/docs/getting-started/state-management"}
 ::
+
+## Using `shallowRef`
+
+If you don't need your state to be deeply reactive, you can combine `useState` with [`shallowRef`](https://vuejs.org/api/reactivity-advanced.html#shallowref). This is usefull for performance optimizations of large data structures
```
danielroe, line 30:
```suggestion
If you don't need your state to be deeply reactive, you can combine `useState` with [`shallowRef`](https://vuejs.org/api/reactivity-advanced.html#shallowref). This will improve performance when your state contains large objects and arrays.
```

docs/3.api/1.composables/use-async-data.md
```diff
@@ -60,6 +61,7 @@ type AsyncDataRequestStatus = 'idle' | 'pending' | 'success' | 'error'
   * _transform_: a function that can be used to alter `handler` function result after resolving
   * _pick_: only pick specified keys in this array from the `handler` function result
   * _watch_: watch reactive sources to auto-refresh
+  * _shallow_: return data in a shallow ref, instead of a deep ref object. This is useful when you want to fine-tune reactivity.
```
danielroe, line 64:
```suggestion
  * _shallow_: return data in a shallow ref, instead of a deep ref object. This can improve performance when your data does not need to be deeply reactive.
```

docs/3.api/1.composables/use-fetch.md
```diff
@@ -72,6 +73,7 @@ All fetch options can be given a `computed` or `ref` value. These will be watche
   * `transform`: a function that can be used to alter `handler` function result after resolving
   * `pick`: only pick specified keys in this array from the `handler` function result
   * `watch`: watch an array of reactive sources and auto-refresh the fetch result when they change. Fetch options and URL are watched by default. You can completely ignore reactive sources by using `watch: false`. Together with `immediate: false`, this allows for a fully-manual `useFetch`.
+  * `shallow`: return data in a shallow ref, instead of a deep ref object. This is useful when you want to fine-tune reactivity.
```
danielroe, line 76:
```suggestion
  * `shallow`: return data in a shallow ref, instead of a deep ref object. This can improve performance when your data does not need to be deeply reactive.
```

packages/nuxt/src/app/plugins/async-data-shallow.ts
```diff
@@ -0,0 +1,9 @@
+import { defineNuxtPlugin } from '#app/nuxt'
+
+export default defineNuxtPlugin({
+  name: 'nuxt:async-data-shallow',
+  enforce: 'pre',
+  setup (nuxtApp) {
+    nuxtApp._asyncDataShallow = true
```
danielroe, line 7: can we use `nuxt.config.mjs` instead?

https://github.com/nuxt/nuxt/blob/a2fb8234ac629c67776f38b691fab273838ef48e/packages/nuxt/src/core/templates.ts#L338-L353

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -149,7 +151,9 @@ export function useAsyncData<
     nuxt.payload._errors[key] ??= null
 
     nuxt._asyncData[key] = {
-      data: ref(getCachedData() ?? options.default!()),
+      data: options.shallow
+        ? shallowRef(getCachedData() ?? options.default!())
+        : ref(getCachedData() ?? options.default!()),
       pending: ref(!hasCachedData()),
       error: toRef(nuxt.payload._errors, key),
```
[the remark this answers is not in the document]
danielroe: we could create a local util until/if it's available in core

packages/nuxt/src/app/plugins/async-data-shallow.ts
danielroe: let's defer this to another PR

docs/3.api/1.composables/use-async-data.md
```diff
@@ -26,7 +26,7 @@ type AsyncDataOptions<DataT> = {
   server?: boolean
   lazy?: boolean
   immediate?: boolean
-  shallow?: boolean
+  deep?: boolean
```
huang-julien, line 29: Isn't it easier to understand with `shallow` or `shallowRef` instead of `deep` ?

danielroe: I was thinking:
1. it mirrors the `deep` option for `watch`
2. it will make more sense later on if we choose to default to false (so `{ deep: true }` would be required to turn on deep reactivity)

PR #23607, opened 2023-10-10
feat(nuxt): auto-install optional features on StackBlitz

antfu:
### ❓ Type of change

### 📚 Description

In StackBlitz we could assume the envoriment is isolated and safe, we could install optional deps when enabled automatically to provide better overall experience when opening playgrounds.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/features.ts
+++ b/packages/nuxt/src/core/features.ts
@@ -1,9 +1,17 @@
 import { addDependency } from 'nypm'
 import { resolvePackageJSON } from 'pkg-types'
 import { logger } from '@nuxt/kit'
-import { isCI } from 'std-env'
+import { isCI, provider } from 'std-env'
 
-export async function ensurePackageInstalled (rootDir: string, name: string, searchPaths?: string[]) {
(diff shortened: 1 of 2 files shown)


fvsch:
Is there a way to test this on StackBlitz?

I have some doubts about the compatibility with WebContainers, since the current prompts are failing when accepted manually.

packages/nuxt/src/core/features.ts
```diff
@@ -1,9 +1,17 @@
...
 
-export async function ensurePackageInstalled (rootDir: string, name: string, searchPaths?: string[]) {
+const isStackblitz = provider === 'stackblitz'
+
+export async function ensurePackageInstalled (
+  rootDir: string,
+  name: string,
+  searchPaths?: string[],
+  // In StackBlitz we install packages automatically by default
+  prompt = !isStackblitz
```
danielroe, line 13: what about making this an options object in case we need to add any more in future?

danielroe (approved):
nice 👌

danielroe (approved):

PR #23617, opened 2023-10-11
fix(nuxt): use import alias when checking if plugin is wrapped

Clarkkkk:
### ❓ Type of change

### 📚 Description
When a plugin use `defineNuxtPlugin` with an import alias like `import { defineNuxtPlugin as n } from "nuxt/app";`, the wraping warning `It is advised to wrap your plugins...` should not occur.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/plugins/plugin-metadata.ts
+++ b/packages/nuxt/src/core/plugins/plugin-metadata.ts
@@ -132,7 +132,7 @@ export const RemovePluginMetadataPlugin = (nuxt: Nuxt) => createUnplugin(() => {
       }
 
       let wrapped = false
-      let importName = ''
+      const wrapperNames = new Set(['defineNuxtPlugin', 'definePayloadPlugin'])
 
       try {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/plugins/plugin-metadata.ts
```diff
@@ -132,13 +132,17 @@ export const RemovePluginMetadataPlugin = (nuxt: Nuxt) => createUnplugin(() => {
       }
 
       let wrapped = false
+      let importName = ''
```
danielroe, line 135: would it make sense to use a Set in case user has an extraneous import - unlikely as it is? e.g. `import { defineNuxtPlugin, definePayloadPlugin as n } from '#app'`

Clarkkkk: Thank you for reviewing. I will use a Set to exclude such edge cases.

danielroe (approved):

PR #23696, opened 2023-10-16
fix(schema): use scule types for `runtimeConfig` type hints

### LINKED ISSUE #23246: RuntimeConfig does not display proper environment variable hint when already using UPPER_SNAKE_CASE
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.7.3
- CLI Version:      3.8.4
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools, runtimeConfig
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-uezcgs?file=nuxt.config.ts

### Describe the bug

Hint reads:
```
You can override this value at runtime with NUXT_PUBLIC_A_PP_NAME_S
```

Should be:
```
You can override this value at runtime with NUXT_PUBLIC_APP_NAMES
```

### Additional context

```ts
  runtimeConfig: {
    public: {
      APP_NAMES: 'xyz'
    }
  }
})
```

Environment variable `NUXT_PUBLIC_APP_NAMES` is properly respected. The issue concerns the documentation exclusively.

### Logs

_No response_

luc122c:
### 🔗 Linked issue

Resolves #23246

### ❓ Type of change

### 📚 Description

In the original code, if the first character was a digit, the code was adding an underscore before the character and then converting it to uppercase. This was causing the additional underscores when the input string was already in upper snake case.

[TypeScript Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogHsATgQwMbAKpkogwgCwFcA7AawB4AVKCBCYgEwGcomkBLYgcwD4oBeKNVrB6zKAAMAJAG9OAMwiIoAQQC+shUqgAhNRIBQUKAH5VNOoxZYcqZEwjkVPI8dO6Loq5Nk2ldh3I2RE5eDTliRWUAJQg2fVc3dx1PMRZpCKioXHDY+MMkpLMM9Vkc2XgkNExsJQISCjzgHgTCtwAuKGIIADclRI7VAc6AchHXTu6+xAMDUEgoAGUAG3ZUCAAZe2AqVO9g0L5BYUtxDK1lUsztPQLkvbPNSO1y6+UAEVa3M3eHljGBsYzCpAVBOiVwq8VmtNttyO8WndjKNxsiur1+nNwNA-IhFsRkKQILh7I4Tl5xAduAAaJbAZCiP5QEZsZCIYAjKAAH2ZywA9gB3JScnkjQi1RCcwQs+nskZHIRM87PS7hC66L7Apm4gKOZwDLUiNI+GQARm5UAATBaAMwWgAsFoArBaAGwWgDsFoAHBaAJwWgAMXyKJquuPxhOJpPIOlpI35QsliNBnUW9MZRu8CcFwotYolqLa9yzjxk0PWWzY5EqKHQuPqZFjPBaTyyTRDxbMFdh1dr1QbRCbOhbTIBxcKxVk4YlkaJJMCceZrLlKYnSXB0-CParO379YljYoI9bMgjBPnMaa8ZXHLX686KVL1gluuboLaxQA+lvZLcP4U4I-jIM44HO0aLvGibCvebRphm0DPsusocgBQJhuE55RgujhLjmSbyp2cHotMqZ0gyiGnP8t4iryuaSgBU5nq+MbOJhs4XhBuFQfRhFIoBkjATqrGnlhl6QXRBH3p0wmBJQLhYgsADy0QAJIAOKqQAcioGwCFAYlceQIwqAACqZX46QAsjAizygYAD0DlJAAeiYinQJQtnUIIhk4cZZmmfZTmue58yed51q+Rx2ExiMyDYMFzluG5HlCN5drRWBnH+SMXB8nyDDqSgPQgEloVpV5iyUI6WVKOBuVICAlD4OwTBKd0WkiOVKUmEAA)

This PR also adds a couple of tests to catch values already in upper snake case (as in the original issue).

### 📝 Checklist


danielroe (approved):
Thank you ❤️

danielroe (changes requested):
Would you check the type fixture failures? 🙏

danielroe:
I believe that is how the type assertion indicates there is a mismatch. This is what the mismatch looks like:

CleanShot 2023-10-16 at 22 55 37@2x

luc122c:
Oh, interesting. I wasn't seeing any errors in the console output so I didn't think the tests were failing. I'll take another look.

luc122c:
@danielroe I've changed the input for the last test that was failing. It was originally `baseAPIToken`, now `baseApiToken`, and I understand why this was struggling. Although we know that 'API' is a common acronym, TypeScript doesn't. It would be possible to add a case to check for this, however I think it may be more beneficial to correctly handle camelCase than to correctly handle this particular case? Please let me know your thoughts.

danielroe:
The intent is to split the string with `_` immediately before the final capital letter of a sequence of capital letters. So:

```
abcDEFghi => abc_DE_FGHI
abcDefGhi => ABC_DEF_GHI
```

luc122c:
> The intent is to split the string with `_` immediately before the final capital letter of a sequence of capital letters.

Ok, i've implemented [this change](https://www.typescriptlang.org/play?#code/PTAEBcE8AcFNQKrTgJwMoDsCGBrWBhLAZ1gB4AVUWAD3FgwBMjQjwUBLDAcwBpQ1wWOlVr0moAOSssKcBNAAfSQBsA9gHdYKeUokBXZFvkBeSdNkSAfKFOUadRswAGAEgDenAGZbQAQQC+7l4+AEL+TgBQIKAxAPx+Ig7iSKgAxsRkvpZRYDFxCfZizu4AjIqgAEzlAMzlACzlAKzlAGzlAOzlABzlAJzlAAzhOXmj8a5uAe4pWpi4BBmkIXwSapralsPRozEAXPyCwoWOKhpG5fqG2iM7efEhiUWgE2jK7KmwADLE4KQAorQUFhUuAZih8AALPQYHBLSybIIYbwoUAAJVgrC2uVud34bw+31Y-0BwNBV0h0NhIXhjxOEgkNxxY2eAH13AE2W4wXM8IQSEsVuY5JtItsmXl9k5OVM3K93l8fsS2KSwRSYXDNpzudheYt0axBYILCLGUz9g9jskrul+dTTeLxtLAm4wqLseLQJKndMrjyFraVmsjCaxbd9gIhPBLcwpEa5PbbuN2c7tfM+WRlqd1lYsR7PaAMLAAG5aBP5iNHUR0oU6LNGMv5CZgm2ZBFc306-0ZwNnDa5s2sn1pRZZFMdtOLTOrXs5t0SxDWxbkbIRCJQOCgAHKkGqqHquxV8SsDjcay2WniCbBFEy6+gV0RfK+C-MZsj7LMi2H4rt4f84+cFwbZ3vq4DDIm94vs8iLIqA+DOqB4E4kmkzOvB7hbkCO7knusKgSaA6FiWKCPjs+y+KReySAyVFEaWa4wPAcoEoqB5JMwAGnjYoBsU8V5Ij4t4CSiD75F+7HQR4wlwc6d4ACJIfkclQfSlG4hRYaSTK6GyviCpEnJBHzqptHFvRq7rvAqa6vyvEnJxvAHJGKk1hcQbaBcBioCYZhxlY3F2ZeMGCbJ0miaA8TPtGC5-q2amRVBExlEoVRKLUSgNEozRKG0SidEoPRKP0ShDJEEETDK1ldgKdZ9mVZFOZWEnTtmnlXDRyGQdFLx6YSvyYSqOGUhqoWwYh9VMvEzH6f1JLYagapUjS0UmXmKGVeONndr5MjCop4pesm7jTX1SpYWSC24SNWqbdVoGGrts5qZp4lPG+tofmtg6oe44UeodP2-rMnbpjVLXBvtozhocUbfjtFjPTs61jqgfqg1O7lPf9BZmSRmkVrDzWubomOI42Q5aC2pCjhT6Ag5OPbZkZmlSrTVM00DdMTgGtVY-m71kMuq4MRuADyqIAJIAOISwAcr4nzcVVoMSL4AAKassvLACyfxoFYpoAHqxCL8DkHrlCmMriyqxrBuhsbps8RbVRW7dKtYMg9vuo7lnO2g5C1G7qP0-yEhcKoqgMFLQJFpA3ujL7jH++QDTB8D3NkBIbCQOQELsEQouFrL9gJ3kScbubAfNOnXNbaQEgAEYZOrEvkKoeAYGXMSO0AA). 

`baseAPIToken` now becomes `BASE_API_TOKEN` however `APP_NAMES`, which is already correct, now becomes `A_PP_NAME_S` which was a case highlighted in the original issue.

luc122c:
Ok, so I think I've got to the bottom of this. The type works fine, except for when the input is already in UPPER_SNAKE_CASE. To get around this, I implemented this check:

```ts
type UpperUpperSnakeCase<T extends string> = T extends Uppercase<T> ? T : UpperSnakeCase<T>
```

Theoretcically, this could be combined into `UpperSnakeCase`, but when I do that, it breaks other tests 😭 [See here](https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAogHsATgQwMbAKpkogwgCwFcA7AawB4AVKCBCYgEwGcomkBLYgcwD4oBeKNVrB6zKAAMAJAG9OAMwiIoAQQC+shUqgAhNRIBQUKAH5VNOoxZYcqZEwjkVPI8dO6Loq5Nk2ldh3I2RE5eDTliRWUAJQg2fVc3dx1PMRZpCKioXHDY+MMkpLMM9Vkc2XgkNExsJQISCjzgHgTCtwAuKGIIADclRI7VAc6AchHXTu6+xAMDUEgoAGUAG3ZUCAAZe2AqVO9g0L5BYUtxDK1lUsztPQLkvbPNSO1y6+UAEVa3M3eHljGBsYzCpAVBOiVwq8VmtNttyO8WndjKNxsiur1+nNwNA-IhFsRkKQILh7I4Tl5xAduAAaJbAZCiP5QEZsZCIYAjKAAH2ZywA9gB3JScnkjQi1RCcwQs+nskZHKAGclpKC4gJkvhmaidJVM87PS7hC66L7Aplq0lOFxFcwiFUZACM3KgACZnQBmZ0AFmdAFZnQA2Z0AdmdAA5nQBOZ0ABi+Nohvgl+MJxMtOlpI35QsliNBnUW9MZdu8WcFwudYolqLa9xLjxk0PWWzY5EqKHQuPqZHIOh4LSeWSa8drZibsNb7eqXaIPb7fHr-xrte+PhkV1xKaJJMCGeZrLleZXSXBAH1ZFdxy2dlPOxLuxR52pzzJNwTt5ampmDxyj8fOiki6qhK6q9tax5ApIL5XLcoJtGeF7hG+qY7o4e5ljm8ojiedIMtAQEymyHJwauiavsm75prumbZsKf4rpMGIzPBuHFqc-w-iKvLlpKJHFEmtiWs4SEUSh6Y0TxWFIoUCHkYJgTCQJShblRaESZhf6dBagSUC4WILAA8tEACSADixkAHIqBsAjATgKmoeQIwqAACi5p5WQAsjAizygYAD0-lJAAeiY+nQJQPnUIIyEfoEzluX5gUhWF8wRVFboxaJcWOCMyDYElQVuKF4VCFFnpZfZlGOSMXB8nyDCmSgPQgIVKWlZFiyUD6lXKdVlojEgICUPg7BMAZ3QWSIbXFal2JlV1-q9Xi-XxQARqSrnGZQfJEsQM3GCVaULZQQbLQ5A0bQ4GDRBsB1QEd82dZQobnatuVgIQa2rKguB8pE7BcPdj0LM9EZvWJ8WiGwf0A0DLjJbNHVRdGEM5U5yDA3NoNRQ6Ma2bFqlOQ6LruljQA)

danielroe:
What about test_APP_NAME?

luc122c:
`test_APP_NAME` becomes `TEST__APP_NAME`. I presume we'd like it to be `TEST_APP_NAME`? I've added an additional test case.

luc122c:
Ok, I think it's done now. All the tests pass, including the new one. 👍

luc122c:
@danielroe The test failure doesn't seem to be related to this PR:
```sh
listen EADDRINUSE: address already in use 127.0.0.1:59823
```
Please can the tests be re-run?

luc122c:
@danielroe Some of these test cases may not be possible with the current logic. For example, `TESTAppName` I can't figure out a way for this to work, since TypeScript doesn't know that 'Test', 'App' and 'Name' are words, so it doesn't know to split 'TESTA' before the A.

danielroe:
The logic isn't related to whether they are words. Note that scule splits these all correctly; it's based on splitting before the change of case.

I think we can use scule's native types for this - but we should ideally expose them upstream.

luc122c:
> The logic isn't related to whether they are words. Note that scule splits these all correctly; it's based on splitting before the change of case.

[This is the closest I've managed to get](https://www.typescriptlang.org/play?ts=5.2.2#code/C4TwDgpgBAKg9gVTJATgZQHYEMDWEDCWAzhADwxQQAewEGAJkVEcCgJYYDmANFAAooIANzZwArk2q0GTFuy5QAvFADkKgHxLYlGnUZQABgBIA3hwBmEFFHxiUgjMAC+pi1agAlCCycGAUFBQAPyGpgAUAsKiEjrS+gAycADuVgDGxGSRIuJE6gGBwVBhtvZ0wLF6TEio6SSkJQ7AmiERgtkxUpWqAPoqhWpQAFytUTkVMqoqQWrDDWXj+iq90yqDSyoAlFtDkxs7ahsuJtVpGfV2jepH8Cfo2HiEdV4svHOOV-6Bg7AA3H5+oEgsG8wAAjFobsgrJhcAQzipaCwAILIAByWAAthAND8oH4APT4gpQAB6QQB4GgMBBACYIYgoXdYY8yCoYABRNAwbpIvh8bqopEAWXZOLxhOJZIpQOpLAAzPTbjCHvDEcAeXyBcLRepcQSiQUpYCqSCACyKxnKuF1BEg7pYZBakVi-WS8nG4EsACsFtQVpZpFtyLR2pdEsN7spnuAADZfdD7tbWWreXx0ViwwbAkao7LgAB2eNMlU2lOa9PY3XirOkyMykEADiL-tVdtTAsxlb14ezdZNLAAnM3EwG2ZyYCiwBXM26gA) separately, just following the logic.

> I think we can use scule's native types for this - but we should ideally expose them upstream.

Does Scule expose itself as a type, [like Defu does](https://github.com/unjs/defu?tab=readme-ov-file#type)?

danielroe:
No - see https://github.com/unjs/scule/pull/58. I pushed a change in [`c5f9a49` (#23696)](https://github.com/nuxt/nuxt/pull/23696/commits/c5f9a491ea6a622e77f3d2ab36ebe96df4d04604) that uses the inferred return type of `snakeCase`, however.

packages/schema/src/types/config.ts
```diff
@@ -4,50 +4,16 @@ import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 import type { Options as VueJsxPluginOptions } from '@vitejs/plugin-vue-jsx'
 import type { SchemaDefinition } from 'untyped'
 import type { NitroRuntimeConfig, NitroRuntimeConfigApp } from 'nitropack'
+// TODO: expose SnakeCase type from scule
```
danielroe, line 7:
```suggestion
```

test/fixtures/basic-types/types.ts
```diff
@@ -271,6 +271,23 @@ describe('runtimeConfig', () => {
     expectTypeOf(val.runtimeConfig!.public!.ids).toEqualTypeOf<undefined | RuntimeValue<Array<number>, 'You can override this value at runtime with NUXT_PUBLIC_IDS'>>()
     expectTypeOf(val.runtimeConfig!.unknown).toEqualTypeOf<unknown>()
   })
+
+  it('correctly converts different kinds of names to snake case', () => {
+    // https://github.com/unjs/nitro/blob/be25a21d79acf272b3629d69332d66795ae90b6e/src/runtime/config.ts#L56
```
danielroe, line 276:
```suggestion
```

test/fixtures/basic-types/types.ts
```diff
@@ -271,6 +271,23 @@ describe('runtimeConfig', () => {
...
+    // require('scule').snakeCase(key).toUpperCase()
```
danielroe, line 277:
```suggestion
```

danielroe (approved):

packages/schema/src/types/config.ts
luc122c: @danielroe Would https://github.com/unjs/scule/pull/58 need to be merged and released first?

danielroe: Yes, it will.

PR #23717, opened 2023-10-17
feat(nuxt): prompt to autoinstall `@nuxt/image` when it is used

danielroe:
### ❓ Type of change

### 📚 Description

This is a preliminary implementation of auto-installation for `@nuxt/image`.

We plan for @pi0 to improve and possibly refactor to nuxi (providing an endpoint for a more interactive/seamless experience that doesn't rely on the CLI).

[autoinstall.mp4](https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/xbfiaQHnuvLSgxHiCxpj/f30d0080-a1ea-482d-a385-4693084c1700/autoinstall.mp4)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-stubs.ts
+++ b/packages/nuxt/src/app/components/nuxt-stubs.ts
@@ -1,17 +1,17 @@
-import { createError } from '#imports'
+import { createError } from '../composables/error'
 
 function renderStubMessage (name: string) {
   throw createError({
     fatal: true,
     statusCode: 500,
-    statusMessage: `<${name}> is provided by @nuxt/image. Check your console to install it or run 'npx nuxi@latest module add @nuxt/image'`
+    statusMessage: `${name} is provided by @nuxt/image. Check your console to install it or run 'npx nuxi@latest module add @nuxt/image'`
   })
 }
 
 export const NuxtImg = {
-  setup: () => renderStubMessage('NuxtImg')
+  setup: () => renderStubMessage('<NuxtImg>')
 }
 
 export const NuxtPicture = {
-  setup: () => renderStubMessage('NuxtPicture')
+  setup: () => renderStubMessage('<NuxtPicture>')
 }
--- a/packages/nuxt/src/app/components/nuxt-stubs.ts
+++ b/packages/nuxt/src/app/components/nuxt-stubs.ts
@@ -1,0 +1,17 @@
+import { createError } from '#imports'
+
+function renderStubMessage (name: string) {
+  throw createError({
+    fatal: true,
+    statusCode: 500,
+    statusMessage: `<${name}> is provided by @nuxt/image. Check your console to install it or run 'npx nuxi@latest module add @nuxt/image'`
+  })
+}
+
+export const NuxtImg = {
+  setup: () => renderStubMessage('NuxtImg')
+}
+
+export const NuxtPicture = {
+  setup: () => renderStubMessage('NuxtPicture')
+}
--- a/packages/nuxt/src/core/features.ts
+++ b/packages/nuxt/src/core/features.ts
@@ -1,6 +1,6 @@
 import { addDependency } from 'nypm'
 import { resolvePackageJSON } from 'pkg-types'
-import { logger } from '@nuxt/kit'
+import { logger, useNuxt } from '@nuxt/kit'
 import { isCI, provider } from 'std-env'
 
 const isStackblitz = provider === 'stackblitz'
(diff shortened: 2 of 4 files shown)


danielroe:
Current dependencies on/for this PR:
* main
  * **PR #23717** <a href="https://app.graphite.dev/github/pr/nuxt/nuxt/23717" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/nuxt/nuxt/23717?utm_source=stack-comment).

atinux (approved):
Awesome!

Gruce:
Could this be for all official modules?

danielroe:
We're thinking of doing it for core Nuxt modules, if there's an appropriate and unique thing we can detect, such as a component or composable. (`useImage` isn't yet supported to auto-install nuxt/image but we plan to make it possible.)

pi0:
> Could this be for all official modules?

Feature Detection is planned to be part of module DB specification so that we can open up both for official and 3rd party known modules, exposing certain composable or component.

packages/nuxt/src/app/components/nuxt-stubs.ts
```diff
@@ -0,0 +1,17 @@
+import { createError } from '../composables/error'
+
+function renderStubMessage (name: string) {
+  throw createError({
+    fatal: true,
+    statusCode: 500,
```
pi0, line 6: Using fatal flag, can cause to have unwanted side-effects like showing verbose Nitro logs and decreasing DX. Also we could use something better than 500 for this like 404 even as we use for placeholder assets too

packages/nuxt/src/core/features.ts
```diff
@@ -39,14 +36,30 @@ export async function ensurePackageInstalled (
...
+// TODO: refactor to Nuxi
+const installPrompts = new Set<string>()
+export function installNuxtModule (name: string, options?: EnsurePackageInstalledOptions) {
+  if (installPrompts.has(name)) { return }
+  installPrompts.add(name)
+  const nuxt = useNuxt()
+  return promptToInstall(name, async () => {
+    const { runCommand } = await import('nuxi')
+    await runCommand('module', ['add', name, '--cwd', nuxt.options.rootDir])
+  }, { rootDir: nuxt.options.rootDir, searchPaths: nuxt.options.modulesDir, ...options })
```
pi0, line 57: We could ensure `import.meta.url` is in the search path. It refers to the `nuxt` package that has direct dependency on `nuxi`. In this case, it would be almost impossible that it won't work or be found...

Also, since rootDir change is happening, was `--cwd` flag also necessary.

danielroe: Yes, it was necessary; it was otherwise picking up the parent directory in a monorepo.

packages/nuxt/src/app/components/nuxt-stubs.ts
danielroe: It is necessary to make the error show on client-side. (as opposed to SSR)

pi0: Yeah. It could be better with placeholder rendering instead of failing whole page but we can certainly iterate to make it happen :)

danielroe: Initially I did show a placeholder but didn't have time to complete design. Switched to full error page until design is complete 😁

packages/nuxt/src/core/features.ts
danielroe: what makes you think that this is a serious bug? we're not executing this nuxi command within the rootDir...

pi0: Not for nuxt but for nuxi which i like to investigate. If `nuxi` CLI is executed with (OS) CWD of `rootDir`, it should run `module add` also within it not wrongly to the monorepo hoisted value. (or perhaps the `runComand` itself could accept CWD for programmatic API) 

Also i guess now would make sense we expose `runCommand` from kit wdyt?

Both above and this are unrelated to PR but worth to track)

danielroe: Could we support passing arbitrary args as part of options, perhaps?

packages/nuxt/src/app/components/nuxt-stubs.ts
```diff
@@ -0,0 +1,17 @@
...
+    statusMessage: `${name} is provided by @nuxt/image. Check your console to install it or run 'npx nuxi@latest module add @nuxt/image'`
```
Gruce, line 7: Is this `... add image` as in the [PR](https://github.com/nuxt/nuxt/pull/23741/files#diff-56eb09419c0f270f530ddd38f7161f0d157341991971611ac7fb060db83d97efR19) or `... add @nuxt/image`?

danielroe: Either way works.

PR #23724, opened 2023-10-17
feat(nuxt): allow configuring default `<NuxtLink>` options

danielroe:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR allows configuring NuxtLink defaults, such as trailing slash behaviour and default classes/prefetch behaviour. All of these are serialisable and initially can be set in `experimental.defaults.nuxtLink` but we will likely want to move to appConfig or `app/` directory later on.

Example:

```ts
export default defineNuxtConfig({
  experimental: {
    defaults: {
      nuxtLink: {
        activeClass: 'nuxt-link-active',
        trailingSlash: 'append'
      }
    }
  }
})
```

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #23725, opened 2023-10-17
feat(nuxt): allow customising defaults for data composables

danielroe:
### ❓ Type of change

### 📚 Description

This allows us to set default options for `useAsyncData` and `useFetch`, including `deep: false` for data composables. (This will enable turning off deeply reactive responses across your app.)

### 📝 Checklist


danielroe:
Current dependencies on/for this PR:
* main
  * **PR #23724** <a href="https://app.graphite.dev/github/pr/nuxt/nuxt/23724" target="_blank">Graphite</a> 
    * **PR #23725** <a href="https://app.graphite.dev/github/pr/nuxt/nuxt/23725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/nuxt/nuxt/23725?utm_source=stack-comment).

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -146,14 +149,15 @@ export function useAsyncData<
 
   options.lazy = options.lazy ?? false
   options.immediate = options.immediate ?? true
+  options.deep = options.deep ?? asyncDataDefaults.deep
```
pi0, line 152: If deep is expected to be default behavior, can't we simply chain with `?? true` here to avoid inlining defaults?

danielroe: I'm not sure that `{ deep: true }` is heavier than inlining `{}` and `asyncDataDefaults.deep ?? true`. Or am I missing something?

packages/schema/src/config/experimental.ts
```diff
@@ -258,7 +258,15 @@ export default defineUntypedSchema({
...
-      }
+      },
+      /**
+       * Options that apply to `useAsyncData` (and also therefore `useFetch`)
+       */
+      useAsyncData: {
+        deep: true
+      },
+      /** @type {Pick<typeof import('ofetch')['FetchOptions'], 'timeout' | 'retry' | 'retryDelay' | 'retryStatusCodes'>} */
+      useFetch: {}
```
pi0, line 269: It is nice to pick only safer options but just leaving as a note, we should explicitly disallow `baseURL` and `headers` at least. I would even type them as `never` explicitly to avoid future confusions. Sharing such options between requests is dangerous...

pi0: (We might also think about some standard type export for safe global options in ofetch later. feel free to make an issue to not forget if i didn't)

packages/nuxt/src/app/composables/asyncData.ts
pi0: Mainly consistency and readability (noticed lazy/immediate defaults are already inlined now...)

packages/schema/src/config/experimental.ts
danielroe: Nice idea. But just checked - it's a type error already as we don't have a `[key: string]: any` signature; adding `?: never` doesn't improve type safety in this particular case.

danielroe: added step to sanitise those in particular.

pi0 (approved):

PR #23751, opened 2023-10-18
fix(nuxt): resolve internal `target: blank` links with base

### LINKED ISSUE #23737: NuxtLink with target="_blank" doesn't render correctly
### Environment

- Operating System: Darwin
- Node Version:     v18.16.1
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  pnpm@8.7.6
- Builder:          -
- User Config:      app
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-uon84v

### Describe the bug

What I do
1. set app.baseURL in nuxt.config.ts,
2. write a NuxtLink with target="_blank", and prop "to" equals path start with slash, which linked to another page in project.

Then the rendered <a> tag's href doesn't prefixed with the value of app.baseURL.

But if I remove target="_blank", or remove the start slash of path, final link prefixed with baseURL, which is as expect

Is there any mistask I made? Or it is a bug?

### Additional context

_No response_

### Logs

_No response_

Jannchie:
### 🔗 Linked issue

- resolve #23737 

### ❓ Type of change

### 📚 Description

Now, such a link will automatically append the baseURL.
``` vue
  <div>
    <NuxtLink to="/about" target="_blank">link with target</NuxtLink>
  </div>
```

This should be a breaking change. because it's possible that someone could explicitly add baseURL to the `to` attribute to get around the bug.

### 📝 Checklist


Mini-ghost:
Hi! I happen to be researching this issue as well. Would you like to consider some other ideas?

```ts
let href = typeof to.value === 'object' ? router.resolve(to.value)?.href ?? null : to.value || null

if (href && href.startsWith('/') && !href.startsWith('//')) {
  const _base = withLeadingSlash(withTrailingSlash(baseURL))
  if (_base !== '/' && !href starts with(_base)) {
    href = joinURL(_base, href)
  }
}
```

This way, the following examples will work as expected:

```vue
<template>
  <NuxtLink to="http://localhost:3000/app/about" target="_blank">
    link with target and protocol
  </NuxtLink>
  <!--
    Output:
    <a href="http://localhost:3000/app/about" rel="noopener noreferrer" target="_blank"> link with target and baseURL </a>
  -->

  <NuxtLink to="//localhost:3000/app/about" target="_blank">
    link with target and starts with `//`
  </NuxtLink>
  <!--
    Output:
    <a href="//localhost:3000/app/about" rel="noopener noreferrer" target="_blank"> link with target and baseURL </a>
  -->

  <NuxtLink to="/app/about" target="_blank">
    link with target and baseURL
  </NuxtLink>
  <!--
    Output:
    <a href="/app/about" rel="noopener noreferrer" target="_blank"> link with target and baseURL </a>
  -->

  <NuxtLink to="/about" target="_blank">
    link with target without baseURL
  </NuxtLink>
  <!--
    Output:
    <a href="/app/about" rel="noopener noreferrer" target="_blank"> link without target </a>
  -->
</template>

```

Jannchie:
> Hi! I happen to be researching this issue as well. Would you like to consider some other ideas?

Yeah, your idea is better. I modified this pull request with reference to your implementation... Thank you!

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -280,8 +279,15 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
         // Resolves `to` value if it's a route location object
         // converts `""` to `null` to prevent the attribute from being added as empty (`href=""`)
-        const href = typeof to.value === 'object' ? router.resolve(to.value)?.href ?? null : to.value || null
+        let href = typeof to.value === 'object' ? router.resolve(to.value)?.href ?? null : to.value || null
 
+        // joins with `baseURL` if it's an relative link
+        if (import.meta.client && href && href.startsWith('/') && !href.startsWith('//')) {
```
Mini-ghost, line 285: Why check only on the client side? This could lead to inconsistent results between the server side and the client side.

```suggestion
-        if (import.meta.client && href && href.startsWith('/') && !href.startsWith('//')) {
+        if (href && href.startsWith('/') && !href.startsWith('//')) {
```

danielroe (changes requested):
Would you be able to add some tests to cover this?

packages/nuxt/src/app/components/nuxt-link.ts
Mini-ghost: Could you please review it once more? I haven't personally encountered the situation where `useRuntimeConfig` is unable to retrieve the `baseURL` on the server side in my environment.

Jannchie: Have you tried this command?

```
pnpm vitest nuxt-link
```

In my environment, the test does not pass without detecting `import.meta.client`:

```
 FAIL  packages/nuxt/test/nuxt-link.test.ts > nuxt-link:propsOrAttributes > `isExternal` is `true` > href > resolves route location object
Error: [nuxt] instance unavailable
 ❯ useNuxtApp packages/nuxt/src/app/nuxt.ts:395:13
    409| }
    410|
    411| export function defineAppConfig<C extends AppConfigInput> (config: C): C {
       |                                             ^
    412|   return config
    413| }
 ❯ Module.useRuntimeConfig packages/nuxt/src/app/nuxt.ts:404:10
 ❯ packages/nuxt/src/app/components/nuxt-link.ts:286:27
 ❯ nuxtLink packages/nuxt/test/nuxt-link.test.ts:46:96
 ❯ packages/nuxt/test/nuxt-link.test.ts:115:16

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[3/3]⎯

 Test Files  1 failed (1)
      Tests  3 failed | 25 passed (28)
   Start at  14:31:52
   Duration  2.81s (transform 179ms, setup 2.22s, collect 167ms, tests 13ms, environment 0ms, prepare 155ms)
```

Jannchie: Have you tried to test it by `pnpm vitest nuxt-link`?

In test cases, it will report an error for lack of context.

May be we should fix it by:

```suggestion
        if (hasInjectionContext() && href && href.startsWith('/') && !href.startsWith('//')) {
```

Mini-ghost: Have you ever attempted to include this in tests?

```
// nuxt-link.test.ts
// Mocks Nuxt `useRuntimeConfig()`
vi.mock('../src/app/nuxt', () => ({
  useRuntimeConfig: () => ({
    app: {
      baseURL: '/'
    }
  })
}))
```

Jannchie: I see, thank you for pointing out the correct approach!

Jannchie:
> Would you be able to add some tests to cover this?

I've added the test cases!

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -280,8 +279,15 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
         // Resolves `to` value if it's a route location object
         // converts `""` to `null` to prevent the attribute from being added as empty (`href=""`)
-        const href = typeof to.value === 'object' ? router.resolve(to.value)?.href ?? null : to.value || null
+        let href = typeof to.value === 'object' ? router.resolve(to.value)?.href ?? null : to.value || null
 
+        // joins with `baseURL` if it's an relative link
+        if (href && href.startsWith('/') && !href.startsWith('//')) {
+          const baseURL = useRuntimeConfig().app.baseURL
+          if (href !== '/' && !href.startsWith(baseURL)) {
```
lihbr, line 287: What happens if it's a blank target to the homepage there? 🤔

Or I assume we want to check `base` here instead?

packages/nuxt/test/nuxt-link.test.ts
```diff
@@ -125,6 +135,21 @@ describe('nuxt-link:propsOrAttributes', () => {
       it('defaults to `null`', () => {
         expect(nuxtLink({ to: 'https://nuxtjs.org' }).props.target).toBe(null)
       })
+
+      it('target="_blank" with baseURL', () => {
```
lihbr, line 139: Small style suggestion:
```
prefixes target="_blank" internal links with baseURL
```

packages/nuxt/test/nuxt-link.test.ts
```diff
@@ -125,6 +135,21 @@ describe('nuxt-link:propsOrAttributes', () => {
...
+          return {
+            app: {
+              baseURL: '/base'
+            }
+          } as any
+        }, () => {
+          expect(nuxtLink({ to: '/to', target: '_blank' }).props.href).toBe('/base/to')
+          expect(nuxtLink({ to: '/base/to', target: '_blank' }).props.href).toBe('/base/to')
+          expect(nuxtLink({ to: 'http://nuxtjs.org/app/about', target: '_blank' }).props.href).toBe('http://nuxtjs.org/app/about')
+          expect(nuxtLink({ to: '//nuxtjs.org/app/about', target: '_blank' }).props.href).toBe('//nuxtjs.org/app/about')
```
lihbr, line 150: Maybe it'd be semantically more correct to check for external links NOT to include the baseURL here.

lihbr (changes requested):
LGTM! just a small concern with the checks, and some optional style suggestions for the tests :)

Thanks for making this!

packages/nuxt/src/app/components/nuxt-link.ts
Jannchie: The '/' doesn't automatically append the baseURL with the current practice, which is indeed a problem.

But then I realized, maybe someone really needs to navigate to a path without baseURL?
Current practice does not seem to allow for this except by using full paths.

lihbr: OK, I think that makes sense, then we just need to update the following for now to my understanding:
```suggestion
          if (baseURL !== '/' && !href.startsWith(baseURL)) {
```

lihbr (approved):
LGTM, thanks!

lihbr (approved):

danielroe (approved):

PR #23780, opened 2023-10-19
chore: update to nitropack 2.7.0

pi0:
https://github.com/unjs/nitro/releases/tag/v2.7.0

Notable: Server bundle size reduced from `304k` to `197k` / `611k` to `504k` as `node-fetch` polyfill removed!

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -83,7 +83,7 @@
     "knitwork": "^1.0.0",
     "magic-string": "^0.30.5",
     "mlly": "^1.4.2",
-    "nitropack": "2.7.0",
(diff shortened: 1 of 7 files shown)


packages/nuxt/package.json
```diff
@@ -83,7 +83,7 @@
     "knitwork": "^1.0.0",
     "magic-string": "^0.30.5",
     "mlly": "^1.4.2",
-    "nitropack": "^2.6.3",
+    "nitropack": "2.7.0",
```
danielroe, line 86: Is this an intentional pin?

danielroe (approved):

PR #23841, opened 2023-10-20
fix(nuxt): handle relative baseURL in nitro runtime config

### LINKED ISSUE #23831: 404 with new App Manifest feature when using custom baseURL
### Environment

- Node Version:     v18.18.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.0
- Package Manager:  pnpm@8.9.0
- Builder:          -
- User Config:      devtools, app, runtimeConfig, experimental
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/amplitudesxd/nuxt-poc / https://stackblitz.com/edit/nuxt-starter-kfgvsc (they are identical)

### Describe the bug

When using Nuxt with the following config, a 404 for the URL `/builds/meta/test.json` occurs when Nuxt tries to access the application manifest:
```
app: {
  baseURL: './',
},
runtimeConfig: {
  app: {
    baseURL: './',
  },
}
```

The baseURL is being set to `./` to allow Nuxt to work with Electron's file:// protocol, per the following module - which I use in the full project, but have excluded for the minimal reproduction:
https://github.com/caoxiemeihao/nuxt-electron/blob/main/src/index.ts#L175

Everything works fine when `experimental.appManifest` is set to false.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23831

### ❓ Type of change

### 📚 Description

Nitro currently doesn't handle or support being passed a relative baseURL. I think it also doesn't particularly make sense for Nitro. So this change simply makes the url absolute instead.

I also ship an improvement in the manifest handling, to use the path utils we already have for giving us the build asset dir rather than reimplementing. However, this change wasn't needed to fix the bug; it's merely an optimisation I discovered in the process of resolving the bug..

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/manifest.ts
+++ b/packages/nuxt/src/app/composables/manifest.ts
@@ -1,10 +1,11 @@
-import { joinURL } from 'ufo'
 import type { MatcherExport, RouteMatcher } from 'radix3'
 import { createMatcherFromExport } from 'radix3'
 import { defu } from 'defu'
-import { useAppConfig, useRuntimeConfig } from '#app'
+import { useAppConfig } from '#app'
 // @ts-expect-error virtual file
 import { appManifest as isAppManifestEnabled } from '#build/nuxt.config.mjs'
+// @ts-expect-error virtual file
+import { buildAssetsURL } from '#build/paths.mjs'
 
 export interface NuxtAppManifestMeta {
   id: string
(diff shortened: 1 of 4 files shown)


pi0 (approved):

abhay-agarwal:
This change breaks our electron app configuration (e.g. nuxt-electron) which requires relative URLs. When downgrading from Nuxt 3.8.1 -> 3.7.0, setting `baseURL: "./"` turns paths in index.html to relative URLs, while in 3.8.1 this setting does nothing. Is there a way to regain this functionality in 3.8.1? Alternatively, is there a way to package a nuxt electron app so that it works with absolute URLs?

danielroe:
You should be able to override this change by setting `nitro.runtimeConfig.app.baseURL` to the previous value. (This PR only changed the _default_.)

abhay-agarwal:
Ah I see, I was setting a different baseURL runtimeConfig variable, not the nitro one.

PR #23856, opened 2023-10-22
fix(nuxt,schema,vite,webpack): analyze w/o overriding config

danielroe:
### ❓ Type of change

### 📚 Description

This resolves two issues:

1. we now have correct types (that is, not `any`) for `build.analyze`
2. we now respect user-passed `build.analyze` options and do not enable analysis just by having these options set in config (once nuxi is updated)

#### Example

```ts
  build: {
    analyze: {
      template: 'sunburst'
    }
  },
})
```

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/webpack.ts
+++ b/packages/schema/src/config/webpack.ts
@@ -1,4 +1,4 @@
-import { join } from 'pathe'
+import defu from 'defu'
 import { defineUntypedSchema } from 'untyped'
 
 export default defineUntypedSchema({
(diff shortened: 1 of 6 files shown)


packages/schema/src/config/webpack.ts
```diff
@@ -1,4 +1,4 @@
-import { join } from 'pathe'
+import defu from 'defu'
```
pi0, line 1:
```suggestion
import { defu } from 'defu'
```

pi0 (approved):

PR #23859, opened 2023-10-22
test: update basic fixture config for vite 5

bluwy:
### 🔗 Linked issue

n/a

### ❓ Type of change

### 📚 Description

In Vite 5, SVGs are inline-able and caused [some tests to fail](https://github.com/vitejs/vite-ecosystem-ci/actions/runs/6583609019/job/17886920571#step:8:2857). I updated the Nuxt config's `vite.build.assetsInlineLimit` option for now so the SVGs are not inlined.

I also ran the tests locally with this change and they passed.

---

PS: If I may have your attention 😅 An upcoming Vite PR (https://github.com/vitejs/vite/pull/11151) is also going to break Nuxt, specifically for this config:

https://github.com/nuxt/nuxt/blob/248e28a1bf1b163c13201f8ffd7951187fcf605b/packages/vite/src/server.ts#L48-L52

`esbuild` doesn't quite like it, but if needed you can pass those via `@rollup/plugin-replace` instead.

### 📝 Checklist

### DIFF
--- a/test/fixtures/basic/nuxt.config.ts
+++ b/test/fixtures/basic/nuxt.config.ts
@@ -123,7 +123,10 @@ export default defineNuxtConfig({
     undefined
   ],
   vite: {
-    logLevel: 'silent'
+    logLevel: 'silent',
+    build: {
+      assetsInlineLimit: 100 // keep SVG as assets URL
+    }
   },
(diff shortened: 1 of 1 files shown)


danielroe:
Thank you ❤️

danielroe (approved):

Issue #20533, opened 2023-04-26
Significant Sass build time difference between Nuxt and Vite


miosenpai:
### Environment

```
------------------------------
- Operating System: Linux
- Node Version:     v16.14.2
- Nuxt Version:     3.4.2
- Nitro Version:    2.3.3
- Package Manager:  npm@7.17.0
- Builder:          vite
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------
```

### Reproduction

Nuxt Project: https://stackblitz.com/edit/github-4zwpte
Vite Project: https://stackblitz.com/edit/github-soapej

### Describe the bug

While working on a Nuxt project, I noticed the project's start up time increased significantly (from a few hundred ms to 9000ms+) after I imported a Sass library. I originally thought this was from the Sass compiler being slow in general but after replicating the same setup in a standalone Vite project I saw very little increase to its start up time, leading me to consider if the performance hit could be Nuxt specific.

Something else I noticed is that with `DEBUG=*`, the Nuxt project logged additional `.scss` entries and the entries also took more time compared to the Vite project (although this could be intended behavior?).

### Additional context

I tried disabling CSS sourcemaps in the Nuxt project after finding https://github.com/vitejs/vite/issues/8656#issuecomment-1160683426 using `sourcemap: false` in `nuxt.config.ts` (https://github.com/nuxt/nuxt/pull/18446), but this didn't seem to help.

I was also able to reproduce the slowdown with other Sass libraries like `bootstrap/scss/bootstrap.scss`.

PixsaOJ:
So is this issue fixed? Or do we know the problem?

manniL:
Should be fixed since 3.8.1, yes.

PixsaOJ:
Well, I am on  "3.8.2" soo...

PR #23861, opened 2023-10-22
fix(vite): remove `postcss-url` and duplicate `postcss-import`

### LINKED ISSUE #15765: Load a external font failed the server response is incorrect
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `pnpm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction
[https://github.com/yyyanghj/nuxt-issue-9357](https://github.com/yyyanghj/nuxt-issue-9357)

### Describe the bug

Importing a CSS file to load a font failed. The server responds with an html file instead of a font file.

If I import the font.css in the `<script>`, everything is fine.

```vue
<script>
// uncomment next line works well.
// import "@fontsource/lora";
</script>

<style>
/* this does not work. */
@import "@fontsource/lora";

body {
  font-family: "Lora";
  font-size: 100px;
}
</style>

```

https://user-images.githubusercontent.com/34713301/204149097-89818bda-76e4-4d54-bf31-563142e19702.mov

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15765

### ❓ Type of change

### 📚 Description

This was a fun deep-dive.

Investigating the linked issue, I found that we are adding `postcss-import` - _and so is vite_. In addition, I believe the key features `postcss-url` (which was causing the bug) are in fact already covered by Vite's list of features:

> Vite improves @import resolving for Sass and Less so that Vite aliases are also respected. In addition, relative url() references inside imported Sass/Less files that are in different directories from the root file are also automatically rebased to ensure correctness.
> https://vitejs.dev/guide/features.html#css-pre-processors

Accordingly, this PR removes both plugins (or rather, doesn't add them).

I suspect removing these two plugins will improve our performance as well as fix the bug. I would also consider we reconsider whether to include `cssnano` or defer to vite's esbuild/lightningcss minificiation step in future (cc: antfu).

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/postcss.ts
+++ b/packages/schema/src/config/postcss.ts
@@ -1,5 +1,3 @@
-import { defu } from 'defu'
-import createResolver from 'postcss-import-resolver'
 import { defineUntypedSchema } from 'untyped'
 
 export default defineUntypedSchema({
@@ -11,29 +9,6 @@ export default defineUntypedSchema({
      * @type {Record<string, any>}
      */
     plugins: {
-      /**
-       * https://github.com/postcss/postcss-import
-       */
-      'postcss-import': {
-        $resolve: async (val, get) => val !== false
-          ? defu(val || {}, {
-            resolve: createResolver({
-              alias: { ...(await get('alias')) },
-              modules: [
-                await get('srcDir'),
-                await get('rootDir'),
-                ...(await get('modulesDir'))
-              ]
-            })
-          })
-          : val
-      },
-
-      /**
-       * https://github.com/postcss/postcss-url
-       */
-      'postcss-url': {},
-
       /**
        * https://github.com/postcss/autoprefixer
        */
--- a/packages/webpack/src/utils/postcss.ts
+++ b/packages/webpack/src/utils/postcss.ts
@@ -1,3 +1,4 @@
+import createResolver from 'postcss-import-resolver'
 import { createCommonJS } from 'mlly'
 import { requireModule } from '@nuxt/kit'
 import type { Nuxt } from '@nuxt/schema'
@@ -25,7 +26,7 @@ const orderPresets = {
   }
 }
 
-export const getPostcssConfig = (nuxt: Nuxt) => {
+export const getPostcssConfig = async async async async (nuxt: Nuxt) => {
   function sortPlugins ({ plugins, order }: any) {
     const names = Object.keys(plugins)
     if (typeof order === 'string') {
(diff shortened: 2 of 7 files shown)


packages/webpack/src/utils/postcss.ts
```diff
@@ -25,7 +26,7 @@ const orderPresets = {
   }
 }
 
-export const getPostcssConfig = (nuxt: Nuxt) => {
+export const getPostcssConfig = async async async async (nuxt: Nuxt) => {
```
johannschopplich, line 29: This seems to be a typo? 🙂 
```suggestion
export const getPostcssConfig = async (nuxt: Nuxt) => {
```

danielroe: thank you! ❤️

packages/schema/src/config/postcss.ts
```diff
@@ -11,29 +9,6 @@ export default defineUntypedSchema({
...
-            })
-          })
-          : val
-      },
-
-      /**
-       * https://github.com/postcss/postcss-url
-       */
-      'postcss-url': {},
-
```
HigherOrderLogic, line 36: Shouldnt we leave the type as is and mark it as deprecated to avoid breaking changes?

mrleblanc101:
Shouldn't this mention/link to #20533 since all the related issue were merged into that one ?

antfu (approved):

antfu:
For CSS minification, I am not sure if users would want full control over which tool to use, maybe could be an option as `'cssnano' | 'esbuild' | 'lightningcss'` where we built a thin layer of adapter?

danielroe:
I think that would be a very nice idea.

nathanchase:
@danielroe @antfu 

> For CSS minification, I am not sure if users would want full control over which tool to use, maybe could be an option as `'cssnano' | 'esbuild' | 'lightningcss'` where we built a thin layer of adapter?

Full control is good, but also with the most performant (sensible) default (likely `lightningcss`). I was recently looking at whether or not something like [nuxt-modules/critters](https://github.com/nuxt-modules/critters) or [nuxt-beastcss](https://github.com/bnznamco/nuxt-beastcss) was even necessary, or if Nuxt 3 already accomplishes the same goals. 

There's not much in the docs other than Nuxt is pre-configured to use `cssnano` for "minification and purge" (at the end of "Using Postcss") https://nuxt.com/docs/getting-started/styling#using-postcss - but nothing specifically about best practices of inlining critical css. 

I know there's the `inlineSSRStyles` [experimental flag](https://nuxt.com/docs/api/nuxt-config#inlinessrstyles) defaulting to true, but unsure if that's doing the same thing that critters / beastcss are or not.

Anyway, ideally Nuxt 3 takes care of all of the above for us (inlined critical CSS, fastest minification and purging), and maybe explains the optimal pre-configured choice (lightning) - but then lets us opt for other less performant options if it's somehow a project requirement.

PR #23868, opened 2023-10-22
fix(nuxt): use route key for loading indicator/view transition

### LINKED ISSUE #21991: loading indicator shows/resolves too quickly when changing query
**Reproduction**: https://stackblitz.com/edit/nuxt-starter-3ab9sw?file=app.vue

Whithout `:throttle="0"`, navigate to "blog" and then to page 2. Loading indication works. But now if you navigate to page 1, 3 or 4 it not works. This pages are in the same component `[page].vue` but with different "page" param. Slugs `/blog/page/1`, `/blog/page/2`...

If you add `:throttle="0"` the indicator appears but with a weird speed... starting/stopping quickly.

I think it's necessary to do an extra check... maybe something like:

```js
if (to === from || to.matched.every((comp, index) => {
  return comp.components && comp.components?.default === from.matched[index]?.components?.default && 
Object.values(to.params) === Object.values(from.params)
})) {
  indicator.finish();
}
```

This works for my example.

_Originally posted by @emaia in https://github.com/nuxt/nuxt/issues/21656#issuecomment-1616829484_

huang-julien:
### 🔗 Linked issue
resolve #21991

### ❓ Type of change

### 📚 Description

hi :wave: this PR makes NuxtLoadingIndicator stop in the `beforeResolve` hook only if:
- it is the same component that is mounted AND route params are different
- to === from

couldn't find a way to write proper tests. 

### 📝 Checklist


packages/nuxt/src/app/components/nuxt-loading-indicator.ts
```diff
@@ -42,7 +42,19 @@ export default defineComponent({
       indicator.finish()
     })
     router.beforeResolve((to, from) => {
-      if (to === from || to.matched.every((comp, index) => comp.components && comp.components?.default === from.matched[index]?.components?.default)) {
```
danielroe, line 45: perhaps we could early return if `to === from` to save the extra work?

packages/nuxt/src/app/components/nuxt-loading-indicator.ts
```diff
@@ -42,8 +42,25 @@ export default defineComponent({
...
+      })
+
+      if (areParamsEqual) {
+        const areComponentsSame = to.matched.every((comp, index) =>
+          comp.components && comp.components.default === from.matched[index]?.components?.default
+        )
+        if (areComponentsSame) {
+          indicator.finish()
+        }
       }
```
danielroe, line 64:
```suggestion
      if (!areParamsEqual) { return }

      const areComponentsSame = to.matched.every((comp, index) =>
        comp.components && comp.components.default === from.matched[index]?.components?.default
      )
      if (areComponentsSame) {
        indicator.finish()
      }
```

danielroe (commented):
This looks like a good improvement.

Could we extract to a shared utility that can be used in:

- `<NuxtLoadingIndicator>`
- view transitions api plugin: https://github.com/nuxt/nuxt/blob/988e75bbc7a19422126ff563b6a6b88235131faa/packages/nuxt/src/app/plugins/view-transitions.client.ts#L13-L15
- and in any other similar kind of logic

Ideally this utility would respect [route keys](https://github.com/nuxt/nuxt/blob/ffd0223583116ad4b9d28898b4f30d73dbb657d7/packages/nuxt/src/pages/runtime/utils.ts#L16-L18) (as this is really what we are testing - whether we will need to rerun setup function for a new page component).

huang-julien:
@danielroe like this ? I think i have an idea to unify the loading state API for nuxt, i'll make it in another PR

danielroe (approved):

PR #23884, opened 2023-10-23
fix(nuxt): don't use local fetch with an external `baseURL`

### LINKED ISSUE #23880: About the isLocalFetch variable of useFetch
### Describe the feature

https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/fetch.ts#L143-L148C6

Currently, `isLocalFetch` is true when the request starts with `/`, and if I redefine `globalThis.$fetch` and set `baseURL`, then it's wrong.

When I tried to fix this I found that `ofetch` couldn't get the `options.baseURL` for the current instance, so I open https://github.com/unjs/ofetch/issues/303#issuecomment-1774885204

cc @pi0 

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23880

### ❓ Type of change

### 📚 Description

This updates the test for `isLocalFetch` to only match if there is no provided baseURL or the baseURL starts with `/`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -140,9 +140,15 @@ export function useFetch<
     controller?.abort?.()
     controller = typeof AbortController !== 'undefined' ? new AbortController() : {} as AbortController
 
-    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/') && (!unref(opts.baseURL) || unref(opts.baseURL)!.startsWith('/'))
+    let _$fetch = opts.$fetch || globalThis.$fetch
+
     // Use fetch with request context and headers for server direct API calls
-    const _$fetch = opts.$fetch || ((import.meta.server && isLocalFetch) ? useRequestFetch() : globalThis.$fetch)
+    if (import.meta.server && !opts.$fetch) {
+      const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/') && (!unref(opts.baseURL) || unref(opts.baseURL)!.startsWith('/'))
+      if (isLocalFetch) {
+        _$fetch = useRequestFetch()
+      }
+    }
 
     return _$fetch(_request.value, { signal: controller.signal, ..._fetchOptions } as any) as Promise<_ResT>
   }, _asyncDataOptions)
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -140,12 +140,9 @@ export function useFetch<
...
     controller = typeof AbortController !== 'undefined' ? new AbortController() : {} as AbortController
 
-    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/')
-    let _$fetch = opts.$fetch || globalThis.$fetch
+    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/') && (!unref(opts.baseURL) || unref(opts.baseURL)!.startsWith('/'))
     // Use fetch with request context and headers for server direct API calls
-    if (import.meta.server && !opts.$fetch && isLocalFetch) {
-      _$fetch = useRequestFetch()
-    }
+    const _$fetch = opts.$fetch || ((import.meta.server && isLocalFetch) ? useRequestFetch() : globalThis.$fetch)
```
pi0, line 145: Small perf: I'm not sure if this would be tree-shaken without modification so worth to double check if you can quickly check. Seems we need `isLocalFetch` computation only in server and without explicit `$fetch` option.

pi0 (approved):

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -140,12 +140,9 @@ export function useFetch<
     controller?.abort?.()
     controller = typeof AbortController !== 'undefined' ? new AbortController() : {} as AbortController
 
-    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/')
-    let _$fetch = opts.$fetch || globalThis.$fetch
+    const isLocalFetch = typeof _request.value === 'string' && _request.value.startsWith('/') && (!unref(opts.baseURL) || unref(opts.baseURL)!.startsWith('/'))
```
KeJunMao, line 143:
```js
useFetch('/users', { $fetch: $fetch.create({ baseURL: 'https://example.com/api/' }) })
// isLocalFetch === true ?
```

packages/nuxt/src/app/composables/fetch.ts
danielroe: yes, it's still tree-shaken.

packages/nuxt/src/app/composables/fetch.ts
danielroe: yes, but it isn't used... will push a perf refactor to avoid computing it at all if not needed...

PR #23899, opened 2023-10-23
fix(nuxt): render a div when client-only hydrates w/o element

### LINKED ISSUE #23876: Cannot read properties of null (reading 'parentNode') with client Component and OnMounted
### Environment

Working directory: /home/projects/nuxt-starter-3ukexv                                                             10:21:36
Nuxt project info:                                                                                                10:21:36

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.0
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

👉 Report an issue: https://github.com/nuxt/nuxt/issues/new                                                       10:21:36

👉 Suggest an improvement: https://github.com/nuxt/nuxt/discussions/new

### Reproduction

https://stackblitz.com/edit/nuxt-starter-3ukexv?file=pages%2Findex.vue

### Describe the bug

We are upgrading from 3.7.4 to 3.8.0 and we have a lot of errors "Cannot read properties of null (reading 'parentNode')"

This errors doesnt appears with 3.7.4

One of them you can see in the reproduction link.

The most of them, are related with .client components

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue

fix #23876 

### ❓ Type of change

### 📚 Description
Hi :wave: 
This PR makes the component cloned from `createClientOnly` render a `<div></div>` if there's no element attached to it's instance.
 
The issue does not happens on hydration because an element is already attached to the component instance on hydration.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/client-only.ts
+++ b/packages/nuxt/src/app/components/client-only.ts
@@ -41,7 +41,7 @@ export function createClientOnly<T extends ComponentOptions> (component: T) {
           : h(res)
       } else {
         const fragment = getFragmentHTML(ctx._.vnode.el ?? null)
-        return process.client ? createStaticVNode(fragment.join(''), fragment.length) : h('div', ctx.$attrs ?? ctx._.attrs)
+        return process.client ? createStaticVNode(fragment.join('') || '<div></div>', fragment.length) : h('div', ctx.$attrs ?? ctx._.attrs)
       }
     }
   } else if (clone.template) {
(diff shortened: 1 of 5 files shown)


packages/nuxt/src/app/components/client-only.ts
```diff
@@ -41,7 +41,7 @@ export function createClientOnly<T extends ComponentOptions> (component: T) {
           : h(res)
       } else {
         const fragment = getFragmentHTML(ctx._.vnode.el ?? null)
```
danielroe, line 43: should we instead make `getFragmentHTML` return null so we can add `|| ['<div></div>']` to this line, so `fragment.length` below is `1`? (I haven't tested this so you may have a reason for doing it as below.)

huang-julien: Sure ! 😅 this was just  the quickest solution i found. But modifying `getFramentHTML` to return `string[]|null` would also work !

danielroe (approved):

PR #20190, opened 2023-04-11
feat(nuxt): scan and register layouts in nested folders

Suv4o:
fix(nuxt): Nuxt 3 does not respect layouts in separate folders (#14275)

### 🔗 Linked issue

resolves #14275

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Fix the issue of not being able to create layouts in different directories, as specified in Issue #14275.

Now, we can define layouts in the following format:

<p dir="auto" style="box-sizing: border-box; margin-top: 0px; margin-bottom: 16px; color: rgb(230, 237, 243); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">For example:</p>

File | Layout name
-- | --
~/layouts/desktop/default.vue | 'desktop-default'
~/layouts/desktop-base/base.vue | 'desktop-base'
~/layouts/desktop/index.vue | 'desktop'

```
<template>
    <NuxtLayout name="desktop-base">
      <NuxtPage />
    </NuxtLayout>
</template>
```

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #23902, opened 2023-10-24
fix(nuxt): preserve hyphens in component/layout kebab names

danielroe:
### ❓ Type of change

### 📚 Description

This fixes a regression in https://github.com/nuxt/nuxt/pull/20190 where hyphen data was missing because we converted kebab -> pascal -> kebab, e.g. `base-1-thing` -> `Base1Thing` -> `base1-thing`. This preserves the initial segments to revert that change.

Because of the value of keeping things consistent, I've also applied the change to the component resolver. This should only affect kebab-casing with numbers, but although it is a behaviour change I think it is a fix (so `test-1-file.vue` can be used as `<test-1-file>` rather than requiring `<test1-file>`).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/scan.ts
+++ b/packages/nuxt/src/components/scan.ts
@@ -1,14 +1,12 @@
 import { readdir } from 'node:fs/promises'
 import { basename, dirname, extname, join, relative } from 'pathe'
 import { globby } from 'globby'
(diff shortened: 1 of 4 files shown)


pi0 (approved):
LGTM

PR #23951, opened 2023-10-26
perf(nuxt): use granular imports for auto-importing composables

danielroe:
### ❓ Type of change

### 📚 Description

This PR aims to improve Vite HMR + cold start/warmup performance by reducing crawling of unneeded files in the nuxt app:
1. within nuxt app, we now entirely use relative paths to import composables _only_ from the file exporting them rather than from the entrypoint
2. within user projects, we now auto-import directly from the file within nuxt app rather than from the entrypoint (`#app`)

### 📝 Checklist


manniL:
@danielroe Is really on a roll providing PRs to all the public repos that will benefit from that change. Amazing job 👏🏻

PR #23952, opened 2023-10-26
repair: nuxi -> nuxt

liangyuetian:
repair: nuxi -> nuxt


danielroe:
`nuxi` is correct in this context. See https://github.com/nuxt/cli for more information. (`nuxt` will also work, but we use `nuxi` by preference as it matches the package name.)

PR #23997, opened 2023-10-28
feat(nuxt): allow configuring nitro preset with `BUILD_PRESET`

### LINKED ISSUE #23996: Add `BUILD_PRESET` alias as `NITRO_PRESET`
IMO it makes sense to have it as an alias of `NITRO_PRESET` in term of DX.

Also noted that Analog.js also have this alias.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23996

### ❓ Type of change

### 📚 Description

This allows configuring the nitro preset with `BUILD_PRESET` environment variable.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -42,7 +42,6 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   )
 
   const nitroConfig: NitroConfig = defu(_nitroConfig, {
-    preset: process.env.BUILD_PRESET || undefined,
     debug: nuxt.options.debug,
     rootDir: nuxt.options.rootDir,
     workspaceDir: nuxt.options.workspaceDir,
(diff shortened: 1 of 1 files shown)


atinux (approved):

pi0:
I think it would be nice to consider some other options such as SERVER_PRESET and have a cross discussion in nitro before proceeding on this.

Atinux:
SERVER_PRESET could make sense if it can support it in development as well.

It makes sense for me that it's called BUILD_PRESET since it's related to the build command but open also for SERVER.

Nuxt goal's always has been to abstract the complexity to give end users a unified experience to create web applications.

Webpack and SSR configuration was the first abstraction and it was for no reason not giving back to Webpack its recognition.

I believe that telling Nuxt users to use NITRO* env variable to build on a specific target is confusing in term of DX (and it is not the first time I mention this).

pi0:
Yeah, Nuxt hides complexities but never terminologies. Builder is still called `vite`, we have `nuxtApp.vueApp`, etc.

But anyway i agree there might be also some points in also abstracting configuration so i guess i am okay that we are doing this.

I appreciate if we move this discussion to https://github.com/unjs/nitro/issues/1860 since there are more important points for **consistency** with future nitro global options and also not diverging Nuxt behavior from Nitro. If introducing a vendorless option, it can directly be in Nitro unless we have a real reason for it.

> SERVER_PRESET could make sense if it can support it in development as well.

Actually, another reason is that for choosing server! I imagine in the future nitro will support development presets as well (as announced in Nitro routemap during Nuxt nation) -- in addition to the plans for adopting `SERVER_*` env support in nitro docs for consistency.

PR #24024, opened 2023-10-30
fix(nuxt): provide route component names to `KeepAlive` cache

### LINKED ISSUE #15214: keepalive `include`/`exclude` props don't work with `<NuxtPage>`
### Environment

- Node Version:     `v16.16.0`
- Nitro Version:    `0.5.4`
- Package Manager:  `npm@8.11.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-xnrjjn?file=pages%2Findex.vue,pages%2Fnotkeptalive.vue

### Describe the bug

App has 2 pages, one page is set to be kept alive with 
`definePageMeta({ keepalive: true });`
the other one is not.

Changing some value in the first page, is not kept if you navigate away from it and if you come back.
Also, there is a weird display bug, as if a new mounted component of the page was added under the first one in the page before the first one being added. It make the page flicker.

If you put the same page meta in the second one (by removing the comment is the reproduction link), now it works, but the two pages are now kept alive and this is not what I want.

In base vue3, I could do something like that :
```
<router-view v-slot="{ Component }">
      <keep-alive include="index">
        <component :is="Component" />
      </keep-alive>
    </router-view>
```

### Additional context

_No response_

### Logs

_No response_

Clarkkkk:
### 🔗 Linked issue

#15214

### ❓ Type of change

### 📚 Description

Hi, this PR is aimed to make the `keepalive` works as expected and fix #15214

Changes:

- Change the nested order of the components in `NuxtPage`, letting `KeepAlive` read the correct component name
- Change the relavant code to let them work the same way as before
- Add tests for the keepalive features

There is one problem I am not sure about. In `RouteProvider`, it uses `previousKey` and `previousRoute` to "Prevent reactivity when the page will be rerendered in a different suspense fork" as the comment says, which I don't really understand. Since `setup` in `RouteProvider` will not rerun after all the changes, I use a `watch` to update `previousKey` and `previousRoute`. I don't know if it works the same way as before, and it seems there is no test case for this scenario.

### 📝 Checklist


Clarkkkk:
A test about HMR failed, working on it.

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -81,29 +94,35 @@ export default defineComponent({
...
-            )).default()
+                onPending: () => nuxtApp.callHook('page:start', routeProps.Component),
+                onResolve: () => {
+                  nextTick(() => nuxtApp.callHook('page:finish', routeProps.Component).finally(done))
+                }
+              }, componentVNode)
+            )
+          ))
+
+          ;(vnode.type as any).__hmrId = (componentVNode.type as any).__hmrId
```
danielroe, line 125: we should probably only do this in dev mode:
```suggestion
          if (import.meta.dev) {
            (vnode.type as any).__hmrId = (componentVNode.type as any).__hmrId
          }
```

package.json
```diff
@@ -17,6 +17,7 @@
     "play": "nuxi dev playground",
     "play:build": "nuxi build playground",
     "play:preview": "nuxi preview playground",
+    "fixtures:basic:dev": "nuxi dev test/fixtures/basic",
```
danielroe, line 20:
```suggestion
```

test/fixtures/basic/pages/keepalive/composables.ts
```diff
@@ -0,0 +1,6 @@
+export function useLifecyleLogs (name: string) {
```
danielroe, line 1: we should probably move this out to `test/fixtures/basic/composables` rather than keep them here, as this will create a page route.

Clarkkkk:
Thanks for the quick review! All done.

Maxbsy:
Tips:This branch is out-of-date with the base branch，You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request.

danielroe (approved):

danielroe (changes requested):
in further testing it seems that this change undoes the purpose of the route provider, which is to ensure a stably injected `route` in a suspense fork.

I've created a test (https://github.com/nuxt/nuxt/pull/24195) which should show the behaviour we want to have - once merged to main and updated in this pr, you can test it with:

```bash
pnpm nuxi dev test/fixtures/basic
```

and navigate to `http://localhost:3000/route-provider/foo` and then navigate to `bar`

Clarkkkk:
Thanks! I will investigate it

Clarkkkk:
After some investigations, i found that the behavior of suspense and async component is heavily relying on the nested order of the components in `page.ts`. The change i've made breaks such behaviors and seems not likely to be fixed.

Another solution that comes to my mind is to set the component name of `route-provider` dynamically in render function. We could set the name same as its child component so that `keep-alive` can cache the right component. It may look confusing in vue dev tools since there are two components with the same name. But it seems acceptable as a temporary solution.

In the long run, to address this issue, `keep-alive` need to be enhanced in vue, and provide the ability to choose which vnode to cache by vnode key or some kind of mark.

Do you have any better idea? @danielroe

danielroe:
@Clarkkkk I think that would be a good solution. And I think we can only get rid of `<RouteProvider>` if this functionality is built into vue-router itself.

Clarkkkk:
I revert the previous changes and set the component name dynamically. `<KeepAlive>` should work as expected now.

The test result is different because `onActivated` is triggered before `onDeactivated` when using `<Suspense>` in `<Keepalive>`. And I believe it is the current behavior of Vue ([example](https://play.vuejs.org/#__SSR__eNrNVE2P2jAQ/StuLrDSEovspaIJhbYr9UNqq26lXnJhkwG869iW7QASyn/v2CHBQUWt1B56AHlmnt/MPE/mGC2Vinc1RLMoNYVmyhIDtlbzXLBKSW3JkWhYk4astazICKGjPvRWVmp6CsTUW45r9GqASAaIpEPkopDCWFKZDclcjvHoPXAuyQ+peflidNMBilprEPYEmqI/pW2pWCQaFirFVxbQIiTdTufHoydtmpSi5b0l2/kDHh9ra6Ugi4Kz4jnLozN9f8oyMiWvSUJmZJpHc7NnttimtL3ZEtKOMf0EoJac7dr86HiojQJhOhs9rU67CVuH+bIMyQkdwhKEATfQ+1M64EtpmC+lffPRbdRqPqlWKn4yUuCTHt2V/BQweTQj3uN8+ArOzqOttcrMKC1KgddKQGodC7BUqIouEEZ1LSyrYFLKanEX38UvsXljQ3cMppo8ark3oGOlZYlMeXQb5KIY2YGeYOslaNB/mvviWpj/IjSowaV32ZtcNKiMNThKa7a50KVAwRkH/UVZhqM20GeFo7j/6H1W19D3UmyheP6F/8kc2p6+avCVBf3bld6AbcP3D5/hgOc+WMmy5qe3uBL8Bkby2tXYwt7UosSyA5yv9oN/ZiY23839weLIdE25Qr0aHu/fw83atdbP5aLagYr9B365LAhfiQ2Othuxea6D1SHFsrBshwNa3qLxDladOdwoeMf9Avh4fEOyORaFbkLcJpAcYi4341H7OfVMuCkQ0/j/liRI81ua8owNic4r5sQarBlPlW6TuWfAJZN4X/AxOvMkmN93/4Fgyb8RLPlrwZKrgjU/AeCBNxc=)). So I update the test result to reflect this.

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -88,20 +88,27 @@ export default defineComponent({
...
+                const providerVNode = h(RouteProvider, {
+                  key: key || undefined,
+                  vnode: routeProps.Component,
+                  route: routeProps.route,
+                  renderKey: key || undefined,
+                  trackRootNodes: hasTransition,
+                  vnodeRef: pageRef
+                })
+                if (import.meta.client && keepaliveConfig) {
+                  (providerVNode.type as any).name = (routeProps.Component.type as any).name || (routeProps.Component.type as any).__name || 'RouteProvider'
```
danielroe, line 108: Should we cover the case that `type` is a string/symbol, etc?

Clarkkkk: Is it necessary? Because in that case `Component.type.name` would be `undefined` and the value falls back to `'RouteProvider'`

danielroe: ah, good point...

danielroe (approved):

freezyh:
> ### 🔗 Linked issue
> #15214
> 
> ### ❓ Type of change
> * [ ]  📖 Documentation (updates to the documentation, readme or JSdoc annotations)
> * [x]  🐞 Bug fix (a non-breaking change that fixes an issue)
> * [ ]  👌 Enhancement (improving an existing functionality like performance)
> * [ ]  ✨ New feature (a non-breaking change that adds functionality)
> * [ ]  🧹 Chore (updates to the build process or auxiliary tools and libraries)
> * [ ]  ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
> 
> ### 📚 Description
> Hi, this PR is aimed to make the `keepalive` works as expected and fix #15214
> 
> Changes:
> 
> * Change the nested order of the components in `NuxtPage`, letting `KeepAlive` read the correct component name
> * Change the relavant code to let them work the same way as before
> * Add tests for the keepalive features
> 
> There is one problem I am not sure about. In `RouteProvider`, it uses `previousKey` and `previousRoute` to "Prevent reactivity when the page will be rerendered in a different suspense fork" as the comment says, which I don't really understand. Since `setup` in `RouteProvider` will not rerun after all the changes, I use a `watch` to update `previousKey` and `previousRoute`. I don't know if it works the same way as before, and it seems there is no test case for this scenario.
> 
> ### 📝 Checklist
> * [x]  I have linked an issue or discussion.
> * [ ]  I have updated the documentation accordingly.



> ### 🔗 Linked issue
> #15214
> 
> ### ❓ Type of change
> * [ ]  📖 Documentation (updates to the documentation, readme or JSdoc annotations)
> * [x]  🐞 Bug fix (a non-breaking change that fixes an issue)
> * [ ]  👌 Enhancement (improving an existing functionality like performance)
> * [ ]  ✨ New feature (a non-breaking change that adds functionality)
> * [ ]  🧹 Chore (updates to the build process or auxiliary tools and libraries)
> * [ ]  ⚠️ Breaking change (fix or feature that would cause existing functionality to change)
> 
> ### 📚 Description
> Hi, this PR is aimed to make the `keepalive` works as expected and fix #15214
> 
> Changes:
> 
> * Change the nested order of the components in `NuxtPage`, letting `KeepAlive` read the correct component name
> * Change the relavant code to let them work the same way as before
> * Add tests for the keepalive features
> 
> There is one problem I am not sure about. In `RouteProvider`, it uses `previousKey` and `previousRoute` to "Prevent reactivity when the page will be rerendered in a different suspense fork" as the comment says, which I don't really understand. Since `setup` in `RouteProvider` will not rerun after all the changes, I use a `watch` to update `previousKey` and `previousRoute`. I don't know if it works the same way as before, and it seems there is no test case for this scenario.
> 
> ### 📝 Checklist
> * [x]  I have linked an issue or discussion.
> * [ ]  I have updated the documentation accordingly.

if I switch other layout, the  keepalive will not work

freezyh:
> I revert the previous changes and set the component name dynamically. `<KeepAlive>` should work as expected now.
> 
> The test result is different because `onActivated` is triggered before `onDeactivated` when using `<Suspense>` in `<Keepalive>`. And I believe it is the current behavior of Vue ([example](https://play.vuejs.org/#__SSR__eNrNVE2P2jAQ/StuLrDSEovspaIJhbYr9UNqq26lXnJhkwG869iW7QASyn/v2CHBQUWt1B56AHlmnt/MPE/mGC2Vinc1RLMoNYVmyhIDtlbzXLBKSW3JkWhYk4astazICKGjPvRWVmp6CsTUW45r9GqASAaIpEPkopDCWFKZDclcjvHoPXAuyQ+peflidNMBilprEPYEmqI/pW2pWCQaFirFVxbQIiTdTufHoydtmpSi5b0l2/kDHh9ra6Ugi4Kz4jnLozN9f8oyMiWvSUJmZJpHc7NnttimtL3ZEtKOMf0EoJac7dr86HiojQJhOhs9rU67CVuH+bIMyQkdwhKEATfQ+1M64EtpmC+lffPRbdRqPqlWKn4yUuCTHt2V/BQweTQj3uN8+ArOzqOttcrMKC1KgddKQGodC7BUqIouEEZ1LSyrYFLKanEX38UvsXljQ3cMppo8ark3oGOlZYlMeXQb5KIY2YGeYOslaNB/mvviWpj/IjSowaV32ZtcNKiMNThKa7a50KVAwRkH/UVZhqM20GeFo7j/6H1W19D3UmyheP6F/8kc2p6+avCVBf3bld6AbcP3D5/hgOc+WMmy5qe3uBL8Bkby2tXYwt7UosSyA5yv9oN/ZiY23839weLIdE25Qr0aHu/fw83atdbP5aLagYr9B365LAhfiQ2Othuxea6D1SHFsrBshwNa3qLxDladOdwoeMf9Avh4fEOyORaFbkLcJpAcYi4341H7OfVMuCkQ0/j/liRI81ua8owNic4r5sQarBlPlW6TuWfAJZN4X/AxOvMkmN93/4Fgyb8RLPlrwZKrgjU/AeCBNxc=)). So I update the test result to reflect this.

switch other layout, the keepalive of before layout will not work

transtone:
the include method still have bug.
If use a array `cacheList` to control the keepalive page, add page name to cacheList will keep the page cache.
but when reduce the cacheList item, the page all refreshed!

```vue
<template>
      <NuxtPage :keepalive="{include:cacheList, max:20}" />
</template>
<script setup>
let cacheList = ref(['a','b','c'])
// the page a, b, c is cached.

const updateCacheList = ()=>{
   cacheList.value = ['a','b']
  // when doing this, the page 'a', 'b' refreshed!
}
</script>
```

PR #24029, opened 2023-10-30
fix(nuxt,schema): ignore files in `buildDir` but not dir itself

danielroe:
### ❓ Type of change

### 📚 Description

This allows users to register paths in `.nuxt` to be watched, e.g. with `ignore: ['!.nuxt/file.d.ts']`.

### 📝 Checklist


pi0:
Do you have more info about why this change/fix was required?

Currently, we don't have so many files in `.nuxt`, but it can easily cause performance implications like when having lots of files in `.nuxt/cache` or `.nuxt/nitro`, etc. And also if a file is frequently being rebuilt there.

If there are certain modules need to do this, maybe we can have some logic or hook to allow intentionally _deoptimizing_ to watch?

danielroe:
Yes, this is originally triggered by @Diizzayy who reached out about a change that affected https://github.com/Diizzayy/nuxt-graphql-client.

I think I can update so we de-optimise if there's a negative pattern in the ignore list.

pi0 (approved):
LGTM, however in my hearth i feel this could be something modules could opt-in (with custom watcher) if really need watching physical files in buildDir as we all tried to reduce usage of this dir. now it is better with auto deoptimize but (poor) users cannot have a choice in this and silently opt-in when use a module...

danielroe:
If you're concerned about perf, I think we can leave it to module authors to register a custom watcher...

pi0:
Would be nice to investigate module-level solution IMO (and if it proven to be a common shared requirement, we can also always think of a kit utility for it)

@Diizzayy if there is an issue or pr relavant, feel free to ping me.

PR #24032, opened 2023-10-30
perf(nuxt): set `pages` on nuxt app and deduplicate calls

danielroe:
### ❓ Type of change

### 📚 Description

This improves performance by reducing the number of times we crawl the filesystem to generate page routes. It does so by adding a new property on `app`: `app.pages`, where we can store the pages, just as we do for components, layouts, middleware, etc.

### 📝 Checklist


pi0 (approved):
Nice idea!

PR #24036, opened 2023-10-30
fix(nuxt): reload nitro when spa loading template updates

danielroe:
### ❓ Type of change

### 📚 Description

This registers the spa loading template as a path to restart nuxt for.

~~(I'll update this PR shortly to use watch hook so it doesn't need to restart, but opening it anyway to track.)~~

### 📝 Checklist


danielroe:
@pi0 do you think there's a way we can programmatically trigger a nitro rebuild to achieve HMR-like effect of updating spa loading template?

packages/schema/src/config/common.ts
```diff
@@ -361,7 +361,15 @@ export default defineUntypedSchema({
...
    */
   watch: {
-    $resolve: val => [].concat(val).filter((b: unknown) => typeof b === 'string' || b instanceof RegExp)
+    $resolve: async (val, get) => {
+      let spaLoadingTemplate = await get('spaLoadingTemplate')
+      if (typeof spaLoadingTemplate !== 'string') {
+        spaLoadingTemplate = 'app/spa-loading-template.html'
+      }
+      return [
+        relative(await get('rootDir'), resolve(await get('srcDir'), spaLoadingTemplate))
```
pi0, line 370: For future maintainability, I would create an array `watchFiles` and push to it (instead of directly contact edge case + filter

pi0 (approved):

pi0:
> do you think there's a way we can programmatically trigger a nitro rebuild to achieve HMR-like effect of updating spa loading template?

AFAIK rollup has no nice API for this and we really re-initialize it's instance internally https://github.com/unjs/nitro/blob/main/src/build.ts#L528

danielroe:
If you're happy with us calling that nitro hook, we could instead just do that when we detect a change to the file. wdyt?

pi0:
Sure you can use `rollup:reload` too if it works 👍🏼 (at least for Nitro 2, for Nitro 3 with other bundlers we might think of something better)

packages/nuxt/src/core/nitro.ts
```diff
@@ -366,6 +366,13 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Init nitro
   const nitro = await createNitro(nitroConfig)
 
+  // Trigger Nitro reload when SPA loading template changes
+  nuxt.hook('builder:watch', (_event, path) => {
+    if (normalize(path) === relative(nuxt.options.rootDir, resolve(nuxt.options.srcDir, 'app/spa-loading-template.html'))) {
+      nitro.hooks.callHook('rollup:reload')
```
pi0, line 372: would be nice to chain the hooks (await) to prevent unhandled promise errors

danielroe: it will still be unhandled as we do not await that hook.

PR #24043, opened 2023-10-30
fix(nuxt): account for delay ≤ 0 in `useCookie`

DamianGlowala:
### ❓ Type of change

### 📚 Description

This PR ensures `maxAge` is respected for negative values and 0, which is a falsy value. For delay less than or equal to 0, `setTimeout` would default it to 1, whereas this PR sets it to `undefined` immediately instead.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#max-agenumber

### 📝 Checklist


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -32,10 +32,11 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
...
-  if (opts.maxAge) {
+
+  if (opts.maxAge !== undefined) {
     delay = opts.maxAge * 1000 // convert to ms for setTimeout
   } else if (opts.expires) {
-    // getTime() already return time in ms
+    // getTime() already returns time in ms
     delay = opts.expires.getTime() - Date.now()
   }
   // use customRef if on client side otherwise use basic ref
```
danielroe, line 42: what about just checking that `delay >= 1` on L43 so we don't use a `cookieRef` in that case?

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -32,14 +32,15 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
   const cookies = readRawCookies(opts) || {}
 
   let delay: number | undefined
-  if (opts.maxAge) {
+
+  if (opts.maxAge !== undefined) {
```
danielroe, line 36: I think you're accounting for 0 here but we don't do anything with it if so, and it doesn't handle null this way.

```suggestion
  if (opts.maxAge) {
```

DamianGlowala: image

Type def disallows `null` though, do we need to handle it?

Also, the initial implementation accounted for 0 and set the value to `undefined`. Reverted to it with some minor tweaks. I kind of prefer letting the `cookieRef` handle the delay _regardless of its value_. Open to your thoughts!

DamianGlowala: The below sentence is a sole reason why I thought about making this PR:
image

danielroe: No, that should definitely be the case, but I'm not sure that `cookieRef` will do that as-is (or is needed to do that).

danielroe: Shall we instead add a handler for that so we do it directly after we initialise the ref?

DamianGlowala: Just pushed an update which I hope is closer to the desirable solution. Let me know your thoughts

danielroe (approved):

PR #24072, opened 2023-11-01
fix(vite): include more of warmup within try/catch

danielroe:
### ❓ Type of change

### 📚 Description

Currently experiencing an issue with `@nuxt/ui` where dependencies from `@nuxt/content` couldn't be resolved... This just makes that error message a little prettier.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/utils/warmup.ts
+++ b/packages/vite/src/utils/warmup.ts
@@ -36,27 +36,31 @@ export async function warmupViteServer (
   const warmedUrls = new Set<String>()
 
   const warmup = async (url: string) => {
-    url = normaliseURL(url, server.config.base)
-
-    if (warmedUrls.has(url)) { return }
-    const m = await server.moduleGraph.getModuleByUrl(url, isServer)
-    // a module that is already compiled (and can't be warmed up anyway)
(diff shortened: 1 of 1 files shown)


packages/vite/src/utils/warmup.ts
```diff
@@ -36,27 +36,31 @@ export async function warmupViteServer (
...
-    // a module that is already compiled (and can't be warmed up anyway)
-    if (m?.transformResult?.code || m?.ssrTransformResult?.code) {
-      return
-    }
-    warmedUrls.add(url)
     try {
+      url = normaliseURL(url, server.config.base)
+
+      if (warmedUrls.has(url)) { return }
+      const m = await server.moduleGraph.getModuleByUrl(url, isServer)
```
pi0, line 43: I guess this is the most likely source of issue (because other lines are mainly optional chaining). So we could simply guard it with `.catch(() => {})` to avoid increasing try/catch scope. But not sure you might want to actually chain error which in case this is right change

danielroe: L58 was the source of the issue I mentioned in the issue description, but I felt that over-logging is likely a better solution here as it will probably (safely) highlight issues with hoisting.

PR #24093, opened 2023-11-02
fix(nuxt): wrap caught asyncData error in `createError`

DamianGlowala:
### ❓ Type of change

### 📚 Description

This PR addresses a case when asyncData handler's promise gets rejected and the error isn't an instance of the `H3Error` class. For instance, when watching the error property of the data fetching composable, it is currently triggered twice in this scenario. Firstly, it has a raw error value assigned, and only then an instance of H3Error, what leads to an inconsistent behavior.

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -212,7 +212,11 @@ export function useAsyncData<
         // If this request is cancelled, resolve to the latest request.
         if ((promise as any).cancelled) { return nuxt._asyncDataPromises[key] }
 
-        asyncData.error.value = error
+        const nuxtError = createError(error)
+
+        nuxt.payload._errors[key] = nuxtError
+
+        asyncData.error.value = nuxtError as DataE
```
danielroe, line 219: do we need to do both of these assignments? `asyncData.error` should already be referencing `nuxt.payload._errors[key]` as it was created with `toRef`....

```suggestion
        asyncData.error.value = createError(error) as DataE
```

DamianGlowala: If the following destructuring doesn't break reactivity, I'll make an amendment.
https://github.com/nuxt/nuxt/blob/4930d273c9b4a4029bf6dcc51c1f3e99b65adfc9/packages/nuxt/src/app/composables/asyncData.ts#L171

danielroe: Thank you! We could add a test to ensure it.

danielroe (approved):

PR #24114, opened 2023-11-03
fix(vite): normalise entry id for `typeCheck` plugin

### LINKED ISSUE #24113: `vite-plugin-checker` overlay not currently working when `typeCheck: true`
### Environment

- Operating System: Darwin
- Node Version:     v20.9.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.0
- Package Manager:  pnpm@8.10.2
- Builder:          -
- User Config:      typescript
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-lcd9px?file=package.json,nuxt.config.ts,app.vue

### Describe the bug

In the presence of a typescript error in a nuxt project with `typeCheck: true` set in the config file, when running `nuxi dev` the error is properly displayed in the terminal through `vue-tsc`, but there is no `vite-plugin-checker` overlay visible on the development page.

### Additional context

The problem is well reproduced in `Stackblitz` and on a local project, as long as it uses the `packaged` nuxt version (js).

Please note it is not reproducible in local `playground` folder, where nuxt is built from ts `source files`.

_**Reason**:_ the statement responsible of detecting the processing of an entrypoint and of subsequently enabling `vite-plugin-checker` operates on an exact match: https://github.com/nuxt/nuxt/blob/c6fc69037d6709ae4ce83f44a7eae85acc743248/packages/vite/src/plugins/type-check.ts#L15

When operating on the `source` version, vite presents the modules with an exact `id`, whereas with a `packaged` nuxt version, vite presents the modules with a query string (for cache busting), c.f. output below after inserting `console.warn` statements.

The plugin thus needs to be updated with a compliant test statement.

### Logs

```shell-script
[10:23:43 PM]  WARN  entry: /dev/nuxt-elements/node_modules/.pnpm/nuxt@3.8.0_@types+node@20.8.10_eslint@8.52.0_typescript@5.2.2_vite@4.5.0_vue-tsc@1.8.22/node_modules/nuxt/dist/app/entry.js


[10:23:43 PM]  WARN  id: /dev/nuxt-elements/node_modules/.pnpm/nuxt@3.8.0_@types+node@20.8.10_eslint@8.52.0_typescript@5.2.2_vite@4.5.0_vue-tsc@1.8.22/node_modules/nuxt/dist/app/entry.js?v=d243727a
```

medfreeman:
### 🔗 Linked issue

resolves #24113

### ❓ Type of change

### 📚 Description

Update `type-check` vite plugin to also operate on modules `id` having a query string, allowing to match app entrypoints as designed.

Update `vite` client configuration so as to not enable the `type-check` plugin when `typeCheck: 'build'` is set, which would provoke a `module not found error` since the plugin is only added to vite configuration in `development` when  `typeCheck: true`: https://github.com/nuxt/nuxt/blob/c6fc69037d6709ae4ce83f44a7eae85acc743248/packages/vite/src/vite.ts#L133-L141

Resolves #24113

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/type-check.ts
+++ b/packages/vite/src/plugins/type-check.ts
@@ -1,7 +1,7 @@
 import MagicString from 'magic-string'
 import type { Plugin } from 'vite'
 
-const MODULE_ID_RE = /([^?]+)(?:[?].+)?/
+const QUERY_RE = /\?.+$/
 
 export function typeCheckPlugin (options: { sourcemap?: boolean } = {}): Plugin {
   let entry: string
--- a/packages/vite/src/plugins/type-check.ts
+++ b/packages/vite/src/plugins/type-check.ts
@@ -14,7 +14,7 @@ export function typeCheckPlugin (options: { sourcemap?: boolean } = {}): Plugin 
       }
     },
     transform (code, id) {
-      if (id.replace(MODULE_ID_RE, '$1') !== entry) { return }
+      if (id.replace(QUERY_RE, '') !== entry) { return }
 
       const s = new MagicString(code)
 
--- a/packages/vite/src/plugins/type-check.ts
+++ b/packages/vite/src/plugins/type-check.ts
@@ -1,6 +1,8 @@
 import MagicString from 'magic-string'
 import type { Plugin } from 'vite'
 
+const MODULE_ID_RE = /([^?]+)(?:[?].+)?/
+
 export function typeCheckPlugin (options: { sourcemap?: boolean } = {}): Plugin {
   let entry: string
   return {
@@ -12,7 +14,7 @@ export function typeCheckPlugin (options: { sourcemap?: boolean } = {}): Plugin 
       }
     },
     transform (code, id) {
-      if (id !== entry) { return }
+      if (id.replace(MODULE_ID_RE, '$1') !== entry) { return }
 
       const s = new MagicString(code)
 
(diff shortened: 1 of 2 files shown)


packages/vite/src/plugins/type-check.ts
```diff
@@ -1,6 +1,8 @@
 import MagicString from 'magic-string'
 import type { Plugin } from 'vite'
 
+const MODULE_ID_RE = /([^?]+)(?:[?].+)?/
```
danielroe, line 4:
```suggestion
const QUERY_RE = /\?.+$/
```

packages/vite/src/plugins/type-check.ts
```diff
@@ -12,7 +14,7 @@ export function typeCheckPlugin (options: { sourcemap?: boolean } = {}): Plugin
       }
     },
     transform (code, id) {
-      if (id !== entry) { return }
+      if (id.replace(MODULE_ID_RE, '$1') !== entry) { return }
```
danielroe, line 17:
```suggestion
      if (id.replace(QUERY_RE, '') !== entry) { return }
```

medfreeman:
BTW [`@histoire/plugin-nuxt`](https://github.com/histoire-dev/histoire/tree/main/packages/histoire-plugin-nuxt) will need a corresponding PR to disable `typeCheck` entirely since it conflicts with the way it's done here, and the part that works (during story collection) is quickly swallowed into oblivion.
I have one ready, i'll try to contribute it soon.

danielroe (approved):

PR #24118, opened 2023-11-04
chore: add `sass` to dependencies

wattanx:
### ❓ Type of change

### 📚 Description
Running `tsc` on a project that does not have `sass` installed will result in an error.
I don't want to get type errors if `sass` is not installed.

```log
node_modules/@nuxt/types/lib/sass-loader/interfaces.d.ts:1:23 - error TS2307: Cannot find module 'sass' or its corresponding type declarations.

1 import * as Sass from 'sass'
                        ~~~~~~


Found 1 error in node_modules/@nuxt/types/lib/sass-loader/interfaces.d.ts:1
```

I wanted to solve this problem without adding a package, but it was difficult.

### 📝 Checklist

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -28,7 +28,8 @@
     "@types/terser-webpack-plugin": "4.2.1",
     "@types/webpack": "^4.41.35",
     "@types/webpack-bundle-analyzer": "3.9.5",
-    "@types/webpack-hot-middleware": "2.25.5"
+    "@types/webpack-hot-middleware": "2.25.5",
+    "sass": "1.69.5"
   },
   "engines": {
     "node": "^14.18.0 || >=16.10.0"


danielroe:
The issue is that this will increase the install size of Nuxt, even if users don't use `sass`.

pi0:
Shall we bundle/inline sass types?

danielroe:
That's probably a good solution, though it does mean we don't respect any given version the user installs themselves.

It's also an issue we face with `@nuxt/schema` in 3.x - we depend on things like webpack or other deps that might not be installed by users, meaning we require `skipLibCheck` to be set. 🤔

packages/types/package.json
```diff
@@ -28,7 +28,8 @@
     "@types/terser-webpack-plugin": "4.2.1",
     "@types/webpack": "^4.41.35",
     "@types/webpack-bundle-analyzer": "3.9.5",
-    "@types/webpack-hot-middleware": "2.25.5"
+    "@types/webpack-hot-middleware": "2.25.5",
+    "sass": "1.69.5"
```
rchl, line 32: This change will most likely not be merged so my comment is not important but I don't want Nuxt pinning this to specific version as then it will cause duplicated packages and maybe even type errors when I use newer version in my project.

rchl:
Dealing with issues like that in my projects over the years, I've accepted that I just have to install those dependencies in my project if I have types to works without `skipLibCheck`. And I think that's much better than libraries like `nuxt` having those as explicit dependencies. Of course only as long as the dependency is only required for types to pass and not during runtime.

PR #24127, opened 2023-11-05
feat(nuxt): allow plugins to specify dependencies

### LINKED ISSUE #24123: add `dependsOn` property for nuxt plugins
### Describe the feature

this feature would allow to make a plugin await hte end of the execution of another plugin before running it.

This can be a good improvement in multiple ways.

**encourage modules authors to set their plugins as parallel**
- In many case, using composition api implies that we'll expect to have some plugins to be ready before using it; So setting plugins as paarallel for module authors bring a sort of incertitude (for its users) of using a composition function which relies on a plugin (such as i18n)

**allow users to have a plugin which relies on multiple parallel plugins**  

**basically encourage users to set every plugins as parallel**

**reduce TTI**

### Additional information

### Final checks

huang-julien:
### 🔗 Linked issue
resolve #24123 

### ❓ Type of change

### 📚 Description

Hello ! 👋 
This PR adds a `dependsOn` field for plugins. It is useful when a plugin need to wait for a parallel plugin before being run. This doesn't change the execution order.
Plugins can also depends on another plugin that is executed after in the array of plugin.

`dependsOn` is also typed

### 📝 Checklist

- [x] test


ineshbose (approved):
This is so great - plugin maps to the rescue! LGTM 🙂

test/fixtures/basic/plugins/dependsOnPlugin.ts
```diff
@@ -0,0 +1,8 @@
+export default defineNuxtPlugin({
+  name: 'depends-on-plugin',
+  dependsOn: ['async-plugin'],
+  async setup () {
+    await new Promise(resolve => setTimeout(resolve, 100))
```
danielroe, line 5: Should we throw an error if `$asyncPlugin` hasn't been injected?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -331,26 +339,44 @@ export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin & ObjectPlug
...
   const errors: Error[] = []
+
   for (const plugin of plugins) {
     if (import.meta.server && nuxtApp.ssrContext?.islandContext && plugin.env?.islands === false) { continue }
-    const promise = applyPlugin(nuxtApp, plugin)
+    const promise = plugin.dependsOn
+      ? Promise.all(plugin.dependsOn.map((name) => {
+        const pluginPromise = pluginMap[name]
+        if (import.meta.dev && !pluginPromise) {
+          console.warn(`[nuxt] ${plugin._name ? `Plugin "${plugin._name}"` : 'A plugin'} depends on "${name}" but "${name}" plugin does not exist or hasn't been registered yet.`)
```
danielroe, line 352: Should we avoid this possibility entirely? Perhaps we eagerly register a map with callbacks or similar? Or otherwise we are dependent on only depending on plugins alphabetically prior. (Or if not, we might advise here to use `order`.)

huang-julien:
@danielroe i've refactored this so we won't have to worry about plugins order anymore, order is still here but we can declare a plugin as dependant of another even if it before the dependant plugin in the array of plugins.

Hope this is ok 🤔

packages/nuxt/src/app/nuxt.ts
```diff
@@ -190,6 +190,14 @@ export interface ObjectPlugin<Injections extends Record<string, unknown> = Recor
    * @default false
    */
   parallel?: boolean
+  /**
+   * Await for plugin to be finished before running this plugin.
+   */
+  dependsOn?: string[]
```
danielroe, line 196: Would be nice to provide strong types for this, as we do for layout names. (Can be a future pr.)

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -113,6 +115,20 @@ export default defineNuxtPlugin({
 })
 ```
 
+### Plugin depending on another plugin
+
+If a plugin needs to await a parallel plugin before being ran. You can add the plugin's name to the `dependsOn` array.
```
danielroe, line 120:
```suggestion
If a plugin needs to await a parallel plugin before being run, you can add the plugin's name to the `dependsOn` array.
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -365,7 +419,7 @@ export function isNuxtPlugin (plugin: unknown) {
  * @param nuxt A Nuxt instance
  * @param setup The function to call
  */
-export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
+export function callWithNuxt<T extends (...args: any[]) => any>(nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
```
danielroe, line 422:
```suggestion
export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -331,26 +339,72 @@ export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin & ObjectPlug
...
 }
 
 /*@__NO_SIDE_EFFECTS__*/
 export function defineNuxtPlugin<T extends Record<string, unknown>> (plugin: Plugin<T> | ObjectPlugin<T>): Plugin<T> & ObjectPlugin<T> {
   if (typeof plugin === 'function') { return plugin }
+
+  const _name = plugin._name || plugin.name
   delete plugin.name
-  return Object.assign(plugin.setup || (() => {}), plugin, { [NuxtPluginIndicator]: true } as const)
+  return Object.assign(plugin.setup || (() => { }), plugin, { [NuxtPluginIndicator]: true, _name } as const)
```
danielroe, line 407:
```suggestion
  return Object.assign(plugin.setup || (() => {}), plugin, { [NuxtPluginIndicator]: true, _name } as const)
```

test/bundle.test.ts
```diff
@@ -71,7 +71,8 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
     const serverDir = join(rootDir, '.output-inline/server')
 
     const serverStats = await analyzeSizes(['**/*.mjs', '!node_modules'], serverDir)
-    expect.soft(roundToKilobytes(serverStats.totalBytes)).toMatchInlineSnapshot('"509k"')
+
```
danielroe, line 74:
```suggestion
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -331,26 +339,72 @@ export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin & ObjectPlug
...
   const errors: Error[] = []
-  for (const plugin of plugins) {
-    if (import.meta.server && nuxtApp.ssrContext?.islandContext && plugin.env?.islands === false) { continue }
-    const promise = applyPlugin(nuxtApp, plugin)
-    if (plugin.parallel) {
-      parallels.push(promise.catch(e => errors.push(e)))
+  let promiseDepth = 0
+
+  // todo move it to build time
+  plugins.forEach((plugin) => {
```
danielroe, line 350:
```suggestion
  for (const plugin of plugins) {
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -331,26 +339,72 @@ export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin & ObjectPlug
...
+    if (plugin.dependsOn && plugin.dependsOn.some(name => !pluginNames.includes(name))) {
+      if (import.meta.dev) {
+        console.warn(`Plugin ${plugin._name} depends on ${plugin.dependsOn.filter(name => !pluginNames.includes(name)).join(', ')} but they are not registered.`)
+      }
+      plugin.dependsOn = plugin.dependsOn.filter(name => pluginNames.includes(name))
+    }
+  })
```
danielroe, line 357:
```suggestion
  }
```

packages/nuxt/src/app/nuxt.ts
huang-julien: done !

danielroe (approved):

danielroe:
Any thoughts @pi0? I see you requested a review.

ineshbose (approved):

pi0:
I have pushed a new test + vary basic check for circular dependency in plugins.

pi0:
Some notes:
- The circular check algorithm is not the most efficient but bundles friendly and I hope to be moved to build-time with static syntax analyzing
- For consistency with rollup plugins, h3 events, etc we could also use `after: []` syntax it is also more open to the future development of including `before: []` but it is my personal opinion, if `dependsOn` is more explicit that's fine
- I wouldn't personally use this kind of hard dependency graph in plugins. Hooks are much more flexible in terms of dependency definition and pretty sure every situation is solvable with correct patterns. Saying I see it as mainly an "escape hatch" for authors looks good addition considering historically it was always a requested feature.

pi0 (approved):

PR #24173, opened 2023-11-07
fix(vite): provide default values for `transformAssetUrls`

### LINKED ISSUE #24038: custom transformAssetUrls removes the default? Intended behavior?
### Environment

- Operating System: Linux
- Node Version:     v20.7.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  pnpm@8.6.5
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/bootstrap-vue-next/bootstrap-vue-next/blob/main/packages/nuxt/src/module.ts#L28
& 
https://github.com/bootstrap-vue-next/bootstrap-vue-next/tree/main/packages/nuxt/playground

### Describe the bug

This could be intended behavior, but adding a custom transformAssetUrls option removes the defaults. 

For example if I do 
```ts
  vite: {
    vue: {
      template: {
        transformAssetUrls: {
          BImg: ['src'],
        },
      },
    },
  },
```
Suddenly `img: ['src']` resolution disappears. 

The simple solution is to just add it back in manually. But this is a bit strange. If transformAssetUrls ever has more defaults, then suddenly they would break too. 

This is a double edged sword, as the question is asked "What if you don't want to transform those urls?". If the default is merged, then you wouldn't be able to override that behavior. But in this context I don't know why you wouldn't want the defaults.

Maybe this is Vites fault, idk. 

### Additional context

_No response_

### Logs

_No response_

darioferderber:
### ❓ Type of change

### 📚 Description

Resolves #24038
### 📝 Checklist


packages/vite/src/vite.ts
```diff
@@ -203,3 +203,17 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
...
+  config.vue = vite.mergeConfig(config.vue || {}, {
+    template: {
+      transformAssetUrls: {
+        video: ['src', 'poster'],
+        source: ['src'],
+        img: ['src'],
+        image: ['xlink:href', 'href'],
+        use: ['xlink:href', 'href']
+      }
+    }
```
danielroe, line 217: how about we add these here instead?

https://github.com/nuxt/nuxt/blob/4f017a5538d1c17864fca5723c5021f13cd9723f/packages/vite/src/vite.ts#L113

darioferderber: sure, it makes more sense

packages/vite/src/server.ts
```diff
@@ -5,7 +5,7 @@ import viteJsxPlugin from '@vitejs/plugin-vue-jsx'
 import { logger, resolvePath, tryResolveModule } from '@nuxt/kit'
 import { joinURL, withTrailingSlash, withoutLeadingSlash } from 'ufo'
 import type { ViteConfig } from '@nuxt/schema'
-import type { ViteBuildContext } from './vite'
+import { type ViteBuildContext } from './vite'
```
danielroe, line 8:
```suggestion
import type { ViteBuildContext } from './vite'
```

danielroe (approved):

PR #24182, opened 2023-11-08
perf(vite): tree-shake `typeof process`

undermoonn:
### ❓ Type of change

### 📝 Checklist


danielroe:
If we add this we should probably also do the same for webpack builder.

https://github.com/nuxt/nuxt/blob/a5dad0d2e83323abc2bed7cfe3541f4dbd495335/packages/webpack/src/presets/base.ts#L234-L237

Not sure whether this would cause any unintended consequences in the ecosystem, or why we might not have done this before. (cc: @pi0)

undermoonn:
In the vite dev server, all the defines will be injected into window. The `'typeof process'` will become the gobal variable which seems wired. It shoube the value `undefined`, not the string value `'undefined'`.

Maybe it's the reason nuxt team not have done this before :)

pi0:
> Not sure whether this would cause any unintended consequences in the ecosystem, or why we might not have done this before. (cc: @pi0)

As far as i remember, @manniL was working on Nuxt 2 version. I think too it is safe to enable by default.

manniL:
Yeah, I had a feeling when I looked into the PR 😂
Let's give it a try 👍🏻

danielroe:
Since we last discussed, Vite has moved away from direct string replacements (which would mean we need to use rollup/plugin replace for this, with the disadvantages that it might be less performant and potentially replace content of strings rather than just code).

It might be worth investigating to see if we can use `define` to handle `process` tree-shaking. (But note that this is more complex as we (and others) use `process.env.SOME_VAR` - which we should likely migrate away from within Vue app code anyway...)

See also https://github.com/nuxt/nuxt/pull/24711.

PR #24231, opened 2023-11-10
chore: enable `noUncheckedIndexedAccess`

DamianGlowala:
### 📚 Description

This PR adds `noUncheckedIndexedAccess` to tsconfig.json.


danielroe:
This is a great initiative - I think it would be worth checking through and seeing whether there are any issues flagged up by this change, and gradually refactoring the code to improve safety ...

However, that will be a fair amount of work!

PR #24253, opened 2023-11-11
fix(nuxt): use max length + iterations for `useCookie` timeout

### LINKED ISSUE #24227: `useCookie` with `maxAge` or `expires` gets removed when updated
### Environment

- Operating System: Darwin
- Node Version:     v18.17.1
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  pnpm@8.10.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

I used the starter project from `pnpx nuxi@latest init` and just replaced the `app.vue` with the following code:

```vue
<script setup lang="ts">
const hasConsent = useCookie('dialog', {
  sameSite: 'strict',
  default: () => 'no',
  maxAge: 2592000, // 30 days in seconds
});
const showConsentAlert = ref(false);

onMounted(() => {
  if (process.server) return;
  if (hasConsent.value === 'yes') return;

  showConsentAlert.value = true;
});

function acceptConsent() {
  hasConsent.value = 'yes';
  showConsentAlert.value = false;
}
</script>

<template>
  <Teleport to="body">
    <div v-if="showConsentAlert">
      <span>We use cookies</span>
      <button type="button" @click="acceptConsent">Accept</button>
    </div>
  </Teleport>
</template>
```

### Describe the bug

When I click the `acceptConsent` button, I set the cookie value from `no` to `yes`. However, instead of saving the `yes` value inside the cookie, it gets removed completely. I tried on Firefox v119.0.1 and Chrome v119.0.6. Both browsers have the exact same behavior.

The strange part is that if I remove the `maxAge` option from `useCookie`, everything works fine. But I would like that the consent cookie expires after 30 days. I also tried with `expires` and the same issue happened.

I'm not sure if I'm doing something wrong or if it really is a bug, but I would expect that if I change the value of the cookie, it must be persisted, but it is getting deleted...

### Additional context

I'm trying to create a cookie consent dialog that shows once and if it gets accepted, it doesn't show again for 30 days. My reproduction code is just a bare-bones demo with the minimal code I could reproduce it with. In the real website this functionality is inside a `CookieConsent.vue` component and it is a Modal.

https://github.com/nuxt/nuxt/assets/47041342/f6146867-31a0-468e-98c7-cb3536c797ba

### Logs

_No response_

ChrisGV04:
### 🔗 Linked issue

Resolves #24227 

### ❓ Type of change

### 📚 Description

This PR fixes an issue introduced when implementing the functionality to make expired cookies `undefined` on the client-side. Since it uses `setTimeout` to apply the delay to expire the cookie, when you pass a long value to `maxAge` or `expires`, it deletes the cookie immediately instead of waiting that time.

This happens because of the maximum delay value allowed by browsers on `setTimeout` [as stated in MDN](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value) and that [@positiveprogrammer](https://github.com/nuxt/nuxt/issues/24227#issuecomment-1806614031) brought to my attention. The issue started happening on v3.8.1 with PR #23549. 

In my case, I was setting a cookie that stores if the user consents to using cookies. That cookie choice should last 30 days, which is above the 24.8 day limit of `setTimeout` so if I updated the value client-side, the cookie instantly became `undefined`.

My proposed solution to work around this issue (since it's a browser limitation unrelated from Nuxt) is to implement extra logic to divide the task into smaller intervals that add up to the total expected delay. That way we never exceed the timeout limit and still wait the desired amount of time before expiring the cookie.

> For example: If the maximum delay value was 5 seconds and you want a cookie with `maxAge` of 12 seconds, we are going to need 2 intervals of 5 seconds (the max value) and a timeout for the remaining 2 seconds. That way we add up to 12 seconds as expected (5 + 5 + 2 = 12) without exceeding the browser limit of 5 seconds of this example.

I'm not sure if this is the best possible solution to the problem, but it seems to work.

### 📝 Checklist


huang-julien (commented):
we could probably wrap this with `import.meta.client`.

ChrisGV04:
Hi @huang-julien! How do you mean? Wrap the the whole `if (intervalsNeeder)`?

danielroe:
(I think we probably don't need to wrap it again as we should only use the `cookieRef` on the client-side anyway.)

I think we can simplify this slightly. Maybe just keep track of a single timeElapsed number, and use a constant for setting a timeout?

ChrisGV04:
Hi @danielroe! Sounds good.

However, I took a closer look and I think there's a fundamental flaw in this functionality to set expired cookies to `undefined` on the client overall, since we always use the original `maxAge` and `expires` to set the timeout, not the actual time that the cookie expires at.

For example, if I set the cookie to expire 5 days from now, when I use the app on day 5 and read the cookie client-side, the timeout is being set to the original 5 days, but it actually only has a few hours left.

If I understand correctly, this is also a browser limitation since there is no easy way to get the actual `expires` date of the cookie on the browser. I don't know if there's a smarter solution to all of this

ChrisGV04:
Hi @danielroe @huang-julien! I know you must be very busy.

What do you think about my last comment about the flaw with expiring the cookie client-side? Maybe you understand the subject better than I do.

danielroe:
Pushed a refactor. Let me know what you think.

You are right about that potential issue; the behaviour is _safe_ in that it shouldn't expire a cookie too early, but the cookie can remain accessible via a `useCookie` called _before_ the expiration after it actually expires.

Ideas welcome to resolve, but we can also defer that to a subsequent PR.

ChrisGV04:
Hi @danielroe!

I believe this solves the current issue at hand, so it's good for now. Thank you very much!

You are right about the other issue, at least it won't expire early. I haven't been able to find a simple solution that doesn't involve storing the expiration date for each cookie.

Since it seems to be impossible to know that real expiration value, we either have to create an additional cookie that stores the expiration for every cookie created with `useCookie` or use a JWT to encode the value and the expiration inside the same cookie, but that would break using the cookie with other external services.

danielroe (approved):

PR #24298, opened 2023-11-15
Update server documentation: auto import/exports caveats

PrimeTimeTran:
### ❓ Type of change

### 📚 Description

Add additional context on what is auto imported/exported from `./server` so that future developers know the caveats of auto import/export

Screenshot 2023-11-15 at 10 42 58 PM

I see builds failing in CI/CD when I change casing from `./server/models/models.model.ts` to `./server/Models/models.model.ts`

Please let me know if this shouldn't be the case.

### 📝 Checklist

### DIFF
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -502,15 +502,14 @@ export default defineNuxtConfig({
 
 Not every directory in `./server` is imported into other files.
 
-`User` is exported from this file automatically.
+`Wizard` is exported from this file automatically.
 
-```ts [./server/models/user.model.ts]
-export const User = ''
+```ts [./server/apis/wizards.get.ts]
+export const Wizard = ''
 ```
 
-`User` is not exported from this file automatically.
+`Wizard` is not exported from this file automatically.
 
-```ts [./server/Models/user.model.ts]
-export const User = ''
+```ts [./server/apis/wizards.get.ts]
+export const Wizard = ''
 ```
-
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -513,3 +513,27 @@ export const Wizard = ''
 ```ts [./server/Models/wizards.model.ts]
 export const Wizard = ''
 ```
+
+### Type suggestions don't work without additional exports from utils.
+
+Types defined in `./server/models/wizard.model.ts` must be exported
+
+```ts [./server/models/wizards.model.ts]
+export type WizardType = {
+  firstName: String
+  lastName: String
+}
+```
+
+Then imported & exported from `./utils/wizard.ts` before they can be used inside of frontend code.
+
+```ts [./utils/wizard.ts]
+export type { WizardType } from '~/server/Models/wizard.model'
+```
+
+But they must also include a const export to get type suggestions.
+
+```ts [./utils/wizard.ts]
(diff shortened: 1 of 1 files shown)


docs/2.guide/2.directory-structure/1.server.md
```diff
@@ -494,3 +494,22 @@ export default defineNuxtConfig({
 })
 ```
 ::
+
+
+
+
+### Cavets on Auto import/export
```
huang-julien, line 501:
```suggestion
### Caveats on Auto import/export
```

docs/2.guide/2.directory-structure/1.server.md
```diff
@@ -494,3 +494,22 @@ export default defineNuxtConfig({
...
+
+Not every directory in `./server` is imported into other files.
+
+`Wizard` is exported from this file automatically.
+
+```ts [./server/apis/wizards.get.ts]
```
huang-julien, line 507: This is the same name as the one bellow, did you meant utils ?

huang-julien (changes requested):
I don't think exported variables from api files are auto-imported in other files

danielroe (commented):
I'm not sure this is needed. We already have this explanation: https://nuxt.com/docs/guide/directory-structure/server#server-utilities.

Maybe we could add another sentence to make it clearer?

docs/2.guide/2.directory-structure/1.server.md
PrimeTimeTran: My bad, I meant to specifically point out that Models & models are different/case sensitive from my observation so far.

PrimeTimeTran:
> I'm not sure this is needed. We already have this explanation: https://nuxt.com/docs/guide/directory-structure/server#server-utilities.
> 
> Maybe we could add another sentence to make it clearer?

I updated the example & also added another interesting thing I discovered.

https://s5.gifyu.com/images/SRHNy.gif

Type suggestions work when a file exports an additional const. Not sure if this is the expected behavior but it caught me off guard and threw me off.

PrimeTimeTran:
@danielroe please let me know if I've been unclear. I'll be more than happy to update to land this.

My thinking: 
- Auto import/export is awesome but isn't defined from what I've found so far.  
- Type definitions in one repo for both front and backend solves a big problem but not getting autocomplete/suggestions due to a missing const export seems strange to me.

Please correct me if I'm wrong. I'm hoping to save future users time wondering why they're not getting them(type suggestions) and their local auto imports work but fail in CI/CD.

I'm on `"nuxt": "^3.8.0",` in case that makes a difference.

danielroe:
I think type imports failing in CI is likely due to a failure to run `nuxi prepare` which generates the `tsconfig.json`.

We should specify that files in `server/utils/` are scanned for auto-imports, [here](https://nuxt.com/docs/guide/directory-structure/server#server-utilities).

danielroe:
Marking as draft just to help me triage PRs better... Feel free to unmark whenever you're ready for me to have a look.

danielroe:
I think as written, these aren't suggestions we'd want to make.

- casing of models/Models shouldn't make a difference to _nuxt_ as we don't do anything special with that directory. But if you are importing from it, you will encounter issues when developing on a case-insensitive OS when you then build in CI. But this isn't related to auto imports.
- As for auto-import of types, this isn't yet supported. Likely this is working for you by accident and we wouldn't want to document the current solution.

Thank you for taking the time to create the PR however, and please do feel free to create another one any time ❤️

PR #24306, opened 2023-11-15
feat(kit): support `prepend` option for `addComponentsDir`

sadeghbarati:
### ❓ Type of change

### 📚 Description

Default Nuxt config for auto-import directories

```
[
  { priority: 1, path: 'D:/Test/nuxt-test/components/islands', island: true },
  { priority: 1, path: 'D:/Test/nuxt-test/components/global', global: true },
  { priority: 1, path: 'D:/Test/nuxt-test/components' },
]
```

In Nuxt [website](https://nuxt.com/docs/guide/directory-structure/components#:~:text=You%20can%20also%20selectively%20register%20some%20components%20globally%20by%20placing%20them%20in%20a%20~/components/global%20directory.&text=You%20can%20also%20selectively%20register%20some%20components%20globally%20by%20placing%20them%20in%20a%20~/components/global%20directory.&text=You%20can%20also%20selectively%20register%20some%20components%20globally%20by%20placing%20them%20in%20a%20~/components/global%20directory.) it's says

> You can also selectively register some components globally by placing them in a `~/components/global` directory.

But if I add `components.dirs[]` options in `nuxt.config.ts` default config will disposed, and in order to have **global** directory I have to define it again

```
components: {
    dirs: [
        {
          path: '~/components/ui',
          prefix: 'Ui',
          pathPrefix: false
        }
    ]
}
```

I want to add a new dir to auto-import directories, to keep the default auto-import directories I prefer using

- `"components:dirs"` hooks (nuxt.config and defineNuxtModule)
- `addComponentsDir` in setup function (defineNuxtModule)

In the hook way, I can push/unshift directories but in `addComponentsDir` way it only has the `push` method

### 📝 Checklist

### DIFF
--- a/packages/kit/src/components.ts
+++ b/packages/kit/src/components.ts
@@ -9,12 +9,12 @@ import { logger } from './logger'
  *
  * Requires Nuxt 2.13+
  */
-export async function addComponentsDir (dir: ComponentsDir) {
+export async function addComponentsDir (dir: ComponentsDir, opts: { prepend?: boolean } = {}) {
   const nuxt = useNuxt()
   await assertNuxtCompatibility({ nuxt: '>=2.13' }, nuxt)
   nuxt.options.components = nuxt.options.components || []
   dir.priority ||= 0
-  nuxt.hook('components:dirs', (dirs) => { dirs.push(dir) })
+  nuxt.hook('components:dirs', (dirs) => { dirs[opts.prepend ? 'unshift' : 'push'](dir) })
 }
 
 export type AddComponentOptions = { name: string, filePath: string } & Partial<Exclude<Component,


sadeghbarati:
I'll share a StackBlitz 🙌

packages/kit/src/components.ts
```diff
@@ -9,12 +9,12 @@ import { logger } from './logger'
  *
  * Requires Nuxt 2.13+
  */
-export async function addComponentsDir (dir: ComponentsDir) {
+export async function addComponentsDir (dir: ComponentsDir, insertMethod: 'push' | 'unshift' = 'push') {
```
danielroe, line 12: Instead, can we support an options object with `prepend` key, in line with other kit utilities?

https://github.com/nuxt/nuxt/blob/1444d89e3b6170eaed56c16494b1ff6b9b68ef9e/packages/kit/src/imports.ts#L14

PR #24307, opened 2023-11-15
fix(nuxt): deeply unwrap headers/query for useFetch key

### LINKED ISSUE #24294: `useFetch` with `computed` query cause hydration mismatch
### Environment

- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  bun@1.0.6
- Builder:          -
- User Config:      devtools, modules, app, runtimeConfig, svgo, image
- Runtime Modules:  @nuxtjs/tailwindcss@6.9.4, nuxt-svgo@3.6.0, @nuxt/image@1.0.0, nuxt-lazy-hydrate@1.0.0
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/sandbox/tender-kalam-244yw7-244yw7?file=%2Fpages%2Findex.vue%3A1%2C1

### Describe the bug

Using a computed as query/param causes hydration mismatch and I noticed it seems to be because upon checking, the hash key is different between server and browser.

As per the documentation:

> All fetch options can be given a computed or ref value. These will be watched and new requests made automatically with any new values if they are updated.

I used this over the `watch` option.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #24300: useFetch data is null in some case depending on the query param type.
### Environment

Nuxt project info:                                                                                                                                            12:16:19

- Node Version:     v18.12.1
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  pnpm@8.6.12
- Builder:          -
- User Config:      devtools, nitro
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-k5ugr3?file=pages%2Fabout.vue

### Describe the bug

two ways to give a param to useFetch

1. give a pure string
2. give a computed vary

two way to go into "about" page
1. click about link in home
2. in about page and refresh to reload

when get about page from other page by route, both useFetch get correct data

when refresh the about page, a pure string param will get a real data immidiate, but a ref make it null.

both useFetch will get data finally if just render in template or shown somewhere, But if the code below rely on the data to make a special logic, it need to always need a watch "pending" to run the logic later, it's a bug or a design?

this will get null data in client console
```
const { data } = await useFetch('/api/user', { query: { param1: id } });
```
this get the correct data
```
const { data } = await useFetch('/api/user', { query: { param1: id.value } });
```

// null or [1,2,3]
// the log in ssr server console is always right.
console.log('data', data.value);

### Additional context

No, just run the reproduction

### Logs

```shell-script
server logs is correct. client logs get null
```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24294
resolves https://github.com/nuxt/nuxt/issues/24300

### ❓ Type of change

### 📚 Description

This makes sure we deeply unwrap headers/params objects when computing hash for automatic fetch key so we don't trigger a mismatch on client-side.

### 📝 Checklist


pi0 (approved):
LGTM. Was thinking if we could use any faster method like JSON serialization but seems safe and fast with manual iteration as i expect these options are also small in size

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -205,3 +205,21 @@ export function useLazyFetch<
...
+  const segments: Array<string | undefined | Record<string, string>> = [
+    toValue(opts.method as MaybeRef<string | undefined> | undefined)?.toUpperCase() || 'GET',
+    toValue(opts.baseURL),
+  ]
+  for (const _obj of [opts.params || opts.query, opts.headers]) {
+    const obj = toValue(_obj)
+    if (obj) {
+      const unwrapped: Record<string, string> = {}
+      for (const key in obj) {
+        unwrapped[key] = toValue(obj[key as keyof typeof obj])
```
pi0, line 219: Have you checked iteration node worngly goes through objects/class instances with a custom constructor such as `Headers`?

danielroe: Nope, good point 👍

xxRockOnxx:
wow that was fast :fire:

PR #24321, opened 2023-11-15
docs: warn about using `useRequestHeaders` in `useFetch`

### LINKED ISSUE #23900: useRequestHeaders and useFetch are incompatible in 3.8.0
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.0
- CLI Version:      3.9.1
- Nitro Version:    2.7.0
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-mgk9r8

### Describe the bug

#23462 introduced `headers` option value as one of the hash components for useFetch requests. This however broke the example of supplying `headers: useRequestHeaders()` as shown in [docs](https://nuxt.com/docs/api/composables/use-request-headers#example), because on server side and client side the supplied value are different (on client side it will always return an empty object). This results in a cache miss and ~~triggers a re-fetch~~ `data` being `null` on hydration.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23900

### ❓ Type of change

### 📚 Description

This updates documentation to avoid using `useRequestHeaders` in `useFetch` calls, following https://github.com/nuxt/nuxt/pull/23462.

We could also consider reverting that change.

Thoughts welcome. 🙏 

### 📝 Checklist

### DIFF
--- a/docs/1.getting-started/6.data-fetching.md
+++ b/docs/1.getting-started/6.data-fetching.md
@@ -371,9 +371,13 @@ The example below adds the request headers to an isomorphic `$fetch` call to ens
 <script setup lang="ts">
 const headers = useRequestHeaders(['cookie'])
 
-const { data } = await useFetch('/api/me', { headers })
+const { data } = await useAsyncData(() => $fetch('/api/me', { headers }))
 </script>
 ```
+
+::callout
+We are not using `useFetch` here as `headers` would then form part of the automatically generated key, causing this result to mismatch between client and server.
+::
 
 ::callout
 Be very careful before proxying headers to an external API and just include headers that you need. Not all headers are safe to be bypassed and might introduce unwanted behavior. Here is a list of common headers that are NOT to be proxied:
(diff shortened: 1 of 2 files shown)


manniL:
> We could also consider reverting that change.

I think that'd be a bit better as it seems like a potential footgun 🤔
The problem is that this would be a breaking change,no ?

webfansplz:
> We could also consider reverting that change.

I agree that we should revert the changes. IMO, [#23462 ](https://github.com/nuxt/nuxt/pull/23462)'s case is even rarer than between the client and server have different request headers. For this case, we recommend that users add a `key` manually to avoid issues.

danielroe:
I think that's reasonable. Let's go ahead and revert adding headers to the auto-generated key. PR welcome 🙏

PR #24322, opened 2023-11-15
feat(nuxt): add payload control to `NuxtIsland` with `useCache` and `setCache`

huang-julien:
### ❓ Type of change

### 📚 Description

Hi :wave: 
This PR gives to the user a way to control `NuxtIsland`'s behavior at a granular level since it uses the app context to share the cache with other islands.

### 📝 Checklist


PedroAugustoRibas (approved):

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -64,6 +64,14 @@ export default defineComponent({
       type: String,
       default: () => undefined
     },
+    /**
+     * use the NuxtIslandResponse which has been cached if available
+     * @default true
+     */
+    useCache: {
```
manniL, line 71: Maybe just me but "useCache" sounds like a composable on the first read :D

huang-julien:
After some discussions, this would probably not be used by many people so i'm going to close it. We probably will re-open it once people ask about this feature

Rigo-m:
Is it possible to cache islands using routeRules instead?

huang-julien:
This is only for  NuxtIsland to always fetch the  island component instead of reusing a previous result

PR #24333, opened 2023-11-16
fix(nuxt): do not compute `useFetch` key from headers (reverting #23462)

### LINKED ISSUE #23900: useRequestHeaders and useFetch are incompatible in 3.8.0

webfansplz:
### 🔗 Linked issue

resolves #23900 #24321

Be related to #23462

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -211,12 +211,12 @@ function generateOptionSegments <_ResT, DataT, DefaultT>(opts: UseFetchOptions<_
     toValue(opts.method as MaybeRef<string | undefined> | undefined)?.toUpperCase() || 'GET',
     toValue(opts.baseURL),
   ]
-  for (const _obj of [opts.params || opts.query, opts.headers]) {
+  for (const _obj of [opts.params || opts.query]) {
     const obj = toValue(_obj)
     if (!obj) { continue }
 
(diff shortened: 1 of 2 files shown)


danielroe (commented):
We'll need to update the tests as well - we can change them from headers -> query param tests.

https://github.com/nuxt/nuxt/blob/562532778bb92418263e2e79e2ff8774e9069646/test/nuxt/composables.test.ts#L259-L273

danielroe (approved):
Thank you!

PR #24369, opened 2023-11-19
feat(nuxt): add useRouteQuery and useRouteParams

### LINKED ISSUE #24365: useQuery and useParams composables
### Describe the feature

useQuery should return all route queries.
useParams should all return all route params.

Currently we have to first call the useRoute composable before getting the route queries and or route params.

### Additional information

### Final checks

saidinusah:
### 🔗 Linked issue
#24365

### ❓ Type of change

### 📚 Description

 I've added useRouteQuery and useRouteParams so we can easily get the route query and params respectively

It prevents us from doing useRoute?.query or useRoute?.params so many times in our code base. Instead with useRouteQuery returns the route queries if there are any and undefined if none.
useRouteParams also returns the route params if there are any and undefined in none

Resolves #24365

### 📝 Checklist


DamianGlowala:
Could you please lint your work with `pnpm lint:fix` (notice many changes in indentation, added semicolons, etc.)?

packages/nuxt/src/app/composables/index.ts
```diff
@@ -1,36 +1,76 @@
...
 export {
   /** @deprecated Import `useHead` from `#imports` instead. This may be removed in a future minor version. */
   useHead,
   /** @deprecated Import `useSeoMeta` from `#imports` instead. This may be removed in a future minor version. */
   useSeoMeta,
   /** @deprecated Import `useServerSeoMeta` from `#imports` instead. This may be removed in a future minor version. */
-  useServerSeoMeta
-} from '@unhead/vue'
+  useServerSeoMeta,
+} from "@unhead/vue";
```
manniL, line 13: Please remove auto-formatting changes (such as this one) from the PR ☺️

TheAlexLichter (changes requested):

packages/nuxt/src/app/composables/index.ts
DamianGlowala: Mentioned precisely that just above 😄

1nusah: pnpm lint:fix doesn't fix it so i'm reverting the changes and working on that again. this time with prettier turned off

1nusah: should be good now

danielroe (commented):
I need to think about whether this adds enough DX improvement, as it's a pretty thin wrapper. 

In addition, both of these destructure the params/queries from the reactive route object so they are then non-reactive. (If you call them and then transition route the values will be incorrect.) We could fix that by using a computed value.

DamianGlowala:
As much as the efforts to provide this PR are appreciated, I'd be personally reluctant to merge these for the following reasons:
- I agree these are really thin wrappers (`useRoute().params` vs `useRouteParams()`), as @danielroe mentioned above
- these two can be added with minimal effort to the `composables/` directory in the users' projects (if they believe these add significant value), as mentioned by @manniL
- some might argue that if we provide these two, why not have `useRouteFullPath()`/`useRoutePath()`/`useRouteName()`, etc.

... and instead recommend to the author to add these two in their own projects.

pi0:
Agree with all the points. It _might_ be useful if we support utilities to access a single reactive param/query by name though like `useRouteQuery(name: string): Ref<string>` since it will **solve** a common mistake of losing reactivity by destructing but probably for top level props of `useRoute().*` it is overkill to have core utils for each or some.

danielroe:
@1nusah apologies for your work on this. I think they are good composables but probably shouldn't be built-in to Nuxt in this case.

PR #24376, opened 2023-11-20
fix(nuxt): handle errors when booting app with `app:error`

antfu:
Without a custom `errorHandler`, Vue will [throw an error and crash itself](https://github.com/vuejs/core/blob/32fb79df41f8c3f5de8ca01545a4dfeb26326709/packages/runtime-core/src/errorHandling.ts#L157-L162) in dev, causing Vue's singleton context has a broken state. In that case, it will cause off-root apps (another `createApp`) to fail to work. Nuxt DevTools creates a separate app to mount itself, and this PR fixes that DevTools does not work on the error page.

This probably also improves the error capture in Production? As in production, [unhandled error are not thrown](https://github.com/vuejs/core/blob/32fb79df41f8c3f5de8ca01545a4dfeb26326709/packages/runtime-core/src/errorHandling.ts#L163-L166), make the `try catch` in `entry.ts` not been able to capture them.


packages/nuxt/src/app/entry.ts
```diff
@@ -76,6 +85,8 @@ if (import.meta.client) {
       nuxt.payload.error = (nuxt.payload.error || err) as any
     }
 
+    vueApp.config.errorHandler = prevVueErrorHandler
```
antfu, line 88: Reset back to undefined. Or should we actually always provide the handler with a better experience?

packages/nuxt/src/app/entry.ts
```diff
@@ -58,6 +58,15 @@ if (import.meta.client) {
...
     const nuxt = createNuxtApp({ vueApp })
 
+    async function handleVueError(err: any) {
+      await nuxt.callHook('app:error', err)
+      nuxt.payload.error = (nuxt.payload.error || err) as any
+    }
+
+    const prevVueErrorHandler = vueApp.config.errorHandler
+    if (!vueApp.config.errorHandler)
+      vueApp.config.errorHandler = handleVueError
```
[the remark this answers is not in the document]
danielroe: The user might provide their own error handler within a plugin but probably won't have provided it by this point.

danielroe (approved):

packages/nuxt/src/app/entry.ts
danielroe: Interesting question. I'd be up for always providing an error handler, but maybe we can get some thoughts from rest of team as well?

danielroe (commented):
Does this also need to be included in server-side fork?

antfu:
`errorHandler` is only used on the client side, so it should be fine to not set it on server

PR #24396, opened 2023-11-21
fix(nuxt): ensure `error` in `useAsyncData` has correct type

DamianGlowala:
### ❓ Type of change

### 📚 Description

Updates error handling composables and defines asyncData's error as `NuxtError` with generic Nuxt error's `data` property.
It also avoids `as any` casting in `createError()` to define `__nuxt_error` by using the `Object.defineProperty`.

(Will double check the docs for relevant updates once it's confirmed this update is good to go.)

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -74,53 +75,53 @@ export type AsyncData<Data, Error> = _AsyncData<Data, Error> & Promise<_AsyncDat
 
 export function useAsyncData<
   ResT,
-  DataE = Error,
+  NuxtErrorDataT = any,
```
danielroe, line 78: what about defaulting to `unknown`?

DamianGlowala: Sure. Should we consider defaulting to `unknown` directly in h3 (https://github.com/unjs/h3/blob/main/src/error.ts) as well?

danielroe: I would also recommend that but I know @pi0 may have a different timetable or opinion on this one.

danielroe (commented):
This seems like a good change but is also a breaking one for users specifying an error type.

See for example: https://github.com/search?q=/use(Lazy)?(AsyncData%7CFetch)+?%3C%5B%5E%3E,%5D,%5B%5E%3E%5D/&type=code

We might use inference to return previous return value if `NuxtErrorDataT` extends `Error` or `NuxtError`. wdyt?

DamianGlowala:
Yes, I agree, good find. Majority of the linked code snippets are React-related, but there were certainly a few Nuxt-related ones in these search results too.

Would the following be fine instead of `NuxtError<NuxtErrorDataT> | null`?
```
(NuxtErrorDataT extends Error | NuxtError ? NuxtErrorDataT : NuxtError<NuxtErrorDataT>) | null
```

PR #24399, opened 2023-11-22
fix(nuxt): avoid recursive ssr errors

### LINKED ISSUE #24393: Infinite loop when plugin/middleware throw error
### Environment

Nitro 2.8.0

### Reproduction

https://stackblitz.com/edit/github-syrpot?file=server%2Fapi%2Ftest.ts,plugins%2Ftest.server.ts

### Describe the bug

If we create a plugin using `useRequestFetch()`, the http headers from the browser are forwarded (`accept: 'text/html'`), and since [this commit](https://github.com/unjs/nitro/commit/0a5911e342a24d5099b6045605913d5d52d30d46) in Nitro, it won't look at it as JSON request.

What is weird is that when calling the API route and rendering the error page, it calls again the Nuxt plugin with the API route as path (I suspected to never do this).

### Additional context

Workaround is to explicit set the text/json header.

```ts
await useRequestFetch()('/api/test', {
  headers: {
    Accept: 'text/json',
  },
}).catch(() => {});
```

### Logs

_No response_

pi0:
### 🔗 Linked issue

resolves #24393

### ❓ Type of change

### 📚 Description

Nitro's default error handler is set to a custom Nuxt handler that unless the incoming request is from a json client, always prefers SSR rendering. If an error happens during SSR rendering of error page by making another fetch call, we will end up with an infinity loop.

This PR fixes issue by setting a custom header to the first error rendering phase and if we see the same header in a subsequent one, we simply fallback to the static HTML one that wouldn't cause recursion.

As a much much better improvement we could (and i believe we should) skip all plugins and nuxt middleware by default when rendering error page and allow opting-in using meta as current behavior can potentially plugins to misbehave in rendering errro pages but i guess is something of a bigger discussion as this would be a breaking change.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/error.ts
@@ -6,7 +6,7 @@ import { useRuntimeConfig } from '#internal/nitro'
 import { useNitroApp } from '#internal/nitro/app'
 import { isJsonRequest, normalizeError } from '#internal/nitro/utils'
 
-export default <NitroErrorHandler> async function errorhandler (error: H3Error, event) {
+export default <NitroErrorHandler>async function errorhandler(error: H3Error, event) {
   // Parse and normalize error
   const { stack, statusCode, statusMessage, message } = normalizeError(error)
 
@@ -45,14 +45,20 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error, 
     return send(event, JSON.stringify(errorObject))
   }
 
+  // Access request headers
+  const reqHeaders = getRequestHeaders(event)
+
+  // Detect to avoid recusrion in SSR rendering of errors
+  const isRenderingError = event.path.startsWith('/__nuxt_error') || !!reqHeaders['x-nuxt-error']
+
   // HTML response (via SSR)
-  const isErrorPage = event.path.startsWith('/__nuxt_error')
-  const res = !isErrorPage
-    ? await useNitroApp().localFetch(withQuery(joinURL(useRuntimeConfig().app.baseURL, '/__nuxt_error'), errorObject), {
-      headers: getRequestHeaders(event) as Record<string, string>,
+  const res = isRenderingError ? null : await useNitroApp().localFetch(
+    withQuery(joinURL(useRuntimeConfig().app.baseURL, '/__nuxt_error'), errorObject),
+    {
+      headers: { ...reqHeaders, 'x-nuxt-error': 'true' },
       redirect: 'manual'
-    }).catch(() => null)
-    : null
+    }
+  ).catch(() => null)
 
   // Fallback to static rendered error page
   if (!res) {
--- a/packages/nuxt/src/core/runtime/nitro/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/error.ts
@@ -48,7 +48,7 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error, 
   // Access request headers
   const reqHeaders = getRequestHeaders(event)
 
-  // Detect to avoid recusrion in SSR rendering of errors
+  // Detect to avoid recursion in SSR rendering of errors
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -45,14 +45,20 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
     return send(event, JSON.stringify(errorObject))
   }
 
+  // Access request headers
+  const reqHeaders = getRequestHeaders(event)
+
+  // Detect to avoid recusrion in SSR rendering of errors
+  const isRenderingError = event.path.startsWith('/__nuxt_error') || !!reqHeaders['x-nuxt-error']
```
danielroe, line 52: Do you think there is a way we can set a property that can't be 'forged'? That is, disallow a user from manually adding this header to a request?

pi0: I'm afraid and tried also not because the attached properties will be lost only the ones  that can be carried over HTTP are always safe to assume will be preserved in recursion.

We might use a random global number btw but it can make troubles for multi threading if source of recusrion is different.

Also i think it should be safe because 1) It is already possible by user agents to switch between json and HTML 2) Static fallback is almost the safest rendering path so adding header for benefit of exploiting an issue should not be reasonable

packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -45,14 +45,20 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
     return send(event, JSON.stringify(errorObject))
   }
 
+  // Access request headers
+  const reqHeaders = getRequestHeaders(event)
+
+  // Detect to avoid recusrion in SSR rendering of errors
```
danielroe, line 51:
```suggestion
  // Detect to avoid recursion in SSR rendering of errors
```

packages/nuxt/src/core/runtime/nitro/error.ts
danielroe: Okay. I wonder if it's worth making available (in future) an explicit way via unenv to pass data/payload to bypass this? I'd guess the issue will be caching but if we effectively opt it out of caching or specify a custom cache key(?)

danielroe (approved):

packages/nuxt/src/core/runtime/nitro/error.ts
pi0: It is possible as of today to also use `context/`__unenv__` init option but let's see in the future if we really need to do this and depend on unenv internals (i am mainly worried about future with multi-threading fetch that makes it harder to assume depending on unenv context)

test/fixtures/basic/plugins/error.server.ts
```diff
@@ -0,0 +1,5 @@
+export default defineNuxtPlugin(async () => {
+  if (useRequestHeaders(['x-recurse'])['x-recurse']) {
```
pi0, line 2: Small code style: I would use `x-test-` to make it more readable it is a test suite specific thing

```suggestion
  if (useRequestHeaders(['x-test-recurse-error'])['x-test-recurse-error']) {
```

test/basic.test.ts
```diff
@@ -814,6 +814,17 @@ describe('errors', () => {
         "url": "/__nuxt_error",
       }
     `)
+
+    it('should not recursively throw an error when there is an error rendering the error page', async () => {
+      const res = await $fetch('/', {
+        headers: {
+          'x-recurse': 'true',
```
pi0, line 821:
```suggestion
          'x-test-recurse-error': 'true',
```

test/fixtures/basic/plugins/error.server.ts
pi0: Don't we also have `useRequestHeader` btw?

danielroe: we do, yes.

Issue #24412, opened 2023-11-22
Types auto imported by Nuxt can't be used for defining components' props


DawidKopys:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.17.1`
- Nuxt Version:     `3.8.2`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.0`
- Package Manager:  `npm@9.6.7`
- Builder:          `-`
- User Config:      `devtools`, `imports`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

1. go to https://github.com/DawidKopys/nuxt-props-with-global-types/ and clone this repo
2. `npm i`
3. `npm run type-check`

### Describe the bug

First of all, let me thank all of the people involved in developing Nuxt. You are doing a great job creating this awesome piece of software!

Now back to the issue:
It seems that types auto imported by Nuxt can't be used for defining components' props (without importing them explicitly, for example from `#imports`). In fact it's even worse - they can be used, we don't get any ts errors, it just doesn't seem to work. It looks that it fails silently.

What is more, it also presents that globally available types can be used for defining props, just not if they are made available globally the way it is currently done in Nuxt.

To keep it brief:
- global types added to the global scope in the way Nuxt does that don't work when used for defining vue components' props. It seems that Nuxt is generating something alike below snippet inside `.nuxt/types/imports.d.ts`
```typescript
declare global {
  // @ts-ignore
  export type { GlobalExternalType1 } from './GlobalExternalType1'
}

```
- it seems that it could work, if Nuxt was generating something like below instead:
```typescript
declare global {
  interface GlobalExternalType2 {
    bar: string
  }
}
```
It would possibly require to copy over entire types' declarations in order to do that. But maybe it wouldn't be a problem since TypeScript is a  structural type system and when comparing types, TypeScript only takes into account the members on the type.

### Additional context

For additional info see https://github.com/DawidKopys/vue-props-with-global-types#readme as well. The same issue but with plain vue project, without Nuxt.

danielroe:
I think this would work automatically if we did explicitly add a type import in the `unimport` transform. We can possibly fix this upstream. What do you think @antfu?

antfu:
@danielroe it's not the same as transform, because this is type checking problem that happens in the editor instead of the pipeline.

antfu:
[...]

PR #24413, opened 2023-11-22
fix(nuxt): improve path resolve for `import.d.ts`

antfu:
Found it in the reproduction of https://github.com/nuxt/nuxt/issues/24412

When a module only exports types, the `.then(r => r.filter(i => !i.type))` filters the module out, causing the generated types contains unresolved raw path:

image

But whenever the module contains a runtime export, the path gets resolved:

Screenshot 2023-11-22 at 14 35 58


packages/nuxt/src/imports/module.ts
```diff
@@ -176,7 +180,7 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
   addTemplate({
     filename: 'types/imports.d.ts',
     getContents: async () => {
-      const imports = await ctx.getImports().then(r => r.filter(i => !i.type))
+      const imports = await ctx.getImports()
```
antfu, line 183: /cc @danielroe do you remember why you added this filter in https://github.com/nuxt/nuxt/pull/22476?

danielroe: No 🤔

Perhaps I was thinking we didn't need to _cache_ the type-only definitions as they wouldn't be emitted here...?

I think we can remove it safely.

danielroe (approved):

PR #24421, opened 2023-11-23
fix(nuxt): skip router middleware/redirections for islands

harlan-zw:
### 🔗 Linked issue

https://github.com/harlan-zw/nuxt-og-image/issues/98

### ❓ Type of change

### 📚 Description

When rendering a Nuxt Island, we provide the SSR context url as the root route `/`. Within the rendering pipeline, something is inheriting the route rules for `/` and applying them. I'm guessing Nitro.

The most obvious issue here is if you have a redirect on your home page, it's going to trigger a redirect on the island render.
You can see a minimal reproduction here: https://stackblitz.com/edit/nuxt-starter-gzuypu?file=pages%2Fabout.vue

While this could probably be dug into to figure out what's causing the redirect and doing an `if (!ssrContext.islandContext)`, I think it's safer for the long term not to re-use this root route.

It's possible that we are using `/` for a reason, maybe to avoid some 404 issues? Would be good to have your input. The updated URL is arbitrary, ideally, we'd pass no route within the context and rework the engine to support that (or escape early with a specific island renderer).

On that note but unrelated to this PR, I think there are some opportunities to refactor the island renderer to try and decouple it outside of the core logic. 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -5,7 +5,7 @@ import { appendResponseHeader } from 'h3'
 import { useHead } from '@unhead/vue'
 import { randomUUID } from 'uncrypto'
 import { joinURL, withQuery } from 'ufo'
-import { useRoute } from '#vue-router'
+import { useRoute } from '#imports'
 import type { FetchResponse } from 'ofetch'
 
 // eslint-disable-next-line import/no-restricted-paths
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -5,6 +5,7 @@ import { appendResponseHeader } from 'h3'
 import { useHead } from '@unhead/vue'
 import { randomUUID } from 'uncrypto'
 import { joinURL, withQuery } from 'ufo'
+import { useRoute } from '#vue-router'
 import type { FetchResponse } from 'ofetch'
 
 // eslint-disable-next-line import/no-restricted-paths
@@ -47,6 +48,7 @@ export default defineComponent({
     }
   },
   async setup (props, { slots }) {
+    const route = useRoute()
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
@@ -119,6 +121,8 @@ export default defineComponent({
       // TODO: Validate response
       // $fetch handles the app.baseURL in dev
       const r = await eventFetch(withQuery(((import.meta.dev && import.meta.client) || props.source) ? url : joinURL(config.app.baseURL ?? '', url), {
+        // url is overrideable
+        url: route.fullPath,
         ...props.context,
         props: props.props ? JSON.stringify(props.props) : undefined
       }))
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -11,7 +11,6 @@ import type { FetchResponse } from 'ofetch'
 import type { NuxtIslandResponse } from '../../core/runtime/nitro/renderer'
 import { useNuxtApp, useRuntimeConfig } from '../nuxt'
 import { prerenderRoutes, useRequestEvent } from '../composables/ssr'
-import { useRoute } from '../../app/composables/router'
 import { getFragmentHTML, getSlotProps } from './utils'
 
 // @ts-expect-error virtual file
@@ -48,7 +47,6 @@ export default defineComponent({
     }
   },
   async setup (props, { slots }) {
-    const route = useRoute()
     const error = ref<unknown>(null)
     const config = useRuntimeConfig()
     const nuxtApp = useNuxtApp()
@@ -121,8 +119,6 @@ export default defineComponent({
       // TODO: Validate response
       // $fetch handles the app.baseURL in dev
       const r = await eventFetch(withQuery(((import.meta.dev && import.meta.client) || props.source) ? url : joinURL(config.app.baseURL ?? '', url), {
-        // url is overrideable
-        url: route.fullPath,
         ...props.context,
         props: props.props ? JSON.stringify(props.props) : undefined
       }))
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -182,7 +182,7 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
   const context = event.method === 'GET' ? getQuery(event) : await readBody(event)
 
   const ctx: NuxtIslandContext = {
-    url: '/',
+    url: event.path,
```
danielroe, line 185: I don't think this is routeRules but rather vue router which is redirecting us, probably because we are still loading the router plugin for use within island.

I'm reluctant to use the `event.path` as this will still have the same issue with any middleware/redirects that apply to the original path that is loading the component.

Instead, I think we should add some more logic to the router plugin to check we're not in an island context, e.g.:

https://github.com/nuxt/nuxt/blob/17ff26ea931e3b430443ab28cc4ff62da205bb78/packages/nuxt/src/pages/runtime/plugins/router.ts#L146

Later we can perhaps opt-out entirely of routing (or, ideally, provide a 'stub' route rather than fully initialising vue-router). But we need to ensure that `useRoute()` continues to work within server components so we keep user-land differences to a minimum.

huang-julien: Yes this would cause an issue if there's rules applied on `/__nuxt_island/**`

Is there a possibility to ignore everything for `/__nuxt_island` ? It implies removing probably both routeRules, nitro global middlewares and the router.

Currently working on server pages, we'll probably need to disable the router plugin and provide a stubbed route for `useRoute` because it currently doesn't work within islands

harlan-zw: Ah yes, that makes sense if Vue Router is handling it. I've pushed up a fix for that.

It's not really clear in this PR but the `event.path` will always be `/__nuxt_island/**`, I've added a comment to make that clearer. I have a strong feeling that we should go with this path (assuming there isn't a good reason for using `/`), it's arbitrary as the user should never interact with it but it does solve another issue and may avoid issues in the future.

For example, if we have a plugin that does something when the user visits the home page directly, it will always get triggered when an island renders, we can work around with using `env: { islands: false }`, but this seems like an easy way to trip up people.

```ts
  setup() {
    const route = useRoute()
    if (route.path === '/') {
      // This will run in an island component
    }
  }
})
```

It feels like we're plugging holes with the implementation so gives us more protection against other possible bugs.

packages/nuxt/src/app/composables/router.ts
```diff
@@ -155,7 +155,7 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
   const nuxtApp = useNuxtApp()
 
   if (import.meta.server) {
-    if (nuxtApp.ssrContext && !nuxtApp.ssrContext.islandContext) {
+    if (nuxtApp.ssrContext) {
```
huang-julien, line 158: i think nuxtApp.ssrContext is always defined in server environment

huang-julien:
@harlan-zw @danielroe what do you think about that ? This is something done on the server page branch. It would allow to be in the same route as the page for an island. 
It's still overrideable and this way we won't need to provide any stub for `useROute`.

packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -135,6 +135,11 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
       await nuxtApp.runWithContext(() => showError(error))
     }
 
+    if (import.meta.server && nuxtApp.ssrContext?.islandContext) {
+      // We're in an island context, so we don't need to do anything else
```
danielroe, line 139:
```suggestion
      // We're in an island context, and don't need to handle middleware or redirections
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -119,6 +121,8 @@ export default defineComponent({
       // TODO: Validate response
       // $fetch handles the app.baseURL in dev
       const r = await eventFetch(withQuery(((import.meta.dev && import.meta.client) || props.source) ? url : joinURL(config.app.baseURL ?? '', url), {
+        // url is overrideable
+        url: route.fullPath,
```
danielroe, line 125:
```suggestion
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -47,6 +48,7 @@ export default defineComponent({
     }
   },
   async setup (props, { slots }) {
+    const route = useRoute()
```
danielroe, line 51:
```suggestion
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -11,6 +11,7 @@ import type { FetchResponse } from 'ofetch'
 import type { NuxtIslandResponse } from '../../core/runtime/nitro/renderer'
 import { useNuxtApp, useRuntimeConfig } from '../nuxt'
 import { prerenderRoutes, useRequestEvent } from '../composables/ssr'
+import { useRoute } from '../../app/composables/router'
```
danielroe, line 14:
```suggestion
```

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -182,7 +182,7 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
   const context = event.method === 'GET' ? getQuery(event) : await readBody(event)
 
   const ctx: NuxtIslandContext = {
-    url: '/',
+    url: event.path, // i.e /__nuxt_island/<component>_<hash>.json, can be ovveriden by user or nuxt island
```
danielroe, line 185: To merge, let's keep this limited to the strict fix for now (and discuss/raise new PR)...

For example, by setting url to `/__nuxt_island` vue router will no longer recognise the path and potentially throw errors.

I would suggest a stub router that prevents code from _breaking_ while effectively doing nothing.

```suggestion
    url: '/',
```

huang-julien: Sure 👌 what do you think about it @harlan-zw ? This should be enough to fix the issue

harlan-zw: Sounds good

harlan-zw:
I'm good with Daniel's changes, seems good to merge then. Will leave you to resolve them @huang-julien

danielroe (approved):

PR #24457, opened 2023-11-24
fix(kit): do not add local `~/modules` directories to `modulesDir`

### LINKED ISSUE #24439: Nuxt doesn't allow auto import types from modules
### Environment

```shell
- Node Version:     v20.9.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.0
- Package Manager:  pnpm@8.10.5
- Builder:          -
- User Config:      devtools, modules, vuetify, experimental
- Runtime Modules:  @unocss/nuxt@0.57.7, ~/layer-modules/another-vuetify, vuetify-nuxt-module@0.6.7
- Build Modules:    -
```

### Reproduction

https://github.com/userquin/vuetify-nuxt-module-types-check

### Describe the bug

Cannot use module hooks if we don't register the module in modules manually.

To reproduce the error, remove `~/modules/vuetify.ts`, run `pnpm nuxt prepare` and restart the window in VSCode or the Vue Language Server in JetBrains IDE and go to the module.

Running `vue-tsc --noEmit` works, why? :thinking: . Thx @BobbieGoede , `vue-tsc`  using `tsconfig.json` not `tsconfig.server.json`

### Additional context

When auto importing modules, the folder is excluded in the `tsconfig.json` file, and so using any module hook will not resolve the hook.

On left, registering the module in the modules, on right no:
imagen

Check the following video (also in linked reproduction):
https://streamable.com/5qdx8g

### Logs

NA: check screenshot and the video

BobbieGoede:
### 🔗 Linked issue
* #24439 

### ❓ Type of change

### 📚 Description
Resolves #24439 

This PR prevents layer `modules` dirs from being added to `modulesDir` which seems to only contain root directories of installed modules, contrary to its name I don't think it should include the `modules` dir itself!
 

### 📝 Checklist


danielroe:
`nuxt.options.modulesDir` is a list of folders to resolve `node_modules` from. You're quite right that we shouldn't include local `~/modules` directories there. But I think in the case of a layer that itself is in `node_modules`, we probably should include it or we might run into issues without `shamefully-hoist` enabled.

packages/kit/src/module/install.ts
```diff
@@ -27,8 +37,8 @@ export async function installModule (moduleToInstall: string | NuxtModule, inlin
   if (typeof moduleToInstall === 'string') {
     nuxt.options.build.transpile.push(normalizeModuleTranspilePath(moduleToInstall))
     const directory = getDirectory(moduleToInstall)
-    if (directory !== moduleToInstall) {
-      nuxt.options.modulesDir.push(getDirectory(moduleToInstall))
+    if (directory !== moduleToInstall && !localLayerModuleDirs.has(directory)) {
```
danielroe, line 40: we might even be able to remove this test entirely and only add directories within `node_modules` 🤔

danielroe (approved):

PR #24462, opened 2023-11-25
perf: cache nuxt config during static site generation

### LINKED ISSUE #10098: nuxt generate execute nuxt.config.js twice when set target=static
### Versions

- nuxt: v2.15.7
- node: v12.14.0

### Reproduction
image

### Steps to reproduce

1. set `target: 'static'` in `nuxt.config.js`
2. add log in `nuxt.config.js` just under the module: `console.log('hello')`
3. build with `nuxt generate` 
4. the shell would show two `hello` 

### What is Expected?
`nuxt.config.js` only executes once.

https://github.com/nuxt/nuxt.js/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/cli/src/commands/generate.js#L71-L75

https://github.com/nuxt/nuxt.js/blob/ab1c6cb442c9418128c85ad2cb14ed162af44933/packages/cli/src/utils/generate.js#L130-L131

`ensureBuild()` and `generate()` both call `getNuxt()` which calls the `cmd.getNuxtConfig()`. `getNuxtConfig` method calls `loadNuxtConfig()` later, which would clear the require cache before loading `nuxt.config.js`. As a result, `nuxt.config.js` would run twice.

Lehoczky:
### 🔗 Linked issue

Resolves #10098

### ❓ Type of change

### 📚 Description

The issue is that [getNuxtConfig()]() is being called 3 times during generation: [here](https://github.com/nuxt/nuxt/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/cli/src/commands/generate.js#L64) and [here](https://github.com/nuxt/nuxt/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/cli/src/utils/generate.js#L123) twice (once by [ensureBuild](https://github.com/nuxt/nuxt/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/cli/src/utils/generate.js#L22) and [generate](https://github.com/nuxt/nuxt/blob/853439ddf945b6ee358d5912e6e5a68009633c27/packages/cli/src/utils/generate.js#L10)).

First, I thought about passing the config as a parameter to `ensureBuild` and `generate` so they don't have to load it again:

```js
const config = await cmd.getNuxtConfig({ dev: false })

...

// Full static
if (config.target === TARGETS.static) {
      await ensureBuild(cmd, config)
      await generate(cmd, config)
      return
}
```

But both functions are heavily mutating the config, so this lead to bugs. I didn't want to install a lib to deep copy this object, so the next best thing that came to my mind is to create an option to not clear the `require` cache. It's not the prettiest solution, because the boolean flag needs to be passed through multiple functions, but so far, this is the easiest fix I can think of. 

### 📝 Checklist

### DIFF
--- a/test/fixtures/full-static/nuxt.config.js
+++ b/test/fixtures/full-static/nuxt.config.js
@@ -1,10 +1,5 @@
 export default {
   target: 'static',
-  export: {
-    payload: {
-      config: true
-    }
-  },
   router: {
     // base: '/test',
   },
@@ -18,7 +13,7 @@ export default {
     x: 123
   },
   hooks: {
-    export: {
+    generate: {
       before ({ setPayload }) {
         setPayload({ shared: true })
       },
--- a/packages/cli/src/command.js
+++ b/packages/cli/src/command.js
@@ -100,7 +100,7 @@ export default class NuxtCommand extends Hookable {
     return this._parsedArgv
   }
 
-  async getNuxtConfig (extraOptions = {}) {
+  async getNuxtConfig (extraOptions = {}, loadOptions = {}) {
     // Flag to indicate nuxt is running with CLI (not programmatic)
     extraOptions._cli = true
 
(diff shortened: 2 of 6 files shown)


test/fixtures/full-static/nuxt.config.js
```diff
@@ -18,7 +13,7 @@ export default {
     x: 123
   },
   hooks: {
-    export: {
+    generate: {
```
Lehoczky, line 16: `export` is deprecated, so I changed it to `generate`

Issue #24196, opened 2023-11-08
Optimizing dependencies - Nuxt 3.8.1


frasza:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v20.9.0
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  pnpm@8.9.0
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://github.com/frasza/nuxt-optimize-deps

### Describe the bug

Since Nuxt 3.8.1 I (and few others on Discord) have noticed on development app boot that Nuxt/Vite does some dependencies optimization and it happens quite often (sometimes even on every page refresh, sometimes I noticed only once per dev server startup). Problem is that is also does full page refresh.

```ts
ℹ Vite server warmed up in 846ms                                                                                                                                                                        18:50:50
ℹ Vite client warmed up in 1195ms                                                                                                                                                                       18:50:50
✔ Nitro built in 548 ms                                                                                                                                                                           nitro 18:50:51
ℹ ✨ new dependencies optimized: @tiptap/vue-3, @tiptap/starter-kit                                                                                                                                     18:50:53
ℹ ✨ optimized dependencies changed. reloading    
```

Provided small reproduction using `tiptap` editor. On my own project I have even more of these packages that produce the issue. 
```ts
[18:43:23] ℹ ✨ new dependencies optimized: @tiptap/vue-3, @tiptap/starter-kit, @tiptap/extension-placeholder, @tiptap/extension-link, primevue/usetoast
ℹ ✨ optimized dependencies changed. reloading   

[18:43:44] ℹ ✨ new dependencies optimized: echarts/core, echarts/renderers, echarts/charts, echarts/components, vue-echarts
ℹ ✨ optimized dependencies changed. reloading 
```

It really does impact DX since its doing full page refresh when it happens. If I revert to Nuxt 3.8.0, issue is gone.

dolbex:
I am having the same issue.

DurtyFree:
Same issue here, this is very uncomfortable, especially on low spec developer machines

manniL:
Please don't write "Me too" or "Same issue here" - instead leave a 👍🏻 on the issue ☺️

nathanchase:
[...]

PR #24584, opened 2023-12-04
fix(nuxt): eager load island components map

harlan-zw:
### 🔗 Linked issue

https://github.com/harlan-zw/nuxt-og-image/issues/117

### ❓ Type of change

### 📚 Description

See https://github.com/nuxt/nuxt/issues/24196#issuecomment-1853870504 for context. The below was a band-aid solution for upstream bugs that is no longer relevant.

This PR Is now just for adding a test for server component HMR.

---

The underlying issue is that HMR is not working for server / island components. This is a regression as it was working correctly previously.

**3.8.2: Broken** - https://stackblitz.com/edit/nuxt-starter-58zyad?file=nuxt.config.ts

**3.8.0: Working** - https://stackblitz.com/edit/nuxt-starter-p136go?file=components%2FFoo.island.vue

It's quite a strange bug and has involved a bit of effort to track down. I've figured out it was introduced within this PR / line https://github.com/nuxt/nuxt/pull/23975/files#diff-b846c59503e5fc45aa8f9cdde89d0a0c89f32a48ec731b1560dbcb334c98cf18R44.

The total scope of how Vite is figuring out the dependency graph is a bit over my head, but what I've gathered is that without this warm-up step, the server / island components are missing their module graph data.

When you modify the component the HMR event is fired but it won't bubble anything up because of that.

Speculating, I would say that the HMR for dynamically imported modules within virtual server modules are broken currently (although Nitro _seems_ to be okay?). I'd need to do some more testing around this.

**Edit Foo.island.vue - HMR Event with warmup**

```ts
{
file: '/home/harlan/packages/nuxt/playground/components/Foo.island.vue', 
  timestamp: 1701690522771,
  modules:
   [ ModuleNode {
       id: '/home/harlan/packages/nuxt/playground/components/Foo.island.vue',
       file: '/home/harlan/packages/nuxt/playground/components/Foo.island.vue',
       importers: [Set],
       clientImportedModules: [Set]
     //...
  }]
}
```

**Edit Foo.island.vue - HMR Event without warmup**

```ts
{ file: '/home/harlan/packages/nuxt/playground/components/Foo.island.vue',
  timestamp: 1701691631721,
  modules: [],
}
```

It seems like we could also manually trigger the invalidation of the `#build/components.islands.mjs` module, but doesn't address the underlying issue.

I'm not sure if my solution here is the right one though, look forward to your thoughts.

### 📝 Checklist


danielroe:
cc: @antfu in case you have any ideas

danielroe:
@harlan-zw Do you think we could add a regression test to ensure this doesn't crop up again? 🙏

harlan-zw:
> @harlan-zw Do you think you could add a regression test to ensure this doesn't crop up again? 🙏

Of course :) I'll give it a go tomorrow.

packages/vite/src/utils/warmup.ts
```diff
@@ -34,6 +34,10 @@ function isBuiltin (id: string) {
   return id.startsWith('node:') || builtins.has(id)
 }
 
+function isVirtual (id: string) {
+  return id.startsWith('virtual:')
```
antfu, line 38: `virtual:` is a community conversion, also it could be `\0` or honestly everything. I wonder if there could be a more general solution to this (I am ok with treating them a bit specifically at this moment, but it might not be a solid solution for the long term)

packages/vite/src/utils/warmup.ts
```diff
@@ -50,7 +54,10 @@ export async function warmupViteServer (
       const m = await server.moduleGraph.getModuleByUrl(url, isServer)
       // a module that is already compiled (and can't be warmed up anyway)
       if (m?.transformResult?.code || m?.ssrTransformResult?.code) {
```
antfu, line 56: Maybe we could check `isServer ? m?.ssrTransformResult?.code : m?.transformResult?.code`?

harlan-zw:
While writing the tests for it, I noticed that the HMR wouldn't work at all with the `virtual:` filter. Not exactly sure why this is and whether it's just an artifact of the modules being local instead of within node_modules themselves.

As a workaround, I filtered for the module existing within node_modules, which works, but as you can imagine adds back a lot of warm-up time, almost 40% slower (5 iterations). Not exactly sure how else to filter it, unless we add some testing exception.

Will think about this.

test/hmr.test.ts
```diff
@@ -21,7 +21,8 @@ if (process.env.TEST_ENV !== 'built' && !isWindows) {
     nuxtConfig: {
       builder: isWebpack ? 'webpack' : 'vite',
       buildDir: process.env.NITRO_BUILD_DIR,
-      nitro: { output: { dir: process.env.NITRO_OUTPUT_DIR } }
+      nitro: { output: { dir: process.env.NITRO_OUTPUT_DIR } },
+
```
danielroe, line 25:
```suggestion
      nitro: { output: { dir: process.env.NITRO_OUTPUT_DIR } }
```

harlan-zw:
Thanks for the updates. Is there anything blocking this going into the next release that you want me to take a look at? 

I think the `node_module` warmup time issue should be more of a non-factor when Nuxt itself is within node_modules but can do some benchmarking if needed

antfu:
I dug into it a bit, and I hope https://github.com/nuxt/nuxt/issues/24196#issuecomment-1853505843 would be the root issue that also fixes your cases, too.

harlan-zw:
> I dug into it a bit, and I hope [#24196 (comment)](https://github.com/nuxt/nuxt/issues/24196#issuecomment-1853505843) would be the root issue that also fixes your cases, too.

This is great news, thank you for your help.

I've modified this PR just to be a test case for server component HMR, let's see if / when it passes :smile:

harlan-zw:
After discussing this with @danielroe, we both agreed adding a "hack" to the warmup code was a bit icky, he indicated this may be an issue with Vite Node caching. I looked into Vite Node further and wasn't able to come up with anything.

What is clear is that Vite is having some issues with the module graph for `#build/components.islands.mjs`. Looking into the code, it's clear this is because the module is within a dynamic import. So scanning the entry isn't going to track the dependency and caches aren't going to get cleared properly (?).

I've pushed up a fix to have the entry scan be able to resolve this module without resolving the dynamic imports and it seems to be working.

In terms of the performance cost of this solution, I think there isn't any as we were already importing all island components on every server request.

Would be great to hear your thoughts on this solution

packages/nuxt/src/core/templates.ts
```diff
@@ -40,7 +40,9 @@ export const appComponentTemplate: NuxtTemplate<TemplateContext> = {
 // TODO: Use an alias
 export const rootComponentTemplate: NuxtTemplate<TemplateContext> = {
   filename: 'root-component.mjs',
-  getContents: ctx => genExport(ctx.app.rootComponent!, ['default'])
+  // TODO: fix upstream in vite - this ensures that vite generates a module graph for islands
+  // but should not be necessary (and has a warmup performance cost). See https://github.com/nuxt/nuxt/pull/24584.
+  getContents: ctx => (ctx.nuxt.options.dev ? "import '#build/components.islands.mjs';\n" : '') + genExport(ctx.app.rootComponent!, ['default'])
```
harlan-zw, line 45: nice

danielroe (approved):

packages/nuxt/src/core/templates.ts
danielroe: ❤️

PR #24673, opened 2023-12-10
fix(nuxt): use fresh route when `<NuxtLayout>` first renders

### LINKED ISSUE #24642: useRoute not correctly populated with async layout
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-bvoawn?file=pages%2Findex.vue,app.vue,layouts%2FTestLayout.vue,pages%2F[profileId].vue

I left a comment in "TestLayout.vue" that you should uncomment to see a fix

### Describe the bug

Upon navigating to a layout that uses top-level await, `useRoute().params`  is not populated correctly on the client side. Upon hard refresh, it is populated correctly

### Additional context

I left a comment in "TestLayout.vue" that you should uncomment to see a fix

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/24642

### ❓ Type of change

### 📚 Description

~~Hi :wave: this PR sync `nuxtApp._route` for `NuxtLayout`~~

**edited**:

this PR updated `useRoute()` within layouts to use the vue-router route if the layout is rendering for the first time in a page transition

I've marked two TODOs in the test - these aren't regressions from this PR but issues I spotted and which we could ultimately fix.

### 📝 Checklist


packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -104,6 +104,9 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
       // so we need to update the route manually
       if (to.matched[0]?.components?.default === from.matched[0]?.components?.default) {
```
danielroe, line 105: Should we instead use the isSameRoute utility you created in another PR?

packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -104,6 +104,9 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
...
         syncCurrentRoute()
+      } else {
+        // assign toValue to avoid reactivity
+        Object.assign(toValue(_route), router.currentRoute.value)
```
danielroe, line 109: I'm not sure this is the correct fix (but do tell me what you're thinking as I could be wrong here). This is omitted initially as the suspense resolution _should_ update the route (and I think we always want it to be reactive).

huang-julien: In that case, should we always sync the route ?

danielroe: We should sync the route, but if we do it here when suspense is involved, it will be too early and update the view before the setup function of the new route has finished

huang-julien: I think assigning this way shouldn't update the view since vue won't trigger reactivity. Is there another way of fixing it ?

packages/nuxt/src/app/components/nuxt-layout.ts
```diff
@@ -51,10 +51,12 @@ export default defineComponent({
...
     // Need to ensure (if we are not a child of `<NuxtPage>`) that we use synchronous route (not deferred)
     const injectedRoute = inject(PageRouteSymbol)
-    const route = injectedRoute === useRoute() ? useVueRouterRoute() : injectedRoute
+    const shouldUseEagerRoute = !injectedRoute /* this should never be true */
+      || injectedRoute === useRoute() /* this is only true if we are not within `<NuxtPage>` */
+    const route = shouldUseEagerRoute ? useVueRouterRoute() as ReturnType<typeof useRoute> : injectedRoute
 
     const layout = computed(() => {
-      let layout = unref(props.name) ?? route.meta.layout as string ?? 'default'
+      let layout = unref(props.name) ?? route!.meta.layout as string ?? 'default'
```
huang-julien, line 59:
```suggestion
      let layout = unref(props.name) ?? route?.meta.layout as string ?? 'default'
```

packages/nuxt/src/app/components/nuxt-layout.ts
```diff
@@ -79,9 +81,14 @@ export default defineComponent({
       nuxtApp._isNuxtLayoutUsed = true
     }
 
+    let lastLayout: string | boolean | undefined
+
     return () => {
       const hasLayout = layout.value && layout.value in layouts
-      const transitionProps = route.meta.layoutTransition ?? defaultLayoutTransition
+      const transitionProps = route!.meta.layoutTransition ?? defaultLayoutTransition
```
huang-julien, line 88:
```suggestion
      const transitionProps = route?.meta.layoutTransition ?? defaultLayoutTransition
```

PR #24674, opened 2023-12-10
feat(nuxt): add path to `error.data` when throwing 404 errors

### LINKED ISSUE #24543: issues output pathname in `~/error.vue`
### Environment

```
Working directory: /home/projects/nuxt-starter-y5gman
Nuxt project info:

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/nuxt-pathname-error-page?file=pages%2Findex.vue

### Describe the bug

I'd like to output the pathname in the case of a 404 being handled by `error.vue`. For example a message like:

```vue
Could not find the page {{ location.pathname }}
```

However, it seems that the pathname is always behind, since the error page begins rendering before the browser's url is updated. There's a similar, but different issue with `$route.fullPath`.

In cases where the route simply does not exist, `location.pathname` gives the correct value, but `$route.fullPath` gives the url for the previous page.

In cases where the navigation fails due to route validation, neither are correct - they both show the previous url. 

### Additional context

_No response_

### Logs

_No response_

TheAlexLichter:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/24543

### ❓ Type of change

### 📚 Description

When having an error thrown by a `false` value from `validate`, no info about the route can be obtained.
Same issue when Nuxt's `router.ts` creates a `404` error.

This PR adds the `fullPath` info from the `to` route as `error.data` in these cases as `path` key.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -246,7 +246,10 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>({
               if (result === false || result instanceof Error) {
                 const error = result || createError({
                   statusCode: 404,
-                  statusMessage: `Page Not Found: ${initialURL}`
+                  statusMessage: `Page Not Found: ${initialURL}`,
+                  data: {
+                    to
+                  }
                 })
                 delete nuxtApp._processingMiddleware
                 return nuxtApp.runWithContext(() => showError(error))
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -248,7 +248,7 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>({
                   statusCode: 404,
                   statusMessage: `Page Not Found: ${initialURL}`,
                   data: {
-                    path: to.fullPath
+                    path: initialURL
                   }
                 })
                 delete nuxtApp._processingMiddleware
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -248,7 +248,7 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>({
                   statusCode: 404,
                   statusMessage: `Page Not Found: ${initialURL}`,
                   data: {
-                    to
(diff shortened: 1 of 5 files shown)


packages/nuxt/src/app/plugins/router.ts
```diff
@@ -246,7 +246,10 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>({
               if (result === false || result instanceof Error) {
                 const error = result || createError({
                   statusCode: 404,
-                  statusMessage: `Page Not Found: ${initialURL}`
+                  statusMessage: `Page Not Found: ${initialURL}`,
+                  data: {
+                    to
```
danielroe, line 251: I think we should likely include just `to.fullPath` (maybe `path: to.fullPath`?) as otherwise the entire route object will (try to) be serialised on an SSR error, which I imagine will throw an error.

manniL: I was thinking about that too first but thought including more info wouldn't hurt (or maybe excluding meta) 🤔

But only `fullPath` sounds good to me!

packages/nuxt/src/pages/runtime/validate.ts
```diff
@@ -15,9 +15,13 @@ export default defineNuxtRouteMiddleware(async (to) => {
   if (import.meta.server) {
     return result
   }
+
   const error = createError({
     statusCode: 404,
-    statusMessage: `Page Not Found: ${to.fullPath}`
+    statusMessage: `Page Not Found: ${to.fullPath}`,
+    data: {
+      to
```
danielroe, line 23:
```suggestion
      path: to.fullPath
```

danielroe (commented):
I think it would also be nice to have a test for this behaviour. Maybe could piggy back on existing tests for validate, etc.?

danielroe (approved):

PR #24707, opened 2023-12-12
feat(kit): load `/module` or `/nuxt` module subpath if it exists

danielroe:
### ❓ Type of change

### 📚 Description

This adds support for detecting a `/module` subpath and loading it where it exists.

This would mean a project that is primarily a library can ship a Nuxt module as `/module` and support adding just the root library name to `modules: []`.

#### Example

```diff
  export default defineNuxtConfig({
    modules: [
-    'some-library/module'
+    'some-library'
    ]
  })
```

(We can, for example, use this straight away for `@nuxt/test-utils/module`.)

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/install.ts
+++ b/packages/kit/src/module/install.ts
@@ -61,19 +61,30 @@ export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, n
   let buildTimeModuleMeta: ModuleMeta = {}
   // Import if input is string
   if (typeof nuxtModule === 'string') {
-    const src = await resolvePath(nuxtModule)
-    try {
+    const paths = [join(nuxtModule, 'module'), nuxtModule]
+    let error: unknown
+    for (const path of paths) {
+      const src = await resolvePath(path)
       // Prefer ESM resolution if possible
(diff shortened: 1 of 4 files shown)


Atinux:
I love the idea 🚀

packages/kit/src/module/install.ts
```diff
@@ -61,18 +61,28 @@ export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, n
   let buildTimeModuleMeta: ModuleMeta = {}
   // Import if input is string
   if (typeof nuxtModule === 'string') {
-    const src = await resolvePath(nuxtModule)
-    try {
+    const paths = [join(nuxtModule, 'module'), nuxtModule]
```
pi0, line 64: It would be nice if we also add `/nuxt` prefix for opt-in modules support from libraries.

pi0:
```suggestion
    const paths = [join(nuxtModule, 'nuxt'), join(nuxtModule, 'module'), nuxtModule]
```

pi0 (approved):
🚀

PR #24709, opened 2023-12-12
feat(nuxt): layers support for spa loading template

### LINKED ISSUE #24275: Support referencing spa loading template from base layer
### Discussed in https://github.com/nuxt/nuxt/discussions/24274

<div type='discussions-op-text'>

<sup>Originally posted by **trc-mathieu** November 13, 2023</sup>
I have a use case where we are reusing the same base layer for all the apps we build in our organization. I would like to reuse the loading indicator that we normally put in the `app/spa-loading-template.html` or at least be able to reference it in the nuxt.config.
As reference, we only look at the `srcDir`:
https://github.com/nuxt/nuxt/blob/8bff8136c16a6e42e7f34590b750b2b6a60730bc/packages/nuxt/src/core/nitro.ts#L517-L520

If it could automatically take the `app/spa-loading-template.html` from the layer, that would be awesome, or allow us to configure it in the nuxt.config.</div>

doganalper:
### 🔗 Linked issue

resolves #24275

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

An option to also extend `spaLoadingTemplate` from `base` layer.

### 📝 Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -525,10 +525,6 @@ function relativeWithDot (from: string, to: string) {
 }
 
 async function spaLoadingTemplatePath (nuxt: Nuxt) {
-  if (typeof nuxt.options.spaLoadingTemplate === "string") {
-    return resolve(nuxt.options.srcDir, nuxt.options.spaLoadingTemplate);
-  }
-
   return (await findPath(
     nuxt.options._layers.map(layer => {
       const layerSourceDir = layer.config.srcDir;
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -378,7 +378,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
 
   // Trigger Nitro reload when SPA loading template changes
   nuxt.hook('builder:watch', async (_event, path) => {
-    if (normalize(path) === spaLoadingTemplatePath(nuxt)) {
+    if (normalize(path) === await spaLoadingTemplatePath(nuxt)) {
       await nitro.hooks.callHook('rollup:reload')
     }
   })
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -532,7 +532,7 @@ function spaLoadingTemplatePath (nuxt: Nuxt) {
   const baseLayer = nuxt.options._layers.find((layer) =>
     layer.cwd.includes("base"),
   );
-  if (baseLayer) {
+  if (baseLayer && existsSync(baseLayer.cwd)) {
     if (typeof baseLayer.config.spaLoadingTemplate === "string") {
       return resolve(
         baseLayer.config.srcDir,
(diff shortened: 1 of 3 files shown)


doganalper:
Possible problem here would be if users give any name other than `base`. I added as `base` because it is what is written on [documentation.](https://nuxt.com/docs/getting-started/layers#usage)

Is it possible to extend multiple local layers? If not one possible solution would be looking local layer and extend `spaLoadingTemplate` from that.

danielroe (changes requested):
This is nice 👌 

One change, though: we shouldn't hard-code the name `base`. Instead, we should iterate over the layers in order to see if any of them has a spa loading template, for example:

https://github.com/nuxt/nuxt/blob/00917a1bd8e40ed93e11b2471f0f2a921e57de99/packages/nuxt/src/core/app.ts#L103-L107

packages/nuxt/src/core/nitro.ts
```diff
@@ -378,7 +378,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
 
   // Trigger Nitro reload when SPA loading template changes
   nuxt.hook('builder:watch', async (_event, path) => {
-    if (normalize(path) === spaLoadingTemplatePath(nuxt)) {
+    if (normalize(path) === await spaLoadingTemplatePath(nuxt)) {
```
danielroe, line 381: Maybe move this outside of the hook so it doesn't rerun every time a file is modified?

packages/schema/src/config/app.ts
```diff
@@ -183,7 +183,8 @@ export default defineUntypedSchema({
    * rendered with `ssr: false`.
    * - If it is unset, it will use `~/app/spa-loading-template.html` if it exists.
    * - If it is false, no SPA loading indicator will be loaded.
-   * - If true, Nuxt will look for `~/app/spa-loading-template.html` file or a default Nuxt image will be used.
+   * - If true, Nuxt will first look for a layer called `base` and its `spa-loading-template` in `nuxt.config` file.
```
danielroe, line 186: We probably need to update these lines...

danielroe (commented):
Nice 👌

packages/schema/src/config/app.ts
doganalper: My bad 😅

danielroe (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -524,16 +525,20 @@ function relativeWithDot (from: string, to: string) {
   return relative(from, to).replace(/^([^.])/, './$1') || '.'
 }
 
-function spaLoadingTemplatePath (nuxt: Nuxt) {
-  return typeof nuxt.options.spaLoadingTemplate === 'string'
-    ? resolve(nuxt.options.srcDir, nuxt.options.spaLoadingTemplate)
-    : resolve(nuxt.options.srcDir, 'app/spa-loading-template.html')
+async function spaLoadingTemplatePath (nuxt: Nuxt) {
+  if (typeof nuxt.options.spaLoadingTemplate === 'string') {
```
doganalper, line 529: @danielroe Nice changes. I removed this part on my last comment because `nuxt.options._layers` has project's initial config (same as `nuxt.options`) as first item and thought looping through `_layers` would bring same output.

danielroe: You're quite right, it would have a similar effect.

But in this case, for consistency, I think we want the behaviour to be the same as other layer options (like error.vue). We will resolve the path automatically if it's _not_ specified in config, and otherwise let `c12` resolve it for us. (The edge case we avoid here would be something like a more prioritised layer specifying an _invalid_ path and it falling back without an error to the lower-level layer.)

trc-mathieu:
Thank you!

PR #24711, opened 2023-12-12
feat(vite): replace browser globals with `undefined` on server

danielroe:
### 🔗 Linked issue

related https://github.com/nuxt/nuxt/pull/24182

### ❓ Type of change

### 📚 Description

With Vite 5 improvements on `define`, we can now define `window`, `document` when used on server, meaning an even wider range of code can be tree shaken (e.g. `if (typeof document === 'undefined')` but also `if (document)`. 

It should also be more accurate to use `define` rather than `replace` as it won't replace non-code (e.g. string literals).

### 📝 Checklist


pi0 (approved):
much better than before!

PR #24713, opened 2023-12-12
chore(kit): remove unnecessary assignment

GalacticHypernova:
### ❓ Type of change

### 📚 Description

In the `resolvePath` method there is a const assignment that is only used once. Considering all the operations essentially happen on the value that is assigned, this PR simply removes this assignment and updates the one time use of that variable with the value itself.

### 📝 Checklist


danielroe:
(We do this to preserve the non-normalised version of the path.)

GalacticHypernova:
> (We do this to preserve the non-normalised version of the path; it is not the same value.)

I see, sorry about that!

PR #24715, opened 2023-12-12
perf(vite): simplify manifest property acccess

GalacticHypernova:
### ❓ Type of change

### 📚 Description

This PR caches each key in the client manifest to improve performance and decrease the manifest handling time. Tested with core-js package.json file, performance is increased by **~90%** (view second attachment):
image

And with native console.time/timeEnd:
image

### 📝 Checklist

### DIFF
--- a/packages/vite/src/manifest.ts
+++ b/packages/vite/src/manifest.ts
@@ -35,13 +35,13 @@ export async function writeManifest (ctx: ViteBuildContext, css: string[] = []) 
   const BASE_RE = new RegExp(`^${escapeRE(buildAssetsDir)}`)
 
   for (const key in clientManifest) {
-    const cliManifestKey=clientManifest[key]
-    if (cliManifestKey.file) {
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
(The failing test has nothing to do with the PR whatsoever, and might be worth fixing in the test case itself)

danielroe:
This is run once after the vite build, and iterates over every file emitted from the vite client build.

I would not imagine there is a significant difference in time between 6x accesses of a deep property and 1x access of deep property and 5x one level less. But I'm happy to be proved wrong.

How did you run the test in the issue?

GalacticHypernova:
@danielroe 
clientManifest is the core-js package-lock.json file (won't paste it here because it's too large), BASE_RE is new RegExp('abcde')
Rest of the code is: 
repeatedIndex: the same as the current file
cachedAccess: this PR
Same goes with all my perf PR's

danielroe (approved):

PR #24718, opened 2023-12-12
perf(nuxt): don't dedupe fewer than two middleware/plugins

GalacticHypernova:
### ❓ Type of change

### 📚 Description

This PR returns early in the `uniqueBy` function that is used to gather all the unique plugins/middleware/etc whenever the input array is empty (which isn't unrealistic, if someone doesn't use any plugins for example) or contains a single file (guaranteed to be unique), avoiding all the computation afterwards and increasing performance in both cases by **~66.67%**.
With the following benchmark I have gotten the following results:
```ts
 function uniqueByNonReg<T, K extends keyof T>(arr: T[], key: K) {
  if (arr.length < 2) {
    return arr;
  }

  const res: T[] = [];
  const seen = new Set<T[K]>();

  for (const item of arr) {
    if (seen.has(item[key])) {
      continue;
    }

    seen.add(item[key]);
    res.push(item);
  }

  return res;
}

 function uniqueBy<T, K extends keyof T> (arr: T[], key: K) {
  const res: T[] = []
  const seen = new Set<T[K]>()
  for (const item of arr) {
    if (seen.has(item[key])) { continue }
    seen.add(item[key])
    res.push(item)
  }
  return res
}

console.time('regularEmpty')
for(let i=0;i<100000;i++){
  uniqueBy([],'name')
}
console.timeEnd('regularEmpty')
console.time('nonRegularEmpty')
for(let i=0;i<100000;i++){
  uniqueByNonReg([],'name')
}
console.timeEnd('nonRegularEmpty')

console.time('regularSingle')
for(let i=0;i<100000;i++){
  uniqueBy([{name:'f'}],'name')
}
console.timeEnd('regularSingle')
console.time('nonRegularSingle')
for(let i=0;i<100000;i++){
  uniqueByNonReg([{name:'f'}],'name')
}
console.timeEnd('nonRegularSingle')

```

image
### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/utils/index.ts
+++ b/packages/nuxt/src/core/utils/index.ts
@@ -2,8 +2,8 @@ export * from './names'
 export * from './plugins'
 
 export function uniqueBy<T, K extends keyof T> (arr: T[], key: K) {
-  if (arr.length === 0) {
-    return [];
+  if (arr.length < 2) {
+    return arr;
   }
   const res: T[] = []
   const seen = new Set<T[K]>()
--- a/packages/nuxt/src/core/utils/index.ts
+++ b/packages/nuxt/src/core/utils/index.ts
@@ -2,6 +2,9 @@ export * from './names'
 export * from './plugins'
 
 export function uniqueBy<T, K extends keyof T> (arr: T[], key: K) {
+  if (arr.length === 0) {
+    return [];
+  }
   const res: T[] = []
   const seen = new Set<T[K]>()
   for (const item of arr) {
--- a/packages/nuxt/src/core/utils/index.ts
+++ b/packages/nuxt/src/core/utils/index.ts
@@ -3,7 +3,7 @@ export * from './plugins'
 
 export function uniqueBy<T, K extends keyof T> (arr: T[], key: K) {
   if (arr.length < 2) {
-    return arr;
+    return arr
   }
   const res: T[] = []
   const seen = new Set<T[K]>()


GalacticHypernova:
I will look into optimizing the entire Set handling (perhaps rework it to a Map?), but from my current benchmarks the results are still inconclusive so I have not added that part yet.

GalacticHypernova:
Update: the new update (the single file early return) yields even faster results, I will soon post new benchmark results

packages/nuxt/src/core/utils/index.ts
```diff
@@ -2,6 +2,9 @@ export * from './names'
 export * from './plugins'
 
 export function uniqueBy<T, K extends keyof T> (arr: T[], key: K) {
+  if (arr.length < 2) {
+    return arr;
```
danielroe, line 6:
```suggestion
    return arr
```

danielroe:
Nuxt itself will always add more than 2 plugins, which means the additional check is likely to increase the time required when resolving plugins.

(Though I don't see the harm in adding this check as it might be relevant for middleware - Nuxt only adds one by default.)

GalacticHypernova:
> which means the additional check is likely to increase the time required when resolving plugins.

Well, the check is O(1), it doesn't add time at all compared to the time it saves because it's a boolean check. I'm not sure about the usability of that function in modules and the likes, but could people use it in their own modules or to register their own middleware/plugins?

GalacticHypernova:
I see. Allow me to benchmark this really quickly.

danielroe (approved):

PR #24725, opened 2023-12-13
perf(nuxt): save reference to options

GalacticHypernova:
### ❓ Type of change

### 📚 Description

There are many repeated references to nuxt.options in both `generateApp` and `resolveApp`, why not save a reference to improve performance? I initially thought about making it a shared variable but decided not to (curious to know if you think it's worth adding as a shared variable).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -8,6 +8,7 @@ import * as defaultTemplates from './templates'
 import { getNameFromPath, hasSuffix, uniqueBy } from './utils'
 import { extractMetadata, orderMap } from './plugins/plugin-metadata'
 
+
 export function createApp (nuxt: Nuxt, options: Partial<NuxtApp> = {}): NuxtApp {
   return defu(options, {
     dir: nuxt.options.srcDir,
@@ -19,11 +20,12 @@ export function createApp (nuxt: Nuxt, options: Partial<NuxtApp> = {}): NuxtApp 
 }
 
 export async function generateApp (nuxt: Nuxt, app: NuxtApp, options: { filter?: (template: ResolvedNuxtTemplate<any>) => boolean } = {}) {
+  const nuxtOptions=nuxt.options
   // Resolve app
   await resolveApp(nuxt, app)
 
   // User templates from options.build.templates
-  app.templates = Object.values(defaultTemplates).concat(nuxt.options.build.templates) as NuxtTemplate[]
+  app.templates = Object.values(defaultTemplates).concat(nuxtOptions.build.templates) as NuxtTemplate[]
 
   // Extend templates with hook
   await nuxt.callHook('app:templates', app)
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
@danielroe Before I continue with my PR's, I wanted to ask you: I decided to go on a little optimization spree (as you can see from the latest 4 PR's). Should I combine them all into a single PR based on the affected package or keep them as is? On one hand I don't want to spam the PR page with my contributions, but on the other hand I remember you asking PR's to be very specific, and to include each additional contribution in a separte PR to keep it on-topic.

packages/nuxt/src/core/app.ts
```diff
@@ -19,11 +20,12 @@ export function createApp (nuxt: Nuxt, options: Partial<NuxtApp> = {}): NuxtApp
 }
 
 export async function generateApp (nuxt: Nuxt, app: NuxtApp, options: { filter?: (template: ResolvedNuxtTemplate<any>) => boolean } = {}) {
+  const nuxtOptions=nuxt.options
```
pi0, line 23: There are no high-level performance improvements by doing this. The non-destructuring pattern is used across our codebases to make it easier to trace back to the source of each variable and increase code-portability/consistency by sharing the same context `nuxt` everywhere.

---

BTW nice to read the article: https://mathiasbynens.be/notes/shapes-ics (in short if in theory say we wanted to call `generateApp` hundreds of times, js engines such as v8 automatically optimize property access using inline-caches)

danielroe:
The separate PRs are totally fine.

But I'm not sure that there is a meaningful performance gain to be made by reducing one level of object property access, compared to the decrease in readability. We deliberately try to access everything off the `nuxt` object, for example, for clarity.

Have you done any measurements that suggest this is a meaningful performance gain?

GalacticHypernova:
> Have you done any measurements that suggest this is a meaningful performance gain?

I can run benchmarks for this for sure, I would assume it's similar to #24715

packages/nuxt/src/core/app.ts
GalacticHypernova: Thanks for the review! I will run a few benchmarks if you don't mind to see whether or not this is an improvement or not.

GalacticHypernova:
@danielroe @pi0 
image

According to the image, it appears that there is a slight performance improvement. I agree that while it's a very little gain, it is still a gain. I'll leave that up to your guys' decision though.

GalacticHypernova:
Update: offline tests present an even bigger improvement in the approach I took:
image
image

But as I said, I'll leave it to your judgement.

packages/nuxt/src/core/app.ts
GalacticHypernova: I also think that despite the inline-caching there's a limit to what it can do, seeing as there are over 15 duplicate indexes, from my benchmarks it seems like it does decrease the performance.

PR #24730, opened 2023-12-13
perf(nuxt): avoid duplicate iterations over layers

GalacticHypernova:
### ❓ Type of change

### 📚 Description

This PR saves the mapped configs (and their reversed counterpart) to avoid the repetitive operations when generating the app, resulting in a **~81.7%** increase in performance.
Benchmark results (offline, from host PC):
```ts
const nuxt={
    options:{
        _layers:[
            {config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true},{config:true}
        ]
    }
}
console.time("regular")
    for(const config of nuxt.options._layers.map(layer => layer.config)){

    }
}
console.timeEnd("regular")
const layerConfigs=nuxt.options._layers.map(layer => layer.config)
console.time("PR")
    for(const config of layerConfigs){}
}
console.timeEnd("PR")
```

image

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -107,8 +107,9 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   }
 
   // Resolve layouts/ from all config layers
+  const layerConfigs = nuxt.options._layers.map(layer => layer.config)
   app.layouts = {}
-  for (const config of nuxt.options._layers.map(layer => layer.config)) {
+  for (const config of layerConfigs) {
     const layoutDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.layouts || 'layouts'
     const layoutFiles = await resolveFiles(config.srcDir, `${layoutDir}/**/*{${nuxt.options.extensions.join(',')}}`)
     for (const file of layoutFiles) {
@@ -119,7 +120,7 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
 
   // Resolve middleware/ from all config layers, layers first
   app.middleware = []
-  for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+  for (const config of layerConfigs.reverse()) {
     const middlewareDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.middleware || 'middleware'
     const middlewareFiles = await resolveFiles(config.srcDir, `${middlewareDir}/*{${nuxt.options.extensions.join(',')}}`)
     app.middleware.push(...middlewareFiles.map((file) => {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/app.ts
```diff
@@ -157,7 +159,7 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
```
GalacticHypernova, line 159: @danielroe in the 2 lines above, is the re-deduplication of the middleware and the plugins needed in lines 171-172?

danielroe: It is there because users may add their own plugins or middleware on L170.

packages/nuxt/src/core/app.ts
```diff
@@ -107,8 +107,10 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   }
 
   // Resolve layouts/ from all config layers
+  const layerConfigs = nuxt.options._layers.map(layer => layer.config)
+  const reversedConfigs=layerConfigs.reverse()
```
danielroe, line 111: (.reverse mutates the array so you would need to spread it...)

Also, would you be able to match the spacing of the existing code 🙏

GalacticHypernova: Sure thing! I apologize, I've been editing it in place.

packages/nuxt/src/core/app.ts
GalacticHypernova: I see, I'll leave it be then.

GalacticHypernova:
I have reverted the reverse update (no pun intended) for the time being.

packages/nuxt/src/core/app.ts
```diff
@@ -119,7 +120,7 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
 
   // Resolve middleware/ from all config layers, layers first
   app.middleware = []
-  for (const config of nuxt.options._layers.map(layer => layer.config).reverse()) {
+  for (const config of layerConfigs.reverse()) {
```
danielroe, line 123: This will mutate `layerConfigs` meaning it will have side effects which is why the tests are failing.

GalacticHypernova: But would the original options._layers not get mutated as well in that case? I mean, the logic that is applied is the same..

GalacticHypernova: What if we use slice()? Or is the only way of making it not have side effects actually calling map each time?

GalacticHypernova: **I know what can be done**

GalacticHypernova:
Turns out all I needed to do was combine the 2 approaches, that way we still save the time of the second reverse,

danielroe (approved):

PR #24734, opened 2023-12-13
perf(schema): avoid duplicate `get` operations

GalacticHypernova:
### ❓ Type of change

### 📚 Description

This PR saves the app that is gotten through the runtimeConfig $resolve async function and is used in place of the current 3 calls, resulting in a **~99.1%** increase in performance.
Benchmark results (tested with the smallest possible timeout of 0 ms which would, in real time, likely take longer than that):
```ts
async function get(name){
        await new Promise((resolve)=>{
            setTimeout(()=>{},0)
            resolve(void 0)
        })
        return {baseURL:"",buildAssetsDir:"",cdnURL:""}
    }
    const val={}
    console.time("current")
       val.baseURL=(await get('app')).baseURL
       val.buildAssetsDir=(await get('app')).buildAssetsDir
       val.cdnURL=(await get('app')).cdnURL
    }
    console.timeEnd("current")
    const app = await get('app')
    console.time("PR")
        val.baseURL=app.baseURL
        val.buildAssetsDir=app.buildAssetsDir
        val.cdnURL=app.cdnURL
     }
     console.timeEnd("PR")
```     
image

### 📝 Checklist


GalacticHypernova:
I will try to get the rest handled as well, but this was the thing that caught my eye the most.

packages/schema/src/config/common.ts
```diff
@@ -50,7 +50,10 @@ export default defineUntypedSchema({
    * It is normally not needed to configure this option.
    */
   workspaceDir: {
-    $resolve: async (val, get) => val ? resolve(await get('rootDir'), val) : await findWorkspaceDir(await get('rootDir')).catch(() => get('rootDir'))
+    $resolve: async (val, get) => {
+      const rootDir = get('rootDir')
+      return val ? resolve(await rootDir, val) : await findWorkspaceDir(await rootDir).catch(() => rootDir)
```
GalacticHypernova, line 55: @danielroe is the catch supposed to return the promise? Is it not meant to be awaited?

danielroe, line 55:
```suggestion
      const rootDir = await get('rootDir')
      return val ? resolve(rootDir, val) : await findWorkspaceDir(rootDir).catch(() => rootDir)
```

danielroe (approved):
Nice 👌

GalacticHypernova:
Could I request this PR to not be merged until I'm done fixing all of schema? Or is it fine if it gets merged in chunks?

danielroe:
Oh, of course.

GalacticHypernova:
Thank you!
I'll try to get this update ready by the end of the day.

GalacticHypernova:
Looking at the webpack schema, I don't think making it as such is necessarily the most performant:
image

Can this be changed to something like this?
```ts
vue:  {
        const vue = await get('vue')
        return {
            transformAssetUrls: {
              video: 'src',
              source: 'src',
              object: 'src',
              embed: 'src'
            },
            compilerOptions: vue.compilerOptions,
            propsDestructure: Boolean(vue.propsDestructure),
            defineModel: Boolean(vue.defineModel)
        }
    }
}
```
Or would it have some kind of side effects? Because the same thing is done with the runtimeConfig:
image
It provides it from the app already so it doesn't have to get everything separately.

packages/schema/src/config/webpack.ts
```diff
@@ -197,15 +197,20 @@ export default defineUntypedSchema({
...
+              video: 'src',
+              source: 'src',
+              object: 'src',
+              embed: 'src'
+            },
+            compilerOptions: vue.compilerOptions,
+            propsDestructure: Boolean(vue.propsDestructure),
+            defineModel: Boolean(vue.defineModel)
+          })
+        }
```
GalacticHypernova, line 213: @danielroe would a change like this make sense?

GalacticHypernova: I can also just wrap everything in a big defu object, like so:
```ts
export default defineUntypedSchema({
    webpack: {
            const webpack = await get('webpack')
            return defu(val, { /* all the properties in here */ })
        }
    }
})
```
Do you think that would be overkill?

GalacticHypernova: Alright, I'll revert it then. Do you have any guidelines for when it's better to make it like this? (For example, the app section in common.ts is resolved the same way as the webpack change I suggested)

GalacticHypernova: I have reverted the update, which means I *think* it is ready for review.

danielroe (approved):

packages/schema/src/config/common.ts
```diff
@@ -342,15 +351,19 @@ export default defineUntypedSchema({
...
-      '**/*.{spec,test}.{js,cts,mts,ts,jsx,tsx}', // ignore tests
-      '**/*.d.{cts,mts,ts}', // ignore type declarations
-      '**/.{pnpm-store,vercel,netlify,output,git,cache,data}',
-      relative(await get('rootDir'), await get('analyzeDir')),
-      relative(await get('rootDir'), await get('buildDir')),
-      await get('ignorePrefix') && `**/${await get('ignorePrefix')}*.*`
-    ].concat(val).filter(Boolean)
+    $resolve: async (val, get) => {
+      const rootDir = await get('rootDir')
+      const ignorePrefix = await get('ignorePrefix')
```
pi0, line 356: Small perf we could use parallel promises: `const[rootDir, ignorePrefix] = await Promise.all(get('rootDir'), get('ignorePrefix'))`

GalacticHypernova: Oh you're right! Thanks, I missed that. I know another place this can be added in.

GalacticHypernova: Is there a way to "append" commits to a merged PR or should I open a new one?

pi0: You probably need a new one..

PR #24743, opened 2023-12-14
fix(nuxt): defensive coding to fix some static builds

### LINKED ISSUE #24724: Nuxt fails to load runtimeConfig in static Storybook build
### Environment

- Operating System: Linux
- Node Version:     v18.18.2
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.10.5
- Builder:          -
- User Config:      devtools, components, modules, runtimeConfig, app
- Runtime Modules:  @nuxtjs/storybook@7.0.0
- Build Modules:    -

### Reproduction

Branch named `storybook-static` on https://github.com/filiphazardous/nuxt-storybook-module-demo
Run `pnpm i` followed by `pnpm build-storybook && pnpm static-storybook` - open http://localhost:6007 and check the console.

### Describe the bug

The bug happens in the lines https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/templates.ts#L317-L323 (as generated by the Storybook build). `useRuntimeConfig()` on line 321 returns an empty object, which sets `appConfig` on the same line to `undefined`. When `baseURL()` on line 323 is called, the web browser throws an error.

### Additional context

Initially, I reported the bug at https://github.com/nuxt-modules/storybook/issues/458  

I'm not sure if the bug is that Nuxt by default provides an empty object as runtime config, or if the storybook addon erroneously managed to set it to an empty object.  

The problem could be avoided by adding `?.` in a couple of places to the `publicPathTemplate` - or just changing line 321 to `: 'const appConfig = useRuntimeConfig().app || {}',`

### Logs

_No response_

filiphazardous:
### 🔗 Linked issue

#24724

### ❓ Type of change

### 📚 Description

Resolves #24724

In the generated paths template in a static production build, runtime config is used to set
baseURL, cdnURL, etc. However, in some obscure cases, runtime config returns an empty object. In theses cases - reading baseURL etc from `useRuntimeConfig().app.baseURL` throws an error.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -318,7 +318,7 @@ export const publicPathTemplate: NuxtTemplate = {
 
       nuxt.options.dev
         ? `const appConfig = ${JSON.stringify(nuxt.options.app)}`
-        : 'const appConfig = useRuntimeConfig().app || {}',
+        : 'const appConfig = useRuntimeConfig()?.app || {}',
 
       'export const baseURL = () => appConfig.baseURL',
       'export const buildAssetsDir = () => appConfig.buildAssetsDir',
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -318,7 +318,7 @@ export const publicPathTemplate: NuxtTemplate = {
 
       nuxt.options.dev
         ? `const appConfig = ${JSON.stringify(nuxt.options.app)}`
-        : 'const appConfig = useRuntimeConfig().app',
+        : 'const appConfig = useRuntimeConfig().app || {}',
 
       'export const baseURL = () => appConfig.baseURL',
       'export const buildAssetsDir = () => appConfig.buildAssetsDir',


danielroe:
I feel like this is a bug in the storybook module.

It needs to set `window.__NUXT__` like this:

https://github.com/nuxt/nuxt/blob/b8bfa60ece189c8d806c4c37c622f02794d1d947/packages/nuxt/src/core/runtime/nitro/renderer.ts#L539

I think probably it's not rendering this script block, which is why this particular issue popped up.

Although this PR might resolve the particular issue, there will be other problems associated with this `__NUXT__` variable not being instantiated in the window and so I think this should be solved in the storybook module.

packages/nuxt/src/core/templates.ts
```diff
@@ -318,7 +318,7 @@ export const publicPathTemplate: NuxtTemplate = {
 
       nuxt.options.dev
         ? `const appConfig = ${JSON.stringify(nuxt.options.app)}`
-        : 'const appConfig = useRuntimeConfig().app',
+        : 'const appConfig = useRuntimeConfig().app || {}',
```
pi0, line 321:
```suggestion
        : 'const appConfig = useRuntimeConfig()?.app || {}',
```

danielroe:
I'm closing it as this isn't an issue with Nuxt but with the particular storybook implementation. I'm more than happy to help resolve that issue though ❤️

PR #24744, opened 2023-12-14
perf(kit, schema, nuxt): rework startsWith to direct index

GalacticHypernova:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

JS's startWith is known to have a performance downgrade compared to traditional character checking (you can see more [here](https://sun0day.github.io/blog/vite/why-vite4_3-is-faster.html#substitute-startswith-endswith-with) how vite also substituted those). And indeed, the performance increase is very much noticeable (**~20%** increase in performance in single character checks and **~15%** increase in performance from starting 2 characters, increasing further as more characters are introduced).
Benchmark results:
```ts
const nuxt={
    _ignorePatterns:["abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!"]
}
console.time("currentSingle")
for (let index = 0; index < 100000; index++) {
    nuxt._ignorePatterns.map(p => p.startsWith('*'))
}
console.timeEnd("currentSingle")
console.time("PRSingle")
for (let index = 0; index < 100000; index++) {
    nuxt._ignorePatterns.map(p=>p[0]==="*")
}
console.timeEnd("PRSingle")
    console.time("currentM")
for (let index = 0; index < 100000; index++) {
    nuxt._ignorePatterns.map(p => p.startsWith('!*'))
}
console.timeEnd("currentM")
console.time("PRM")
for (let index = 0; index < 100000; index++) {
    nuxt._ignorePatterns.map(p => p[0] === '!' && p[1] === '*')
}
console.timeEnd("PRM")
```

image

[...]

PR #24746, opened 2023-12-14
perf(vite): rework endsWith to direct index

GalacticHypernova:
### 🔗 Linked issue

https://github.com/nuxt/framework/pull/8529

### ❓ Type of change

### 📚 Description

Similar to #24744, endsWith also has a performance downside when checking single characters. This reworks it to direct index in strings, which inccreases performance by **~40%**.
Benchmark results:

```ts
const nuxt={
    _ignorePatterns:["abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!","abcdefghijklmnopqrstuvwxyz*!"]
}
console.time("currentSingle")
    nuxt._ignorePatterns.map(p => p.endsWith('*'))
}
console.timeEnd("currentSingle")
console.time("PRSingle")
    nuxt._ignorePatterns.map(p=>p[p.length-1]==="*")
}
console.timeEnd("PRSingle")
    console.time("currentM")
    nuxt._ignorePatterns.map(p => p.endsWith('!*'))
}
console.timeEnd("currentM")
console.time("PRM")
    nuxt._ignorePatterns.map(p => p[p.length-2] === '!' && p[p.length-1] === '*')
}
console.timeEnd("PRM")
```

image

And yes, here too we can make a tradeoff of partial readability for performance if needed.

### 📝 Checklist


packages/vite/src/plugins/composable-keys.ts
```diff
@@ -116,7 +116,8 @@ export const composableKeysPlugin = createUnplugin((options: ComposableKeysOptio
           }
 
           // TODO: Optimize me (https://github.com/nuxt/framework/pull/8529)
-          const endsWithComma = code.slice(codeIndex + (node as any).start, codeIndex + (node as any).end - 1).trim().endsWith(',')
+          const newCode = code.slice(codeIndex + (node as any).start, codeIndex + (node as any).end - 1).trim()
```
GalacticHypernova, line 119: @danielroe this entire slice/trim operation, is it even needed? Can we not just
```ts
code[codeIndex + (node as any).end - 1]===','
```

danielroe: yes, I think it's possible it will end with whitespace...

GalacticHypernova: Hmm, I'll try to run some tests with this. There's gotta be a way to make it work without it..

danielroe: you could iterate backwards in a while loop... 🤔

GalacticHypernova: That could be possible, it will at the very least reduce the amount of characters "digested" by trim I suppose, but I wonder if there's a way to make it even faster, like if the expression itself will not contain whitespace, in which case last index could be used directly.

GalacticHypernova: image

For starter optimization, we can use trimEnd which increase performance by ~50%

GalacticHypernova:
@danielroe would you like me to also fix this in tests? Or keep this in packages only?

danielroe:
We can ignore tests. 👌

GalacticHypernova:
:+1:

GalacticHypernova:
Alright. I'm getting some odd results and even odder deviations in my analysis. We should count on offline benchmarks, right? Ones made on a PC rather than on the browser?

danielroe:
Benchmarks should be on node and bun, for changes to kit/schema/nuxt build.

I would take perf optimisations with an extreme grain of salt as runtimes frequently optimise newer features in future... In general avoiding sweeping statements.

Ideally reproducible benchmarks should be provided - might be worth using something like codspeed, and adding a comment to remove when things change.

GalacticHypernova:
I had a feeling it was just the online benchmarks faking because on node it is marginally faster.


> might be worth using something like codspeed, and adding a comment to remove when things change.

I can do that, but considering endsWith was implemented in ECMA6 and hasn't been optimized since i can say pretty cerainly this isn't likely to change anytime soon.

GalacticHypernova:
Well, a little underwhelming that there was only one thing to change that actually makes any sort of difference, but I guess this happens =)

danielroe (approved):

PR #24771, opened 2023-12-15
perf(schema): use parallel promises

GalacticHypernova:
### 🔗 Linked issue

#24734 

### ❓ Type of change

### 📚 Description

Following the linked PR, @pi0 brought to my attention that we can use parallel promises, and indeed the benchmark shows a significant **~30%** improvement in performance, so all credits of this PR are to him.
```ts
async function main(){
async function get(name:string){
    return 42;
  }
  console.time("current")
  for (let index = 0; index < 1000000; index++) {
   const a = await get('a')
   const b = await get('b')
   const c = await get('c')
   const d = await get('d')
    
  }
console.timeEnd("current")
console.time("PR")
for (let index = 0; index < 1000000; index++) {
    const [a,b,c,d]=await Promise.all([get('a'), get('b'), get('c'), get('d')])
    
  }
  console.timeEnd("PR")
}
main()
```
image

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -136,10 +136,9 @@ export default defineUntypedSchema({
     $default: ['node_modules'],
     $resolve: async (val, get) => {
       const rootDir = await get('rootDir')
-      const currentDir = process.cwd()
       return [
         ...await Promise.all(val.map(async (dir: string) => resolve(rootDir, dir))),
-        resolve(currentDir, 'node_modules')
+        resolve(process.cwd(), 'node_modules')
       ]
     }
   },
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -313,7 +313,7 @@ export default defineUntypedSchema({
    */
   alias: {
     $resolve: async (val, get) => {
-      const [srcDir, rootDir, dirAssets, dirPublic] = await Promise.all(get('srcDir'), get('rootDir'), get('dir.assets'), get('dir.public'))
+      const [srcDir, rootDir, dirAssets, dirPublic] = await Promise.all([get('srcDir'), get('rootDir'), get('dir.assets'), get('dir.public')])
       return {
         '~': srcDir,
         '@': srcDir,
@@ -353,7 +353,7 @@ export default defineUntypedSchema({
    */
   ignore: {
     $resolve: async (val, get) => {
-      const[rootDir, ignorePrefix] = await Promise.all(get('rootDir'), get('ignorePrefix'))
+      const[rootDir, ignorePrefix] = await Promise.all([get('rootDir'), get('ignorePrefix')])
       return [
         '**/*.stories.{js,cts,mts,ts,jsx,tsx}', // ignore storybook files
         '**/*.{spec,test}.{js,cts,mts,ts,jsx,tsx}', // ignore tests
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -134,10 +134,14 @@ export default defineUntypedSchema({
    */
   modulesDir: {
     $default: ['node_modules'],
-    $resolve: async (val, get) => [
-      ...await Promise.all(val.map(async (dir: string) => resolve(await get('rootDir'), dir))),
-      resolve(process.cwd(), 'node_modules')
-    ]
+    $resolve: async (val, get) => {
+      const rootDir = await get('rootDir')
+      const currentDir = process.cwd()
+      return [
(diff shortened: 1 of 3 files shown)


packages/schema/src/config/common.ts
```diff
@@ -309,10 +313,7 @@ export default defineUntypedSchema({
    */
   alias: {
     $resolve: async (val, get) => {
-      const srcDir = await get('srcDir')
-      const rootDir = await get('rootDir')
-      const dirAssets = await get('dir.assets')
-      const dirPublic = await get('dir.public')
+      const [srcDir, rootDir, dirAssets, dirPublic] = await Promise.all(get('srcDir'), get('rootDir'), get('dir.assets'), get('dir.public'))
```
pi0: Code style: usually variables names are like `assetsDir` and `publicDir` in reverse

GalacticHypernova: Thanks!

GalacticHypernova:
Pretty sure that's everything

packages/schema/src/config/common.ts
```diff
@@ -134,10 +134,14 @@ export default defineUntypedSchema({
    */
   modulesDir: {
     $default: ['node_modules'],
-    $resolve: async (val, get) => [
-      ...await Promise.all(val.map(async (dir: string) => resolve(await get('rootDir'), dir))),
-      resolve(process.cwd(), 'node_modules')
-    ]
+    $resolve: async (val, get) => {
+      const rootDir = await get('rootDir')
+      const currentDir = process.cwd()
```
danielroe, line 139: I don't think this needs to be split out; it's not async or used more than once.

GalacticHypernova: Oh you're right, my bad. I thought for a second that it was part of the promise, I apologize!

packages/schema/src/config/vite.ts
```diff
@@ -39,7 +39,8 @@ export default defineUntypedSchema({
         if (val) {
           consola.warn('Directly configuring the `vite.publicDir` option is not supported. Instead, set `dir.public`. You can read more in `https://nuxt.com/docs/api/nuxt-config#public`.')
         }
-        return val ?? resolve((await get('srcDir')), (await get('dir')).public)
+        const [srcDir, dir] = await Promise.all([get('srcDir'), get('dir')])
+        return val ?? resolve(srcDir, dir.public)
```
danielroe, line 43: This will always resolve these even if they aren't needed - I think you could amend to:

```suggestion
        return val ?? await Promise.all([get('srcDir'), get('dir')])
.then(([srcDir, dir]) => resolve(srcDir, dir.public))
```

GalacticHypernova: I have not tested .then performance, does it add any sort of overhead? I can run some benchmarks real quick

packages/schema/src/config/vite.ts
```diff
@@ -92,14 +93,17 @@ export default defineUntypedSchema({
...
-            await get('buildDir'),
-            await get('srcDir'),
-            await get('rootDir'),
-            await get('workspaceDir'),
-            ...(await get('modulesDir')),
-            ...val ?? []
-          ]
+          $resolve: async (val, get) => {
+            const [buildDir, srcDir, rootDir, workspaceDir, modulesDir] = await Promise.all([get('buildDir'), get('srcDir'), get('rootDir'), get('workspaceDir'), get('modulesDir')])
+              [
```
danielroe, line 98:
```suggestion
            return [
```

GalacticHypernova: Thanks! I missed it.

packages/schema/src/config/vite.ts
```diff
@@ -92,14 +92,17 @@ export default defineUntypedSchema({
...
+            const [buildDir, srcDir, rootDir, workspaceDir, modulesDir] = await Promise.all([get('buildDir'), get('srcDir'), get('rootDir'), get('workspaceDir'), get('modulesDir')])
+            return  [
+                buildDir,
+                srcDir,
+                rootDir,
+                workspaceDir,
+                ...(modulesDir),
+                ...val ?? []
+              ]
+          }
```
danielroe, line 105:
```suggestion
            return [
              buildDir,
              srcDir,
              rootDir,
              workspaceDir,
              ...(modulesDir),
              ...val ?? []
            ]
          }
```

danielroe (approved):

GalacticHypernova:
Ah I see, looking back at it. I could've sworn without parentheses was fine when I tested but I mistook it for something

PR #24781, opened 2023-12-15
feat(nuxt): `useRequestHeader` utility

GalacticHypernova:
### ❓ Type of change

### 📚 Description

The current request header retrival process of `useRequestHeaders` is a bit unintuitive in situations that require reading headers. It doesn't make a lot of sense from a developmental standpoint as more often than not you would need only a single header, mostly for "reading" operations (I myself have a couple of such use cases, for example, authorization middleware).
This PR provides that possibility to simplify and improve DX. With this a person could simply do
```ts
const cookie = useRequestHeader('cookie')
```
That being said, I would like to reiterate and clarify that this is **not** a replacement of `getRequestHeaders`. This can be used to directly read header values when dealing with authorization related middleware for example, whereas getRequestHeaders can be used to proxy the headers and use them in API calls, due to its object return.

### 📝 Checklist

### DIFF
--- a/docs/3.api/2.composables/use-request-header.md
+++ b/docs/3.api/2.composables/use-request-header.md
@@ -16,7 +16,7 @@ const authorization = useRequestHeader('authorization')
 ```
 
 ::callout
-In the browser, `useRequestHeader` will return an empty string.
+In the browser, `useRequestHeader` will return `undefined`.
 ::
 
 ## Example
--- a/docs/3.api/2.composables/use-request-header.md
+++ b/docs/3.api/2.composables/use-request-header.md
@@ -11,7 +11,7 @@ links:
 You can use the built-in [`useRequestHeader`](/docs/api/composables/use-request-header) composable to access any incoming request header within your pages, components, and plugins.
 
 ```ts
-// Get the authorization request header
+// Get the cookie request header
 const cookie = useRequestHeader('cookie')
 ```
 
--- a/docs/3.api/2.composables/use-request-header.md
+++ b/docs/3.api/2.composables/use-request-header.md
@@ -8,7 +8,7 @@ links:
     size: xs
 ---
 
-You can use built-in [`useRequestHeader`](/docs/api/composables/use-request-header) composable to access any incoming request header within your pages, components, and plugins.
+You can use the built-in [`useRequestHeader`](/docs/api/composables/use-request-header) composable to access any incoming request header within your pages, components, and plugins.
 
 ```ts
 // Get the authorization request header
--- a/docs/3.api/2.composables/use-request-header.md
+++ b/docs/3.api/2.composables/use-request-header.md
@@ -25,7 +25,7 @@ We can use `useRequestHeader` to easily figure out if a user is authorized or no
 
 The example below reads the `authorization` request header to an isomorphic `$fetch` call.
 
-```vue [middleware/authorized-only.ts]
+```ts [middleware/authorized-only.ts]
 export default defineNuxtRouteMiddleware((to, from) => {
   if(!useRequestHeader('authorization')){
     return navigateTo('/not-authorized')
--- a/docs/3.api/2.composables/use-request-header.md
+++ b/docs/3.api/2.composables/use-request-header.md
@@ -12,7 +12,7 @@ You can use built-in [`useRequestHeader`](/docs/api/composables/use-request-head
 
 ```ts
 // Get the authorization request header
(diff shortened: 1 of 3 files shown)


GalacticHypernova:
Marking this a draft to write the documentation.

GalacticHypernova:
Should I also clarify in the docs that this is used as an abstraction of `getRequestHeaders` in specific header-related operations (like the example I gave in the docs)?

docs/3.api/2.composables/use-request-header.md
```diff
@@ -0,0 +1,34 @@
...
+    icon: i-simple-icons-github
+    to: https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/ssr.ts
+    size: xs
+---
+
+You can use the built-in [`useRequestHeader`](/docs/api/composables/use-request-header) composable to access any incoming request header within your pages, components, and plugins.
+
+```ts
+// Get the cookie request header
+const cookie = useRequestHeader('cookie')
```
danielroe, line 15: I'd recommend using a different header as an example as we already have `useCookie`. Maybe `authorization`?

docs/3.api/2.composables/use-request-header.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Example
+
+We can use `useRequestHeader` to easily figure out if a user is authorized or not.
+
+The example below reads the `authorization` request header to find out if a person can access a restricted resource.
+
+```ts [middleware/authorized-only.ts]
+export default defineNuxtRouteMiddleware((to, from) => {
+  if(!useRequestHeader('authorization')){
```
danielroe, line 30:
```suggestion
  if (!useRequestHeader('authorization')) {
```

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -17,6 +17,13 @@ export function useRequestHeaders (include?: any[]) {
   return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
 }
 
+export function useRequestHeader(include: string) {
+  if (import.meta.client) { return '' }
+  const event = useRequestEvent()
+  const header = event ? getRequestHeader(event, include) : ''
+  return header
```
danielroe, line 24:
```suggestion
  if (import.meta.client) { return undefined }
  const event = useRequestEvent()
  return event ? getRequestHeader(event, include) : undefined
```

danielroe:
Sure, it would be good to point people to the upstream utility used 👍

packages/nuxt/src/app/composables/ssr.ts
GalacticHypernova: Should this return undefined or an empty string?

docs/3.api/2.composables/use-request-header.md
GalacticHypernova: Done, I added authorization in stead, thanks!

docs/3.api/2.composables/use-request-header.md
GalacticHypernova: Thank you!

Atinux:
Thank you! Indeed it is useful to be able to get one specific header and mention that it uses `useRequestHeaders` under the hood 👍

packages/nuxt/src/app/composables/ssr.ts
Atinux: I would prefer returning an empty string for consistency.

danielroe: I think undefined is mroe consistent as `getRequestHeader` returns undefined if a header doesn't exist, as does `useRequestHeaders(['unknown-header'])['unknown-header']`.

docs/3.api/2.composables/use-request-header.md
```diff
@@ -0,0 +1,34 @@
...
+
+You can use the built-in [`useRequestHeader`](/docs/api/composables/use-request-header) composable to access any incoming request header within your pages, components, and plugins.
+
+```ts
+// Get the authorization request header
+const authorization = useRequestHeader('authorization')
+```
+
+::callout
+In the browser, `useRequestHeader` will return an empty string.
```
DamianGlowala, line 19:
```suggestion
In the browser, `useRequestHeader` will return `undefined`.
```

danielroe (approved):

Issue #21855, opened 2023-06-29
Nuxt Render Composables


Hebilicious:
### Describe the feature

related issues
https://github.com/nuxt/nuxt/issues/14507
https://github.com/nuxt/nuxt/issues/21370
https://github.com/nuxt/nuxt/issues/13949

# Motivation : 

There's isn't a good API to define SSG/SSR logic that should happen within a page/component. When we communicate and talk about rendering as developers, we use the terms SSR and SSR and these terms have become widely used and understood.

Currently, we can use `process.server`, as well as vues `onServerPrefetch` to define server-only logic. However, conditional logic with `process` is "heavy handed" and will be applied while pre-rendering and ssr-ing, and process is technically semantically incorrect for non node runtimes. Additionally there's no way to specify something like caching.
[`onServerPrefetch`](https://vuejs.org/api/composition-api-lifecycle.html#onserverprefetch) is a "low level" utility, and its timing is different than `process.server`.
AFAIK, it's also currently not possible to pass ssg data from a nuxt hook to a component/page.

# Proposal: 

Introduce new hooks to help developer express logic, `onSSG` and `onSSR`.

`onSSG` can be used for SSG logic and runs at build time only. It can be used to control the pre-rendering logic. The common scenario of fetching n pages (such as products/blog articles) to pre-render a home page for example.
`onSSR` (and `onClient`) can be used as semantically correct replacements of `if(process.server){}`
Design : 

```ts
onSSG(()=> {
console.log("This runs at build time only")
})

onSSR(()=> {
console.log("This runs every time the nuxt app is rendered server side")
})

onClient(() => {
console.log("This runs on client only")
})
```

# Example : 

Logic to fetch posts : 
```ts
const getPosts = async () => {
    const endpoint = 'https://jsonplaceholder.typicode.com/posts'
    const response = await $fetch(endpoint)
    const articles = response.data.slice(0, 10);
    const slugs = articles.map((each) => {
        return '/articles/' + each.id
    })
    return {articles, slugs}
}
```

Before, this had to happen within a nuxt hook, in nuxt.config or in a module.
```ts
export default defineNuxtConfig({
    hooks: {
        async 'nitro:config'(nitro) {
            if (nitro.dev) return
            const { slugs } = await getPostRoutes();
            nitro.prerender.routes.push(...slugs)
        }
    }
})
```

After this can happen in any page 
Payload data can be passed down to other components while pre-rendering

articles/index.vue
```html
<script setup>
[...]

PR #24787, opened 2023-12-15
feat(nuxt): add `callOnce` util to allow running code only once

atinux:
### 🔗 Linked issue

Should help to resolves many issues with Pinia store actions or anything code that does not return data but needs to be run once.

### ❓ Type of change

### 📚 Description

The `callonce` function is designed to execute a given function or block of code only a single time in a specific scenario:
- During server-side rendering but not during hydration
- On client-side navigation

This is useful for code that should only be executed once, such as logging an event or setting up a global state.

```vue [app.vue]
<script setup>
const websiteConfig = useState('config')

await callOnce(async () => {
  console.log('This will only be logged once')
  websiteConfig.value = await $fetch('https://my-cms.com/api/website-config')
})
</script>
```

Note that `stats` won't be reactive in this case and if an error happen you will [have to handle it](/docs/getting-started/error-handling#error-page), if you want to do data fetching and have it be reactive, you should use [`useAsyncData`](/docs/api/composables/use-async-data) or [`useFetch`](/docs/api/composables/use-fetch).

Example on Stackblitz: https://stackblitz.com/edit/nuxt-once?file=app.vue

### 📝 Checklist


docs/1.getting-started/7.state-management.md
```diff
@@ -57,9 +57,67 @@ const counter = useState('counter', () => Math.round(Math.random() * 1000))
...
+::
+
+:read-more{to="/docs/api/utils/once"}
+
+### Usage with Pinia
+
+In this example, we leverage the [Pinia module](/modules/pinia) to create a global store and use it across the app.
+
+::callout
+Make sure to install the Pinia module with `npx nuxi@latest module add pinia`
```
GalacticHypernova, line 85:
```suggestion
Make sure to install the Pinia module with `npx nuxi@latest module add pinia @pinia/nuxt`
```
Pinia has an official module that I feel like should also be included, as that's the module that goes in the `modules` array.

GalacticHypernova, line 85: Should this also include regular installation steps for those that don't use the cli for whatever reason?

Atinux: I link to the Pinia module actually so I think that should be enough no?

GalacticHypernova: I guess so. Perhaps it can say something like "or follow the module's installation stete"? Some people might mistake it for a typo in the docs if it's not made clear that it's equivalent, especially those whose native language isn't English.

GalacticHypernova:
It looks promising, but what would be the difference between this and nuxt 
.server plugins for example? You stated in the docs that it can replace nuxtServerInit, but can't server plugins do this as well? Is this basically syntactic sugar?

docs/1.getting-started/7.state-management.md
Atinux: Updated

Atinux:
@GalacticHypernova server plugins can do this but it is harder to understand than making a global fetch in the main component .

You can also use `once` in pages and components.

GalacticHypernova:
I see. That's actually gonna be pretty useful for things that shouldn't require another plugin

Atinux:
It does not act only a server plugin, the code inside `once` will run during server-side rendering **but not during hydration** if the code ran when rendering the page on the server. Then, if you go to another page that uses `once`, the code will be executed on client-side.

We make a demo to showcase.

Atinux:
Example on Stackblitz: https://stackblitz.com/edit/github-ljcghe?file=app.vue,pages%2Findex.vue,pages%2Fabout.vue,stores%2Fapp.ts,package.json

GalacticHypernova:
I see, this is actually even more useful than I originally comprehended. It's like an expansion of server plugins, that's very nice 👍

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -149,7 +149,8 @@ const getSPARenderer = lazyCachedFunction(async () => {
       _errors: {},
       serverRendered: false,
       data: {},
-      state: {}
+      state: {},
+      once: {}
```
danielroe, line 153: I wonder if we use instead a set of ids rather than kv pairs?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -224,6 +225,7 @@ export function createNuxtApp (options: CreateOptions) {
     payload: reactive({
       data: {},
```
danielroe, line 226: the payload (when payload extraction is enabled) will only include `data`.

if that's the desired behaviour (that is, the same as `useState`) then we should probably document it. we could also make it clearer by storing it _within_ state as `state.once`

docs/1.getting-started/7.state-management.md
```diff
@@ -57,9 +57,67 @@ const counter = useState('counter', () => Math.round(Math.random() * 1000))
 To globally invalidate cached state, see [`clearNuxtState`](/docs/api/utils/clear-nuxt-state) util.
 ::
 
-### Advanced Usage
+### Init with Async Data
```
danielroe, line 60:
```suggestion
### Initializing State
```

docs/3.api/3.utils/once.md
```diff
@@ -0,0 +1,50 @@
...
+::callout{to="/docs/getting-started/state-management#usage-with-pinia"}
+`once` is useful in combination with the [Pinia module](/modules/pinia) to call store actions.
+::
+
+:read-more{to="/docs/getting-started/state-management"}
+
+::callout{color="info" icon="i-ph-warning-duotone"}
+Note that `once` won't return anything, you should use [`useAsyncData`](/docs/api/composables/use-async-data) or [`useFetch`](/docs/api/composables/use-fetch) if you want to do data-fetching during SSR.
+::
+
```
danielroe, line 41:
```suggestion

::callout
`once` is a composable meant to be called directly in a setup function, plugin, or route middleware, because it needs to add data to the Nuxt payload to avoid re-calling the function on the client when the page hydrates.
::

```

packages/nuxt/src/imports/presets.ts
```diff
@@ -38,7 +38,7 @@ const granularAppPresets: InlinePreset[] = [
     from: '#app/composables/asyncData'
   },
   {
-    imports: ['useHydration'],
+    imports: ['once', 'useHydration'],
```
danielroe, line 41: we could also make this its own file:  `#app/composables/once`

danielroe (commented):
I love it!

Also linking https://github.com/nuxt/nuxt/issues/21855 as it's similar in scope.

test/basic.test.ts
```diff
@@ -916,6 +916,12 @@ describe('middlewares', () => {
     expect(html.headers.get('location')).toEqual('/')
     expect(html.status).toEqual(307)
   })
+  it('should run code once', async () => {
+    const html = await $fetch('/once')
+
+    expect(html).toContain('once.vue')
+    expect(html).toContain('once: 1')
```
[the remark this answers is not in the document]
Atinux: Could you guide me how to do this?

packages/nuxt/src/core/runtime/nitro/renderer.ts
Atinux: Yes we should use Set(), was not sure about the hydration with the new json payload

packages/nuxt/src/app/composables/hydrate.ts
```diff
@@ -1,6 +1,37 @@
...
+  if (!_key || typeof _key !== 'string') {
+    throw new TypeError('[nuxt] [once] key must be a string: ' + _key)
+  }
+  if (fn !== undefined && typeof fn !== 'function') {
+    throw new Error('[nuxt] [once] fn must be a function: ' + fn)
+  }
+  const nuxt = useNuxtApp()
+  if (import.meta.server) {
+    await fn()
+    // Store it no error is thrown
```
DamianGlowala, line 25: This comment might need a rewording.

packages/nuxt/src/app/composables/hydrate.ts
```diff
@@ -1,6 +1,37 @@
...
```
Atinux, line 25:
```suggestion
```

packages/nuxt/src/app/composables/hydrate.ts
```diff
@@ -1,6 +1,36 @@
...
+export function once (fn?: (() => any | Promise<any>)): Promise<void>
+export async function once (...args: any): Promise<void> {
+  const autoKey = typeof args[args.length - 1] === 'string' ? args.pop() : undefined
+  if (typeof args[0] !== 'string') { args.unshift(autoKey) }
+  const [_key, fn] = args as [string, (() => any | Promise<any>)]
+  if (!_key || typeof _key !== 'string') {
+    throw new TypeError('[nuxt] [once] key must be a string: ' + _key)
+  }
+  if (fn !== undefined && typeof fn !== 'function') {
+    throw new Error('[nuxt] [once] fn must be a function: ' + fn)
```
posva, line 20: Is there a reason this one is an `Error` while the one above is a `TypeError`?

docs/1.getting-started/7.state-management.md
```diff
@@ -57,9 +57,68 @@ const counter = useState('counter', () => Math.round(Math.random() * 1000))
 To globally invalidate cached state, see [`clearNuxtState`](/docs/api/utils/clear-nuxt-state) util.
 ::
 
-### Advanced Usage
+### Initializing State
 
-In this example, we use a composable that detects the user's default locale from the HTTP request headers and keeps it in a `locale` state.
+Most of the time you will want to initialize your state with async data, you can use the [`app.vue`](/docs/guide/directory-structure/app) component to do so with the [`once`](/docs/api/utils/once) util.
```
DamianGlowala, line 62:
```suggestion
Most of the time, you will want to initialize your state with data that resolves asynchronously. You can use the [`app.vue`](/docs/guide/directory-structure/app) component to do so with the [`once`](/docs/api/utils/once) util.
```

docs/1.getting-started/7.state-management.md
```diff
@@ -57,9 +57,68 @@ const counter = useState('counter', () => Math.round(Math.random() * 1000))
...
+const websiteConfig = useState('config')
+
+await once(async () => {
+  websiteConfig.value = await $fetch('https://my-cms.com/api/website-config')
+})
+</script>
+```
+
+::callout
+This is similar to Nuxt 2 `nuxtServerInit` action to fill the store with data before rendering the page.
```
DamianGlowala, line 75: Would be nice to add a link for anyone unfamiliar with Nuxt 2 :) Reworded it slightly, feel free to modify even further if needed!
```suggestion
This is similar to the [`nuxtServerInit` action](https://v2.nuxt.com/docs/directory-structure/store/#the-nuxtserverinit-action) in Nuxt 2, which allows filling the initial state of your store server-side before rendering the page on the client side.
```

posva:
Given how generic the name `once` is, I think it would be important to choose a different name. **Starting with `use` would also make sense** given it's supposed to be used within components.
This method is only useful for SSR apps, so making that part more apparent in the name would benefit its discovery and understanding when reading it somewhere for the first time. Names that use server-specific terms in the name, such as `useFromServer()`, `useServerDedupe()` or `useHydrated()`, wouldn't be used by someone building an SPA with `ssr: false`.
I'm only giving these names based on words that make me think of this feature: server, deduping, hydration, ensure.

packages/nuxt/src/app/nuxt.ts
Atinux: I don't think it makes sense regarding `state.once` since you can also use this to log things and they may not be related to state management.

It is fine being outside data anyway since only `useFetch` or `useAsyncData` should be extracted for payload extraction.

PS: I tested an example with `nuxt generate` and it works like intended.

packages/nuxt/src/app/composables/hydrate.ts
Atinux: It is similar to https://github.com/nuxt/nuxt/blob/a82a724baafb18dca0b07e3e6a7d31d54cec95e6/packages/nuxt/src/app/composables/state.ts#L21

For me I am ok to update it to TypeError, let's see what @danielroe thinks about it

Atinux:
I do like `useServerDedupe()`, one thing though is composable should not return a promise but I guess with Nuxt we already have `useFetch` and `useAsyncData` that return promises.

danielroe:
I think this is also relevant for a non SSR app.

For example, it would only run once no matter how many times a component mounts, or how many times you navigate to a page.

That could be very useful, but not be conceptually related to server rendering.

Atinux:
Renamed to `callOnce()`

Atinux:
**lint checker** shall be ok once merged

DamianGlowala:
> Renamed to `callOnce()`

Shall the composable's filename be updated accordingly?

Atinux:
That's what I did @DamianGlowala

DamianGlowala:
This is what I meant more precisely:
image

Atinux:
Not sure for this since we don't follow prefix the files (ex: `asyncData.ts`, `state.ts`, etc)

DamianGlowala:
Ah, you're right 💯 Forgot that composables with a prefix other than `use` don't include it in the filename either.

packages/nuxt/src/app/composables/once.ts
```diff
@@ -0,0 +1,28 @@
+import { useNuxtApp } from '../nuxt'
+
+/**
+ * An SSR-friendly utility to call a method once
+ * @param key a unique key ensuring the function can be properly de-duplicated across requests
+ * @param fn a function to call
+ * @see https://nuxt.com/docs/api/utils/once
```
DamianGlowala, line 7:
```suggestion
 * @see https://nuxt.com/docs/api/utils/call-once
```

DamianGlowala (approved):

posva:
> For example, it would only run once no matter how many times a component mounts, or how many times you navigate to a page.

I thought it would only apply the first time. e.g. after hydrating, any successive visit would rerun the callback. Otherwise, people might end up with outdated, stale data, especially on SSG apps. Or maybe I'm missing something.

danielroe:
@posva I thought it would be more useful only to run once. But you're pretty much the expert here on stores 😆

packages/nuxt/src/app/composables/once.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (fn !== undefined && typeof fn !== 'function') {
+    throw new Error('[nuxt] [callOnce] fn must be a function: ' + fn)
+  }
+  const nuxt = useNuxtApp()
+  // If key already ran
+  if (nuxt.payload.once.has(_key)) {
+    return
+  }
+  await fn()
+  nuxt.payload.once.add(_key)
```
danielroe, line 27: Because the key isn't added until after the function completes, it's possible for the function to be run more than once if called again before it finishes.

Instead, we should probably store the promise while it is running and return the original promise if `callOnce` is called again while running.

danielroe: @Atinux I've added an implementation - see what you think.

Atinux: Good for me! I thought about it too, thank you for implementing it ❤️

posva:
@danielroe you are right. I commented after discussing with Sebastien and I didn't have everything clear. `useAsyncData()` already covers the use-case I was talking about

danielroe (approved):

TheAlexLichter (approved):

ovchinnikov-lxs:
Hi,

I found a problem with the `callOnce` function when I use the `@nuxtjs/i18n` package with a URL change strategy when changing the locale.

According to the documentation, the key is generated based on the file name and line number, and I noticed that it is not called again on the `pages/index.vue` (or others) when changing the route from `/` to `/en`. A temporary solution was to create a dynamic key that depends on the current locale. However, is it worth using such a solution?

I am considering creating a wrapper for `callOnce`, where I could modify the key. However, this may not seem like the most elegant solution, as it looks like a hack. What do you think? Will this be an appropriate solution for my project or do I need to create an issue for the `callOnce` fix, since `useAsyncData` works correctly when changing locale.

GalacticHypernova:
@ovchinnikov-lxs callOnce is meant to only run once, if you make a navigation it won't trigger it again, it's by design.

Issue #24834, opened 2023-12-20
In `addComponent` the `mode` is not resolved from `filePath`


becem-gharbi:
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-w5kk6d?file=modules%2Fmy-module%2Findex.ts

### Describe the bug

As described in [docs](https://nuxt.com/docs/guide/directory-structure/components#client-components), adding `.client` suffix to the component filename makes it client only. However, when registering a component with `addComponent` this feature is not working. 

Explicitly setting `mode` to `client` works well. However I think this is a bug given that `addPlugin` does not have this problem. 

Adding a normalization step similar to [this](https://github.com/nuxt/nuxt/blob/a073bec6b1b0bf82f435345306d7187f7f4f4b0f/packages/kit/src/plugin.ts#L11) may solve the issue.

PR #24835, opened 2023-12-20
fix(kit): ensure `mode` from `filePath` for `addComponent`

becem-gharbi:
### 🔗 Linked issue

#24834 

### ❓ Type of change

### 📚 Description

Regarding the `addComponent` Kit helper, this PR adds support for `mode` resolution from `filePath`.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/components.ts
+++ b/packages/kit/src/components.ts
@@ -31,6 +31,11 @@ export async function addComponent (opts: AddComponentOptions) {
   await assertNuxtCompatibility({ nuxt: '>=2.13' }, nuxt)
   nuxt.options.components = nuxt.options.components || []
 
+  if (!opts.mode) {
+    const [, mode = 'all'] = opts.filePath.match(/\.(server|client)(\.\w+)*$/) || []
+    opts.mode = mode as 'all' | 'client' | 'server'
+  }
+
   // Apply defaults
   const component: Component = {
     export: opts.export || 'default',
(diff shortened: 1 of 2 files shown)


TheAlexLichter (approved):

packages/kit/src/components.ts
```diff
@@ -31,6 +31,11 @@ export async function addComponent (opts: AddComponentOptions) {
   await assertNuxtCompatibility({ nuxt: '>=2.13' }, nuxt)
   nuxt.options.components = nuxt.options.components || []
 
+  if (!opts.mode) {
+    const [, mode = 'all'] = opts.filePath.match(/\.(server|client)(\.\w+)*$/) || []
+    opts.mode = mode as 'all' | 'client' | 'server'
```
manniL, line 36: As we use the mode quite a bit now ([plugins](https://github.com/nuxt/nuxt/blob/a073bec6b1b0bf82f435345306d7187f7f4f4b0f/packages/kit/src/plugin.ts#L37-L38) + components), I wonder if an extra type would help in case more modes will be added later on.

danielroe (approved):

packages/kit/src/components.ts
danielroe: actually, I remembered why it's this way - it's so we have a very visual guide to ensure the regexp matches the type assertion... there is already an actual type for `mode` that this 'checks' against

PR #24845, opened 2023-12-20
perf(nuxt, kit, vite): Enhance Performance with Concurrent Execution of Async Functions 🏎

sanjaiyan-dev:
### 📚 Description

Hello, this pull request aims to provide a minor performance boost by concurrently executing asynchronous functions using `Promise.all`. This approach enhances the utilization of hardware resources, leading to improved efficiency and effectiveness. Thank you for your consideration.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/define.ts
+++ b/packages/kit/src/module/define.ts
@@ -143,10 +143,12 @@ function nuxt2Shims (nuxt: Nuxt) {
         templateVars: templates.templateVars
       }
     }
+    const promises: Promise<void>[] = [];
     for await (const template of virtualTemplates) {
       const contents = await compileTemplate({ ...template, src: '' }, context)
       await fsp.mkdir(dirname(template.dst), { recursive: true })
-      await fsp.writeFile(template.dst, contents)
+      promises.push(fsp.writeFile(template.dst, contents))
     }
+    await Promise.all(promises)
   })
 }
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -296,11 +296,13 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
         }
 
         await fsp.mkdir(join(tempDir, 'meta'), { recursive: true })
-        await fsp.writeFile(join(tempDir, 'latest.json'), JSON.stringify({
-          id: buildId,
-          timestamp: buildTimestamp
-        }))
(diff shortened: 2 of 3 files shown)


packages/kit/src/module/define.ts
```diff
@@ -143,10 +143,12 @@ function nuxt2Shims (nuxt: Nuxt) {
         templateVars: templates.templateVars
       }
     }
+    const promises: Promise<void>[] = [];
     for await (const template of virtualTemplates) {
       const contents = await compileTemplate({ ...template, src: '' }, context)
       await fsp.mkdir(dirname(template.dst), { recursive: true })
```
GalacticHypernova, line 149: We can probably place this in another Promise.all that would run before the Promise.all(promises) if we want to really speed it up.

GalacticHypernova:
image

Appears to be a bit mixed, but I do agree that for the initial installation it would make a difference

sanjaiyan-dev:
> image
> 
> Appears to be a bit mixed with the current method being sometimes faster, but I do agree that for the initial installation it would make a difference Also, due to the similarity this has in nature to #24771 , is it possible to perhaps modify that changelog to include this? I don't mind if all credit would go to sanjaiyan and my commit would get overwritten, or if there's a way to combine them (commit1 + commit2) in the changelog, since the schema update also did this, so it would only make sense to group it in some way.

Hi @GalacticHypernova ,

Thank you for acknowledging the credit, but your contribution is more significant.
I still wonder why concurrent processing experienced a small performance drop ‽

GalacticHypernova:
> I still wonder why concurrent processing experienced a small performance drop ‽

I wouldn't count on it, it seems to give out very inaccurate reports.

sanjaiyan-dev:
> I wouldn't count on it, it seems to give out very inaccurate reports.

Understood, thank you.

danielroe (changes requested):
we did a small benchmark on this but couldn't find that making FS access parallel was beneficial as writing to the file system generally has to be done sequentially anyway. (Though compiling the templates in parallel seems useful.)

The bit you've changed here is for Nuxt 2. The way we improved this for Nuxt 3 here:

https://github.com/nuxt/nuxt/blob/5877e11c89fa3f6cf5699ca1a96d1f8b67605f8f/packages/nuxt/src/core/app.ts#L43-L75.

I'd welcome a PR that ports that improvement to Nuxt 2 within `packages/kit/src/module/define.ts`.

(We can skip the change in `packages/nuxt/src/core/nitro.ts`.)

Marking as draft just for my own triage - feel free to unmark as draft whenever it's ready to review 🙏

PR #24853, opened 2023-12-21
perf(nuxt): avoid multiple transformations of headers

GalacticHypernova:
### ❓ Type of change

### 📚 Description

When receiving the headers in useRequestHeaders there are multiple computations being done, which can be combined.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -14,7 +14,10 @@ export function useRequestHeaders (include?: any[]) {
   const event = useRequestEvent()
   const headers = event ? getRequestHeaders(event) : {}
   if (!include) { return headers }
-  return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
+  return Object.fromEntries(include.map(key => {
+    const lowerKey = key.toLowerCase()
+    return [lowerKey, headers[lowerKey]]
+  }).filter([key, value] => value !== undefined))
 }
 
 export function useRequestHeader(header: string) {
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -15,6 +15,7 @@ export function useRequestHeaders (include?: any[]) {
   const headers = event ? getRequestHeaders(event) : {}
   if (!include) { return headers }
   return Object.fromEntries(include.map(key => {
+   // Testing codspeed 
     const lowerKey = key.toLowerCase()
     return [lowerKey, headers[lowerKey]]
   }).filter(([_, value]) => value !== undefined))
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -17,7 +17,7 @@ export function useRequestHeaders (include?: any[]) {
   return Object.fromEntries(include.map(key => {
     const lowerKey = key.toLowerCase()
     return [lowerKey, headers[lowerKey]]
-  }).filter([key, value] => value !== undefined))
+  }).filter(([key, value]) => value !== undefined))
 }
 
 export function useRequestHeader(header: string) {
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -17,7 +17,7 @@ export function useRequestHeaders (include?: any[]) {
   return Object.fromEntries(include.map(key => {
     const lowerKey = key.toLowerCase()
     return [lowerKey, headers[lowerKey]]
-  }).filter(([key, value]) => value !== undefined))
+  }).filter(([_, value]) => value !== undefined))
 }
 
 export function useRequestHeader(header: string) {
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
I will fix the issues soon, currently in a bit of a hurry.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -14,7 +14,10 @@ export function useRequestHeaders (include?: any[]) {
   const event = useRequestEvent()
   const headers = event ? getRequestHeaders(event) : {}
   if (!include) { return headers }
-  return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
+  return Object.fromEntries(include.map(key => {
+    const lowerKey = key.toLowerCase()
+    return [lowerKey, headers[lowerKey]]
+  }).filter([key, value] => value !== undefined))
```
GalacticHypernova, line 20: Is this filter even necessary? If useRequestHeader returns undefined on the browser, does it justify the additional overhead in stead of returning undefined?

GalacticHypernova, line 20: Is this filter even necessary? If useRequestHeader returns undefined on the browser, does it justify the additional overhead in stead of returning undefined?

danielroe: We already return early on browser in L13.

GalacticHypernova: Which is exactly why I'm asking, what does this filter actually do? If it's about removing the unused headers, what would be the benefit of filtering if either way they will get a "undefined-ish" result?

GalacticHypernova:
How come it degraded by 30% from updating the branch?

GalacticHypernova:
There is no way adding a comment fixed the alleged regression. There might be a bit of a misconfiguration in codspeed which is probably worth fixing to not get inaccurate results.

danielroe:
I have already disabled it on the main branch as it seems to give wildly disparate results, probably because we're doing things like interacting with the FS in the benchmarks. We'll be better off using it to benchmark lower-level code we're fully in control of.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -14,7 +14,11 @@ export function useRequestHeaders (include?: any[]) {
   const event = useRequestEvent()
   const headers = event ? getRequestHeaders(event) : {}
   if (!include) { return headers }
-  return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
+  return Object.fromEntries(include.map(key => {
+   // Testing codspeed 
+    const lowerKey = key.toLowerCase()
+    return [lowerKey, headers[lowerKey]]
+  }).filter(([_, value]) => value !== undefined))
```
danielroe, line 21: I think this iterate approach would be more performant still.

```suggestion
  const headers = Object.create(null)
  for (const _key of include) {
    const key = _key.toLowerCase()
    const header = headers[key]
    if (header) {
      headers[key] = header
    }
  }
```

GalacticHypernova, line 21:
```suggestion
})
```
This would just return a list of all the headers where the empty headers would be undefined, right?

danielroe:
(We could, for example, create a benchmark for individual composables.)

packages/nuxt/src/app/composables/ssr.ts
GalacticHypernova: I still have a question about why this sort of filtering at the end is needed.. It will just return undefined either way.

danielroe: It creates a smaller object

GalacticHypernova: I see. I'm gonna run some benchmarks for this

danielroe: It only iterates once instead of 3 times in your example, and 4 in the original code.

GalacticHypernova: @danielroe I made some benchmarks, and Object.create(null) is ~10x slower than {}. Is there a reason we want to avoid the built-in properties for the headers if it's not used outside of the function, meaning for everything else it will be a readonly table?
image

danielroe: it's to avoid prototype pollution. admittedly i can't think of an exploit as the headers should all be _strings_ but it's safer when assigning user-controlled keys to an object. And it's only initialised 1x per `useRequestHeaders()` call.

GalacticHypernova: Well, yea, I guess in such case this bit of performance can be sacrificed in favor of security.

GalacticHypernova: Is this good? Should we maybe use the headers returned from getRequestHeaders?

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -13,9 +13,17 @@ export function useRequestHeaders (): Readonly<Record<string, string>>
 export function useRequestHeaders (include?: any[]) {
   if (import.meta.client) { return {} }
   const event = useRequestEvent()
-  const headers = event ? getRequestHeaders(event) : {}
-  if (!include) { return headers }
-  return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
+  const _headers = event ? getRequestHeaders(event) : {}
+  if (!include) { return _headers }
```
danielroe, line 17:
```suggestion
  if (!include || !event) { return _headers }
```

Issue #23594, opened 2023-10-10
NuxtData causes signficant increase in page render times


rstainsby:
### Environment

- Operating System: Darwin
- Node Version:     v18.16.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.5.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Reproduction repository: https://github.com/rstainsby/nuxt-data-performance-issue.

To reproduce the issue:
- Run the project `npm run dev`
- Click the 'fast page' link at the top of the page and take note of the response time
- Click the 'data load' link and wait for the response to be received (the page will show the status of the request)
- Click the 'fast page' link again and note the response time

### Describe the bug

It appears that `NuxtData`, the caching system utilised by `useFetch` and `useAsyncData` causes a significant drop in page load times, directly correlated to the size of the data being cached i.e. the bigger the cache the slower the response times.

Whilst I would expect this to be the case when caching significantly large objects, it is surprising to see significant impact when 15MB and over is cached.

### Additional context

On my machine, after following the reproduction steps, the initial response time is ~5ms and the second response time (after fetching 17MB of data) is now anywhere between 180 and 250ms for every page navigation, even on a page that is making 0 network requests and is only 20 lines of HTML.

This slow response time is resolvable by calling `clearNuxtData` and does not occur when using `$fetch`, which is what leads me to believe it is directly tied to `NuxtData`.

This issue does not occur if the `useFetch` is client-side only or if the page is rendered in SPA mode but will occur if the page is client-side rendered (navigated to after initial SSR) but has the SRR set to true.

OrbisK:
@danielroe I tried to reproduce the issue, and it seems to have been resolved with Nuxt 4. I was able to reproduce with nuxt 3.

danielroe:
nuxtathon closed @OrbisK

PR #24894, opened 2023-12-25
feat(nuxt): add experimental sharedPrerenderData option

### LINKED ISSUE #18615: Can't persist data between routes with useNuxtData() when using npm nuxi generate

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/18615

### ❓ Type of change

### 📚 Description

This allows automatically sharing a Nuxt payload between initial prerenders, which should significantly improve performance across a very large site with multiple pages that fetch the same data.

~~Note that data will not be shared across the initial batch of concurrent renders, although we could implement this if desired.~~

### Testing

This can be tested in the playground with this `nuxt.config`:

```ts
  experimental: {
    sharedPrerenderData: true
  },
  nitro: {
    prerender: {
      concurrency: 4,
      routes: Array.from({ length: 50 }).map((_, i) => `/posts/${i + 1}`),
    }
  }
})
```

... and this `app.vue`:

```vue
<script setup lang="ts">
const { data } = await useAsyncData(async () => {
  console.log('running async data')
  await new Promise((resolve) => setTimeout(resolve, 1000))
  return { message: 'Hello World' }
})
</script>

<template>
  <pre>{{ data }}</pre>
</template>
```

### 📝 Checklist


docs/2.guide/3.going-further/1.experimental-features.md
```diff
@@ -321,7 +321,41 @@ You can also set this to `chokidar` to watch all files in your source directory.
...
   experimental: { 
-      watcher: 'chokidar-granular' // 'chokidar' or 'parcel' are also options
+    watcher: 'chokidar-granular' // 'chokidar' or 'parcel' are also options
   } 
 }) 
 ```
+
+## sharedPrerenderData
+
+Enabling this feature automatically shares payload _data_ between pages that are prerendered. This can result
```
danielroe, line 331:
```suggestion
Enabling this feature automatically shares payload *data* between pages that are prerendered. This can result
```

Atinux:
This is the kind of feature where I am like "Why didn't we think about this before?!"

I believe this is very good since during pre-rendering, the data are the same (no auth involved).

One thing for DX though, could we improve `useAsyncData` to warn if we have `import.meta.prerender` and that the key is auto-generated?

danielroe:
looks like we have some performance issues with enormous payloads. I have an idea to refactor to get payload data _on-demand_ which should resolve both concurrency issues and performance with _enormous_ paylods.

manniL:
> looks like we have some performance issues with enormous payloads

Related: https://github.com/nuxt/nuxt/issues/23594

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -280,6 +280,10 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
     islandContext
   }
 
+  if (import.meta.prerender && process.env.NUXT_SHARED_DATA) {
+    ssrContext._sharedStorage = useStorage('internal:nuxt:prerender:shared')
```
pi0, line 284: Herer, we could expose a wrapped cached function for the handler, which can handle concurrency as well. Supporting concurrency over cache I think could make this feature significantly more useful!

danielroe: Nice idea!

My only worry was memory consumption but I guess we could use an lru cache? Wdyt?

pi0: PS: Using `.cache/nitro` as key, we could benefit from single configuration to define where build cache goes.

pi0: > My only worry was memory consumption but I guess we could use an lru cache? Wdyt? 

Surely! We could also mount an LRU for a specific subpath and configure it for nuxt pretender specifically. I am thinking also for main `.cache` we could do the same later.

danielroe: Also, would we ever prerender in a different context, meaning we needed storage to share data?

pi0: At some point yes, but it will also come with an internal API that we can integrate with the storage layer so we won't "loose" this feature since we rely on storage API.

danielroe: implemented wrapped function handling with storage layer underneath. ✅ 

the more I think about it, the more I think we should have a hybrid memory/FS storage for all of these caches - I can implement in separate PR. The idea would be that we have an LRU cache backed up by a FS cache so we _never_ have to rerun a fetcher function. The reason we need FS backing is that the other caches (payload particularly) _must_ retain the data - there is no fallback for them. And on sufficiently large sites the payloads could end up taking gigabytes of memory.

pi0: Using FS as the main cache + hybrid memory/LRU (or TTLed) makes sense for this usage. 

It could be nice also to see where we can have the cache so as not to reinvent the same thing in multiple places. Nitro caching API also handles concurrent promises with a storage-backed cache and TTL support (it will be a standalone lib) unstorage/nitro interaction should support multi-layer caching we could collaborate on these efforts.

pi0 (approved):
LGTM lets iterate 🚀

prakashpie:
Wants to learn more about generating dynamic routes with Nitro prerender hooks, with or without the use of payloads in Nuxt 3 ? 🤔

👉 Check out this repository where I've added examples for both scenarios.

[https://github.com/codepie-io/nuxt3-dynamic-routes](https://github.com/codepie-io/nuxt3-dynamic-routes)

PR #24931, opened 2023-12-27
fix(nuxt): stop loading indicator if page keys are the same

### LINKED ISSUE #24928: Loading indicator is triggered on anchor navigation
### Environment

- Node Version:     v20.10.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.12.1
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://nuxt.com/docs/api/utils/navigate-to#type

- click on a link of the table of content
- see the loading indicator

### Describe the bug

The loading indicator is triggered without being stopped when clicking on an anchor

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #25039: NuxtLoadingIndicator doesn't ends when going back with nested route
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

also

- Node Version:     v18.18.2
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      modules, pinia, components, imports, css, postcss, runtimeConfig, googleFonts, vite, typescript, ssr, veeValidate
- Runtime Modules:  @vueuse/nuxt@10.7.0, @nuxtjs/google-fonts@3.1.3, @pinia/nuxt@0.4.11, @vee-validate/nuxt@4.12.3
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-r5biil?file=pages%2Fuser.vue

### Describe the bug

1. Click 'Go To UserPage'. LoadingIndicator works well.
2. Click 'Go To User 1111'. It goes nested route /user/[id].vue. and works well.
3. But if you go to previous page (press back button), LoadingIndicator loads slowly and it doesn't ends.

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix #24928 
fix #25039 

### ❓ Type of change

### 📚 Description
Hi :wave: this PR stops the loading indicator by triggering `page:loading:end` if page keys are the same when re-rendering

### 📝 Checklist


packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -88,6 +89,10 @@ export default defineComponent({
           }
 
           const key = generateRouteKey(routeProps, props.pageKey)
+          if (!nuxtApp.isHydrating && previousPageKey === key) {
```
danielroe, line 92: should we use `isChangingPage`?

huang-julien: I don't know 🤔 , `isChangingPage` is relying on `to` and `from` from a router guard. 

This implementation is relying on whether we force rerender the page or not by comparing the generated key.

Do you think we should use a router guard instead ?

BilalShah3: whats the solution

manniL: Do we want to have a different behavior compared to `isChangingPage` for checking? If not, let's use it.

huang-julien: I think we want to stop the loading indicator only if the setup of the new page won't be run. 

Using a router guard with beforeResolve don't necessaraly means that the setup will be re-run 🤔.

danielroe:
Is there a way to add some tests for loading indicator behaviour and these loading hooks? (Ensuring they are called in the right order, not over-called, etc.)

Atinux:
We also have the issue on https://nuxt.com/modules when clicking on the categories:

https://volta.s3.fr-par.scw.cloud/Clean_Shot_2024_01_02_at_13_24_58_71ddda020b.mp4

huang-julien:
I'm trying to add some tests, but we'd need to at least wait for 300ms to get the failing case on the current main branch.

300ms seems to be way too long for the CI and i can't reduce it otherwise tests are becoming false positive.

just FYI, this PR also fix the issue with queries

manniL:
@huang-julien FYI - we should also check nested page behavior with the `back` button. Ideally this PR solves that too (see https://github.com/nuxt/nuxt/issues/25039)

danielroe:
@huang-julien Do you think we could reduce that by updating loading indicator props, e.g. `throttle` and `duration`?

huang-julien:
> @huang-julien FYI - we should also check nested page behavior with the `back` button. Ideally this PR solves that too (see #25039)

Yeah... this does also stop the loading indicator in nested pages since the higher one will trigger the stop hook...

> @huang-julien Do you think we could reduce that by updating loading indicator props, e.g. `throttle` and `duration`?

Sadly it does not work 😢

thezumrad:
My temporary solution is this:
Just hiding it for certain page

`onMounted(() => {
  setTimeout(() => {
    const loadingIndicator = document.querySelector('.nuxt-loading-indicator');
    if (loadingIndicator) {
      loadingIndicator.classList.add('!opacity-0')
    }
  }, 1000);
})

onBeforeUnmount(() => {
  const loadingIndicator = document.querySelector('.nuxt-loading-indicator');
  if (loadingIndicator) {
    loadingIndicator.classList.remove('!opacity-0')
  }
})`

manniL:
@thezumrad it should already work in 3.9.1 ☺️

thezumrad:
Ahah yeah
I've 3.9.0. Did not pay attention 

Thanks)

PR #24935, opened 2023-12-28
fix(nuxt): don't show transition on initial page load

### LINKED ISSUE #24934: isChangingPage() from '#app/components/utils' is not checking for first load and wrongfully returns true on first hit / page refresh
### Environment

[01:36:27] Working directory: /home/projects/github-4luk6a
[01:36:27] Nuxt project info: 

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools, app, router
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-4luk6a?file=app%2Frouter.options.ts

### Describe the bug

The function should check if the `from` route is `START_LOCATION` from `vue-router` to detect first page load. First page load or page reload shouldn't 

### Additional context

I stumbled upon this whilst investigating why hash-links to different pages are inconsistent and why the router won't scroll down to hash-links on first hit or page reload. This isn't directly related to the bug (I will submit another PR for the router.options.ts separately because there need to be additional checks there for `START_LOCATION` in e.g. the `hasTransition()` method).

### Logs

_No response_

bernhardberger:
check if `from` route is `START_LOCATION` from vue-router in isChangingPage to detect first page load (like page refresh or first hit). When there is no previous route then there is no changing page. This causes pages to not scroll down to the hash on pages that are not on the / route.

### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/24934

### ❓ Type of change

### 📚 Description

Add check if `from` route is `START_LOCATION` vom `vue-router` in isChangingPage utility. Without this check it returns a false positive on page refresh or initial load.
Resolves #24934 

### 📝 Checklist

- [X] I have updated the documentation accordingly.


TheAlexLichter (approved):

packages/nuxt/src/app/components/utils.ts
```diff
@@ -4,6 +4,7 @@ import type { Component, RendererNode } from 'vue'
 import { isString, isPromise, isArray, isObject } from '@vue/shared'
 import destr from 'destr'
 import type { RouteLocationNormalized } from '#vue-router'
+import { START_LOCATION } from 'vue-router'
```
danielroe, line 7: we might need to add a stub for this to avoid including `vue-router` dependency in build when pages integration isn't enabled.

manniL: Good catch! Also via the `#vue-router` import?

danielroe: We currently use `#build/pages.mjs`.

https://github.com/nuxt/nuxt/blob/7c6b6771ab84686078624bcda089677a680c24cd/packages/nuxt/src/pages/module.ts#L91-L94

I'll push a quick fix.

danielroe (approved):

PR #24952, opened 2023-12-29
perf(nuxt): avoid double path lookup when generating imports

GalacticHypernova:
### ❓ Type of change

### 📚 Description

By using `else` we can avoid a double path lookup when checking if it's absolute, and we won't need to explicitly make another if statement as the return value is a boolean.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -153,7 +153,11 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
         return
       }
       let path = resolveAlias(from)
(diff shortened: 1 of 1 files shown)


DamianGlowala:
Good spot :) I'm also personally a fan of placing non-negated expressions in the `if ()` condition first and using `else` for the negated version of it (for improved readability).

GalacticHypernova:
> Good spot :) I'm also personally a fan of placing non-negated expressions in the `if ()` condition first and using `else` for the negated version of it (for improved readability).

Thanks! I agree with you, I do that too, I just didn't want to change the structure too much but I see there's no harm in reversing the order of operations here.

PR #24953, opened 2023-12-29
perf(nuxt): prefer iterating rather than `Object.fromEntries`

GalacticHypernova:
### ❓ Type of change

### 📚 Description

Many parts of nuxt require some sort of object manipulation, and therefore use Object.fromEntries and some more filter methods. This, especially in larger inputs, can really slow it down. This PR solves these issues by eliminating unnecessary iterations.

### 📝 Checklist


GalacticHypernova:
@danielroe Should I group all of these updates into this one PR? (Marking this a draft so that in case you say yes I would be able to add them all here).

GalacticHypernova:
Sorry this took so long, personal life has been very demanding! 
I would like to hear your opinion on whether this PR should be an all-around "perf PR" and kept around to group all improvements (kind of like a monorepo but in PR form) or separate them by topics (like this one currently is for Object.fromEntries)

GalacticHypernova:
Thanks for the refactor! 
I had planned to do it but got a bit caught up with something, sorry about that 😅 
Although I do wonder why with reduce 2 checks were failing...

GalacticHypernova:
May I ask why the revert? I get that smaller arrays would provide less of a performance improvement, but it's an improvement nonetheless. And I'm not even sure, could the bundles be considered small? Is it not user-input dependant?

danielroe (approved):

GalacticHypernova:
Well, in that case I suppose it's reasonable. Thanks for the feedback!

PR #24954, opened 2023-12-29
feat(nuxt): server-only pages

huang-julien:
### ❓ Type of change

### 📚 Description

Hi :wave: this PR allows users to define pages as server page with `definePageMeta()` and the `.server.vue` suffix in the filename.

~~in `definePageMeta` this PR took https://github.com/nuxt/nuxt/pull/25037 as reference for the field name : `mode`. For now it only accept undefined or `'server'`.~~ For the client only implementation, see https://github.com/nuxt/nuxt/pull/25037

### 📝 Checklist

- [x] test
- [x] refactor
- [ ] ???


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -13,6 +13,7 @@ import { join } from 'pathe'
 import type { NuxtIslandResponse } from '../../core/runtime/nitro/renderer'
 import { useNuxtApp, useRuntimeConfig } from '../nuxt'
 import { prerenderRoutes, useRequestEvent } from '../composables/ssr'
+import { useRoute } from '../../app/composables/router'
```
danielroe, line 16:
```suggestion
import { useRoute } from '../composables/router'
```

packages/nuxt/src/pages/utils.ts
```diff
@@ -87,6 +88,11 @@ export async function generateRoutesFromFiles (files: ScannedFile[], options: Ge
     // Array where routes should be added, useful when adding child routes
     let parent = routes
 
+    if(segments[segments.length - 1].endsWith('.server')) {
```
danielroe, line 91: One note: We might need to refactor this in future to support multiple ending segments. (But any change can be deferred to that point.)

```suggestion
    if (segments[segments.length - 1].endsWith('.server')) {
```

packages/schema/src/types/hooks.ts
```diff
@@ -28,6 +28,7 @@ export type NuxtPage = {
   alias?: string[] | string
   redirect?: RouteLocationRaw
   children?: NuxtPage[]
+  mode?: 'server'
```
danielroe, line 31: Could we support `all` as well, matching component mode syntax?

```suggestion
  mode?: 'server' | 'all'
```

packages/nuxt/src/pages/utils.ts
huang-julien: sure !

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -160,6 +162,7 @@ export default defineComponent({
       // TODO: Validate response
       // $fetch handles the app.baseURL in dev
       const r = await eventFetch(withQuery(((import.meta.dev && import.meta.client) || props.source) ? url : joinURL(config.app.baseURL ?? '', url), {
+        url: route.fullPath,
```
danielroe, line 165: I think this will make it much harder to cache server islands as the hash we create will vary by path.

Could we instead pass this as part of the context via `createIslandPage` instead?

huang-julien: I think we should not skip query 🤔, userss tend to relly on it within pages composables. Maybe can we pass an object like
```ts
{
  path: string,
query?: Record<string, string>
}
```
?

danielroe: The issue is (for example) that we will make a separate request for every query variation, which wouldn't be compatible with static sites and allow cache-busting by passing arbitrary query params or hashes ... (We should definitely not pass hash.) We could document this as a limitation of server pages... Or at a minimum, skip passing the full route if the route matches prerender routeRule?

huang-julien: I see, let's start by passing only the path in the context then 🤔

danielroe (commented):
this looks great to me.

I do think we probably could respect the route query as long as it's not a prerendered page. (We can use [`isPrerendered`](https://github.com/nuxt/nuxt/blob/25f1787922c726c6e166a4952ddb286cd013d624/packages/nuxt/src/app/composables/payload.ts#L77) to tell us.).

And I think we should drop the support for defining server-only pages within `definePageMeta`, which means that it's impossible to tree-shake this code out of the bundle, if even one page has any metadata at all.

danielroe (approved):

brc-dd:
There seems to be an issue with this if you try to directly access the server-only page in dev:

```
[nitro] [unhandledRejection] FetchError: [GET] "/_nuxt/builds/meta/dev.json": 404 Page not found: /_nuxt/builds/meta/dev.json
    at async $fetch2 (file:///Users/brc-dd/nuxt-app/node_modules/.pnpm/ofetch@1.3.3/node_modules/ofetch/dist/shared/ofetch.00501375.mjs:261:15)
[Vue warn]: Component foo is missing template or render function.
```

It works fine if you open home first and then go to server-only page via link 👀

PR #24993, opened 2024-01-02
docs: add an example of provide in object syntax plugins

Youhan:
### 🔗 Linked issue

There is no open issue for this.

### ❓ Type of change

### 📚 Description

The doc has an example of `provide`. The example can't be used when the plugin uses object syntax. This adds an example for that case.

### 📝 Checklist

### DIFF
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -169,8 +169,8 @@ export default defineNuxtPlugin(() => {
 ```
 ```ts [plugins/hello-object-syntax.ts]
 export default defineNuxtPlugin({
-  name: "hello",
-  setup(nuxtApp) {
+  name: 'hello',
+  setup () {
     return {
       provide: {
         hello: (msg: string) => `Hello ${msg}!`
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -155,7 +155,7 @@ Normally, Vue.js composables are bound to the current component instance while p
 
 ## Providing Helpers
 
-If you would like to provide a helper on the [`NuxtApp`](/docs/api/composables/use-nuxt-app) instance, return it from the plugin under a `provide` key. You can also use `nuxtApp.provide()` if you are using object syntax.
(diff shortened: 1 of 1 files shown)


docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -166,6 +167,15 @@ export default defineNuxtPlugin(() => {
   }
 })
 ```
+```ts [plugins/hello-object-syntax.ts]
+  name: "hello",
+  setup(nuxtApp) {
+    nuxtApp.provide('hello', (msg: string) => `Hello ${msg}!`)
```
danielroe, line 174: I wouldn't recommend this, but instead return an object, which allows type inference.

Is there a reason that isn't working for you?

Youhan: You are right! My issue was that in my reproduction/test I had `async setup` instead of `setup`. With async, typescript said that the setup should return `void`, which confused me. Should we mention that too in the docs you think?

Youhan: I fixed that with the latest commit.

danielroe: it should allow a return even if setup is async...

Youhan: I'll create a reproduction and if that's the case I'll open an issue :)

Youhan: Nope! not a bug! It my my mistake in my project. Somewhere in the plugin, I was doing `if(..) return`, which caused the typescript to go mad at me!

Youhan:
I still think having the extra example worth it to clarify that there is no difference. Defer to you to keep or close this PR.

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -166,6 +167,19 @@ export default defineNuxtPlugin(() => {
   }
 })
 ```
+```ts [plugins/hello-object-syntax.ts]
+  name: "hello",
+  setup(nuxtApp) {
```
danielroe, line 173:
```suggestion
  name: 'hello',
  setup () {
```

danielroe (approved):

PR #25007, opened 2024-01-02
fix(nuxt): clone cookie to detect changes within object

### LINKED ISSUE #24969: useCookie doesn't change value of an object stored in cookie
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-4fsr9x?file=app.vue

### Describe the bug

Same as https://github.com/nuxt/nuxt/issues/22001 
When I store an object in a cookie using useCookie composable and then want to change the value of a key of that object it doesn't effect the cookie in the web browser. The value in the ref is good but value registered in the cookie browser it not updated.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24969

### ❓ Type of change

### 📚 Description

When a cookie is an object, changing a property within it will change the original referenced object as well, meaning `isEqual` doesn't detect the change. Instead, we can clone the value straight away so we can detect differences.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -6,6 +6,7 @@ import { deleteCookie, getCookie, getRequestHeader, setCookie } from 'h3'
 import type { H3Event } from 'h3'
 import destr from 'destr'
 import { isEqual } from 'ohash'
+import { klona } from 'klona'
 import { useNuxtApp } from '../nuxt'
 import { useRequestEvent } from './ssr'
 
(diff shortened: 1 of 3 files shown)


TheAlexLichter (approved):
The whole stream approves 😛

Issue #25016, opened 2024-01-02
Adding node polyfills to nuxt


paro-paro:
### Environment

- Operating System: Darwin
- Node Version:     v20.10.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.12.1
- Builder:          -
- User Config:      ssr, devtools, vite
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Repro:

1. [vite-vue](https://stackblitz.com/edit/vitejs-vite-frjpo1)
2. [vite-nuxt](https://stackblitz.com/edit/github-y9hcb5) (SPA)

Both cases use vite 5.x (nuxt 3.9.0)

### Describe the bug

Hi!

I am trying to integrate the following authentication library [@web3auth/modal](https://www.npmjs.com/package/@web3auth/modal) into my nuxt app.

Despite being distributed as a "browser" library... it requires adding a few [polyfills](https://web3auth.io/docs/troubleshooting/vite-issues) in order to work, so I am using the [vite-plugin-node-polyfills](https://github.com/davidmyersdev/vite-plugin-node-polyfills) to handle them.

Unfortunately, I can only seem to make it work in a simple vite+vue application but not in nuxt (SPA mode).

Both cases share the same simple `App.vue` / `app.vue` file (see below) and in both cases the vite/nuxt dev server is run successfully.

The problem seems to be related with the build stage. Running `vite build` in the first case (vite+vue) will successfully build the application and `vite preview` will serve a fully functional production app. 

But in nuxt, despite setting `ssr` option to `false` and running `nuxt generate` (which successfully builds the app in SPA mode) (.output/public chunks are almost identical to the first case, as expected) serving the static app will throw an error when clicking in the sign in button =>  `exports are not defined`

I do understand that polyfills are messy, that browser-library authors should avoid them ([not this time, unfortunately](https://github.com/Web3Auth/web3auth-web/issues/1430#issuecomment-1489805525)) and that the vite plugin that I am using is not an official one (I have tried many other polyfills plugins/libraries and custom configurations with no luck so far...) but I was kind of hoping that maybe this "simple" use case might have a quick solution if I find the right nuxt configuration option or something like that.

Any help or guidance will be really appreciated. 

Thanks in advanced and happy new year!

### Additional context

`App.vue`- `app.vue` (same in both cases)

```vue
<script setup lang="ts">
const onSignIn = async () => {
[...]

PR #25028, opened 2024-01-03
feat: experimental Node.js imports compatibility for client-side

### LINKED ISSUE #12786: Support unenv for client to polyfill Node.js imports
[unjs/unenv](https://github.com/unjs/unenv) provides a powerful shimming framework for Node.js internals and other dependencies to be Node-free. We are already using it for server and nitro while we can use preset in webpack and vite to avoid issues when a library is depending on Node.js imports or wrongly importing from node built-ins

pi0:
### 🔗 Linked issue

resolves #12786, #25016

### ❓ Type of change

### 📚 Description

This PR allows using libraries for code that depends on Node.js API by polyfilling them using [unjs/unenv](https://github.com/unjs/unenv) which is already used for server bundles (via nitro and presets).

Notes: 
- All `node:` and node imports are supported
- No global injections are supported. `process`, `Buffer`, etc needs to be manually imported  

Injecting globals:

To avoid accidentally increasing bundle size if only a code-syntax depends on `Buffer` check (that adds `~42Kb`, and Vue deps already do this!), auto injections are disabled. But they can be easily imported from `node:`

```ts
import { Buffer } from 'node:buffer'
import process from 'node:process'
``` 

Inside a Nuxt plugin, they can be also polyfilled (by User's risk!)

```ts
// plugins/node.client.ts
import { Buffer } from 'node:buffer'
import process from 'node:process'

globalThis.Buffer = Buffer
globalThis.process = process
```

In next iterations we might find smarter ways to safely inject them.

### 📝 Checklist


test/bundle.test.ts
```diff
@@ -19,7 +19,7 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
   for (const outputDir of ['.output', '.output-inline']) {
     it('default client bundle size', async () => {
       const clientStats = await analyzeSizes('**/*.js', join(rootDir, outputDir, 'public'))
-      expect.soft(roundToKilobytes(clientStats.totalBytes)).toMatchInlineSnapshot(`"103k"`)
+      expect.soft(roundToKilobytes(clientStats.totalBytes)).toMatchInlineSnapshot(`"145k"`)
```
pi0, line 22: This is buffer overhead :( Probably testing with something smaller...

manniL: That's unfortunate! But then we can keep it behind a flag I suppose? Or would you consider it too unreliable/unsafe?

pi0: For now i have added an example how to polyfill it is rather simple for end-users and also makes it easier for Nuxt side to iterate and eventually do it by default.

Besides Vue to get rid of buffer, and us to find smarter injections (I'm not sure easily possible), I also opened https://github.com/unjs/unenv/issues/160 to reduce Buffer polyfill size more but still even 1Kb is unacceptable for users don't need Buffer.

packages/vite/src/client.ts
```diff
@@ -48,15 +51,17 @@ export async function buildClient (ctx: ViteBuildContext) {
...
     },
     optimizeDeps: {
       entries: [ctx.entry]
     },
     resolve: {
       alias: {
         '#build/plugins': resolve(ctx.nuxt.options.buildDir, 'plugins/client'),
-        '#internal/nitro': resolve(ctx.nuxt.options.buildDir, 'nitro.client.mjs')
+        '#internal/nitro': resolve(ctx.nuxt.options.buildDir, 'nitro.client.mjs'),
+        ..._env.alias
```
danielroe, line 64: The code here overrides user config so think we should add these subsequently via `defu` to allow the user to override these settings, granularly.

packages/schema/src/config/build.ts
```diff
@@ -54,6 +54,24 @@ export default defineUntypedSchema({
...
+   *
+   * ```ts
+   * import { Buffer } from 'node:buffer'
+   * import process from 'node:process'
+   *
+   * globalThis.Buffer = globalThis.Buffer || Buffer
+   * globalThis.process = globalThis.process || process
+   * ```
+   */
+  unenv: {},
```
danielroe, line 73: should we type this?

danielroe (commented):
I love this. I am wondering if we should default it to `false`, however - so it can be an escape hatch for users who need it.

I am thinking it would still be better for users to be aware of the fact that they are using node dependencies in the browser, so they can potentially bail out or switch to a different library.

What do you think?

packages/schema/src/config/build.ts
pi0: Ah somehow removed in last comit will revert

packages/vite/src/client.ts
pi0: Good catch! For now we only have this option so added simple override to periororitze user conf (besides, it will be configurable also using `unenv: {}` option for alias

pi0:
I am also happy to keep it disabled by default feel free to commit before merge. But it kinda takes away the point of build-in integration because users can already easily do same unenv client support from `nuxt.config` and `alias`

Currently we do same mocks on server-side for non Nodejs targets it makes the behavior more consistant for library support when universally imported in SSR.

Meanwhile i agree about keeping users aware of that, i guess as a followup we can show a warn about dependencies that use Node.js API and guiding users about prefering an alternative lib (if they are lucky enough to have alt!) with dep graph this should be straight forward.

pi0:
(i am also not happy about top level `unenv` key as it doesn't indicate client config. any idea welcome!)

danielroe:
what if we use a single `unenv` key with `client`/`server` subkeys that configures unenv? (We pass the server config across to ntiro..., and the client config to vite/webpack...)

We could also consider making the key more generic, as the average user who wants to configure this might be unaware of unenv or what it is or does...

Finally, if there's more testing to be done, we could release this under `experimental`...

pi0:
Mind you, in the current state what we do is only enable access to `node:` (and without proto) imports which previously would only make an error for the vite client-side bundle and there is no injection magic. It is all up to the user.

While I think it should be something really safe to be enabled by default, I have moved it behind `experimental.clientNodeCompat` flag. Also, the flag is now only a boolean as the only effect (addition of aliases) is configurable by users using the same `alias` config to override some or more.

danielroe (approved):
beautiful. 👌

I would actually be fine enabling by default (with a warning if needed) given there's no bundle cost - but let's see usage and we can always enable in future.

PR #25037, opened 2024-01-03
feat(nuxt): client only page

### LINKED ISSUE #24866: Client-only page
### Describe the feature

Importing a library that is not compatible with SSR in a page may result in an error like `window is undefined`.

```vue
<!-- /pages/page1.vue -->
<script setup>
import L from 'L' // L is a library that does not support SSR
</script>

<template>
<div id='L'>
  <!-- ... -->
</div> 
</template>
```

We need to manually extract it to a separate component to the `components` folder and import it in the page, even though this component may only be used once.

```vue
<!-- /components/PageComponent.vue -->
<script setup>
import L from 'L' // L is a library that does not support SSR
</script>

<template>
<div id='L'>
  <!-- ... -->
</div> 
</template>
```
```vue
<!-- /pages/page1.vue -->
<script setup>
</script>

<template>
<ClientOnly>
  <PageComponent />
</ClientOnly> 
</template>
```

Maybe we can add a `clientOnly: boolean` field to `definePageMeta` macro to improve DX. Wrapping this page to `ClientOnly` for end users.

I know we already have `routeRules`. However, I still want to server render the layout when navigating to this page.

### Additional information

### Final checks

logotip4ik:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/24866

### ❓ Type of change

### 📚 Description

resolves https://github.com/nuxt/nuxt/issues/24866

### 📝 Checklist


docs/2.guide/2.directory-structure/1.pages.md
```diff
@@ -299,6 +299,10 @@ You may define a name for this page's route.
 
 You may define a path matcher, if you have a more complex pattern than can be expressed with the file name. See [the `vue-router` docs](https://router.vuejs.org/guide/essentials/route-matching-syntax.html#custom-regex-in-params) for more information.
 
+#### `clientOnly`
+
+Will wrap your page in [the Nuxt `<ClientOnly>` component](/docs/api/components/client-only) if you set `clientOnly: true` in your `definePageMeta`. This might be useful if you trying to import library that depends on `window` in page file.
```
manniL, line 304: I'd word the "library" part a bit more general and give examples then. Something like

```suggestion
Will wrap your page in [the Nuxt `<ClientOnly>` component](/docs/api/components/client-only) if you set `clientOnly: true` in your `definePageMeta`. This might be useful if you trying to import library that has no support for server-side rendering, e.g. needs `document` or `window`.
```

packages/nuxt/src/pages/runtime/composables.ts
```diff
@@ -38,6 +38,8 @@ export interface PageMeta {
   path?: string
   /** Set to `false` to avoid scrolling to top on page navigations */
   scrollToTop?: boolean | ((to: RouteLocationNormalizedLoaded, from: RouteLocationNormalizedLoaded) => boolean)
+  /** Make this page run on client only */
```
manniL, line 41:
```suggestion
  /** Render this page on the client-side only */
```

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -8,6 +8,7 @@ import { toArray } from './utils'
 import type { RouterViewSlotProps } from './utils'
 import { generateRouteKey, wrapInKeepAlive } from './utils'
 import { RouteProvider } from '#app/components/route-provider'
+import ClientOnly from '#app/components/client-only'
```
manniL, line 11: I wonder if we have any negative impact by now always importing `ClientOnly` 🤔

logotip4ik: I also had doubts about the size of the bundle, but the minimal size fixture seems to pass. Maybe we can have a build flag, that will enable/disable this feature, but then it won't be as seamless as is

danielroe: The minimal fixture doesn't use `<NuxtPage>` so it wouldn't be affected.

danielroe (changes requested):
Whoa - thank you! ❤️ 

I wonder if we can implement this at build-time to avoid even including these pages in the server bundle (at the moment, they will be).

So, when we scan pages here:

https://github.com/nuxt/nuxt/blob/e084ee7b3a5b9adcb5eccb4945b8948d509ab0d1/packages/nuxt/src/pages/utils.ts#L340

we could instead use a virtual file that is handled by a build plugin. That build plugin would generate code that wraps the pages with `createClientOnly`.

I'm suggesting this because I think it will pair better with the implementation of server pages, which can be done in the same way (cc: @huang-julien)

logotip4ik:
@danielroe Implemented something like what you said. But it has one issue, when wrapped in `Transition` it will log that: "component inside `Transition` is rendering non root component". That's why later i moved `_wrapIf` from user page component to transition.

wdyt about build file that reexports `ClientOnly` component or plain function and moving `_wrapIf` back to `Transition`

danielroe:
I think that would probably be a good solution 👌

huang-julien (approved):

danielroe (changes requested):
I think it would be better to support detection from the file path (e.g. `~/pages/index.client.vue`) and add a `mode` to the `NuxtPage` interface, which would also allow setting this at build-time via `pages:extend`.

(`mode` could then be set to either `all`, `client` or `server`, mirroring plugins, components, etc.)

logotip4ik:
Yeah, that would be also great. Plus, it will be more consistent with components `.(server|client).vue`.

Will implement this closer to weekend 😅

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -14,6 +14,8 @@ import { _wrapIf } from '#app/components/utils'
 import { LayoutMetaSymbol, PageRouteSymbol } from '#app/components/injections'
 // @ts-expect-error virtual file
 import { appKeepalive as defaultKeepaliveConfig, appPageTransition as defaultPageTransition } from '#build/nuxt.config.mjs'
+// @ts-expect-error virtual file
+import { ClientOnly } from '#build/pages-wrapper.mjs'
```
huang-julien, line 18: So we bundle `createClientOnly` even if the page is not client only in some cases ?

logotip4ik: It should bundle `ClientOnly` component only if at least one of the pages are client only

huang-julien: yeah... can't we instead wrap the page component competly with `createClientOnly` ?

Because we'd bundle `ClientOnly` even if we're going to a page that is not client only

logotip4ik: Daniel also suggested that 😅. It did work, but the main problem was that the Vue showed warnings with page transitions enabled. Client only component isn't working well under transitions :(

huang-julien: I see, maybe can we wrap with a div so vue can correctly apply transitions ?

logotip4ik: Tried doing that and now even just wrapping page component into `createClientOnly` seems not working 😅

huang-julien: I'm currently testing island pages by defining it within the `route.mjs` with a dynamic import and a `.then`. I'm pretty sure we can do the same thing for `createClientOnly`

packages/nuxt/src/components/runtime/client-component.ts
```diff
@@ -0,0 +1,19 @@
...
+/*@__NO_SIDE_EFFECTS__*/
+export const createClientPage = (loader: AsyncComponentLoader) => {
+  const page = defineAsyncComponent(loader);
+
+  return defineComponent({
+    inheritAttrs: false,
+    setup() {
+      return () => h('div', [
+        h(ClientOnly, undefined, {
+          default: () => h(page)
```
huang-julien, line 14: can we find a way to send down page props ?

logotip4ik: sure thing :+1:

danielroe (approved):

PR #25071, opened 2024-01-05
feat(nuxt): warn when using data-fetching composables outside setup/plugin

atinux:
### ❓ Type of change

### 📚 Description

Added JSDoc comments for a better DX:

| before | after |
| - | - |
| CleanShot 2024-01-05 at 16 06 22@2x | CleanShot 2024-01-05 at 16 01 26@2x |
| CleanShot 2024-01-05 at 16 06 26@2x | CleanShot 2024-01-05 at 16 01 32@2x |

Added a warning when `useFetch` or `useAsyncData` is used on client-side on a component already mounted (or instance unavailable):

CleanShot 2024-01-05 at 16 05 20@2x

**Notes:**

It looks good on my tests in the playground, but before merging it want to test it on bigger projects to make sure I did not forgot anything.

### 📝 Checklist


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -131,7 +143,8 @@ export function useFetch<
     immediate,
     getCachedData,
     deep,
-    watch: watch === false ? [] : [_fetchOptions, _request, ...(watch || [])]
+    watch: watch === false ? [] : [_fetchOptions, _request, ...(watch || [])],
+    _useFetch: true
```
pi0, line 147: One solution to hide it is to use a Symbol

Atinux: Tell me how 😄

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -259,6 +326,9 @@ export function useAsyncData<
   if (import.meta.client) {
     // Setup hook callbacks once per instance
     const instance = getCurrentInstance()
+    if (!instance || instance?.isMounted) {
```
danielroe, line 329:
```suggestion
    if (import.meta.dev && (!instance || instance?.isMounted)) {
```

packages/nuxt/src/app/composables/fetch.ts
danielroe: we could also avoid typing it entirely

DamianGlowala:
Spotted in a few places:
'a SSR-friendly' -> 'a**n** SSR-friendly'

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -42,15 +42,51 @@ export interface AsyncDataOptions<
   PickKeys extends KeysOf<DataT> = KeysOf<DataT>,
   DefaultT = null,
 > {
+  /**
+   * Whether to fetch on server side.
```
DamianGlowala, line 46:
```suggestion
   * Whether to fetch on the server side.
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -42,15 +42,51 @@ export interface AsyncDataOptions<
...
+   * When set to false, will prevent the request from firing immediately
+   * @default true
+   */
   immediate?: boolean
+  /**
+   * Return data in a deep ref object (it is true by default). It can be set to false to return data in a shallow ref object, which can improve performance if your data does not need to be deeply reactive.
+   */
   deep?: boolean
+  /**
+   * Avoid fetching same key more than once at a time
```
DamianGlowala, line 87:
```suggestion
   * Avoid fetching the same key more than once at a time
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -113,6 +168,13 @@ export function useAsyncData<
   handler: (ctx?: NuxtApp) => Promise<ResT>,
   options?: AsyncDataOptions<ResT, DataT, PickKeys, DefaultT>
 ): AsyncData<PickFrom<DataT, PickKeys> | DefaultT, (NuxtErrorDataT extends Error | NuxtError ? NuxtErrorDataT : NuxtError<NuxtErrorDataT>) | null>
+/**
+ * Provides access to data that resolves asynchronously in a SSR-friendly composable.
+ * See {@link https://nuxt.com/docs/api/composables/use-async-data}
+ * @param key a unique key to ensure that data fetching can be properly de-duplicated across requests.
```
DamianGlowala, line 174:
```suggestion
 * @param key A unique key to ensure that data fetching can be properly de-duplicated across requests.
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -102,6 +150,13 @@ export function useAsyncData<
...
```
DamianGlowala, line 156:
```suggestion
 * @param key A unique key to ensure that data fetching can be properly de-duplicated across requests.
```

DamianGlowala (changes requested):

DamianGlowala (changes requested):

danielroe (approved):

danielroe:
It shouldn't. Please raise an issue if you encounter an issue and I'll look into it straight away.

HADB:
I'm using useFetch in a custom composable and will call this composable on each page mounted. I will get this warning when switching pages. Is this as expected?

danielroe:
Indeed, it's warning you not to call the composable in the mounted hook.

HADB:
I'm doing some common response code handling in my custom composable and which will be called both in or out the mounted hook. What should I use to call the apis if `useFetch` is not allowed to use in the custom composable?

danielroe:
useFetch returns a refresh function that can be called at any time. If you don't want to run it in setup, you can pass `immediate: false` to defer running the function until a time of your choosing.

dalirnet:
> useFetch returns a refresh function that can be called at any time. If you don't want to run it in setup, you can pass `immediate: false` to defer running the function until a time of your choosing.

This warning is also displayed when the refresh function is used.

danielroe:
Are you sure about that? If so, would you raise an issue with a reproduction? 🙏

matt-clegg:
Is there anyway of knowing where in the codebase this warning is being thrown from? We have updated Nuxt and spotted several of these warnings but it's not immediately clear which files/components are throwing the warning logs.

AnthonyRuelle:
I have warnings everywhere with this version upgrade. I don't use useFetch in .vue files for the simple reason that I have a very large application and I split that via the repository pattern. What do you recommend for sharing API call code without duplicating everything .. ?

manniL:
@matt-clegg set a debugger at the warning or use console.trace there (locally) ☺️ 

@AnthonyRuelle This is better suited in a discussion including a reproduction :)

AnthonyRuelle:
> @matt-clegg set a debugger at the warning or use console.trace there (locally) ☺️ 
> 
> @AnthonyRuelle This is better suited in a discussion including a reproduction :)


Yes of course, I created a discussion with an example in my project

https://github.com/nuxt/nuxt/discussions/25625

Issue #21532, opened 2023-06-12
useFetch & useAsyncData reactive key


c-schwan:
hi @pi0, @danielroe 

is it possible to pass the "key" on calling refresh/execute or making the key computed in options?

In my case i got a computed header with Authorization and custom key

`const headers = computed(() => token.value ? { Authorization: `Bearer ${token.value}` } : undefined)`
`const key = hash([opts.baseURL, request, unref(opts.params || opts.query), unref(headers)])`

so the fetchOptions are reactive and are watched by default. If the token changed on client the initialized key isn't changed.

94726:
#22348 would probably also be solvable by this.

connerblanton:
+1 for this. Ran into an issue with this when using the new `getCachedData` as seen here - https://stackblitz.com/edit/github-wyzfak?file=pages%2Findex.vue

Using the `getCachedData` with query params can produce incorrect results when navigating back to a page

Amar-Gill:
+1 I have same issue.

### My Situation

I am reactively changing a `ref()` in my top level `<Nav>` component. That ref is used in my `useFetch()` call as a reactive query param:

```ts
params: {
  searchParam: someReactiveRef,
}
```

The resulting data from `useFetch` is used to render a list of links that correspond to the specific query param. The data is re-fetched as required and I can reactively render the correct data in the list. This is good. ✔️ 

### What's not working

Since the key is not being updated beyond the first run of `useFetch`, any use of `useNuxtData(key)` and `refreshNuxtData(key)` elsewhere in the app (typically inside a `<NuxtPage />` will not get / refresh the data, even though it is present inside the `<Nav />` component.

Specifically, the key is being derived from page route params.

### Reproduction

Here is a focused reproduction of what my app is doing: https://stackblitz.com/edit/nuxt-starter-buzcel?file=composables%2FuseResource.ts,app.vue,pages%2Fa.vue

Notes:
- the returned data is re-fetched and renders correctly inside `app.vue` as you select different options from the `<select>` element
- the keyed state from `useAcyncData` never changes beyond initial `useFetch` call (see screenshot of Nuxt devtools):

![Screenshot 2023-12-02 at 1 25 27 PM](https://github.com/nuxt/nuxt/assets/39676845/47f9e48f-59bf-45d6-afad-df147d979fdd)
- notes on the screenshot:
  - key is `resource-undefined` because on page load the selected value of `<select>` element is undefined
[...]

PR #25079, opened 2024-01-06
feat(nuxt): allow passing dynamic route params in `useFetch` via options

DamianGlowala:
### ❓ Type of change

### 📚 Description

Currently, to make a `useFetch` request with a path containing dynamic segments, we need to use string interpolation/concatenation to substitute these with real values. This PR allows for keeping the request path **unmodified**.

This might unlock the benefits of type-safety integrations between data fetching composables and corresponding server endpoints, where path no longer needs to change.

It exposes a `routeParams` property (key-value pairs), which provides autocompletion for the keys by extracting dynamic segments from the request's path. Dynamic segments of the unmodified path are substituted internally with the real values provided via this property.

An example:
```
await useFetch("/api/posts/:post-id/comments/:comment-id/**/**:slug", {
    routeParams: {
        ":post-id": 1,
        ":comment-id": 2,
        "**": "catch-all",
        "**:slug": "named-catch-all"
    }
})
```

The current syntax for `routeRules` requires the keys to be unique, therefore making it impossible to specify different values for two different segments named in the same way, e.g `/api/posts/:id/comments/:id` or `/api/posts/**/comments/**`. However, as Nuxt allows for flexibility in naming, these two examples could be taken care of by the user and updated to e.g. `/api/posts/:post-id/comments/:comment-id` or `/api/posts/**:posts/comments/**:comments`.

I would very much value the team's feedback on this, whether this is something worth proceeding with or not.

Things to consider (if feedback is positive):
- making it an experimental feature
- applying it to `$fetch` as well
- improving DX by providing a warning when dynamic segments are left with no `routeParams` provided
- deciding on whether to keep the `routeParams` keys exactly the same as in the path or come up with a different format and map accordingly

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -26,6 +26,12 @@ interface NitroFetchOptions<R extends NitroFetchRequest, M extends AvailableRout
 
 type ComputedFetchOptions<R extends NitroFetchRequest, M extends AvailableRouterMethod<R>> = ComputedOptions<NitroFetchOptions<R, M>>
 
+type ExtractRouteParams<T extends string> = T extends `${infer _Start}:${infer Param}/${infer Rest}`
+  ? Param | ExtractRouteParams<Rest>
+  : T extends `${infer _Start}:${infer Param}`
+  ? Param
+  : never;
+
 export interface UseFetchOptions<
   ResT,
   DataT = ResT,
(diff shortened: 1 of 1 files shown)


manniL:
Related: https://github.com/nuxt/nuxt/issues/21532

Amar-Gill:
I would use this in my projects.

If the `routeParams` option is introduced, should we consider deprecating the `params` option? It is currently an alias for the `query` option and might become confusing.

DamianGlowala:
Closing in favor of https://github.com/nuxt/nuxt/issues/21532.

Issue #24950, opened 2023-12-29
[Docs] Add minimal nuxt version to feature/utility - may be a label or a badge


mkierdev:
### Describe the feature

Nuxt 3.9 introduced new utility called callOnce. https://nuxt.com/docs/api/utils/call-once

However, if you are reading that page you have no idea that Nuxt 3.9 is required in order to use that utility. Maybe it would be possible to mark minimal nuxt version for a feature to work? It can be implemented as a badge or label. 

This information can be:
- "global" for whole docs page, providing ```minimalVersion``` inside frontmatter
- "local" for newly introduced feature extending functionality like getCachedData for useFetch (Nuxt 3.8 if I remember correctly?) https://nuxt.com/docs/api/composables/use-fetch#params

I believe that simple ```:badge[v3.9]``` from Docus is more than enough

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
A great idea - and @Atinux and I have already talked about this...

If we can't automate it, then it would be nice to annotate.

manniL:
I remember this was a topic a few times ago and I'm also all for it. Helpful for upgrading versions or when new (future) features are committed but not released yet 👍🏻

manniL:
@Atinux not sure if that fully closes the issue as documenting the "first version available" of most features would be nice.

mkierdev:
callOnce was just an example of how important it is to provide minimal version info across every feature. I understand that it would be impossible to keep versioning whole docs to each nuxt version for various reasons, so this is why a small notice would be helpful.
 
Small projects can often be up-to-date, but when we talk about large-scale apps with a lot of dependencies it is not so obvious that they can be updated on day 1.

Atinux:
Sure thing! Happy to re-open it as we can improve the docs along the way for each composable / feature with the minimal version associated to it 👍

warflash:
Linking https://github.com/nuxt/nuxt/issues/23189

luc122c:
[...]

PR #25086, opened 2024-01-06
docs(nuxt): add `@since` annotations to exported composables

luc122c:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/24950

### ❓ Type of change

### 📚 Description

This PR is a first step, adding `@since` annotations to exported composables. 

#### What is `@since`?

[As specified by JSDoc](https://jsdoc.app/tags-since)
> The `@since` tag indicates that a class, method, or other symbol was added in a specific version.

#### Methodology
- Look for exported functions
- Git blame to see when the function was created
- Look for an associated PR / Milestone / Release 
- If the code came across from Nuxt 2 or a Release Candidate, use 3.0.0

### 📝 Checklist


DamianGlowala:
I am wondering whether we should inform about `3.0.0`, which is likely to be assumed as the default.

luc122c:
@DamianGlowala I thought about this myself. I opted to add them 1) so that they would show up in IntelliSense and 2) for completeness. I'm happy to be overruled by the Nuxt team though 👍

danielroe:
I'm happy to have 3.0.0 for the record (then, if we ever omit one, it means for sure that we need to add a `@since` rather than that it's definitely present since 3.0.0)... (More lint-friendly.)

luc122c:
AI just blew my mind. I asked Phind:
> I want to find when this function was added to the codebase

It replied:

> To find out when a specific function was added to the codebase, you can use the `git blame` command followed by the filename. This command shows what revision and author last modified each line of a file.
> However, since you're looking for the addition of a specific function, you might want to use the `git log` command with the `-S` option, which allows you to search for changes in the amount of occurrences of a string in the source code.
> In your case, you can run the following command in your terminal:
> 
> ```bash
> git log -S'preloadRouteComponents' -- ./packages/nuxt/src/app/composables/preload.ts
> ```
> 
> This command will show the commit history where the string 'preloadRouteComponents' appears in the specified file. The first appearance of this string would likely be the addition of the function.

And in the terminal:
```
% git log -S'preloadRouteComponents' -- ./packages/nuxt/src/app/composables/preload.ts
commit 2a4ebfb18bdffdf9f61e613a2706bc83c4b277c8
Author: Daniel Roe <daniel@roe.dev>
Date:   Mon Oct 17 12:15:29 2022 +0100

    perf(nuxt): improve link prefetching (#8225)
 ```
 Which led me to nuxt/framework#8225 which was included in `3.0.0-rc.12`

File: packages/nuxt/src/app/composables/asyncData.ts [hunk 1]
```diff
@@ -499,4 +504,4 @@ function pick (obj: Record<string, any>, keys: string[]) {
     (newObj as any)[key] = obj[key]
   }
   return newObj
-}
+}
```
luc122c, line 507: I'm not sure what's happened here?

luc122c:
All composables exported from `packages/nuxt/src/app/composables/index.ts` now have a `@since` annotation.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -26,19 +28,22 @@ export function useRequestHeaders (include?: any[]) {
...
 
+/** @since 3.2.0 */
 export function useRequestFetch (): typeof global.$fetch {
   if (import.meta.client) {
     return globalThis.$fetch
   }
   return useRequestEvent()?.$fetch as typeof globalThis.$fetch || globalThis.$fetch
 }
 
+/** @since 3.0.0 */
```
luc122c, line 46: When a function has multiple overloads, does each one need it's own annotation? I see that the last overload has a `@deprecated` annotation which does not affect the first.
Also, what about the case where a new overload is added in a newer version, can that have it's own `@since` annotation?

packages/nuxt/src/app/composables/asyncData.ts
danielroe:
```suggestion
}

```

packages/nuxt/src/app/composables/ssr.ts
danielroe: I'm happy to stick with just one for now.

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -1,30 +1,53 @@
-import type { FetchError, FetchOptions } from 'ofetch'
```
danielroe, line 1: Would you be able to revert the code style changes in this file?

(And also, we need to investigate why CI lint step isn't catching this 🤔)

luc122c: Argh, sorry about this. I thought I'd turned off the formatter 🤦 Yes, i'll sort it 👍

danielroe (approved):
Thank you! ❤️

PR #25109, opened 2024-01-08
chore(kit): improve types for addTemplate

ineshbose:
### ❓ Type of change

### 📚 Description

Providing better type definition with the `options` property along with the right definition for `getContents` parameter, and cleaned up redundant code where we were typing the context.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -366,7 +366,7 @@ export default defineNuxtModule({
     addTemplate({
       filename: 'routes.mjs',
       getContents ({ app }) {
-        if (!app.pages) return ''
+        if (!app.pages) return 'export default []'
         const { routes, imports } = normalizeRoutes(app.pages)
         return [...imports, `export default ${routes}`].join('\n')
       }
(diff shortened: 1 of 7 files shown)


packages/nuxt/src/pages/module.ts
```diff
@@ -366,6 +366,7 @@ export default defineNuxtModule({
     addTemplate({
       filename: 'routes.mjs',
       getContents ({ app }) {
+        if (!app.pages) return ''
```
danielroe, line 369:
```suggestion
        if (!app.pages) return 'export default []'
```

danielroe:
This [failure](https://github.com/nuxt/ecosystem-ci/actions/runs/7472440918/job/20334638207) looks like an issue in nuxt/content - nicely discovered through this PR. 🙌

danielroe (approved):

danielroe:
Oh - one comment. IIRC, we changed the type signature in Nuxt 3 vs Nuxt 2. We should confirm that this type is accurate for Nuxt 2 projects as well 🤔

PR #25174, opened 2024-01-12
chore: use nuxt.new URLs in issue template

TheAlexLichter:
### ❓ Type of change

### 📚 Description

The URLs in the bug report issue template were still the "old template URLs". A we use `v3` now everywhere (e.g. on `nuxt.new`), I've updated them.

### 📝 Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -9,8 +9,8 @@ body:
         👉 https://nuxt.com/docs/community/reporting-bugs
 
         Please use a template below to create a minimal reproduction
-        👉 https://stackblitz.com/github/nuxt/starter/tree/v3/
-        👉 https://codesandbox.io/s/github/nuxt/starter/tree/v3/
+        👉 https://stackblitz.com/github/nuxt/starter/tree/v3
+        👉 https://codesandbox.io/s/github/nuxt/starter/tree/v3
   - type: textarea
     id: bug-env
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -9,8 +9,8 @@ body:
         👉 https://nuxt.com/docs/community/reporting-bugs
 
         Please use a template below to create a minimal reproduction
-        👉 https://stackblitz.com/github/nuxt/starter/tree/v3-stackblitz
-        👉 https://codesandbox.io/s/github/nuxt/starter/v3-codesandbox
+        👉 https://stackblitz.com/github/nuxt/starter/tree/v3/
+        👉 https://codesandbox.io/s/github/nuxt/starter/tree/v3/
```
DamianGlowala, line 13:
```suggestion
        👉 https://stackblitz.com/github/nuxt/starter/tree/v3
        👉 https://codesandbox.io/s/github/nuxt/starter/tree/v3
```

danielroe, line 13: if we do this we can likely remove these branches from the starter. I think they exist in case there's anything specific, like https://github.com/nuxt/starter/blob/v3-codesandbox/sandbox.config.json...

manniL: But then we should also use them for `nuxt.new` (and keep them all up to date)

danielroe: Exactly. I think if both are working well we can just use the v3 branch.

danielroe (approved):

PR #25190, opened 2024-01-13
refactor(nuxt): refactor `NuxtIslandResponse` and improve nuxt-island render mechanism

huang-julien:
### ❓ Type of change

### 📚 Description
Hi :wave: this PR brings some improvement to nuxt islands and added more comment to make it easier to understand for newcomers if they are working on islands.

refactor for slots:
- use a component just like how client component are working
- move to attributes like `data-island-**`
- get it's own interface in the `NuxtIslandResponse`
- the serverside teleport component for slots can be used in non island components.
- move props from html to the json response

refactor of `NuxtIslandResponse`.
- move slots and client component's data into their own interfaces instead of having props/components at the root of the object.
  - former props and components are moved to the interface for client components

the server-side teleport component for client components has been improved to be usable in non island components. So we can propagate the `nuxt-client` attribute globally (in another experimental flag and in another PR)

improvement of the render mechanism:
- the new system relies on `withMemo` to have a better control on the static and teleports rendering

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -58,10 +58,18 @@ export interface NuxtIslandContext {
   name: string
   props?: Record<string, any>
   url?: string
-  // chunks to load components
-  chunks: Record<string, string>
-  // props to be sent back
-  propsData: Record<string, any>
+  slots: Record<string, Omit<NuxtIslandSlotResponse, 'html' | 'fallback'>>
+  clients: Record<string, Omit<NuxtIslandClientResponse, 'html'>>
```
danielroe, line 62: I'm not 100% sure about `clients` - can you think of another word?

danielroe:
Tagging @harlan-zw because of the `og-image` failure - though this looks like a snapshot mismatch rather than anything more serious.

danielroe (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
huang-julien: Do you have any suggestion ?

harlan-zw:
> Tagging @harlan-zw because of the `og-image` failure - though this looks like a snapshot mismatch rather than anything more serious.

The actual image buffer snapshots aren't modified so looks good for the og-image module. I can rework these tests if it's a blocker.

packages/nuxt/src/core/runtime/nitro/renderer.ts
harlan-zw: only having a minimal knowledge of how this is all working, clients hints at a list of socket connections or something

huang-julien: Maybe selectiveClient or simply `components`?

harlan-zw (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
harlan-zw: `components` sounds good to me then, it's clear enough that non-hydrated HTML won't have a component

huang-julien: in that case, should we also rename data-island-client to data-island-component ?

danielroe: oops, I missed this 🤦 

opened https://github.com/nuxt/nuxt/pull/25232

Issue #25196, opened 2024-01-14
`<NuxtLayout /> component has not been used.` Warning when using `<ClientOnly>` in `app.vue`


manniL:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-issues-25196

```vue
<template>
  <ClientOnly>
    <NuxtLayout>
      <NuxtPage />
    </NuxtLayout>
  </ClientOnly>
</template>
```

### Describe the bug

When using a `<ClientOnly>` component around the `App.vue` (only SSR the app shell), the NuxtLayout warning will appear even though the component is used.

### Additional context

Related: https://github.com/nuxt/nuxt/issues/24912

danielroe:
Yes, this is a limitation in how we are checking whether the layout is used.

zeezzh:
I get the warning when using middleware call navigateTo().

danielroe:
@zeezzh Would you provide a reproduction? 🙏

zeezzh:
@danielroe See [https://stackblitz.com/edit/github-bnh9zl](url).It happens when navigateTo called on server side.

danielroe:
We could bail on checking entirely if `<ClientOnly>` is used somewhere?

PR #25197, opened 2024-01-14
fix: check for layout after nextTick

TheAlexLichter:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25196

### ❓ Type of change

### 📝 Checklist


packages/nuxt/src/app/plugins/check-if-layout-used.ts
```diff
@@ -1,3 +1,4 @@
+import { nextTick } from "vue"
```
danielroe, line 1:
```suggestion
import { nextTick } from 'vue'
```

packages/nuxt/src/app/plugins/check-if-layout-used.ts
```diff
@@ -16,7 +17,10 @@ export default defineNuxtPlugin({
       }
     }
     if (import.meta.server) {
-      nuxtApp.hook('app:rendered', checkIfLayoutUsed)
+      nuxtApp.hook("app:rendered", async () => {
+        await nextTick()
+        checkIfLayoutUsed()
+      });
```
danielroe, line 23: seems like there might be something wrong with our eslint setup 🤔 

```suggestion
      nuxtApp.hook('app:rendered', async () => {
        await nextTick()
        checkIfLayoutUsed()
      })
```

packages/nuxt/src/app/plugins/check-if-layout-used.ts
```diff
@@ -16,7 +17,10 @@ export default defineNuxtPlugin({
       }
     }
     if (import.meta.server) {
-      nuxtApp.hook('app:rendered', checkIfLayoutUsed)
+      nuxtApp.hook('app:rendered', async () => {
+        await nextTick()
+        checkIfLayoutUsed()
+      })
```
danielroe, line 23: there might be some value in not awaiting the nextTick as that will block `app:rendered`...

```suggestion
      nuxtApp.hook('app:rendered', () => { nextTick(checkIfLayoutUsed) })
```

manniL: Makes sense 👍

danielroe (approved):

PR #25210, opened 2024-01-15
feat(nuxt): experimentally extract route metadata at build time

### LINKED ISSUE #24770: expose some route metadata at build-time
### Describe the feature

Currently `definePageMeta` is a purely runtime macro and can't be consumed by modules. But _some_ of that metadata, particularly the route metadata (`alias`, `name`, `path`), would be very useful for modules to have access to and modify at build time. For example, see https://github.com/nuxt-modules/i18n/pull/2603.

The downside here would be performance cost, and of course it would only work with static strings/arrays rather than variables or conditional assignment.

Might be worth exposing option for metadata extraction (e.g. useful for extracting component metadata), and investigate caching or sharing parsing between utilities/plugins?

It would resolve https://github.com/nuxt/nuxt/issues/24359.

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24770

### ❓ Type of change

### 📚 Description

This lets us extract `name`, `alias`, `redirect` and `path` properties from with `definePageMeta` at _build-time_, meaning modules can read/write these values in the `pages:extend` hook. This allows tighter (and better) build-time integrations with modules that work with routing, for example with `@nuxtjs/i18n`.

I've added this under an experimental flag for testing for now.

```ts
  experimental: {
    scanPageMeta: true,
  },
  hooks: {
    'pages:extend' (routes) {
      console.log(routes)
    }
  }
})
```

### 📝 Checklist


BobbieGoede:
Nice! On initial testing it looks like this make it possible to fix https://github.com/nuxt-modules/i18n/issues/2581. 💪

packages/nuxt/src/pages/utils.ts
```diff
@@ -128,25 +134,74 @@ export function extractScriptContent (html: string) {
 
 const PAGE_META_RE = /(definePageMeta\([\s\S]*?\))/
 
-async function getRouteName (file: string) {
-  const script = extractScriptContent(file)
-  if (!script) { return null }
+const metaCache: Record<string, Partial<Record<keyof NuxtPage, any>>> = {}
+async function getRouteMeta (contents: string, absolutePath?: string): Promise<Partial<Record<keyof NuxtPage, any>>> {
```
BobbieGoede, line 138: I think `absolutePath` is always passed to `getRouteName`, would using it as key for `metaCache` be more performant?

BobbieGoede (approved):
As far as I have tested it works well, apart from the comment it all looks good to me 🚀

packages/nuxt/src/pages/utils.ts
danielroe: I think the metadata might vary based on the contents of the file. (So unless we can invalidate based on a fs watcher, we probably can't key to the file name...) But we could consider another approach, like keying to a hash of the file contents.

BobbieGoede: >I think the metadata might vary based on the contents of the file. 

Ah, of course 😅..

>But we could consider another approach, like keying to a hash of the file contents.

I'm not sure how performance compares between using large key or hashing large strings, my guess is that the current approach would do better (unless the file hash is already available someplace else). 

Maybe something in between could be to combine `absolutePath` and `script` (hashed or not, a placeholder if no script is present) to use as a key? But then `extractScriptContent` would always need to be called 😅

BobbieGoede:
While this feature is already an improvement for modules relying on route meta during build in its current state, is there a reason for limiting the extracted keys to `['name', 'path', 'alias', 'redirect']` other than performance and/or predictability?

In the i18n module we implement a macro called [`defineI18nRoute`](https://i18n.nuxtjs.org/api/compiler-macros) which I just realized works almost the same as the changes in this PR. I think it would be ideal if it were possible to instead use `definePageMeta({ i18n: { ... } })`, to provide static i18n route options. 

Maybe this is very specific to the i18n module, I am not aware of other modules implementing their own compiler macros to work around this issue.

BobbieGoede:
I think there are no tests for routes without `file`, something we should be wary of as the changes/refactors I made have touched that logic.

packages/nuxt/src/pages/utils.ts
```diff
@@ -301,47 +377,99 @@ function prepareRoutes (routes: NuxtPage[], parent?: NuxtPage, names = new Set<s
...
+          if (route[key] == null) {
+            delete metaRoute[key]
+            continue
+          }
+
+          metaRoute[key] = route[key]
+        }
+      } else {
+        if (route.meta != null) {
+          metaRoute.meta = `{ ...(${metaImportName}) || {}), ...${route.meta} }`
```
eriksLapins, line 460: typo on line 460 (extra ")" after "...(${metaImportName}) || {}"

danielroe: Nice spot. It's already fixed in nightly releases, if you are encountering a bug here.

PR #25219, opened 2024-01-15
fix(nuxt): render user-inserted links in island responses

danielroe:
### ❓ Type of change

### 📚 Description

Previously, the only `<link>`s that were inserted in island responses were vite dev stylesheet. This PR removes the filtering to allow inserting links into island responses (for example, in case of `<NuxtImg>` which can preload/prefetch images that are used).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -229,9 +229,8 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
   }
 
   // Check for island component rendering
-  const islandContext = (process.env.NUXT_COMPONENT_ISLANDS && event.path.startsWith('/__nuxt_island'))
-    ? await getIslandContext(event)
-    : undefined
+  const isRenderingIsland = (process.env.NUXT_COMPONENT_ISLANDS as unknown as boolean && event.path.startsWith('/__nuxt_island'))
+  const islandContext = isRenderingIsland ? await getIslandContext(event) : undefined
 
   if (import.meta.prerender && islandContext && event.path && await islandCache!.hasItem(event.path)) {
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -430,9 +430,13 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
       style: []
     }
     for (const tag of await head.resolveTags()) {
-      if (tag.tag === 'link' && tag.props.rel === 'stylesheet' && tag.props.href.includes('scoped') && !tag.props.href.includes('pages/')) {
+      if (import.meta.dev && tag.tag === 'link' && tag.props.rel === 'stylesheet' && tag.props.href.includes('scoped') && !tag.props.href.includes('pages/')) {
```
huang-julien, line 433: What would happen in production if an island use `useHead()` with a stylesheet <link> ? (can't test as i'm at work) Is it going to be ignored ?

danielroe: Yes, it would be ignored - but that wouldn't be inserted even now - it only inserts hrefs with `scoped` in their name which I believe is dev-only 🤔

I would be totally up for refactoring this so islands can insert more kinds of things in `<head>`, now that I've disabled the default resource hints from rendering there...

huang-julien: oh i see that make sense :+1:

I think we could allow island to insert things into `head`. If this is too late for 3.9.2, this can probably be done in another PR ?

huang-julien (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
danielroe: good point - updated ✅

huang-julien (approved):

PR #25220, opened 2024-01-15
fix: reuse serialized values in `normalizedRoutes`

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/nuxt/pull/25210

### ❓ Type of change

### 📚 Description
Built on https://github.com/nuxt/nuxt/pull/25210, I haven't looked into other changes for the PR yet.

This is refactor of `normalizedRoutes` mostly for legibility with some tiny/negligible changes for perf. 

* Changes the initial loop to skip serializing `route.children` as its being overwritten in the next few lines
  * ~Should this be changed so that empty strings also get stringified? I think they are skipped as they're falsy~
  * Based on the PR history, yes they should be stringified to allow empty paths (and names?)
* Removes the second serialization loop, serialization has already occurred in the previous loop
* Rewrite of `route` assignments to reuse serialized values

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -362,9 +362,9 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> = 
     routes: genArrayFromRaw(routes.map((page) => {
       const route: Record<Exclude<keyof NuxtPage, 'file'>, string> & { component?: string } = Object.create(null)
       for (const [key, value] of Object.entries(page)) {
-        if (key !== 'file' && (Array.isArray(value) ? value.length : value)) {
-          route[key as Exclude<keyof NuxtPage, 'file'>] = JSON.stringify(value)
-        }
+        if (['file', 'children'].includes(key)) continue
+        if (!(Array.isArray(value) ? value.length : value)) continue
+        route[key as Exclude<keyof NuxtPage, 'file'>] = JSON.stringify(value)
       }
 
       if (page.children?.length) {
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -388,7 +388,7 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> = 
       metaImports.add(genImport(`${file}?macro=true`, [{ name: 'default', as: metaImportName }]))
 
       if (overrideMeta) {
-        route.name = route.name ?? `${metaImportName}?.name ?? undefined`
+        route.name = route.name ?? `${metaImportName}?.name`
         route.path = route.path ?? `${metaImportName}?.path ?? ''`
       } else {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/utils.ts
```diff
@@ -361,10 +361,16 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> =
     imports: metaImports,
     routes: genArrayFromRaw(routes.map((page) => {
       const route: Record<Exclude<keyof NuxtPage, 'file'>, string> & { component?: string } = Object.create(null)
+      const skipSerializeValue = (val: unknown | unknown[]) => {
+        if (val === undefined) return true
+        if (Array.isArray(val)) return val.length === 0
+        return typeof val !== 'string' && !!val
+      }
```
danielroe, line 368: Looks like this utility can live outside the arrow function here - maybe at top level.

packages/nuxt/src/pages/utils.ts
```diff
@@ -361,10 +361,16 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> =
...
+
       for (const [key, value] of Object.entries(page)) {
-        if (key !== 'file' && (Array.isArray(value) ? value.length : value)) {
-          route[key as Exclude<keyof NuxtPage, 'file'>] = JSON.stringify(value)
-        }
+        if (['file', 'children'].includes(key)) continue
+        if (skipSerializeValue(value)) continue
+        route[key as Exclude<keyof NuxtPage, 'file'>] = JSON.stringify(value)
```
danielroe, line 373:
```suggestion
        route[key as Exclude<keyof NuxtPage, 'file' | 'children'>] = JSON.stringify(value)
```

packages/nuxt/src/pages/utils.ts
```diff
@@ -373,32 +379,31 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> =
...
 
-      let aliasCode = `${metaImportName}?.alias || []`
-      const alias = toArray(page.alias).filter(Boolean)
-      if (alias.length) {
-        aliasCode = `${JSON.stringify(alias)}.concat(${aliasCode})`
-      }
       if (overrideMeta) {
-        route.name = page.name !== undefined ? JSON.stringify(page.name) : `${metaImportName}?.name`
-        route.path = page.path !== undefined ? JSON.stringify(page.path) : `${metaImportName}?.path`
+        route.name = route.name ?? `${metaImportName}?.name ?? undefined`
```
danielroe, line 390:
```suggestion
        route.name = route.name ?? `${metaImportName}?.name`
```

danielroe (approved):

packages/nuxt/src/pages/utils.ts
danielroe: Those would be nice changes. 👌 

I think we can optimise quite a bit of this.

PR #25222, opened 2024-01-15
fix(nuxt): await async payload revivers

cernymatej:
### ❓ Type of change

### 📚 Description

This PR resolves an issue where the payload reviver was not awaited if it was an asynchronous function. This could be useful for creating more versatile payload revivers (utilizing dynamic imports), for example. However, prior to this change, the revived value on the client would be a promise instead of its resolved value.

It would probably also be a good idea to add tests for this issue. However, I noticed that `definePayloadReducer` and `definePayloadReviver` are listed in `skippedComposables` in the tests. Is there a reason why they are not tested?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/payload.ts
+++ b/packages/nuxt/src/app/composables/payload.ts
@@ -102,7 +102,7 @@ export async function getNuxtClientPayload () {
     return {}
   }
 
-  const inlineData = parsePayload(el.textContent || '')
+  const inlineData = await parsePayload(el.textContent || '')
 
   const externalData = el.dataset.src ? await _importPayload(el.dataset.src) : undefined
 
@@ -115,8 +115,8 @@ export async function getNuxtClientPayload () {
   return payloadCache
 }
 
-export function parsePayload (payload: string) {
-  return parse(payload, useNuxtApp()._payloadRevivers)
+export async function parsePayload (payload: string) {
+  return await parse(payload, useNuxtApp()._payloadRevivers)
 }
 
 /**


danielroe (approved):

manniL:
@CernyMatej There are tests in place though, see https://github.com/nuxt/nuxt/blob/629461b7f768c17fa2b5815723a9d17e305b0b99/test/fixtures/basic/plugins/custom-type-registration.ts#L3-L4 and https://github.com/nuxt/nuxt/blob/629461b7f768c17fa2b5815723a9d17e305b0b99/test/fixtures/basic/plugins/custom-type-assertion.client.ts#L1-L2

TheAlexLichter (approved):

CernyMatej:
@manniL thanks for pointing that out! Didn't notice them. However, those are only fixtures. Where are they used in tests? I'd be happy to write a test case for it if you could tell me where to put it.

manniL:
@CernyMatej That's it! the first link shows where it is set up and the 2nd link shows where it is checked in the fixture, throwing an error and failing tests when the reviving would fail ☺️
Maybe it'd be already enough to make that case async though 🤔

CernyMatej:
I noticed that I had made changes to the `devalue` source code to support asynchronous revivers before making the changes in Nuxt and forgot about them. 🤦‍♂️ 

At the moment, `parse()` from `devalue` isn't asynchronous and [doesn't actually await](https://github.com/Rich-Harris/devalue/blob/50af63e2b2c648f6e6ea29904a14faac25a581fc/src/parse.js#L60) the reviver which causes the revived value to stay as a fulfilled promise. I'm wondering what would be the best solution for this... 🤔

Changing `parse()` to be asynchronous would be a breaking change, wouldn't it? I'm thinking maybe include the parsing code in Nuxt and change it to be asynchronous - so that other users of devalue are not affected by this change?

manniL:
@CernyMatej I think it would unless introducing a parseAsync function 😋

PR #25245, opened 2024-01-17
fix(nuxt): check for uid before replacing data-island-uid attribute in island

huang-julien:
### 🔗 Linked issue
https://github.com/danielroe/roe.dev/pull/1350/commits/b33100c8b9300dd5a01660fa34f1105b49d58cab

### ❓ Type of change

### 📚 Description

Hi :wave: this PR fix an issue where nested islands can be buggy and render invalid html (double attributes).

This is due to the fact that we were replacing all data-island-uid without checking if those were already set.

I'll add a test later.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -20,7 +20,7 @@ import { remoteComponentIslands, selectiveClient } from '#build/nuxt.config.mjs'
 
 const pKey = '_islandPromises'
 const SSR_UID_RE = /data-island-uid="([^"]*)"/
-const DATA_ISLAND_UID_RE = /data-island-uid(="[^"]*")?/g
+const DATA_ISLAND_UID_RE = /data-island-uid(="")?(?!="[^"])/g
 const SLOTNAME_RE = /data-island-slot="([^"]*)"/g
 const SLOT_FALLBACK_RE = / data-island-slot="([^"]*)"[^>]*>/g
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -20,7 +20,7 @@ import { remoteComponentIslands, selectiveClient } from '#build/nuxt.config.mjs'
 
 const pKey = '_islandPromises'
 const SSR_UID_RE = /data-island-uid="([^"]*)"/
-const DATA_ISLAND_UID_RE = /data-island-uid/g
+const DATA_ISLAND_UID_RE = /data-island-uid(="[^"]*")?/g
```
danielroe, line 23: What about something like:

```suggestion
const DATA_ISLAND_UID_RE = /data-island-uid(?!=)/g
```

Then we wouldn't need to pass a function matcher in...

huang-julien: :+1: Just to be sure. Is it possible that vue render something like `data-island-uid=""` instead of `data-island-uid` ?

danielroe: we could update to account for that:

```suggestion
const DATA_ISLAND_UID_RE = /data-island-uid(?!="[^"])/g
```

huang-julien: We still need the function matcher in that one 🤔. I guess we could go with your first suggestion. I havn't seen vue rendering `<div attribute="">` instead of simply `<div attribute>` yet

huang-julien: nevermind ...
image

danielroe: in that case, this would probably work:

```suggestion
const DATA_ISLAND_UID_RE = /data-island-uid(="")?(?!="[^"])/g
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -183,7 +183,14 @@ export default defineComponent({
...
         cHead.value.style = res.head.style
-        ssrHTML.value = res.html.replaceAll(DATA_ISLAND_UID_RE, `data-island-uid="${uid.value}"`)
+        ssrHTML.value = res.html.replaceAll(DATA_ISLAND_UID_RE, (full, match) => {
+          if (match) {
+            // avoid replacing uid that are already set. This can happen with nested islands
+            return full
+          }
+          return `data-island-uid="${uid.value}"`
+        })
+
```
huang-julien, line 193:
```suggestion
        ssrHTML.value = res.html.replaceAll(DATA_ISLAND_UID_RE, `data-island-uid="${uid.value}"`)
```

danielroe (approved):

PR #25255, opened 2024-01-17
fix(nuxt): inability to update same useCookie with same initial value

### LINKED ISSUE #25254: useCookie stops to set cookie with equal value after page reload
### Environment

- Operating System: Linux
- Node Version:     v20.10.0
- Nuxt Version:     3.9.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  yarn@4.0.2
- Builder:          -
- User Config:      alias, extends, runtimeConfig, modules, avModule, typescript
- Runtime Modules:  @misc/base-vue-module/module@5.0.0-beta.14, @nuxt/test-utils/module
- Build Modules:    -

### Reproduction

https://github.com/daniluk4000/nuxt-cookies-bug

### Describe the bug

https://github.com/nuxt/nuxt/assets/10450717/401e77bf-36bd-4e13-8818-f08c5d889dd0

### Additional context

channel.onmessage never works on the same page within same channel

### Logs

_No response_

daniluk4000:
### 🔗 Linked issue

#25254

### ❓ Type of change

### 📚 Description

Resolves #25254

Have also updated tests to cover this case.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -14,7 +14,7 @@ type _CookieOptions = Omit<CookieSerializeOptions & CookieParseOptions, 'decode'
 
 export interface CookieOptions<T = any> extends _CookieOptions {
   decode?(value: string): T
-  encode?(value: string): string
+  encode?(value: T): string
   default?: () => T | Ref<T>
   watch?: boolean | 'shallow'
   readonly?: boolean
@@ -26,7 +26,7 @@ const CookieDefaults = {
   path: '/',
   watch: true,
   decode: val => destr(decodeURIComponent(val)),
-  encode: val => encodeURIComponent(val)
+  encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val))
 } satisfies CookieOptions<any>
 
 export function useCookie<T = string | null | undefined> (name: string, _opts?: CookieOptions<T> & { readonly?: false }): CookieRef<T>
@@ -62,9 +62,8 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?: 
       if (opts.readonly || isEqual(cookie.value, cookies[name])) { return }
       writeClientCookie(name, cookie.value, opts as CookieSerializeOptions)
 
-      const rawCookie = typeof cookie.value === "string" ? cookie.value : JSON.stringify(cookie.value)
-      cookies[name] = destr(rawCookie)
-      channel?.postMessage(opts.encode(rawCookie))
+      cookies[name] = klona(cookie.value)
+      channel?.postMessage(opts.encode(cookie.value as T))
     }
 
     let watchPaused = false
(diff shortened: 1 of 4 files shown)


daniluk4000:
Reason behind this is that BroadcastChannel doesn't work on same page, so cookies[name] were never updated.

To solve this, we could also create second BroadcastChannel, but I believe that my solution is better on performance. 

Speaking of performance, we probably also want to avoid double JSON.stringify here.

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -61,7 +61,10 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
     const callback = () => {
       if (opts.readonly || isEqual(cookie.value, cookies[name])) { return }
       writeClientCookie(name, cookie.value, opts as CookieSerializeOptions)
-      channel?.postMessage(opts.encode(cookie.value as T))
+
+      const rawCookie = typeof cookie.value === "string" ? cookie.value : JSON.stringify(cookie.value)
+      cookies[name] = destr(rawCookie)
```
daniluk4000, line 66: We have to create new object here to avoid old object reference. Otherwise objects will be equal.

danielroe: Would klona be a better idea, as we already use it above?

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -14,7 +14,7 @@ type _CookieOptions = Omit<CookieSerializeOptions & CookieParseOptions, 'decode'
 
 export interface CookieOptions<T = any> extends _CookieOptions {
   decode?(value: string): T
-  encode?(value: T): string
+  encode?(value: string): string
```
danielroe, line 17: I think we want the ability for users to pass an encoder that handles non-strings...

```suggestion
  encode?(value: T): string
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -26,7 +26,7 @@ const CookieDefaults = {
   path: '/',
   watch: true,
   decode: val => destr(decodeURIComponent(val)),
-  encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val))
+  encode: val => encodeURIComponent(val)
```
danielroe, line 29: Let's avoid changing this behaviour and just handle with `klona` or etc. below

```suggestion
  encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val))
```

packages/nuxt/src/app/composables/cookie.ts
daniluk4000: Sure, but to avoid duplicate JSON.stringify encode will only receive string now, I've moved JSON.stringify from default encode to main function.

This ability is still preset for user, but encode will only receive strings from now on

```typescript
const rawCookie = typeof cookie.value === "string" ? cookie.value : JSON.stringify(cookie.value)
channel?.postMessage(opts.encode(rawCookie))
```

packages/nuxt/src/app/composables/cookie.ts
daniluk4000: Gotcha

danielroe (approved):

PR #25259, opened 2024-01-17
feat(nuxt): throw error if `setInterval` is used on server

### LINKED ISSUE #25256: DX: warn or throw error when using `setInterval` during SSR
### Describe the feature

It is easy to make a mistake and use `setInterval` in a Vue component without wrapping it into `onMounted()`.

This will let of memory leaks and `nuxt generate` being stuck.

I suggest to mock `setInterval` for the application part to throw an error and suggest using it inside `onBeforeMount` or `onMounted` Vue hooks.

Example of reproduction:

```vue
<script setup>
const index = ref(1);

setInterval(() => {
  if (index.value === 8) index.value = 1
  else index.value = index.value + 1;
}, 1000)
</script>
```

### Additional information

### Final checks

atinux:
### 🔗 Linked issue

Resolves #25256 

### ❓ Type of change

### 📚 Description

Throw an error if using `setInterval` in the Vue app during SSR to avoid memory leaks or getting the node process stuck during `nuxt generate`.

### 📝 Checklist


packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -56,3 +57,10 @@ export function prerenderRoutes (path: string | string[]) {
   const paths = toArray(path)
   appendHeader(useRequestEvent(), 'x-nitro-prerender', paths.map(p => encodeURIComponent(p)).join(', '))
 }
+
+export const setInterval = import.meta.client ? window.setInterval : () => {
+  throw createError({
+    statusCode: 500,
+    message: 'setInterval should not be used on server side, please wrap it with `onBeforeMount` or `onMounted` Vue Lifecycle hooks.'
```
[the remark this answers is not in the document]
Atinux: Happy to update it if you know how to do it 😄 

So far it looks like this:
CleanShot 2024-01-17 at 14 59 42@2x

Would be nice also to get the source map to link back here to `app.vue` instead of `app.js`, but same, not sure how to do this.

Can be handled in another PR to move one faster though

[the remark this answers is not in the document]
manniL: Yeah, let's go with the initial version and possibly raise an issue for improvements 👍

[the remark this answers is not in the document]
pi0: You might try `Error.captureStackTrace` if it can help (https://v8.dev/docs/stack-trace-api)

danielroe:
I love it! ❤️

packages/nuxt/src/app/composables/ssr.ts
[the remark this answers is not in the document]
Atinux: Hum does not help:

CleanShot 2024-01-17 at 15 27 59@2x

I think it's fine in the current step since the message is quite explicit.

danielroe, line 64:
```suggestion
    message: '`setInterval` should not be used on the server. Consider wrapping it with an `onNuxtReady`, `onBeforeMount` or `onMounted` lifecycle hook, or ensure you only call it in the browser by checking `import.meta.client`.'
```

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -56,3 +57,10 @@ export function prerenderRoutes (path: string | string[]) {
   const paths = toArray(path)
   appendHeader(useRequestEvent(), 'x-nitro-prerender', paths.map(p => encodeURIComponent(p)).join(', '))
 }
+
+export const setInterval = import.meta.client ? window.setInterval : () => {
+  throw createError({
```
danielroe, line 62: In production, perhaps don't throw an error, but just do nothing?

```suggestion
  if (!import.meta.dev) {
    return
  }
  throw createError({
```

pi0: I would suggest to show a warning in production as well. If anyhow setInterval slips away it can cause memory leaks that are more critical in production.

manniL: That way at least any kind of error tracking (or QA) could catch it 👍

Atinux: Well if we disable only in dev, we could still hang `nuxt generate`.

It might be tough, but they should never use it on ssr anyway

pi0: Good idea about generate. For prerender only (`import.meta.prerender`) we can directly return

gaetansenn:
> I love it! ❤️

Was my fault 😂

packages/nuxt/src/app/composables/ssr.ts
danielroe: what would be the use case of allowing when generating?

pi0: Only if suggestion in this comment (dev ~> return) i said (dev||prerender ~> return). Since it is an auto import only for Nuxt apps, i also agree with always throwing 👍🏼

danielroe: ah, my suggestion was to throw error _only_ in dev

danielroe (approved):

PR #25264, opened 2024-01-17
feat(nuxt): controllable view transition

### LINKED ISSUE #25263: Allow controlling the viewTransition on a per-page bases
### Describe the feature

Currently view-transitions are enabled globaly when the `experimental.viewTransition` property is set in the nuxt.config.ts.

It would be nice to have a way to disable/enable it on a per-page basis, or even in a middleware. Similary how the `pageTransition` and the `layoutTransition` values work.

This functionality would also be nice, to improve the view-transition implementation in [this PR](https://github.com/elk-zone/elk/pull/2211).

### Additional information

### Final checks

horvbalint:
### 🔗 Linked issue

resolve https://github.com/nuxt/nuxt/issues/25263

### ❓ Type of change

### 📚 Description

To solve the problem mentioned in the linked issue, this PR adds support for the `viewTransition` property inside PageMeta. 
Similary how pageTransition works, this can be enabled inside nuxt.config.ts's app property globaly, or on a per-page basis using the `definePageMeta` composable.
View transitions can also be enabled/disabled in a middleware using the `to.meta.viewTransition` value (again, similary how pageTransitions work).

~~After this PR the `experimental.viewTransition` property does not by itself enables view-transitions on the pages, but only loads the view-transition plugin inside nuxt. One have to enable view transitions using the `app.viewTransition` property or on a per-page basis.~~
As suggested in the code review below, if `experimental.viewTransition` is set, `app.viewTransition` uses its value as default.

We are not sure how to conditionaly enable the viewTransition property inside the d.ts files, so that typescript knows that the `PageMeta.viewTransition` is only available when the `experimental.viewTransition` value is set to `true`.

Initially, we also added checks for the `prefer-reduced-motion` user setting, so that nuxt would skip the transition if set to `reduce`. Later we found [this PR](https://github.com/nuxt/nuxt/pull/22292) which does the same, so we reverted that. If that PR gets merged, then this PR should be updated to allow `always` inside `PageMeta.viewTransition` , instead of leaving that in `experimental.viewTransition`, since we use that only as a feature guard.

We also tried to add documentation for the changes.

We are not very familiar with testing frameworks and are not sure if this functionality is testable or not. So if tests are needed, we would like to get some help :)

(we checked the breaking change checkbox, but since this is an experimental feature, this might not count as one)

### 📝 Checklist


packages/schema/src/config/app.ts
```diff
@@ -152,6 +152,16 @@ export default defineUntypedSchema({
...
 
+    /**
+     * Default values for view transitions.
+     *
+     * This can be overridden with `definePageMeta` on an individual page.
+     * Only booleans are allowed.
+     * @see https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental
+     * @type {typeof import('../src/types/config').NuxtAppConfig['viewTransition']}
+     */
+    viewTransition: false,
```
huang-julien, line 163: I think this should be default to `true` if the feature is enabled to avoid breaking changes for others projects that are already using `viewTransitionAPI`.

packages/schema/src/config/app.ts
```diff
@@ -152,6 +152,16 @@ export default defineUntypedSchema({
      */
     pageTransition: false,
 
+    /**
+     * Default values for view transitions.
+     *
+     * This can be overridden with `definePageMeta` on an individual page.
+     * Only booleans are allowed.
+     * @see https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental
```
huang-julien, line 160:
```suggestion
     * This can be overridden with `definePageMeta` on an individual page.
     * @see https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental
```

huang-julien (changes requested):

packages/schema/src/config/app.ts
horvbalint: I implemented this, but am not sure if I did it the correct way 😅

huang-julien (approved):
good to me ! thank you !

lukewarlow:
The PR with prefers reduced motion changes has been merged fyi

horvbalint:
Thanks!
I moved the `always` option from Experimental to PageMeta and with this I had to move the check into the `router.beforeResolve` callback, since `always` can now be set on a per page basis.

norbertSG:
Hello, I faced to an issue related to viewTransition, on android mobile(samsung) if we have a sticky bottom container, the elements/components inside are dissapear and appear on page change.
Take a look and let me know if there is any solution for this.

https://stackblitz.com/edit/nuxt-view-transitions-5q9zqa?file=app.vue

horvbalint:
@norbertSG this is not an issue with Nuxt, you can read about it here: https://www.nicchan.me/blog/view-transitions-and-stacking-context/
Basically you have to include the sticky elements in the transition as well :)

danielroe (approved):

PR #25282, opened 2024-01-18
fix(nuxt): fix import path of named components

### LINKED ISSUE #25281: Cannot register named exported components as client-only
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-jzao4p?file=modules%2Fsome-module%2Findex.ts

### Describe the bug

The reproduction contains a nuxt module that adds 2 components. `DComp` that is exported as `default` and `NComp` that is exported as named `NComp`. When registering these components as client-only via `addComponent` the `DComp` works correctry. However `NComp` is apparently not found.

### Additional context

_No response_

### Logs

```shell-script
[Vue warn]: Invalid vnode type when creating vnode: undefined.
[Vue warn]: Component <Anonymous> is missing template or render function.
```

becem-gharbi:
### 🔗 Linked issue

resolves #25281 

### ❓ Type of change

### 📚 Description

If a component is registered as client-only then when scanning a placeholder is created for it on server-side:
https://github.com/nuxt/nuxt/blob/2bfe01d26f962f0f296741ab2aad601d0aa577b6/packages/nuxt/src/components/module.ts#L175

When loading this component the import path depends on the `export` property:
https://github.com/nuxt/nuxt/blob/2bfe01d26f962f0f296741ab2aad601d0aa577b6/packages/nuxt/src/components/loader.ts#L79

The problem is that the placeholder is exported as `default` and the component is exported as named. So the import path is incorrect.

### 📝 Checklist


danielroe:
Any chance you could add a regression test to the basic/ test suite? (Just an example usage that previously would have failed.)

becem-gharbi:
I have added test named `import components` that checks if components are loaded correctly considering import type (default/named) and rendering mode (client/server/all)

PR #25286, opened 2024-01-18
fix(nuxt): fix lazy import of named components

### LINKED ISSUE #23768: Module-added component throws 'src.replace is not a function' if prefixed with Lazy and is not a vue file
### Environment

Stackblitz

### Reproduction

https://stackblitz.com/edit/nuxt-starter-5f85qa?file=app.vue

### Describe the bug

Given two components that produce the same output:

**JsComponent.js**
```js
import { h, defineComponent } from 'vue'

export const JsComponent = defineComponent({
  render() {
    return h('div', null, this.$slots.default?.())
  },
})
```

**VueComponent.vue**
```vue
<template>
  <div><slot /></div>
</template>
```

Added to the Nuxt app via `addComponent` in a Nuxt module definition:

```ts
addComponent({
  name: 'JsComponent',
  export: 'JsComponent',
  filePath: resolve('./JsComponent.js'),
})

addComponent({
  name: 'VueComponent',
  filePath: resolve('./VueComponent.vue'),
})
```

If prefixed with Lazy:

```vue
<template>
  <LazyVueComponent>Hello</LazyVueComponent>
  <LazyJsComponent>Hello</LazyJsComponent>
</template>
```

The .vue one works as expected, the .js one throws the following SSR error: `src.replace is not a function`.

After dealing with this a bit, I found out that if the js component has no child, it doesn't throw any error:

```vue
<LazyJsComponent></LazyJsComponent>
```

But even a single char or a comment will break things again. 

### Additional context

_No response_

### Logs

```shell-script
[Vue warn]: Invalid vnode type when creating vnode: undefined.
[nuxt] [request error] [unhandled] [500] src.replace is not a function
```
```

becem-gharbi:
### 🔗 Linked issue

resolves #23768 

### ❓ Type of change

### 📚 Description

When lazy loading components the dynamic import always return the exported `default`. So named exported components cannot be lazy loaded.

https://github.com/nuxt/nuxt/blob/8e4905c2fe4531f27aeb45b1b06d2fc03868e2a3/packages/nuxt/src/components/loader.ts#L77

### 📝 Checklist


danielroe:
Any chance you could add a regression test to the basic/ test suite? (Just an example usage that previously would have failed.)

becem-gharbi:
I'm learning to use `@nuxt/test-utils`. I'm not sure if there is a setup function to register components with `addComponent`. If there is a guide or a similar test please let me know.

becem-gharbi:
I have added test named `lazy import components` that checks if components are lazy loaded correctly considering import type (named) and rendering mode (client/server/all)

danielroe (approved):

Issue #25302, opened 2024-01-19
`defineModel` shoule enable by default


baixiaoyu2997:
### Environment

no need

### Reproduction

no need

### Describe the bug

defineModel stable since `vue v3.4`

manniL:
It is!

https://stackblitz.com/edit/nuxt-issues-25302

If you are on an earlier nuxt version, make sure you update your Vue version / regenerate the lockfile ;)

baixiaoyu2997:
thanks for reply,  I actually mean the documentation, and the type hints for the configuration file.
<img width="1234" alt="image" src="https://github.com/nuxt/nuxt/assets/9149999/7fac1461-4f69-4c3c-8ec9-3515c6403ca2">
![image](https://github.com/nuxt/nuxt/assets/9149999/89839446-dbe8-4508-9acd-4070a2c33c7c)
<img width="802" alt="image" src="https://github.com/nuxt/nuxt/assets/9149999/058dc200-c1a7-4069-924b-035e2f4193fd">

manniL:
Good point 👍

manniL:
Resolved via #25306

PR #25306, opened 2024-01-19
chore: remove reference to "experimental" `defineModel`

TheAlexLichter:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25302

### ❓ Type of change

### 📚 Description

As `defineModel` is stable now, we can remove (or alternatively deprrecate) the configuration option for it, as well as remove the docs reference.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/app.ts
+++ b/packages/schema/src/config/app.ts
@@ -28,13 +28,6 @@ export default defineUntypedSchema({
      * @type {boolean}
      */
     propsDestructure: false,
-
-    /**
-     * Vue Experimental: Enable macro `defineModel`
-     * @see [Vue RFC#503](https://github.com/vuejs/rfcs/discussions/503)
-     * @type {boolean}
-     */
-    defineModel: false
   },
 
   /**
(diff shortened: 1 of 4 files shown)


packages/schema/src/config/app.ts
```diff
@@ -28,13 +28,6 @@ export default defineUntypedSchema({
      * @type {boolean}
      */
     propsDestructure: false,
-
-    /**
-     * Vue Experimental: Enable macro `defineModel`
-     * @see [Vue RFC#503](https://github.com/vuejs/rfcs/discussions/503)
-     * @type {boolean}
```
manniL, line 35: As a note: Also happy to *deprecate* it to avoid breaking type-changes (as it does nothing anyway right now)

danielroe (approved):

Revadike:
Am I doing something wrong here, or should I submit a bug report, as this is not working:

```html
<script setup>
  const [modelValue, modelModifiers] = defineModel({
    // get() omitted as it is not needed here
    set(value) {
      // if the .capitalize modifier is used, return capitalized value
      if (modelModifiers.capitalize) {
        return value.charAt(0).toUpperCase() + value.slice(1)
      }
      // otherwise, return the value as-is
      return value
    }
  })
</script>

<template>
  <v-container>
    <v-text-field v-model.capitalize="modelValue" />
  </v-container>
</template>
```


https://vuejs.org/api/sfc-script-setup.html#definemodel
https://vuejs.org/guide/components/v-model.html

PR #25318, opened 2024-01-19
fix(nuxt): load plugin that depends on loaded and unloaded plugins

### LINKED ISSUE #25305: Plugin that `dependsOn` both resolved and unresolved plugins is not loaded
### Discussed in https://github.com/nuxt/nuxt/discussions/25301

<div type='discussions-op-text'>

<sup>Originally posted by **hitochan777** January 19, 2024</sup>
I am new to Nuxt3 (though I have been using Nuxt2) and I am not sure if this is a bug or expected behavior so I am posting as discussion.

### Environment
Node Version: v18.18.0
Nuxt Version: 3.9.3

### Reproduction
https://stackblitz.com/edit/github-pwakuy?file=plugins%2Fb.ts

### Describe the behavior

I have defined three plugins `a.ts`, `b.ts`, `c.ts`.
`b` depends on `a` and `c` so I used `dependsOn` options as follows.
Other plugins are almost the same as plugin `b` except they do not define any dependencies.

```ts
  name: 'b',
  dependsOn: ['a', 'c'],
  setup() {
    console.log('setup plugin b');
  },
});
```

I expect plugins to be loaded in the order of `a`, `c`, `b`.
To check if it is loaded in this order I put console.log in setup function.
But when I ran the dev server and server receives a request, it just shows the following in the console log and it seems plugin `b` is not loaded.
If I remove the dependency on `a`, plugin `b` is loaded.

```
setup plugin a
setup plugin c
```

I have read the [Nuxt code on dependsOn](https://github.com/nuxt/nuxt/blob/6736decba0dc8bbc7276848a8a26973440ebdee4/packages/nuxt/src/app/nuxt.ts#L375), and it seems after an plugin is applied it deletes itself from `dependsOn` of unresolved plugins, and only if `dependsOn` becomes empty it applies the unresolved plugins.
But if `dependsOn` contains plugins that are executed earlier than the unresolved plugin, `dependsOn` will never be empty and its plugin will never be applied.
Using the example above, when plugin c is applied it deletes itself from `dependsOn` of plugin b and it becomes `["a"]`, not empty. Thus, plugin b will never be applied.

I am sorry if my understanding is incorrect. But is this expected behavior or a bug?</div>

hitochan777:
### 🔗 Linked issue

#25305

### ❓ Type of change

### 📚 Description

Resolves #25305 

Originally, when a plugin is loaded, if there is any unresolved plugin depended by the plugin all its dependencies including the ones already resolved are marked as resolve needed.
Thus even when all its unresolved plugin are resolved, it is not loaded because it still has plugins that are by mistake considered unresolved but actually have already been resolved.

I made change to the code so that only unresolved plugins are marked as resolve needed when a plugin is loaded.

### 📝 Checklist

### DIFF
--- a/test/nuxt/plugin.test.ts
+++ b/test/nuxt/plugin.test.ts
@@ -249,19 +249,19 @@ describe('plugin dependsOn', () => {
     const nuxtApp = useNuxtApp()
     const sequence: string[] = []
     const plugins = [
-      pluginFactory('A', undefined, sequence),
-      pluginFactory('B', ['A', 'C'], sequence),
-      pluginFactory('C', undefined, sequence)
+      pluginFactory('A', undefined, sequence, false),
+      pluginFactory('B', ['A', 'C'], sequence, false),
+      pluginFactory('C', undefined, sequence, false),
     ]
     await applyPlugins(nuxtApp, plugins)
 
     expect(sequence).toMatchObject([
       'start A',
-      'start C',
       'end A',
+      'start C',
       'end C',
       'start B',
-      'end B'
+      'end B',
     ])
   })
 })
--- a/test/nuxt/plugin.test.ts
+++ b/test/nuxt/plugin.test.ts
@@ -244,4 +244,24 @@ describe('plugin dependsOn', () => {
       'end B'
     ])
   })
+
+  it('expect B to execute after A, C when B depends on A and C', async () => {
+    const nuxtApp = useNuxtApp()
+    const sequence: string[] = []
+    const plugins = [
+      pluginFactory('A', undefined, sequence),
+      pluginFactory('B', ['A', 'C'], sequence),
+      pluginFactory('C', undefined, sequence)
+    ]
+    await applyPlugins(nuxtApp, plugins)
+
+    expect(sequence).toMatchObject([
+      'start A',
+      'start C',
+      'end A',
+      'end C',
+      'start B',
+      'end B'
+    ])
+  })
 })
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -368,8 +368,9 @@ export async function applyPlugins (nuxtApp: NuxtApp, plugins: Array<Plugin & Ob
   let promiseDepth = 0
 
   async function executePlugin (plugin: Plugin & ObjectPlugin<any>) {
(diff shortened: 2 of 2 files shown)


test/nuxt/plugin.test.ts
```diff
@@ -244,4 +244,24 @@ describe('plugin dependsOn', () => {
       'end B'
```
huang-julien, line 244: strange i havn't been able to make this test fail on the main branch

test/nuxt/plugin.test.ts
```diff
@@ -244,4 +244,24 @@ describe('plugin dependsOn', () => {
...
+    await applyPlugins(nuxtApp, plugins)
+
+    expect(sequence).toMatchObject([
+      'start A',
+      'start C',
+      'end A',
+      'end C',
+      'start B',
+      'end B'
+    ])
```
huang-julien, line 265:
```suggestion
    const plugins = [
      defineNuxtPlugin({
        name: 'A',
        async setup () {
          sequence.push('start A')
          sequence.push('end A')
        }
      }),
      defineNuxtPlugin({
        name: 'B',
        dependsOn: ['A', 'C'],
        async setup () {
          sequence.push('start B')
          sequence.push('end B')
        }
      }),
      defineNuxtPlugin({
        name: 'C',
        async setup () {
          sequence.push('start C')
          sequence.push('end C')
        }
      })
    ]
    await applyPlugins(nuxtApp, plugins)

    expect(sequence).toMatchObject([
      'start A',
      'end A',
      'start C',
      'end C',
      'start B',
      'end B',
    ])
```

huang-julien: `pluginFactory` is returning plugins that are parallels by default and are waiting 10ms before resolving. The reproduction only contains plugins that are not parallel

hitochan777: > The reproduction only contains plugins that are not parallel

Yes, you are right!
But isn't it better to test both cases (parallel and non-parallel case), or is it just enough to test the reproduction in the issue?

huang-julien: I think the case should be already tested by the test at line 124, i'm ok with keeping both.

hitochan777: I see. I started to think keeping both is a bit redundant I will keep only your suggestion. Thanks.

huang-julien (approved):

Issue #25000, opened 2024-01-02
TS comments from schema are not passed into bundled types


TheAlexLichter:
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.9.0`
- Nuxt Version:     `3.9.0`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `pnpm@8.5.0`
- Builder:          `-`
- User Config:      `runtimeConfig`, `routeRules`, `modules`, `nitro`, `plausible`, `content`, `tailwindcss`, `devtools`, `experimental`, `future`, `watch`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.10.3`, `@vueuse/nuxt@10.7.1`, `nuxt-icon@0.6.8`, `@nuxt/content@2.10.0`, `nuxt-simple-sitemap@4.2.3`, `nuxt-schema-org@3.3.0`, `nuxt-og-image@3.0.0-rc.14`, `@nuxt/image@1.1.0`, `@nuxtjs/plausible@0.2.4`
- Build Modules:    `-`

### Reproduction

Not on StackBlitz because TS support is flaky IIRC.

1. `npx nuxi init my-project` + follow process
2. Change your nuxt config to the following:

```ts
export default defineNuxtConfig({
  experimental: {
    headNext: true
  }
})
```

3. Check the type definitions and see that there is no info, only the default value

<img width="501" alt="image" src="https://github.com/nuxt/nuxt/assets/640208/170b912d-3bdc-4267-8e13-fed9b58d452b">

---

Also [visible on UnPKG](https://unpkg.com/browse/@nuxt/schema@3.9.0/dist/index.d.ts#L993).

### Describe the bug

Lots of types have a great TSDoc description (see e.g. [headNext](https://github.com/nuxt/nuxt/blob/4dd87f6e281bacacafa9566a21388d5a7a53f911/packages/schema/src/config/experimental.ts#L241-L247) in the source code) but it is not shown nor exported to the dist types.

It should be though!

danielroe:
This seems to be an upstream issue in untyped -  see https://github.com/nuxt/nuxt.com/issues/1415.

Norbiros:
@danielroe it seems to be resolved in the latest version

PR #25319, opened 2024-01-19
fix(schema): pin jiti until babel bug is resolved

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25000
https://github.com/nuxt/nuxt.com/issues/1415

### ❓ Type of change

### 📚 Description

This temporarily pins `jiti` to ensure we can generate comments from docs, until the upstream issue is resolved.

(Seems likely an issue in Babel itself, but we need to reproduce and file if possible.)

cc: @manniL, @pi0 

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@nuxt/schema": "workspace:*",
     "@nuxt/vite-builder": "workspace:*",
     "@nuxt/webpack-builder": "workspace:*",
+    "jiti": "1.20.0",
     "rollup": "^4.9.5",
     "nuxt": "workspace:*",
     "vite": "5.0.11",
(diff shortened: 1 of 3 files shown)


pi0:
Have you tried https://github.com/nuxt/nuxt.com/issues/1415#issuecomment-1858520738 as fix in untyped?

danielroe:
Oops, sorry for the review ping. No, I didn't try that.

pi0:
It can be dangerous if we hide jiti v1 possible regressions for Nuxt to be caught in monorepo because of this.

danielroe:
Indeed, that's true. We could also pin this only before releasing/building schema 😆

pi0:
Checking issue now let's see if we can solve it easily otherwise sure pinning only for schema makes sense as hotfix 👍🏼

pi0:
🐛 https://github.com/unjs/untyped/commit/7a941412c2ca6c78f1fc1bc6a7fb9f2701d4be2a

PR #25390, opened 2024-01-23
fix(nuxt): set `useFetch` accept default to `application/json`

### LINKED ISSUE #23539: "No match found for location" is logged at least 8 times
### Environment

- Node Version:     v18.12.1
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Builder:          -
- User Config:      modules, devtools
- Runtime Modules:  @nuxtjs/i18n@8.0.0-rc.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-lip1ot?file=pages%2Findex.vue,error.vue,app.vue 

### Describe the bug

When a not existing route/path is opened, the warning `[Vue Router warn]: No match found for location with path "/not-existing"` is logged 8 times. In our project (with translated paths via @nuxtjs/i18n and i18n pages) it logs even move times.

1. click the link "Link to not available page" in the preview of stackblitz to navigate to a not existing route
2. reload the page in preview window
3. see server log that logs 8 times `[Vue Router warn]: No match found for location with path "/not-existing"`

### Additional context

Im investigating a memory leak at server side. And this is potentially part of the issue. It seems like that multiple instances of the router are created.

When using a native Vue application with separat installed vue-router package the warning is only logged once.

### Logs

`[Vue Router warn]: No match found for location with path "/not-existing"` x 8

### LINKED ISSUE #24609: API error data gets rendered as HTML during SSR
### Environment

- Operating System: Linux
- Node Version:     v21.2.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.0
- Package Manager:  bun@1.0.13

### Reproduction

https://stackblitz.com/edit/nuxt-starter-typhrm

### Describe the bug

When I throw an error with `createError` in an API handler, and then want to retrieve its data in an app, instead of JSON response I receive an HTML page. This makes it impossible to retrieve any additional data I passed in with the error, and also bloats the `__NUXT_DATA__` object.  

Happens only during SSR, client side requests do not have this problem. Also works when the error is actually thrown, as simply responding with an error response status works fine.

Renders HTML ❌
Screenshot_20231205_155553

Renders JSON ✅
Screenshot_20231205_155614

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #24742: Vue Router warning when throwing error in api route
### Environment

- Node Version:     v18.17.1
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.6.7
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-jpycru?file=server%2Fapi%2Findex.get.ts

### Describe the bug

Error Output:
`[Vue Router warn]: No match found for location with path "/api"`

This error has only occurred recently. When I return an error in an API route or utility function, a Vue Router warning appears in the console.

### Additional context

_No response_

### Logs

```shell-script
[Vue Router warn]: No match found for location with path "/api"
```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24742
resolves https://github.com/nuxt/nuxt/issues/24609
resolves https://github.com/nuxt/nuxt/issues/23539

### ❓ Type of change

### 📚 Description

When using `useFetch`, we automatically use the request event fetch, which preserves headers from the initial request. Unfortunately, this also proxies the `accept` header, which is `text/html` from the browser, and can result in all local fetches with `useFetch` returning HTML error messages rather than JSON.

This PR also normalises the headers object (which was not previously working with nested refs).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -21,7 +21,7 @@ type ComputedOptions<T extends Record<string, any>> = {
 }
 
 interface NitroFetchOptions<R extends NitroFetchRequest, M extends AvailableRouterMethod<R> = AvailableRouterMethod<R>> extends FetchOptions {
-  method?: M;
+  method?: M
 }
 
 type ComputedFetchOptions<R extends NitroFetchRequest, M extends AvailableRouterMethod<R>> = ComputedOptions<NitroFetchOptions<R, M>>
@@ -130,7 +130,7 @@ export function useFetch<
   } = opts
 
   const _fetchOptions = reactive({
-    ...fetchDefaults,
+    ...fetchDefaults as typeof fetchOptions,
     ...fetchOptions,
     cache: typeof opts.cache === 'boolean' ? undefined : opts.cache
   })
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -170,16 +170,28 @@ export function useFetch<
...
     let _$fetch = opts.$fetch || globalThis.$fetch
+    const headers = Object.create(null)
 
     // Use fetch with request context and headers for server direct API calls
     if (import.meta.server && !opts.$fetch) {
       const isLocalFetch = typeof _request.value === 'string' && _request.value[0] === '/' && (!toValue(opts.baseURL) || toValue(opts.baseURL)![0] === '/')
       if (isLocalFetch) {
         _$fetch = useRequestFetch()
+        // Set default `accept` value to avoid it defaulting to `text/html` (from initial browser request)
+        headers.accept = 'application/json'
```
pi0, line 181: Maybe instead avoid irritating accept? This seems a bug that we proxy browser's accept header down to API calls. ofetch defaults to the right accept header based on body but if we explicitly ovveride here, it cannot do this anymore.

danielroe: Okay, let's fix it upstream then 👍

danielroe: ~> https://github.com/unjs/h3/pull/646

danielroe:
closing ~> https://github.com/unjs/h3/pull/646

PR #25395, opened 2024-01-23
test: avoid explicit timeouts

danielroe:
### ❓ Type of change

### 📚 Description

This attempts to address some persistent test failures, some of which I think are attributable to explicit timeouts.

This uses a custom assertion to wait for the global route having changed within the nuxt app, rather than looking at url or just for network idleness. I hope this should be less flaky.

But would appreciate a check just to see whether there was an explicit reason for some of the timeouts, e.g. on L366-367.

### 📝 Checklist

### DIFF
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -7,7 +7,6 @@ import { join, normalize } from 'pathe'
 import { $fetch, createPage, fetch, isDev, setup, startServer, url, useTestContext } from '@nuxt/test-utils/e2e'
 import { $fetchComponent } from '@nuxt/test-utils/experimental'
 
-import type { ConsoleMessage } from 'playwright-core'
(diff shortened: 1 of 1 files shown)


test/basic.test.ts
```diff
@@ -363,8 +362,7 @@ describe('pages', () => {
     // don't expect any errors or warning on client-side navigation
     const { page: page2, consoleLogs: consoleLogs2 } = await renderPage('/')
     await page2.locator('#to-client-only-components').click()
-    // force wait for a few ticks
-    await page2.waitForTimeout(50)
```
huang-julien, line 367: Yeah i think i wanted to wait for several ticks, I wans't ware of the `waitFor` at that time 😅

huang-julien (approved):

PR #24922, opened 2023-12-27
feat(nuxt): add `pages:routerOptions` hook

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24141

### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This adds a new hook: `pages:routerOptions`. It allows modifying the detected `app/router.options.ts` files from layers, as well as adding new router options (for example, from a module).

If `optional: true` is set, then this will not force pages integration. (Which is ideal for a custom scroll behaviour, for example.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #25397, opened 2024-01-23
fix(nuxt): add router.options files in definite order

danielroe:
### ❓ Type of change

### 📚 Description

In debugging https://github.com/nuxt/nuxt/actions/runs/7628326647/job/20779171448?pr=25395, realised that we in fact were resolving `app/router.options` files in a non-reproducible order (since https://github.com/nuxt/nuxt/pull/24922).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -38,10 +38,10 @@ export default defineNuxtModule({
       // Add default options
       context.files.push({ path: resolve(runtimeDir, 'router.options'), optional: true })
 
-      await Promise.all(nuxt.options._layers.map(async layer => {
+      for (const layer of nuxt.options._layers) {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/module.ts
```diff
@@ -38,10 +38,10 @@ export default defineNuxtModule({
       // Add default options
       context.files.push({ path: resolve(runtimeDir, 'router.options'), optional: true })
 
-      await Promise.all(nuxt.options._layers.map(async layer => {
+      for (const layer of nuxt.options._layers) {
```
manniL, line 41: Couldn't we use await Promise.all and then loop over the resolved promises?

danielroe: Oops, yes, we could indeed.

PR #25398, opened 2024-01-23
refactor(nuxt): replace Nuxt payload's error type with NuxtError

DamianGlowala:
### ❓ Type of change

### 📚 Description

Enforces a `NuxtError` type on the `NuxtPayload`'s error property for consistency.

### 📝 Checklist


danielroe (commented):
Looks great! We should also update the error handling docs where we describe the prop type in `error.vue`.

DamianGlowala:
Added! As the `nuxt-error-page.vue` isn't using TypeScript, we probably need to leave it as it is:
https://github.com/nuxt/nuxt/blob/a995f724eadaa06d5443b188879ac18dfe73de2e/packages/nuxt/src/app/components/nuxt-error-page.vue#L9

docs/1.getting-started/8.error-handling.md
```diff
@@ -86,8 +86,10 @@ Customize the default error page by adding `~/error.vue` in the source directory
 
 ```vue [error.vue]
 <script setup lang="ts">
+import type { NuxtError } from '#app'
+
 const props = defineProps({
-  error: Object
+  error: NuxtError
```
danielroe, line 92: This should be `Object as () => NuxtError` here and below.

DamianGlowala: Good point. Do we want to switch to the following though?
```
const props = defineProps<{
  error: NuxtError
}>()
```

danielroe: No, I think not. It will require vue to have typescript installed as a peerDependency, and doesn't add any additional validation or type safety

danielroe (approved):

PR #25410, opened 2024-01-24
fix(nuxt): add missing type for `head` in `defineNuxtComponent`

### LINKED ISSUE #24811: Type missing: for `head(nuxtApp) {}`
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.5.0`
- Nuxt Version:     `3.8.2`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `npm@10.2.3`
- Builder:          `-`
- User Config:      `devtools`, `build`, `modules`, `vite`
- Runtime Modules:  `()`
- Build Modules:    `-`

### Reproduction

use the code from the documentation: https://nuxt.com/docs/api/utils/define-nuxt-component#head

### Describe the bug

As shown in the documentation: https://nuxt.com/docs/api/utils/define-nuxt-component#head

Using that code (`head(nuxtApp) {}`), then `nuxtApp` has no type, implicitly "any".

Bildschirmfoto 2023-12-18 um 14 35 42

### Additional context

_No response_

### Logs

_No response_

mateuszkulpa:
### 🔗 Linked issue

#24811

### ❓ Type of change

### 📚 Description
Fixes missing type definition for `head` function in the `defineNuxtComponent`

Resolves #24811

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/types/augments.d.ts
+++ b/packages/nuxt/src/app/types/augments.d.ts
@@ -1,3 +1,4 @@
+import type { UseHeadInput } from "@unhead/vue";
 import type { NuxtApp, useNuxtApp } from '../nuxt'
 
 interface NuxtStaticBuildFlags {
@@ -31,5 +32,8 @@ declare module 'vue' {
   }
   interface ComponentInternalInstance {
     _nuxtOnBeforeMountCbs: Function[]
+  }
+  interface ComponentCustomOptions {
+    head?(nuxtApp: NuxtApp): UseHeadInput
   }
 }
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/types/augments.d.ts
```diff
@@ -32,4 +33,7 @@ declare module 'vue' {
   interface ComponentInternalInstance {
     _nuxtOnBeforeMountCbs: Function[]
   }
+  interface ComponentCustomOptions {
+    head?(nuxtApp: NuxtApp): UseHeadInput
```
danielroe, line 37: I guess the only issue here is that this will mark this as a valid type for `defineComponent` as well, right? Is there a way we can disallow that, or perhaps we could add a JSdoc comment here?

mateuszkulpa: I've added a comment. I tried to extend the type exclusively for `defineNuxtComponent`, but this would have required copying many types from `vue`, as `defineComponent` has numerous overloads. If you have any ideas for a better approach, I am open to suggestions for improvement.

PR #25431, opened 2024-01-25
fix(nuxt): on `callOnce` only run once callbacks with falsy return values

### LINKED ISSUE #25419: `callOnce` run multiple times in client-only components
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-6wtgz2?file=app.vue

### Describe the bug

The reproduction contains a client-only component. This component calls `callOnce` on setup.
The problem is when multiple instances are rendered the callbacks of `callOnce` are all executed. 
I'm not sure if this is intentional or an issue with reactivity.

### Additional context

_No response_

### Logs

_No response_

becem-gharbi:
### 🔗 Linked issue

resolves #25419 

### ❓ Type of change

### 📚 Description

This PR is just a suggestion to solve the related issue in code instead of on documentation. 

When registering a callback with falsy return value (undefined), it gets triggered multiple times by `callOnce` (same key). As explained in issue, the callback should return a truthy value in order to handle the evaluation below correctly. For better DX, I think the return value should be irrelevant to the logic. 

https://github.com/nuxt/nuxt/blob/44fcacba846a6cdf33e1475fc8413f6ae49d1fe4/packages/nuxt/src/app/composables/once.ts#L28

The proposed solution is to add the key in `payload.once` before the evaluation. Plus this makes sure that if the callback throws an error it will not get called again.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/once.ts
+++ b/packages/nuxt/src/app/composables/once.ts
@@ -24,9 +24,9 @@ export async function callOnce (...args: any): Promise<void> {
   if (nuxtApp.payload.once.has(_key)) {
     return
   }
+  nuxtApp.payload.once.add(_key)
   nuxtApp._once = nuxtApp._once || {}
   nuxtApp._once[_key] = nuxtApp._once[_key] || fn()
   await nuxtApp._once[_key]
-  nuxtApp.payload.once.add(_key)
   delete nuxtApp._once[_key]
 }
--- a/packages/nuxt/src/app/composables/once.ts
+++ b/packages/nuxt/src/app/composables/once.ts
@@ -24,9 +24,9 @@ export async function callOnce (...args: any): Promise<void> {
   if (nuxtApp.payload.once.has(_key)) {
     return
   }
-  nuxtApp.payload.once.add(_key)
   nuxtApp._once = nuxtApp._once || {}
   nuxtApp._once[_key] = nuxtApp._once[_key] || fn() || true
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/composables/once.ts
```diff
@@ -24,9 +24,9 @@ export async function callOnce (...args: any): Promise<void> {
   if (nuxtApp.payload.once.has(_key)) {
     return
   }
+  nuxtApp.payload.once.add(_key)
   nuxtApp._once = nuxtApp._once || {}
   nuxtApp._once[_key] = nuxtApp._once[_key] || fn()
```
danielroe, line 29: How about instead:

```suggestion
  nuxtApp._once[_key] = nuxtApp._once[_key] || fn() || true
```

This adds a 'true' default value for non-async once callbacks, instead of early returning when a promise has not been resolved.

Otherwise we will not await the promise if one is running in parallel.

becem-gharbi: Now I get it. When running in parallel all `callOnce` with the same key should resolve at the same time. Thanks for the clarification.

danielroe (approved):

PR #25435, opened 2024-01-25
feat(nuxt): `trailingSlash` config & `createRouteResolver` util

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/15462

### ❓ Type of change

### 📚 Description

There are several DX improvements we can make within the core for users who want to use appending trailing slashes in their Nuxt apps.

#### Nuxt Config

We keep existing config types from `NuxtLink`, in that end users can configure a trailing slash as either:
- `append` - add trailing slash `/foo/`
- `remove` - remove trailing slash `/foo`
- `undefined` - as is

Config is added to `app` namespace.

```ts
  app: {
    trailingSlash: 'append',
  },
})
```

Side note: I'd prefer to have this as a `boolean` where it is always `false` or `true` but I think keeping consistent config APIs is more important.

#### Opportunities

- Setting slashes for `<NuxtLink>` and `navigateTo` paths
- Set `NuxtPage` paths to use trailing slash, which in turn generates the correct types when using `experimental.typedPages`.
- Giving modules a core config that they can use instead of rolling their own
- :warning: Redirecting non-slash URLs to slashed version (Not currently in scope of this PR)

#### createRouteResolver

In introducing new logic on how paths are resolved, we're introducing new logic to `<NuxtLink>` and `navigateTo`. To reduce the boilerplate, make testing easier, and improve core APIs, we introduce the `createRouteResolver` composable. 
This composable allows the core, module authors, and end users to resolve paths while respecting the trailing slashes and the base path.

We need to have a composable to create the resolver as we need to instantiate `useRouter()` within this function that won't be accessible async. 

```ts
const resolveRoute = createRouteResolver({ trailingSlash })
const path = computed(() => {
 return resolveRoute(props.to)
})
```

This is inspired by the logic of the [nuxt-site-config](https://github.com/harlan-zw/nuxt-site-config/blob/main/packages/site-config/src/urls.ts) module which has been battle-tested at this point.

:information_source: Looking for feedback on this.

#### Additional Notes

- Trailing slashes on file links are currently broken `<NuxtLink to="/file.pdf">` -> `href="/file.pdf/`
- The logic for handling whether links are external is currently inconsistent between `navigateTo` and `NuxtLink`
- - `NuxtLink` - any link that has a protocol or has a `target` that isn't `self` (new tab)
- - `navigateTo` - any link that has a protocol 

#### To Do

- [ ] Fix logic (hacked away at it, very likely broken)
- [ ] Tests (likely we're introducing regressions with current code).
- [ ] Documentation (will need docs for `createRouteResolver` and `app.trailingSlash`)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -1,6 +1,6 @@
 import type { ComputedRef, DefineComponent, InjectionKey, PropType } from 'vue'
 import type { RouteLocationRaw } from '#vue-router'
-import type { NuxtAppConfig } from "@nuxt/schema"
+import type { NuxtAppConfig } from 'nuxt/schema'
 import { computed, defineComponent, h, inject, onBeforeUnmount, onMounted, provide, ref, resolveComponent } from 'vue'
 import { hasProtocol, joinURL, parseQuery, parseURL } from 'ufo'
 import { preloadRouteComponents } from '../composables/preload'
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -5,7 +5,7 @@ import { computed, defineComponent, h, inject, onBeforeUnmount, onMounted, provi
 import { hasProtocol, joinURL, parseQuery, parseURL } from 'ufo'
 import { preloadRouteComponents } from '../composables/preload'
 import { onNuxtReady } from '../composables/ready'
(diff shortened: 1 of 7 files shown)


danielroe:
Note that we already have partial support for this via https://github.com/nuxt/nuxt/pull/23724.

harlan-zw:
> Note that we already have partial support for this via #23724.

Yes, Nuxt SEO already implements it. This is to solve outstanding module inconsistencies mainly (and vue-router types).

Issue #25451, opened 2024-01-26
On production build, CSS does not get inlined in SSR response in some cases when using `defineAsyncComponent`


tibineagu:
### Environment

```
Working directory: /home/projects/nuxt-starter-ta6mqt
Nuxt project info:

------------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------
```

### Reproduction

[Reproduction repository](https://stackblitz.com/edit/nuxt-starter-ta6mqt)

- run `npm run build && npm run preview`
- access the page and look at the HTML response
- notice that the styles for both `layout.vue` and `AsyncCompParent.vue` are NOT inlined in the response

### Describe the bug

There seems to be an issue when using a common mapping object to create associations to various components that need to be loaded dynamically.

This is a rather complex system so please follow along carefully.
- we have an enum in `@/utils/name-map.js`
- this enum is used in a parent component (`AsyncCompParent`) that will decide which other component to render based on an association map
    - these child components (`ChildComponentA`) are declared using `defineAsyncComponent` in order to achieve code splitting
- The async component (`ChildComponentA`) uses a composable (`useSayMyName.js`) that, in turn, ALSO imports the enum

The setup above will cause the following:
- no styles declared in `AsyncCompParent` are inlined in the SSR HTML response
- no styles in `layout.vue` (which is the _parent_ of `AsyncCompParent`) are inlined in the SSR HTML response

#### Expected behaviour
The styles for `layout.vue` and `AsyncCompParent.vue` should be inlined in the SSR HTML response, since these components are always present on the page and rendered at load time.

### Additional context

The repo above is an extremely simplified version of code we ported from a Nuxt 2 app, where it functions as expected.

More background / things we discovered:
- if, in any of the two files that import from `@/utils/name-map`, we instead redeclare a local copy of that constant, the build goes ahead as expected and the CSS is inlined (see comment in the code)
- if we don't use `defineAsycnComponent`, and just use plain a `import` statement for `ChildComponentA`, the build goes ahead as expected
- if we don't import `useSayMyName` in `ChildComponentA`, and instead import the enum directly there, the build goes ahead as expected

[...]

Issue #25420, opened 2024-01-24
missing important inline styles from components used across multiple pages


gibkigonzo:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  yarn@1.22.19
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction
**!!! It only appears in production mode. Run** `yarn build && yarn preview`

https://codesandbox.io/p/devbox/affectionate-lederberg-vz2xsy

### Describe the bug
When a component is used across multiple pages, the styles associated with that component do not appear as inline styles. This issue arises because these components are likely candidates for quick loading. For instance, I have an icon component that appears on every page. From a performance perspective, it's crucial to load the styles from this component as soon as possible, similar to critical CSS. In a real application, there may be components responsible for establishing the page layout, and they should ideally be loaded as soon as possible.

In reproduction, you'll notice that the "IconComponent" turns red once the styles from the file are applied. This delay increases the First Contentful Paint (FCP). Those styles should be loaded as critical css (which worked that way in nuxt 2).

![Screenshot from 2024-01-24 16-12-44](https://github.com/nuxt/nuxt/assets/42383376/f42ea46b-f876-4979-9d0d-01e564d3cd54)

However, if you delete the `test.vue` page, the styles from the "IconComponent" are added as inline CSS.

![Screenshot from 2024-01-24 16-24-12](https://github.com/nuxt/nuxt/assets/42383376/90d77cd6-fbae-482d-9aaf-c5c2517447b4)

Is this behavior expected? Is there a method to enforce inline styles for certain components?

Could this issue be related to https://github.com/nuxt/nuxt/issues/22507 ?

### Additional context

Even when I add this component as suggested here https://github.com/nuxt/nuxt/issues/22507#issuecomment-1757833275
Those styles are still not included as inline.
https://codesandbox.io/p/devbox/gifted-noyce-j3rncx

![Screenshot from 2024-01-24 16-42-50](https://github.com/nuxt/nuxt/assets/42383376/d963b661-9d14-42f0-9162-c1498b206042)

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/25455

PR #25394, opened 2024-01-23
fix(vite): inline styles for imports used only on server

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/24480
resolves https://github.com/nuxt/nuxt/issues/25330

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Not quite sure why, but we were previously not extracting styles for components that didn't have a counterpart in the client build. This change avoids the early return, but needs some extra testing to see what I was thinking in https://github.com/nuxt/nuxt/pull/21573 🤔 

It also needs a simple test.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

Issue #25330, opened 2024-01-19
Styling of Regular Component Lost in Production When Nested Inside Global Server Component


vsergiu93:
### Environment

- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.1.0
- Builder:          -
- User Config:      devtools, experimental, modules, srcDir
- Runtime Modules:  @nuxtjs/tailwindcss@6.10.4
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-rkhot3?file=README.md

### Component Setup: 
- **Component A**: This is a global server component (A.server.global.vue)
- **Component B**: A regular component with specific scoped styling. (B.vue)

### Behavior Observed:

- In development mode (`nuxt dev`), everything works as expected. component B's styling is correctly applied when it is used inside component A.
- In production mode (`npm run build && node .output/server/index.mjs`), the styling of component B is not applied when it is used inside component A.

I provided a stackblitz reproduction link, but that starts directly in dev mode, and as I said above everything works in dev mode, the stackblitz needs to be started in production mode.

huang-julien:
Hi :wave: is it reproductible with a non global server component ? If so, this may be a duplicate of https://github.com/nuxt/nuxt/issues/24480

vsergiu93:
Hi @huang-julien, thanks for checking, you are right, it is reproductible with a non global server component, but please note that for me this only happens in production build not in dev mode.

imnaK:
I may not have an in-depth understanding of every aspect, but I will make an effort to respond to all requests for additional information.

Does my test repository replicate the same issue presented here, or should I create a new issue for it?

**Repo:**
https://github.com/imnaK/nuxt-minimal-starter

**Bug in short:**
My style ref will not be applied on a scroll listener on production/preview server. It will work as expected if I switch pages or on dev server.

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/25455

PR #25455, opened 2024-01-26
fix(vite): extract styles for shared chunks

danielroe:
### 🔗 Linked issue

- [x] resolves https://github.com/nuxt/nuxt/issues/25420
- [x] resolves https://github.com/nuxt/nuxt/issues/25451
- [x] resolves https://github.com/nuxt/nuxt/issues/22507
- [x] resolves https://github.com/nuxt/nuxt/issues/24480
- [x] resolves https://github.com/nuxt/nuxt/issues/25330

closes https://github.com/nuxt/nuxt/pull/25394

### ❓ Type of change

### 📚 Description

We weren't extracting styles for chunks without a `facadeModuleId`, but it's safe to do so for individual module ids. This is a PR that simply adds these additional module ids to be extracted, as long as they are Vue files (which are the ones tracked in ssr context).

### 📝 Checklist

- [x] added tests


huang-julien (approved):

test/basic.test.ts
```diff
@@ -1950,6 +1952,11 @@ describe('component islands', () => {
       expect(result.head).toMatchInlineSnapshot(`
         {
           "link": [
+            {
+              "href": "/_nuxt/components/SharedComponent.vue?vue&type=style&index=0&scoped=3ee84738&lang.css",
+              "key": "island-link",
+              "rel": "stylesheet",
+            },
```
danielroe, line 1959: This is not a product of changes in the code; this is vite behaviour for adding a shared component (also on main).

danielroe:
Seems that this PR also resolves issue in https://github.com/nuxt/nuxt/pull/25394 (added regression test from that PR here as well).

packages/vite/src/utils/index.ts
```diff
@@ -11,6 +13,39 @@ export function isCSS (file: string) {
   return IS_CSS_RE.test(file)
 }
 
+export function isVue (id: string, opts: { type?: Array<'template' | 'script' | 'style'> } = {}) {
```
huang-julien, line 16: Can we use directly the util from the core ?

huang-julien: `core/utils/plugins`

danielroe: It's a different package and I'm reluctant to expose the export. But maybe we can inline it instead

huang-julien (approved):

PR #25473, opened 2024-01-28
perf(nuxt): remove empty optional route properties

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/nuxt/pull/25210
* https://github.com/nuxt-modules/i18n/issues/2738

### ❓ Type of change

### 📚 Description
Maybe I'm missing something but if we extract meta and directly set it to the routes, then I think we should be able to safely remove empty optional route properties? This issue https://github.com/nuxt-modules/i18n/issues/2738 reminded me of the [extract meta feature](https://github.com/nuxt/nuxt/pull/25210), removing the empty optional properties can reduce the entry file size quite a bit, especially so for projects using the i18n module.

I have changed the tests to use snapshots as it makes more sense in this case and allows us to see how much of a difference this makes even on a small set of routes (see https://github.com/nuxt/nuxt/commit/e08ebf0c03306ea45dfb067d52cf59643462a80e#diff-e02641b7437949321cefa187a31035cc6380d98cd5b8616311a43d02a68417a1). Another example would be this project https://github.com/WordPress/openverse/pull/3646, a lot of routes are generated due to the amount of locales, with this change the entry file size is reduced from `1456 KB` to `1063 KB`.

The tests should probably be expanded to include routes that use the `alias`, `redirect` properties. 

Also this change makes me wonder why we still have fallbacks using the meta variable for `name` and `path` if the values used should have been extracted to begin with.

### 📝 Checklist


danielroe:
We currently have the possibility of more dynamic values (eg a ref or computed - or a value set conditionally). These can't be extracted so we need to have a fall back for them.

I'd be up for removing that possibility in Nuxt v4 depending on wherever people have a valid use case for it.

BobbieGoede:
> We currently have the possibility of more dynamic values (eg a ref or computed - or a value set conditionally). These can't be extracted so we need to have a fall back for them.

Oh I don't think I have seen this before, are there examples of what this looks like? Are the `console.debug` statements in `getRouteMeta` called when this is the case? 

If we can at least reliably detect a meta key is present it should be possible to conditionally add/remove the fallback (and/or property), does that sound right?

BobbieGoede:
Okay I have changed the implementation to try and conditionally fallback or remove the properties as described. I haven't checked if this works in other projects, my code definitely needs a close look by someone else 😅..

Also, I'm not sure what the automated fixes CI is trying to do (and why it is failing).

danielroe (approved):

PR #24960, opened 2023-12-29
fix(nuxt): improve default router `scrollBehaviour`

bernhardberger:
handle first page load and page refresh scenarios (restore scroll position without jumps), fixes hash-navigation when changing pages and browser back/forward positions, especiall when page transitions are involved.

### 🔗 Linked issue

fixes #24941 
fixes #22487 
fixes #19664
fixes #25030

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

handle first page load and page refresh scenarios (restore scroll position without jumps), fixes hash-navigation when changing pages and browser back/forward positions, especially when page transitions are involved.

### 📝 Checklist

- [X] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #25483, opened 2024-01-29
fix(nuxt): prevent initial scroll jump when loading page

### LINKED ISSUE #19664: app.pageTransition Refreshing page scrolls will to the top
### Environment

mac m1

### Reproduction

- [stackblitz](https://stackblitz.com/edit/nuxt-starter-8svbbt?file=nuxt.config.ts)

On the right window, there is a refresh button on the left of the URL

### Describe the bug

- `nuxt.config.js`

If pageTransition is set. Refreshing page scrolls will to the top

```ts
  app: {
    pageTransition: { name: 'page', mode: 'out-in' },
  },
});
```

But disable pageTransition. the scroller will be at the position before refreshing

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #22487: Scroll do weird jumping to the top of the web page on Nuxt3 page refresh
### Environment

- Node Version:     v16.15.0
- Nuxt Version:     3.6.2
- Nitro Version:    2.5.2
- Package Manager:  npm@8.5.5
- Builder:          vite
- User Config:      app, components, devtools, modules, swiper, css, vite
- Runtime Modules:  @pinia/nuxt@0.4.11, nuxt-swiper@1.2.0
- Build Modules:    -

### Reproduction

You can check the effect here:
https://stackblitz.com/edit/nuxt-starter-cgquz4?file=pages%2Findex.vue

### Describe the bug

I have a problem with scroll behavior in Nuxt.js 3. When I refresh the page (when static content is served), the scroll jumps abruptly to the top of the page. However, after serving the static content, there is an automatic scroll to the previous position (the position from which the refresh was triggered), which creates an unpleasant "jumping" effect on the website.
If I use SSR = False, the scroll doesnt "jump" and stay on the same position when the page is refreshed.

I assume that Nuxt3 should prevent the scroll from returning to the top of the document when refreshing my page and keep the scroll at the same position.

When I try to do the same in Nuxt2, the scroll behaviour is normal.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #25030: Restoring scroll position takes too long
### Environment

Same as `nuxt.com` because Im using `nuxt.com` the example.

### Reproduction

https://nuxt.com

Steps to reproduce:
1. Scroll down a bit
2. Press `ctrl`+`R`
3. Page should reload and stay in same scroll position but it takes about 400-500ms to restore scroll position

### Describe the bug

Nuxt needs to wait until fully hydrated in order to restore scrolling position.
Other web frameworks are not waiting for full hydration and do it instantly on page load.

### Additional context

GIF Example:
CleanShot 2024-01-02 at 21 35 31

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19664
resolves https://github.com/nuxt/nuxt/issues/25030
resolves https://github.com/nuxt/nuxt/issues/22487

### ❓ Type of change

### 📚 Description

When performing an initial load of a page, there are _two_ navigation events that happen, first when the router plugin initialises, and then after all plugins run (to give a chance for plugins to register middleware).

This causes a jump of the scroll to the top and then after all plugins run, it restores the scroll position. However, this is unaesthetic and a bad user experience.

Rather than adjusting the scrollBehavior itself, we can set `window.history.scrollRestoration` to `auto` until these first two navigations have taken place.

https://github.com/vuejs/router/blob/main/packages/router/src/router.ts#L378-L381

This supersedes some of https://github.com/nuxt/nuxt/pull/24960 but by no means all. (cc: @bernhardberger)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/plugins/router.ts
+++ b/packages/nuxt/src/pages/runtime/plugins/router.ts
@@ -78,13 +78,25 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
           startPosition = savedPosition
           return
         }
-        // reset scroll behavior to initial value
-        router.options.scrollBehavior = routerOptions.scrollBehavior
-        return routerOptions.scrollBehavior?.(to, START_LOCATION, startPosition || savedPosition)
+        if (routerOptions.scrollBehavior) {
+          // reset scroll behavior to initial value
+          router.options.scrollBehavior = routerOptions.scrollBehavior
+          if ('scrollRestoration' in window.history) {
+            const unsub = router.beforeEach(() => {
+              unsub()
+              window.history.scrollRestoration = 'manual'
+            })
+          }
+          return routerOptions.scrollBehavior(to, START_LOCATION, startPosition || savedPosition)
+        }
       },
       history,
       routes
     })
+
+    if (import.meta.client && 'scrollRestoration' in window.history) {
+      window.history.scrollRestoration = 'auto'
+    }
     nuxtApp.vueApp.use(router)
 
     const previousRoute = shallowRef(router.currentRoute.value)


TheAlexLichter (approved):
Really good catch! And also, the stream agrees 😁

bernhardberger:
Thanks! I overlooked this in my PR.

PR #25106, opened 2024-01-08
docs: suggest using `<NuxtPage>` rather than `<RouterView>`

toto6038:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25089

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Describe the problem mentioned in https://github.com/nuxt/nuxt/issues/25089.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have added tests (if possible).
- [x] I have updated the documentation accordingly.

cc @manniL

PR #25490, opened 2024-01-29
chore(nuxt): show warning if `NuxtPage` is not used with pages enabled

### LINKED ISSUE #25089: Warn if `<RouterView>` is used instead of `<NuxtPage>`
### Describe the feature

When using Vue Routers "native" `<RouterView>` component instead of `<NuxtPage>`, it can result in route desync (see e.g. [this Discord thread](https://discord.com/channels/473401852243869706/1193483058456825976/1193546383630807060)). It would be nice to warn when using `<RouterView>` (and maybe `<RouterLink>`) instead of the idiomatic Nuxt components.

### Additional information

### Final checks

nozomuikuta:
### 🔗 Linked issue

Resolves #25089

### ❓ Type of change

### 📚 Description

This PR makes Nuxt show a warning message if pages feature is enabled but `<NuxtPage />` is never used, like it already does for `<NuxtLayout />`.

In the original issue, @manniL suggested that swap `<RouterView />` in dev mode by a proxy wrapper so that we can insert a logic to be run when `RouterView` is instantiated. 

In this PR, I took another approach that detection logic is inserted into `<NuxtPage />`, so that component tree can be same between dev and production, and that we can avoid to maintain dev-only component for external library.

Although my approach requires 2 flags (dev-only runtime config and in-component flag), I believe it's simple and reasonable enough. What do you think?

### 📝 Checklist

I tested if this works in `playground` directory.

- add `pages/index.vue` with any content
- Run `pnpm run dev` 👉 warning message is shown
- Add `<RouterView />` to `app.vue` 👉 warning message is shown
- Replace `<RouterView />` with `<NuxtPage />` 👉 warning message is not shown

Update on documentation is already handled by #25106


danielroe (approved):
Thank you ❤️ 

I moved the plugin to the pages module to remove that extra runtimeConfig flag, but otherwise 👌

robert-gruner:
@NozomuIkuta We installed v3.10.0 and now we get this warning when running the app in dev mode although we are using NuxtPage in our code normally. Is the check working correctly?

You can see it in a StackBlitz too: https://stackblitz.com/~/edit/github-wgg4gm-sdfzwd

Or are we doing something wrong? Thanks in advance for the info. ✌️

NozomuIkuta:
@robert-gruner 

Thank you for reporting and providing the reproduction!

I confirmed that this warning does not work expectedly with `setPageLayout` call.

I created a follow-up issue to track: #25617 🙋‍♂️

PR #25514, opened 2024-01-30
perf(nuxt): invalidate `getRouteMeta` cache on page contents change

BobbieGoede:
### ❓ Type of change

### 📚 Description
Instead of using file contents as key, we cache the contents separately using `absolutePath` as key. While I'm not sure if there is much performance gained by not using large strings as keys, using the contents as key will add an entry on every file change, so with the current implementation the cache will keep on growing during development as old entries won't be removed. 

(changed PR title to better communicate the change)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -135,18 +135,25 @@ export function extractScriptContent (html: string) {
 const PAGE_META_RE = /(definePageMeta\([\s\S]*?\))/
 const DYNAMIC_META_KEY = '__nuxt_dynamic_meta_key' as const
 
+const pageContentsCache: Record<string, string> = {}
 const metaCache: Record<string, Partial<Record<keyof NuxtPage, any>>> = {}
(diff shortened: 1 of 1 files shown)


danielroe (changes requested):
We definitely need to fix the ever-growing cache. But we can't cache based on the path as the path will stay the same even if the contents changes (e.g. if the user updates the meta and the site hot-reloads).

BobbieGoede:
> We definitely need to fix the ever-growing cache. But we can't cache based on the path as the path will stay the same even if the contents changes (e.g. if the user updates the meta and the site hot-reloads).

I know, this change uses a second cache to store the contents, if the contents have changed the metaCache entry is cleared. 

This is still indirectly using contents to cache metaCache, but doing so by caching contents by absolutePath separately. This way the keys aren't large, but the bigger gain is not having the cache grow on every change, only compared on every change.

danielroe:
Ahh. Great!

BobbieGoede:
I thought I had those flaky scroll to top tests figured out 🤦 now its failing in Ubuntu..

danielroe (approved):

PR #25527, opened 2024-01-31
ci: fix playwright cache

BobbieGoede:
### ❓ Type of change

### 📚 Description
I think playwright cache in the ci action has been ignored (always downloadin & installing) since replacing `playwright` with `playwright-core`. This (should) fix the cache.

Main thing I wanted to fix is that the action sometimes fails running this command:
```shell
$env:PLAYWRIGHT_VERSION="$(pnpm ls --depth 0 --json -w playwright | jq --raw-output '.[0].unsavedDependencies[\"playwright\"].version')"
```
With the following error:
```shell
jq: error: syntax error, unexpected INVALID_CHARACTER (Windows cmd shell quoting issues?) at <top-level>, line 1:
.[0].unsavedDependencies[\"playwright\"].version                         
jq: 1 compile error
```
Honestly I'm not sure why this command doesn't consistently error, and when it doesn't error it resolves to `null`, and I think that since there is no local playwright installed with version `null` it downloads and installs instead. This will likely reduce flakiness for the erroring actions.

### 📝 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,7 +229,7 @@ jobs:
         if: runner.os == 'Windows'
         run: |
           echo "PLAYWRIGHT_BROWSERS_PATH=$HOME\.cache\playwright-bin" >> $env:GITHUB_ENV
-          $env:PLAYWRIGHT_VERSION="$(pnpm ls --depth 0 --json -w playwright-core | jq --raw-output '.[0].devDependencies[\"playwright-core\"].version')"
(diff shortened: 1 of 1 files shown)


BobbieGoede:
Maybe it should still be using `unsavedDependencies` instead of `devDependencies`?

BobbieGoede:
Looks like caching is working now for the Ubuntu actions, but not quite for Windows.. Not sure why that is, at least playwright version parsing is working for Windows now as well.

danielroe (approved):

BobbieGoede:
Actually caching should work for Windows actions as well, I think the cache just still unset on the first run 😅

PR #25543, opened 2024-01-31
ci: retry flaky test when running in Windows with Webpack

BobbieGoede:
### ❓ Type of change

### 📚 Description
I was afraid this would be the case but it makes sense that this flakiness is present for both scroll to top tests, should have included it in the previous PR. This one likely got triggered less as the previous test failed more often, preventing this test from being reached?

### 📝 Checklist

### DIFF
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -619,29 +619,33 @@ describe('nuxt links', () => {
     { retry: isWebpack && isWindows ? 10 : 0 }
   )
 
-  it('expect scroll to top on nested pages', async () => {
-    // #20523
-    const page = await createPage('/nested/foo/test', {
-      viewport: {
-        width: 1000,
-        height: 1000
-      }
-    })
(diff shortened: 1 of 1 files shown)


danielroe (approved):

DamianGlowala:
@BobbieGoede, WDYT about adding this one too?

https://github.com/nuxt/nuxt/actions/runs/7733731134/job/21086407948?pr=25548#step:11:70

BobbieGoede:
@DamianGlowala 
Honestly, we might as well add retries to all known flaky tests if you ask me, it may not be pretty but we're rerunning the jobs anyway. Going to have to change the retry conditional for the test in this PR as well based on this: https://github.com/nuxt/nuxt/actions/runs/7732935327/job/21084024589?pr=25514

Maybe we should also open issues for tests showing flaky behaviour if we allow them to retry?

danielroe:
Yes, we should definitely aim to resolve flakiness rather than just 'allowing' it. Sometimes flaky tests identify genuine bugs.

PR #25568, opened 2024-02-01
fix: remove return type of promise from refresh and execute.

### LINKED ISSUE #25558: Type-definition of refresh and execute method of useAsyncData incorrect
### Environment

3.10.0 and earlier

### Reproduction

Reproduction repo: https://codesandbox.io/p/devbox/modest-gates-2fxpt3?file=%2Fapp.vue%3A7%2C25

### Describe the bug

[asyncData.ts](https://github.com/nuxt/nuxt/blob/f665434781d767d4b6d85135246a0fe17c04d00c/packages/nuxt/src/app/composables/asyncData.ts#L110) contains type-definitions of the `refresh` and `execute` functions. These definitions dictate that these functions also return the data of the asynchronous feature, but they don't.

Looking at the implementation of `useAsyncData`, it seems that the type-definitions are wrong.

### Additional context

_No response_

### Logs

_No response_

tdekoning:
### 🔗 Linked issue

#25558

### ❓ Type of change

### 📚 Description

The TypeScript definition didn't match the implementation. The promise chain of `useAsyncData` does not return a value, but updates the returned `Ref` values.

This solves the problem that developers expect these functions to return data.

Resolves #25558

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -107,8 +107,8 @@ export interface AsyncDataExecuteOptions {
 export interface _AsyncData<DataT, ErrorT> {
   data: Ref<DataT>
   pending: Ref<boolean>
-  refresh: (opts?: AsyncDataExecuteOptions) => Promise<DataT>
-  execute: (opts?: AsyncDataExecuteOptions) => Promise<DataT>
+  refresh: (opts?: AsyncDataExecuteOptions) => Promise<void>
+  execute: (opts?: AsyncDataExecuteOptions) => Promise<void>
   error: Ref<ErrorT | null>
   status: Ref<AsyncDataRequestStatus>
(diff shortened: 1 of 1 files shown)


danielroe (commented):
We could also consider returning the data.

manniL:
@danielroe Do you think that'd be beneficial or rather result in misuse of e.g. `useFetch`? 🤔

manniL:
@tdekoning could you also update the types in the docs?

tdekoning:
@manniL you mean [here](https://github.com/nuxt/nuxt/blob/main/docs/3.api/2.composables/use-async-data.md) ? They already seem to be properly typed.

manniL:
Ah, perfect! Already strange that they were out of sync then! Thanks for checking 👍

TheAlexLichter (approved):

Issue #25532, opened 2024-01-31
NuxtLink external / internal logic issues


harlan-zw:
### Environment

-----------------------------
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      app, devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-qntwss?file=package.json

### Describe the bug

:information_source: Still thinking through this issue and how to best articulate the higher level problems.

## Several edge cases

Within the `NuxtLink` component, we have several bugs related to how "external" links are treated. Within the reproduction, you can see the following issues:

1. Trailing slashes are applied to implicit external links 
2. Explicit external links do not when using vue-router object
3. Explicit external links (that are relative) that use vue-router object add base and trailing slash
4. Explicit external for relative path trailing slashes is applied
5. Using navigate() with external link does not work

The majority of these are edge cases and have workarounds. The number of edge cases, however, does point us toward the insight that the logic within NuxtLink may need some attention.

## The deeper issue

While working on adding support for [app.trailingSlash config](https://github.com/nuxt/nuxt/pull/25435) I was trying to understand how the NuxtLink component works. While digging through the code, it became clear there was some inconsistency in how we're handling logic in a couple of cases.

### External Links

We have the `external` prop that will switch the rendering between `a` and `RouterLink`.

> external: Forces the link to be considered as external (true) or internal (false). This is helpful to handle edge-cases

We also have a computed property called `isExternal` which toggles the same behavior (rendering an anchor tag instead of RouterLink), which is enabled with:
1. Has been marked explicitly as `external`, meaning the end-user wants to bypass vue-router (probably?)
2. Is implicitly external from being an absolute URL (has a protocol)
3. Has a `target` attribute and it doesn't equal `_self` ( :warning: This seems like broken logic as vue-router handles this)

```vue
<template>
  <!-- behaves correctly -->
  <NuxtLink to="/foo" external> <!-- isExternal: true -->
  <NuxtLink to="https://example.com/foo"> <!-- isExternal: true -->
  <!-- things start getting weird -->
  <NuxtLink :to="{ path: 'https://google.com/foo' }" external> <!-- isExternal: false -->
[...]

PR #25599, opened 2024-02-03
refactor(nuxt): improve `NuxtLink` types

harlan-zw:
### 🔗 Linked issue

Related to https://github.com/nuxt/nuxt/issues/25532

### ❓ Type of change

### 📚 Description

- Exposed types from vue-router are being re-implemented, we should avoid doing this for maintenance / inconsistency reasons
- Types aren't documented, copied docs from the site
- Added auto-completion for `rel` for most common privacy, security & SEO use cases
- Improved type safety for the router link props we generate

### 📝 Checklist


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -17,52 +24,90 @@ const firstNonUndefined = <T> (...args: (T | undefined)[]) => args.find(arg => a
...
+    to: string,
+    resolve: Router['resolve']
+  ): string
+  function resolveTrailingSlashBehavior (
+    to: RouteLocationRaw,
+    resolve: Router['resolve']
+  ): Omit<RouteLocationRaw, string>
+  function resolveTrailingSlashBehavior (
     to: RouteLocationRaw,
-    resolve: (to: RouteLocationRaw) => RouteLocation & { href?: string }
```
harlan-zw, line 64: Couldn't figure out why this `href` was here @manniL, maybe you know?

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -25,6 +25,23 @@ In this example, we use `<NuxtLink>` component to link to another page of the ap
...
 
+### Handling 404s
+
+When using `<NuxtLink>` for `/public` directory files or when pointing to a different app on the same domain, you should use the `external` prop.
+
+Using `external` forces the link to be rendered as an `a` tag instead of a Vue Router `RouterLink`.
+
+```vue [pages/index.vue]
+<template>
+  <NuxtLink to="/the-important-report.pdf" external>
```
harlan-zw, line 36: It's possible that we implement this logic automatically for files, not sure if Vue Router handles file links at all?

harlan-zw: We can definitely switch it to use `a` if the `download` attribute is set which I plan to do in a separate PR

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -25,6 +25,23 @@ In this example, we use `<NuxtLink>` component to link to another page of the ap
 </template>
 ```
 
+### Handling 404s
+
+When using `<NuxtLink>` for `/public` directory files or when pointing to a different app on the same domain, you should use the `external` prop.
```
harlan-zw, line 30: I couldn't think of any other valid use cases for manually using `external`, let me know if you think of anything

harlan-zw:
Not too sure why but it seems like using the `RouterLinkOptions['to']` type does not provide the `typedPages` augmentation :thinking:

docs/3.api/1.components/4.nuxt-link.md
posva: FYI vue-router doesn't differentiate any kind of link. They are all meant to be resolved with `router.resolve()`

harlan-zw: I see. Not sure the initial thinking of the original `NuxtLink` design but I'd like to not see this error being thrown for valid links.

> [Vue Router warn]: No match found for location with path "/the-important-report.pdf"

Guessing their may be some minor performance improvements by not using RouterLink for links that don't need them (?).

posva:
Other than the comment I left, I don't think my review is valuable here. But I appreciate you thought it could be 😄 . Let me know if you have other vue-router questions

harlan-zw:
> Other than the comment I left, I don't think my review is valuable here. But I appreciate you thought it could be 😄 . Let me know if you have other vue-router questions

Thanks for reviewing! Your sanity check is more than appreicated

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -314,12 +362,11 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
               if (!href) { return undefined }
 
               const url = parseURL(href)
-              return {
+              return <RouteLocation & { href: string }> {
```
danielroe, line 365: maybe we could use `satisfies` which would ensure we actually do return values compatible with the vue router types?

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -102,17 +130,18 @@ You can then use `<MyNuxtLink />` component as usual with your new defaults.
...
   exactActiveClass?: string;
   prefetchedClass?: string;
   trailingSlash?: 'append' | 'remove'
-}) => Component
+}
+function defineNuxtLink(options: NuxtLinkOptions): Component {}
 ```
 
-- `componentName`: A name for the defined `<NuxtLink>` component.
+- `componentName`: A name for the component. Default is `<NuxtLink>`.
```
danielroe, line 144:
```suggestion
- `componentName`: A name for the component. Default is `NuxtLink`.
```

danielroe (approved):

harlan-zw:
Thanks for the merge satisfies fix 👍

PR #25600, opened 2024-02-03
fix(nuxt): `NuxtLink` no default rel for relative external links

harlan-zw:
### 🔗 Linked issue

Relates to https://github.com/nuxt/nuxt/issues/25532

### ❓ Type of change

### 📚 Description

The default `rel` tag we apply to `external` NuxtLink's is quite interesting, here are some details about it if you don't know:
- We have used it because this security issue: https://mathiasbynens.github.io/rel-noopener/
- The [noreferrer](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) will automatically apply `noopener`, providing both are not required. I'd just use `noreferrer` but it's probably worth keeping consistency with larger sites in this case (i.e GitHub / Twitter still uses both).
- The [noopener](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener) has been applied by default when using `target="_blank"` in browsers since 2021, so providing all together will not be necessary at some point.  

Anyway to the point of this PR, it's to fix an edge case where it's applied when not needed and document this logic.

```vue
<template>
  <NuxtLink to="/foo" external> <!--- <a href="/foo" rel="noopener noreferrer"> -->
</template>
```

Wasn't sure whether to support relative links that open in a separate window context for extra security, interested to hear others' thoughts (i.e you're hosting an untrusted app on your domain).

```vue
<template>
  <NuxtLink to="/naughty-app" external target="_blank"> <!--- <a to="/naughty-app" rel="noopener noreferrer"> -->
</template>
```

### 📝 Checklist

### DIFF
--- a/docs/3.api/1.components/4.nuxt-link.md
+++ b/docs/3.api/1.components/4.nuxt-link.md
@@ -34,7 +34,7 @@ In this example, we use `<NuxtLink>` component to link to a website.
   <NuxtLink to="https://nuxtjs.org">
     Nuxt website
   </NuxtLink>
-  <!-- <a href="https://nuxtjs.org" rel="noopener noreferrer">...</a> -->
+  <!-- <a href="https://nuxtjs.org" rel="noreferrer">...</a> -->
 </template>
 ```
 
@@ -47,7 +47,7 @@ In this example, we use `<NuxtLink>` with `target`, `rel`, and `noRel` props.
   <NuxtLink to="https://twitter.com/nuxt_js" target="_blank">
     Nuxt Twitter
   </NuxtLink>
-  <!-- <a href="https://twitter.com/nuxt_js" target="_blank" rel="noopener noreferrer">...</a> -->
+  <!-- <a href="https://twitter.com/nuxt_js" target="_blank" rel="noreferrer">...</a> -->
 
   <NuxtLink to="https://discord.nuxtjs.org" target="_blank" rel="noopener">
     Nuxt Discord
@@ -62,7 +62,7 @@ In this example, we use `<NuxtLink>` with `target`, `rel`, and `noRel` props.
   <NuxtLink to="/contact" target="_blank">
     Contact page opens in another tab
   </NuxtLink>
-  <!-- <a href="/contact" target="_blank" rel="noopener noreferrer">...</a> -->
+  <!-- <a href="/contact" target="_blank" rel="noreferrer">...</a> -->
 </template>
 ```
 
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -294,10 +295,16 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
         checkPropConflicts(props, 'noRel', 'rel')
-        const rel = (props.noRel)
-          ? null
+        const rel = firstNonUndefined<string | null>(
           // converts `""` to `null` to prevent the attribute from being added as empty (`rel=""`)
-          : firstNonUndefined<string | null>(props.rel, options.externalRelAttribute, href ? DEFAULT_EXTERNAL_REL_ATTRIBUTE : '') || null
+          props.noRel ? '' : undefined, props.rel,
+          options.externalRelAttribute,
+          /*
+          * A fallback rel of `noreferrer` is applied for external links or links that open in a new tab.
```
danielroe, line 303:
```suggestion
          * A fallback rel of `noopener noreferrer` is applied for external links or links that open in a new tab.
```

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -34,13 +34,19 @@ In this example, we use `<NuxtLink>` component to link to a website.
   <NuxtLink to="https://nuxtjs.org">
     Nuxt website
   </NuxtLink>
-  <!-- <a href="https://nuxtjs.org" rel="noopener noreferrer">...</a> -->
+  <!-- <a href="https://nuxtjs.org" rel="noreferrer">...</a> -->
```
danielroe, line 37:
```suggestion
  <!-- <a href="https://nuxtjs.org" rel="noopener noreferrer">...</a> -->
```

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -34,13 +34,19 @@ In this example, we use `<NuxtLink>` component to link to a website.
...
 </template>
 ```
 
-## `target` and `rel` Attributes
+## Best Practice Rel
```
danielroe, line 41:
```suggestion
## `target` and `rel` Attributes
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -294,10 +295,16 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
         // Resolves `rel`
         checkPropConflicts(props, 'noRel', 'rel')
-        const rel = (props.noRel)
-          ? null
+        const rel = firstNonUndefined<string | null>(
           // converts `""` to `null` to prevent the attribute from being added as empty (`rel=""`)
-          : firstNonUndefined<string | null>(props.rel, options.externalRelAttribute, href ? DEFAULT_EXTERNAL_REL_ATTRIBUTE : '') || null
+          props.noRel ? '' : undefined, props.rel,
```
danielroe, line 300:
```suggestion
          props.noRel ? '' : props.rel,
```

danielroe (approved):

PR #25616, opened 2024-02-04
fix(nuxt): warn if `inheritAttrs: false` is set when using `useId`

### LINKED ISSUE #25605: useId bug when inheritAttrs: false
### Environment

- Operating System: Darwin
- Node Version:     v20.7.0
- Nuxt Version:     3.10.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1

### Reproduction

https://stackblitz.com/~/edit/nuxt-starter-1b6v9b

### Describe the bug

when we add `inheritAttrs: false` in the component, `data-n-ids` not added to element, that caused hydration error

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25605

### ❓ Type of change

### 📚 Description

This adds a warning when `inheritAttrs: false` is set in a component where we are using `useId`. This is not compatible with our current implementation.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/id.ts
+++ b/packages/nuxt/src/app/composables/id.ts
@@ -26,7 +26,7 @@ export function useId (key?: string): string {
   const instanceIndex = key + ':' + instance._nuxtIdIndex[key]++
 
   if (import.meta.server) {
-    if (!import.meta.dev && instance.vnode.type && typeof instance.vnode.type === 'object' && 'inheritAttrs' in instance.vnode.type && instance.vnode.type.inheritAttrs === false) {
+    if (import.meta.dev && instance.vnode.type && typeof instance.vnode.type === 'object' && 'inheritAttrs' in instance.vnode.type && instance.vnode.type.inheritAttrs === false) {
       console.warn('[nuxt] `useId` is not compatible with components that have `inheritAttrs: false`.')
     }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/id.ts
```diff
@@ -26,6 +26,9 @@ export function useId (key?: string): string {
   const instanceIndex = key + ':' + instance._nuxtIdIndex[key]++
 
   if (import.meta.server) {
+    if (!import.meta.dev && instance.vnode.type && typeof instance.vnode.type === 'object' && 'inheritAttrs' in instance.vnode.type && instance.vnode.type.inheritAttrs === false) {
```
huang-julien, line 29: we don't warn in dev ?

manniL: I think we should consider warning only in dev, no? 🤔

danielroe: yes! 🤦

danielroe:
```suggestion
    if (import.meta.dev && instance.vnode.type && typeof instance.vnode.type === 'object' && 'inheritAttrs' in instance.vnode.type && instance.vnode.type.inheritAttrs === false) {
```

TheAlexLichter (approved):

PR #25679, opened 2024-02-07
docs: enable twoslash for some code snippets

antfu:
This enables TwoSlash support for nuxt.com in https://github.com/nuxt/nuxt.com/pull/1486

Here is an example: 
Screenshot 2024-02-07 at 16 43 52

I manually checked every code snippet I enabled. Since the docs and nuxt.com are different repos, currently, you have to manually link them to verify the works (TwoSlash also verifies the types and throws when there are type errors, this could be handly to typecheck the docs later on). I am working with Orta to bring `twoslash-cli` so that later, we can verify the code snippets in CI to guard against type errors.

This PR is pure additional, should not break existing usages and is good to merge.


danielroe (approved):

danielroe:
This makes me very happy.

PR #25727, opened 2024-02-09
fix(nuxt): respect baseUrl for client components within server components

### LINKED ISSUE #25676: Client component within server component not working, when app.baseUrl is set
### Environment

- Operating System: `Linux`
- Node Version:     `v21.5.0`
- Nuxt Version:     `3.10.1`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `yarn@1.22.21`
- Builder:          `-`
- User Config:      `devtools`, `app`, `experimental`
- Runtime Modules:  `-`
- Build Modules:    `-`

flags:
`{  experimental: { componentIslands: { selectiveClient: true } }`

### Reproduction

https://stackblitz.com/~/github.com/marccremer/nuxt_islands_issue

### Describe the bug

When you rerender a server component it doesnt work on the inner client components.
In the example you change the props by clicking on the button.
You expect "heyfalse" to turn into "heytrue".
Nothing happens

### Additional context

The given example works when you remove the baseUrl.
It seems that on rererend er the dynmic import doesnt take the baseUrl.
It should be `import("http://localhost:3000"+baseUrl+"component.vue")`
But it does instead this
`import("http://localhost:3000"+"component.vue")`
And doesnt find the import.

### Logs

```shell-script
Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: http://localhost:3000/_nuxt/@fs/home/redacted/nuxt_island_issue/components/ClientShow.vue?import
```

marccremer:
### 🔗 Linked issue

resolves #25676 

### ❓ Type of change

### 📚 Description
Client components in server component use dynamic imports that are defined in nuxt-island.
The problematic function didn't take the app.baseURL config into account.

### 📝 Checklist


huang-julien (approved):
okay for me thank you ! :heart:

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -35,7 +35,7 @@ async function loadComponents (source = '/', paths: NuxtIslandResponse['componen
   for (const component in paths) {
     if (!(components!.has(component))) {
       promises.push((async () => {
-        const chunkSource = join(source, paths[component].chunk)
+        const chunkSource = join(source,appBaseURL, paths[component].chunk)
```
danielroe, line 38:
```suggestion
        const chunkSource = join(source, appBaseURL, paths[component].chunk)
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -35,7 +35,7 @@ async function loadComponents (source = '/', paths: NuxtIslandResponse['componen
   for (const component in paths) {
     if (!(components!.has(component))) {
       promises.push((async () => {
-        const chunkSource = join(source, paths[component].chunk)
+        const chunkSource = join(source, appBaseURL, paths[component].chunk)
```
danielroe, line 38: shouldn't this instead be that `loadComponents` is _passed_ `appBaseURL` rather than it joins it to the source?

I imagine a remote source might not have the same baseURL

huang-julien: oh indeed !

marccremer: Should it look maybe like this 
```js
let chunkSource
if(isRemoteSource){
      chunkSource = join(source, paths[component].chunk)
 }else{
     chunkSource = join(source, appBaseURL, paths[component].chunk)
}     
```
I assume that props.source would contain the remote url like `https://my.cdn.com/nuxt`
The signature would be the same and it makes sense that we dont need the baseUrl when source starts with "http"

huang-julien: probably something like 
making source undefined by default in the signature then
```ts
if (!source) {
  source = appBaseURL ?? '/'
}
```

?

danielroe (approved):

danielroe (approved):

PR #22157, opened 2023-07-15
docs: standardize indentation and tag positioning

shahabbasian:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

I believe that newcomers tend to internalize the code patterns present in the documentation as they read it. This will indeed help them write clean code by following established patterns.

- [x] The <template> tag has been added to the code snippets in the "Class And Style Bindings" section of the documentation.
- [x] Standardize indentation and tag positioning across all pages.
- [x] All <script setup> changed to <script setup lang="ts">

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #25750, opened 2024-02-12
docs: <script setup> changed to <script setup lang="ts">

shahabbasian:
### ❓ Type of change

### 📚 Description

due to the:
https://github.com/nuxt/nuxt/pull/22157

lang="ts" By default, it was placed in the content of the document

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Simple, intuitive and powerful, Nuxt lets you write Vue components in a way that
 Example of an `app.vue`:
 
 ```vue
-<script setup>
+<script setup lang="ts">
 useSeoMeta({
   title: 'Meet Nuxt',
   description: 'The Intuitive Vue Framework.'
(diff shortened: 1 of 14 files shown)


TheAlexLichter (changes requested):
Thanks for the contribution! I'd only change it in case we use TS in there though so people could copy and paste it without using TS as well ☺️

shahabbasian:
> Thanks for the contribution! I'd only change it in case we use TS in there though so people could copy and paste it without using TS as well ☺️

Please see the link below:
https://github.com/nuxt/nuxt/pull/22157

danielroe:
Without `lang="ts"` a lot of the magic just won't work. I think it's safe to use even if people aren't writing actual TypeScript. What do you think @manniL?

manniL:
Fair point. Let's stick with it then 👍

TheAlexLichter (approved):

Issue #25747, opened 2024-02-12
`instanceof` operator not working properly in prod. build when using Nuxt Layers


cernymatej:
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.9.0`
- Nuxt Version:     `3.10.1`
- CLI Version:      `3.10.0`
- Nitro Version:    `2.8.1`
- Package Manager:  `pnpm@8.15.1`
- Builder:          `-`
- User Config:      `devtools`, `extends`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-gcietl?file=child%2Fapp.vue,parent%2Futils%2FisInstanceOfZodObject.ts,child%2Fnuxt.config.ts

### Describe the bug

In a production build where the code is transformed using esbuild, the behavior of the `instanceof` operator can break in certain situations. This issue is likely related to how dependencies are managed in projects using Nuxt Layers.

For instance, when a variable is checked with `instanceof` ZodObject, and `'zod'` is installed in both the parent layer and the child layer, one of them will have a suffix added (`ZodObject` becomes `ZodObject$1`). If a utility function `isInstanceOfZodObject(any)` from the parent layer is used in the child layer, it will return `false` in the production build, unlike in the development mode, where the code is not transformed.

This issue is not an esbuild bug, as discussed in [this esbuild issue](https://github.com/evanw/esbuild/issues/3333). Although it's not necessarily a "bug" in Nuxt either, I believe it should be addressed in some way because it can "silently" break important logic. Any suggestions?

### Steps to Reproduce
1. Open the reproduction
2. `cd` into the `parent/` directory and run `npm i`
3. `cd` into the `child/` directory and run `npm i` and then `npm run dev` (**both checks return `true` in development mode**)
4. Stop the dev server, then run `npm run build`
5. Run `npm run preview` (**in the production build, the check performed by a function in the parent layer returns `false`, which is "incorrect"**)

danielroe:
I think this is no longer reproducible (on `main`) as nitro v3 bundles everything.

PR #25752, opened 2024-02-12
fix(vite): optimise layer dependencies with vite

### LINKED ISSUE #19426: Can't use dependencies in nuxt layers
### Environment

- Node Version:     `v16.19.0`
- Nuxt Version:     `^3.2.3`
- Nitro Version:    `-`
- Builder:          `webpack`
- User Config:      `ssr`, `extends`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

[dayjs-in-nuxt-layer-reporoduction.zip](https://github.com/nuxt/nuxt/files/10882990/dayjs-in-nuxt-layer-reporoduction.zip)
In `./nuxt-project`:
```bash
yarn && yarn dev
```
image

### Describe the bug

Can't use dependencies in package layer. So far I've noticed that the problem occurs with packages like: `axios`, `humps`, `qs`, `dayjs`, but I suspect there are many more.

### Additional context

Workaround (in `nuxt.config.ts`):
```ts
	// ...
	extends: [
		'<layer-name>'
	],
	alias: {
		axios: 'axios',
		humps: 'humps',
		qs: 'qs',
		dayjs: 'dayjs',
	},
});
```

### Logs

```shell-script
Uncaught SyntaxError: The requested module '/_nuxt/node_modules/dayjs/dayjs.min.js?v=17986f0a' does not provide an export named 'default' (at useDayjs.ts?v=17986f0a:1:8)
```

### LINKED ISSUE #21714: Ambiguous Indirect Export error when using elastic-builder in nuxt layer
### Environment

- Operating System: Linux
- Node Version:     v16.14.2                                                                                                                             12:48:46
- Nuxt Version:     3.5.3
- Nitro Version:    2.5.1
- Package Manager:  yarn@1.22.5
- Builder:          vite
- User Config:      extends, vite
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/thibaultrey/nuxt-extends-import-bug

### Describe the bug

If you import [elastic-builder](https://www.npmjs.com/package/elastic-builder) package in a layer extended as a NPM package, you get the following error when load the page :
`
Uncaught SyntaxError: ambiguous indirect export: default
`

This error doesn't appear when you using the same layer locally via a relative path.

The error seems similar to [#19267](https://github.com/nuxt/nuxt/issues/19267) but added elastic-builder package in the build.transpile settings doesn't works.

In the relative path extend way vite seems to _optimize_ the elastic-builder; this action never happens when you extends throught a NPM Package.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/19426
resolves https://github.com/nuxt/nuxt/issues/21714

### ❓ Type of change

### 📚 Description

This PR triggers a resolve call when importing dependencies into layers, to trigger vite to optimise any dependencies that are imported within a layer, just as if they were imported by a user...

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -138,6 +138,19 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
     ctx.config.build!.watch = undefined
   }
 
+  // Trigger vite to optimize any dependencies that are imported within a layer, just as if they were imported by a user
+  if (nuxt.options.dev) {
+    ctx.config.plugins!.push({
+      name: 'nuxt:optimize-layer-deps',
+      enforce: 'pre',
+      async resolveId (source, importer) {
+        if (importer && nuxt.options._layers.some(layer => layer.cwd !== nuxt.options.rootDir && importer.includes(layer.cwd))) {
+          await this.resolve(source, join(nuxt.options.srcDir, 'index.html'), { skipSelf: true }).catch(() => null)
+        }
+      },
+    })
+  }
+
   // Add type-checking
   if (!ctx.nuxt.options.test && (ctx.nuxt.options.typescript.typeCheck === true || (ctx.nuxt.options.typescript.typeCheck === 'build' && !ctx.nuxt.options.dev))) {
     const checker = await import('vite-plugin-checker').then(r => r.default)
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -138,27 +138,29 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
     ctx.config.build!.watch = undefined
   }
 
-  // Identify which layers will need to have an extra resolve step.
-  const layerDirs: string[] = []
-  const delimitedRootDir = nuxt.options.rootDir + '/'
-  for (const layer of nuxt.options._layers) {
-    if (layer.config.srcDir !== nuxt.options.srcDir && !layer.config.srcDir.startsWith(delimitedRootDir)) {
-      layerDirs.push(layer.config.srcDir + '/')
+  if (nuxt.options.dev) {
+    // Identify which layers will need to have an extra resolve step.
+    const layerDirs: string[] = []
+    const delimitedRootDir = nuxt.options.rootDir + '/'
+    for (const layer of nuxt.options._layers) {
+      if (layer.config.srcDir !== nuxt.options.srcDir && !layer.config.srcDir.startsWith(delimitedRootDir)) {
+        layerDirs.push(layer.config.srcDir + '/')
+      }
     }
-  }
-  if (layerDirs.length > 0) {
-    ctx.config.plugins!.push({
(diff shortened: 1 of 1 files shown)


packages/vite/src/vite.ts
```diff
@@ -138,13 +138,23 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
     ctx.config.build!.watch = undefined
   }
 
-  // Trigger vite to optimize any dependencies that are imported within a layer, just as if they were imported by a user
-  if (nuxt.options.dev) {
```
manniL: Not only in dev anymore?

danielroe: ah. yes. probably best to keep in dev for now until tested better. we might later expand out of dev to try and resolve dependencies from final layer (would resolve https://github.com/nuxt/nuxt/issues/25747, for example)...

darvids0n:
Hi @danielroe, I believe this change is the direct cause of some serious build performance issues e.g. #26211, #27106

Experiment for any users who have a slow dev server startup: Comment out this code chunk from
`@nuxt/vite-builder/dist/shared/vite-builder.CHUNK_ID.mjs`

```ts
if (nuxt.options.dev) {
    const layerDirs = [];
    const delimitedRootDir = nuxt.options.rootDir + "/";
    for (const layer of nuxt.options._layers) {
      if (layer.config.srcDir !== nuxt.options.srcDir && !layer.config.srcDir.startsWith(delimitedRootDir)) {
        layerDirs.push(layer.config.srcDir + "/");
      }
    }
    if (layerDirs.length > 0) {
      ctx.config.plugins.push({
        name: "nuxt:optimize-layer-deps",
        enforce: "pre",
        async resolveId(source, _importer) {
          if (!_importer) {
            return;
          }
          const importer = normalize(_importer);
          if (layerDirs.some((dir) => importer.startsWith(dir))) {
            await this.resolve(source, join(nuxt.options.srcDir, "index.html"), { skipSelf: true }).catch(() => null);
          }
        }
      });
    }
  }
```

Result for me: Build performance significantly improves from ~40-80s, down to ~2-3s

Something about these resolves is very inefficient it seems.

danielroe:
@darvids0n Thanks for digging into this. There is definitely space to optimise this. 

Do you have a reproduction I can use to test?

darvids0n:
> @darvids0n Thanks for digging into this. There is definitely space to optimise this.
> 
> Do you have a reproduction I can use to test?

I just made one on Stackblitz for you. Link here: https://stackblitz.com/edit/nuxt-starter-ujzgtq

All I've done is take the nuxt starter, and pull in 3 layers which provide some ready-to-go components - `oxecore`, `nuxt-lego` and `supauth`. I added them to `extends`. The `npm run dev` performance is roughly 13 seconds in before Nitro completed build, and 20 seconds for Nuxt ssr+client.

I added the exact instructions for commenting out the layer deps chunk in README.

PR #25765, opened 2024-02-13
fix(nuxt): don't add an extra `baseURL` on server `useRequestURL`

### LINKED ISSUE #25763: useRequestURL() returns double baseUrl on the server side
### Environment

- Node Version:     v20.11.0
- Nuxt Version:     3.10.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@10.3.0
- Builder:          -
- User Config:      app
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/

### Describe the bug

I am trying to use  useRequestURL() composable, but it returns different url on the server and client if the baseUrl is set in nuxt config `app: { baseURL: "/my-app/" }`. On the server, it returns url with path  `/my-app/my-app/`, on the client, it returns the correct url with path `/my-app/`.
I found source file: [`nuxt/packages/nuxt/src/app/composables/url.ts`](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/url.ts):
```ts
import { getRequestURL } from 'h3'
import { joinURL } from 'ufo'
import { useRuntimeConfig } from '../nuxt'
import { useRequestEvent } from './ssr'

/** @since 3.5.0 */
export function useRequestURL () {
  if (import.meta.server) {
    const url = getRequestURL(useRequestEvent()!)
    url.pathname = joinURL(useRuntimeConfig().app.baseURL, url.pathname)
    return url
  }
  return new URL(window.location.href)
}
```
There may be a problem in this line:
`url.pathname = joinURL(useRuntimeConfig().app.baseURL, url.pathname)`

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25763

### ❓ Type of change

### 📚 Description

I suspect this was triggered by an upstream change after `useRequestURL` was introduced, but currently `event.path` returns full URL so we do not have to add `baseURL` if app is running with a custom baseURL.

cc: @pi0 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/url.ts
+++ b/packages/nuxt/src/app/composables/url.ts
@@ -1,14 +1,10 @@
 import { getRequestURL } from 'h3'
-import { joinURL } from 'ufo'
-import { useRuntimeConfig } from '../nuxt'
 import { useRequestEvent } from './ssr'
 
 /** @since 3.5.0 */
 export function useRequestURL () {
   if (import.meta.server) {
-    const url = getRequestURL(useRequestEvent()!)
-    url.pathname = joinURL(useRuntimeConfig().app.baseURL, url.pathname)
-    return url
+    return getRequestURL(useRequestEvent()!)
   }
   return new URL(window.location.href)
 }
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -593,7 +593,7 @@ describe('nuxt links', () => {
     await page.close()
   })
 
-  it('expect scroll to top on routes with same component', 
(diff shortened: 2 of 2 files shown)


pi0 (approved):
I'm not sure anything related to these changes in upstream (maybe Nuxt SSR internals?) 

BTW seems a good change to me 👍🏼  (to also not diverge util behavior specially it has same name!)

PS: We have `withBase` that can ensure double base won't happen in similar cases.

PR #25766, opened 2024-02-13
fix(schema): use `rootDir` rather than `process.cwd` for `modulesDir`

### LINKED ISSUE #25408: [Documentation] Nuxt Configuration isn't entirely correct leading to unexpected behaviour
### Environment

N/A

### Reproduction

N/A

### Describe the bug

https://nuxt.com/docs/api/nuxt-config states the page is generated from source, however it's over simplifying it.

Examples:
[rootDir](https://nuxt.com/docs/api/nuxt-config#rootdir)
Default: `/<rootDir>`

It references itself, which can't be possible. If you look at the [source](https://github.com/nuxt/nuxt/blob/a42aaecfb5eb44966c67a925b9ee4085bea8329f/packages/schema/src/config/common.ts#L41) you can see the default is using the cwd (current working directory). That's a very different default than documented.

Another example:
[serverDir](https://nuxt.com/docs/api/nuxt-config#serverdir)
Default: `/<rootDir>/server`

The default is entirely incorrect, the default isn't `/<rootDir>/server`, but rather `/<srcDir>/server`, [source](https://github.com/nuxt/nuxt/blob/a42aaecfb5eb44966c67a925b9ee4085bea8329f/packages/schema/src/config/common.ts#L102)

There are numerous inconsistencies in the documentation, so the whole function that is generating this page (and likely others) should be revisited to fix the incorrect defaults.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25408

### ❓ Type of change

### 📚 Description

This addresses a number of schema issues.

1. It adds a minor perf improvement by deduplicating `fs.allow` and `modulesDir` arrays.
2. It adds `<srcDir>` and `<workspaceDir>` aliases to clean up generated documentation
3. It does not add `process.cwd()` to modulesDir but instead rootDir (which defaults to `process.cwd()` if no value is provided). This is the most significant change and means that Nuxt's behaviour will change when running `nuxi some/directory`. It will no longer pull in the `node_modules` from the directory _where_ you are running nuxi but will instead pull in the `node_modules` from the Nuxt project you are running, which seems to be the better behaviour.

I wonder if the 

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -137,8 +137,8 @@ export default defineUntypedSchema({
     $resolve: async (val: string[] | undefined, get): Promise<string[]> => {
       const rootDir = await get('rootDir') as string
       return [
-        ...await Promise.all((val || []).map(async (dir: string) => resolve(rootDir, dir))),
-        resolve(process.cwd(), 'node_modules')
+        ...(val || []).map((dir: string) => resolve(rootDir, dir)),
+        resolve(rootDir, 'node_modules')
       ]
     }
   },
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -136,10 +136,10 @@ export default defineUntypedSchema({
     $default: ['node_modules'],
     $resolve: async (val: string[] | undefined, get): Promise<string[]> => {
       const rootDir = await get('rootDir') as string
-      return [
+      return [...new Set([
         ...(val || []).map((dir: string) => resolve(rootDir, dir)),
         resolve(rootDir, 'node_modules')
-      ]
+      ])]
     }
   },
 
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -92,14 +92,14 @@ export default defineUntypedSchema({
         allow: {
           $resolve: async (val: string[] | undefined, get) => {
             const [buildDir, srcDir, rootDir, workspaceDir, modulesDir] = await Promise.all([get('buildDir'), get('srcDir'), get('rootDir'), get('workspaceDir'), get('modulesDir')]) as [string, string, string, string, string]
-            return [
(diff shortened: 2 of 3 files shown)


packages/schema/src/config/vite.ts
```diff
@@ -92,14 +92,14 @@ export default defineUntypedSchema({
         allow: {
           $resolve: async (val: string[] | undefined, get) => {
             const [buildDir, srcDir, rootDir, workspaceDir, modulesDir] = await Promise.all([get('buildDir'), get('srcDir'), get('rootDir'), get('workspaceDir'), get('modulesDir')]) as [string, string, string, string, string]
-            return [
+            return [...new Set([
```
pi0, line 95: Just curious was it making a bug or just is for docs?

packages/schema/src/config/common.ts
```diff
@@ -136,10 +136,10 @@ export default defineUntypedSchema({
     $default: ['node_modules'],
     $resolve: async (val: string[] | undefined, get): Promise<string[]> => {
       const rootDir = await get('rootDir') as string
-      return [
-        ...await Promise.all((val || []).map(async (dir: string) => resolve(rootDir, dir))),
-        resolve(process.cwd(), 'node_modules')
-      ]
+      return [...new Set([
+        ...(val || []).map((dir: string) => resolve(rootDir, dir)),
+        resolve(rootDir, 'node_modules')
```
pi0, line 141: Nice find!

pi0 (approved):

packages/schema/src/config/vite.ts
danielroe: For the docs mostly but I did see a related vite issue where adding extra entries to an array increased startup time disproportionately so thought it was worth resolving.

PR #25647, opened 2024-02-05
perf: don't manipulate an empty value

GalacticHypernova:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

We should not iterate and make O(n) operations on an empty values (strings, arrays), just to receive the same empty value back.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

PR #25771, opened 2024-02-13
perf: performance overhaul

GalacticHypernova:
## NOTE:
This is the previous pr, new one is at #26114 .
This one had some unexplained issues with certain commits (for example, https://github.com/nuxt/nuxt/pull/25771#issuecomment-1974822436) and overall had some edge cases, so I have decided to stop commiting on this one and opening the new one to take care of all the issues more carefully.

### ❓ Type of change

### 📚 Description

This is a PR that aims to improve both build time and runtime performance to the max by refactoring everything that comes to mind with performance best practices, some of which include smarter and fewer object/array iterations, reduced indexes, smarter variable usage for constant values, and more. 

Some key improvements in this PR:
- Refactors of known costly javascript methods:
   In some places throughout the app, there are operations that rely on certain javascript methods that are known to be more expensive at certain conditions than manual implementation 
   (similar to `endsWith` in  #24746 
    and `startsWith` in  #24744). 
    One key change under this category is the usage of  `Array.prototype.includes`.
    `Array.prototype.includes` appears to be slower when checking for existence of a certain input between at the very least 2 elements than directly checking each element with strict equality (like when dealing with strings).
    After thorough benchmarking, both browser and node, direct comparisons showed marginally better performance, benchmarks attached below in that same order with the following benchmark:
    ```ts
    const a = "#text"
    console.time("start")
    for(let i = 0; i< 1000000; i++) {
      const inc = ['#comment', '#text'].includes(a)
    }
    console.timeEnd("start")
    console.time("stop")
    for(let i = 0; i< 1000000; i++) {
      const inc  = a === "#comment" || a === "#text"
    }
    console.timeEnd("stop")
    ```
    Browser:
    
    image
    
    Node:
    
    image

- Reduced and reworked iterations
   Originally the main change in this PR, many parts in nuxt involve some sort of array/object manipulations, and in some places they are either repeated, redundant, or simply costly in terms of performance. The main examples of such cases are the small side PR's I have submitted like #24717, #24730,  and numerous others. I couldn't get all of it at the same time so I have decided to group the rest here, and also handle the iteration logic as opposed to mere count.
   
 - Reduced indexes:
    Some places in nuxt include working with indexes, namely the `length` property of array/strings. While that alone isn't expensive, it still makes that repeated index, which especially with bigger inputs (like in HTML parsing or AST traversal) will gradually cost more and more microseconds, and might eventually become slightly noticeable in page performance.
    
- Early returns
   Similar to #25647, there are some places where some iterations are made on potentially empty values, which increase the overhead and therefore decrease performance. This PR adds guards to ensure there won't be iterations over an empty value.
### 📝 Checklist


GalacticHypernova:
I'll fix the errors when I get the chance,  made it a draft on purpose

packages/nuxt/src/components/client-fallback-auto-id.ts
```diff
@@ -19,10 +19,10 @@ export const clientFallbackAutoIdPlugin = createUnplugin((options: LoaderOptions
     name: 'nuxt:client-fallback-auto-id',
     enforce: 'pre',
     transformInclude (id) {
-      if (exclude.some(pattern => pattern.test(id))) {
+      if (exclude.length && exclude.some(pattern => pattern.test(id))) {
```
GalacticHypernova, line 22: Note to self: worth investigating a manual implementation of `some`, as local node benchmarks showed noticeable improvement with a custom `for of` loop and a break statement, which leads me to believe `Array.prototype.some` is O(n) as opposed to worst-case O(n).

GalacticHypernova: Still needs to be checked how many elements need to be in an array for it to be beneficial, but looks like the number is unlimited.

packages/nuxt/src/core/templates.ts
```diff
@@ -142,11 +158,14 @@ const adHocModules = ['router', 'pages', 'imports', 'meta', 'components', 'nuxt-
...
+    for (const m of nuxt.options._installedModules) {
+      const meta = {...m.meta}
+      const impName = m.entryPath || meta?.name
+      if (meta.configKey && meta.name && !adHocModules.includes(meta.name)) {
+         moduleInfo.push({...meta, importName: impName})
+      }
+    }
     const relativeRoot = relative(resolve(nuxt.options.buildDir, 'types'), nuxt.options.rootDir)
     const getImportName = (name: string) => (name[0] === '.' ? './' + join(relativeRoot, name) : name).replace(/\.\w+$/, '')
     const modules = moduleInfo.map(meta => [genString(meta.configKey), getImportName(meta.importName)])
```
GalacticHypernova, line 171: TODO: revisit this, as the type checks fail but the regular tests don't, which might indicate a false negative.

GalacticHypernova (commented):
Notes to self

packages/nuxt/src/core/templates.ts
GalacticHypernova: Update: fixed, missed a line in the previous refactors

GalacticHypernova:
Still need to figure out why the dev builds fail...

GalacticHypernova:
Update: I decided to rework this more carefully, I will open a new PR for it and close this one.

PR #25777, opened 2024-02-13
ci: typecheck code samples in docs

antfu:
### ❓ Type of change

### 📚 Description

This PR adds typecheck for TwoSlash blocks in our docs Markdown. Follow up on #25679 to make it easier to work with without linking to nuxt.com repo. This would help us provide code examples that are type-safe in the future.

### 📝 Checklist


.nuxtrc
```diff
@@ -1 +1,2 @@
 telemetry.enabled=false
+pages=true
```
danielroe, line 2: this is enabling `pages` also for the vitest runtime tests we have - maybe instead run `nuxt-content-twoslash` from within `test/fixtures/basic-types` or similar?

antfu: I originally tried to use `test/fixtures/basic-types` but the problem is that the types are generated with different paths, that are relative to the project root (`../../../../node_modules/*`) - which would need a heavy rewrite to every dts file to get them working in twoslash, so I thought it might be better to use the types from root dir to also be more consistent with the final users' env.

danielroe (approved):

.github/workflows/docs.yml
```diff
@@ -32,3 +32,9 @@ jobs:
 
       - name: Lint (docs)
         run: pnpm lint:docs
+
+      - name: Build Nuxt
+        run: pnpm build
+
+      - name: Typecheck (docs)
+        run: pnpm typecheck:docs
```
danielroe, line 40: Maybe there is an opportunity here to move docs workflow back into main ci.yml but use `tj-actions/changed-files` instead to granularly enable workflows.

PR #25809, opened 2024-02-15
feat(kit): allow module configuration from c12

ineshbose:
### ❓ Type of change

### 📚 Description

Long pending PR about wanting to split Nuxt modules configuration to keep a cleaner `nuxt.config`, and now with support for `/.config` dir 😄 

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/install.ts
+++ b/packages/kit/src/module/install.ts
@@ -2,6 +2,7 @@ import { existsSync, promises as fsp, lstatSync } from 'node:fs'
 import type { ModuleMeta, Nuxt, NuxtModule } from '@nuxt/schema'
 import { dirname, isAbsolute, join, resolve } from 'pathe'
 import { defu } from 'defu'
+import { loadConfig } from 'c12'
 import { isNuxt2 } from '../compatibility'
 import { useNuxt } from '../context'
 import { requireModule } from '../internal/cjs'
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for this PR but i think generally it is not the best idea to push it this way:

1. It is too specific for tailwind module that requires an external configuration file for IDE support. By default, all nuxt config goes into `nuxt.config` interface that is extendable too
2. For a convention to split nuxt.config itself into a nested structure, would be nice but probably we should think of it directly in c12 so that it can handle it consistently

ineshbose:
> It is too specific for tailwind module that requires an external configuration file for IDE support.

Not thinking specifically for Tailwind module in this scenario, but even so, the module will take options from `tailwindcss.nuxt.config` while Tailwind CSS takes from `tailwind.config`; the module options for `tailwindcss.nuxt.config` are a bit different from `tailwind.config` (its a superset with `config: {}` property).

> By default, all nuxt config goes into nuxt.config interface that is extendable too

In those cases, we'd encourage to use JSDoc possibly like `/* @type {import('nuxt/schema').NuxtConfig['configKey']} */` or ideally we want to provide a `defineModuleConfig('configKey', {})` utility.

> For a convention to split nuxt.config itself into a nested structure, would be nice but probably we should think of it directly in c12 so that it can handle it consistently.

Modules aren't a c12 concept though, right? Or if I am to understand correctly (taking https://nuxt.com/docs/api/nuxt-config#baseurl as a reference) you want to provide support to have `nuxt.config`, `app.nuxt.config`, `baseUrl.app.nuxt.config` (something like that, a bit like how the tailwind module splits the config into subpath files)?

pi0:
I think that c12 can support `.config/nuxt/[key].[ex]` when loading the configuration. It can basically give same effect as this PR while not limiting it to modules not requiring additional c12 loads.

pi0:
Made https://github.com/unjs/c12/issues/137 to followup

In the meantime, if the main case we are thinking is tailwind module, I think it would be really really better if the module itself directly handles it. (long configs are in most of the times a no-no for Nuxt. Nuxt is designed to be minimal and I like to know what other cases you might think can benefit and often require a long configuration that doesn't logically fits into main nuxt config)

ineshbose:
Thanks!

Yes I am planning to add c12 to the tailwind module itself either anyway. 🙂 

Closing this for now..

PR #25817, opened 2024-02-15
fix(nuxt): scroll to top on page navigation when scroll-behavior: smo…

### LINKED ISSUE #19239: Nuxt Link does not scroll to top on new page...
### Environment

- Operating System: `Darwin`
- Node Version:     `v19.1.0`
- Nuxt Version:     `3.1.1`
- Nitro Version:    `2.1.0`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `pages`, `modules`, `plugins`, `build`, `googleFonts`, `i18n`, `strapi`, `vite`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.3.0`, `@nuxtjs/google-fonts@3.0.0-1`, `@nuxtjs/strapi@1.7.1`, `@nuxtjs/i18n@8.0.0-beta.9`
- Build Modules:    `-`

### Reproduction

Simply add normal `<NuxtLink :to"newPage"></NuxtLink>` or in my situation with i18n :
`<NuxtLink :to="localePath({ name: newPage })"></NuxtLink>`

### Describe the bug

Hi,
I know this was already added in bugs, but none of solutions I found here worked for me...
When clicking a classical NuxtLink, page does not scroll to top.

Solutions I found and tried but not working :

Plugin use : `/plugins/router.ts`
```
import { defineNuxtPlugin } from "#app";

  nuxtApp.$router.options.scrollBehavior = () => {
    return { left: 0, top: 0 };
  };
});
```

Middleware use : `/middleware/route.global.ts`
```
  useNuxtApp().hook("page:finish", () => {
    if (history.state.scroll) {
      setTimeout(() => window.scrollTo(history.state.scroll), 0);
    } else {
      setTimeout(() => window.scrollTo(0, 0), 0);
    }
  });
});
```

And finally router options use : `/app/router.options.ts`
```
import type { RouterConfig } from "@nuxt/schema";

export default <RouterConfig>{
  scrollBehavior(to, _from, savedPosition) {
    return new Promise((resolve, _reject) => {
      setTimeout(() => {
        if (savedPosition) {
          resolve(savedPosition);
        } else {
          if (to.hash) {
            resolve({
              el: to.hash,
              top: 80,
            });
          } else {
            resolve({ top: 0 });
          }
        }
      }, 100);
    });
  },
};
```

or :

```
import type { RouterOptions } from "@nuxt/schema";

export default <RouterOptions>{
  scrollBehavior(to, from, savedPosition) {
    if (savedPosition) {
      return savedPosition;
    } else {
      return {
        top: 0,
        behavior: "smooth",
      };
    }
    return { top: 0 };
  },
};
```

Non of these solutions are working and page never goes top after nuxt link click.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #25816: Scroll to top not working in firefox when `scroll-behavior: smooth` is applied
### Environment

Firefox 122.0.1

- Operating System: Darwin
- Node Version:     v20.10.0
- Nuxt Version:     3.10.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      alias, devServer, postcss, css, modules, app, build, ignore, runtimeConfig, srcDir, ssr, telemetry
- Runtime Modules:  @pinia/nuxt@0.4.11
- Build Modules:    -

### Reproduction

[stackblitz reproduction](https://stackblitz.com/edit/nuxt-starter-4f9fpe?file=app.vue)

### Describe the bug

When the style `scroll-behavior: smooth` is applied, navigation to new pages doesn't scroll to top in Firefox.

The issue is that awaiting `nextTick` in [router.options.ts](https://github.com/nuxt/nuxt/blob/599081eda20bf374923cd076e9a5e02828ca9fc2/packages/nuxt/src/pages/runtime/router.options.ts#L46) doesn't actually always play nice with Firefox for some reason, I'll open a PR for this shortly since I found a straightforward fix.

### Additional context

https://github.com/nuxt/nuxt/issues/19239 discusses this in detail (and many other issues throughout the life of Nuxt). The issue does [exist in vue-router](https://github.com/vuejs/router/issues/1411), (reproduced [in stackblitz](https://stackblitz.com/edit/github-6h8ymm?file=src%2Fpages%2FHome.vue)). However, vue router seems to have the issue (when reproduced this way) regardless of whether the style includes `scroll-behavior: smooth` or not. Since there's an easy fix on Nuxt's side ([pr](https://github.com/nuxt/nuxt/pull/25817)), I think it's worth fixing it.

### Logs

_No response_

Chalks:
### 🔗 Linked issue

fixes #25816 
fixes #19239 

### ❓ Type of change

### 📚 Description

In Firefox, if you have `scroll-behavior: smooth` style set on `html`, navigation between pages doesn't scroll to top ([reproduction](https://stackblitz.com/edit/nuxt-starter-4f9fpe?file=app.vue)). The root cause of the problem is that awaiting `nextTick` in [router.options.ts](https://github.com/nuxt/nuxt/blob/599081eda20bf374923cd076e9a5e02828ca9fc2/packages/nuxt/src/pages/runtime/router.options.ts#L46) doesn't work in that case. It's not clear to me why since the [nextTick implementation](https://github.com/vuejs/core/blob/272ab9fbdcb1af0535108b9f888e80d612f9171d/packages/runtime-core/src/scheduler.ts#L46-L58) doesn't seem to be doing anything too strange. That said, replacing `await nextTick` with awaiting a `setTimeout` fixes things. This happens to be related to something being addressed in a [currently open PR](https://github.com/nuxt/nuxt/pull/24960#pullrequestreview-1807867839) where both myself and @bernhardberger found that a `setTimeout` was necessary.

I'll admit I'm not particularly familiar with why `nextTick` is specifically being called here, but as far as I can tell replacing it the way I did fully resolves #25816.

In addition, I fixed a slight missed edge case, if you click on the same link 20 times in a row that all link to the same page, the next time you click on another link, the 20 `hookOnce`s that you created are all resolved at the same time returning undefined. It doesn't seem to cause any issues, but there's no reason to enter line 45+ if we're not going to scroll anyways when you click the same link.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/router.options.ts
+++ b/packages/nuxt/src/pages/runtime/router.options.ts
@@ -1,5 +1,4 @@
 import type { RouteLocationNormalized, RouterScrollBehavior } from '#vue-router'
-import { nextTick } from 'vue'
 import type { RouterConfig } from 'nuxt/schema'
 import { useNuxtApp } from '#app/nuxt'
 import { isChangingPage } from '#app/components/utils'
@@ -36,6 +35,8 @@ export default <RouterConfig> {
       if (to.hash) {
         return { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
       }
+      // The route isn't changing so keep current scroll position
+      return false;
     }
 
     // Wait for `page:transition:finish` or `page:finish` depending on if transitions are enabled or not
@@ -43,7 +44,7 @@ export default <RouterConfig> {
     const hookToWait = (hasTransition(from) && hasTransition(to)) ? 'page:transition:finish' : 'page:finish'
     return new Promise((resolve) => {
       nuxtApp.hooks.hookOnce(hookToWait, async () => {
-        await nextTick()
+        await new Promise((resolve) => setTimeout(resolve, 0));
         if (to.hash) {
           position = { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
         }
--- a/packages/nuxt/src/pages/runtime/router.options.ts
+++ b/packages/nuxt/src/pages/runtime/router.options.ts
@@ -36,7 +36,7 @@ export default <RouterConfig> {
         return { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
       }
       // The route isn't changing so keep current scroll position
-      return false;
+      return false
     }
 
     // Wait for `page:transition:finish` or `page:finish` depending on if transitions are enabled or not
@@ -44,7 +44,7 @@ export default <RouterConfig> {
     const hookToWait = (hasTransition(from) && hasTransition(to)) ? 'page:transition:finish' : 'page:finish'
     return new Promise((resolve) => {
       nuxtApp.hooks.hookOnce(hookToWait, async () => {
-        await new Promise((resolve) => setTimeout(resolve, 0));
+        await new Promise(resolve => setTimeout(resolve, 0))
         if (to.hash) {
           position = { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
         }


packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -36,14 +35,16 @@ export default <RouterConfig> {
       if (to.hash) {
         return { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
       }
+      // The route isn't changing so keep current scroll position
+      return false;
```
danielroe, line 39:
```suggestion
      return false
```

packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -36,14 +35,16 @@ export default <RouterConfig> {
       if (to.hash) {
         return { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
       }
+      // The route isn't changing so keep current scroll position
```
danielroe, line 38: Note that a query might have changed (e.g. `?filter=content` to `?filter=navigation`)

Chalks: Fair, but in that case today's code also wouldn't scroll anywhere because [this won't set `position`](https://github.com/nuxt/nuxt/blob/3d5c845cb2897221d05212f2eaac17b483c928bf/packages/nuxt/src/pages/runtime/router.options.ts#L27-L29) because [`isChangingPage`](https://github.com/nuxt/nuxt/blob/3d5c845cb2897221d05212f2eaac17b483c928bf/packages/nuxt/src/app/components/utils.ts#L20-L45) doesn't look at the query either.

I don't like returning `false` here, because I think that if you click on a link (even if it's the same page you're on) it should _always_ scroll to top, but since this is my first PR into this repo, I didn't want to make functional behavior changes.

My ideal solution is to mimic what happens if you click a regular link, which normally would put you at the top of the page, even if you already are on that page. To do that, I would replace [this chunk of code](https://github.com/nuxt/nuxt/blob/cf27356f9a6de97d0ac01839c670f3db01f6a71d/packages/nuxt/src/pages/runtime/router.options.ts#L30-L40) with:

```
    // Hash routes on the same page, no page hook is fired so resolve here
    if (to.path === from.path) {
      if (to.hash) {
        return { el: to.hash, top: _getHashElementScrollMarginTop(to.hash), behavior }
      }
      // The route isn't changing but we clicked a link, so scroll to top
      return { left: 0, top: 0 }
    }
```

danielroe: Okay, let's revisit that.

TheAlexLichter (approved):

CernyMatej:
What @fabianwohlfart mentioned seems reasonable to me .Awaiting a `setTimeout 0` indeed causes issues where the scroll happens a few ms after the page actually switches. @danielroe, do you see any issues with replacing it with `await new Promise(resolve => requestAnimationFrame(resolve))` ? Based on my testing, it solves the current issue. I believe you may have addressed a similar problem in one of your Twitch streams a few months ago. Was it this one?

byhndo:
How do I disable scroll to top in my blog's route and slug? I've tried various methods, but none of them work. I've used router options and meta on my blog and slug, but scroll to top still works. How do I disable it? 

router.options.ts
import type { RouterOptions } from '@nuxt/schema'

export default <RouterOptions>{
  scrollBehavior(to, from, savedPosition) {
    
    if (savedPosition) return savedPosition
    
    return false
  }
}

blog/index.vue and blog/[slug].vue

definePageMeta({
  scrollToTop: false
})

PR #25822, opened 2024-02-16
fix(nuxt): in dev mode, don't add links to `css` files that have been inlined

### LINKED ISSUE #24932: Nuxt ignores Vite's `?inline` option for CSS imports
### Environment

```
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

Vite Example: https://stackblitz.com/edit/vitejs-vite-5dsgmk?file=src%2FApp.vue
Nuxt Repro: https://stackblitz.com/edit/github-iqq3hp?file=pages%2Findex.vue

### Describe the bug

Vite offers a way for developers to import a CSS file into a variable (without having the CSS injected into `head`) using the `?inline` option documented [here](https://vitejs.dev/guide/features#disabling-css-injection-into-the-page). This can be useful if the developer wants finer control over how a CSS import is handled (e.g. I use it to add CSS to an iframe).

Nuxt seems to ignore the `?inline` option on CSS imports and injects the CSS into `head` regardless of `?inline`'s presence. For example the Nuxt repro above has a red background, even though I injected the corresponding CSS file with `?inline`.

### Additional context

_No response_

### Logs

_No response_

OnlyWick:
fix #24932.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -13,7 +13,7 @@ import { appendResponseHeader, createError, getQuery, getResponseStatus, getResp
 import devalue from '@nuxt/devalue'
 import { stringify, uneval } from 'devalue'
 import destr from 'destr'
-import { joinURL, withoutTrailingSlash } from 'ufo'
+import { getQuery as getURLQuery, joinURL, withoutTrailingSlash } from 'ufo'
 import { renderToString as _renderToString } from 'vue/server-renderer'
 import { hash } from 'ohash'
 import { renderSSRHead } from '@unhead/ssr'
@@ -394,7 +394,7 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
       const resource = styles[style]
       if (!import.meta.dev || !isRenderingIsland || (resource.file.includes('scoped') && !resource.file.includes('pages/'))) {
         // support for the ?inline query parameter for Vite version > 5.0
-        if (import.meta.dev && 'inline' in getQuery(resource.file)) continue
+        if (import.meta.dev && 'inline' in getURLQuery(resource.file)) continue
 
         link.push({ rel: 'stylesheet', href: renderer.rendererContext.buildAssetsURL(resource.file) })
       }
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -393,6 +393,9 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
     for (const style in styles) {
       const resource = styles[style]
       if (!import.meta.dev || !isRenderingIsland || (resource.file.includes('scoped') && !resource.file.includes('pages/'))) {
+        // support for the ?inline query parameter for Vite version > 5.0
+        if (process.env.NUXT_VITE_NODE_OPTIONS && resource.file.endsWith("?inline")) break;
```
danielroe, line 397:
```suggestion
        if (import.meta.dev && resource.file.endsWith('?inline')) continue
```

danielroe: we should probably also parse the url in case there are other queries using `getQuery` from `ufo`...

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -393,6 +393,9 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
     for (const style in styles) {
       const resource = styles[style]
       if (!import.meta.dev || !isRenderingIsland || (resource.file.includes('scoped') && !resource.file.includes('pages/'))) {
+        // support for the ?inline query parameter for Vite version > 5.0
+        if (import.meta.dev && resource.file.endsWith('?inline')) continue
```
danielroe, line 397:
```suggestion
        if (import.meta.dev && 'inline' in getQuery(resource.file)) continue
```

packages/nuxt/src/core/runtime/nitro/renderer.ts
OnlyWick: yeah, we really should parse the url. But, I don't know much about `h3` and `ufo`, you can switch to their api.

danielroe (approved):

danielroe (approved):

Issue #25033, opened 2024-01-03
Multiple `useLazyAsyncData` within a composable becomes non lazy


huang-julien:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v20.2.0
- Nuxt Version:     3.7.4
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.12.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-t7zx5g?file=composables%2Ftest.ts

- click on to lazy
- the navigation is blocked

### Describe the bug

Using 2 `useLazyAsyncData` make the page change blocking

gibkigonzo:
it looks like `getCurrentInstance` doesn't exist for second `useLazyAsyncData`. So instead of going here https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L359-L362 it goes to https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L363-L365

jamesdawsonWD:
I believe that this is working correctly and that the issue lies with how it is documented. The example you provided @huang-julien uses `await` on `useLazyAsyncData` but these composables are not async and I think the awaits are causing the problem.

I believe the awaits make the vue component instance become undefined for any sequential `await` calls (ref: https://nuxt.com/docs/api/composables/use-nuxt-app ). And since instance is undefined, lazyAsyncData has to revert to just asyncData which blocks the route change because it can't call `instance._nuxtOnBeforeMountCbs.push(initialFetch)`. This is also why one `useLazyAsyncData` with `await` works fine because the component instance is there on the first call. 

Playground with working multiple `useLazyAsyncData`: https://stackblitz.com/edit/nuxt-starter-3ojz7s?file=pages%2Findex.vue

If this is correct, I suggest we make a simple doc change to remove the superfluous `await`. https://nuxt.com/docs/api/composables/use-lazy-async-data

I created a pull request with the documentation change :) https://github.com/nuxt/nuxt/pull/25851

huang-julien:
Didn't had the time to investigate yet, but it can be a unctx issue

jamesdawsonWD:
[...]

PR #25851, opened 2024-02-18
docs: remove superfluous await from useLazyAsync page

jamesdawsonWD:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25033

### ❓ Type of change

### 📚 Description

I have been investigating why multiple `useLazyAsyncData` composables inside composition setup are blocking the route change and are not "lazy". I believe the issue is with how it is documented rather than a problem with the composable itself. The documentation recommends using await on a non async function `useLazyAsyncData`. My change is to remove the `await` from the `useLazyAsyncData` documentation page

I believe the awaits make the vue component instance become undefined for any sequential `await` calls, I read [here](https://nuxt.com/docs/api/composables/use-nuxt-app) that vue should handle undefining and setting the component instance when an await is called but this seems to not be the case for these composables when wrapped in `await`.  Since the instance is undefined, `lazyAsyncData` has to revert to `asyncData` which blocks the route change because it cannot call `instance._nuxtOnBeforeMountCbs.push(initialFetch)`. This is also why one `useLazyAsyncData` with `await` works fine because the component instance is there on the first call. 

Playground with working multiple `useLazyAsyncData`: https://stackblitz.com/edit/nuxt-starter-3ojz7s?file=pages%2Findex.vue

### 📝 Checklist

### DIFF
--- a/docs/3.api/2.composables/use-lazy-async-data.md
+++ b/docs/3.api/2.composables/use-lazy-async-data.md
@@ -25,7 +25,7 @@ By default, [`useAsyncData`](/docs/api/composables/use-async-data) blocks naviga
 /* Navigation will occur before fetching is complete.
   Handle pending and error states directly within your component's template
 */
-const { pending, data: count } = await useLazyAsyncData('count', () => $fetch('/api/count'))
+const { pending, data: count } = useLazyAsyncData('count', () => $fetch('/api/count'))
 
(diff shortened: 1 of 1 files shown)


huang-julien (changes requested):
Removing await exposes you to XRSP

PR #25864, opened 2024-02-19
fix(nuxt): align `pending` with `status` value for v4

### LINKED ISSUE #25225: Remove `pending` in favor of `status` in data fetching composables
### Describe the feature

Currently, data fetching composables expose pending information in two ways: `pending` and `status` (`status.value === 'pending'`). It would be great to consider removing `pending` for consistency, which currently has a duplicated source of truth. As this would be a breaking change, it's probably best to address it in the next version of Nuxt.

([Update in this comment](https://github.com/nuxt/nuxt/issues/25225#issuecomment-2182690498))

### Additional information

### Final checks

drowhann:
### 🔗 Linked issue
resolves #25225 

### ❓ Type of change

### 📚 Description

I have removed `pending` in favor of `status` as discussed in issue #25225.   

### 📝 Checklist


docs/3.api/3.utils/clear-nuxt-data.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: 'clearNuxtData'
-description: Delete cached data, error status and pending promises of useAsyncData and useFetch.
+description: Delete cached data and error status promises of useAsyncData and useFetch.
```
DamianGlowala, line 3: I think this might be left unchanged, as it refers to pending _promises_.

drowhannn: Yeah, i think i misunderstood the part.

DamianGlowala (approved):

UfukUstali:
If there are going to be breaking changes in the useFetch's signature I think it would be also be good to at least **consider** returning a reactive object rather than individual refs.

My reasoning is because currently returned values are refs we cannot leverage discriminated unions. With a reactive object however because it is not wrapped with refs typescript can build discriminated unions around it's type.

I had a working wrapper around the current useFetch already that is properly typed to leverage the benefits of the change I will post the repo here (or in the issue that we decide to track this in if that were to happen) once I clean it up a bit more and if there is any interest.

**Benefits:**
Simplified type checks around useFetch because we can just check the status and be sure that data is also there (for clarification this is already the case just typescript is not aware of it)

Improved DX for the reasons listed above

**Downsides:**
We lose the ability to destructure the useFetch call due to reactivity breaking

Potentially and **most** importantly a rather tedious migration experience (both in terms of code and convention/habits) for people that were using the destructuring pattern extensively

These were the ones that were obvious to me at first glance I am sure that there are more implications around this kind of a change that I am not aware of.

DamianGlowala (changes requested):

PR #25906, opened 2024-02-22
feat(nuxt): sort numbered middleware/plugins from all layers

markthree:
### 🔗 Linked issue

#25891 

### ❓ Type of change

### 📚 Description

Only sort the global middleware, if there are numbers, they will be moved forward in order, Resolves #25891

### 📝 Checklist


packages/nuxt/src/core/templates.ts
```diff
@@ -208,14 +208,21 @@ export const layoutTemplate: NuxtTemplate = {
...
     return [
       ...globalMiddleware.map(mw => genImport(mw.path, genSafeVariableName(mw.name))),
       `export const globalMiddleware = ${genArrayFromRaw(globalMiddleware.map(mw => genSafeVariableName(mw.name)))}`,
       `export const namedMiddleware = ${namedMiddlewareObject}`
     ].join('\n')
+    function sortGlobalMiddleware (globalMiddleware: NuxtMiddleware[]): NuxtMiddleware[] {
+      const reg = /^\d/
+      const withOrdergGobalMiddleware = globalMiddleware.filter(m => reg.test(m.name)).toSorted((l, r) => l.name > r.name ? 1 : -1)
+      const withoutOrdergGobalMiddleware = globalMiddleware.filter(m => !reg.test(m.name))
+      return  [...withOrdergGobalMiddleware, ...withoutOrdergGobalMiddleware]
```
DamianGlowala, line 224:
```suggestion
      const orderedGlobalMiddleware = globalMiddleware.filter(m => reg.test(m.name)).toSorted((l, r) => l.name > r.name ? 1 : -1)
      const unorderedGlobalMiddleware = globalMiddleware.filter(m => !reg.test(m.name))
      return  [...orderedGlobalMiddleware, ...unorderedGlobalMiddleware]
```

danielroe (changes requested):
Thank you! ❤️

We create the middleware here:

https://github.com/nuxt/nuxt/blob/b96fe1ececaf65fd276da26c35864df9c729a704/packages/nuxt/src/core/app.ts#L130-L143

We could sort them there, but I think it would (also?) be better to sort the results coming from `resolveFiles`:

https://github.com/nuxt/nuxt/blob/a2ef3091e856a2adf27cf5efcd13b56718e23ac1/packages/kit/src/resolve.ts#L163-L166

It would also be amazing if you could add a test case here:

https://github.com/nuxt/nuxt/blob/b784336f8aec5c9b60a375536189c3bf50bdea7c/packages/nuxt/test/app.test.ts#L152-L183

markthree:
@danielroe Thanks. I'll try it.

markthree:
@danielroe，It's done, and while this way of writing it isn't the best for performance, it's clear enough and also introduces as little break change as possible. 🥰

Maybe in the future we can have resolveFiles upper level tools that can handle it, but I can't think of a good design right now (without introducing break change) 🤣

packages/nuxt/src/core/app.ts
```diff
@@ -239,3 +240,19 @@ export function checkForCircularDependencies (_plugins: Array<NuxtPlugin & Omit<
...
+function sortMiddleware(middleware: NuxtMiddleware[]) {
+  const globalMiddleware = middleware.filter(mw => mw.global)
+  const namedMiddleware = middleware.filter(mw => !mw.global)
+  return [
+    ...sortOrderedMiddleware(globalMiddleware),
+    ...sortOrderedMiddleware(namedMiddleware)
+  ]
+  function sortOrderedMiddleware (middleware: NuxtMiddleware[]) {
+    const reg = /^\d/
+    const orderedMiddleware = middleware.filter(m => reg.test(m.name)).toSorted((l, r) => l.name > r.name ? 1 : -1)
```
danielroe, line 254:
```suggestion
    const orderedMiddleware = middleware.filter(m => reg.test(m.name)).sort((l, r) => l.name > r.name ? 1 : -1)
```

packages/nuxt/src/core/app.ts
```diff
@@ -239,3 +240,19 @@ export function checkForCircularDependencies (_plugins: Array<NuxtPlugin & Omit<
     checkDeps(name)
   }
 }
+
```
danielroe, line 243:
```suggestion
```

danielroe (approved):

danielroe:
I'm rethinking this. Plugins explicitly don't behave this way:

https://github.com/nuxt/nuxt/blob/460daa0f5c49d22fb2588109bdb42edba38155ef/packages/nuxt/test/app.test.ts#L138-L148

I think if we want numbered plugins/middleware to be sorted and hoisted out, they should have the same behaviour.

danielroe (approved):

markthree:
Thank you @danielroe , I've been so busy lately, I meant to take a look at it when I had some free time. 😭

packages/nuxt/src/core/app.ts
```diff
@@ -252,3 +254,51 @@ export function checkForCircularDependencies (_plugins: Array<NuxtPlugin & Omit<
...
+    const name = plugin.name ? plugin.name : filename(plugin.src)
+    const mode = plugin.mode ? plugin.mode : 'all'
+    const flag = `${name}.${mode}`
+    if (pluginFlags.has(flag)) {
+      continue
+    }
+    pluginFlags.add(flag)
+    bucket.push(plugin)
+  }
+  return bucket.reverse()
```
Kronhyx, line 303:
```suggestion
function uniquePlugins (plugins: NuxtPlugin[]) {
  const pluginFlags = new Set<string>()
  const bucket: NuxtPlugin[] = []

  for (const plugin of [...plugins].reverse()) {
    const name = plugin.name ? plugin.name : filename(plugin.src)
    const mode = plugin.mode ? plugin.mode : 'all'
    const flag = `${name}.${mode}`

    if (pluginFlags.has(flag)) {
      continue
    }

    pluginFlags.add(flag)
    bucket.push(plugin)
  }

  return bucket.reverse()
```

Kronhyx: add line breaks to improve readability

Kronhyx (changes requested):

packages/nuxt/src/core/app.ts
oemer-aran: Any reason that this cannot be merged?

mkierdev:
@danielroe any news on this? Should we create middlewares in plugins instead?

PR #25924, opened 2024-02-22
fix(schema): add vueCompilerOptions property to TSConfig in defineNuxtConfig

### LINKED ISSUE #25874: 'vueCompilerOptions' does not exist in type 'tsConfig' in 'defineNuxtConfig'
### Describe the feature

# vueCompilerOptions in nuxt.config.ts

Currently, the `defineNuxtConfig`'s `typescript.tsConfig.vueCompilerOptions` property is not officially supported on type-level: the compiler will yell at you if you try to use it.
However, the configuration options defined in `vueCompilerOptions` still work as intended.

Enhancing the `tsConfig` interface locally to include the `vueCompilerOptions` property is a possible workaround (the `VueCompilerOptions` type is currently available and exported from vue's languageModule), but an official solution would be much appreciated.

You can find this reproduction [on stackblitz](https://stackblitz.com/edit/github-58xcrx?file=README.md)

### Prerequisites

- A minimal Nuxt app started with `npx nuxi@latest init`
- `vue-tsc` and `typescript` installed as dev dependencies (see [Nuxt's TypeScript documentation](https://nuxt.com/docs/guide/concepts/typescript#type-checking))
- A `nuxt.config.ts` file with the `vueCompilerOptions` property added to the `tsConfig` property. To observe the behavior, the easiest way is to add the `strictTemplates` property to the `vueCompilerOptions` object.

### Steps to reproduce

1. Create a component and define a prop in the component using the `defineProps` macro and passing it the `Props` type argument
2. Use your new component in a page and pass it the prop you defined (this is not really necessary, but it's a good way to observe the behavior)
3. Pass your component a prop that is not defined in the component's `props`, (e.g. `:nonExistentProp="true"`)
4. Save and run `npx nuxi typecheck`. You l'll get one error from the `nuxt.config.ts` file, saying that the `vueCompilerOptions` property is not allowed, and one from the page where you used the component, saying that `non-existent-prop` is not defined in the component's `Props` type.

### Additional information

### Final checks

Evobaso-J:
### 🔗 Linked issue
#25874

### ❓ Type of change

### 📚 Description

Resolves #25874

`VueCompilerOptions` type added to the `defineNuxtConfig` schema

### 📝 Checklist


Evobaso-J:
@manniL Hi there! 
This one is ready for code review. However, I'm not sure whether documentation on the VueCompilerOptions is needed or if a type addition, in general, is self-explanatory 🤔 

By the way, I must say I couldn't find any reference to these options anywhere [but on Volar's changelog](https://github.com/vuejs/language-tools/blob/master/CHANGELOG.md), so having those documented somewhere could be nice!

Please let me know what you think :v:
TY

packages/schema/build.config.ts
```diff
@@ -53,6 +53,7 @@ export default defineBuildConfig({
     'sass-loader',
     'c12',
     'unenv',
+    "@vue/language-core",
```
danielroe, line 56:
```suggestion
    '@vue/language-core',
```

packages/schema/src/types/hooks.ts
```diff
@@ -19,6 +20,9 @@ export type HookResult = Promise<void> | void
 export type TSReference = { types: string } | { path: string }
 
 export type WatchEvent = 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir'
+export type VueTSConfig = TSConfig & {
+  vueCompilerOptions?: VueCompilerOptions;
+};
```
danielroe, line 25: If the user doesn't have this dependency installed, this (and mention above in schema) will be typed as any. We could use something like this type:

```ts
type IsAny<T> = 0 extends 1 & T ? true : false
```

to conditionally augment tsConfig only when `VueCompilerOptions` is not any...

Evobaso-J: @danielroe changes made!
I've been pondering about it for some days now but I still don't quite understand one thing: how can the user not have the dependency installed if they're installing nuxt? 🤔
Also, why don't we need to make the same check for other dependencies as well?

danielroe: The user will only have this dependency installed if they add `vue-tsc`. Nuxt doesn't ship with `vue-tsc` by default.

danielroe (approved):

packages/schema/src/types/hooks.ts
Evobaso-J: Oh, I see! Thank you 😄

danielroe:
Thank you ❤️

PR #25932, opened 2024-02-23
feat(nuxt): custom loading reset/hide delay + force `finish()`

### LINKED ISSUE #24938: [feat] useLoadingIndicator force reset function
### Describe the feature

Hello,

As discussed in #24900.

There should be an option for `useLoadingIndicator.start({ force: true })` so when it's called the Loading indicator should start from scratch.

Currently, after the page load it's not possible to manually start `useLoadingIndicator` because there is `setTimeout` which is basically preventing the start:

https://github.com/nuxt/nuxt/blob/f5676fba5668c3dbff8c1c40e988d469740ee82e/packages/nuxt/src/app/composables/loading-indicator.ts#L16-L23

As suggested by @manniL I could use a workaround and wait 1 second

```javascript
  // eslint-disable-next-line unused-imports/no-unused-vars
  async setup(nuxtApp) {},
  hooks: {
    // You can directly register Nuxt app runtime hooks here
    'page:loading:end': async function () {
      console.log('Calling startLoading function from a plugin!');
      await new Promise(r => setTimeout(r, 1000))
      startLoading();
    },
  },
});
```

Honestly, this makes loading indicator as lagging which is weird.

Thanks!

### Additional information

### Final checks

divine:
### 🔗 Linked issue
Fix: #24938

### ❓ Type of change

### 📝 Checklist


packages/nuxt/src/app/composables/loading-indicator.ts
```diff
@@ -50,7 +61,12 @@ function createLoadingIndicator (opts: Partial<LoadingIndicatorOpts> = {}) {
 
   let _throttle: any = null
 
-  const start = () => set(0)
+  const start = (force: boolean | null = true) => {
```
danielroe, line 64: What about updating the signature so we call this with `start({ force: true})` (and also below with `finish`)? This makes it more explicit and allows passing more options in future.

danielroe: Sorry, to clarify I don't mean passing it as an option to `useLoadingIndicator` but to the `start` and `finish` functions.

divine: Hi Daniel!

I thought about it as well but the better would be passing force as parameter to loading indicator instead.

Let's say:
1. LoadingIndicator starts
2. We try to start the new one with `start({force: true})`.
3. LoadingIndicators starts again.
4. `_hide` function is being called and LoadingIndicator is being hidden because it was still running because of setTimeout...

Let me know what do you think!

Thanks!

danielroe: The fewer options that `useLoadingIndicator` accepts, the better, because they will influence the global singleton.

I think we should pass the override to start/finish and calling `_hide` or `_reset` should instead clear any timeout that is still running. (We probably should have implemented it that way in the first place.)

divine: Hi @danielroe,

I've applied your suggestions, let me know wdyt.

Thanks!

danielroe: Thank you!

danielroe (approved):

PR #25945, opened 2024-02-25
fix(nuxt): align `error` param in `showError`/`createError` with H3

### LINKED ISSUE #25942: Nuxt `createError` param signature does not match h3's
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-qz4t9r?file=test.ts

### Describe the bug

In `h3` docs error handling advises to use `status` in `createError` (https://h3.unjs.io/guide/event-handler#error-handling). When you try do the same in some vue page or component, you will get type error, despite [Nuxt using h3's `createError` under the hood](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/error.ts#L61). This is because `h3` also allows `status` and `statusText` along side other `H3Error` props.

`createError` params [in Nuxt](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/error.ts#L59) and [in h3](https://github.com/unjs/h3/blob/main/src/error.ts#L70).

This is minor thing, but still results in unpleasant dx, because you need to remember to use different params in vue and server side

### Additional context

_No response_

### Logs

_No response_

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25942

### ❓ Type of change

### 📚 Description

Aligns `error` param of Nuxt's  `showError` and `createError` with H3's. 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/error.ts
+++ b/packages/nuxt/src/app/composables/error.ts
@@ -56,7 +56,7 @@ export const isNuxtError = <DataT = unknown>(
 
 /** @since 3.0.0 */
 export const createError = <DataT = unknown>(
-    error: string | Partial<NuxtError<DataT>>
+    error: Parameters<typeof createH3Error<DataT>>[0]
 ) => {
   const nuxtError: NuxtError<DataT> = createH3Error<DataT>(error)
 
--- a/packages/nuxt/src/app/composables/error.ts
+++ b/packages/nuxt/src/app/composables/error.ts
@@ -13,10 +13,10 @@ export interface NuxtError<DataT = unknown> extends H3Error<DataT> {}
 
 /** @since 3.0.0 */
 export const showError = <DataT = unknown>(
-    error: string | Error | (Partial<NuxtError<DataT>> & {
-        status?: number;
-        statusText?: string;
-    })
+  error: string | Error | (Partial<NuxtError<DataT>> & {
+    status?: number;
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/error.ts
```diff
@@ -13,7 +13,7 @@ export interface NuxtError<DataT = unknown> extends H3Error<DataT> {}
 
 /** @since 3.0.0 */
 export const showError = <DataT = unknown>(
-    error: string | Error | Partial<NuxtError<DataT>>
+    error: Parameters<typeof createH3Error<DataT>>[0]
```
danielroe, line 16: Could we align it rather than use `Parameters`? 🙏  I might be wrong but worry this might cause issues with either TS perf or non-hoisted deps (if h3 isn't installed in root level).

Also, it's useful to have `NuxtError` as a separate interface that can be extended.

We can add a type assertion that isn't exported that will fail the build if these two aren't identical, so we keep the safety of ensuring we don't diverge from h3.

danielroe (approved):

Issue #25949, opened 2024-02-25
Generated ID from `useId` is no valid CSS selector


wokalek:
### Environment

- Operating System: Linux
- Node Version:     v20.11.1
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.15.4
- Builder:          -
- User Config:      telemetry, devtools, srcDir, app, extends, modules, vite, appConfig, runtimeConfig, eslint, tailwindcss, css, postcss, colorMode, device, content, pwa, robots, sitemap, site
- Runtime Modules:  @nuxtjs/eslint-module@4.1.0, @nuxtjs/tailwindcss@6.11.4, @nuxtjs/robots@3.0.0, @nuxtjs/sitemap@5.1.0, @nuxtjs/color-mode@3.3.2, @nuxtjs/fontaine@0.4.1, @nuxtjs/device@3.1.1, @nuxt/content@2.12.0, @nuxt/image@1.3.0, @vueuse/nuxt@10.8.0, @vite-pwa/nuxt@0.5.0, nuxt-typed-router@3.6.1, nuxt-svgo@4.0.0, nuxt-schema-org@3.3.4, nuxt-lodash@2.5.3
- Build Modules:    -

### Reproduction

use `useId`, then try for querySelectorAll

### Describe the bug

I get this error when I use querySelectorAll

```
Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Document': 'JfvpJe1itr:5' is not a valid selector.
```

But `document.getElementById('JfvpJe1itr:18')` is working.

Google suggests that this is [not a valid selector](https://stackoverflow.com/a/63179238/7274005).

Would it be better to use `_` insted of `:`?

manniL:
Try `document.getElementById('JfvpJe1itr\:18')`

wokalek:
> Try `document.getElementById('JfvpJe1itr\:18')`

Yes, I already found that you can escape with `\\`, like this 

```js
document.getElementById('JfvpJe1itr\\:18')
```

wokalek:
I just think that since it was meant to conveniently use `useId()` for linking inside libraries, such as PhotoSwipe, which need to pass `id`, or they themselves take the id of the element from the attributes (and you have no control over this, internally they can use querySelectorAll), then it would be best to make the row generation as universal as possible.

wokalek:
[...]

PR #25969, opened 2024-02-26
fix(nuxt): produce valid css selector from `useId`

TheAlexLichter:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25949

### ❓ Type of change

### 📚 Description

This PR adds an `n` in front of all results from `useId` and ensures that `_` is used as separator instead of  `:`, ensuring that the result can be used as CSS selector without sanitizing.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/id.ts
+++ b/packages/nuxt/src/app/composables/id.ts
@@ -3,7 +3,7 @@ import { useNuxtApp } from '../nuxt'
 import { clientOnlySymbol } from '#app/components/client-only'
 
 const ATTR_KEY = 'data-n-ids'
-const SEPARATOR = '_'
+const SEPARATOR = '-'
 
 /**
  * Generate an SSR-friendly unique identifier that can be passed to accessibility attributes.
--- a/packages/nuxt/src/app/composables/id.ts
+++ b/packages/nuxt/src/app/composables/id.ts
@@ -3,6 +3,7 @@ import { useNuxtApp } from '../nuxt'
 import { clientOnlySymbol } from '#app/components/client-only'
 
 const ATTR_KEY = 'data-n-ids'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/id.ts
```diff
@@ -26,11 +28,11 @@ export function useId (key?: string): string {
   instance._nuxtIdIndex ||= {}
   instance._nuxtIdIndex[key] ||= 0
 
-  const instanceIndex = key + ':' + instance._nuxtIdIndex[key]++
+  const instanceIndex = key + SEPARATOR + instance._nuxtIdIndex[key]++
 
   if (import.meta.server) {
     const ids = JSON.parse(instance.attrs[ATTR_KEY] as string | undefined || '{}')
-    ids[instanceIndex] = key + ':' + nuxtApp._id++
+    ids[instanceIndex] = key + SEPARATOR + nuxtApp._id++
```
danielroe, line 35: The separator needs to be different from L57 to avoid collisions.

packages/nuxt/src/app/composables/id.ts
```diff
@@ -3,6 +3,7 @@ import { useNuxtApp } from '../nuxt'
 import { clientOnlySymbol } from '#app/components/client-only'
 
 const ATTR_KEY = 'data-n-ids'
+const SEPARATOR = '_'
```
manniL, line 6:
```suggestion
const SEPARATOR = '-'
```

packages/nuxt/src/app/composables/id.ts
manniL: Updated as `-` 👍🏻

danielroe (approved):

PR #25978, opened 2024-02-27
docs(nuxt): move play online under installation

moshetanzer:
### ❓ Type of change

### 📚 Description
Hi,

Seems to be that play online should be under installation, since many users come back when installing a new Nuxt project and obviously want local install properly. Play online is cute, but IMHO not most important and note most used. Thus, i think installation guide should come first. 
Another option is to put a little callout with a play online anchor

### 📝 Checklist


danielroe:
I think the play online links are short enough that it makes sense to keep them at the top, particularly given it's the page linked to from 'Get Started' on homepage which is more generic than 'Installation'. Do you think they stop people from finding out how to install Nuxt?

What do you think @atinux?

moshetanzer:
Hi @daniel,
Nope i dont think it would stop people installing nuxt😂, just think that
greatest use of page - and most used part of page is installing nuxt not
playing in an online ide.
Though what trust what you think. As i suggested that even if you want to
mention it there maybe do it in a card with anchor. Just my opinion, Keep
up your work!

danielroe (approved):
Reworded it slightly to make the purpose of this clearer, and added a link directly within the getting started introduction as well.

Thank you ❤️

PR #26045, opened 2024-03-02
feat(nuxt): add `page:view-transition:start` hook

### LINKED ISSUE #26044: Provide a way to obtain the ongoing ViewTransition object
### Describe the feature

The current implementation of the experimental ViewTransition plugin is great, but it can be limiting not having the actual transition object returned by 'document.startViewTransition'. This object can be used to wait for certain events during the transition (eg.: finished).

It would be great to receive this object from nuxt when one is created.
This would be needed in Elks (in-progress) view transition implementation as described in this issue here: https://github.com/elk-zone/elk/pull/2211#issuecomment-1974770960

### Additional information

### Final checks

horvbalint:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/26044

### ❓ Type of change

### 📚 Description

This is a small PR that introduces a new client lifecycle hook, which is called right after a [ViewTransition object](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition) is created by the [experimental view transition plugin](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental).

Using this hook an app can obtain the transition object.

Resolves #26044

### 📝 Checklist

### DIFF
--- a/docs/3.api/6.advanced/1.hooks.md
+++ b/docs/3.api/6.advanced/1.hooks.md
@@ -28,6 +28,7 @@ Hook                   | Arguments           | Environment     | Description
 `page:loading:start`   | -                   | Client          | Called when the `setup()` of the new page is running.
 `page:loading:end`     | -                   | Client          | Called after `page:finish`
 `page:transition:finish`| `pageComponent?`    | Client          | After page transition [onAfterLeave](https://vuejs.org/guide/built-ins/transition.html#javascript-hooks) event.
+`page:view-transition:start` | `transition`        | Client          | Called after document.startViewTransition is called by the [experimental nuxt plugin](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental).
 
 ## Nuxt Hooks (build time)
 
--- a/docs/3.api/6.advanced/1.hooks.md
+++ b/docs/3.api/6.advanced/1.hooks.md
@@ -28,7 +28,7 @@ Hook                   | Arguments           | Environment     | Description
 `page:loading:start`   | -                   | Client          | Called when the `setup()` of the new page is running.
 `page:loading:end`     | -                   | Client          | Called after `page:finish`
 `page:transition:finish`| `pageComponent?`    | Client          | After page transition [onAfterLeave](https://vuejs.org/guide/built-ins/transition.html#javascript-hooks) event.
-`page:view-transition:start` | `transition`        | Client          | Called after document.startViewTransition is called by the [experimental nuxt plugin](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental).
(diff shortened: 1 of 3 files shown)


horvbalint:
I am not sure, how to solve the not exported interface error in a way so that it can be imported in projects using nuxt.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -45,6 +45,7 @@ export interface RuntimeNuxtHooks {
   'page:finish': (Component?: VNode) => HookResult
   'page:transition:start': () => HookResult
   'page:transition:finish': (Component?: VNode) => HookResult
+  'page:view-transition:start': (transition: ViewTransition) => HookResult
```
GalacticHypernova, line 48: How about providing an exported type in nuxt and importing it here?
```ts
export interface ViewTransition {
    ready: Promise<void>
    finished: Promise<void>
    updateCallbackDone: Promise<void>
  }
  ```
  
  ```ts
import {ViewTransition} from "wherever you'll place it"
```

horvbalint: Yeah I think I overthinked this one 😅 I tought it has to be importable as "import type {ViewTransition} from 'nuxt'"

GalacticHypernova: Happens to the best of us =)

packages/nuxt/src/app/plugins/view-transitions.client.ts
```diff
@@ -35,6 +35,7 @@ export default defineNuxtPlugin((nuxtApp) => {
       changeRoute()
       return promise
     })
+    nuxtApp.callHook('page:view-transition:start', transition)
```
huang-julien, line 38:
```suggestion
    await nuxtApp.callHook('page:view-transition:start', transition)
```
We probably need to await the hook

docs/3.api/6.advanced/1.hooks.md
```diff
@@ -28,6 +28,7 @@ Hook                   | Arguments           | Environment     | Description
 `page:loading:start`   | -                   | Client          | Called when the `setup()` of the new page is running.
 `page:loading:end`     | -                   | Client          | Called after `page:finish`
 `page:transition:finish`| `pageComponent?`    | Client          | After page transition [onAfterLeave](https://vuejs.org/guide/built-ins/transition.html#javascript-hooks) event.
+`page:view-transition:start` | `transition`        | Client          | Called after document.startViewTransition is called by the [experimental nuxt plugin](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental).
```
danielroe, line 31:
```suggestion
`page:view-transition:start` | `transition`        | Client          | Called after `document.startViewTransition` is called when [experimental viewTransition support is enabled](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental).
```

danielroe (approved):

PR #26051, opened 2024-03-03
docs(nuxt): update additional notes

oumarbarry:
Takeover Mode has been deprecated.
"TypeScript Vue Plugin" extension has been deprecated.

PR #26052, opened 2024-03-03
fix(schema): disable `typescript.shim` by default in favour of volar features

oumarbarry:
Takeover Mode has been deprecated.
"TypeScript Vue Plugin" extension has been deprecated.

closes https://github.com/nuxt/nuxt/pull/26051


packages/schema/src/config/typescript.ts
```diff
@@ -50,8 +50,7 @@ export default defineUntypedSchema({
     /**
      * Generate a `*.vue` shim.
      *
-     * We recommend instead either enabling [**Take Over Mode**](https://vuejs.org/guide/typescript/overview.html#volar-takeover-mode) or adding
-     * TypeScript Vue Plugin (Volar)** 👉 [[Download](https://marketplace.visualstudio.com/items?itemName=Vue.vscode-typescript-vue-plugin)].
+     * We recommend instead installing the [official Vue extension](https://marketplace.visualstudio.com/items?itemName=Vue.volar).
      */
     shim: true
```
danielroe, line 55: We should update the default to false in this case.

oumarbarry: yeah for sure

PR #26068, opened 2024-03-04
fix(nuxt): cache-bust payloads with build id

### LINKED ISSUE #15427: Adding hashes to the _payload.js files
### Describe the feature

When using a CDN like Cloudflare it's normal to use hashes on .js files in order for the cache to be automatically cleared on redeployment as the file name (and the hash) changes.

Now it's not really possible to do that so it's hard to manage re-deployment as we have to manually clean the cache each time. At least that's what I see from the docs.

It would be nice if Nuxt would generate the payload files with a hash, like this: `_payload.{hash}.js` by default. Or at least an option to specify the file name would be nice to have.

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15427

### ❓ Type of change

### 📚 Description

This moves payload hashes to a query string and uses the build ID to provide cache-busting capability.

### 📝 Checklist


packages/nuxt/src/app/composables/payload.ts
```diff
@@ -50,14 +51,14 @@ export function preloadPayload (url: string, opts: LoadPayloadOptions = {}) {
...
-const extension = renderJsonPayloads ? 'json' : 'js'
+const filename = renderJsonPayloads ? '_payload.json' : '_payload.js'
 function _getPayloadURL (url: string, opts: LoadPayloadOptions = {}) {
   const u = new URL(url, 'http://localhost')
   if (u.host !== 'localhost' || hasProtocol(u.pathname, { acceptRelative: true })) {
     throw new Error('Payload URL must not include hostname: ' + url)
   }
-  const hash = opts.hash || (opts.fresh ? Date.now() : '')
-  return joinURL(useRuntimeConfig().app.baseURL, u.pathname, hash ? `_payload.${hash}.${extension}` : `_payload.${extension}`)
+  const hash = opts.hash || (opts.fresh ? Date.now() : (useAppConfig().nuxt as any)?.buildId)
```
pi0, line 60: I haven't tested but this wouldn't affect on domain prerenderin (in runtime) and only append buildId as hash in `prerender` env right?

pi0 (approved):

packages/nuxt/src/app/composables/payload.ts
danielroe: No, this would always add the hash, so that when a pre-rendered payload is re-requested it bypasses browser cache

axelerod12:
Is it possible to include the build ID here as well?https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/runtime/nitro/renderer.ts#L434
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/core/runtime/nitro/renderer.ts#L394

danielroe:
That's a good idea. I'd welcome a PR to do it @axelerod12 🙏

PR #24717, opened 2023-12-12
perf(kit): avoid duplicate `join` operation

GalacticHypernova:
### 🔗 Linked issue

#24707

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

There is a duplicate filesystem operation in the nuxt module resolving process. This PR saves the reference to prevent the redundancy and thereby improve the performance.

NOTE: While the performance improvement here is way more miniscule, it still exists. I couldn't test it with the methods nuxt uses (I will make a benchmark with that soon) but with native fs's methods there's a ~10ms decrease in execution time, which is much more noticeable is longer iterations (**~8.38%** performance increase in 1000 iterations):
```ts
import { existsSync, promises as fsp } from "node:fs"
import { dirname, join } from "node:path"
import { fileURLToPath } from "node:url"
async function main() {
    async function test() {
        let buildTimeModuleMeta
        if (existsSync(join(dirname(fileURLToPath(import.meta.url)), 'Test2.json'))) {
            buildTimeModuleMeta = JSON.parse(await fsp.readFile(join(dirname(fileURLToPath(import.meta.url)), 'Test2.json'), 'utf-8'))
        }
    }
    async function test2() {
        let buildTimeModuleMeta
        const modB = join(dirname(fileURLToPath(import.meta.url)), 'Test2.json')
        if (existsSync(modB)) {
            buildTimeModuleMeta = JSON.parse(await fsp.readFile(modB, 'utf-8'))
        }
    }
    console.time("Current")
    for (let index = 0; index < 1000; index++) {
        await test()
    }
    console.timeEnd("Current")
    console.time("New")
    for (let index = 0; index < 1000; index++) {
        await test2()
    }
    console.timeEnd("New")
}
main()
```
```ts
//Test.json
{
"name":"test"
}
```
image

I don't know if pathe implements it differently, so I don't mind this not being merged for the time being (until I make the tests), however even my mind tells me a duplicate operation, even if insanely fast, will still take longer than saving a reference to it.

### 📝 Checklist

[...]

PR #26114, opened 2024-03-06
perf: performance overhaul

GalacticHypernova:
## NOTE:
This is a continuation and a rework of the previous PR that had some unexplainable issues so I decided to rewrite it. I have opened this to rework it more carefully and take care of all the edge cases and random errors. (For more info, please refer to #25771)

## NOTE 2:
The 70 docs commits are due to me forgetting to flush the original patch-9 branch when I was done with that one. I usually start these PR's remotely, which means I don't get to name them something unique, and I just got onto the github app to try and re-run the failed tests, which still had the old patch-9 commits. That's my bad. I used the origin's files in every conflict so it ended up just not changing anything (as can be seen in the files changed).

### 🔗 Linked issue

Related: #25771 

### ❓ Type of change

### 📚 Description

This is a PR that aims to improve both build time and runtime performance to the max by refactoring everything that comes to mind with performance best practices, some of which include smarter and fewer object/array iterations, reduced indexes, smarter variable usage for constant values, and more. 

Some key improvements in this PR:
- Refactors of known costly javascript methods:
   In some places throughout the app, there are operations that rely on certain javascript methods that are known to be more expensive at certain conditions than manual implementation 
   (similar to `endsWith` in  #24746 
    and `startsWith` in  #24744). 
    One key change under this category is the usage of  `Array.prototype.includes`.
    `Array.prototype.includes` appears to be slower when checking for existence of a certain input between at the very least 2 elements than directly checking each element with strict equality (like when dealing with strings).
    After thorough benchmarking, both browser and node, direct comparisons showed marginally better performance, benchmarks attached below in that same order with the following benchmark:
    ```ts
    const a = "#text"
    console.time("start")
    for(let i = 0; i< 1000000; i++) {
      const inc = ['#comment', '#text'].includes(a)
    }
    console.timeEnd("start")
    console.time("stop")
    for(let i = 0; i< 1000000; i++) {
      const inc  = a === "#comment" || a === "#text"
    }
    console.timeEnd("stop")
    ```
    Browser:
    
    image
    
    Node:
    
    image

- Reduced and reworked iterations
   Originally the main change in this PR, many parts in nuxt involve some sort of array/object manipulations, and in some places they are either repeated, redundant, or simply costly in terms of performance. The main examples of such cases are the small side PR's I have submitted like #24717, #24730,  and numerous others. I couldn't get all of it at the same time so I have decided to group the rest here, and also handle the iteration logic as opposed to mere count.
   
 - Reduced indexes:
    Some places in nuxt include working with indexes, like the `length` property of array/strings. While that alone isn't expensive, it still makes that repeated index, which especially with bigger inputs (like in HTML parsing or AST traversal) will gradually cost more and more microseconds, and might eventually become slightly noticeable in page performance.
    
- Early returns
   Similar to #25647, there are some places where some iterations are made on potentially empty values, which increase the overhead and therefore decrease performance. This PR adds guards to ensure there won't be iterations over an empty value.
   
   
 - Avoiding amortized operations where possible
     Methods like `Array.prototype.push` have an amortized complexity (O(1) in the case of `push`), which for the most part would serve most use cases well. But if we want to really fine-tune performance and take care of that amortization, we can remove them in favor of plain complexity if we know the size of the array. In multiple places around the Nuxt source, there are usages of push where the exact array size is known. This PR uses a hybrid approach that refactors said parts with known sizes by using the `new Array(size)` constructor with direct index modification, which would result in plain O(1) complexity, potentially outperforming the amortized O(1) complexity of `push`, while keeping push for dynamic arrays, variable sizes, and other parts where the size cannot be accurately obtained without causing increased time complexity (like `Object.keys` to obtain the size of an object).

### 📝 Checklist


GalacticHypernova (commented):
Notes to self:

GalacticHypernova:
An important note:
The added commits are due to me forgetting to flush the original patch-9 branch when I was done with that one. That's my bad, however I used the origin's files in every conflict so it ended up just not changing anything (as can be seen in the files changed).

packages/nuxt/src/core/nitro.ts
```diff
@@ -291,7 +308,7 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
           const filteredRules = {} as Record<string, any>
           for (const routeKey in _routeRules[key]) {
             const value = (_routeRules as any)[key][routeKey]
-            if (['prerender', 'redirect', 'appMiddleware'].includes(routeKey) && value) {
+            if ((routeKey === 'prerender' || routeKey === 'redirect' || routeKey === "appMiddleware") && value) {
               if (routeKey === 'redirect') {
                 filteredRules[routeKey] = typeof value === 'string' ? value : value.to
               } else {
```
GalacticHypernova, line 314: I will rework this structure soon, there is a way to simplify it

GalacticHypernova (commented):
chore: attempt to revert to an earlier commit

GalacticHypernova:
@danielroe Just got back to this, I don't see anything that I changed that can cause the autofix.ci to error stating "appId" is not exported? Which file could I have messed that up in?

GalacticHypernova:
Hey @danielroe , any idea why it fails to import /public.svg? I haven't changed anything there..

GalacticHypernova:
I can keep finding more areas to improve and more files to change (already set my sights on a couple as of writing this message), but I know that it's impossible to get it all completely because new things are always being added and I find new ways to improve stuff every day, so in stead of keeping this a draft for eternity, I'll likely take care of a last few of the most important and most impactful areas, and it'll be ready for review 😊

packages/nuxt/src/app/composables/state.ts
```diff
@@ -44,7 +44,7 @@ export function clearNuxtState (
 ): void {
   const nuxtApp = useNuxtApp()
   const _allKeys = Object.keys(nuxtApp.payload.state)
-    .map(key => key.substring(useStateKeyPrefix.length))
+    .map(key => key.substring(2))
```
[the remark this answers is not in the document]
GalacticHypernova: This has since been fixed and used

packages/nuxt/src/core/nitro.ts
GalacticHypernova: This has been incorporated (somewhat)

packages/nuxt/src/core/app.ts
```diff
@@ -194,18 +197,14 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
...
       ...(config.plugins || []),
       ...config.srcDir
         ? await resolveFiles(config.srcDir, [
-          `${pluginDir}/*{${nuxt.options.extensions.join(',')}}`,
-          `${pluginDir}/*/index{${nuxt.options.extensions.join(',')}}`,
+          `${pluginDir}/*{${extensions}}`,
+          `${pluginDir}/*/index{${extensions}}`,
         ])
         : [],
     ].map(plugin => normalizePlugin(plugin as NuxtPlugin)))
```
GalacticHypernova, line 210: Question: in L222-L223, is that even necessary? Can we not call them just once in L238-L239?

GalacticHypernova:
Accidentally hit the wrong button, sorry about that!

GalacticHypernova:
I believe as an initial refactor round this is good enough. It would definitely serve as good foundation for a more targeted refactor round that will involve structural changes (there were already a few in this one). Please let me know if there's anything you think I should add/change/revert!

GalacticHypernova:
Actually, just realized an area that could potentially **really** benefit from a refactor, looking into it.

packages/ui-templates/lib/render.ts
```diff
@@ -82,19 +82,19 @@ export const RenderPlugin = () => {
         // Serialize into a js function
         const chunks = html.split(/\{{2,3}[^{}]+\}{2,3}/g).map(chunk => JSON.stringify(chunk))
         const hasMessages = chunks.length > 1
-        let templateString = chunks.shift()
+        const templateString = [chunks.shift()]
```
danielroe, line 85: I'm not sure this is a more performant approach... Let's skip this particular change for now.

packages/vite/src/vite.ts
```diff
@@ -103,8 +103,14 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
             composables: nuxt.options.optimization.keyedComposables,
           }),
           replace({
-            ...Object.fromEntries([';', '(', '{', '}', ' ', '\t', '\n'].map(d => [`${d}global.`, `${d}globalThis.`])),
```
danielroe, line 106: I would prefer to keep this as is, though it would be a nice improvement to split it out of the function.

packages/ui-templates/lib/render.ts
GalacticHypernova: Sure thing! Generally speaking pushing to an array has a better space complexity than concataneting strings because each concatenation creates a new string instance, whereas pushing to an array uses the same instance, so the memory usage in this specific execution, especially if templateString is iterate over a lot, would be far lower for the array method than the string method. But sure, it can be kept as is until the more impactful areas are taken care of

danielroe: Understood. TIL. However this is build time code (while bundling up templates) and never impacts a user. I am happy to keep it as is.

packages/vite/src/vite.ts
GalacticHypernova: In my opinion, the question that should be asked here is whether or not this will change. Are there dynamic replacements planned to be added? If it's all static, this method doesn't hurt readability, it doesn't use iterations, and doesn't create intermediate arrays. I'll leave the decision up to you of course, but in my opinion this is an example of unmandatory iterations  that would just decrease runtime performance and increase memory usage (albeit, a little, but increase regardless) with no reasonably equivalent benefits.

packages/ui-templates/lib/render.ts
GalacticHypernova: Sounds good to me!

packages/schema/src/config/experimental.ts
```diff
@@ -89,7 +89,12 @@ export default defineUntypedSchema({
       async $resolve (val, get) {
         // TODO: remove in v3.10
         val = val ?? await (get('experimental') as Promise<Record<string, any>>).then((e: Record<string, any>) => e?.inlineSSRStyles)
-        if (val === false || (await get('dev')) || (await get('ssr')) === false || (await get('builder')) === '@nuxt/webpack-builder') {
+        const [dev, ssr, builder] = await Promise.all([
```
danielroe, line 92: this seems less performant in the most common case (running in dev mode) as previously we short-circuited if any of those values were true.

GalacticHypernova: True, this might be a bit less performant in best case scenario, but it's difficult to make it more performant for the worst case scenario without sacrificing a tiny bit. Due to it being parallelized, the potential impact in best case justified the potential improvement in worst case

GalacticHypernova: Alright, the benchmark seems to show that the second approach is much slower. (difference 10 seconds). That was my mistake, I apologize. Here's the benchmark I used:
```ts
async function get(key) {
  return new Promise((resolve) => {
    setTimeout(() => {
      const values = {
        dev: Math.random() > 0.5,
        ssr: Math.random() > 0.5,
        builder: '@nuxt/webpack-builder'
      };
      resolve(values[key]);
    }, 100);
  });
}

async function snippet1(val) {
  if (val === false || (await get('dev')) || (await get('ssr')) === false || (await get('builder')) === '@nuxt/webpack-builder') {
    return true;
  }
  return false;
}

async function snippet2(val) {
  const [dev, ssr, builder] = await Promise.all([
    get('dev'),
    get('ssr'),
    get('builder'),
  ]);
  if (val === false || dev || ssr === false || builder === '@nuxt/webpack-builder') {
    return true;
  }
  return false;
}

async function benchmark() {
  const iterations = 100;
  let start, end;
  
  console.time("e")
  for (let i = 0; i < iterations; i++) {
    await snippet1(false);
  }
  console.timeEnd("e")
  
  console.time("F")
  for (let i = 0; i < iterations; i++) {
    await snippet2(false);
  }
  console.timeEnd("F")
}
```
image

GalacticHypernova: Actually, when testing with val = true, I get that the new version is almost twice as fast:
image

GalacticHypernova: I pushed an update that combines the best of both worlds. If val is false it will immediately return false before awaiting the Promise.all, as per the results of the benchmark when val is set to false.
Otherwise it awaits promise.all and handles everything else. This actually doesn't impact the best case scenario at all, when dev is true
image

And it does make sense when I think about it, considering Promise.all parallelizes it, so each promise is handled in a different core, separate from the rest. There is no dependency between each promise so there's nothing to cause slowdown. Of course, as the if conditions progress to the next, the performance difference becomes much bigger (for example, here it is when dev is hardcoded to false)
image

GalacticHypernova:
Marking as draft to resolve type issues

GalacticHypernova:
Marking this as ready for review. The type issues don't appear to be related to a change of mine, seeing as it also impacts a production app I have:
image

(Cropped image to show that this is, in fact, not a blank project)

packages/schema/src/config/experimental.ts
danielroe: I don't think your benchmark is right. It has an arbitrary delay of 100ms for any `get` operation, which is not accurate, and doesn't account for the fact that this is only running _once_ when starting Nuxt.

In general, I would rather not waste too much time optimising something that only runs once if it adds complexity to the code.

And in the case I'm not sure it's the right 'happy-path' because the vast majority of calls to resolve the schema is running the dev server.

GalacticHypernova:
I'm curious, why not just make all the changes here and then merge as a whole? It would be less time consuming compared to making multiple pr's, which I'm sure you have way more important things to invest that time on

danielroe:
I find it easier to review in smaller batches, and there's less risk of merge conflicts.

Plus, there are a lot of judgement calls here, where I need to edit the code - it's not a straightforward single feature or change to review.

GalacticHypernova:
I see. Well, all I can offer is my help, if you would like me to make any changes so you won't have time to spend time making them I'll happily make them

PR #26132, opened 2024-03-07
fix: provide a valid useId type signature

ghost:
### ❓ Type of change

### 📚 Description

Current implementation gives a TypeScript error when I try to pass in a key as parameter, just to then throw at runtime because the key is not provided 😬

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/id.ts
+++ b/packages/nuxt/src/app/composables/id.ts
@@ -8,7 +8,7 @@ const SEPARATOR = '-'
 /**
  * Generate an SSR-friendly unique identifier that can be passed to accessibility attributes.
  */
-export function useId (key: string): string {
+export function useId (key: string = 'nuxt'): string {
   if (typeof key !== 'string') {
     throw new TypeError('[nuxt] [useId] key must be a string.')
   }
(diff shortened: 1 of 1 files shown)


danielroe:
The key should be provided by the Nuxt compiler, which is why it is disallowed. Can you open an issue with a reproduction instead? 🙏

Is there a reason you want to provide the key instead of using the auto generated one?

remonke:
I wanted to use this composable with Radix Vue. Unfortunately it throws an error when prefix is not provided.

https://stackblitz.com/edit/nuxt-starter-he5ner?file=nuxt.config.ts%2Cpages%2Fno-prefix.vue%2Capp.vue%2Cpages%2Fprefix.vue

danielroe:
I think you could do something like this:

```vue
<template>
  <ConfigProvider :use-id="providedUseId">
    <TooltipProvider>
      <TooltipRoot>
        <TooltipTrigger class="IconButton"> Hover me </TooltipTrigger>
        <TooltipPortal>
          <TooltipContent as-child class="TooltipContent" :side-offset="5">
            Add to library
            <TooltipArrow class="TooltipArrow" size="8" />
          </TooltipContent>
        </TooltipPortal>
      </TooltipRoot>
    </TooltipProvider>
  </ConfigProvider>
</template>

<script setup>
const nuxtApp = useNuxtApp()
const providedUseId = () => nuxtApp.runWithContext(() => useId())
</script>
```

PR #26172, opened 2024-03-10
perf(nuxt): use faster approach to check cache exists

Mini-ghost:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

Currently, the implementation of `hasCachedData` inside `useAsyncData` is as follows:

```ts
const hasCachedData = () => ![null, undefined].includes(options.getCachedData!(key) as any)
```

Perhaps we can tune it for better performance by simplifying the condition like this:

```ts
const hasCachedData = () => options.getCachedData!(key) != null
```

For reference, here are the benchmark test results:

```
check via `value != null` x 993,997,209 ops/sec ±0.10% (98 runs sampled)
check via `![null, undefined].includes(value)` x 74,629,930 ops/sec ±0.35% (98 runs sampled)
Fastest is check via `value != null`
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -243,7 +243,7 @@ export function useAsyncData<
     console.warn('[nuxt] `boolean` values are deprecated for the `dedupe` option of `useAsyncData` and will be removed in the future. Use \'cancel\' or \'defer\' instead.')
   }
 
-  const hasCachedData = () => ![null, undefined].includes(options.getCachedData!(key) as any)
+  const hasCachedData = () => options.getCachedData!(key) != null
 
   // Create or use a shared asyncData entity
   if (!nuxtApp._asyncData[key] || !options.immediate) {


danielroe (approved):

TheAlexLichter (approved):

TheAlexLichter (changes requested):
What if `0`, `false` or the like are returned? 🤔

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -243,7 +243,7 @@ export function useAsyncData<
     console.warn('[nuxt] `boolean` values are deprecated for the `dedupe` option of `useAsyncData` and will be removed in the future. Use \'cancel\' or \'defer\' instead.')
   }
 
-  const hasCachedData = () => ![null, undefined].includes(options.getCachedData!(key) as any)
+  const hasCachedData = () => options.getCachedData!(key) != null
```
manniL, line 246:
```suggestion
  const hasCachedData = () => {
    const data = options.getCachedData!(key)
    return data !== null && data !== undefined
 }
```

Maybe explicitly check null and undefined instead?

Mini-ghost: In fact, `value != null` is used in some places inside Nuxt. I think this perhaps clear enough to identify `undefined` or `null` 🤔

manniL: But what if the cached data should return `false`, `0` or `''`? 🤔

Mini-ghost: However, it's worth noting that encountering a `null` or `undefined` return value will trigger a fetch, as clearly outlined in the documentation. May I suggest that adhering to this established behavior might be the preferable approach?

https://github.com/nuxt/nuxt/blob/8bb8d824c73a7b2e4fcc4f0b870e6e036318e6a6/packages/nuxt/src/app/composables/asyncData.ts#L59-L64

danielroe: @manniL I also had to check, but I think it's good:

CleanShot 2024-03-10 at 10 23 02@2x

manniL: 🤦  I think I accidentally checked for `!= undefined` in my tests.

 LGTM!

TheAlexLichter (approved):

PR #26191, opened 2024-03-11
refactor(nuxt): simplify `request` computation inside `useFetch`

Mini-ghost:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

Currently, the implementation of `_request` inside `useFetch` is as follows:

https://github.com/nuxt/nuxt/blob/ac54da242f47c466372017be0573777857d374a6/packages/nuxt/src/app/composables/fetch.ts#L97-L103

This code is almost equivalent to `toRef(request)`, should we just use `toRef` as a cleaner alternative?

I am open to feedback regarding whether this adjustment will yield significant benefits. Should the evaluation indicate that the drawbacks outweigh the benefits, please feel free to close this PR.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -1,7 +1,7 @@
 import type { FetchError, FetchOptions } from 'ofetch'
 import type { NitroFetchRequest, TypedInternalResponse, AvailableRouterMethod as _AvailableRouterMethod } from 'nitropack'
 import type { MaybeRef, Ref } from 'vue'
-import { reactive, toValue, toRef } from 'vue'
+import { reactive, toRef, toValue } from 'vue'
 import { hash } from 'ohash'
 
 import { useRequestFetch } from './ssr'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -1,7 +1,7 @@
 import type { FetchError, FetchOptions } from 'ofetch'
 import type { NitroFetchRequest, TypedInternalResponse, AvailableRouterMethod as _AvailableRouterMethod } from 'nitropack'
 import type { MaybeRef, Ref } from 'vue'
-import { computed, reactive, toValue } from 'vue'
+import { computed, reactive, toValue, toRef } from 'vue'
```
danielroe, line 4:
```suggestion
import { reactive, toValue, toRef } from 'vue'
```

Mini-ghost:
I'd like to understand if using a getter function as the parameter for `request`, especially when the calculations are very complex, could actually worsen performance. This is because every time `_request` is accessed, the getter function would be executed again .

Of course, for users passing in a type of `Ref<ReqT> | ReqT`, this should not pose a problem.

danielroe (approved):

danielroe:
Does `toRef` re-execute the function every time `.value` is accessed?

Mini-ghost:
I attempted to write and execute the following code:

```ts
const text = toRef(() => {
  console.log('accessed test');
  return 'test';
});

console.log(text.value);
console.log(text.value);
console.log(text.value);
```

I accessed `text.value` three times, and it resulted in the following output:

accessed `text.value` three times

Therefore, using a getter indeed causes multiple executions.

danielroe (approved):

PR #26193, opened 2024-03-11
fix(nuxt): improved plugin annotating warnings

harlan-zw:
### ❓ Type of change

### 📚 Description

I had a valid use case for needing runtime meta data for a plugin, when making it, a cryptic warning was being thrown 3 times. 

```sh
WARN  Could not resolve /home/harlan/packages/nuxt-seo/src/runtime/nuxt/plugin/defaults.ts.
WARN  Could not resolve /home/harlan/packages/nuxt-seo/src/runtime/nuxt/plugin/defaults.ts.
WARN  Could not resolve /home/harlan/packages/nuxt-seo/src/runtime/nuxt/plugin/defaults.ts.
```

I honestly had no idea what it was meant to mean and could only debug it by looking through the Nuxt source.

This PR aims to make this warning easier to interpret.

New warnings:

```sh
WARN  Failed to parse static properties from plugin plugins/sample.ts, falling back to non-optimized runtime meta. Learn more: https://nuxt.com/docs/guide/directory-structure/plugins#object-syntax-plugins
```

OR

```sh
WARN  Failed to extract meta data from plugin ../../../src/runtime/nuxt/plugin/defaultsWaitI18n. ENOENT: no such file or directory, open '/home/harlan/packages/nuxt-seo/src/runtime/nuxt/plugin/defaultsWaitI18n'
```

Note: I did try and change it to only throw once by making sure we don't annotate code blocks that have already been parsed but ran into too many errors from the test suite. Not too sure what went wrong.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -205,9 +205,9 @@ export async function annotatePlugins (nuxt: Nuxt, plugins: NuxtPlugin[]) {
         ...await extractMetadata(code),
         ...plugin
       })
-    } catch (e: Error) {
+    } catch (e) {
       const relativePluginSrc = relative(nuxt.options.rootDir, plugin.src)
-      if (e.message === 'Invalid plugin metadata') {
+      if ((e as Error).message === 'Invalid plugin metadata') {
         logger.warn(`Failed to parse static properties from plugin \`${relativePluginSrc}\`, falling back to non-optimized runtime meta. Learn more: https://nuxt.com/docs/guide/directory-structure/plugins#object-syntax-plugins`)
       } else {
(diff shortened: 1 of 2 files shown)


danielroe (approved):

harlan-zw:
Not directly related to this PR but I think if we plan to stop supporting runtime meta in plugins we should include when it will no longer be supported. The docs are quite ambigious

> he properties may be statically analyzed in future to produce a more optimized build. So you should not define them at runtime.

danielroe:
"you should not define them at runtime" is meant to be quite explicit but I would welcome any better wording 🙏

harlan-zw:
I meant the first bit "the properties may be statically analyzed in future", aren't they currently being analyzed?

danielroe:
Good point! Yes. 👌

Issue #26102, opened 2024-03-06
`useRequestURL()` doesn't provide the correct hostname in when `swr` is enabled (route rules)


bpolaszek:
### Environment

- Operating System: `Darwin`
- Node Version:     `v21.6.2`
- Nuxt Version:     `3.10.3`
- CLI Version:      `3.10.1`
- Nitro Version:    `2.9.1`
- Package Manager:  `yarn@1.22.21`
- Builder:          `-`
- User Config:      `ssr`, `pages`, `devtools`, `runtimeConfig`, `routeRules`, `sourcemap`, `modules`, `vite`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.11.4`, `nuxt-icon@0.6.8`, `nuxt-schema-org@3.3.4`, `@pinia/nuxt@0.5.1`
- Build Modules:    `-`

### Reproduction

```
# /etc/hosts
127.0.0.1    example.local
```

```vue
// app.vue
<template>
  {{ hostname }}
</template>

<script setup>
const {hostname} = useRequestURL()
</script>
```

```console
yarn dev --host 127.0.0.1
```

If you visit `http://example.local:3000`, `example.local` will be displayed (from SSR).

### Describe the bug

Now, add an [`swr` directive](https://nuxt.com/docs/guide/concepts/rendering#hybrid-rendering) to add cache headers: 
```ts
// nuxt.config.ts
export default defineNuxtConfig({
  // ...
  routeRules: {
    '/**': {
      swr: 3600,
    },
  },
})
```

Then do it again:
```console
yarn dev --host 127.0.0.1
```

If you visit `http://example.local:3000`, `localhost` will be displayed instead of `example.local` (from SSR).

danielroe:
Would you see if you can reproduce this in pure Nitro ([reproduction sandbox](https://stackblitz.com/github/unjs/nitro/tree/main/examples/hello-world)), and if so, raise [there](https://github.com/unjs/nitro/issues/new/choose)? 🙏

pi0:
Hi, dear @bpolaszek. Yes, and it is an expected limitation you are facing.

When you opt-in to a caching mechanism, we have to make sure the "same" cache response is served to every request hitting this URL. And therefore it should not "vary" based on incoming headers.

By default Nitro removes all incoming headers as a safety measure to avoid cache leaks. 

You can add varies support for `header` host (but consider `useRequestURL` depends on **more headers** including `x-forwarded-*` ones for full support)

```js
[...]

PR #26197, opened 2024-03-11
docs(nuxt): add cache.varies docs for multi-tenant use case

adamdehaven:
### ❓ Type of change

### 📚 Description

`useRequestURL` will return `localhost` for the `host` header when utilizing cache strategies, which isn't completely obvious. (It took me 3 days to figure out why my `host` headers were returning `localhost` 😅 )

The `cache.varies` option for Nitro's `routeRules` needs to be better documented to help understand the cache behavior and discarded headers, especially related to a multi-tenant implementations.

The explanation is not currently obvious when reading through the docs; however, was better explained in https://github.com/nuxt/nuxt/issues/26102#issuecomment-1980625833 by @pi0. This PR surfaces the explanation and provides docs visitors a keyword of "tenant" in the docs they can search for as a hint. 

I also created a corresponding PR for Nitro: https://github.com/unjs/nitro/pull/2241

### 📝 Checklist

### DIFF
--- a/docs/3.api/2.composables/use-request-url.md
+++ b/docs/3.api/2.composables/use-request-url.md
@@ -10,6 +10,12 @@ links:
 
 `useRequestURL` is a helper function that returns an [URL object](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) working on both server-side and client-side.
 
+::important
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
Thank you! And sorry for not getting back to you sooner on Twitter 🙏

adamdehaven:
No worries! 

It does still seem a bit odd that I have to pass in the host to a custom header when calling a Nuxt-defined server endpoint; calling `getRequestHeader(event, 'host')` or `getRequestURL(event).host` still return `localhost`. Is that expected? 

Also trying to find the best place to store the host once I get it the first time to be able to access it in future requests instead of having to parse it via middleware, etc.

PR #26204, opened 2024-03-11
feat(nuxt): allow generating metadata for nuxt components

### LINKED ISSUE #15015: Retrieve auto-imported component / function names
### Describe the feature

Hi, I'm not sure if this is a feature request or just a question, but I didn't find any related issues / docs.

**Context:** 

I have an ESLint configuration and it contains rules [`no-undef`](https://eslint.org/docs/latest/rules/no-undef) and [`vue/no-undef-components`](https://eslint.vuejs.org/rules/no-undef-components.html) that report undeclared functions/variables/components.

some-component vue - cardplace local_220925161713

Nuxt automatically imports many things and I like it, I don't want to import them explicitly.
[This is slightly related issue](https://github.com/nuxt/nuxt.js/issues/13413), but I don't need to import them explicitly, I need an option to retrieve a **names** of all auto-imported components/functions to declare them as global in `.eslintrc` file.

**My current workaround** is define them manually and declare as global like that:

● Untitled-1 - cardplace local_220925163652

Although this is a working solution, there are some limitations:
- It needs to be filled manually
- It doesn't update automatically if I register a new module that adds a global component (such as `@nuxt/image`) or new composables (such as `@vueuse/nuxt`)
- It doesn't update if I set custom dirs to import in `nuxt.config.ts` via `imports.dirs` and create a new file inside it.
- It doesn't update if new release of Vue/Nuxt includes a new features

So, a lot of manual work every time something was changed.
This can be done using custom module, but I believe it should be done by Nuxt itself.

**Feature request:**

Expose an auto-imported functions / components names to allow import them and mark as global in `.eslintrc`

● Untitled-1 - cardplace local_220925170128

**Additional info**

I know it's recommended to disable `no-undef` for TypeScript at all, but currently, as far I know, there is no robust way to separate `<script lang="ts">` and `<script>` in ESLint confgurations for `.vue` files. 

Also Nuxt isn't only TypeScript framework. and global components still needs to be configured

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15015

### ❓ Type of change

### 📚 Description

This adds a new option: `components.generateMetadata` which writes `~/.nuxt/components.json` which could be consumed by other integrations to get information about the auto-imported components in your app.

The original use case of this (eslint) will likely be implemented differently so it may be worth reconsidering this and/or closing the PR. What do you think @antfu?

### 📝 Checklist

### DIFF
--- a/docs/3.api/5.kit/5.components.md
+++ b/docs/3.api/5.kit/5.components.md
@@ -41,6 +41,11 @@ interface ComponentsDir {
   transpile?: 'auto' | boolean
 }
 
+// You can augment this interface (exported from `@nuxt/schema`) if needed
+interface ComponentMeta {
+  [key: string]: unknown
+}
+
 interface Component {
   pascalName: string
   kebabName: string
@@ -54,6 +59,7 @@ interface Component {
   island?: boolean
   mode?: 'client' | 'server' | 'all'
   priority?: number
+  meta?: ComponentMeta
 }
 ```
 
--- a/packages/kit/src/components.ts
+++ b/packages/kit/src/components.ts
@@ -48,6 +48,7 @@ export async function addComponent (opts: AddComponentOptions) {
     mode: 'all',
     shortPath: opts.filePath,
     priority: 0,
+    meta: {},
     ...opts
   }
 
--- a/packages/nuxt/src/components/module.ts
+++ b/packages/nuxt/src/components/module.ts
@@ -5,7 +5,7 @@ import type { Component, ComponentsDir, ComponentsOptions } from 'nuxt/schema'
 
 import { distDir } from '../dirs'
 import { clientFallbackAutoIdPlugin } from './client-fallback-auto-id'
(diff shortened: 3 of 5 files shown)


antfu:
If we end up recommending the eslint config generation approach (https://github.com/nuxt/eslint-config/pull/332), where the module can directly grab the necessary data from module hooks, I guess we don't need core to support that - unless there would be other cases to consume those data

pi0:
I think it makes sense only if we can make it opt-in and expandable for component extracted meta (similar to [nuxt-component-meta](https://www.npmjs.com/package/nuxt-component-meta)). 

The advantage of being in core is that there is a unified protocol to augment and access component data and it is probably valuable.

pi0:
additional `meta` object to be expandable with hooks sound a nice idea. Also for sure yes being opt-in makes sense.

PR #26223, opened 2024-03-13
feat(nuxt): enable component islands when server pages/components are present

danielroe:
### ❓ Type of change

### 📚 Description

This enables components whenever there are components with `mode: server` or server pages. The setting is 'auto'. (It won't affect bundle size when these aren't used.)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -374,10 +374,13 @@ export const nuxtConfigTemplate: NuxtTemplate = {
       baseURL: undefined,
       headers: undefined
     }
+    const shouldEnableComponentIslands = ctx.nuxt.options.experimental.componentIslands && (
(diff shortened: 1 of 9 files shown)


packages/nuxt/src/core/templates.ts
```diff
@@ -374,10 +374,13 @@ export const nuxtConfigTemplate: NuxtTemplate = {
       baseURL: undefined,
       headers: undefined
     }
+    const shouldEnableComponentIslands = ctx.nuxt.options.experimental.componentIslands && (
+      ctx.nuxt.options.dev || ctx.nuxt.options.experimental.componentIslands !== 'auto' || ctx.app.pages?.some(p => p.mode === 'server') || ctx.app.components?.some(c => c.mode === 'server')
```
huang-julien, line 378: `ctx.nuxt.options.experimental.componentIslands !== 'auto'` we can set `false` to `componentIsland`, not sure if anyone does that but it can enable island

danielroe: I think the first check prevents that - `ctx.nuxt.options.experimental.componentIslands &&`

huang-julien: oh indeed !

huang-julien (approved):
:heart:

PR #21934, opened 2023-07-04
fix(nuxt): regenerate imports after template

antfu:
### 🔗 Linked issue

fixes #18897

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #26249, opened 2024-03-14
fix(nuxt): auto imports regeneration timing

antfu:
### ❓ Type of change

### 📚 Description

In the review comments

### 📝 Checklist


packages/nuxt/src/imports/module.ts
```diff
@@ -119,22 +131,20 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
       if (composablesDirs.some(dir => dir === path || path.startsWith(dir + '/'))) {
-        await updateTemplates({
-          filter: template => templates.includes(template.filename)
-        })
+        await regenerateImports()
       }
     })
 
-    nuxt.hook('app:templatesGenerated', async () => {
-      await regenerateImports()
```
antfu, line 137: Previously we did regeneration (rescan) after all templates to avoid race conditions in #21934. But since unimport also use templates to generate those dts etc, so it means we missed the change to update them - until this get triggered - we are always one "tick" behind.

This PR makes sure `regenerateImports` always update (only) the unimport templates. Thus we also need some mechanism to dedupe and avoid circular update

packages/nuxt/src/imports/module.ts
```diff
@@ -119,22 +131,20 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
+    // Watch for template generation
+    nuxt.hook('app:templatesGenerated', async (_app, templates) => {
+      // Filter out templates of imports to avoid circular regeneration
+      const nonDtsTemplates = templates.filter(t => !isImportsTemplate(t))
```
danielroe, line 144: maybe .some instead?

danielroe (approved):

PR #26259, opened 2024-03-14
feat(nuxt): add `clear` function to async data

### LINKED ISSUE #26195: Add `reset` fn to data fetching composables
### Describe the feature

In specific scenarios, it might be handy to reset the state and cancel any pending promises of a declared data fetching composable. 

For example, it would come useful when we declare a data fetching composable inside a dialog which remains in the DOM when closed AND yet we would like to conveniently reset everything inside a dialog to prepare it for when it becomes opened by the user again. Otherwise, I need to reset everything by hand.

An alternative approach is to pass a key explicitly and use it when calling `clearNuxtData`:

```
const key = '...'
const fn = useAsyncData(key, ...)
const isDialogOpen = defineModel<boolean>({ required: true })

whenever(
  () => !isDialogOpen.value,
  {
    clearNuxtData(key)
  }
)
```

### Additional information

### Final checks

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26195

### ❓ Type of change

### 📚 Description

Allows clearing Nuxt data from within the async data composable (hence removing a need for knowing and/or passing a key). Original issue describes _resetting_ Nuxt data instead (by respecting the optional `default` prop rather than always setting `data.value` to `undefined`). Thought this might be better to start with though; I'm open to any suggestions!

Will update the docs knowing this is a desirable addition.

### 📝 Checklist


manniL:
Seems like a nice DX improvement to me 👍 

With regards to `clear` (undefined) vs. `reset` (set to default value) - I was wondering if we need a distinction or could rely on resetting it to the default value straight away. But if that should be consistent, `clearNuxtData` should behave the same way, which would be tricky at the moment 🤔

DamianGlowala:
I certainly agree - this raises a question whether `clearNuxtData` should respect the default value. We have a couple of options:

- make `clearNuxtData` respect the default value (I personally prefer this behavior; would be a breaking change though)
- add `reset` function to async data which would differ from `clear` (setting it to the default value vs `undefined`)
- add optional param to `clear` function which would allow either reverting to the default value or not (again, some could ask why we wouldn't have the same param for `clearNuxtData`)

manniL:
I think both, clear and reset, might be confusing

Maybe updating it for v4 could be nice? 🤔

DamianGlowala:
They will be confusing, I'm sure.

Created https://github.com/nuxt/nuxt/issues/26269 to discuss updating it for v4.

danielroe (approved):

docs/1.getting-started/6.data-fetching.md
```diff
@@ -150,6 +150,7 @@ Read more about `useAsyncData`.
 - `data`: the result of the asynchronous function that is passed in.
 - `pending`: a boolean indicating whether the data is still being fetched.
 - `refresh`/`execute`: a function that can be used to refresh the data returned by the `handler` function.
+- `clear`: a function that can be used to empty the state of `data` (and prevent any currently pending requests from updating it)
```
DamianGlowala, line 153: `data` on its own might be confusing as the scope is broader than one property, WDYT? What about '... to reset the state of the data request (and cancel any currently pending requests)'?

danielroe: we could be even more explicit:

```suggestion
- `clear`: a function that can be used to set `data` to undefined, set `error` to null, set `pending` to false, set `status` to `idle`, and mark any currently pending requests as cancelled
```

DamianGlowala: Perfect. Shall we add backticks to undefined, null, and false? Also, there's probably less value in telling the user about _marking_ as cancelled, and more in putting an emphasis on the actual result: '... and cancel any currently pending requests'.

danielroe: happy either way. the reason I wasn't wanting to do that as it doesn't actually stop them running, and a few queries have come in previously by people who want somehow to 'revoke' a promise and stop its execution

DamianGlowala: Totally understandable, let's keep your version for clarity 👍

PR #26289, opened 2024-03-15
fix(nuxt): use build plugin to detect usage of `<NuxtPage>` and `<NuxtLayout>`

### LINKED ISSUE #25912: "NuxtPage / NuxtLayout component has not been used" warn when NuxtLayout async show up
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.2
- CLI Version:      3.10.1
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-2vasos?file=app.vue

### Describe the bug

I used `<NuxtPage />` and `<NuxtLayout />` components in the below way:
```vue
    <div v-if="loading">loading...</div>
    <NuxtLayout v-else>
      <NuxtPage />
    </NuxtLayout>
```

But it shows up the below warning in console log.
```
[nuxt] Your project has pages but the `<NuxtPage />` component has not been used. You might be using the `<RouterView />` component instead, which will not work correctly in Nuxt. You can set `pages: false` in `nuxt.config` if you do not wish to use the Nuxt `vue-router` integration.
[nuxt] Your project has layouts but the `<NuxtLayout />` component has not been used.
```

### Additional context

My purpose is fetching some environment data from server before NuxtLayout and NuxtPage rendered and shows a loading layout.

ps. I disabled the `ssr` options.

### Logs

_No response_

IonianPlayboy:
### 🔗 Linked issue

Fixes #25912

### ❓ Type of change

#### The initial problem

Currently, the `check-if-layout-used` and `check-if-page-unused` plugins check if their respective component has been mounted when the Nuxt App is ready. (with `onNuxtReady()`). If this is not the case and the project has layouts/pages, they log a message in the console to warn the user to make them aware they should probably make some changes. (i.e. use `NuxtLayout`/`NuxtPage` in their app, or change their config)

This approach works well in most cases, but it raises a false positive if the `NuxtLayout`/`NuxtPage` components are not mounted right away, which can happen if there is a loading state that needs to be resolved before switching to the `NuxtLayout`/`NuxtPage` display. In that case, checking if the component is mounted is not enough to guarentee that it is not used in the app.

#### The idea

The best way to be sure that `NuxtLayout`/`NuxtPage` are indeed used in the user app, conceptually, is to check if they are imported in the user code. Since Nuxt is also responsible for configuring Vite/Webpack to load the app dependencies/modules, it makes sense to make use of it by checking if at some point Vite/Webpack has to resolve the `NuxtLayout`/`NuxtPage` components to determine if they are used or not.

However, I found it quite tricky to relay this information from the `loader` plugin all the way to the `NuxtApp` : I could not add new properties to the `NuxtApp` from the components module (which is where components are resolved for the whole app), and I could not access the `Nuxt` instance from the  `check-if-layout-used` and `check-if-page-unused` plugins to hook into the Vite/Webpack config.

#### The solution

I ended up creating two virtual files with `addTemplate` to carry the data between the Nuxt and the NuxtApp instances:

- `detected-nuxt-layout-usage.mjs`, which exports a const named `isNuxtLayoutUsed`
- `detected-nuxt-page-usage.mjs`, which exports a const named `isNuxtPageUsed`

They are only created in dev mode (since the warnings are only intended for development) and are updated when the Vite/Webpach loader detect that the corresponding component is being imported. I thought about putting both variables inside a single virtual file, but decided against it because I was worried about concurrency issues.

#### Final thoughts

I am not 100% sure this is the best way to fix this issue, but with my current knowledge of Nuxt's codebase this is the best I could do on my own. I am interested in learning more about it though, so if there is anything to improve or maybe another solution that would be better than my current fix, feedback would be appreciated. 😊

### 📝 Checklist


packages/nuxt/src/components/loader.ts
```diff
@@ -123,3 +126,12 @@ function findComponent (components: Component[], name: string, mode: LoaderOptio
   // and createServerComponent above
   return otherModeComponent
 }
+
+function detectComponentUsage (component: Component, detectedComponentsUsage: LoaderOptions['detectedComponentsUsage']) {
+  if (process.env.NODE_ENV !== 'development' || !(component.pascalName in detectedComponentsUsage)) { return }
```
danielroe, line 131: we would probably want to access `nuxt.options.dev` rather than NODE_ENV.

danielroe (commented):
This is a nice idea. We're getting enough false positives on the warning that I wonder if it would be worth instead refactoring our implementation to use a compile-time check _only_ rather than purely runtime.

The main issue here is that the NuxtLayout/NuxtPage might be detected _across_ a project but not necessarily in the particular route being rendered. (Nevertheless, this is might still be fine because we're trying to check for the 'easy' case where a user is unaware they need to use NuxtPage/NuxtLayout rather than the more advanced cases.)

We'd also probably to update the implementation to detect any usage of NuxtLayout/NuxtPage that is _imported_ (so not necessarily injected by components loader). It could be a separate vite plugin rather than adding into the existing component loader.

huang-julien (commented):
Should we remove the runtime `nuxtApp._isNuxtLayoutUsed` since we analyse imports ?

edit: woops didn't see @danielroe 's comment

IonianPlayboy:
> We're getting enough false positives on the warning that I wonder if it would be worth instead refactoring our implementation to use a compile-time check only rather than purely runtime.
> ...
> We'd also probably to update the implementation to detect any usage of NuxtLayout/NuxtPage that is imported (so not necessarily injected by components loader). It could be a separate vite plugin rather than adding into the existing component loader.

This makes perfect sense to me. I tried to stay as close to the existing implementation as possible with my first attempt, but I agree that we should have all the data needed to emit an accurate warning without any runtime logic. 

In the `NuxtPage` case, we could even add a check for a `RouterView` import to be even more precise in the warning message, since the main goal is to inform the user they should never have to use it in their Nuxt app.

I am going to try to update this PR to follow this new direction, and I'll report back if I encounter some issues or if I am stuck. 😊

IonianPlayboy:
I have updated the PR with the discussed changes. 😊

Things worth mentioning : 
- I could not rely on `app.layouts` to detect if they were layouts in the user project, since we are deliberately ignoring `index.vue` layout here. I was not able either to find a way to import the layouts from `#build/layouts` inside `core/nuxt.ts`, so I ended up adding a new property to the app in order to check if they were any layout in the project or not. If there is a better way to do this, let me know and I'll change it.
- The `hasBeenInitialized` check is necessary because when the `app:templates` run for the first time, the Vite/Webpack plugin has not been instancied yet - which means that we did not have the opportunity to detect anything yet. It works fine afterwards.
- This  Vite/Webpack plugin shares a lot of logic with the `ImportProtectionPlugin`. It might make sense to extract it into an `utils`function, but I was not sure if this was relevant or not, where this should live and how it should be called.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -197,8 +198,47 @@ async function initNuxt (nuxt: Nuxt) {
...
+      // Exclude top-level resolutions by plugins
+        join(nuxt.options.rootDir, 'index.html'),
+        // Keep only imports coming from the user's project (inside the rootDir)
+        new RegExp(`^(?!${escapeRE(nuxt.options.rootDir)}${escapeRE(sep)}).+[^\n]+$`)
+      ],
+      detectedComponents
+    }
+
+    addVitePlugin(() => DetectComponentUsagePlugin.vite(componentDetectionConfig))
+    addWebpackPlugin(() => DetectComponentUsagePlugin.webpack(componentDetectionConfig))
```
huang-julien, line 218: We could probably use `addBuildPlugin`

IonianPlayboy:
Ah, I just realized that this does not work without the `setTimeout` I was using in my playground to emulate loading, my bad. The `app:templates` hook is never triggered again in that case, which means that I never have the opportunity to check the component usage from `core/nuxt.ts`. I guess I'll have to use a plugin to do a runtime check then, and I'll remove the `app._hasLayouts` property since I will be able to use `#build/layouts` here.

packages/nuxt/src/core/nuxt.ts
IonianPlayboy: That's a good idea, thanks for the suggestion.

This pattern is repeated alot across the codebase (`addVitePlugin` + `addWebpackPlugin` instead of `addBuildPlugin`), maybe a future PR could refactor them to make plugin declaration consistent.

packages/nuxt/src/core/app.ts
```diff
@@ -123,6 +123,7 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   for (const config of layerConfigs) {
     const layoutDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.layouts || 'layouts'
     const layoutFiles = await resolveFiles(config.srcDir, `${layoutDir}/**/*{${nuxt.options.extensions.join(',')}}`)
+    app._hasLayouts = layoutFiles.length > 0
```
danielroe, line 126: I think you can rely on `app.layouts` later on instead - we do not support `~/layouts/index.vue` as a layout at all.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -197,8 +198,46 @@ async function initNuxt (nuxt: Nuxt) {
...
+    const detectedComponents = new Set<string>()
+    const cache = new Set<string>()
+
+    const componentDetectionConfig = {
+      rootDir: nuxt.options.rootDir,
+      exclude: [
+      // Exclude top-level resolutions by plugins
+        join(nuxt.options.rootDir, 'index.html'),
+        // Keep only imports coming from the user's project (inside the rootDir)
+        new RegExp(`^(?!${escapeRE(nuxt.options.rootDir)}${escapeRE(sep)}).+[^\n]+$`)
```
danielroe, line 212: `sep` will always be `/` in `pathe` - you can hard-code it.

packages/nuxt/src/core/plugins/detect-component-usage.ts
```diff
@@ -0,0 +1,39 @@
...
+  exclude?: Array<RegExp | string>
+  detectedComponents: Set<string>
+}
+
+export const DetectComponentUsagePlugin = (options: DetectComponentUsageOptions) => createUnplugin(() => {
+  const importersToExclude = options?.exclude || []
+
+  const detectComponentUsagePatterns: Array<[importPattern: string | RegExp, name: string]> = [
+    ['../packages/nuxt/src/pages/runtime/page', 'NuxtPage'],
+    ['../packages/nuxt/src/app/components/nuxt-layout', 'NuxtLayout']
```
danielroe, line 15: these paths will work in the framework repo + tests but won't work in a user project as they will be `dist` paths instead. We expose the `distDir` in `src/dirs` however so you could use that.

danielroe (commented):
Looking good!

IonianPlayboy:
@danielroe I have found out that my current approach is flaky after trying a couple of scenarios. I have made some changes to use a runtime plugin that checks component usage on `app:mounted`to make it more robust. Can I push my changes here ?

IonianPlayboy:
I have updated the PR to take the feedback into account and to put the detection in a runtime plugin with an `app:mounted` hook - this should be more robust than the previous approach. Let me know if there is anything to change. 🙂

test/fixtures/runtime-compiler/app.vue
IonianPlayboy, line 1: This was causing an issue with tests checking that the `runtime-compiler` app would render without any logs, since it was not using the `NuxtPage` component to display its routes.

Since this is the correct behavior and adding this `app.vue` to the fixture does not break any other test, I believe the best solution is to add this file to solve the logs related tests failing.

test/fixtures/basic/app.vue
IonianPlayboy, line 1: Same idea here, I think the best option is to add a `app.vue` that uses a `NuxtLayout` and a `NuxtPage` to avoid unwanted logs in testing.

danielroe: we shouldn't need this as nuxt injects a default `app.vue` with layout + page when no one is provided:

https://github.com/nuxt/nuxt/blob/4712e994112a97feba447b3edb4afb65e968ed63/packages/nuxt/src/pages/runtime/app.vue#L1-L5

This is probably an indication that the check isn't working correctly.

IonianPlayboy: Oh thanks for pointing that out, I'll take a look to understand why it was still logging the warning then.

IonianPlayboy:
@danielroe I found out why the logging happened in the test when it was not supposed to: the `exclude` patterns were filtering out `pages/runtime/app.vue`, since it's not coming from the user's project. I added an `include` pattern to specifically keep it, and it has been enough on my end to fix the issue. 😊

This should be the last thing that needed to be fixed, so this PR should now be ready to be merged, unless I missed something. 🙂

IonianPlayboy:
If there's anything I can do to help this PR get reviewed/merge, let me know. 😊

danielroe:
Don't worry - this is with us now to review. 🙏

packages/nuxt/src/core/plugins/detect-component-usage.ts
```diff
@@ -0,0 +1,43 @@
...
+  ]
+
+  return {
+    name: 'nuxt:detect-component-usage',
+    enforce: 'pre',
+    resolveId (id, importer) {
+      if (!importer) { return }
+      if (id[0] === '.') {
+        id = join(importer, '..', id)
+      }
```
danielroe, line 29: This logic doesn't quite look right... What about directories beginning with `.` - and if it's a relative path surely we could just do `join(importer, id)` rather than inserting an extra level up?

packages/nuxt/src/core/plugins/detect-component-usage.ts
```diff
@@ -0,0 +1,43 @@
...
+      const isExcludedImporter = importersToExclude.some(p => typeof p === 'string' ? importer === p : p.test(importer))
+      const isIncludedImporter = importersToInclude.some(p => typeof p === 'string' ? importer === p : p.test(importer))
+      if (isExcludedImporter && !isIncludedImporter) { return }
```
danielroe, line 32: Why do we need the include/exclude pattern? (If we do need it for some reason, we would probably need to add quite a few more directories, including all layer root directories, and the paths to all auto-imported components.)

rocketiscool:
In need of this, would love for it to get finished up.

PR #26315, opened 2024-03-17
fix(nuxt): useId() causing hydration-error when using value multiple times

ToBinio:
### ❓ Type of change

### 📚 Description

When using the value of `useId()` multiple times a hydration-error happens.

```vue
<script setup lang="ts">
const id = useId()
</script>

<template>
  <div :id="id">
    text
  </div>
  <div :id="id">
    text
  </div>
</template>
```

```
warn [Vue warn]: Hydration text content mismatch on 
<div id="id" data-v-inspector="playground/app.vue:6:3"> 
  - rendered on server: nE3KxD8SGyR-0
  - expected on client: nE3KxD8SGyR_0 
  at <App key=3 > 
  at <NuxtRoot>
```

this only occurs when using the ID multiple times.
working example:

```vue
<script setup lang="ts">
const id = useId()
</script>

<template>
  <div :id="id">
    text
  </div>
</template>
```

note:
I do not really understand the code so there might be a reason why a different separator is used.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/id.ts
+++ b/packages/nuxt/src/app/composables/id.ts
@@ -54,5 +54,5 @@ export function useId (key?: string): string {
   }
 
   // pure client-side ids, avoiding potential collision with server-side ids
-  return key + '_' + nuxtApp._id++
+  return key + SEPARATOR + nuxtApp._id++
 }


packages/nuxt/src/app/composables/id.ts
```diff
@@ -54,5 +54,5 @@ export function useId (key?: string): string {
   }
 
   // pure client-side ids, avoiding potential collision with server-side ids
-  return key + '_' + nuxtApp._id++
+  return key + SEPARATOR + nuxtApp._id++
```
danielroe, line 57: The reason is so that purely client side calls to useId don’t run the risk of being the same id.

I think the fix needs to be different here.

danielroe:
Looked at this a little more and the issue is that you do not have a single root element - which is required for `useId`. (That's how the id is passed from server -> client.)

I've updated the docs in https://github.com/nuxt/nuxt/commit/401370b3a04b0651370dc5d4726b0d8dd5a70511.

ToBinio:
I see.

Thanks a lot for the quick response and docs update and thanks for the great work!

packages/nuxt/src/app/composables/id.ts
xak2000: I think the problem with `_` vs `-` is that the error message is confusing. 99% people see the difference in 1 character and think it is a bug or something like this. It's not clear from the hydration mismatch error that the ID value was actually generated purely on the client and the server value was ignored because of mutli-root element.

I'm not sure how this problem could be solved. Maybe changing `_` to `_client_` could make it more clear, but of course it's silly "solution". :) Ideally the error message should clearly indicate this, but I'm not sure at the time it is generated, it has access to the information about `useId()` and if it was used at all.

PR #26386, opened 2024-03-19
fix(nuxt): correctly move v-if from slot to wrapper component

### LINKED ISSUE #26361: Setting `selectiveClient: 'deep'` breaks components with conditional slots
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.11.1
- CLI Version:      3.11.0
- Nitro Version:    2.9.4
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-romtaf

### Describe the bug

Setting `componentIslands.selectiveClient` to `deep` currently breaks things like conditionally rendered slots because the transform isn't handling directives properly. It's transforming `v-if="foo"` to ``:props="[{ v-if: `foo` }]"``

### Additional context

Also, if I refactor to template + v-if + slot (https://stackblitz.com/edit/github-romtaf-cwixc1), I'm getting warnings like ` WARN  [Vue warn]: Extraneous non-props attributes (data-v-875a1e80) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.` However, there is no warning if I don't enable that option.

### Logs

```
[vite-node] [plugin:vite:vue] [SyntaxError] ./components/Foo.vue:6:0

<script setup lang="ts">
import { vforToArray as __vforToArray } from '#app/components/utils'
import NuxtTeleportIslandComponent from '#app/components/nuxt-teleport-island-component'
import NuxtTeleportSsrSlot from '#app/components/nuxt-teleport-island-slot'
const foo = true;
</script>

<template>
  <NuxtTeleportSsrSlot name="default" :props="[{ v-if: `foo` }]"><slot v-if="foo" /></NuxtTeleportSsrSlot>
</template>
```

huang-julien:
### 🔗 Linked issue

fix #26361 

### ❓ Type of change

### 📚 Description

Hi :wave: this fix a bug where v-if are not correctly moved from the slot to the wrapper component

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/islandsTransform.ts
+++ b/packages/nuxt/src/components/islandsTransform.ts
@@ -106,7 +106,7 @@ export const islandsTransform = createUnplugin((options: ServerOnlyComponentTran
             s.appendLeft(startingIndex + loc[0].start, `<NuxtTeleportSsrSlot${vIf ? ` v-if="${vIf}"` : ''} name="${slotName}" :props="${bindings}">`)
             // remove v-if from first tag
             if (vIf) {
-              s.overwrite(startingIndex + loc[0].start,  startingIndex + loc[0].end, code.slice( startingIndex + loc[0].start,  startingIndex + loc[0].end).replace(`v-if="${vIf}"`, ''))
+              s.overwrite(startingIndex + loc[0].start, startingIndex + loc[0].end, code.slice(startingIndex + loc[0].start, startingIndex + loc[0].end).replace(`v-if="${vIf}"`, ''))
             }
             s.appendRight(startingIndex + loc[1].end, '</NuxtTeleportSsrSlot>')
           } else if (options.selectiveClient && ('nuxt-client' in node.attributes || ':nuxt-client' in node.attributes)) {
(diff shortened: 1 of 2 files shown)


brc-dd:
There might be cases like v-else-if or v-show too. Is there some generic way to solve this instead of having to list out each directive manually? 👀 

Also, I believe `getPropsToString` in islandsTransform doesn't handle props with hyphen properly:

```
getPropsToString({'foo-bar': 'baz'}) // [{ foo-bar: `bar` }]"
```

which is invalid JS.

huang-julien:
indeed thank you ! I'm currently working on changing the way slots and client works internally to make this work for non-SFC too.

For now I think we only need to extract a few directive, I believe we still want to leave directive on the component (other than v-if/else/elseif)

brc-dd:
> Also, I believe getPropsToString in islandsTransform.ts doesn't handle props with hyphen properly:

This is still not fixed. Or is it guaranteed that the passed bindings won't contain any name having any character other than :a-zA-Z 👀  If not then something like this at line 162 of `packages/nuxt/src/components/islandsTransform.ts` might be better:

```ts
  const content = Object.entries(bindings).filter(b => b[0] && b[0] !== '_bind').map(([name, value]) => isBinding(name) ? `[\`${name.slice(1)}\`]: ${value}` : `[\`${name}\`]: \`${value}\``).join(',')
```

huang-julien:
oops sorry, i completly forgot about that ! thanks !

danielroe (approved):

PR #26388, opened 2024-03-19
fix(nuxt): make pageMeta ast undefineable in transformation

### LINKED ISSUE #26377: Browser Console Warning: definePageMeta() is a compiler-hint helper that is only usable inside the script block of a single file component which is also a page.
### Environment

Nuxt Version: 3.11.1

### Reproduction

https://stackblitz.com/edit/nuxt-starter-yd5hhb?file=pages%2Findex.client.vue

### Describe the bug

When defining Client-Only Pages, Received this warning.
`definePageMeta() is a compiler-hint helper that is only usable inside the script block of a single file component which is also a page. Its arguments should be compiled away and passing it at runtime has no effect.`

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix #26377

### ❓ Type of change

### 📚 Description

Hi :wave: this PR makes the meta AST undefineable in the transform plugin.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/plugins/page-meta.ts
+++ b/packages/nuxt/src/pages/plugins/page-meta.ts
@@ -122,32 +122,33 @@ export const PageMetaPlugin = createUnplugin((options: PageMetaPluginOptions) =>
           const name = 'name' in node.callee && node.callee.name
           if (name !== 'definePageMeta') { return }
 
-          const meta = node.arguments[0] as Expression & { start: number, end: number }
+          const meta = node.arguments[0] as (Expression & { start: number, end: number }) | undefined
+          let contents = `const __nuxt_page_meta = ${meta ? code!.slice(meta.start, meta.end) || 'null' : 'null'}\nexport default __nuxt_page_meta` + (options.dev ? CODE_HMR : '')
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/plugins/page-meta.ts
```diff
@@ -122,32 +122,33 @@ export const PageMetaPlugin = createUnplugin((options: PageMetaPluginOptions) =>
...
-
-          let contents = `const __nuxt_page_meta = ${code!.slice(meta.start, meta.end) || 'null'}\nexport default __nuxt_page_meta` + (options.dev ? CODE_HMR : '')
-
-          function addImport (name: string | false) {
-            if (name && importMap.has(name)) {
-              const importValue = importMap.get(name)!.code
-              if (!addedImports.has(importValue)) {
-                contents = importMap.get(name)!.code + '\n' + contents
-                addedImports.add(importValue)
+          const meta = node.arguments[0] as (Expression & { start: number, end: number }) | undefined
```
danielroe, line 125: This plugin should already be doing it:

https://github.com/nuxt/nuxt/blob/ff1bb56e3f90472ff7aae69b214e21197c78400b/packages/nuxt/src/core/plugins/tree-shake.ts#L29-L46

It would be worth investigating why it is not in this case.

PR #26407, opened 2024-03-20
fix(nuxt): pass `joinRelativeURL` to bundle renderer and share paths with nitro

danielroe:
### ❓ Type of change

### 📚 Description

Previously we switched across to `joinRelativeURL` in `build/paths.mjs` but didn't do so for `paths.ts` which is used in the nitro renderer.

This PR:
1. updates the implementations to be the same
2. deduplicates these two files on the server when building

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/paths.ts
```diff
@@ -1,4 +1,4 @@
-import { joinURL } from 'ufo'
+import { joinRelativeURL as joinURL } from 'ufo'
```
pi0, line 1: I would suggest using `joinRelativeURL` as the canonical name to make today's (and future) tracing easier.

packages/nuxt/src/app/entry.ts
```diff
@@ -3,7 +3,7 @@ import type { App } from 'vue'
 
 // These files must be imported first as they have side effects:
 // 1. (we set __webpack_public_path via this import, if using webpack builder)
-import '#build/paths.mjs'
+import '#paths'
```
pi0, line 6: i'm sure nitro at least will keep using `#internal/nitro/` scoped virtuas but it _might_ conflict with any other alias by modules we could use `#nuxt-paths` or something non conflicting. totally up to you.

vitest.config.ts
```diff
@@ -8,7 +8,7 @@ export default defineConfig({
   resolve: {
     alias: {
       '#build/nuxt.config.mjs': resolve('./test/mocks/nuxt-config'),
-      '#build/paths.mjs': resolve('./test/mocks/paths'),
+      '#paths': resolve('./test/mocks/paths'),
```
pi0, line 11: i'm sure nitro at least will keep using `#internal/nitro/` scoped virtuas but it _might_ conflict with any other alias by modules we could use `#nuxt-paths`, `#app-paths`, `#app/paths` or something non conflicting. totally up to you.

danielroe: maybe `#internal/nuxt/paths`? what do you think?

pi0 (approved):
🆒 💯

PR #26421, opened 2024-03-21
perf(nuxt): don't tree-shake `useServerHead` in dev

danielroe:
### ❓ Type of change

### 📚 Description

Spotted this opportunity to improve performance - we deliberately do not tree-shake any other composables in dev mode, but weren't checking before inserting `useServerHead`.

This should mean we don't need to run that vite plugin at all and it won't be initialised:

https://github.com/nuxt/nuxt/blob/3c7e68c84637d3aca9273921f8176e87e8423696/packages/nuxt/src/core/nuxt.ts#L187-L190

cc: @harlan-zw in case there's a reason we should tree-shake it in dev mode

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/head/module.ts
+++ b/packages/nuxt/src/head/module.ts
@@ -1,5 +1,6 @@
 import { resolve } from 'pathe'
 import { addComponent, addImportsSources, addPlugin, addTemplate, defineNuxtModule, tryResolveModule } from '@nuxt/kit'
+import { defu } from 'defu'
 import { distDir } from '../dirs'
 
 const components = ['NoScript', 'Link', 'Base', 'Title', 'Meta', 'Style', 'Head', 'Html', 'Body']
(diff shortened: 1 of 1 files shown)


harlan-zw:
Nice, should be good

packages/nuxt/src/head/module.ts
```diff
@@ -29,9 +30,11 @@ export default defineNuxtModule({
     }
 
     // allow @unhead/vue server composables to be tree-shaken from the client bundle
-    nuxt.options.optimization.treeShake.composables.client['@unhead/vue'] = [
-      'useServerHead', 'useServerSeoMeta', 'useServerHeadSafe'
-    ]
+    if (!nuxt.options.dev) {
+      nuxt.options.optimization.treeShake.composables.client['@unhead/vue'] = defu(nuxt.options.optimization.treeShake.composables.client['@unhead/vue'], [
+        'useServerHead', 'useServerSeoMeta', 'useServerHeadSafe'
+      ])
```
danielroe, line 36:
```suggestion
      nuxt.options.optimization.treeShake.composables.client['@unhead/vue'] = [
        'useServerHead', 'useServerSeoMeta', 'useServerHeadSafe'
      ]
```

packages/nuxt/src/head/module.ts
```diff
@@ -1,5 +1,6 @@
 import { resolve } from 'pathe'
 import { addComponent, addImportsSources, addPlugin, addTemplate, defineNuxtModule, tryResolveModule } from '@nuxt/kit'
+import { defu } from 'defu'
```
danielroe, line 3:
```suggestion
```

harlan-zw (approved):

PR #26439, opened 2024-03-22
docs: update 4.styling.md

chenfan0:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/1.getting-started/4.styling.md
+++ b/docs/1.getting-started/4.styling.md
@@ -51,7 +51,7 @@ The stylesheets will be inlined in the HTML rendered by Nuxt, injected globally 
 
 ### Working With Fonts
 
-Place your local fonts files in your `~/public/` directory, for example in `~/public/fonts`. You can then reference them in your stylesheets using `url()`.
+Place your local fonts files in your `public/` directory, for example in `public/fonts`. You can then reference them in your stylesheets using `url()`.
 
 ```css [assets/css/main.css]
 @font-face {


nopeless:
regex to find all instances referring to path

```
`~?/?(\.nuxt|\.output|assets|components|composables|content|layouts|middleware|modules|node_modules|pages|plugins|public|server|utils)(/[.*/]*)?`
```

regex to find all instances that aren't enclosed in []
```
`(?<!\[)`~?/?(\.nuxt|\.output|assets|components|composables|content|layouts|middleware|modules|node_modules|pages|plugins|public|server|utils)(/[.*/]*)?`
```

image


if you want to pass this pr

u should probably replace all of these as well

```
7.migration\5.plugins-and-middleware.md
71:Much like Nuxt 2, route middleware placed in your `~/middleware` folder is automatically registered. You can then specify it by name in a component. However, this is done with `definePageMeta` rather than as a component option.
80:1. Any global middleware (such as in your `nuxt.config`) can be placed in your `~/middleware` folder with a `.global` extension, for example `~/middleware/auth.global.ts`.

1.getting-started\4.styling.md
54:Place your local fonts files in your `~/public/` directory, for example in `~/public/fonts`. You can then reference them in your stylesheets using `url()`.

1.getting-started\10.deployment.md
86:1. Load the HTML of your application's root route (`/`), any non-dynamic pages in your `~/pages` directory, and any other routes in the `nitro.prerender.routes` array.

1.getting-started\4.assets.md
34:In your application's code, you can reference a file located in the [`assets/`](/docs/guide/directory-structure/assets) directory by using the `~/assets/` path.

2.guide\2.directory-structure\3.error.md
28:Although it is called an 'error page' it's not a route and shouldn't be placed in your `~/pages` directory. For the same reason, you shouldn't use `definePageMeta` within this page. That being said, you can still use layouts in the error file, by utilizing the [`NuxtLayout`](/docs/api/components/nuxt-layout) component and specifying the name of the layout.

2.guide\1.concepts\1.auto-imports.md
135:Nuxt also automatically imports components from your `~/components` directory, although this is configured separately from auto-importing composables and utility functions.
139:To disable auto-importing components from your own `~/components` directory, you can set `components.dirs` to an empty array (though note that this will not affect components added by modules).

3.api\5.kit\11.nitro.md
377:just like the `~/server` folder is.

2.guide\2.directory-structure\1.components.md
145:By default, only the `~/components` directory is scanned. If you want to add other directories, or change how the components are scanned within a subfolder of this directory, you can add additional directories to the configuration:
160:    // to sub-directories of `~/components`.

2.guide\2.directory-structure\1.pages.md
16:Nuxt will automatically create a route for every page in your `~/pages/` directory.
```

danielroe:
Is there a particular reason you want to remove `~/`? It's meant to communicate that the public directory is underneath your source folder, which may be the same as your root directory, or might be `src/` or something else...

nopeless:
I have no idea why they closed this but there are a lot of inconsistencies

image

danielroe:
Not every instance needs to be exactly the same, but if any of these are confusing, I would welcome PRs...

PR #26441, opened 2024-03-22
feat(nuxt): warn when accessing private runtimeConfig on client

nopeless:
### 🔗 Linked issue

None

### ❓ Type of change

### 📚 Description

I wasted a bit too much time having trouble why server hydration and client view was different until I realized that I was accessing `useRuntimeConfig().<some prop>` in client as well. This would prevent future issues like that

### 📝 Checklist


danielroe (commented):
I think this is a great idea! Thank you ❤️ 

However, I think as written it will warn when these keys are accessed on server side as well given that we are manipulating the config within the SSR renderer.

Maybe just warn on client side? (We could add the guard within the `import.meta.client` block within `app/nuxt`.)

nopeless:
> I think this is a great idea! Thank you ❤️ 
> 
> However, I think as written it will warn when these keys are accessed on server side as well given that we are manipulating the config within the SSR renderer.
> 
> Maybe just warn on client side? (We could add the guard within the `import.meta.client` block within `app/nuxt`.)

SSR context is untouched. In fact, it exposes even more properties, specifically .nitro (which in my opinion makes more sense, let me know if this behaviour is not what you want)

only SPA context is altered

nopeless:
bump

danielroe:
(This PR is waiting for a minor release - probably Nuxt v3.12.)

danielroe (approved):

PR #26468, opened 2024-03-24
feat(nuxt): delayed/lazy hydration support

### LINKED ISSUE #24242: Lazy Hydration in Nuxt Core
## What?

This proposal introduces lazy hydration feature as a first-class citizen in Nuxt Core.

> This RFC is inspired by a [Vue 3 Lazy Hydration Plugin](https://github.com/freddy38510/vue3-lazy-hydration) from Freddy Escobar (that was inspired by another plugin from Markus Oberlehner). All the credit belongs to these amazing developers. I am literally suggesting implementing it in the core with some tiny DX improvements.

## Why?

Large JavaScript bundles are one of the key reasons behind poor performance of Single Page Applications. Every interactive element you add to the project at the component level is by default loaded eagerly during the application start-up, often blocking the critical rendering path.

The truth is - we rarely need interactivity upront. It's usually triggered - sometimes by an event like click, sometimes element appears in the viewport, sometimes it can be loaded when the browser is done with more important tasks and goes idle. Despite interactivity not being eagerly needed in most cases, we are lacking first-class solution allowing to control hydration process in Nuxt. There is a `<NuxtIsland>` component that moves the needle into the right direction but it's a 0/1 tool. If we're already going this direction, let's release a complete solution that tackles the problem as a whole.

We see lazy hydration being implemented as a first-class citizen in (meta)frameworks like Astro or Qwik with great results that prove that interactivity is rarely needed eagerly.

We need similar solution in Nuxt, coming from the core, integrated well with other Nuxt features and benefiting from framework-level optimizations.

## How?

A natural temptation would be to introduce components similar to the one from Vue3 Lazy Hydrate library but this solution is not perfect DX-wise.

When you have multiple components requireing different treating you end up doubling the size of your files by creating a separate wrapper for each component. 

Here is an example from [Vue Storefront's boilerplate](https://github.com/vuestorefront/storefront-nuxt3-boilerplate/blob/develop/apps/web/pages/product/%5Bslug%5D.vue) for Nuxt 3.

```html
<div>
  <section class="grid-in-left-top md:h-full xl:max-h-[700px]">
    <NuxtLazyHydrate when-idle>
      <Gallery :images="product?.gallery ?? []" />
    </NuxtLazyHydrate>
  </section>
  <section class="mb-10 grid-in-right md:mb-0">
    <NuxtLazyHydrate when-idle>
      <UiPurchaseCard v-if="product" :product="product" />
    </NuxtLazyHydrate>
  </section>
  <section class="grid-in-left-bottom md:mt-8">
    <UiDivider class="mb-6" />
    <NuxtLazyHydrate when-visible>
      <ProductProperties v-if="product" :product="product" />
    </NuxtLazyHydrate>
    <UiDivider class="mt-4 mb-2 md:mt-8" />
    <NuxtLazyHydrate when-visible>
      <ProductAccordion v-if="product" :product="product" />
    </NuxtLazyHydrate>
  </section>
  <UiDivider class="mt-4 mb-2" />
</div>
<section class="mx-4 mt-28 mb-20">
  <NuxtLazyHydrate when-visible>
    <RecommendedProducts v-if="recommendedProducts" :products="recommendedProducts" />
  </NuxtLazyHydrate>
</section>
```
A directive would be a much more elegant approach that fixes this small but definitely annoying DX flaw.

### `hydrate` directive

The idea is simple - introduce global `hydrate` directive that can control when components are hydrated.

```html
<!-- Hydrate when certain event is triggered -->
<MyComponent hydrate:on="['click', 'touchstart']" />
<!-- Hydrate when certain condition is met -->
<MyComponent hydrate:when="shouldHydrate === true" />
<!-- Hydrate when element is in the viewport -->
<MyComponent hydrate:when-visible />
<!-- Hydrate when element is in the viewport (with an offset) -->
<MyComponent hydrate:when-visible="{ rootMargin: '100px' }" />
<!-- Hydrate when browser is idle (requestIdelCallback) -->
<MyComponent hydrate:when-idle />
<!-- Hydrate when browser is idle with maximum timeout -->
<MyComponent hydrate:when-idle="4000" />
<!-- Never Hydrate, works just like <NuxtIsland> -->
<MyComponent hydrate:never />
```

### Hydration callback

Having a hydration callback could make it easier to react to components hydration from it's parent.

```html
<MyComponent @hydrated="callback" />
```
It would be useful when we want to hydrate a set of compoennts in particular order.

```html
<template>
    <MyComponentA hydrate:when="hydrateA" @hydrated="hydrateB = true" />
    <MyComponentB hydrate:when="hydrateB" />
    <button @click="hydrateA = true">Start hydration</button>
</template>

<script setup>
const hydrateA = ref(false)
const hydrateB = ref(false)
</script>
```

### Hydration Wrapper

We can use already existing `NuxtIsland` component to wrap multiple components/nodes at once.

```html
<NuxtIsland hydrate:never>
    <div> {{ message }} </div>
    <MyComponentA />
    <MyComponentB />
    <MyComponentC />
</NuxtIsland>
```

### Framework-level optimizations

Thanks to component auto-imports we can statically analyze the code and specify which components chunks should be lazy-loaded as build-time optimization.

### Additional information

### Final checks

GalacticHypernova:
### 🔗 Linked issue

Resolves #24242 
### ❓ Type of change

### 📚 Description

Lazy components are great for controlling the chunk sizes in your app, but they don't always enhance runtime performance, as they still load eagerly unless conditionally rendered. In real-world applications, some pages may include a lot of content and a lot of components, and most of the time not all of them need to be interactive as soon as the page is loaded. Having them all load eagerly can negatively impact performance.

In order to optimize your app, you may want to delay the hydration of some components until they're visible, or until the browser is done with more important tasks.

Nuxt supports this using lazy (or delayed) hydration, allowing you to control when components become interactive.

### Hydration Strategies

Nuxt provides a range of built-in hydration strategies. Only one strategy can be used per lazy component.

::warning
Currently Nuxt's built-in lazy hydration only works in single-file components (SFCs), and requires you to define the prop in the template (rather than spreading an object of props via `v-bind`).
::

#### `hydrate-on-visible`

Hydrates the component when it becomes visible in the viewport.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-on-visible />
  </div>
</template>
```

::read-more{to="https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver" title="IntersectionObserver options"}
Read more about the options for `hydrate-on-visible`.
::

::note
Under the hood, this uses Vue's built-in [`hydrateOnVisible` strategy](https://vuejs.org/guide/components/async.html#hydrate-on-visible).
::

#### `hydrate-on-idle`

Hydrates the component when the browser is idle. This is suitable if you need the component to load as soon as possible, but not block the critical rendering path.

You can also pass a number which serves as a max timeout.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-on-idle />
  </div>
</template>
```

::note
Under the hood, this uses Vue's built-in [`hydrateOnIdle` strategy](https://vuejs.org/guide/components/async.html#hydrate-on-idle).
::

#### `hydrate-on-interaction`

Hydrates the component after a specified interaction (e.g., click, mouseover).

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-on-interaction="mouseover" />
  </div>
</template>
```

If you do not pass an event or list of events, it defaults to hydrating on `pointerenter` and `focus`.

::note
Under the hood, this uses Vue's built-in [`hydrateOnInteraction` strategy](https://vuejs.org/guide/components/async.html#hydrate-on-interaction).
::

#### `hydrate-on-media-query`

Hydrates the component when the window matches a media query.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-on-media-query="(max-width: 768px)" />
  </div>
</template>
```

::note
Under the hood, this uses Vue's built-in [`hydrateOnMediaQuery` strategy](https://vuejs.org/guide/components/async.html#hydrate-on-media-query).
::

#### `hydrate-after`

Hydrates the component after a specified delay (in milliseconds).

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent :hydrate-after="2000" />
  </div>
</template>
```

#### `hydrate-when`

Hydrates the component based on a boolean condition.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent :hydrate-when="isReady" />
  </div>
</template>
<script setup lang="ts">
const isReady = ref(false)
function myFunction() {
  // trigger custom hydration strategy...
  isReady.value = true
}
</script>
```

#### `hydrate-never`

Never hydrates the component.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-never />
  </div>
</template>
```

### Listening to Hydration Events

All delayed hydration components emit a `@hydrated` event when they are hydrated.

```vue [pages/index.vue]
<template>
  <div>
    <LazyMyComponent hydrate-on-visible @hydrated="onHydrated" />
  </div>
</template>

<script setup lang="ts">
function onHydrate() {
  console.log("Component has been hydrated!")
}
</script>
```


packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,56 @@
...
+          unobserve?.()
+          unobserve = null
+        })
+      })
+      onBeforeUnmount(() => {
+        unobserve?.()
+        unobserve = null
+      })
+      return () => h('div', { ref: el }, [
+        isIntersecting.value ? h(componentLoader, attrs) : null,
```
huang-julien, line 27: Do we want to lazy hydrate or to lazy load here ?

Lazy hydration would need to render the component in SSR. Keep it "static"/non-interactive in client until loaded

GalacticHypernova: Yea, we probably want all of these wrappers to lazy load as opposed to lazy hydrate. Can this be lazy loaded via 
```ts
h(await import(compName))
```?

Comment (huang-julien) [reply packages/nuxt/src/components/runtime/client-delayed-component.ts:27]: It's more about how `createClientIO` should work.

The issue with the current state is that it is a client-only until observed because it is never rendered in SSR.  So when missused end users can experience a reduced UX (like when using `CLientOnly`). And also having an reduced webperf score if the lazyhydrated component is within the viewport at loading.

I think what #24242 was expecting is to render the component in SSR and then hydrate when observed.

So we could probably render the component in server side. Then until observed, it would be a static vnode (you can check how to do it in nuxt-island). Once observed, you'll be able to render the component.

I think the PR probably just need some changes within `createClientIO`

Comment (GalacticHypernova) [reply packages/nuxt/src/components/runtime/client-delayed-component.ts:27]: Perhaps something along the lines of:
```ts
return () => h('div', { ref: el }, [
   isIntersecting.value ? h(defineAsyncComponent(import('./Comp.vue').then(c=>c.default || c)), attrs) : null
])
```

GalacticHypernova: > The issue with the current state is that it is a client-only until observed because it is never rendered in SSR. So when missused, end users can experience a reduced UX (like when using CLientOnly). And also having an reduced webperf score if the lazyhydrated component is within the viewport at loading.
Though the question here would be, would lazy loading not inherently provide better performance as opposed to lazy hydration? We can use a static vnode of course, but wouldn't lazy loading still reduce bundle? Of course everything can be misused, it's up to the individual developers to use the tools as intended..

huang-julien: It's more about the meaning behind "lazy hydration", in my opinion it implies that the component is still rendered in SSR.

The current implementation is more like a "lazy component until in viewport" 🤔 .

> Though the question here would be, would lazy loading not inherently provide better performance as opposed to lazy hydration?

Probably slightly better with lazy loading. But rendering a static vnode isn't very costly in term of performance.

> There is no reason to make components inside the viewport async regardless, that in and of itself would hurt performance, even if it doesn't use any special wrappers. How can it hurt UX, especially if used properly?

Indeed but some users may use it for a component that is in small breakpoint -> meaning that in desktop it would load the component after the page load resulting in some layout shift.

danielroe:
🎉

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,94 @@
+import { createStaticVNode, createVNode, defineComponent, getCurrentInstance, h, onBeforeUnmount, onMounted, ref } from 'vue'
+import type { Component, Ref } from 'vue'
+// import ClientOnly from '#app/components/client-only'
+import { useObserver } from '#app/utils'
+import { getFragmentHTML } from '#app/components/utils'
+import { useNuxtApp } from '#app/nuxt'
+
+// todo find a better way to do it ?
```
huang-julien, line 8: This is probably the best wait to do it. Can remove this comment :+1:

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,94 @@
...
+          idleHandle = null
+        })
+      })
+      onBeforeUnmount(() => {
+        if (idleHandle) {
+          cancelIdleCallback(idleHandle as unknown as number)
+          idleHandle = null
+        }
+      })
+      return () => h('div', {}, [
```
huang-julien, line 89: Do we need a div here ?

GalacticHypernova: We do not, I was thinking about it here, I'll remove it

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: Removed

GalacticHypernova:
I was thinking of changing up the way visibility based component behaves, to ensure we don't pollute the DOM unnecessarily with extra nodes. What do you think? Do note I wrote this on my phone right in this message so it may have some mistakes, but I'm talking about the general idea.

```ts
return () => {
  if (!isIntersecting.value && (!instance.vnode.el || !nuxt.isHydrating)) {
    return h('div', { ref: el });
  }
  if (isIntersecting.value) {
    return h(componentLoader, attrs);
  } else {
    const fragmentHTML = getFragmentHTML(instance.vnode.el ?? null, true)?.join('') || '';
    return createVNode(createStaticVNode(fragmentHTML, 1));
  }
};
```

packages/nuxt/src/app/utils.ts
```diff
@@ -1,4 +1,49 @@
...
+import { useNuxtApp } from './nuxt'
+
 export function toArray<T> (value: T | T[]): T[] {
   return Array.isArray(value) ? value : [value]
 }
+
+type CallbackFn = () => void
+type ObserveFn = (element: Element, callback: CallbackFn) => () => void
+
+export function useObserver (): { observe: ObserveFn } | undefined {
```
harlan-zw, line 10: I'd rename this to useIntersectionObserver, maybe it's something Nuxt can export in the future.

packages/nuxt/src/app/utils.ts
```diff
@@ -1,4 +1,49 @@
...
+  if (import.meta.server) { return }
```
harlan-zw, line 11: Having a single return type would be better, we could make a noop `observe` function, otherwise, end users need to do type juggling

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,87 @@
...
+    setup (_, { attrs }) {
+      if (import.meta.server) {
+        return () => h(componentLoader, attrs)
+      }
+      const nuxt = useNuxtApp()
+      const instance = getCurrentInstance()!
+      const isIdle = ref(false)
+      let idleHandle: number | null = null
+      onMounted(() => {
+        idleHandle = requestIdleCallback(() => {
```
harlan-zw, line 72: you need to use the `app/compat/idle-callback.ts` polyfill

harlan-zw: it may make sense to register this `onNuxtReady` also

harlan-zw:
Nice job overall, the code looks good.

In terms of the API, I have a couple of issues:
- End users have no way to configure the visibility and idle triggers which can block them from fine-grain performance tuning. For example, setting more appropriate visibility boundaries or timeouts on the idle.
- Component names are very tightly coupled to behaviour, if we wanted to add another optimized way to lazy load a component it starts getting complicated (i.e `mouseover`, `click`, etc)

Personally, I'd like to see a lower-level implementation first that exposes the `defineAsyncComponent` internals on Lazy prefixed components (or a new component prefix). Nuxt can then export an idle / visibility loader that can be configured with types.

```vue
<template>
	<LazyFoo :loader="createIdleLoader({ timeout: 3000 )">
</template>
```

I'd say if someone is reaching for a lazy component then they are already thinking about the best way to load the component, I feel like the extra effort to manually use/configure a loader versus just using `<LazyIdleFoo>` won't be a big deal. IMO the more lazy-loaded components have the worst UX so they should be used sparingly where they are needed.

I need more time to sit down on this properly and prototype if it's feasible, I would be interested to hear your thoughts though.

packages/nuxt/src/app/utils.ts
GalacticHypernova: Sure thing! I'll rename it.

packages/nuxt/src/app/utils.ts
GalacticHypernova: Yea I was thinking about that, could that cause any issues if the server render returns a blank observer function?

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: You mean `onNuxtReady` in stead of onMounted or I Mounted inside on NuxtRenderHTMLContext?

GalacticHypernova:
First things first, thank you so much for the feedback!
> * End users have no way to configure the visibility and idle triggers which can block them from fine-grain performance tuning.

Yea, this is something I had thought about. The issue with this is that these extended lazy components are all based on the regular "sync" component, and in order to even have this extra loader option there would need to be type magic on nuxt's end so that it would be accepted as a prop, and to ensure that it isn't used in non-lazy components as that could cause conflicts which I haven't managed to solve quite yet (maybe you have an idea about them?). Although I agree that having these fine tuned performance opportunities is valuable, at the moment it doesn't seem feasible without making a reserved component that would act as a wrapper, which is what most delayed hydration modules/libraries do. The obvious issue with this approach is that it adds **a ton** of boilerplate, so the current implementation extends the base component. 

That being said, I can add the functionality of it with no issue, like the `create*Loader` and have it used in the components. The main issue would be DX, specifically around type safety

packages/nuxt/src/app/utils.ts
GalacticHypernova: Should I create another overload of this that won't cache the observer in nuxtApp._observer? To allow different options?

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: This has been implemented

packages/nuxt/src/app/utils.ts
GalacticHypernova: Reused the same one, saw that the logic matches

packages/nuxt/src/app/utils.ts
GalacticHypernova: Renamed to useIntersectionObserver

GalacticHypernova:
Documentation updates to include the new `loader` prop and the corresponding utilities following harlan's reviews will be added soon. I'll mark it as draft for the time being until I take care of everything reviewed

harlan-zw:
Thanks for your feedback on my comment, it makes sense and they are some good points. The changes look good and I think you have things headed in the right direction.

One final idea I'll throw in the ring though (sorry :laughing:), what are your thoughts on being able to create the components explicitly? Like how we do with custom Nuxt Links.

```ts
// MyLazyIdleComponent
export default defineLazyComponent({
  componentName: 'FooComponen',
  loader: createIdleLoader({
    idle: 1200,
  })
})
```

```vue
<template>
  <MyLazyIdleComponent />
</template>
```

GalacticHypernova:
> being able to create the components explicitly

I can definitely see this being useful in advanced use cases. Though I do sense some slight issues with it.
Functionality wise it is very easy to set it up, we can convert the componentLazyHydration flag to accept either a boolean or an object with the info, or have an array where the user can input multiple. In loader.ts, this could easily be checked through the nuxt options. The pattern won't matter because the full name could be checked based on the full name provided in the options.

Type-wise however, I haven't really thought about quite yet. Ideally it should recognize the components specified within the IDE, which means ideally it should be added to the type declarations. I could try to look into it, but once again the functionality itself is fully feasible.

GalacticHypernova:
Event, idle, and visibility based delayed hydration all appear to work again. All of the components have been documented, as well as their loaders, so I am making this as ready to review again

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,129 @@
...
+          cancelIdleCallback(idleHandle as unknown as number)
+          idleHandle = null
+        }
+      })
+      return () => isIdle.value ? h(componentLoader, attrs) : (instance.vnode.el && nuxt.isHydrating) ? createVNode(createStaticVNode(getFragmentHTML(instance.vnode.el ?? null, true)?.join('') || '', 1)) : null
+    },
+  })
+}
+
+const eventsMapper = new WeakMap<ComponentInternalInstance, (() => void)[]>()
```
GalacticHypernova, line 93: Thinking about whether or not this is necessary. Theoretically, if each instance is guaranteed to be unique, this could save edge cases where the same instance is registered more than once, which is why this is used to keep track of it internally. But the question is, is it even possible for it to happen?

GalacticHypernova:
Marking as draft to fix custom loaders

GalacticHypernova:
Component names have been added to templates so they will now get auto detected. I'll look into adding the additional loader prop to the templates as well.

GalacticHypernova:
All type matters should also be fixed now. That includes loaders, templates, type definition, intellisense, and everything else

GalacticHypernova:
To summarize:
All components and their loaders are fully functional now, including custom triggers. They are all auto imported (loaders and components), and they have added type definitions to autocomplete the loader prop. Experimental config has been normalized to delayedHydration, and the components usage, loaders, and overriding defaults examples were documented.
Would love to hear if there are any performance implications or edge cases with the latest changes!

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -119,6 +119,79 @@ const show = ref(false)
 </template>
 ```
 
+## Delayed Hydration
+
+In real world applications, some pages may include a lot of content and a lot of components, and most of the time not all of them need to be interactive as soon as the page is loaded. Having them all load eagerly can negatively impact performance and increase bundle size.
```
harlan-zw, line 124: It may be useful to outline the difference with Lazy prefixed components as it might seem like Lazy should solve this already.

Some rough prose, feel free to word however you like

> While Lazy prefixed components are useful for controlling the chunk size of your app, they won't necessarily improve the loading performance. This is because the async component is loaded eagerly with the rest of the page's scripts.

> Delayed hydration components fix this, allowing you to load the component just in time to maintain a good user experience while providing the performance benefit of not loading the component immediately.

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,129 @@
...
+  return defineComponent({
+    inheritAttrs: false,
+    setup (_, { attrs }) {
+      if (import.meta.server) {
+        return () => h(componentLoader, attrs)
+      }
+
+      const nuxt = useNuxtApp()
+      const instance = getCurrentInstance()!
+      const isIntersecting = ref(false)
```
harlan-zw, line 29: slightly misleading, should be `hasIntersected` (it doesn't get toggled to false if it's not in the viewport anymore)

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -119,6 +119,79 @@ const show = ref(false)
...
+```
+
+### Custom hydration triggers
+
+If you would like to override the default hydration triggers when dealing with delayed hydration, like changing the timeout, the options for the intersection observer, or the events to trigger the hydration, you can do so by supplying a `loader` prop to your lazy components.
+
+```vue [pages/index.vue]
+<template>
+  <div>
+    <LazyIdleMyComponent :loader="createIdleLoader({timeout: 3000})"/>
```
harlan-zw, line 175: The `loader` prop only makes sense if we're passing in a promise, which would only make sense for a generic component. 

```ts
<LazyMyComponent :loader="createIdleLoader({timeout: 3000})"/>
```

If you prefer not to create a low-level implementation, that's okay, but I think it's probably better to use a different prop name and the function wouldn't be required.

packages/nuxt/src/app/composables/hydrate.ts
```diff
@@ -23,3 +23,21 @@ export const useHydration = <K extends keyof NuxtPayload, T = NuxtPayload[K]> (k
     })
   }
 }
+
+/**
+ * A `requestIdleCallback` options utility, used to determine custom timeout for idle-callback based delayed hydration.
+ * @param opts the options object, containing the wanted timeout
+ */
```
harlan-zw, line 30: As previous comment, this only makes sense if they return a promise.

docs/2.guide/2.directory-structure/1.components.md
GalacticHypernova: I was thinking about it, but since there are defaults for all the components the loader prop is there to allow overrides. Should I name it differently?

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: Good spot, fixed!

packages/nuxt/src/app/composables/hydrate.ts
GalacticHypernova: I don't think I understand why it needs to be a promise? Why can it not be a synchronous override?

docs/2.guide/2.directory-structure/1.components.md
GalacticHypernova: Rephrased it, let me know what you think!

packages/nuxt/src/app/composables/hydrate.ts
harlan-zw: from a naming point of view* i'd go for something with a define prefix if it's just for the types

docs/2.guide/2.directory-structure/1.components.md
harlan-zw: Yes renaming, keep in mind that `loader` is a prop for defineAsyncComponent in Vue which is a promise, we want to remove ambiguity around what the prop does

harlan-zw: I suggested using `loader` with the thinking that the user would be providing a promise so the signature would match

packages/nuxt/src/app/composables/hydrate.ts
GalacticHypernova: Renamed to `hydrate`

GalacticHypernova:
Was bored so I added another type of delayed hydration for general purpose conditional hydration. For never-hydrated components I don't quite know if it's worth making a wrapper over just using a .server component, but I would love to hear your opinion!

GalacticHypernova:
Looks like there was some regression, the tests in dev fail now despite passing previously

packages/nuxt/src/app/utils.ts
```diff
@@ -1,4 +1,50 @@
...
+export function useIntersectionObserver (options?: Partial<IntersectionObserverInit>): { observe: ObserveFn } {
+  if (import.meta.server) { return { observe: () => () => {} } }
+
+  const nuxtApp = useNuxtApp()
+  if (nuxtApp._observer) {
+    return nuxtApp._observer
+  }
+
+  let observer: IntersectionObserver | null = null
+  const callbacks = new Map<Element, CallbackFn>()
```
huang-julien, line 21: Maybe can we use a WeakMap ?

GalacticHypernova: I think so, don't see a reason why not

danielroe:
Thank you so much for your work on this!

 Linking https://github.com/vuejs/core/pull/11458. Can we rework this PR so it uses that under the hood? 🙏

GalacticHypernova:
Thanks for linking!
This will likely simplify it a lot

GalacticHypernova:
@danielroe 
I'm linking here this PR https://github.com/vuejs/core/pull/11530 that took care of the full customization options for delayed hydration strategies. Should I rework this ahead of time to use the expected 3.5 beta.1 release?

danielroe:
You can include the changes to `package.json` resolutions from https://github.com/nuxt/nuxt/pull/28285/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R53-R58:

```json
    "vue": "3.5.0-beta.1",
    "@vue/compiler-core": "3.5.0-beta.1",
    "@vue/compiler-dom": "3.5.0-beta.1",
    "@vue/compiler-sfc": "3.5.0-beta.1",
    "@vue/compiler-ssr": "3.5.0-beta.1",
    "@vue/shared": "3.5.0-beta.1"
```

danielroe:
Yes, do feel free to rebase on that branch so the tests pass!

GalacticHypernova:
A point for discussion would be supporting async/promise conditions in conditional hydration. I don't imagine there are many use cases for that but I could probably add support for it if it's deemed necessary.

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,102 @@
...
+    setup (_, { attrs }) {
+      if (import.meta.server) {
+        return () => h(defineAsyncComponent(loader), attrs)
+      }
+      const ready = ref(false)
+      const nuxt = useNuxtApp()
+      const instance = getCurrentInstance()!
+      onNuxtReady(() => ready.value = true)
+      // This is a hack to prevent hydration mismatches for all hydration strategies
+      return () => ready.value ? h(defineAsyncComponent({ loader, hydrate: hydrateOnVisible(attrs.hydrate as IntersectionObserverInit | undefined) })) : nuxt.isHydrating && instance.vnode.el ? h('div', attrs) : null
```
GalacticHypernova, line 18: While this hack prevents hydration mismatches for all types of hydrations, it also appears to ignore them altogether, and in stead it treats everything as a regular dynamic import (meaning it doesn't defer anything outside the initial client render)

GalacticHypernova: Marking as resolved for the time being due to the new solution. This could probably be explored a little further

GalacticHypernova: This is now extra resolved 👍

GalacticHypernova:
The failing test fixture appears to be related to some intermittent fetch issue, updating branch would likely resolve it.

danielroe:
The link checker failure can be ignored. No need to update branch.

GalacticHypernova:
Yea, the link checker was broken from even before that, I referenced https://github.com/nuxt/nuxt/actions/runs/10861880723/job/30144306768

GalacticHypernova:
Reverting to draft because there is an issue with reactivity that I need to fix

packages/nuxt/src/app/composables/hydrate.ts
```diff
@@ -23,3 +23,21 @@ export const useHydration = <K extends keyof NuxtPayload, T = NuxtPayload[K]> (k
...
+ * An `IntersectionObserver` options utility, used to determine custom viewport-based delayed hydration behavior.
+ * @param opts the options object, containing the wanted viewport options
+ */
+export const createVisibleLoader = (opts: Partial<IntersectionObserverInit>) => opts
+
+/**
+ * A utility used to determine which event/events should trigger hydration in components with event-based delayed hydration.
+ * @param events an event or array of events that will be used to trigger the hydration
+ */
+export const createEventLoader = (events: keyof HTMLElementEventMap | Array<keyof HTMLElementEventMap>) => events
```
GalacticHypernova, line 43: I wonder if this is still necessary given that the props are now typed. I don't think there is much added use for this besides another way of supplying hydration overrides.

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,175 @@
...
+    },
+  })
+}
+
+/* @__NO_SIDE_EFFECTS__ */
+export const createLazyIfComponent = (loader: AsyncComponentLoader) => {
+  return defineComponent({
+    inheritAttrs: false,
+    props: {
+      hydrate: {
```
HigherOrderLogic, line 96: Is the prop name `hydrate` used on all of these components for consistency? Otherwise, I suggest changing the prop name to other name that are more "clear", for example, `condition` for `<LazyIfXxx>`, `promise` for `<LazyPromiseXxx>`, ...

GalacticHypernova: Yes, that is indeed what I was aiming for. It makes for a unified and easily-adoptable feature that people won't need to look around in the docs for the exact prop for each hydration strategy. The fact this allows to reuse a type to reduce the size of component IntelliSense is an added bonus.

HigherOrderLogic: > Though I am curious, is there any issue that you foresee with this naming convention?

If any, I think that this name is too generic and maybe used by a feature of Vue in the future (though this should be rare, as Evan is aware of Nuxt).

GalacticHypernova: >  maybe used by a feature of Vue in the future

This did happen, but back then the prop was named as `loader` (which conflicted with the loader in `defineAsyncComponent`), which is actually why I renamed it to `hydrate`. I have to say, that the chances of Vue introducing something like that are very slim, especially because the delayed hydration system is pretty much stable now in terms of API (maybe not so much in terms of overall features). 

What name would you think would be better? Ideally it should be a shared name because of the consistency, which would bring a lot more benefit to newcomers.

GalacticHypernova: > What do you think about the name precondition?

This is definitely a unique name. May I ask why you think it's better? Hydrate at least has some context awareness as for its purpose, whereas precondition is pretty vague in this sense.

I'm not entirely opposed to it, I'm just curious to hear your opinion =)

packages/nuxt/src/app/composables/hydrate.ts
GalacticHypernova: I've decided to remove this given the props themselves are now correctly tied thanks to the refactor from attrs to props

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: I've been thinking about this a little.

I do still believe that some variation of `hydrate(*?)` will be the best fit, because after all, it's for delayed hydration.
I also believe that keeping it uniform will make it much easier for newcomers to get along with it and learn about it, and it can also make it much more convenient to document.

I suppose `hydrateOn` is an equally decent name, but it sounds almost promise-condition related. What do you think?

Tofandel:
May I ask what is the logic behind using component prefixes instead of a new async wrapper component that can take props to determine hydration method?

I fear that those prefixes may not be very portable, they might conflict with lint rules and prevent ide autocompletion from working properly

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -119,6 +119,184 @@ const show = ref(false)
...
+```vue [pages/index.vue]
+<template>
+  <div>
+    <LazyPromiseMyComponent />
+  </div>
+</template>
+```
+
+Nuxt's delayed hydration system is highly flexible, allowing each developer to build upon it and implement their own hydration strategy.
+If you have highly specific hydration triggers that aren't covered by the default strategies, or you want to have conditional hydraion, you can use the general purpose `LazyIf` prefix:
```
Tofandel, line 209:
```suggestion
If you have highly specific hydration triggers that aren't covered by the default strategies, or you want to have conditional hydration, you can use the general purpose `LazyIf` prefix:
```

packages/nuxt/src/components/runtime/client-delayed-component.ts
Tofandel: May I ask what is the logic behind using fake component prefixes instead of fake wrapper component that can take props to determine hydration method?

I fear that those prefixes may not be very portable, they might conflict with lint rules and prevent ide autocompletion from working properly

For example do you see any problem with this: `<TimerHydrate :time="3000"><MyComponent></TimerHydrate>` ?

We have control on the vue template loader in nuxt, so we could definitely transform this and avoid prop name conflicts in the first place, and this solves a lot of headaches

GalacticHypernova: Hey @Tofandel ! Thanks for voicing your concern. From your example, there is more boilerplate involved as you need to add more to your template. Compare that with the following:
`<LazyTimeMyComponent :hydrate="3000" />`

I'm not entirely sure what you mean by prevent IDE autocompleting from working properly and which lint rules they could conflict with, so if you could elaborate on that it would be appreciated

docs/2.guide/2.directory-structure/1.components.md
GalacticHypernova: Thanks for catching it!

packages/nuxt/src/components/runtime/client-delayed-component.ts
Tofandel: Ide provide autocompletion on props name via the imported components in the Vue file or types of global components

So for the autocompletion to work that means all possible components names need to be globally typed, this is what you are already doing a bit lower right? 

Linters might try to understand if a component used exists and report errors if it doesn't

Ide provide autocompletion on props name via the imported components in the Vue file or types of global components

I'm just wondering if then having so many types added might be problematic (imagine I have 300 components, before they were there only twice right? The Lazy version and the normal version, 600 types is still okay, but what happens when now all 300 are multiplied 9 times with this PR?, seems like it would snowball quickly)

Maybe using a modifier on lazy components then could be a better solution like

<LazyMyComponent v-hydrate:timer="3000">

Because then we only need to provide types for the modifiers once and we still don't have the problem with the 'hydrate' and prop naming and we can correctly type each prop

GalacticHypernova: You bring some important points, but unfortunately they come with their own downsides.

While your approach could help with the amount of types, I can't help but to think about the performance impact.

Let's start with the obvious: it's very rare to have 300 components in an app, no matter the size. 20? Very much possible. 40? Sure. 70?? Maybe if it covers so many things and is extremely multi purpose. But 300 components seems like quite a jump. Furthermore, I don't believe having these 2700 types would be too much of a concern for people, regardless of where it's deployed, for 3 main reasons:
1. It's not bundled with the app so it doesn't add any additional bundle size.
2. It's written very infrequently and is only used in development, which usually happens on a developer's local computer or a dedicated server, which don't have a memory bottleneck.
3. If you take a look at all the built-in type declarations for every type in web development, whether it be NodeJS types, DOM types, and all others, you can see it adds up to way more than 2700 types, and yet it works and performs perfectly fine in development, which leads me to believe the impact from the types might not be that significant. 

Now let's go over the performance aspect:
While having a directive could make it seem better, this could actually slow it down.

In the current approach, each delayed hydration wrapper is made exclusively for a single purpose. This allows for maximal optimizations, as it only needs to take care of a single scenario.

In your approach, the handling of the `<Lazy*>` components would need to account for regular, non-delayed async components, as well as all possible delayed hydration strategies, all from the same place. This inherently makes it impossible to optimize for ideal performance as it needs to have broader handling that might result in something for a certain input and might result in something else for a different one.

Tofandel: > In your approach, the handling of the <Lazy*> components would need to account for regular, non-delayed async components, as well as all possible delayed hydration strategies, all from the same place. This inherently makes it impossible to optimize for ideal performance as it needs to have broader handling that might result in something for a certain input and might result in something else for a different one.

You misunderstood slightly the approach, it would be only syntactic sugar directive (it would never work as a real directive), it would actually be transformed at build time by `@vue/sfc-compiler` or `vite` and during the transformation it would not be a problem to use a different wrapper component, this means all that you did with the wrappers stays mostly the same, it only needs a transformer and the type generation is not needed anymore, we could still let the wrappers be used via imports as well in case someones want to use a wrapper component outside of a template like a fully scripted component

It would not have any performance drawback compared to this method

GalacticHypernova: Ah, I see what you're referring to. At first you stated it would be transformed at runtime, which is where the confusion arose. Doing it at build time does indeed sound better but the directive also should be unique to ensure it doesn't conflict with other libraries (I think I recall some external library using `v-hydrate`).

> My 16Gb of rams would like to disagree, I'm constantly swapping when developing with a single project taking in the 10Gb of ram

Well, that's definitely odd. Obviously I can't outright say this is wrong, but I am simply speaking from experience with an experiment that added thousands of types, and all seemed to be fine and fast. My IDE didn't appear to be slower either

Tofandel: My bad, definitely meant build time

> Doing it at build time does indeed sound better but the directive also should be unique to ensure it doesn't conflict with other libraries (I think I recall some external library using v-hydrate).

I think it shouldn't be a problem because it's a directive that we will only transform on a Lazy prefixed component, so even if naming conflicts with real directives, because of the context in which it's used is scoped to a specific nuxt feature and we will be transforming only those instances it shouldn't cause issues, and the other library might become redundant if they are providing the same feature as well

For performance it really depends on your env and your ide, they are not all created equal and some try to provide more features for the DX like AI autocompletion at the cost of performance or memory, it's a sensitive balance

In any case I think v-hydrate:how is also easier to remember and understand for the end users

nathanchase: > Let's start with the obvious: it's very rare to have 300 components in an app, no matter the size. 100? Very much possible. 200? Sure. 250?? Maybe if it covers so many things and is extremely multi purpose. But 300 components doesn't seem like an everyday occurrence. Furthermore, I don't believe having these 2700 types would be too much of a concern for people, regardless of where it's deployed, for 3 main reasons:
> 

For what it's worth, I don't consider my app _**that**_ large, and yet...
image

GalacticHypernova: > In any case I think v-hydrate:how is also easier to remember and understand for the end users

This would probably vary. Ideally, as I've stated in the past, a uniform naming would make it easier for newcomers because they won't need to figure out which option is for which strategy.

> For what it's worth, I don't consider my app that large, and yet...

Well, that's definitely a rare occurrence, but I agree there are cases that do reach it, mainly with UI libraries.

All in all, I'd have to look into the plugin approach, especially from the side of the DX with the directive intellisense, because this would be a tricky aspect that I do not want to go unaddressed.

nathanchase: > Well, that's definitely a rare occurrence, but I agree there are cases that do reach it, mainly with UI libraries.

I'm not using any UI libraries. That's just normal userland, custom Vue SFC components.

Tofandel: > a uniform naming would make it easier for newcomers because they won't need to figure out which option is for which strategy.

Which is exactly what I was advocating v-hydrate:[name-of-strategy]="propValue" achieves, because then you don't have any option name involved at all

I do think a normal sized app (website with maybe a blog or ecommerce) easilly reaches the 200 or 300 count

This is me and it's also not a big app and it's still in development so likely there will be 50 more components by the end of it
image

packages/nuxt/src/components/plugins/loader.ts
```diff
@@ -17,10 +17,11 @@
   experimentalComponentIslands?: boolean
 }
 
-const REPLACE_COMPONENT_TO_DIRECT_IMPORT_RE = /(?<=[ (])_?resolveComponent\(\s*["'](lazy-|Lazy(?=[A-Z]))?([^'"]*)["'][^)]*\)/g
+const REPLACE_COMPONENT_TO_DIRECT_IMPORT_RE = /(?<=[ (])_?resolveComponent\(\s*["'](lazy-|Lazy(?=[A-Z]))?(Idle|Visible|idle-|visible-|Event|event-|Media|media-|If|if-|Never|never-|Time|time-|Promise|promise-)?([^'"]*)["'][^)]*\)/g
 export const LoaderPlugin = (options: LoaderOptions) => createUnplugin(() => {
   const exclude = options.transform?.exclude || []
   const include = options.transform?.include || []
+  const clientDelayedComponentRuntime = resolve(distDir, 'components/runtime/client-delayed-component')
```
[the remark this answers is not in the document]
GalacticHypernova: This will be removed in the next push, as it was legacy code from before Nuxt was used the new unplugin vfs. Therefore, I am marking this as resolved now, before making the commit.

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: > v-hydrate:[name-of-strategy]="propValue"

To clarify, this part still needs the unique identifier being the strategy. I meant uniform in the prop, which is why I initially opted for the prefixes with :hydrate, as people won't need to figure out which prop is the one for time-based hydration and which one is for promise-based hydration. In either case however, the actual strategy is still unique.

On the bright side, I've been able to make a prototype implementation using a build plugin as you suggested. A demonstration is provided below:
 
image

https://github.com/user-attachments/assets/ced2be81-8e53-4b8f-9c65-ffc64c1da7a5

Tofandel: Amazing ❤️, let me know if you need help creating the type for the directive

GalacticHypernova: > I see you used a prop instead of a directive

For now, yes, because I've been looking into typing the directive. I'm also taking into account the warning not to use custom directives on components because it can have unexpected results with multiple root nodes, so there is something which I would need to carefully review.

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,175 @@
...
+        type: [String, Array],
+        required: false,
+        default: 'mouseover',
+      },
+    },
+    emits: ['hydrated'],
+    setup (props, { attrs, emit }) {
+      const hydrated = () => { emit('hydrated') }
+      // @ts-expect-error Cannot type HTMLElementEventMap in props
```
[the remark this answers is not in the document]
Tofandel: That's because HTMLElementEventMap is not a type, it's an object

You need to use `keyof HTMLElementEventMap` but in this case I would recommend to use for all the types this:

```js
        type: null as PropType<Arguments<typeof hydrateOnInteraction>[0]>,
```

So you get the type directly from the hydration methods of vue

packages/nuxt/src/components/runtime/client-delayed-component.ts
```diff
@@ -0,0 +1,175 @@
...
+  })
+}
+
+/* @__NO_SIDE_EFFECTS__ */
+export const createLazyPromiseComponent = (loader: AsyncComponentLoader) => {
+  return defineComponent({
+    inheritAttrs: false,
+    props: {
+      hydrate: {
+        type: Promise,
```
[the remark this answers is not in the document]
Tofandel: See above suggestion

packages/nuxt/src/components/runtime/client-delayed-component.ts
GalacticHypernova: Hmm, I have noticed an issue with this. In the previous approach with the component prefixes, you could simply use the prefix without any other props for the default. However, if we rework this into a prop, we end up with:
```vue
<LazyMyComponent hydrate:time />
```

If we want to allow for defaults, this means the type would need to allow boolean for each hydration strategy, otherwise a type error will occur. While it's not a big deal, I wonder if it's the best approach for this

GalacticHypernova: @Tofandel I'm not entirely sure if this is what you were looking for, but feel free to review it and let me know if there's anything else you think I should change.

I can't fully say I am extremely pleased with the implementation because there are still edge cases that can't be fixed during the build process without overly impacting the performance (for example, message above), and I would still like to sit on these issues a little and try to see if I can fix them. It's much less straightforward than when it was prefix-based, that's for sure, and I am hesitant to use an AST traversal in the plugin because it's more expensive than regex, which for the most part should suffice for the general guards, but still more performant

Tofandel: I'll have a look, but I'm quite surprised to see a regex in there, normally you already have full AST available if you build the plugin for the sfc compiler, I'll maybe see too send a PR to your fork to demonstrate

GalacticHypernova: Where do you see the full AST? When making such plugins you get the raw code as a string and would have to walk the AAT using acorn or another AST library

Tofandel: Basically you need to pass into the Vue vite plugin a function into an option under compilerOptions.nodeTransforms

GalacticHypernova: Ah, that's where the issue lies I assume. Build plugins happen before the Vue plugin is ran. You can see that other plugins in the Nuxt codebase traverse the AST on their own with Acorn rather than relying on Vue

Tofandel: You can create the vite plugin that simply edits the config of the other Vue plugin I think. Acorn is meant to parse javascript or ts, not vue templates

GalacticHypernova: Maybe, but this seems way too overcomplicated, and I'llhave to carefully test it.. I'm almost considering maybe reverting to the prefix approach and just refining it. We can make use of mapped types to easily type out all possible strats within a single type definitions which will massively reduce total types (it's actually used in here as well so it could be applied the same way). Something that would look roughly like this:

MyComponent: regular import,
LazyMyComponent: regular async import,
Lazy${k in keyof types}MyComponent: import with hydrate: ${typeMap[k]}

Would this not essentially answer your concern about total types?

Tofandel: `Lazy${k in keyof types}MyComponent: import with hydrate: ${typeMap[k]}` is this possible to do with a top level type?

Just sit tight, I'll send in a PR to your branch tomorrow night with some tweaks (like userland extensibility of the hydration methods) and the AST improvement

GalacticHypernova: Hey!

After careful consideration, I decided to revert the directive approach and return to the original, for a couple main reasons:
1. The directive approach brings a lot of complexity to the process, and would make it more difficult to maintain and extend it. There could be some unaccounted edge cases and having another AST manipulation, in my opinion, isn't worth the performance degradation, as some projects can have tens of delayed hydration components at a time in a single page, which would be quite the overhead for the AST traversal.
2. There is no concrete evidence that suggests the types are a major bottleneck. Most IDE's are already optimized for type definitions, and Microsoft themselves are already optimizing TypeScript's internal overhead, and there is no indication that it would actually cause a problem that warrants a complete build time transformation.

That being said, feel free to still send in the PR, if you believe it is worth it, and of course this could be reworked again, once everything has been planned out, to avoid bringing more harm than good. I will still help you work on the PR, if you'd like.

However, there should be some additional testing and real-world cases to accurately determine whether the types cause too major of an issue.

Hopefully you're not too upset about my decision, and of course I would love to hear any insight you may have.

GalacticHypernova: > we can apply transforms to the default slot to make all the components inside lazy and import the slot on demand

I haven't really tried that, but do you think as it currently is won't allow for slots?

GalacticHypernova: Just letting you know, I cloned the original branch to patch-21-backup so we could continue working on it. In the meantime, I believe some functional implementation is better than no implementation, especially with the performance considerations the current approach takes care of. We could always make another PR to refactor the delayed hydration process, but I didn't want this to be a road block from implementing it.

GalacticHypernova:
It once again seems to work fine in local tests but fails in job runners... That is very odd.

GalacticHypernova:
I decided to keep this prefix approach for now as I review a way to simplify it. I figured it would be better to have an initial well tested and proven approach with minimal overhead (as this directly corresponds to the components that would be used and it doesn't need to go through another build plugin), as opposed to choosing a less thoroughly tested approach which I couldn't prove made any real difference. Of course, we could reiterate this and mature the implementation as time goes on, which is why this is still in experimental.

GalacticHypernova:
Hey @mkierdev !

The PR, in its current implementation, is pretty much ready. As I stated, I will look into ways simplify the integration for end-developers but I've gotten pretty busy with personal matters and haven't had adequate time to properly review any other implementation strategy, and I'd much rather provide a slightly more direct implementation (such as the one I used here, with each hydration strategy corresponding to its own prefix) than to make something potentially better in some aspects but wasn't thoroughly tested.

danielroe:
I'm looking to merge this for v3.16, but I plan to refactor slightly so that we take something that looks like a directive rather than generating multiple component names as at present.

(@GalacticHypernova - no need to take action, I've got this final small refactor.)

Tofandel:
@danielroe I already started a big cleanup and directive approach in my branch of this PR if you want to take a look, I'm also quite busy right now so didn't have time to finish it up

Tofandel:
What is missing in the branch

1. It needs to be rebased/merged with the updates (mostly the tests added)
2. I wanted to use the node transform to add the imports directly in the vue file instead of just replacing the component to the pascal case version and then using the regex

danielroe:
Current TODOs:

- [ ] ensure we do not preload/prefetch chunks for components that are lazily hydrated
- [ ] reduce type augmentation - look at directive- or prop-style approach
- [ ] update tests to use locators + avoid hard coding timings

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -123,6 +123,158 @@ const show = ref(false)
...
+2. **Conditional Rendering:** When using `v-if="false"` on a lazy component, you might not need delayed hydration. You can use a delayed hydration prop.
+
+3. **Shared State:** Be mindful of shared state (`v-model`) across multiple components. Updating the model in one component can trigger hydration in all components bound to that model.
+
+4. **Use Each Strategy's Intended Use Case:** Each strategy is optimized for a specific purpose.
+    * `hydrate-when` is best for components that might not always need to be hydrated.
+    * `hydrate-after` is for components that can wait a specific amount of time.
+    * `hydrate-on-idle` is for components that can be hydrated when the browser is idle.
+
+5. **Avoid `LazyNever` on interactive components:** If a component requires user interaction, it should not be set to never hydrate.
```
cernymatej, line 276: > You can use a delayed hydration prop.

What exactly does this mean in the `Conditional Rendering` section? I find it a bit confusing, especially since the previous sentence states, "you might not need delayed hydration."  

> 5. **Avoid `LazyNever` on interactive components:** If a component requires user interaction, it should not be set to never hydrate.

this might be a remnant from the previous implementation

packages/nuxt/src/components/plugins/loader.ts
```diff
@@ -14,103 +15,243 @@ interface LoaderOptions {
...
+          await walk(ast, (node) => {
+            if (node.type !== 1 /* ELEMENT_NODE */) {
+              return
+            }
+            if (!/^(?:Lazy|lazy-)/.test(node.name)) {
+              return
+            }
+            const pascalName = pascalCase(node.name.slice(4))
+            if (!components.some(c => c.pascalName === pascalName)) {
+              // not auto-imported
```
TheAlexLichter, line 74: Does that meant that lazy hydration will only work for auto-imported components? 🤔

TheAlexLichter: I guess that should be documented then (or did I miss that part?)

Hmm, maybe analyzing the import and creating a wrapper for the imported component "on the fly"? 🤔

danielroe: I think that would be the right solution 👍

danielroe (approved):

PR #26480, opened 2024-03-25
fix(nuxt): check vnode element instead of hydration state for islands

### LINKED ISSUE #26434: Server components cause hydration error when used inside resolved component
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.11.1
- CLI Version:      3.11.1
- Nitro Version:    2.9.4
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-v2dhfq

### Describe the bug

Since Nuxt version 3.10 server components cause an hydration mismatch and don't render when used inside resolved components.

### Additional context

_No response_

### Logs

```shell-script
[Vue warn]: Hydration node mismatch:
- rendered on server: <!--[--> (start of fragment) 
- expected on client: div 
  at <NuxtIsland name="Icon" lazy=false props= Proxy(Object)  ... > 
  at <Icon> 
  at <Container.global> 
  at <AsyncComponentWrapper> 
  at <App key=3 > 
  at <NuxtRoot>
```

huang-julien:
### 🔗 Linked issue
fix #26434

### ❓ Type of change

### 📚 Description

Hi :wave: this PR prefer using `vnode.el` instead of `nuxtApp.isHydrating`

Not sure why we do have this issue with `resolveComponent`.

This PR however fix the case where nuxtIsland is called within an async component

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -109,14 +109,14 @@ export default defineComponent({
       components: {}
     }
 
-    if (nuxtApp.isHydrating) {
+    if (instance.vnode.el) {
       payloads.slots = toRaw(nuxtApp.payload.data[`${props.name}_${hashId.value}`])?.slots ?? {}
       payloads.components = toRaw(nuxtApp.payload.data[`${props.name}_${hashId.value}`])?.components ?? {}
     }
 
     const ssrHTML = ref<string>('')
 
-    if (import.meta.client && nuxtApp.isHydrating) {
+    if (import.meta.client && instance.vnode.el) {
       ssrHTML.value = getFragmentHTML(instance.vnode?.el ?? null, true)?.join('') || ''
       const key = `${props.name}_${hashId.value}`
       nuxtApp.payload.data[key] ||= {}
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -109,14 +109,14 @@ export default defineComponent({
...
+    if (instance.vnode.el) {
       payloads.slots = toRaw(nuxtApp.payload.data[`${props.name}_${hashId.value}`])?.slots ?? {}
       payloads.components = toRaw(nuxtApp.payload.data[`${props.name}_${hashId.value}`])?.components ?? {}
     }
 
     const ssrHTML = ref<string>('')
 
-    if (import.meta.client && nuxtApp.isHydrating) {
+    if (import.meta.client && instance.vnode.el) {
       ssrHTML.value = getFragmentHTML(instance.vnode?.el ?? null, true)?.join('') || ''
```
danielroe, line 120:
```suggestion
    if (import.meta.client && instance.vnode?.el) {
      ssrHTML.value = getFragmentHTML(instance.vnode.el, true)?.join('') || ''
```

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -250,7 +250,7 @@ export default defineComponent({
           // this is used to force trigger Teleport when vue makes the diff between old and new node
           const isKeyOdd = teleportKey.value === 0 || !!(teleportKey.value && !(teleportKey.value % 2))
 
-          if (uid.value && html.value && (import.meta.server || props.lazy ? canTeleport : mounted.value || nuxtApp.isHydrating)) {
+          if (uid.value && html.value && (import.meta.server || props.lazy ? canTeleport : mounted.value || instance.vnode.el)) {
```
danielroe, line 253:
```suggestion
          if (uid.value && html.value && (import.meta.server || props.lazy ? canTeleport : mounted.value || instance.vnode?.el)) {
```

danielroe:
Might be the same issue as https://github.com/vuejs/core/issues/6638.

huang-julien:
@danielroe it's probably the same issue. Is it better to let vue/core fix it ?

danielroe:
I think this is a good fix in the meantime. But fixing upstream would be 👌

danielroe (approved):

PR #26481, opened 2024-03-26
docs: add a note for https call with self-signed certificate

maximepvrt:
### ❓ Type of change

### 📝 Checklist


docs/3.api/2.composables/use-fetch.md
```diff
@@ -72,6 +72,10 @@ const { data, pending, error, refresh } = await useFetch('/api/auth/login', {
 
 :link-example{to="/docs/examples/features/data-fetching"}
 
+::note
+If you use `useFetch` to call a https URL with a self-signed certificate in development, you will need to set `NODE_TLS_REJECT_UNAUTHORIZED=0` in your environment.
```
danielroe, line 76: This is true but also relevant for `$fetch` - should this be on the data fetching page or do you think it should be true here too?

Also, maybe we could support a more verbose log with this information in dev mode when this is the error...

maximepvrt: Certainly! I believe it's essential to maintain this information within the useFetch since it primarily initiates server calls (with SSR). While I had considered adding it to $fetch.

However, I agree that having a more verbose logging mechanism, especially in development mode, would be beneficial for tracking HTTP calls made by the server. This could greatly assist in debugging and understanding the causes of any call failures.

maximepvrt: @danielroe

danielroe (approved):

PR #26519, opened 2024-03-27
perf(nuxt): only resolve initial route once

### LINKED ISSUE #26425: When non-existing URL is visited then "No match found for location with path" is logged 8 times
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.11.1
- CLI Version:      3.11.0
- Nitro Version:    2.9.4
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-bejpvy

### Describe the bug

On clean installation of `Nuxt 3.11.1` I just added `pages` directory with `blog.vue`. When you visit wrong URL (for example `"/"` in that case) then "No match found for location" is logged exactly 8 times.

I suspect that this bug causes CPU leak on my personal project as I have many non-existing URLs crawled by bots.

### Additional context

There is closed issue #23539 that was probably not resolved by https://github.com/nuxt/nuxt/pull/25390.

### Logs

```shell-script
WARN [Vue Router warn]: No match found for location with path "/"
```

### LINKED ISSUE #26499: Nuxt router calls resolve from vue-router twice
### Environment

- Operating System: Darwin
- Node Version:     v20.11.1
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.9.1

### Reproduction

Any nuxt with any routes

### Describe the bug

The router of nuxt calls resolve twice when resolving a location
https://github.com/nuxt/nuxt/blob/3522c9867779bfc10b319f0947a740b733645c03/packages/nuxt/src/pages/runtime/plugins/router.ts#L115
and again when using router.push
https://github.com/nuxt/nuxt/blob/3522c9867779bfc10b319f0947a740b733645c03/packages/nuxt/src/pages/runtime/plugins/router.ts#L145

### Additional context

In environments with a lot of routes the resolve takes some amount of time, e.g. in my current environment around 100ms.

The second call to resolve is therefore very costy

If I read the code from vue-router correctly it is also possible to pass and already resolved route and it will skip the heavy part where it matches the regexes.

Is there any reason for this ? I did not create a pull request because I'm still lacking the knowledge a little bit on how to do it better, but I think especially in big environments it is crucial to reduce the calls to those functions in the router

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26499
resolves https://github.com/nuxt/nuxt/issues/26425

### ❓ Type of change

### 📚 Description

This seeks to improve performance by not calling `router.resolve` as often.

~~The one area I'm a bit worried about and would welcome testing on is whether this makes it tricky to resolve routes that are added dynamically by plugins (as the final `router.replace` exists to run middleware 'freshly' on the resolved route).~~

Having checked, this still works in the same manner when adding routes within plugins. There will (still) be warnings printed but it should continue to work.

When hitting a 404 on a route (not from a browser) (such as an API request 404), we will now log two warnings:

https://github.com/nuxt/nuxt/blob/789bb6b1a66d430c7d364381cf2380b2d9c030af/packages/nuxt/src/pages/runtime/plugins/router.ts#L116-L118
https://github.com/nuxt/nuxt/blob/789bb6b1a66d430c7d364381cf2380b2d9c030af/packages/nuxt/src/pages/runtime/plugins/router.ts#L263-L266

From a browser, we'll get four (2x for the initial request, and then 2x for the error page we render and send back instead), from the same lines.

### 📝 Checklist


manniL:
@danielroe Do you have any modules in mind which add routes dynamically? 🤔

danielroe:
Here's a list of public repos:

https://github.com/search?q=/router%5C.addRoutes/&type=code

danielroe:
Having checked, this still works in the same manner when adding routes within plugins. There will (still) be warnings printed but it should continue to work.

When hitting a 404 on a route (not from a browser) (such as an API request 404), we will now log two warnings:

https://github.com/nuxt/nuxt/blob/789bb6b1a66d430c7d364381cf2380b2d9c030af/packages/nuxt/src/pages/runtime/plugins/router.ts#L116-L118
https://github.com/nuxt/nuxt/blob/789bb6b1a66d430c7d364381cf2380b2d9c030af/packages/nuxt/src/pages/runtime/plugins/router.ts#L263-L266

From a browser, we'll get four (2x for the initial request, and then 2x for the error page we render and send back instead), from the same lines.

PR #26522, opened 2024-03-27
feat: add `useLink` to `NuxtLink`

### LINKED ISSUE #22169: Provide RouterLink-compatible `useLink` method
### Describe the feature

https://github.com/vuetifyjs/vuetify/issues/17490

vue-router has a method call `RouterLink.useLink`, but it doesn't exist in `NuxtLink`.

### Additional information

### Final checks

userquin:
### 🔗 Linked issue

resolves #22169

### ❓ Type of change

### 📚 Description

Check title: I need to add a test and include `useLink` type

### 📝 Checklist


userquin:
We need to move the second test to some fixture

vitest.nuxt.config.ts
```diff
@@ -1,6 +1,11 @@
 import { defineVitestConfig } from '@nuxt/test-utils/config'
 
 export default defineVitestConfig({
+  resolve: {
+    alias: {
+      '#vue-router': 'vue-router'
```
danielroe, line 6: This shouldn't be needed, I think...

userquin: check last failing commit: https://github.com/nuxt/nuxt/pull/26522/commits/302d91b3c78f548d0587e2e05b7b34c11210d0f0

danielroe: The reason this is failing is that we also have a version of Nuxt where there _is no_ vue-router. This 'universal' router needs to work when there is no `pages/` directory, so we can't directly import from or depend on vue-router in that case. We need to have a 'stub' version of `useLink` in that case...

danielroe: implementing - will push shortly

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -207,6 +249,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
         required: false,
       },
     },
+    useLink: (props: UseLinkOptions) => useNuxtLink(props as NuxtLinkProps),
     setup (props, { slots }) {
```
danielroe, line 253: ideally we should not only expose this but also use `useLink` internally within the setup function, to reduce code duplication and ensure the logic remains the same

vitest.nuxt.config.ts
danielroe: the thing I pushed is an example implementation - but as you refactor to share logic across NuxtLink and useLink, you might find it needs to be updated.

packages/nuxt/src/app/components/nuxt-link.ts
userquin: yeah, check linked vuetify PR: https://github.com/vuetifyjs/vuetify/pull/19514

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -207,43 +280,11 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
         required: false,
       },
     },
+    useLink: useNuxtLink,
```
harlan-zw, line 283: What does this do?

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -120,6 +120,79 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
     return resolvedPath
   }
 
+  function useNuxtLink (props: NuxtLinkProps) {
```
harlan-zw, line 123: Just FYI this logic has some issues, see https://github.com/nuxt/nuxt/issues/25532

I wouldn't consider fixing them as part of this PR but it will likely be updated in the future.

harlan-zw:
Looks good, nice job. Didn't sanity check all of the logic but if tests are passing it should be good :shrug:

packages/nuxt/src/app/components/nuxt-link.ts
danielroe: This matches `RouterLink` behaviour (which exposes a 'bound' composable which other libraries, like vuetify, can use to create custom links).

packages/nuxt/src/app/components/nuxt-link.ts
danielroe: Exactly. ❤️  Just wanted to  check that this implementation would be compatible with the tree-shaking/bundle size improvements you have in mind.

PR #26558, opened 2024-03-29
refactor(nuxt): simplify `runtimeConfig` initialization of client side

Mini-ghost:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

Currently, when we initialize `nuxtApp.$config` on the client side, we use `reactive(nuxtApp.payload.config!)`.

https://github.com/nuxt/nuxt/blob/50c92fbe6cfb5ce4d5bd51ffdbc7f6da2b1ddf65/packages/nuxt/src/app/nuxt.ts#L340-L342

However, `nuxtApp.payload` is set to `reactive` during initialization, meaning `nuxtApp.payload.config` is already a reactive object on the client side, so it does not need to be processed with `reactive` again.

https://github.com/nuxt/nuxt/blob/50c92fbe6cfb5ce4d5bd51ffdbc7f6da2b1ddf65/packages/nuxt/src/app/nuxt.ts#L241-L247

Thank you for reviewing this. If there's anything I might have overlooked, please don't hesitate to let me know.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -338,7 +338,7 @@ export function createNuxtApp (options: CreateOptions) {
   }
 
   // Expose runtime config
-  const runtimeConfig = import.meta.server ? options.ssrContext!.runtimeConfig : reactive(nuxtApp.payload.config!)
+  const runtimeConfig = import.meta.server ? options.ssrContext!.runtimeConfig : nuxtApp.payload.config!
   nuxtApp.provide('config', runtimeConfig)
 
   return nuxtApp


danielroe (approved):
This is good. ❤️ However, a note that in v4 we will likely remove the deep reactivity of the payload and this will likely need to return.

Mini-ghost:
Could I please understand why `config` needs to be reactivity? Isn't it used for storing data like from `defineNuxtConfig` and `defineAppConfig`?

Mini-ghost:
@danielroe 

Thank you for the heads up. I was actually hoping to propose removing the reactivity of `config` in v4, as I've observed that the data stored there tends to be more like environment variables or settings for things like Nuxt Content, which I guess don't require reactivity.

Therefore, I'm not quite sure in what scenarios the reactivity of `config` would be utilized. Could you please provide some guidance on this matter? I would greatly appreciate it 🥹

PR #26576, opened 2024-03-30
feat(nuxt): introduce `imports.scan` option

### LINKED ISSUE #26525: Add feature toggle to disable autoImports
### Describe the feature

Hello, thank you very much for the product you make. 

When working with `nuxt`, I don't really like the concept of auto-imports and would like to disable it. I have several reasons why I want to do this:
1) Auto imports are secret magic. It’s not entirely clear what you are importing and from where
2) I use `jest` in my project and there are big problems with it when setting up mocks for `#imports`

I saw a previous [issue](https://github.com/nuxt/nuxt/issues/14505) on this topic, but as I see it did not end in anything. Currently, there are settings to disable auto-imports of components, but there are no settings to disable other auto-imports: `vue` re-export, `@unhead/vue` re-export, `composables` folder re-export in a project, `utils` folder re-export in a project.

I would like to have some setting that allows me to disable all re-exports from `#imports` except `nuxt` re-exports.

### Additional information

### Final checks

antfu:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26525

### 📚 Description

I think auto imports for local composables are the main source of implicitness we have for auto imports. 

This PR introduced the `imports.scan: false` option to disable that feature while keeping the explicitly registered auto imports like `vue` or module-contributed imports. (The naming could be better tho, open for suggestions)

It's also possible for each layer to opt out of the registry of its own composable folders.

I personally would even vote to disable the scanning by default in Nuxt 4 (where I also made https://github.com/antfu/nuxt-eslint-auto-explicit-import, which could be useful for migration)  - but let's discuss that in the next team meeting


danielroe (approved):

packages/nuxt/src/imports/module.ts
```diff
@@ -51,30 +52,36 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
           continue
         }
-        composablesDirs.push(resolve(layer.config.srcDir, dir))
+        composablesDirs.push(resolve(layer.config.srcDir, 'composables'))
+        composablesDirs.push(resolve(layer.config.srcDir, 'utils'))
+        for (const dir of (layer.config.imports?.dirs ?? [])) {
+          if (!dir) {
+            continue
+          }
+          composablesDirs.push(resolve(layer.config.srcDir, dir))
```
userquin, line 67: how about nuxt aliases?

danielroe: what are you thinking?

PR #26589, opened 2024-03-31
fix(nuxt): remove cast on rollup input for island chunks

### LINKED ISSUE #26220: TypeError in `islandsTransform.ts` affecting storybook
### Environment

Nuxt project info: 

- Operating System: Linux
- Node Version:     v20.11.1
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.9.3
- Package Manager:  pnpm@8.15.4
- Builder:          -
- User Config:      runtimeConfig, modules, $test, devtools, experimental, vue, tailwindcss, app, nitro, typescript, pinia
- Runtime Modules:  @nuxtjs/tailwindcss@6.11.5-1709108338.2ae28e5, @vueuse/nuxt@10.9.0, nuxt-headlessui@1.1.5, @pinia/nuxt@0.5.1, @nuxtjs/fontaine@0.4.1, @formkit/auto-animate/nuxt@0.8.1
- Build Modules:    -

👉 Report an issue: https://github.com/nuxt/nuxt/issues/new

👉 Suggest an improvement: https://github.com/nuxt/nuxt/discussions/new

👉 Read documentation: https://nuxt.com



### Reproduction

On a project using NuxtIslands (via server components), and storybook, running the `build-storybook` command fails with the following error:

```
=> Failed to build the preview                                                                                                                                                                                                            10:43:12 AM
TypeError: Cannot create property 'ActionMessage' on string './node_modules/.pnpm/@storybook+builder-vite@7.6.17_typescript@5.4.2_vite@5.1.6/node_modules/@storybook/builder-vite/input/iframe.html'                                      10:43:12 AM
    at config (file://./node_modules/.pnpm/nuxt@3.10.3_@types+node@18.19.23_eslint@8.57.0_typescript@5.4.2_vite@5.1.6_vue-tsc@1.8.27/node_modules/nuxt/dist/index.mjs:1866:62)
```

### Describe the bug

In `packages/nuxt/src/components/islandsTransform.ts`, the `componentsChunkPlugin` function makes the following type assertion ` (config.build.rollupOptions.input as Record<string, string>)`.


```typescript
export const componentsChunkPlugin = createUnplugin((options: ComponentChunkOptions) => {
  const { buildDir } = options
  return {
    name: 'componentsChunkPlugin',
    vite: {
      async config (config) {
        const components = options.getComponents()
        config.build = config.build || {}
        config.build.rollupOptions = config.build.rollupOptions || {}
        config.build.rollupOptions.output = config.build.rollupOptions.output || {}
        config.build.rollupOptions.input = config.build.rollupOptions.input || {}
        // don't use 'strict', this would create another "facade" chunk for the entry file, causing the ssr styles to not detect everything
        config.build.rollupOptions.preserveEntrySignatures = 'allow-extension'
        for (const component of components) {
          if (component.mode === 'client' || component.mode === 'all') {
            (config.build.rollupOptions.input as Record<string, string>)[component.pascalName] = await resolvePath(component.filePath)
          }
        }
      },
```

`config.build.rollupOptions.input` is declared as `type InputOption = string | string[] | { [entryAlias: string]: string }`. When using storybook and `@storybook/builder-vite`, `config.build.rollupOptions.input` is a `string`, with a value similar to `/node_modules/.../@storybook/builder-vite/input/iframe.html`.

Would a PR that checks first for the type of `config.build.rollupOptions.input` make sense? Something along the lines of:

```typescript
   if (component.mode === 'client' || component.mode === 'all') {
      if ('object' === typeof config.build.rollupOptions.input && !Array.isArray(config.build.rollupOptions.input))      
        (config.build.rollupOptions.input as Record<string, string>)[component.pascalName] = await resolvePath(component.filePath)
```

### Additional context

Disabling server components is a workaround that fixes the issue when running or building storybook. I have them enabled when building the app, but disabled when building statically with storybook.

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix #26220

### 📚 Description

Hey :wave: this PR removes the cast on the rollup input config.


packages/nuxt/src/components/islandsTransform.ts
```diff
@@ -188,9 +188,21 @@ export const componentsChunkPlugin = createUnplugin((options: ComponentChunkOpti
...
         config.build.rollupOptions.preserveEntrySignatures = 'allow-extension'
+        const { rollupOptions } = config.build
+
+        // convert input to object syntax
+        if (!rollupOptions.input) { rollupOptions.input = {} }
+
+        if (typeof rollupOptions.input === 'string') {
+          rollupOptions.input = { [rollupOptions.input]: rollupOptions.input }
+        } else if (typeof rollupOptions.input === 'object' && Array.isArray(rollupOptions.input)) {
+          rollupOptions.input = rollupOptions.input.reduce<{ [key: string]: string }>((acc, input) => { acc[input] = input; return acc }, {})
```
huang-julien, line 199: I'm absolutely unsure about this conversion.

danielroe (commented):
We know for sure that `rollupOptions.input` must be a Record because we set it in the first place:

https://github.com/nuxt/nuxt/blob/8c3159f4ba506fd40cac3252b81bd0aa9c51f8b6/packages/vite/src/server.ts#L82

If that's not the case then Nuxt will fail to build.

It feels like this normalisation is something that should happen in the storybook module rather than in Nuxt.

huang-julien:
Oooops I linked the wrong issue 🤦 . 

So you think storybook should send a record instead of a string ?

PR #26602, opened 2024-04-02
docs: add usePreviewMode explanation

### LINKED ISSUE #26351: 📚 Docs: explain what preview mode is
People asked today on Discord what [preview mode](https://nuxt.com/docs/api/composables/use-preview-mode) actually is. I think the documentation could be extended with a description, and maybe a complete example.

adam-hudak:
### 🔗 Linked issue
resolves #26351 

### ❓ Type of change
* [x]  📖 Documentation (updates to the documentation, readme or JSdoc annotations)
* [ ]  🐞 Bug fix (a non-breaking change that fixes an issue)
* [ ]  👌 Enhancement (improving an existing functionality like performance)
* [ ]  ✨ New feature (a non-breaking change that adds functionality)
* [ ]  🧹 Chore (updates to the build process or auxiliary tools and libraries)
* [ ]  ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
Add clarification for usePreviewMode composable

### 📝 Checklist
* [x]  I have linked an issue or discussion.
* [ ]  I have added tests (if possible).
* [x]  I have updated the documentation accordingly.


docs/3.api/2.composables/use-preview-mode.md
```diff
@@ -5,12 +5,63 @@ description: "Use usePreviewMode to check and control preview mode in Nuxt"
...
+```
+
+Then you can see your preview page by adding the query param defined in `shouldEnable` to the end of the page you want to see once:
+
+```js
+?customPreview=true
+```
+
+::note
+usePreviewMode should be tested locally with nuxi generate and not yarn dev
```
danielroe, line 62:
```suggestion
`usePreviewMode` should be tested locally with `nuxi generate` and then `nuxi preview` rather than `nuxi dev`.
```

docs/3.api/2.composables/use-preview-mode.md
```diff
@@ -5,12 +5,63 @@ description: "Use usePreviewMode to check and control preview mode in Nuxt"
...
+
+The example below creates a page where part of a content is rendered only in `Preview mode`.
+
+```vue [pages/some-page.vue]
+<script setup>
+const route = useRoute()
+const { enabled, state } = usePreviewMode({
+  shouldEnable: () => {
+    return route.query.customPreview === 'true'
+  },
```
danielroe, line 26: maybe it would be worth leaving the example below as an example of `shouldEnable` and modify this one so it shows how it works out of the box, without `shouldEnable`?

docs/3.api/2.composables/use-preview-mode.md
adam-hudak: Thank you for your suggestion. I will update the note, but I will use **nuxi start**, because for junior developer, it might seem that **nuxi preview** is somehow related to usePreviewMode and might be surprised why they don't see preview code right away

docs/3.api/2.composables/use-preview-mode.md
adam-hudak: I've been thinking about this while making my PR and the idea was to show in the example as much as possible. But I understand that might be overwhelming. I will update the example. Thank you

adam-hudak: I think it's not necessary to have whole example at the bottom, because there already is an example  for `shouldEnable` :

```
export function useMyPreviewMode () {
  return usePreviewMode({
    shouldEnable: () => {
      return !!route.query.customPreview
    }
  });
}
```

danielroe: sounds good

danielroe (approved):

danielroe (approved):

PR #26583, opened 2024-03-30
chore: migrate to ESLint flat config

luc122c:
### 🔗 Linked issue
Resolves #25351

### 📚 Description

As discussed in #25351, eslint are upgrading their config system in eslint v9. The new syntax is available now but is opt-in. We can opt in by replacing `.eslintrc` with a `eslint.config.js` file or setting the environment variable `ESLINT_USE_FLAT_CONFIG=true` before running eslint.

I have followed eslint's [configuration migration guide](https://eslint.org/docs/latest/use/configure/migration-guide) to replace the configuration. This PR is intended to be a straight swap, not modifying the configuration in any way, just upgrading the syntax. There may be additional work needed to take advantage of the benefits the new syntax brings. 

### 🧪 How to test

1. Pull the branch
2. Install dependencies
3. Run `ESLINT_USE_FLAT_CONFIG=true pnpm run lint`

### ⬇️ Dependencies 

Nuxt depends on several configs and plugins. Below is the status of flat config support for each dependency. ([Data sourced from tracker](https://github.com/eslint/eslint/issues/18093))

| Name | Support | Issue/PR |
| ------ | -------- | --------- |
| [jsdoc](https://github.com//gajus/eslint-plugin-jsdoc) | ✅ | https://github.com/gajus/eslint-plugin-jsdoc/pull/1177 |
| import | ? |  https://github.com/import-js/eslint-plugin-import/pull/2714 |
[unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn) | ✅ | https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1886 |
| [no-only-tests](https://github.com/levibuzolic/eslint-plugin-no-only-tests) | ? | 
| [@nuxt/eslint-config](https://github.com/nuxt/eslint/) | 🚧 | https://github.com/nuxt/eslint/releases/tag/v0.3.0-beta.7 | 
| [eslint-config-standard](https://github.com/standard/eslint-config-standard) | 📦 🛑 | https://github.com/standard/eslint-config-standard/issues/351 | 

### ⚠️ Issues

Current issues:
- Standard needs to release a new package
- @nuxt/eslint-config is still in beta

PR #26653, opened 2024-04-05
chore: cleanup eslint rules with latest `@nuxt/eslint-config`

antfu:
### 🔗 Linked issue

Continues #26583

### 📚 Description

- Updated to the latest `@nuxt/eslint-config`, which contains plugins like jsdoc and unicorn.
- Remove standard config, move to ESLint Stylistic for future proof
- Cleanup eslint.config, make it more readable and extensible
- Disable a few rules to align more with the original behavior and keep the changes minimal.

In the meantime, I guess it could be a good chance for us to review some rules and do the linting once?


eslint.config.mjs
```diff
@@ -1,280 +1,221 @@
...
-    languageOptions: {
-      globals: {
-        NodeJS: 'readonly',
-        $fetch: 'readonly'
-      }
+export default createConfigForNuxt({
+  features: {
+    stylistic: {
+      // TODO: Anthony highly suggests to set this to 'always'
+      commaDangle: 'never'
```
antfu, line 15: I would highly recommend enabling comma dangling, to have a more stable diff and would save time on working with array items.

eslint.config.mjs
```diff
@@ -1,280 +1,221 @@
...
-        }
+      '@typescript-eslint/triple-slash-reference': 'off',
+      '@typescript-eslint/unified-signatures': 'off',
+      ...{
+        // TODO: Discuss if we want to enable this
+        '@typescript-eslint/ban-types': 'off',
+        // TODO: Discuss if we want to enable this
+        '@typescript-eslint/no-explicit-any': 'off',
+        // TODO: Discuss if we want to enable this
+        '@typescript-eslint/no-invalid-void-type': 'off'
```
antfu, line 88: These are the rules that creating quite a few warnings - we might want to enable them eventually but would requires some updates to the code.

eslint.config.mjs
```diff
@@ -1,280 +1,221 @@
...
-      'vue/valid-v-for': 'off'
+      // TODO: Discuss if we want to enable this
+      '@stylistic/brace-style': 'off',
+      '@stylistic/indent-binary-ops': 'off',
+      '@stylistic/max-statements-per-line': 'off',
+      '@stylistic/operator-linebreak': 'off',
+      // TODO: Anthony suggests to set this to 'consistent'
+      '@stylistic/quote-props': ['error', 'as-needed'],
+      // TODO: Should we enable this and set to 'never'?
+      '@stylistic/space-before-function-paren': 'off'
```
antfu, line 117: There are a few stylistic rules to discuss; they are not very important, tho.

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "skipLibCheck": true,
     "target": "ESNext",
     "module": "ESNext",
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
DamianGlowala, line 9: Is this relevant? https://x.com/mattpocockuk/status/1775138087443595689

danielroe (approved):

tsconfig.json
antfu: Kind of yes, it was because `node` does not resolve types for `@nuxt/eslint-config/flat` - while I can fix it on the plugin side, I think it's more future-proof to change to Bundler. I could send another PR to do it

eslint.config.mjs
danielroe: I picked some choices here - we can also discuss later but merging to avoid pending this pr 🙏

PR #26676, opened 2024-04-06
docs: remove step of installing deps in new nuxt project

RollingTL:
Dependencies are installed by Nuxi init.


docs/1.getting-started/2.installation.md
```diff
@@ -63,32 +63,8 @@ Or change directory into your new project from your terminal:
 cd <project-name>
 ```
 
-Install the dependencies:
+You don't need to install dependencies, they are already installed.
```
danielroe, line 67:
```suggestion
```

danielroe (approved):

PR #26680, opened 2024-04-06
fix(nuxt): re-initialise `nuxt._ignore` after all modules run

ineshbose:
### 📚 Description

(I raised this in Discord just about an hour ago, thought I'd make a PR anyway)

The `nuxt.options.ignore` is not modifiable by Nuxt modules as it is resolved and populated before the modules are installed and `modules:done` is called. This is because [we are resolving layer modules using `resolveFiles` which calls `isIgnored` that populates `nuxt._ignore` if already not populated](https://github.com/nuxt/nuxt/blob/bfb7eeac078be435acc713ce151ca3846a6d9f29/packages/nuxt/src/core/nuxt.ts#L259). So, when the modules are installed and any of them modify `nuxt.options.ignore`, the `!nuxt._ignore` condition is never true and options are not updated. This fix will check if `nuxt._ignorePattern` (as `nuxt._ignore` is of different type) is outdated from `nuxt.options.ignore` and re-populate accordingly. Added tests as well.


packages/kit/src/ignore.ts
```diff
@@ -1,7 +1,9 @@
 import { existsSync, readFileSync } from 'node:fs'
 import ignore from 'ignore'
 import { join, relative, resolve } from 'pathe'
-import { tryUseNuxt } from './context'
+import { tryUseNuxt, useNuxt } from './context'
+
+const checkIgnoreOutdated = (nuxt = useNuxt()) => nuxt._ignorePatterns?.join(',') !== nuxt.options.ignore.join(',')
```
danielroe, line 6: This will always return false, right? It doesn't resolve the patterns (e.g. `.{js,mjs}`).

ineshbose: Oh, you're quite right; forgot to consider that - good spot! Fixed 👍

packages/kit/src/ignore.ts
```diff
@@ -1,7 +1,9 @@
 import { existsSync, readFileSync } from 'node:fs'
 import ignore from 'ignore'
 import { join, relative, resolve } from 'pathe'
-import { tryUseNuxt } from './context'
+import { tryUseNuxt, useNuxt } from './context'
+
+const checkIgnoreOutdated = (nuxt = useNuxt()) => !nuxt._ignorePatterns?.join(',').startsWith(nuxt.options.ignore.flatMap(resolveGroupSyntax).join(','))
```
danielroe, line 6: I think we need a more performant check as `isIgnored` will be called _lots_ of times - once for every single file that globby or chokidar scans.

ineshbose: very good point - made improvements now

danielroe (commented):
The more I think about this, the more I think we should initiate nuxt ignorer after modules have finished running (maybe directly before build?) - and we should no longer check the contents of `nuxt.options.ignore`.

manniL:
@danielroe you mean for nuxt 4?

danielroe:
I think we can do it in a non breaking way for Nuxt 3.

danielroe (approved):

PR #26687, opened 2024-04-07
feat(nuxt): add `opts` param to `useRequestURL`

### LINKED ISSUE #26674: useRequestURL doesn't accept options of h3 getRequestURL
### Environment

Nuxt >= 3.5.0

### Reproduction

–

### Describe the bug

`useRequestURL()` doesn't accept options offered by h3 `getRequestURL()` (`xForwardedHost: boolean` and `xForwardedProto: boolean`). Esp. the `xForwardedHost` option is imho vital if Nuxt is run behind a reverse proxy (we're using Varnish in front of Nuxt), otherwise the URLs produced during SSR are different than those returned on client side.

### Additional context

_No response_

### Logs

_No response_

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26674

### 📚 Description

Adds `opts` param to `useRequestURL` to align with H3.

### DIFF
--- a/packages/nuxt/src/app/composables/url.ts
+++ b/packages/nuxt/src/app/composables/url.ts
@@ -2,10 +2,7 @@ import { getRequestURL } from 'h3'
 import { useRequestEvent } from './ssr'
 
 /** @since 3.5.0 */
-export function useRequestURL (opts?: {
-  xForwardedHost?: boolean
-  xForwardedProto?: boolean
-}) {
+export function useRequestURL (opts?: Parameters<typeof getRequestURL>[1]) {
   if (import.meta.server) {
     return getRequestURL(useRequestEvent()!, opts)
   }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/url.ts
```diff
@@ -2,9 +2,12 @@ import { getRequestURL } from 'h3'
 import { useRequestEvent } from './ssr'
 
 /** @since 3.5.0 */
-export function useRequestURL () {
+export function useRequestURL (opts?: {
+  xForwardedHost?: boolean
+  xForwardedProto?: boolean
+}) {
```
danielroe, line 8: we could probably use the type directly from `getRequestURL` in this case given that it's a pure wrapper.

DamianGlowala: updated 👍

danielroe (approved):

PR #26694, opened 2024-04-09
fix(nuxt): use runtime routes to hint more accurate routes to prerender

### LINKED ISSUE #25594: Prerendering fails if file-based routing is disabled, but there are .vue files in pages directory
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-sgbtc2

```shell
npm run generate
```

### Describe the bug

During prerendering, it tries to pre-render pages corresponding to `.vue` files in `/pages` directory, but since the server is responding with 404 status code, the whole process fails.

This can be cured by adding all the pseudo-routes to `nitro.prerender.ignore` array of nuxt config file.

### Additional context

_No response_

### Logs

_No response_

RyanClementsHax:
### 🔗 Linked issue

Fixes: #25594

### 📚 Description

- move route selection logic for prerendering to a runtime plugin enabled only during prerendering to allow for files like router.options.ts to override configured routes during prerendering


packages/nuxt/src/pages/runtime/plugins/prerender.ts
```diff
@@ -0,0 +1,40 @@
...
+import _routes from "#build/routes";
+// @ts-expect-error virtual file
+import routerOptions from "#build/router.options";
+import type { RouterConfig } from "@nuxt/schema";
+import { joinURL } from "ufo";
+
+const OPTIONAL_PARAM_RE = /^\/?:.*(\?|\(\.\*\)\*)$/;
+
+export default defineNuxtPlugin(() => {
+  if (!import.meta.server || !import.meta.prerender) {
```
RyanClementsHax, line 13: I need to add guards for the condition below https://github.com/nuxt/nuxt/blob/5d071a70d3b8afdf450e62489c19ba22055480ae/packages/nuxt/src/pages/module.ts#L270

I'm still figuring out how to get all of these options available in a plugin, but wanted to throw up what I had in progress.

danielroe: A plugin is a great approach! Let me push a couple of changes, and I can think we can go ahead and merge 🙏

RyanClementsHax:
@danielroe With the most recent changes, does this only waiting on https://github.com/unjs/nitro/pull/2348 now?

playground/package.json
```diff
@@ -4,8 +4,7 @@
   "scripts": {
     "dev": "nuxi dev",
     "build": "nuxi build",
-    "start": "nuxi preview",
-    "generate": "nuxi generate"
```
RyanClementsHax: @danielroe I'm curious, how come we don't want this in the package.json?

danielroe: Perfectly fine to have it, but I try to restrict changes in PRs to the single thing the PR is doing.

danielroe:
I don't think the nitro discussion needs to delay us here. At a minimum, this isn't a regression, I think. In future we might consider how to support auto-detection in cases where there is no `/` route in a user project.

danielroe (approved):

RyanClementsHax:
> This was nice work - thank you ❤️
> 
> I don't think the nitro discussion needs to delay us here. At a minimum, this isn't a regression, I think. In future we might consider how to support auto-detection in cases where there is no `/` route in a user project.

If you write up an issue, i wouldn't mind taking this on next.

packages/nuxt/src/pages/module.ts
```diff
@@ -266,36 +263,14 @@ export default defineNuxtModule({
...
-          for (const route of nitro.options.prerender.routes) {
-            // Skip default route value as we only generate it if it is already
-            // in the detected routes from `~/pages`.
-            if (route === '/') { continue }
-            prerenderRoutes.add(route)
-          }
-        }
-        nitro.options.prerender.routes = Array.from(prerenderRoutes)
-      })
+    nuxt.hook('app:resolve', (app) => {
```
RyanClementsHax, line 266: Is `nitro.options` only valid after `app:resolve`?

packages/nuxt/src/pages/module.ts
```diff
@@ -266,36 +263,14 @@ export default defineNuxtModule({
       })
     })
 
-    nuxt.hook('nitro:init', (nitro) => {
-      if (nuxt.options.dev || !nitro.options.static || nuxt.options.router.options.hashMode) { return }
```
RyanClementsHax, line 270: Is there anything that preserves the `nuxt.options.dev` or `nitro.options.static` guards?

I'm guessing the `.server.ts` extension serves the purpose of `nuxt.options.dev`.

packages/nuxt/src/pages/module.ts
```diff
@@ -266,36 +263,14 @@ export default defineNuxtModule({
...
-            // in the detected routes from `~/pages`.
-            if (route === '/') { continue }
-            prerenderRoutes.add(route)
-          }
-        }
-        nitro.options.prerender.routes = Array.from(prerenderRoutes)
-      })
+    nuxt.hook('app:resolve', (app) => {
+      const nitro = useNitro()
+      if (nitro.options.prerender.crawlLinks) {
```
RyanClementsHax, line 268: This condition plus nitro not scanning `/` by default means the pages directory routes are no longer prerendered automatically. Was this intentional?

PR #26744, opened 2024-04-11
feat(kit): provide type support for module options in `installModule`

ineshbose:
### 📚 Description

These changes relies and builds on top of #18416 (my first PR on Nuxt 😄); it extracts the options from there after module authors run prepare.

Try this in the repo playground:

```ts
import { installModule } from "@nuxt/kit";

installModule('@nuxt/devtools', { customTabs: [] })
```


test/fixtures/basic-types/nuxt.config.ts
```diff
@@ -41,6 +41,15 @@ export default defineNuxtConfig({
         filename: 'test.d.ts',
         getContents: () => 'declare type Fromage = "cheese"',
       })
+      function _test () {
+        installModule('~/modules/example', {
```
ineshbose, line 45: this doesn't seem to be picking up because `.nuxt/types/schema.d.ts` now seems to be using resolved paths, i.e.

```ts
declare module 'nuxt/schema' {
  interface NuxtConfig {
    modules?: (["/home/nuxt/nuxt/test/fixtures/basic-types/modules/example", Exclude<NuxtConfig["sampleModule"], boolean>] | ["@nuxt/devtools", Exclude<NuxtConfig["devtools"], boolean>])[],
  }
}
```

so where `T = '~/modules/example'`, it is not equal to `/home/nuxt/nuxt/test/fixtures/basic-types/modules/example`; so this will work for package names more than local modules, sure, but I'm not sure how the usage ahead of `['~/modules/example, { typeTest: () {} }]` worked even when the path isn't same 🤔

ineshbose: > but I'm not sure how the usage ahead of `['~/modules/example, { typeTest: () {} }]` worked even when the path isn't same

it's likely that its just the TypeScript is quite relaxed with all the union options.

ineshbose: it should work now; also note that (as I had thought), we're doing a dynamic import of `NuxtConfig` within types

BobbieGoede: Looks like the dynamic import https://github.com/nuxt/nuxt/pull/26744/files#diff-84e752bfeb5965269318fc84ddc24d24fa76cdca52fcc7d4e61f9ea5e7efae21R17 ends up using `NuxtConfig` from `@nuxt/schema` instead of `nuxt/schema` after build, but `nuxt/schema` is augmented by the generated types https://github.com/nuxt/nuxt/pull/26744/files#diff-2176c7ab69dcd5ffc94e79963693b007471788f25591aeb72e7533240ebad2c2R168.

If you use a type import instead of the dynamic import and set `nuxt/schema` as external it seems to work as expected, but I have no idea if this has other unintended results. 

I couldn't open a PR on your branch for some reason so you can check out my changes in this commit instead [BobbieGoede/nuxt@`f1ba5e1` (#5)](https://github.com/BobbieGoede/nuxt/pull/5/commits/f1ba5e18017c1cb60c28487d9fbdfa8fc34dc915).

ineshbose: You're absolutely amazing ❤️

packages/kit/build.config.ts
```diff
@@ -7,6 +7,7 @@ export default defineBuildConfig({
   ],
   externals: [
     '@nuxt/schema',
+    'nuxt/schema',
```
danielroe, line 10: we need to use `@nuxt/schema` within Kit as there is no `nuxt` dependency. The end user is the only one who should use `nuxt/schema` as a type import/augmentation...

danielroe: Yes, I understand. We need to find a different solution than using `nuxt/schema` with `@nuxt/kit`...

danielroe (approved):

danielroe:
Nice work! And also debugging the type issue. This will be a nice DX improvement ❤️

PR #26776, opened 2024-04-14
chore: update `devcontainer.json` based on documentation

iiio2:
In the documentation, `Dockerfile` is given such way. This is given in the same way of documentation. 

https://code.visualstudio.com/docs/devcontainers/create-dev-container#_dockerfile  https://containers.dev/implementors/json_reference/#image-specific


danielroe (approved):
Thank you!

PR #26821, opened 2024-04-17
perf(nuxt): unsubscribe from asyncData watch when component is unmounted

### LINKED ISSUE #26820: useAsyncData is not cleared when using defineNuxtComponent
### Environment

- Operating System: Linux
- Node Version:     v20.11.1
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/daniluk4000/nuxt-refresh-data-duplicate

1. `yarn`
2. `yarn dev`
3. Go to index page
4. Go to second page
5. Go back to index  page
6. Watch second page async data to be repeated
7. If you do that many times, you will find many duplicate repeates when coming back to first page

### Describe the bug

When you use useAsyncData on one page and navigate to another, and then try to use refreshNuxtData, all previous async datas are triggered even when page was already unmounted, which leads to falsey errors and/or duplicate requests.

### Additional context

I was like "HAHA FOUND YA LENNY" when I got the reason for this. We are facing weird issues in our project and I wasn't able to reproduce it on clear project until I finally got the reason for this.

### Logs

_No response_

daniluk4000:
### 🔗 Linked issue

Resolves #26820

### 📚 Description

Fixes performance degradation inproduced in https://github.com/nuxt/nuxt/pull/26554

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -384,6 +384,9 @@ export function useAsyncData<
         await asyncData.refresh()
       }
     })
+    if (instance) {
+      onUnmounted(off)
+    }
     if (hasScope) {
       onScopeDispose(off)
     }
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -375,7 +375,9 @@ export function useAsyncData<
     const hasScope = getCurrentScope()
     if (options.watch) {
       const unsub = watch(options.watch, () => asyncData.refresh())
-      if (hasScope) {
+      if (instance) {
+        onUnmounted(unsub)
+      } else if (hasScope) {
         onScopeDispose(unsub)
       }
     }
@@ -386,8 +388,7 @@ export function useAsyncData<
     })
     if (instance) {
       onUnmounted(off)
-    }
-    if (hasScope) {
+    } else if (hasScope) {
       onScopeDispose(off)
     }
   }


danielroe (approved):

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -384,6 +384,9 @@ export function useAsyncData<
         await asyncData.refresh()
       }
     })
+    if (instance) {
+      onUnmounted(off)
+    }
     if (hasScope) {
```
danielroe, line 390:
```suggestion
    } else if (hasScope) {
```

danielroe: we should also apply the same fix to L378-380

daniluk4000: What if it has both instance and scope?

danielroe: we want to remove when it is unmounted - it will be reinitialised when component is mounted again. the scope is at a higher level (at nuxt app level, or within a pinia store).

daniluk4000: Thanks! Agreed then, will do a bit later

daniluk4000: Done!

danielroe (approved):

daniluk4000:
Can confirm original issue is fixed in latest nightly build

PR #26845, opened 2024-04-18
docs: advise correct extension for vitest config

### LINKED ISSUE #822: This isn't Nuxt specific here, but to import `@nuxt/test-utils` in your vitest config, you either need to have to have `type: "module"` specified in your `package.json` or name it `vitest.config.m{ts,js}`.
This isn't Nuxt specific here, but to import `@nuxt/test-utils` in your vitest config, you either need to have to have `type: "module"` specified in your `package.json` or name it `vitest.config.m{ts,js}`.

We should make this clear in the docs.

_Originally posted by @danielroe in https://github.com/nuxt/test-utils/issues/704#issuecomment-1890993131_

navanjr:
adds `tip`
  add `type: module` to package.json || rename vitest.config filename

Resolves https://github.com/nuxt/test-utils/issues/822

adds a tip section to the setup docs

### DIFF
--- a/docs/1.getting-started/11.testing.md
+++ b/docs/1.getting-started/11.testing.md
@@ -59,6 +59,11 @@ We currently ship an environment for unit testing code that needs a [Nuxt](https
    })
    ```
 
+::tip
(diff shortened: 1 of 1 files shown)


danielroe (approved):
Thank you!

PR #26850, opened 2024-04-19
feat(kit,schema): improve type safety of `defineNuxtModule` options

DamianGlowala:
### 📚 Description

Currently, defining the defaults for options in `defineNuxtModule` does not impact the type of `resolvedOptions` param within the `setup` function. Their types are disconnected, which this PR aims to address.

As of today, TypeScript does not support partial inference of generic type params. Due to this fact, I'd like to suggest adding an extra function call (named `with`), which can handle merging inferred defaults with the module options generic type param and therefore improve both the type safety and DX:
```
type ModuleOptions = {
  option1: string
  option2: number
}

defineNuxtModule<ModuleOptions>().with({
  defaults: {
    option1: 'foo'
  },
  setup: (resolvedOptions) => {}
         // ^? { option1: string, option2?: number | undefined }
})
```

I put effort to add it in a non-breaking way. It would be good to double check nothing changed in the current implementation.


danielroe (approved):
This looks great!

I think we should add some docs to the module author guide. (Can be a subsequent PR.)

PR #26854, opened 2024-04-19
docs: add third parties + scripts documentation

danielroe:
### 📚 Description

🚧 This is a WIP pull request to add documentation for coming nuxt third-parties/scripts implementations.

It's very much subject to change, so don't rely too much on this for now. 😉


harlan-zw:
This needs a bit of refactoring:
- we don't expose `useStyles` anymore
- inline asset strategy is not supported
- the composable names convention is to be prefixed with `useScript` which is already updated upstream

We need quite a bit of documentation and I don't think the Nuxt core repo is the place for it, so I'm making a dedicated docs site. I intend to pick any usable content from this draft into that.

So not sure what you want to do with this PR

danielroe:
For now, I think we need authoritative source of docs in standalone nuxt/scripts docs website. Ultimately we might reconsider merging them back into nuxt.com when API surface is stable and docs changes are few or nuxt/scripts is automatically installed as a dependency of Nuxt.

But until then, we should document key utilities like `useScript` and any other composables that will trigger auto-install of nuxt/scripts with minimal docs pointing to the separate scripts docs website (like we do with nuxt/image).

docs/1.getting-started/5.third-parties.md
```diff
@@ -0,0 +1,153 @@
...
+
+```vue
+<script setup lang="ts">
+useStyles(
+  ['https://www.example.com/styles.css'],
+  { assetStrategy: 'inline' }
+)
+</script>
+```
+
```
huang-julien, line 85:
```suggestion
```

`useStyles` has been removed since the rework of `@nuxt/scripts`

danielroe: Feel free to push directly 🙏❤️

atinux:
What shall we do about this PR?

danielroe:
Thank you for the reminder!

I would love to see these documented on `nuxt.com`, with links to the latest info on https://scripts.nuxt.com, much as we do for `nuxt/image`.

If someone has the time to sync the PR with the latest state on scripts.nuxt.com, that would be perfect 🙏

atinux:
cc @Mini-ghost is this something you feel you can tackle?

Mini-ghost:
It’s my honor! I will do my best to understand the implementation of nuxt scripts and try to complete this part. 💪

Issue #26443, opened 2024-03-22
set `{ deep: false }` by default for data fetching composables


danielroe:

manniL:
@danielroe Would you consider removing the `deep` option all together (or deprecate it at least)?

danielroe:
implemented in https://github.com/nuxt/nuxt/pull/26925 under a flag (to be removed when v4 lands)

Issue #25339, opened 2024-01-19
Emit absolute paths in `builder:watch` hook


manniL:
### Describe the feature

Related PR with discussion: https://github.com/nuxt/nuxt/pull/22333

Nuxt v4 would be the time to update that. Might need discussion before though.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
implemented in https://github.com/nuxt/nuxt/pull/26925 under a flag (to be removed when v4 lands)

Issue #25338, opened 2024-01-19
Remove `templateUtils` 


manniL:
### Describe the feature

`templateUtils` for compiling templates is deprecated and can be removed in v4

https://github.com/nuxt/nuxt/blob/648ef069931336b8ee0c0e48fef3ee700ebc3dbe/packages/nuxt/src/core/app.ts#L37-L38

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
👉 Recommend `knitwork` for manually creating templates.

danielroe:
implemented in https://github.com/nuxt/nuxt/pull/26925 under a flag (to be removed when v4 lands)

Issue #25332, opened 2024-01-19
Remove `compileTemplate` for kit v4


manniL:
### Describe the feature

We can remove `compileTemplate` for Nuxt 4

https://github.com/nuxt/nuxt/blob/648ef069931336b8ee0c0e48fef3ee700ebc3dbe/packages/kit/src/internal/template.ts#L10-L27

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
implemented in https://github.com/nuxt/nuxt/pull/26925 under a flag (to be removed when v4 lands)

PR #26925, opened 2024-04-24
feat(schema): enable opt-in to Nuxt v4 features/flags with `future. compatibilityVersion`

danielroe:
### 📚 Description

This enables users (for testing purposes) to opt-in to the following coming changes:

- [x] https://github.com/nuxt/nuxt/issues/25332
- [x] https://github.com/nuxt/nuxt/issues/26443
- [x] https://github.com/nuxt/nuxt/issues/25338
- [x] https://github.com/nuxt/nuxt/issues/25339

... with potentially more to come in the next weeks.

#### Usage

```ts
  future: {
    compatibilityVersion: 4,
  },
  // Setting the `compatibilityVersion` above flips all the options below to `false`
  // you can granularly re-enable them when testing. Please file issues if so, so that
  // we can address in Nuxt or in the ecosystem.
  // experimental: {
  //   compileTemplate: true,
  //   templateUtils: true,
  //   relativeWatchPaths: true,
  //   defaults: {
  //     useAsyncData: {
  //       deep: true
  //     }
  //   }
  // }
})
```


manniL:
@danielroe could we have a more gradual approach so v4 could be set to `true` or to an object containing the different changes? That might be easier to upgrade + "future-safe" in case we introduce more changes in a patch/minor.

danielroe:
Good idea. Note that we probably do need to have granular control over things that we plan to allow people to configure on v4+, e.g. directory structure. But others will fully go away.

I think I'll move this to `experimental` and make it more granular, for now. But this is likely to be very short lived as an option.

manniL:
> Good idea. Note that we probably do need to have granular control over things that we plan to allow people to configure on v4+, e.g. directory structure. But others will fully go away.
>
> I think I'll move this to `experimental` and make it more granular, for now. But this is likely to be very short lived as an option.

While others might go away, it might help with migration to toggle changes "step by step". Grouping some smaller ones is also fine IMO but that could make migration easier. `experimental` seems like a great place 👍🏻

TheAlexLichter (approved):

manniL:
Great solution to have both, a "progressive" flag enabling more future's by default and the detail-level in case something doesn't work or break 👏

PR #26989, opened 2024-04-29
feat(nuxt): enable `renderSSRHeadOptions` configuration for `unhead`

### LINKED ISSUE #23085: Nuxt doesn't minify html properly (does not remove line-breaks)
### Environment

- Operating System: Darwin
- Node Version:     v16.18.1
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  yarn@3.2.4
- Builder:          vite
- User Config:      app, routeRules, modules, vue, vite, nitro, css, postcss, runtimeConfig
- Runtime Modules:  @pinia/nuxt@0.4.11, @nuxt/devtools@0.8.2, @vueuse/nuxt@10.3.0, nuxt-jsonld@2.0.8
- Build Modules:    -

### Reproduction

`import { fileURLToPath, URL } from 'url';
import svgLoader from 'vite-svg-loader';

  app: {
    baseURL: process.env.BASE_URL || '/',
    head: {
      htmlAttrs: {
        lang: 'de-DE',
      },
      meta: [
        {
          name: 'viewport',
          content: 'width=device-width, initial-scale=1, user-scalable=no',
        },
        {
          name: 'format-detection',
          content: 'telephone=no',
        },
      ],
      link: [
        {
          rel: 'icon',
          type: 'image/x-icon',
          href: '/favicon.ico',
        },
      ],
    },
  },
  routeRules: {
    '/': { prerender: true },
    '/**': { cors: true },
  },
  modules: [
    '@pinia/nuxt',
    '@nuxt/devtools',
    '@vueuse/nuxt',
    [
      'nuxt-jsonld',
      { disableOptionsAPI: true },
    ],
  ],
  vue: {
    compilerOptions: {
      isCustomElement: tag => tag.startsWith('cookie-banner-component'),
    },
  },
  vite: {
    resolve: {
      alias: {
        '~': fileURLToPath(new URL('./', import.meta.url)),
      },
    },
    plugins: [svgLoader()],
    assetsInclude: ['**/*.mdx'],
  },
  nitro: {
    compressPublicAssets: {
      brotli: false,
      gzip: true,
    },
    storage: {
      redis: {
        driver: 'redis',
        port: process.env.REDIS_CONFIG_PORT,
        host: process.env.REDIS_CONFIG_HOST,
        username: '',
        password: process.env.REDIS_CONFIG_PASSWORD,
        db: 0,
        tls: {},
      },
    },
  },
  css: [
    '~/assets/styles/css/tailwind.css',
    '~/assets/styles/scss/main.scss',
  ],
  postcss: {
    plugins: {
      tailwindcss: {},
      autoprefixer: {},
    },
  },
});
`

### Describe the bug

While using `nuxi build`, nuxt minifies the html as follows,

`<!doctype html>
<html data-n-head-ssr lang="en" data-n-head="%7B%22lang%22:%7B%22ssr%22:%22en%22%7D%7D">
  <head>
    <title>Website</title><meta data-n-head="ssr" charset="utf-8"><meta data-n-head="ssr" name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no"><meta data-n-head="ssr" name="description" content="Website"><meta data-n-head="ssr" name="format-detection" content="telephone=no"><link data-n-head="ssr" rel="icon" type="image/x-icon" href="/favicon.ico">`
    
As you can see the line breaks after <!doctype html>, <html>, <head> aren't removed. Why is that? And how do I remove the line breaks? Thanks.

This Problem was mentioned in Nuxt 2 back then as well.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #26983: support configure `omitLineBreaks` of `@unhead`
### Describe the feature

As of `@unhead` v1.8.10, we are able to set `omitLineBreaks` to `true` to eliminate line breaks in the head tags.

Perhaps we could configure this feature via `features.omitLineBreaks` in the Nuxt configuration. The default setting would remain `false` to prevent any issues during page debugging.

This feature is mentioned in the discussion and issue below:
- https://github.com/nuxt/nuxt/discussions/25303
- https://github.com/nuxt/nuxt/issues/23085

### Additional information

### Final checks

Mini-ghost:
### 🔗 Linked issue

- Resolves #25303
- Resolves #23085
- Resolves #26983

### 📚 Description

This PR support configure `renderSSRHeadOptions ` of `unhead` to help us adjust the output of head tags, such as get more compact HTML result.

```ts
  features: {
    omitLineBreaks: true
  }
})
```

In production environment , when `meta.renderSSRHeadOptions.omitLineBreaks` is `false`:

```html
<!DOCTYPE html><html><head><meta charset="utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1">
<link rel="modulepreload" as="script" crossorigin href="/_nuxt/Cc_1Avt4.js">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/CW1WwFCW.js">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/BbxhR_VA.js">
<link rel="prefetch" as="script" crossorigin href="/_nuxt/Cz_w0jng.js">
<script type="module" src="/_nuxt/Cc_1Avt4.js" crossorigin></script></head>
```

In production environment , when `meta.renderSSRHeadOptions.omitLineBreaks` is `undefined` or `true`:

```html
<!DOCTYPE html><html><head><meta charset="utf-8"><meta name="viewport" content="width=device-width, initial-scale=1"><link rel="modulepreload" as="script" crossorigin href="/_nuxt/Bzb4FGiR.js"><link rel="prefetch" as="script" crossorigin href="/_nuxt/CX8zz9fU.js"><link rel="prefetch" as="script" crossorigin href="/_nuxt/bk47NN6-.js"><link rel="prefetch" as="script" crossorigin href="/_nuxt/ImDE-rvC.js"><script type="module" src="/_nuxt/Bzb4FGiR.js" crossorigin></script></head>

```


docs/2.guide/3.going-further/1.features.md
```diff
@@ -33,6 +33,18 @@ export default defineNuxtConfig({
 })
 ```
 
+### omitLineBreaks
+
+Omits line breaks in `<head>` when rendering HTML.
```
xjccc, line 38: Set `omitLineBreaks` to `true` if you prefer to render the head tags without line breaks.

Same with useHead Docs. Is it more appropriate?

danielroe:
I'm wondering about the best location for this (and possibly other options) for the head - maybe instead in `head` or in `app.head`. I'm worried this will clutter `features`. What do you think @harlan-zw?

harlan-zw:
If a `features` Nuxt line break config was going to be added it should handle the line breaks for the entire document, not just the head so I'd say this isn't the right place for it.

Having a way to pass options to the Unhead functions would be the most future proof and it would make sense for that config to exist within the head modules config that is under the `meta` key.  (although we should probably rename this to `head` or `unhead` for v4)

```ts
  meta: {
    renderSSRHead: { // typeof RenderSSRHeadOptions @unhead/schema
      ssrOmitLineBreaks: true,
    }
  }
})
```

I don't really see a reason why this can't be enabled by default in a production environment for v4 either :thinking:

Mini-ghost:
I am currently following the conclusions of https://github.com/unjs/unhead/pull/297 and have set `omitLineBreaks` to be disabled by default. However, I think it would be great if `omitLineBreaks` could be enabled by default in the production environment!

Also, concerning the type of `meta.renderSSRHead`, if it is expected to align with the `RenderSSRHeadOptions` from `@unhead/schema`, perhaps we could structure the configuration as follows to ensure consistency. Additionally, would it be more appropriate to use `renderSSRHeadOptions` instead of `renderSSRHead`?

```ts
  meta: {
    renderSSRHeadOptions: { // typeof RenderSSRHeadOptions @unhead/schema
      omitLineBreaks: true,
    }
  }
})
```

Moreover, as we anticipate the name will change to `head` or `unhead` in v4, could we consider adopting this naming convention in v3 to maintain continuity?

danielroe:
I don't think either `meta` or `head` currently exists in the Nuxt config options, so I think we're good to add a new key without any issues (though `head` would conflict with _Nuxt 2_ types - so it might create issues for module that support both Nuxt 2/Bridge and Nuxt 3+ 🤔)...

Mini-ghost:
Oops! I didn't notice that this would conflict with `@nuxt/bridge`. It seems that in v3, we can only use `meta` as the key!

Mini-ghost:
Wouldn't it be better for us to use `meta` as the key in v3? Additionally, should we set the schema part in [`schema/src/config/app.ts`](https://github.com/nuxt/nuxt/blob/main/packages/schema/src/config/app.ts)?

Mini-ghost:
It appears that our decision to use `meta` as the key in v3 might be a more suitable choice.

Additionally, should we set the schema part in [`schema/src/config/app.ts`](https://github.com/nuxt/nuxt/blob/main/packages/schema/src/config/app.ts)?

docs/2.guide/3.going-further/1.features.md
Mini-ghost: This paragraph has been removed due to settings tweaks, so I'll mark it as resolved.

packages/schema/src/config/app.ts
```diff
@@ -318,4 +318,14 @@ export default defineUntypedSchema({
...
+
+  meta: {
+    /**
+     * An object that will be passed to `renderSSRHead` to customize the output.
+     *
+     * @see https://unhead.unjs.io/setup/ssr/installation#options
+     * @type {typeof import('@unhead/schema').RenderSSRHeadOptions}
+     */
+    renderSSRHeadOptions: undefined,
+  },
```
Mini-ghost, line 330: Comments here will automatically generate documentation. I want to know how to write better?

Mini-ghost:
I didn’t put much thought into the name initially, but now that I’ve thought it over, `unhead` actually sounds like a better choice. It clearly points out where our dependencies come from and helps us avoid any mix-ups.

https://github.com/nuxt/nuxt/blob/572c36745597ee45e9b57f8f32a24dd99a5b9d42/packages/schema/src/config/app.ts#L108-L110

Should we change it to `unhead` then?

danielroe:
The bit in schema is an (untyped) backwards-compatible bit. We'd likely need to keep the backwards-compatibility, but people really shouldn't be using it at this point.

danielroe:
We've decided to add `unhead` key to schema that allows configuring the unhead nuxt module directly, which will unblock configuration of advanced use cases (linking https://github.com/nuxt/nuxt/issues/27031 for `headNext` support). Would you be happy to update this PR? 🙏

packages/nuxt/src/head/module.ts
```diff
@@ -68,9 +75,23 @@ export default import.meta.server ? [CapoPlugin({ track: true })] : [];`
       },
     })
 
+    addTemplate({
+      filename: 'unhead.config.mjs',
+      getContents (ctx) {
+        const renderSSRHeadOptions = defu(options.renderSSRHeadOptions, {
+          omitLineBreaks: !(ctx.nuxt.options.dev || ctx.nuxt.options.test),
+        })
```
danielroe, line 83: I think this can be moved up into `defaults` function above `setup` and it will be pre-normalised for us...

Mini-ghost: OK! I've updated it 😁

danielroe (approved):
Perfect! Though will wait for @harlan-zw's thoughts on API. 🙏

harlan-zw:
Looks good minus the comment. I'm guessing this will land in v3.12 so probably we want to minimise any new behaviour until v4.

packages/nuxt/src/head/module.ts
```diff
@@ -1,12 +1,25 @@
...
-export default defineNuxtModule({
+export default defineNuxtModule<HeadModuleOptions>({
   meta: {
     name: 'meta',
+    configKey: 'unhead',
+  },
+  defaults (nuxt) {
+    return {
+      renderSSRHeadOptions: {
+        omitLineBreaks: !(nuxt.options.dev || nuxt.options.test),
```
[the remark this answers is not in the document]
Mini-ghost: You are right, I have set the default value to `false` and added a `TODO` comment. Thank you for your reminder.

PR #27005, opened 2024-04-30
fix: add "-D" for command, Update 4.styling.md

zer0fire:
🔗 Linked issue
No issue, this is just a misunderstanding

📚 Description
image
When I used the CLI tool, it indicated that I should install development dependencies; however, the documentation did not mention this, which could lead to some misunderstanding.


danielroe:
It doesn't matter whether it's a dev dependency or not, as your `package.json` only contains dependencies used in development and when building. (Nuxt itself is installed as a normal dependency by default in newer versions.)

However, there is probably merit in aligning the docs with the Vue plugin's prompt.

PR #27010, opened 2024-04-30
feat(nuxt): prompt to install `nuxt/scripts` on usage

danielroe:
### 📚 Description

In preparation for `nuxt/scripts` initial release, this adds a prompt when adding a `scripts: {}` option to Nuxt config or using one of:

- `useConsentScriptTrigger`
- `useAnalyticsPageEvent`
- `useElementScriptTrigger`
- `useScript`
- `useScriptGoogleAnalytics`
- `useScriptPlausibleAnalytics`
- `useScriptCloudflareWebAnalytics`
- `useScriptFathomAnalytics`
- `useScriptMatomoAnalytics`
- `useScriptGoogleTagManager`
- `useScriptSegment`
- `useScriptFacebookPixel`
- `useScriptXPixel`
- `useScriptIntercom`
- `useScriptHotjar`
- `useScriptStripe`
- `useScriptLemonSqueezy`
- `useScriptVimeoPlayer`
- `useScriptYouTubeIframe`
- `useScriptGoogleMaps`
- `useScriptNpm`

(We also need to add stub docs in https://github.com/nuxt/nuxt/pull/26854 for these, pointing to the source.)

cc: @harlan-zw - do these utilities look like the ones we should stub, or do you think there are any others to include here?


harlan-zw:
Nice, this looks really good.

The only thing we may consider is stubbing registry scripts. Adding all of them is too much maintenance effort but maybe we can support the TPC scripts `useScriptGoogleAnalytics` and `useScriptGoogleTagManager`.

Components could also be added but starting to get a bit bloated so not sure, it would be for YoutubeEmbed and GoogleMaps.

danielroe:
One thought: what do you think about `@nuxt/scripts` exporting the names of any key composables? We could inline the names, at least, directly from `@nuxt/scripts` in nuxt itself, which might avoid the need to update manually.

harlan-zw:
v0.3.0 exports [/registry](https://unpkg.com/browse/@nuxt/scripts@0.3.0/dist/registry.mjs) which includes all of the Registry Scripts. It's a little heavy as it's shipping the icons used for the DevTools and the docs, will trim it down if you're going to use it.

Unrelated to that, as part of this PR I wonder if you'd consider adding support for "stubbing" via Nuxt Config. I imagine a common way for people to use Nuxt Scripts between sites will just be to copy + paste config over. 

For example my personal sites all use Fathom, I don't use any of the functions provided by Fathom itself so registering the script via nuxt config makes the most sense, I just need to provide the site config. This is similar to Nuxt sites using Plausible.

The config would look like 

```ts
  scripts: {
    registry: {
      fathomAnalytics: {
        site: 'YOUR_TOKEN_ID'
      }
    }
  }
})
```

harlan-zw:
I reckon if we have the nuxt config stub we can skip the registry script composables.

shinGangan:
Hi @danielroe , I tried to use the `ScriptYouTubePlayer` component in Nuxt v3.12.4 but Auto-imports was not possible.

Looking at the implementation of `.nuxt/components.d.ts` and `.nuxt/imports.d.ts`, it seems that Nuxt Scripts components are not included as targets for Auto-imports.

Will this be added in the future? Or is the plan to only add composables?
If it's not supported yet, I'll create a new PR.

Issue #26995, opened 2024-04-30
Further Customization Options on Teleport and Root Tags


woodbrettm:
### Describe the feature

Hi there, I know the `nuxt.config.ts` file includes the option to customize `app.rootTag` and `app.rootId`. It'd be great to have the option to provide custom attributes to the rootTag. Same goes for the `teleportId`, and `teleportTag` options.

Perhaps instead of `rootId`, and `teleportId`, we could be allowed to provide a function which returns an object containing keyvalue pairs of attributes the element would contain?

Looks like I should be able to do this with a module if needed, but posting the request as something like `app.rootAttrs` and `app.teleportAttrs` keys would be useful directly from the config.

Similar discussion was in #26609

Thanks!,
Brett

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

woodbrettm:
Sure a tuple or object would work great too thanks :)

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/27014

PR #27014, opened 2024-05-01
feat(nuxt): support configurable root/teleport element attributes

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/26995

### 📚 Description

This PR allows users to add any arbitrary attribute to the root element (`<div id="__nuxt">`) and the teleport element (`<div id="teleports">`.

```ts
  app: {
    rootAttrs: {
       class: 'w-full min-h-screen'
    }
  }
})
```

- Consider deprecating `teleportId` and `rootId` with these changes for a smaller config surface.
- We prefer `rootAttrs`, `teleportAttrs` for parity with Unhead `htmlAttrs`, `bodyAttrs`
- Question: Should & will root element always have a unique id?


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -232,15 +258,8 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
...
-const HAS_APP_TELEPORTS = !!(appTeleportTag && appTeleportId)
-const APP_TELEPORT_OPEN_TAG = HAS_APP_TELEPORTS ? `<${appTeleportTag} id="${appTeleportId}">` : ''
-const APP_TELEPORT_CLOSE_TAG = HAS_APP_TELEPORTS ? `</${appTeleportTag}>` : ''
-
-const APP_ROOT_OPEN_TAG = `<${appRootTag}${appRootId ? ` id="${appRootId}"` : ''}>`
-const APP_ROOT_CLOSE_TAG = `</${appRootTag}>`
-
 const PAYLOAD_URL_RE = process.env.NUXT_JSON_PAYLOADS ? /\/_payload.json(\?.*)?$/ : /\/_payload.js(\?.*)?$/
-const ROOT_NODE_REGEX = new RegExp(`^${APP_ROOT_OPEN_TAG}([\\s\\S]*)${APP_ROOT_CLOSE_TAG}$`)
+const ROOT_NODE_REGEX = new RegExp(`^<body[^>]*>\\s*<${appRootTag}[^>]*>([\\s\\S]*?)<\\/${appRootTag}>\\s*<\\/body>$`)
```
harlan-zw, line 262: Does this seem safe @huang-julien :thinking:

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -232,15 +241,8 @@ async function getIslandContext (event: H3Event): Promise<NuxtIslandContext> {
...
-const HAS_APP_TELEPORTS = !!(appTeleportTag && appTeleportId)
-const APP_TELEPORT_OPEN_TAG = HAS_APP_TELEPORTS ? `<${appTeleportTag} id="${appTeleportId}">` : ''
-const APP_TELEPORT_CLOSE_TAG = HAS_APP_TELEPORTS ? `</${appTeleportTag}>` : ''
-
-const APP_ROOT_OPEN_TAG = `<${appRootTag}${appRootId ? ` id="${appRootId}"` : ''}>`
-const APP_ROOT_CLOSE_TAG = `</${appRootTag}>`
-
 const PAYLOAD_URL_RE = process.env.NUXT_JSON_PAYLOADS ? /\/_payload.json(\?.*)?$/ : /\/_payload.js(\?.*)?$/
-const ROOT_NODE_REGEX = new RegExp(`^${APP_ROOT_OPEN_TAG}([\\s\\S]*)${APP_ROOT_CLOSE_TAG}$`)
+const ROOT_NODE_REGEX = new RegExp(`^<${appRootTag}[^>]*id\\s*=\\s*["']${appRootId}["'][^>]*>([\\s\\S]*)<\\/${appRootTag}>$`)
```
harlan-zw, line 245: This regex needs a look, it requires an id to be set for the root tag.

danielroe: I think we don't really need the ID - we're just stripping the outer tag from the HTML within this line:

https://github.com/nuxt/nuxt/blob/e9181116a3e25f5507ba1df11ef5515ed05cf297/packages/nuxt/src/core/runtime/nitro/renderer.ts#L637-L643

Which means, I think, that we can be pretty imprecise on the RE.

harlan-zw: There can be multiple multiple root (body > div) tags both prepended and appended from the "root" node? This logic wants only the "root", not the teleports right or the other appended / prepended Unhead templates :thinking: 

https://github.com/nuxt/nuxt/blob/e9181116a3e25f5507ba1df11ef5515ed05cf297/packages/nuxt/src/core/runtime/nitro/renderer.ts#L561

https://github.com/nuxt/nuxt/blob/e9181116a3e25f5507ba1df11ef5515ed05cf297/packages/nuxt/src/core/runtime/nitro/renderer.ts#L471-L473

danielroe: The way the regexp is written, it only looks at the first item in `html.body` (not html.bodyPrepend, or the other bits), and it intends to match the entire string (hence the `^` and `$`).

I don't think it would work (as written currently) if the user prepended an item in `body` within the `render:html` hook. (But if the user doesn't do that, we know for sure it will be just the result from `renderToString().html`.)

harlan-zw:
Thanks for the help, updates look good :100:

danielroe (approved):

PR #27022, opened 2024-05-01
feat(nuxt): check for compatible builder when installing modules

HigherOrderLogic:
### 🔗 Linked issue

N/A

### 📚 Description

Some modules are only compatible with specific builder (eg: plugins available only for Vite) so it may not work when using other builders.


docs/3.api/5.kit/3.compatibility.md
```diff
@@ -25,6 +25,7 @@ async function checkNuxtCompatibility(
 interface NuxtCompatibility {
   nuxt?: string;
   bridge?: boolean;
+  builder: 'vite' | 'webpack' | 'all';
```
danielroe, line 28:
```suggestion
  builder?: 'vite' | 'webpack' | 'all';
```

docs/3.api/5.kit/3.compatibility.md
```diff
@@ -25,6 +25,7 @@ async function checkNuxtCompatibility(
 interface NuxtCompatibility {
   nuxt?: string;
   bridge?: boolean;
+  builder?: 'vite' | 'webpack' | 'all';
```
danielroe, line 28: Setting a single valid builder option makes sense now with only vite + webpack as options, but in future if we have rspack as well, or some other builder, it will be difficult to express the constraint that way.

Maybe instead we could have `vite` or `webpack` as top-level items allowing modules to express not just that they do work (or not!) with them, but additionally which version they require?

What do you think @pi0?

danielroe (approved):

docs/3.api/5.kit/3.compatibility.md
```diff
@@ -25,6 +25,12 @@ async function checkNuxtCompatibility(
 interface NuxtCompatibility {
   nuxt?: string;
   bridge?: boolean;
+  builder?: {
+    // Set `false` if your module is not compatible with a builder
+    // or a semver-compatible string version constraint
+    vite?: false | string;
+    webpack?: false | string;
+  };
```
HigherOrderLogic, line 33: https://github.com/nuxt/nuxt/blob/9b1bfe69ed4a146e2a12131f53019618fa115085/packages/schema/src/types/config.ts#L80

Shouldnt we add one more field, maybe `custom`, to support user-provided builder in the form of `{ bundle: (nuxt) => {} }`?

danielroe: I think user provided builder (in form of string) would work automatically:

```ts
builder: {
  'my-custom-builder': false
}
```

But I'm not sure how we could apply a constraint to a custom `bundle` function.

danielroe: Or did you mean just fully disabling with `custom: false`?

HigherOrderLogic: Yeah, my original idea when making this PR is to totally deny or accept custom builder, since they are user-provided and we cannot control or know what features they have.

danielroe: Is there an example you've come across? I think if a user is supplying a custom builder then they are probably aware that not everything might work as expected.

Perhaps we could merge as-is and iterate?

HigherOrderLogic: I havent seen any real life example of custom builder but I added it just in case.

If you're okay with the current PR, you can just merge it as is.

danielroe: Great! Let's review if/when it becomes an issue.

Thank you so much for this initiative ❤️

tobiasdiez:
Thanks @HigherOrderLogic for this PR. This is a nice addition 🚀

PR #27037, opened 2024-05-02
feat(nuxt): add `onPrehydrate` lifecycle hook

danielroe:
### 📚 Description

This adds a new composable lifecycle hook:`onPrehydrate`.

It allows running a callback on the client immediately before Nuxt hydrates the page.

The callback will be stringified and inlined in the HTML so it should not have any external
dependencies (such as auto-imports) or refer to variables defined outside the callback.

The callback will run before Nuxt runtime initializes so it should not rely on the Nuxt or Vue context.

#### Example

```vue
<script setup lang="ts">
// onPrehydrate is guaranteed to run before Nuxt hydrates
onPrehydrate(() => {
  console.log(window)
})

// As long as it only has one root node, you can access the element
onPrehydrate((el) => {
  console.log(el.outerHTML)
  // <div data-v-inspector="app.vue:15:3" data-prehydrate-id="b3qlvSiBeH"> Hi there </div>
})

// For _very_ advanced use cases (such as not having a single root node) you
// can access/set `data-prehydrate-id` yourself
const prehydrateId = onPrehydrate((el) => {})
</script>
```


Mini-ghost:
Just a thought. Have you thought about possibly naming it `definePreHydrate`?

For names like `definePageMeta`, `defineProps`, `defineEmits`, and `defineOption`, these get special treatment by the compiler. 

This is how I categorize them. Some names that begin with `define` help users specify types, such as `defineNuxtComponent` and `defineComponent`. Others relate to compiler macros as previously mentioned.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -65,3 +68,47 @@ export function prerenderRoutes (path: string | string[]) {
...
+ *
+ * The callback will be stringified and inlined in the HTML so it should not have any external
+ * dependencies (such as auto-imports) or refer to variables defined outside the callback.
+ *
+ * The callback will run before Nuxt runtime initializes so it should not rely on the Nuxt or Vue context.
+ * @since 3.12.0
+ */
+export function onPrehydrate (callback: (el: HTMLElement) => void): void
+export function onPrehydrate (callback: string | ((el: HTMLElement) => void), key?: string): undefined | string {
+  if (import.meta.client) { return }
```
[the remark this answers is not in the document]
huang-julien: + a warning like definePageMeta ?

harlan-zw:
This is very cool and the code looks good.

Naming I can go either way.

docs/3.api/2.composables/on-prehydrate.md
```diff
@@ -0,0 +1,61 @@
...
+Nuxt hydrates the page."
+links:
+  - label: Source
+    icon: i-simple-icons-github
+    to: https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/ssr.ts
+    size: xs
+---
+
+::important
+This composable will be available in Nuxt v3.12+.
```
manniL, line 13: Might be worth mentioning the nightly channel here?

docs/3.api/2.composables/on-prehydrate.md
```diff
@@ -0,0 +1,61 @@
...
+`onPrehydrate` is a composable lifecycle hook that allows you to run a callback on the client immediately before
+Nuxt hydrates the page.
+
+::note
+This is an advanced utility and should be used with care.
+::
+
+## Usage
+
+`onPrehydrate` can be called directly in the setup function of a Vue component, or in a plugin. It will only have an
```
manniL, line 25: "setup function"  also refers to "inside `<script setup>`" but I see the risk that this might not be clear for people not aware of `setup()` without the syntactic sugar

Issue #27064, opened 2024-05-05
Improve ignorePrefix JSDoc


volk234:
### Environment

https://stackblitz.com/edit/github-4pefag?file=package.json&view=editor

### Reproduction

https://stackblitz.com/edit/github-4pefag?file=package.json&view=editor

### Describe the bug

1. Create file "-test.txt" in public directory
2. Build project
3. The public directory does not contain a file "-test.txt"

manniL:
I'd say this is intended because `-` is the `ignorePrefix`. The [JSDOC for it](https://github.com/nuxt/nuxt/blob/b330a296b8c91d1dd6dbc3dc74a84819120830a8/packages/schema/src/config/common.ts#L399-L405) should be updated though.

PR #27065, opened 2024-05-05
docs: update jsdoc for ignore-prefix to clarify ignored files

sossost:
### 🔗 Linked issue

#27064 

### 📚 Description
This commit updates the JSDoc comments to clarify the behavior of the `ignorePrefix` setting in the configuration. The updated documentation now explicitly states that any file in `pages/`, `layouts/`, `middleware/`, `store/`, and `public/` directories will be ignored during the build process if its filename starts with the prefix specified by `ignorePrefix`.

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -39,7 +39,7 @@ export default defineUntypedSchema({
    * It is normally not needed to configure this option.
    */
   rootDir: {
-    $resolve: val => typeof val === 'string' ? resolve(val) : process.cwd(),
+    $resolve: val => (typeof val === 'string' ? resolve(val) : process.cwd()),
   },
 
   /**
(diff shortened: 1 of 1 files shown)


TheAlexLichter (changes requested):
please remove stylistic changes from the PR ☺️

packages/schema/src/config/common.ts
```diff
@@ -397,8 +442,10 @@ export default defineUntypedSchema({
   ignoreOptions: undefined,
 
   /**
-   * Any file in `pages/`, `layouts/`, `middleware/` or `store/` will be ignored during
-   * building if its filename starts with the prefix specified by `ignorePrefix`.
+   * Any file in `pages/`, `layouts/`, `middleware/`, `store/`, and `public/` directories will be ignored during
```
manniL, line 445: I think `store` can be removed here (was Nuxt 2 specific). 
Besides that, it'd be good to check in the code which other parts are affected

sossost: @manniL oh. I'm sorry. I ran lint:fix to see if stylistic changes are allowed.

manniL: Oh, if they are from lint:fix that that's fine 🙈

danielroe (approved):

Issue #21635, opened 2023-06-19
Multi-app support (TBA)


danielroe:

jeantil:
As far as I understand this feature is the only hope to have full microfrontend support.
- Layers imply build time dependencies
- vite-federation-plugin can import bits of a nuxt app but the [scoped css support is broken](https://github.com/originjs/vite-plugin-federation/issues/361) making it difficult/impossible to use for larger projects wich are the ones that benefit most of micro-frontend.

The best we could come up with for our migration from pure vue 3 apps to nuxt is to use iframes which has practical limits.

localusercamp:
I hope this will be a great opportunity to easily scaffold a DDD application.

OlaAlsaker:
> Is this app as in the concept of App in Vue, or apps like plugin apps on a CMS?

As I understand it, the multi-app feature is designed for combining multiple Nuxt apps on a single page. In large applications, different teams often work on different parts of the same app. Instead of each team working on the same Nuxt app, they can work on their own app, and Nuxt can automatically merge these apps together.

jeantil:
The blog post for [3.11](https://nuxt.com/blog/v3-11) says 3.11 is the last before nuxt 4. for which a first release is expected 2024-Q1 according to the [roadmap](https://nuxt.com/docs/community/roadmap). 

Does it mean Nuxt 3.x will never have multi application support ? 
if 3.x never gets it, is multi-app support something of a goal for nuxt 4 ?

danielroe:
Features aim to land in minor releases. Breaking changes in major ones. Our feature roadmap + plans isn't really affected by our v4 plans; v4 is just an opportunity to make strategic (and gentle) breaking changes.

nicolaspayot:
@danielroe I'd like to give a try on the first task: namespace access to `useNuxtApp('app-name')`

danielroe:
@nicolaspayot Great! 🎉

aekovalev:
@danielroe Daniel, hello! Is there any information about the appearance of support for microfrontends in nuxt out of the box? when will this happen? after the release of nuxt v4?

danielroe:
It's in active development. 😊

odranoelBR:
Best nuxt feature =).

cheindl:
That ist really great news to hear! But i wonder which of the following scenarios will multi-app nuxt support?

1) Multiple instances of different nuxt apps included within one page
[...]

PR #27068, opened 2024-05-06
feat(nuxt): allows the support of multiple shared runtime context

nicolaspayot:
### 🔗 Linked issue

First task of #21635 

### 📚 Description

These changes allows the support of multiple shared runtime contexts by namespacing them under a new `buildId` option, instead of `nuxt-app` as default. It's one of the first task for multi-app support.


packages/nuxt/src/app/nuxt.ts
```diff
@@ -20,11 +20,16 @@ import type { LoadingIndicator } from '../app/composables/loading-indicator'
...
 
-import type { NuxtAppLiterals } from '#app'
+import { type NuxtAppLiterals } from '#app'
+// @ts-expect-error virtual import
+import { buildId } from '#build/nuxt.config.mjs'
 
-const nuxtAppCtx = /* @__PURE__ */ getContext<NuxtApp>('nuxt-app', {
-  asyncContext: !!__NUXT_ASYNC_CONTEXT__ && import.meta.server,
-})
+/* @__NO_SIDE_EFFECTS__ */
```
nicolaspayot, line 27: Here I'm not so sure about this, should I have kept `/* @__PURE__ */` or is `/* @__NO_SIDE_EFFECTS__ */` fine? I see it's rather used for functions.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -458,31 +464,31 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp |
...
 /**
- * Returns the current Nuxt instance.
+ * Returns the Nuxt instance from the given app name if provided.
  *
  * Throws an error if Nuxt instance is unavailable.
  * @since 3.0.0
  */
-export function useNuxtApp (): NuxtApp {
-  const nuxtAppInstance = tryUseNuxtApp()
+export function useNuxtApp (appName?: string): NuxtApp {
```
nicolaspayot, line 490: From the outside, `appName` might make more sense than `buildId` but as it's defaulting to `buildId` for now, should we actually use `buildId`?

danielroe: I think `appName` makes sense here 👍

packages/nuxt/src/app/nuxt.ts
danielroe: I think we can remove it now.

```suggestion
```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -20,11 +20,15 @@ import type { LoadingIndicator } from '../app/composables/loading-indicator'
 import type { RouteAnnouncer } from '../app/composables/route-announcer'
 import type { ViewTransition } from './plugins/view-transitions.client'
 
-import type { NuxtAppLiterals } from '#app'
-
-const nuxtAppCtx = /* @__PURE__ */ getContext<NuxtApp>('nuxt-app', {
-  asyncContext: !!__NUXT_ASYNC_CONTEXT__ && import.meta.server,
-})
+import { type NuxtAppLiterals } from '#app'
```
danielroe, line 23:
```suggestion
import type { NuxtAppLiterals } from '#app'

```

packages/nuxt/src/app/nuxt.ts
```diff
@@ -242,6 +243,7 @@ export interface CreateOptions {
 export function createNuxtApp (options: CreateOptions) {
   let hydratingCount = 0
   const nuxtApp: NuxtApp = {
+    name: buildId,
```
nicolaspayot, line 246: Ah, that's a good addition 👍

vitest.nuxt.config.ts
```diff
@@ -13,6 +13,7 @@ export default defineVitestConfig({
     environmentOptions: {
       nuxt: {
         overrides: {
+          buildId: 'nuxt-app',
```
danielroe, line 16: **Note to self**: before release we need to update `@nuxt/test-utils` to set `buildId` to `nuxt-app` in the vitest environment...

docs/3.api/2.composables/use-nuxt-app.md
```diff
@@ -18,6 +18,14 @@ const nuxtApp = useNuxtApp()
 
 If runtime context is unavailable in your scope, `useNuxtApp` will throw an exception when called. You can use [`tryUseNuxtApp`](#tryusenuxtapp) instead for composables that do not require `nuxtApp`, or to simply check if context is available or not without an exception.
 
+::note
+By default, the shared runtime context of Nuxt is namespaced under the [`buildId`](/docs/api/nuxt-config#buildid) option. It allows the support of multiple runtime contexts.
+::
+
+## Params
+
+- `appName`: an optional application name. If you do not provide it, the Nuxt `buildId` option is used. Otherwise, it must match with an existing `buildId`.
```
danielroe, line 27: I think for now let's not document this. It won't fully work yet and it will confuse people as a result. I will comment it out so we can add back in a future PR.

danielroe (approved):

PR #27029, opened 2024-05-01
feat(nuxt,schema): support v4 folder structure

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26444

### 📚 Description

This implements the new folder structure, if `future.compatibilityVersion` is set to `4` (which will be enabled by default when Nuxt v4 is released) - see https://github.com/nuxt/nuxt/pull/26925.

The main points (all only relevant when v4 compatibility is enabled):

* the new Nuxt default `srcDir` is `app/` by default, and most things are resolved from there.
* `serverDir` now defaults to ` <rootDir>/server` rather than `<srcDir>/server`
* `modules` and `public` are resolved relative to `<rootDir>` by default. If you want to revert this, set `dir.modules` to `'src/modules'` and `dir.public` to `'src/public'`.
* a new `dir.app` is added, which is the directory we look for `router.options.ts` and `spa-loading-template.html` - this defaults to `<srcDir>/` (under v4) and `<srcDir>/app` in v3 mode
* I have not changed `app.vue` to `main.vue` (as in the RFC) to avoid additional breaking changes
* in v4 mode, if there is no `app/` directory, we default to previous behaviour (and restart if the `app/` directory is added)

🚨 **Implementation details might change before the release of v4.** 🚨

<details>

<summary>The v4 folder structure.</summary>

```
.output/
.nuxt/
app/
  assets/
  components/
  composables/
  layouts/
  middleware/
  pages/
  plugins/
  utils/
  app.vue
  router.options.ts
modules/
node_modules/
public/
server/
  api/
  middleware/
  plugins/
  routes/
  utils/
nuxt.config.ts
```

</details>

PR #27077, opened 2024-05-06
docs: put comment on right line

edimitchel:
### 🔗 Linked issue

Just a docs issue

### 📚 Description

Comment was on the wrong place since following change was added: https://github.com/nuxt/nuxt/pull/27029/files#diff-5e083470b457ea806258cab809ae6fe30eae38cdd7a1e8ba98e86969e5476226

### DIFF
--- a/docs/2.guide/3.going-further/1.features.md
+++ b/docs/2.guide/3.going-further/1.features.md
@@ -55,11 +55,11 @@ export default defineNuxtConfig({
   future: {
     compatibilityVersion: 4,
   },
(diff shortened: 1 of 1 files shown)


danielroe (commented):
I believe the comment is on the correct line. Have you tried it?

manniL:
I also think the comment is on the right line, (given the new default directory structure) ☺️

edimitchel:
My bad I thought the comment was for default values!

PR #27117, opened 2024-05-08
feat(kit): add `useRuntimeConfig` and `updateRuntimeConfig` utils

### LINKED ISSUE #24224: Runtime config is not replaced by env variables in the module
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.1
- CLI Version:      3.9.1
- Nitro Version:    2.7.2
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools, runtimeConfig
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-anbt7g?file=modules%2Ftest.ts

### Describe the bug

Expected: `feature` should be equal to `bar`
Actual: `feature` is equal to `foo`

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #26960: Resolve `runtimeConfig` within nuxt options
At the moment runtimeConfig is resolved by Nitro which means that Nuxt modules can't access the resolved config from environment.

We can resolve this in Nuxt directly to improve module author experience and promote consistency, without changing build/runtime behaviour.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26960
resolves https://github.com/nuxt/nuxt/issues/24224

### 📚 Description

In line with [this suggestion](https://github.com/unjs/nitro/pull/2404#issuecomment-2086887658), this PR adds two new Kit utilities:

* `useRuntimeConfig` - this allows accessing `nuxt.options.nitro.runtimeConfig`, with the configuration resolved/expanded with environment variables
* `updateRuntimeConfig` - this allows deep-updating `nuxt.options.nitro.runtimeConfig` (and also triggering a Nitro HMR if it has already initialised)

In addition, `nuxt.options.runtimeConfig` is synced with `nuxt.options.nitro.runtimeConfig` and Nitro runtimeConfig resolution is moved into `@nuxt/schema` rather than kept directly within Nuxt.


packages/kit/src/runtime-config.ts
pi0, line 1: Usually i would add a comment to the source and commit of the forked code for reference and let's hope we can eventually soon use same utils.

packages/kit/src/runtime-config.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type EnvOptions = {
+  env?: Record<string, unknown>
+  prefix?: string
+  altPrefix?: string
+  envExpansion?: boolean
+}
+
+function getEnv (key: string, opts: EnvOptions) {
+  const env = opts.env ?? process.env
```
pi0, line 48: discussion as https://github.com/unjs/nitro/pull/2404: it is not really reasonable to me that we are accepting an env object that is not system environment variables yet call it env. it really should be called something else or directly augment `runtimeConfig` object with any source that is not system environment variables for purpose of updating it like for testing env.

PS: It seems both getEnv/applyEnv internals are currently not exposed so I appreciate keeping consistency with Nitro and discuss on it later if you believe differently.

danielroe: Yes, this isn't exposed and doesn't allow using anything other than `process.env` in this PR.

pi0: if the final goal is to reuse it into kit, it won't allow this. the purpose of this util is get a system/runtime environment variable by name.

danielroe: This configuration option isn't exposed in this PR. But as I commented above, it is very useful in a context like `@nuxt/test-utils` to be able to get overrides from a source other than `process`.

pi0: > This configuration option isn't exposed in this PR

So I guess no harm in not introducing it until waiting for the upstream discussion. 

---

I'm afraid by introducing this change, we **WILL** be diverged from nitro. I also understand your use case and proposed a better alternative (`updateRuntimeConfig()`) to update the runtime config from different sources. changing the definition of environment variables is something I never would opt nitro for (nor wish nuxt does..)

danielroe: Oh, I misunderstood your earlier comment:

> It seems both getEnv/applyEnv internals are currently not exposed so I appreciate keeping consistency with Nitro and discuss on it later if you believe differently.

I read that as you appreciated that I _had_ kept consistency with Nitro and were happy to discuss later on.

Which I appreciated.

But I realise that's not what you meant. Happy to remove the unused option.

In any case it did not diverge from Nitro behaviour, and of course would have switched to Nitro implementation when it was available.

pi0 (approved):

PR #27131, opened 2024-05-09
feat(nuxt): add `useResponseHeader` composable

### LINKED ISSUE #27075: Suggestion for new composables: `useResponseHeader(s)` similar to existing `useRequestHeader` composable
### Describe the feature

My feature suggestion is to add 1 or 2 new composables:

- `useResponseHeader`
- `useResponseHeaders`

Which would both be very similar to their existing equals:
- [useRequestHeader](https://nuxt.com/docs/api/composables/use-request-header)
- [useRequestHeaders](https://nuxt.com/docs/api/composables/use-request-headers)

Without having looked extensively at these existing composables, I think you might even be able to reuse a lot of the same scaffolding for creating `useResponseHeader(s)`.

Although I have not contributed before, I am experienced in building with Nuxt and would be willing (glad!) to implement this feature myself.

Motivation for this feature comes from my own desire to have a simple way to set response headers. For what it's worth, I did a quick search for issues and more people are asking about this:
- https://github.com/nuxt/nuxt/discussions/25091
- https://github.com/nuxt/nuxt/discussions/25425 
- https://github.com/nuxt/nuxt/discussions/21920

(It's not impossible right now of course, but having a dedicated composable for this in the docs could increase discoverability/usability.)

### Additional information

### Final checks

jelmerdemaat:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/27075

### 📚 Description

Adds 2 new composables:

- `useResponseHeader`
- `useResponseHeaders`

See referenced issue for details.

I tried my best to cover enough ground to make a draft PR:

1. Proposed code for new composables
2. Written 2 new documentation pages

But I still need some help :) With the following questions:

- I'm not familiar with writing tests, is it expected that I add tests and if so, where do I start?
- Right now I propose to return nothing from the functions, but we could also return `true` if successful and `false` if failed? Any ideas for better return values?
- On naming the functions. The name "use" responseHeader kinda implies a return value, as it implies consuming something... Would it be better to name them `setResponseHeader`? That would be in line with the h3 library but, be out of character for a Nuxt composable.
- Do I put the Nuxt versions above the functions myself, or are these added later?
- Typings: right now I used `string` for both header `name` and `value`. But the h3 packages includes a type `HTTPHeaderName` which seems perfect for `name`. But the other composables about headers also didn't use this type, so I didn't use it. Should I use it? Is this something we could improve for all composables about headers?


danielroe:
I agree with you - I think that `useResponseHeader` doesn't make sense as a way of setting response headers. I think it should more likely return a computed value, which allows both getting and setting response headers.

For example:

```ts
const someHeader = useResponseHeader('x-some-header')
console.log(someHeader.value) // existing value
someHeader.value = 'new value'
```

jelmerdemaat:
@danielroe Any thoughts on my proposed solution before I start building it?

danielroe:
Yes, that looks good to me!

jelmerdemaat:
@danielroe I have updated the `useResponseHeader` code to reflect my proposal. Also updated the docs for it.

To be honest I am unsure about the added value of the plural `useResponseHeaders` and also unsure about how to implement it. We could also go forward with only the singular form? Or I would be glad to update the code for `useResponseHeaders` if you have any tips about implementation (e.g. with a computed, or only as a setter with an object parameter, or...)

danielroe:
I agree, let's drop `useRequestHeaders`...

jelmerdemaat:
I deleted the plural form. 

How do I proceed? I think I need to do some checks/tasks before merging?

jelmerdemaat:
I added a very basic test. I really don't know much about testing Nuxt, so if further work is required, I need some help.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -61,6 +61,27 @@ export function setResponseStatus (arg1: H3Event | number | undefined, arg2?: nu
   }
 }
 
+/** @since 3.13.2 */
```
danielroe, line 64:
```suggestion
/** @since 3.14.0 */
```

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -61,6 +61,43 @@ export function setResponseStatus (arg1: H3Event | number | undefined, arg2?: nu
...
+    }
+    return ref()
+  }
+
+  const event = useRequestEvent()
+  if (!event) {
+    if (import.meta.dev) {
+      return computed({
+        get: () => undefined,
+        set: () => console.warn('[nuxt] Setting response headers is not supported in the browser.'),
```
jelmerdemaat, line 81: I think the warning text needs to be changed here? `import.meta.client` was already checked in the code above. In this scenario the `event` is missing

danielroe: you are quite right!

PR #27138, opened 2024-05-09
docs: improve testing component examples 

### LINKED ISSUE #27120: docs: improve testing component example to avoid auto import misunderstanding
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools, modules
- Runtime Modules:  @nuxt/test-utils/module@3.12.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-aygvkp?file=components%2Fcomponent.nuxt.test.ts

Open the stackblitz and notice that the editor complains:

```
Cannot find name 'SomeComponent'. Did you mean 'component'?(2552)
component.nuxt.test.ts(10, 9): 'component' is declared here.
```

Run the test suite and notice that the tests fail with:

```
ReferenceError: SomeComponent is not defined
```

### Describe the bug

I have followed the instructions on https://nuxt.com/docs/getting-started/testing and the way I read [this section](https://nuxt.com/docs/getting-started/testing#%EF%B8%8F-helpers) `SomeComponent` should be auto-imported into that test.

The test suite runs successfully and all tests pass if I uncomment the SomeComponent import on line 4 but according to the example in the doc, that isn't necessary. Also, when I uncomment line 4, I get the following error:

```
Cannot find module './SomeComponent' or its corresponding type declarations.(2307)
```

Compare this to a similar import in `SomeWrapperComponent.vue` which does not cause the error.

### Additional context

_No response_

### Logs

_No response_

michaelmcandrew:
### 🔗 Linked issue

Resolves #27120.

Thanks @danielroe for the pointer on the import shortcut.

I made a PR with some related tweaks - hopefully improvements.

- I removed the declaration of the Components and the associated Component type import since they are no longer necessary.
- I split the first example in two since the comments indicate they are separate files.
- I used the @ directory alias to do the app import (`import App from '@/app.vue'`) but maybe there is another way?

### DIFF
--- a/docs/1.getting-started/11.testing.md
+++ b/docs/1.getting-started/11.testing.md
@@ -160,20 +160,27 @@ export default defineVitestConfig({
 `mountSuspended` allows you to mount any Vue component within the Nuxt environment, allowing async setup and access to injections from your Nuxt plugins. For example:
 
 ```ts twoslash
-import type { Component } from 'vue'
 import { it, expect } from 'vitest'
-declare const SomeComponent: Component
-declare const App: Component
 // ---cut---
 // tests/components/SomeComponents.nuxt.spec.ts
 import { mountSuspended } from '@nuxt/test-utils/runtime'
+import { SomeComponent } from '#components'
 
 it('can mount some component', async () => {
     const component = await mountSuspended(SomeComponent)
     expect(component.text()).toMatchInlineSnapshot(
-        'This is an auto-imported component'
+        '"This is an auto-imported component"'
     )
 })
+
+```
+
+```ts twoslash
+import { it, expect } from 'vitest'
+// ---cut---
+// tests/components/SomeComponents.nuxt.spec.ts
+import { mountSuspended } from '@nuxt/test-utils/runtime'
+import App from '~/app.vue'
 
 // tests/App.nuxt.spec.ts
 it('can also mount an app', async () => {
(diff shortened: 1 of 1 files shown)


docs/1.getting-started/11.testing.md
```diff
@@ -215,13 +220,11 @@ it('can render some component', async () => {
...
 
 ```ts twoslash
-import type { Component } from 'vue'
 import { it, expect } from 'vitest'
-declare const SomeComponent: Component
-declare const App: Component
 // ---cut---
 // tests/App.nuxt.spec.ts
 import { renderSuspended } from '@nuxt/test-utils/runtime'
+import App from '@/app.vue'
```
danielroe, line 227:
```suggestion
import App from '~/app.vue'
```

danielroe (commented):
This looks great - just need to confirm it will pass twoslash type testing - it might need some more basic comments.

Also (tiny thing) we generally prefer the tilde alias in the docs.

docs/1.getting-started/11.testing.md
michaelmcandrew: Thanks - I updated the @s to ~s and gave it a conventional commit message and force pushed (which now that I have properly read the contribution docs I am thinking might not have been necessary since it'll use the PR title).

Anyhow, hopefully all good now. G

danielroe (approved):

danielroe:
Thank you ❤️

PR #27149, opened 2024-05-10
fix: index.d.ts for plugins (#app → nuxt/app)

Cheaterman:
### 📚 Description

I tried to manually type my plugin (I'm making a custom $fetch) because *.ts is linted by non-Nuxt-aware tslint/tsserver/etc instead of volar and therefore (I assume) can't simply infer types from the defineNuxtPlugin call.

I banged my head against the wall for a good while until I realized `#app` should actually read `nuxt/app`. :-)

### DIFF
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -216,7 +216,7 @@ If you need to use a provided helper _within_ another plugin, you can call [`use
 For advanced use-cases, you can declare the type of injected properties like this:
 
 ```ts [index.d.ts]
-declare module '#app' {
+declare module 'nuxt/app' {
   interface NuxtApp {
     $hello (msg: string): string
   }


danielroe:
I think we would recommend instead to point tslint to your `tsconfig.json` - is there something that isn't working with that?

danielroe:
I'm going to close this for now as it's intentionally `#app` but please do open an issue so we can discuss - maybe with a reproduction 🙏

Cheaterman:
@danielroe Aha, I might have found something. The part of my setup that actually broke `'#app'` is using `"baseUrl": "."` in `tsconfig.json` like so:
```json
{
  // https://nuxt.com/docs/guide/concepts/typescript
  "extends": "./.nuxt/tsconfig.json",
  "compilerOptions": {
    "plugins": [{ "name": "@vue/typescript-plugin" }],
    "baseUrl": "."
  }
}
```
Is that not supposed to work? (the idea was to have absolute imports relative to my project root folder, eg `import type { Xyz } from 'composables/useXyz'`)

danielroe:
No, we do not recommend it or specifically handle this use case. From the TypeScript docs:

> This feature was designed for use in conjunction with AMD module loaders in the browser, and is not recommended in any other context. As of TypeScript 4.1, `baseUrl` is no longer required to be set when using paths.
>
> https://www.typescriptlang.org/tsconfig/#baseUrl

Cheaterman:
Well, thanks a lot for the clarification! I assume if I want absolute imports with Nuxt, I instead should use `@` or `~` syntax. Thanks again :-)

Issue #12414, opened 2023-01-13
Consistent behaviour of scanning files


ineshbose:
### Describe the feature

Scanning files in `composables/` and `plugins/` have different behaviours - only top level files are scanned in `composables/`, while `[...]/index.ts` is also scanned in `plugins/` (https://nuxt.com/docs/guide/directory-structure/composables#how-files-are-scanned, https://nuxt.com/docs/guide/directory-structure/plugins#which-files-are-registered). I know that it can be configured in `imports.dirs` but the default behaviour could be configured to have these as the same.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

danielroe:
Now that your PR is merged - thank you! - let's track the progress of this issue in https://github.com/nuxt/nuxt/issues/14358.

PR #27187, opened 2024-05-13
fix(nuxt): scan folder indices for middleware

### LINKED ISSUE #14358: Scan plugins from top level `/plugins` folder only
Similar to nuxt/framework#6025, this makes directory scanning simpler and more consistent between dirs. Plugins nested in dirs can be manually added with `plugins: []` in `nuxt.config` or a reexport to top level `plugins/` directory.

### LINKED ISSUE #25331: Scan `<subdir>/index` files for composables, plugins, middleware, etc.
### Describe the feature

This PR will align scanning behavior for `plugins` with existing behavior for e.g. composables or middleware by only scanning top-level folders by default.

https://github.com/nuxt/nuxt/blob/648ef069931336b8ee0c0e48fef3ee700ebc3dbe/packages/kit/src/plugin.ts#L23-L27

Related: https://github.com/nuxt/nuxt/pull/18418/files

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25331
closes https://github.com/nuxt/nuxt/issues/14358
related: https://github.com/nuxt/nuxt/issues/12414

### 📚 Description

This moves ahead to undo the previous move we made to remove scanning of `middleware/*/index` files, in the interests of consistency.

It might be nice also to update behaviour for `components/` and `utils/` so things are fully consistent across the board (what do you think @antfu?).

### DIFF
--- a/packages/nuxt/src/core/app.ts
+++ b/packages/nuxt/src/core/app.ts
@@ -179,7 +179,12 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   app.middleware = []
   for (const config of reversedConfigs) {
     const middlewareDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.middleware || 'middleware'
-    const middlewareFiles = await resolveFiles(config.srcDir, `${middlewareDir}/*{${nuxt.options.extensions.join(',')}}`)
+    const middlewareFiles = await resolveFiles(config.srcDir, [
+      `${middlewareDir}/*{${nuxt.options.extensions.join(',')}}`,
+      ...nuxt.options.future.compatibilityVersion === 4
+        ? [`${middlewareDir}/*/index{${nuxt.options.extensions.join(',')}}`]
+        : [],
+    ])
     for (const file of middlewareFiles) {
       const name = getNameFromPath(file)
       if (!name) {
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/core/app.ts
```diff
@@ -179,7 +179,12 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
   app.middleware = []
   for (const config of reversedConfigs) {
     const middlewareDir = (config.rootDir === nuxt.options.rootDir ? nuxt.options : config).dir?.middleware || 'middleware'
-    const middlewareFiles = await resolveFiles(config.srcDir, `${middlewareDir}/*{${nuxt.options.extensions.join(',')}}`)
+    const middlewareFiles = await resolveFiles(config.srcDir, [
+      `${middlewareDir}/*{${nuxt.options.extensions.join(',')}}`,
+      ...nuxt.options.future.compatibilityVersion === 4
```
manniL, line 184: Do we need a fine-grained flag for it too?

antfu:
I would be happy to see that move, yes! Thanks

packages/nuxt/src/core/app.ts
danielroe: I don't think so, as I've provided a hook. The fine grained flags will be removed with v4 release but the hook will continue to work.

TheAlexLichter (approved):

PR #27208, opened 2024-05-14
docs: add custom useFetch recipe

atinux:
Migrate from https://notes.atinux.com/nuxt-custom-fetch to a Nuxt recipe as it has a lot of traffic and helps many people on this important feature.


MuhammadM1998:
I'm using a similar setup with both `$fetch` and `useFetch` wrappers inside the same composable without plugins and it's working correctly. Is there a drawback of this that I'm missing? Maybe that would be good to clarified in the docs too

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -0,0 +1,94 @@
+---
+navigation.title: 'Custom useFetch'
+title: Custom useFetch in Nuxt
+description: How to create a custom fetcher for calling your external API in Nuxt 3.
+---
+
+When working with Nuxt, most of the time your are making the frontend and fetching an API made with another language (ex: PHP with Laravel).
```
DamianGlowala, line 7:
```suggestion
When working with Nuxt, most of the time you are making the frontend and fetching an API made with another language (e.g. PHP with Laravel).
```

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -0,0 +1,94 @@
...
+
+When we created the [`useFetch`](/docs/api/composables/use-fetch) composable and [`$fetch`](/docs/api/utils/dollarfetch) utility function, we made them immutable on purpose to avoid any side-effects. This is important to keep a consistent behaviour through your application and avoid any modules to break your API calls, for example.
```
DamianGlowala, line 9:
```suggestion
When we created the [`useFetch`](/docs/api/composables/use-fetch) composable and [`$fetch`](/docs/api/utils/dollarfetch) utility function, we made them immutable on purpose to avoid any side effects. This is important to keep a consistent behavior throughout your application and avoid any modules to break your API calls, for example.
```

DamianGlowala (changes requested):
I would generally suggest replacing vast majority of occurences of 'we' (e.g. 'we created', 'we are storing') with a form more commonly used in the docs.

E.g. given the following sentence:

> We are currently discussing with the core team (...)

... who are 'we' to the readers if it isn't the core team?

manniL:
Might be worth linking https://www.youtube.com/watch?v=jXH8Tr-exhI here too (happy to link back to the blog post instead of your personal one after this was published 👌)

danielroe (commented):
I've pushed a couple of changes, including to the example (for handling different possibilities of `HeadersInit`). It would be worth checking that everything still works as expected.

what do you think @Atinux?

maximepvrt:
@Atinux Will this duplicate https://nuxt.com/docs/examples/advanced/use-custom-fetch-composable ? Should this example be deleted ?

Atinux:
@maximepvrt I created a PR on https://github.com/nuxt/examples/pull/75 to update the link

I think we could update the example based on Daniel changes though.

@danielroe it would be nice to have a utility for that because having 3 lines to do the same thing is quite frustrating. @pi0 is this something that `ofetch` could provide?

Thinking of:

```ts
const api = $fetch.create({
    baseURL: 'https://api.nuxt.com',
    onRequest({ request, setHeader, error }) {
      if (session.value?.token) {
        setHeader('Authorization', `Bearer ${session.value?.token}`)
      }
    }
})

**Comment by Atinux** [comment]
@DamianGlowala applied the changes.

@manniL linked your video :)

**Comment by pi0** [comment]
Having standard way in ofetch context to set/get headers seems good idea. I'm thinking to adopt usage of `Header` (via a lazy) getter. (Therefore using `ctx.headers` instead of restructuring args in example would be more performant suggestion)

**Comment by Atinux** [comment]
@manniL I added a redirect instead :)

**Comment by Anton-Plagemann** [comment]
Hi @Atinux,

thanks for adding the recipe, it helped a lot! 🙏 
While implementing it, I came across an optimization regarding the `useAPI`-composable that I thought would be valuable to share:

Instead of this:

```ts [composables/useAPI.ts]
import type { UseFetchOptions } from 'nuxt/app'

export function useAPI<T>(
  url: string | (() => string),
  options: Omit<UseFetchOptions<T>, 'default'> & { default: () => T | Ref<T> },
) {
  return useFetch(url, {
    ...options,
    $fetch: useNuxtApp().$api
  })
}
```

I used this:

```ts
export const useAPI: typeof useFetch = function (url, options) {
  return useFetch(url, {
    ...options,
    $fetch: useNuxtApp().$api,
  });
};
```

By using automatic inference the code becomes much more clean and also more type-safe (as url isn't just a string now, but a NitroFetchRequest).
Feel free to add it to the docs if you feel like it would be an enhancement 😊

Issue #26444, opened 2024-03-22
Improved Directory Structure


Atinux:
### Discussed in https://github.com/nuxt/nuxt/discussions/20251

<div type='discussions-op-text'>

<sup>Originally posted by **Atinux** April 13, 2023</sup>
Nuxt 3 is now a fullstack framework thanks to [Nitro](https://nitro.unjs.io) with the ability to have [auto-imported modules](https://nuxt.com/docs/guide/directory-structure/modules) and [layers](https://nuxt.com/docs/getting-started/layers), I believe it's time to rethink the directory structure to give more separation for the "app" level.

This is the proposal I am thinking of:

```bash
.output/
.nuxt/
app/
  assets/
  components/
  composables/
  layouts/
  middleware/
  pages/
  plugins/
  utils/
  main.vue
  router.options.ts
content/
layers/
modules/
node_modules/
public/
server/
  api/
  middleware/
  plugins/
  routes/
  utils/
nuxt.config.ts
```

Do you see any downside of this approach?

The minimal starter would then be:

```bash
app/
  main.vue
public/
  favicon.ico
nuxt.config.ts
```

HigherOrderLogic:
Shouldn't we also move the `content/` dir into the `app/` dir? I think that way, it would align better with the usage of each dir.

Atinux:
Not sure for `content/` as I found it similar to `public/` and the data is not bundled.

It also allows having a minimal usage when used with a Nuxt Layer:

```ts
content/
  index.md
nuxt.config.ts <-- Using extends: 'my-docs-theme'
```

Zihan-Hu:
(BTW should `middleware` dir be `middlewares`?)

they-cloned-me:
Love it! 💚

What will be the best approach to share common utils that can be used in `/server` and `/app` e.g some string manipulation utilities.
Put’em in `/layers`?

Strift:
Love the new proposal!

Would you need to follow this structure to upgrade from v3 to v4?

Laravel recently changed its boilerplate folder structure. Although the new version works with the previous folder structure, users fear that if they don't adapt to the new structure, future code might break because they have things in the wrong place.

Atinux:
I believe we can easily detect the old format and have a way to upgrade to the new one by running either `nuxt dev` or with a `nuxi upgrade` command.

ennioVisco:
> ### Discussed in #20251
[...]

PR #27221, opened 2024-05-14
feat(nuxt): auto-register layers in `layers/` directory

danielroe:
### 🔗 Linked issue

follow up on https://github.com/nuxt/nuxt/issues/26444

### 📚 Description

This auto-registers layers located in a top-level `~~/layers` directory. To avoid collisions with user layer configuration, it adds a new internal `_extends` key to nuxt options, and then instructs `c12` to read from that. This seems like a more straightforward/correct way than attempting to add an auto-registration feature to `c12` itself.

Would love your thoughts particularly @pi0 with regard to c12 approach.

### DIFF
--- a/packages/kit/src/loader/config.ts
+++ b/packages/kit/src/loader/config.ts
@@ -4,6 +4,8 @@ import type { ConfigLayer, ConfigLayerMeta, LoadConfigOptions } from 'c12'
 import { loadConfig } from 'c12'
 import type { NuxtConfig, NuxtOptions } from '@nuxt/schema'
 import { NuxtConfigSchema } from '@nuxt/schema'
+import { globby } from 'globby'
+import defu from 'defu'
 
 export interface LoadNuxtConfigOptions extends LoadConfigOptions<NuxtConfig> {}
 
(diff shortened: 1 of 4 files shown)


Atinux:
Seems good to me at first sight.

I would maybe named it `_layers` to be similar to the layers directory for consistency.

pi0:
- This looks like a good approach in Nuxt to try
- I like the idea of `_layers` @Atinux mentioned it also marks it better for me as not-namespaced layers and might even eventually become standard for c12 and nitro

danielroe:
`._layers` is already used internally for all resolved layers, so unfortunately I felt we had to avoid it.

It should be fine as we are not exposing this or even allowing users to provide their own config.

Rigo-m:
Will each layer have it's own tsconfig?

Atinux:
> `._layers` is already used internally for all resolved layers, so unfortunately I felt we had to avoid it.
> 
> It should be fine as we are not exposing this or even allowing users to provide their own config, so it's purely internal within this file.

Damn, then it's good for me!

atinux (approved):

PR #27224, opened 2024-05-14
feat(nuxt): check key dependencies when loading nuxt

huang-julien:
### 📚 Description

Hey :wave: 

This PR integrates https://github.com/huang-julien/nuxt-core-version-check into nuxt core. The reason is that we can forget to update @nuxt/kit or @nuxt/schema when working with a lot of workspaces in a big monorepo.

Having a version mismatch on theses two key modules can trigger vite pre-transform error, resulting in errors that are complicated to debug (there's an issue for that)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -621,4 +628,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
...
 }
 
+async function checkDependencyVersion(name: string, nuxtVersion: string): Promise<void> {
+  const path = await resolvePath(name).catch(() => null)
+
+  if (!path) return
+  const { version } = await readPackageJSON(path)
+
+  if (version !== nuxtVersion) {
+    console.error(`Version mismatch for ${name} and nuxt: expected ${nuxtVersion} (nuxt) but got ${version}`)
```
manniL, line 638: Can we somehow hint people what to do on a version mismatch?

huang-julien: Do you have any specific message in mind ? Should we say to check the package.json (or more ) ?

manniL: @huang-julien Maybe ask to check the package.json and/or refresh the lockfile? 🤔

packages/nuxt/src/core/nuxt.ts
```diff
@@ -621,4 +628,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
...
 }
 
+async function checkDependencyVersion (name: string, nuxtVersion: string): Promise<void> {
+  const path = await resolvePath(name).catch(() => null)
+
+  if (!path) { return }
+  const { version } = await readPackageJSON(path)
+
+  if (version !== nuxtVersion) {
+    console.error(`Version mismatch for ${name} and nuxt: expected ${nuxtVersion} (nuxt) but got ${version}`)
```
danielroe, line 638: shall we use `console.warn` instead?

packages/nuxt/src/core/nuxt.ts
```diff
@@ -621,4 +628,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
   return nuxt
 }
 
+async function checkDependencyVersion (name: string, nuxtVersion: string): Promise<void> {
+  const path = await resolvePath(name).catch(() => null)
+
+  if (!path) { return }
+  const { version } = await readPackageJSON(path)
+
+  if (version !== nuxtVersion) {
```
danielroe, line 637: shouldn't we accept _newer_ versions of kit + schema? We could use semver `gte` to check...

packages/nuxt/src/core/nuxt.ts
```diff
@@ -621,4 +628,15 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
...
+    console.warn(`Version mismatch for ${name} and nuxt: expected ${nuxtVersion} (nuxt) but got ${version}. This might lead to unexpected behavior. Check your package.json or refresh your lockfile.`)
```
danielroe, line 638: Might be worth adding a few lines in the docs, and linking to it so we can update with workarounds or more explanation later on.

danielroe (approved):

PR #27258, opened 2024-05-16
feat(nuxt,schema): add new `appId` option and improve chunk determinism

danielroe:
### 📚 Description

This adds a new `appId` option and passes both `appId` and `buildId` via runtimeConfig. This means we don't have all chunk hashes changing on _every_ build. (Caused by entry file chunk hash changing with new buildId, and then every other file that imported the entry file changing because _that_ file name changed.)

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -535,12 +535,11 @@ export default defineUntypedSchema({
    */
   runtimeConfig: {
     $resolve: async (val: RuntimeConfig, get): Promise<Record<string, unknown>> => {
-      const [app, appId, buildId] = await Promise.all([get('app') as Promise<Record<string, string>>, get('appId') as Promise<string>, get('buildId') as Promise<string>])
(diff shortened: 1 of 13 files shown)


packages/schema/src/config/common.ts
```diff
@@ -154,6 +154,13 @@ export default defineUntypedSchema({
     $resolve: async (val: string | undefined, get): Promise<string> => resolve(await get('rootDir') as string, val || '.nuxt'),
   },
 
+  /**
+   * For multi-app projects, the unique name of the Nuxt application.
+   */
+  appId: {
+    $resolve: (val: string) => val ?? 'nuxt-app',
```
nicolaspayot, line 161: Don't we have a risk of name collision in case `appId` is not filled in in user land and default value `nuxt-app` is used multiple times?

JCtapuk: well, I would add Symbol('nuxt-app') so that the name does not conflict

PR #27273, opened 2024-05-19
docs: add additional example for setting `baseURL` and `cdnURL`

itchina110:
Correct the Example of baseURL and cdnURL


manniL:
Using environment variables to set these values should work and is correct I think 🤔

danielroe:
The example should be correct. If it is not working for you, would you mind opening an issue with a reproduction? 🙏

itchina110:
As a beginner with Nuxt, I find the documentation at ["Nuxt Configuration"](https://nuxt.com/docs/api/nuxt-config) suggesting configuration in the `nuxt.config.ts` file. However, the examples for `baseURL` and `cdnURL` in the documentation are clearly not suitable for `nuxt.config.ts`. This can be very confusing for beginners. I believe the official documentation should be improved to consider various scenarios and provide clear descriptions.

danielroe:
Understood! I've updated the PR so it adds an extra example rather than removing the one about overriding at runtime.

Issue #25984, opened 2024-02-27
Add partitioned option to useCookie


achenieux-eyein:
### Describe the feature

Hello, currently Nuxt 3 useCookie composable does not have the option for a cookie to be partitioned.

Google Chrome is phasing out third-party cookies and it should be the case by the end of 2024. 

I have a use case of a website that must work inside an iframe, so I have to set SameSite to "None" but it's then considered as a 3rd party cookie. So the new CHIPS partitioned option is the solution as presented here: [https://developers.google.com/privacy-sandbox/3pcd/chips?hl=en](https://developers.google.com/privacy-sandbox/3pcd/chips?hl=en)

I want to be able to do something like this:
`const cookie = useCookie('mycookie', {secure: true, sameSite: 'none', httpOnly: false, partitioned: true})`

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

DamianGlowala:
There's a pending PR for adding `Partitioned` attribute:
https://github.com/unjs/cookie-es/pull/21

achenieux-eyein:
Any news on this? I see the unjs/cookie-es PR you sent is now merged.

DamianGlowala:
You can already do this in Nuxt:
```
useCookie('foo', {
  partitioned: true
})
```

achenieux-eyein:
Oh thank you, I just tested it and it seems to work. I just checked the documentation and it's not here yet: [https://nuxt.com/docs/api/composables/use-cookie](https://nuxt.com/docs/api/composables/use-cookie)

PR #27297, opened 2024-05-21
docs: document `partitioned` attribute of `useCookie`

DamianGlowala:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/25984#issuecomment-2123481120 (thanks @achenieux-eyein!)

### 📚 Description

Adds missing documentation for `partitioned` attribute of `useCookie` composable (and removes irrelevant line breaks in other parts).

### DIFF
--- a/docs/3.api/2.composables/use-cookie.md
+++ b/docs/3.api/2.composables/use-cookie.md
@@ -87,7 +87,7 @@ Be careful when setting this to `true`, as compliant clients will not send the c
 
 ### `partitioned`
 
-Specifies the `boolean` value for the [`Partitioned` `Set-Cookie`](rfc-cutler-httpbis-partitioned-cookies) attribute. When truthy, the `Partitioned` attribute is set, otherwise it is not. By default, the `Partitioned` attribute is not set.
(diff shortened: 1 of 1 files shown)


danielroe (approved):

docs/3.api/2.composables/use-cookie.md
```diff
@@ -74,22 +71,29 @@ If neither of `expires` and `maxAge` is set, the cookie will be session-only and
...
 
 ::warning
-Be careful when setting this to `true`, as compliant clients will not send the cookie back to
-the server in the future if the browser does not have an HTTPS connection. This can lead to hydration errors.
+Be careful when setting this to `true`, as compliant clients will not send the cookie back to the server in the future if the browser does not have an HTTPS connection. This can lead to hydration errors.
+::
+
+### `partitioned`
+
+Specifies the `boolean` value for the [`Partitioned` `Set-Cookie`](rfc-cutler-httpbis-partitioned-cookies) attribute. When truthy, the `Partitioned` attribute is set, otherwise it is not. By default, the `Partitioned` attribute is not set.
```
danielroe, line 90: Looks like this link needs to be updated.

DamianGlowala, line 90:
```suggestion
Specifies the `boolean` value for the [`Partitioned` `Set-Cookie`](https://datatracker.ietf.org/doc/html/draft-cutler-httpbis-partitioned-cookies#section-2.1) attribute. When truthy, the `Partitioned` attribute is set, otherwise it is not. By default, the `Partitioned` attribute is not set.
```

DamianGlowala: Good spot, replaced the link 👍

Issue #25084, opened 2024-01-06
Nuxt3 global middleware called twice for 404 page


alex-zabyshny:
### Environment

- Operating System: Darwin
- Node Version:     v18.16.0
- Nuxt Version:     3.9.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.7.5
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/alex-zabyshny/nuxt3-404-page-issue

### Describe the bug

Clone project and perform the following steps:

1. `pnpm install`
2. `pnpm build`
3. `pnpm preview`

Then open the project in a browser and go to any non-existent page. For example: `http://localhost:3000/noexistingpage`
In the server console, you will see that the middleware is called twice for **the same** route:
<img width="447" alt="Screenshot 2024-01-06 at 18 16 12" src="https://github.com/nuxt/nuxt/assets/93944018/32cd9135-c0ec-4222-9830-c5f5264d4778">
If you try to follow the "/" route, everything will be fine, and the middleware only calls ones.

**Expeted result:**
Middleware is called only once if the page is not found.

toto6038:
I would like to supplement for @alex-zabyshny that the situation can be observed on dev server as well.

<img width="607" alt="image" src="https://github.com/nuxt/nuxt/assets/50100922/32cc325e-a7eb-44a0-869d-2966242c8b68">

saifobeidat:
we're facing this, and its a bit important as we're receiving many visits (fake) on 404 routes, which causes the plugins/middleware run twice if you visit a 404 route, this lead to have many duplicate API calls and makes the CPU consumption very high

danielroe:
When an error page is rendered, it is done so as an entirely separate 'page load'. So you can check in your middleware if `import.meta.server && useError().value` is true, and avoid additional API calls if that is the case.

PR #27304, opened 2024-05-22
perf(nuxt): abort vue render when plugins throw error

### LINKED ISSUE #25518: Requests are fired twice in the error page
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      modules
- Runtime Modules:  nuxt-swiper@1.2.2, @pinia/nuxt@0.4.11
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-reload-data-with-usefetch-ugvxdv

### Describe the bug

Requests are fired twice on the error page on the server side:
1. Go to the link attached in reproduction section
2. Add something to the url where reproduction is hosted. in order to get 404 page
3. Check the logs 
image

### Additional context

Reproduces after the build as well

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25518

### 📚 Description

It turns out that when we were throwing an error in plugins (or anywhere before the vue app started rendering) we were then rendering the error page twice - once on the initial request (which is discarded) and then again when we are intentionally rendering an error page.

A possible performance improvement might be using that initial render instead of running plugins twice. But for now, I think it's safer to keep the separated error render for consistency.

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-root.vue
+++ b/packages/nuxt/src/app/components/nuxt-root.vue
@@ -1,7 +1,8 @@
 <template>
   <Suspense @resolve="onResolve">
+    <div v-if="abortRender" />
     <ErrorComponent
-      v-if="error"
+      v-else-if="error"
       :error="error"
     />
     <IslandRenderer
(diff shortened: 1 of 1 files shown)


saifobeidat:
here's a similar issue but for middlewares, will this PR fix it or its just for plugins? thanks
https://github.com/nuxt/nuxt/issues/25084

PR #27305, opened 2024-05-22
feat(nuxt): display banner when running with `compatibilityVersion`

### LINKED ISSUE #27302: Show compatibility flag when starting the server
### Describe the feature

In addition to the Nuxt and Nitro version, the compatibility flags could be shown in the terminal when starting the Nuxt dev server, so people using the "future" flags are sure they work as expected.

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/27302

### 📚 Description

When users are running Nuxt with a different compatibilityVersion than the current one, this adds a banner.

CleanShot 2024-05-22 at 17 53 31@2x

Wording improvements are particularly welcome 🙏

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -560,7 +560,7 @@ async function initNuxt (nuxt: Nuxt) {
   // Show compatibility version banner when Nuxt is running with a compatibility version
   // that is different from the current major version
   if (!(satisfies(nuxt._version, nuxt.options.future.compatibilityVersion + '.x'))) {
-    console.info(`running with compatibility version \`${nuxt.options.future.compatibilityVersion}\``)
+    console.info(`Running with compatibility version \`${nuxt.options.future.compatibilityVersion}\``)
   }
 
   await nuxt.callHook('ready', nuxt)
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -13,7 +13,7 @@ import fse from 'fs-extra'
 import { withTrailingSlash, withoutLeadingSlash } from 'ufo'
 
 import defu from 'defu'
-import { gt } from 'semver'
+import { gt, satisfies } from 'semver'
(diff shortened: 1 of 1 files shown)


danielroe:
cc: @pi0 in case you think it's better in nuxt/cli, but I figured that this is very much tied to the nuxt instance itself and its options, which might be manipulated by modules (e.g. not like Nuxt/Nitro versions which CLI can resolve).

packages/nuxt/src/core/nuxt.ts
```diff
@@ -557,6 +557,12 @@ async function initNuxt (nuxt: Nuxt) {
     addPlugin(resolve(nuxt.options.appDir, 'plugins/payload.client'))
   }
 
+  // Show compatibility version banner when Nuxt is running with a compatibility version
+  // that is different from the current major version
+  if (!(satisfies(nuxt._version, nuxt.options.future.compatibilityVersion + '.x'))) {
+    console.info(`running with compatibility version \`${nuxt.options.future.compatibilityVersion}\``)
```
manniL, line 563:
```suggestion
    console.info(`Running with compatibility version \`${nuxt.options.future.compatibilityVersion}\``)
```

Would probably capitalize this - otherwise LGTM 👍🏻

TheAlexLichter (approved):

packages/nuxt/src/core/nuxt.ts
manniL, line 563: Some "Learn more" link would be sweet - though tough to generalize

danielroe: Agreed. I think we can skip though, as users will have explicitly opted in to `compatibilityVersion` in order for banner to show, so I think they will know where to look.

pi0 (approved):
LGTM (perhaps a shortlink in mesage to docs could be helpful for future notes and updates?)

danielroe:
Okay - let's definitely do that then. Merging this and will update once I create the short link. 🙏

PR #27313, opened 2024-05-22
feat(nuxt): handle nuxt route injection for `this.$route`

### LINKED ISSUE #27272: [Performance] Using this.$route inside a computed triggers reactivity while we are changing pages
### Environment

- Operating System: Darwin
- Node Version:     v20.10.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Package Manager:  pnpm@8.13.1
- Builder:          -
- User Config:      ssr, experimental, sourcemap, app, hooks, devServer, modules, devtools, imports, vite, nitro, components, pinia, i18n, runtimeConfig
- Runtime Modules:  @internal/cms-renderer@1.0.0, @internal/module-proxy@1.0.0, @nuxtjs/i18n@8.2.0, @nuxt/devtools@1.0.6
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-89dmst?file=pages%2F[...slug].vue

### Describe the bug

Issue: 
 - When using options API with Vue and creating a computed that utilizes `this.$route` inside it, the computed will trigger when we click on a `NuxtLink` while we are still changing pages. Being inside a computed seems special because if you inject `$route` directly into the template it does not update until page change.
 
 Observing the issue should be easy with the provided reproduction.
 
 Notes: 
  - The same cannot be said when we use Composition API and the `useRoute` composable, if we create a computed that uses the value of the composable inside it will not trigger during a route change (correct behavior for me).
  - The first query object is a direct template injection of `$route`, the second one is a computed using `this.$route` inside and you can see that it updates while changing pages, the third one is a computed using the `useRoute` composable value and it only updates after route change.
  
  Impact: 
   - Currently our big E-Commerce website has some heavy computations associated with the `this.$route` while we are migrating to full Composition API, each page change it's triggering some computeds and, by consequence, patching a full component with a lot of data. 
   - I got a workaround by having a proxy route computed that checks if the route the computed is re-calculating it's the same route the component was originally created at (not scalable for our use case).
   
   

### Additional context

_No response_

### Logs

_No response_

jakubednar:
### 🔗 Linked issue

Resolves #27272, following up on https://github.com/nuxt/nuxt/pull/21585

### 📚 Description

Added support for `this.$route` in script in options API to ensure that within each suspense fork we have different route objects.


danielroe (commented):
This looks great - thank you! ❤️ 

My only worry is that this might have a false positive - e.g. replacing `this.$route` in a blog post about vue-router. Maybe we could use `strip-literal` to handle this?

See example here:

https://github.com/nuxt/nuxt/blob/f209158352b09d1986aa320e29ff36353b91c358/packages/nuxt/src/core/plugins/tree-shake.ts#L33-L36

jakubednar:
Should I use `stripLiteral` for both `$route` and `this.$route` or just `this.$route` ?

RodrigoProjects:
> Should I use `stripLiteral` for both `$route` and `this.$route` or just `this.$route` ?

You lose nothing by doing  it first thing after the Magic String creation so why not? Makes sense to me!

jakubednar:
> > Should I use `stripLiteral` for both `$route` and `this.$route` or just `this.$route` ?
> 
> You lose nothing by doing it first thing after the Magic String creation so why not? Makes sense to me!

I thought the same, so I implemented it for both cases 🙌

danielroe (approved):

Issue #25812, opened 2024-02-15
environment configuration alias docs (`$production`, `$development` and `$test`)


lassediercks:
In [Getting Started with Unit Tests](https://nuxt.com/docs/getting-started/testing#registerendpoint)  using `$test` is the suggested way of handling api requests. 

@danielroe provided an [example in a PR discussion](https://github.com/nuxt/nuxt/pull/25805#discussion_r1490992101) but besides that there is no documentation about that feature

```ts
export default defineNuxtConfig({
  $test: {
    runtimeConfig: {
      apiBase: ''
    }
  }
})
```

![image](https://github.com/nuxt/nuxt/assets/1765584/85e8eb2b-704b-4249-8d30-78e0530868ae)

riddla:
There is some documentation over at https://github.com/unjs/c12#environment-specific-configuration. At least I think that this is the underlying feature used in the `nuxt.config`.

manniL:
It is also documented in https://nuxt.com/docs/getting-started/configuration#environment-overrides, which should suffice. Feel free to point out what is missing if you disagree ☺️

maartenvanhunsel:
@manniL is it possible to change the envname withing Nuxt? https://github.com/unjs/c12?tab=readme-ov-file#envname
For example I would like to use the `NUXT_SITE_ENV` variable to determine the environment.

manniL:
@maartenvanhunsel follow along https://github.com/nuxt/cli/issues/357 ☺️

lassediercks:
I do agree that the environment overrides section does greatly document the feature. 

Following my personal scenario: from a perspective of someone that explores the nuxt docs for the first time there is still not clear indication what `$test` is refering to 
<img width="770" alt="image" src="https://github.com/nuxt/nuxt/assets/1765584/3b06d77b-2612-4c0e-bc11-0e2d77d884e3">


Here is my idea how to battle that: https://github.com/nuxt/nuxt/pull/27315

PR #27315, opened 2024-05-22
docs: introduce hint and link to environment override feature (#25812)

lassediercks:
### 🔗 Linked Issue

Reference: #25812

### 📚 Description

This small quality of life improvement is aimed at helping those who are not very familiar with the Nuxt ecosystem.

Rephrasing the scenario from the issue:

A visitor stumbles upon the hint in the [testing documentation](https://nuxt.com/docs/getting-started/testing):

> Note: If your requests in a component go to external API, you can use baseURL and then make it empty using Nuxt Environment Config ($test) so all your requests will go to Nitro server.

Searching for `$test` or `Environment Config` does not yield any results, which can be a potential dead end.

As @manniL stated in the issue, it is perfectly documented in [Environment Overrides](https://nuxt.com/docs/getting-started/configuration#environment-overrides).

This PR introduces a small augmentation to the hint to make sure it is clear.

In hindsight, I understand the hint, and one might call it fussy to cross-link docs for such a small example. However, I would argue that overriding configurations at a framework level with a feature like this is not yet commonly understood by developers and therefore deserves a link like this.

### DIFF
--- a/docs/1.getting-started/11.testing.md
+++ b/docs/1.getting-started/11.testing.md
@@ -373,7 +373,7 @@ registerEndpoint('/test/', {
 })
 ```
 
-> **Note**: If your requests in a component go to external API, you can use `baseURL` and then make it empty using Nuxt Environment Config (`$test`) so all your requests will go to Nitro server.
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
Thank you - really nice 👌

Issue #20972, opened 2023-05-20
Prompting user to reload when the site has been updated


edwh:
### Describe the feature

Apologies if I've missed a feature which does this.

Some users will load a site and then have it sit in an open tab for days.  Meanwhile, the site may be changing under their feet, perhaps with important fixes which they won't pick up.  

Sometimes it would be good if the user was prompted to "reload when convenient"; more rarely it would be good if the site did a hard refresh under their feet.

It would be possible to roll my own code for this - for example Netlify has an API end point which I could poll to see if the `deploy_id` has changed since the build which the code is currently running.  

But this might be of wider interest, including in other deployment environments, so I'm opening this for discussion.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

warflash:
I recommend checking out https://github.com/nuxt/nuxt/pull/19086 and https://nuxt.com/docs/getting-started/error-handling#errors-downloading-js-chunks, perhaps those can accomplish what you're looking for already

maou-shonen:
You might want to learn about PWA

edwh:
@warflash I'm already commenting on #19086.  I'm not talking here about the case where there are issues downloading chunks after new deploys.  I'm talking about the opposite case where you have downloaded everything you need and just sit there using the site.  So the default nuxt behaviour of reloading on chunk errors, or the kinds of variations discussed in #19806, will never kick in.  

But meanwhile the site has moved on, and unless you refresh the page you won't notice.  I'm suggesting that most if not all Nuxt apps would benefit from a built-in mechanism to spot that.

@maou-shonen I could be wrong, but I don't think PWAs affect this significantly.  You can roll your own PWA updating mechanism (e.g. https://stackoverflow.com/questions/59708511/how-to-update-a-pwa-when-has-been-already-installed) but I don't think that you get one for free.

killjoy1221:
Here's a primitive update mechanism that uses the build time to check for changes.

```ts
// nuxt.config.ts
export defaultNuxtConfig({
  appConfig: {
    buildTimestamp: Date.now()
  }
})
```
```ts
// server/api/updateCheck.ts
export default defineEventHandler(() => {
[...]

PR #27324, opened 2024-05-23
feat: make checking for outdated interval configurable

IsraelOrtuno:
### 🔗 Linked issue

#20972 and more specifically this comment https://github.com/nuxt/nuxt/issues/20972#issuecomment-1828502798

### 📚 Description

The current checking for outdated builds is hardcoded to 1 hour. This allows to customise the interval or even disable this feature entirely. The default interval is `1000 * 60 * 60`.

```ts
experimental: {
    checkOutdatedBuildInterval: 1000 * 60 * 60
    // or
    checkOutdatedBuildInterval: false
}
```

I wasn't sure where to put this option and I thought under the `build` option made sense but open to move it somewhere else.


packages/schema/src/config/build.ts
```diff
@@ -118,6 +118,14 @@ export default defineUntypedSchema({
...
       },
     },
+
+    /**
+     * Set the time interval (in ms) to check for new builds.
+     *
+     * Set to `false` to disable.
+     * @type {number | false}
+     */
+    checkOutdatedBuildInterval: 1000 * 60 * 60,
```
danielroe, line 128: let's make it `experimental.appManifestInterval` for now ...

packages/nuxt/src/app/plugins/check-outdated-build.client.ts
```diff
@@ -4,16 +4,18 @@ import type { NuxtAppManifestMeta } from '../composables/manifest'
 import { onNuxtReady } from '../composables/ready'
 // @ts-expect-error virtual file
 import { buildAssetsURL } from '#internal/nuxt/paths'
+// @ts-expect-error virtual file
+import { outdatedBuildInterval } from '#build/nuxt.config.mjs'
 
 export default defineNuxtPlugin((nuxtApp) => {
-  if (import.meta.test) { return }
+  if (import.meta.test || !outdatedBuildInterval) { return }
```
danielroe, line 11: I think if it's set explicitly to `false` we can completely omit adding this plugin, here:

https://github.com/nuxt/nuxt/blob/5b1ca73886d942db634fc8f939a4ee3db16e47f1/packages/nuxt/src/core/nuxt.ts#L502

danielroe (approved):
Perfect! 👌

PR #27343, opened 2024-05-24
fix(vite): ignore layer `tsconfig.json` when resolving vue prop types

### LINKED ISSUE #26449: [Bug] [`@vue/compiler-sfc`] Read files based on Nuxt project/VFS in a monorepo
Reproduction: https://stackblitz.com/edit/github-ntpcmz?file=nuxt.config.ts,app.vue

Code: https://github.com/nuxt/nuxt/blob/62a1fee8174e039d75e9cdd04f8e283b8e0c5773/packages/vite/src/vite.ts#L115 (we want to pass `script.fs` here)

context:
- https://github.com/nuxt/ui-pro/issues/118
- https://github.com/nuxt/cli/pull/309

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26449

### 📚 Description

There are two issues here:

1. Vue's resolution algorithm. This PR exists to resolve that issue; we can force it to resolve the `tsconfig.json` of the parent project.

2. `tsconfck` (used in vite) expects to find a valid `tsconfig.json` for any file that is not in `node_modules/`.

   https://github.com/vitejs/vite/blob/6f77b2b22012ad1b810f4ec0511609ead35363dd/packages/vite/src/node/plugins/esbuild.ts#L451-L476

   Because the layer is outside of `node_modules/`, vite throws an error. I think we have to require that layers in your source directory have valid TS configs in this case. 

@ineshbose The reproduction you provided still has an issue with point 2 above, so this doesn't totally resolve it, but from your comment in https://github.com/nuxt/cli/pull/309 I think you were referring to a bigger picture. Could you provide a reproduction or test this fix on that situation and let me know whether it resolves things for you?

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -1,13 +1,15 @@
-import { existsSync } from 'node:fs'
+import { existsSync, readFileSync, realpathSync } from 'node:fs'
 import * as vite from 'vite'
 import { dirname, join, normalize, resolve } from 'pathe'
 import type { Nuxt, NuxtBuilder, ViteConfig } from '@nuxt/schema'
 import { addVitePlugin, isIgnored, logger, resolvePath } from '@nuxt/kit'
 import replace from '@rollup/plugin-replace'
 import { sanitizeFilePath } from 'mlly'
-import { withoutLeadingSlash } from 'ufo'
+import { withoutLeadingSlash, withTrailingSlash } from 'ufo'
 import { filename } from 'pathe/utils'
 import { resolveTSConfig } from 'pkg-types'
+import defu from 'defu'
+import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 
 import { buildClient } from './client'
 import { buildServer } from './server'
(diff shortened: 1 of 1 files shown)


ineshbose:
I'm afraid this solution may not suitable IMO.

> This PR exists to resolve that issue; we can force it to resolve the tsconfig.json of the parent project.

This is fine; I feel it may have side-effects that may break existing layers - we're under the assumption that all layers are subdirectories of a base-root Nuxt project.

> tsconfck (used in vite) expects to find a valid tsconfig.json for any file that is not in node_modules/. Because the layer is outside of node_modules/, vite throws an error. I think we have to require that layers in your source directory have valid TS configs in this case.

We have this error for any layers shipped as a NPM package with Vue SFCs in their dist, so they'll be in `node_modules` and we'll have this error still. In a way, the above can solve it, but that's still under the assumption that the root is a Nuxt project (this won't work for, say, a monorepo).

> The reproduction you provided still has an issue with point 2 above, so this doesn't totally resolve it, but from your comment in https://github.com/nuxt/cli/pull/309 I think you were referring to a bigger picture. Could you provide a reproduction or test this fix on that situation and let me know whether it resolves things for you?

I can't say on bigger picture - as I'm thinking to work around the code of `@vue/compiler-sfc` given the spec and how Nuxt allows layers to be shipped, but I should definitely work on a possibly easier/better/verbose reproduction. My thinking for the fix for this may have been that we configure `fs.readFile` instead where we replace layer.src with the rootDir.. still may not be ideal, so easiest becomes preparing `tsconfig.json` for each layer.

danielroe:
Agreed - makes sense. Let me know when you have a better reproduction and I can investigate further. 🙏

PR #27354, opened 2024-05-25
chore: add `baseUrl: "."` to .nuxt/tsconfig.json

sadeghbarati:
### 🔗 Linked issue

https://github.com/radix-vue/shadcn-vue/issues/291#issuecomment-1901788144

### 📚 Description

This PR will add default `"baseUrl": "."` to **.nuxt/tsconfig.json**

Will update description for context please check shadcn-vue issue


danielroe:
There are a number of reasons we choose not to  set baseURL, including the fact that typescript recommends against it.

Let's see if we can fix the upstream issue another way.

PR #27356, opened 2024-05-25
fix(vite): disable `optimizeDeps` in ssr

antfu:
### 🔗 Linked issue

related to #26783, but not sure if completely fixes 

### 📚 Description

We never need `optimizeDeps` in ssr, as `vite-node` externalize deps and use native node imports. This was probably a mistake. We found this out with @patak-dev during the test with Vite 6, which surfaces this issue. This might be a perf improvement as well.


danielroe (approved):

danielroe:
Nice find - thank you ❤️🙏

PR #27418, opened 2024-06-01
fix: update createError parameter signature to match h3 version in Nuxt.

Ray0907:
### 🔗 Linked issue
#25942 

### 📚 Description
This pull request updates the createError parameter signature to align with the h3 version used in Nuxt. The previous signature did not match, which could cause inconsistencies and potential issues when creating errors. This change ensures compatibility and improves the robustness of error handling.
ref: https://h3.unjs.io/guide/event-handler#error-handling

### DIFF
--- a/docs/1.getting-started/8.error-handling.md
+++ b/docs/1.getting-started/8.error-handling.md
@@ -162,8 +162,8 @@ const { data } = await useFetch(`/api/movies/${route.params.slug}`)
 
 if (!data.value) {
   throw createError({
-    statusCode: 404,
-    statusMessage: 'Page Not Found'
+    status: 404,
+    message: 'Page Not Found'
   })
 }
 </script>
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -285,8 +285,8 @@ export default defineEventHandler((event) => {
 
   if (!Number.isInteger(id)) {
     throw createError({
-      statusCode: 400,
-      statusMessage: 'ID should be an integer',
+      status: 400,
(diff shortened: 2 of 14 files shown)


danielroe:
`statusCode` is used consistently in Nitro, e.g.:

https://github.com/unjs/nitro/blob/d20ffcbd16fc4003b774445e1a01e698c2bb078a/src/runtime/internal/task.ts#L37-L49

The move from statusMessage -> message looks good.

What do you think @pi0?

pi0:
If i'm not wrong `statusMessage` was Node.js flavor.

[`status`](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) and [`statusText`](https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText) are Web Response props. i would try to adopt them consistently everywhere.

danielroe:
Contra my previous message, we should not update `statusMessage` -> `message`. It should be `statusMessage` -> `statusText`. `message` is a separate field that governs the _content_ of the body.

I'm a little reluctant to update the docs yet as the type of the error _returned_ from `createError` still has `statusCode` and `statusMessage` fields, which is possibly just confusing for people. (For example, an error thrown with `{ status: 400 }` would have `{ statusCode: 400 }` and no `status` property.)

`statusCode` and `status` would be used in different places. (Still required by Nitro render response, for example.) And the error object provided by Nuxt in `error.vue` and elsewhere still expects those properties. Likely we'd need to implement backwards compatibility with a getter in h3/nuxt.

As a result, I think we need to defer this PR for now. Depending on what we do in `h3`, we might still merge it in time for 4.x. What do you think @pi0?

PR #27428, opened 2024-06-02
fix(nuxt): missing `onNuxtReady` deadline types

harlan-zw:
### 📚 Description

The `onNuxtReady` function is a SSR / browser-compatable way to use `requestIdleCallback`. `requestIdleCallback` is great for queuing low-priority work to be processed in the background.

When using `onNuxtReady` to queue sync work, we don't have a type-safe way to process it without avoiding the current idle callback timeout, as the `deadline` object is not typed.

See https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API for background on this but we can see a simple example as follows:

```ts
const tasks = new Set<() => void>()
const processWork = (deadline: IdleDeadline) => {
  const task = tasks.values().next().value
  if (deadline.timeRemaining() > 0) {
    task()
    tasks.delete(task)
  } else {
    // queue work for next idle callback - avoid hitting 50ms timeout
    onNuxtReady(processWork)
  }
}

onNuxtReady(processWork)
```

It's not clear if this was not typed intentionally so it may be missing some context.


danielroe:
I'm torn on this one. It would obviously be quite a nice thing to expose this in terms of what it would enable, but semantically it's not a replacement for `requestAnimationFrame` as it doesn't run at all on the server.

I think your example would work nicely like this:

```diff
  const tasks = new Set<() => void>()
  const processWork = (deadline: IdleDeadline) => {
    const task = tasks.values().next().value
    if (deadline.timeRemaining() > 0) {
      task()
      tasks.delete(task)
    } else {
      // queue work for next idle callback - avoid hitting 50ms timeout
-     onNuxtReady(processWork)
+     requestAnimationFrame(processWork)
    }
  }
  
- onNuxtReady(processWork)
+ onNuxtReady(() => processWork())
```

harlan-zw:
I don't follow what you mean or how it relates to the PR exactly, I may not have explained it properly or am missing something.

The `onNuxtReady` function accepts a function and passes the `idleDeadline` as the first argument already, it's just missing the types for it. There is no logic changing. We may break downstream user types if they have typed the function the same.

I have provided extra context and a code sample to show how that type can be used, the example is a direct equivalent to the code recommended for running _[background tasks](https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API)_. This is not the same as `requestAnimationFrame`.

>  but semantically it's not a replacement for requestAnimationFrame as it doesn't run at all on the server.

Idle timeout doesn't run on the server? Not sure what this means

danielroe:
Sorry - my mistake - somehow got the wrong method in my mind! But my point I was trying to make is that `onNuxtReady` doesn't run until the Nuxt app is ready. It doesn't execute the calllback at all on the sever. So it's not quite the same as `requestIdleCallback`, though it could be used to dispatch a first call to `requestIdleCallback`.

I think it would be reasonable to polyfill `requestIdleCallback` but I wouldn't want to suggest `onNuxtReady` is a direct replacement for it, and adding the types would tie `onNuxtReady` quite tightly to its implementation.

harlan-zw:
Yes, this is not for work to be done on the server. This is browser background tasks.

`requestIdleCallback` is already polyfilled, this is why it's preferred to do this via `onNuxtReady` rather than re-implementing the `requestIdleCallback` callback yourself (and that it's SSR safe).

`onNuxtReady` is already coupled to `requestIdleCallback`, I'd suggest either we need to call the `onNuxtReady` hook without the `deadline: IdleDeadline`  OR just add the types. I don't see any downsides in just adding the types though

harlan-zw:
Here's a Stackblitz demo that maybe will help: https://stackblitz.com/edit/nuxt-starter-pk9gdo?file=app.vue

danielroe:
Makes sense!

Let's call it without args for now, and I'm happy to revisit later if this blocks anything. (For now, I think it's just a very slight change to your code: https://github.com/nuxt/nuxt/pull/27428#issuecomment-2157738000.)

PR #27449, opened 2024-06-04
feat(nuxt): register plugin hooks before executing plugins

hendrikheil:
### 🔗 Linked issue

implements #27438 

### 📚 Description

This PR implements the changes described in discussion #27438. It serves as a reference implementation for that discussion, mainly to have a concrete example of the proposed changes.

I'll add tests as the discussion matures.

~As this is a breaking change, I think it is best to target this change for 4.x~
This now targets 3.12.0


danielroe:
This seems really good to me ... What makes you think it might be a breaking change?

hendrikheil:
I might be overly cautious, but theoretically this change might suddenly run hook listeners that didn't run before this change.

Since it isn't really documented behavior, it might be ok to do this change anyway

docs/2.guide/2.directory-structure/1.plugins.md
```diff
@@ -81,8 +81,9 @@ Watch a video from Alexander Lichter about the Object Syntax for Nuxt plugins.
 ::
 
 ::note
-If you are using the object-syntax, the properties may be statically analyzed in future to produce a more optimized build. So you should not define them at runtime. :br
+If you are using the object-syntax, the properties may be statically analyzed in the future to produce a more optimized build. So you should not define them at runtime. :br
```
danielroe, line 84:
```suggestion
If you are using the object-syntax, the properties are statically analyzed to produce a more optimized build. So you should not define them at runtime. :br
```

danielroe (approved):

PR #27469, opened 2024-06-06
fix(nuxt): use buildAssetsDir in nuxt-teleport-island-component

andresilva-cc:
### 📚 Description

It's common in some scenarios, for example when multiple projects are running on the same domain, to define a different value for `app.buildAssetsDir`. When using `componentIslands.selectiveClient` in development, the `nuxt-teleport-island-component` uses a hardcoded value (`/_nuxt/`).

This PR changes the component to use the `buildAssetsDir` config. If the config is not set, it'll still use `/_nuxt/` as before, because that's the default value of the config.


andresilva-cc:
I opened this as a draft for 2 reasons:

1. Not sure if it's recommended to get `buildAssetsDir` from `nuxtApp.$config`
2. The test `test/bundle.test.ts > minimal nuxt application > default server bundle size (inlined vue modules)` is failing because the size was increased to `76.2k`. Not sure if this is something I need to update or if it gets updated automatically.

huang-julien (approved):

PR #27499, opened 2024-06-08
docs: updated documentation to include missing parameter

joypal23jkp:
### 📚 Description

Updated documentation for `useId` composable. In the documentation the optional  _string_ parameter `key` was missing. 
So I added that information to this `PR`.

### DIFF
--- a/docs/3.api/2.composables/use-id.md
+++ b/docs/3.api/2.composables/use-id.md
@@ -38,7 +38,7 @@ const id = useId()
 - `key`: An optional string parameter.
 
 ::warning
-If `key` is not a string then you will be thrown a **TypeError** showing `[nuxt] [useId] key must be a string`.
+If `key` is set and *not a string* then you will be thrown a **TypeError** showing `[nuxt] [useId] key must be a string`.
 ::
 
 ## Returns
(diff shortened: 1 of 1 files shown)


docs/3.api/2.composables/use-id.md
```diff
@@ -35,7 +35,11 @@ const id = useId()
 
 ## Parameters
 
-`useId` does not take any parameters.
+- `key`: An optional string parameter.
+
+::warning
+If `key` is not a string then you will be thrown a **TypeError** showing `[nuxt] [useId] key must be a string`.
```
manniL, line 41: Maybe making it more clear that this applies only when `key` is set?

```suggestion
If `key` is set and *not a string* then you will be thrown a **TypeError** showing `[nuxt] [useId] key must be a string`.
```

danielroe (commented):
The key is generated by Nuxt and shouldn't be provided by the user. (It's why it's not visible in the TypeScript function signature - we deliberately do not expose it.)

danielroe:
Closing for now - but please do follow up if you think there's a good reason to expose the key to the user?

PR #27512, opened 2024-06-10
feat(nuxt,schema): add `compatibilityDate` flag for future support

danielroe:
### 📚 Description

This adds support for `compatibilityDate` (defaulting to today's date), which should enable unlocking more features in future.

@pi0, do you think we should prompt user to add to `nuxt.config` using `magicast` immediately, or only from v4/v4-compat?

**TODO:**

- [ ] prompt user to add to `nuxt.config` using `magicast`
- [ ] print out compatibility updates in future using https://github.com/unjs/compatx

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -5,6 +5,7 @@ import { isDebug, isDevelopment, isTest } from 'std-env'
 import { defu } from 'defu'
 import { findWorkspaceDir } from 'pkg-types'
 import { randomUUID } from 'uncrypto'
+import { formatDate } from 'compatx'
 import type { RuntimeConfig } from '../types/config'
 
 export default defineUntypedSchema({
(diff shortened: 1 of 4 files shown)


packages/schema/src/config/common.ts
```diff
@@ -20,6 +21,18 @@ export default defineUntypedSchema({
...
+   * Specify a compatibility date for your app.
+   *
+   * This is used to control the behavior of presets in Nitro, Nuxt Image
+   * and other modules that may change behavior without a major version bump.
+   *
+   * We plan to improve the tooling around this feature in the future.
+   *
+   * @type {string | Record<string, string>}
+   */
+  compatibilityDate: formatDate(new Date()),
```
pi0, line 34: - I think here defaulting to undefined would be the best so we know user either set it or not and later add prompt
- formatDate maybe can be in `$resolve`?
- comptax has `DateString` we can use to add type checks for valid input

pi0 (approved):

packages/schema/src/config/common.ts
```diff
@@ -20,6 +20,18 @@ export default defineUntypedSchema({
...
 
+  /**
+   * Specify a compatibility date for your app.
+   *
+   * This is used to control the behavior of presets in Nitro, Nuxt Image
+   * and other modules that may change behavior without a major version bump.
+   *
+   * We plan to improve the tooling around this feature in the future.
+   *
+   * @type {typeof import('compatx').DateString | Record<string, typeof import('compatx').DateString>}
```
pi0, line 31: My bad forgot to mention we have `CompatibilityDateSpec` as well, it is string or object map from known platform keys to version we should used it

pi0: ~> https://github.com/nuxt/nuxt/pull/27521

PR #27564, opened 2024-06-12
feat(kit): add `isNuxtMajorVersion` compatibility util, deprecate `isNuxt2` and `isNuxt3`

lihbr:
### 🔗 Linked issue

/

### 📚 Description

This PR adds a `isNuxt4` utils that works like the existing `isNuxt2` and `isNuxt3` ones to `@nuxt/kit`.

Maybe a bit early to add it but I had to check for it preemptively already on a module so maybe that's what we want?

I've tried to check if this was done somewhere else already but couldn't, sorry if this is duplicate 🙏


DamianGlowala:
I'd suggest introducing a single util instead for any future Nuxt versions moving forward:
```
export function isNuxtVersion (majorVersion: number, nuxt: Nuxt = useNuxt()) {
  const version = getNuxtVersion(nuxt)
  return version.startsWith(`${majorVersion}.`)
}
```

lihbr:
I like that also! Happy to refactor in that direction if we agree on that

danielroe (approved):

danielroe:
Thank you for this - and yes, I like @DamianGlowala's idea too!

lihbr:
Will refactor in the evening then ☺️ hold on!

lihbr:
Done:
1. I know we've been avoiding `startsWith` when possible for performance reasons, I believe it's not applicable here as it could break the logic when we reach Nuxt 10 (well, not happening tomorrow ^^'). Maybe we want to compromise here and keep using `version[0]`?
2. `@nuxt/kit` appears with a leading `\` in the deprecation notice as otherwise it's considered as a TSDoc tag an rendered differently. The escape character doesn't show as expected when hovering the definition 👌

packages/kit/src/compatibility.ts
```diff
@@ -98,19 +98,26 @@ export async function hasNuxtCompatibility (constraints: NuxtCompatibility, nuxt
...
 /**
- * Check if current nuxt instance is version 2 legacy
+ * Check if current Nuxt instance is of specified major version
  */
-export function isNuxt2 (nuxt: Nuxt = useNuxt()) {
+export function isNuxtMajorVersion (majorVersion: 2 | 3 | 4, nuxt: Nuxt = useNuxt()) {
   const version = getNuxtVersion(nuxt)
-  return version[0] === '2' && version[1] === '.'
+
+  return version.startsWith(`${majorVersion}.`)
```
DamianGlowala, line 106: Following your advice, let's revert it to a more performance-oriented version! 😄 
```suggestion
  return version[0] === majorVersion.toString() && version[1] === '.'
```

DamianGlowala (changes requested):

danielroe (approved):

DamianGlowala (approved):

Issue #27594, opened 2024-06-13
useAsyncData unexpected behaviour when callback return falsy value


VladStepanov:
### Environment

- Operating System: Darwin
- Node Version:     v20.2.0
- Nuxt Version:     3.12.1
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.3.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

[...]

PR #27595, opened 2024-06-13
fix(nuxt): more accurate asyncData deduplication

dsfcoll:
### 🔗 Linked issue

#27594 

### 📚 Description

This MR handles more accurate asyncData client deduplication check for the cases when `useAsyncData` return falsy values


danielroe:
I think we can close this with the changes for v4 - only `undefined` should trigger re-fetching.

PR #27596, opened 2024-06-13
fix(nuxt): use portal to sync nitro/nuxt runtimeConfig + routeRules

### LINKED ISSUE #27589: @bg-dev/nuxt-directus not working with Nuxt 3.12.2
### Environment

Nuxt project info:                                                                                                                                                                  15:55:48

- Operating System: Linux
- Node Version:     v20.5.1
- Nuxt Version:     3.12.1
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.1.1
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Install a base Nuxt App.
Add @bg-dev/nuxt-directus: "^2.4.3" to package.json
Add
```
{
  modules: ['@bg-dev/nuxt-directus'],
  directus: {
    graphql: {
      enabled: false
    }
  }
}
```
App is working with nuxt 3.11.2 but not with 3.12.0 or 3.12.1

### Describe the bug

When upgrading Nuxt from 3.11.2 to 3.12.1, we get the following error for @bg-dev/nuxt-directus:
Cannot read properties of undefined (reading 'rest')

For other modules, we get similar errors (e.g. Nuxt Content or Nuxt Seo).

### Additional context

@bg-dev/nuxt-directus will do the following in module.ts, setup function:
```
nuxt.options.runtimeConfig = defu(
  nuxt.options.runtimeConfig,
  {
    app: {},
    public: {
      apollo: {},
      directus: options,
    },
  },
)
```
But then in the runtime plugin rest.ts, it starts with 
```
  const config = nuxtApp.$config.public.directus as PublicConfig & { auth: { enabled: true } }
```
But nuxtApp.$config gives:
```
{ app: { baseURL: '/', buildId: 'dev', buildAssetsDir: '/_nuxt/', cdnURL: '' },
  nitro:
   { envPrefix: 'NUXT_',
     routeRules:
      { '/__nuxt_error': [Object],
        '/_nuxt/builds/meta/**': [Object],
        '/_nuxt/builds/**': [Object] } },
  public: {} }
```
so public is an empty object
In Nuxt 3.11.2 public is `public: { apollo: {}, directus: { rest: [Object], graphql: [Object], auth: [Object] } }`

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/27589

### 📚 Description

As a result of the refactor to runtimeConfig in https://github.com/nuxt/nuxt/pull/27117, it was possible to 'detach' these values if overwriting `nuxt.options.runtimeConfig`.

This PR first prevents overwriting `nitro` option (forcing `Object.assign`) and then also configures `runtimeConfig` and `routeRules` so they will always be kept in sync.

We might consider porting to `untyped`.

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -4,7 +4,7 @@ import ignore from 'ignore'
 import type { LoadNuxtOptions } from '@nuxt/kit'
 import { addBuildPlugin, addComponent, addPlugin, addRouteMiddleware, addServerPlugin, addVitePlugin, addWebpackPlugin, installModule, loadNuxtConfig, logger, nuxtCtx, resolveAlias, resolveFiles, resolveIgnorePatterns, resolvePath, tryResolveModule, useNitro } from '@nuxt/kit'
 import { resolvePath as _resolvePath } from 'mlly'
-import type { Nuxt, NuxtHooks, NuxtModule, NuxtOptions, RuntimeConfig } from 'nuxt/schema'
+import type { Nuxt, NuxtHooks, NuxtModule, NuxtOptions } from 'nuxt/schema'
 import type { PackageJson } from 'pkg-types'
 import { readPackageJSON, resolvePackageJSON } from 'pkg-types'
 import { hash } from 'ohash'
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Nice idea!

(re untyped, i was thinking of portal support since early versions but not sure how repeated this pattern is and honestly I'm not investing much on it nowadays)

toniengelhardt:
Hey @danielroe, you pointed me to this issue from [here](https://github.com/nuxt/nuxt/issues/27616#issuecomment-2168164199), but I still see the errors in `v3.12.2`. 

Should I report this is `nuxt-og-image` ?

Screenshot 2024-06-16 at 14 00 03

danielroe:
@toniengelhardt Would you open an issue with a reproduction on nuxt/nuxt?

toniengelhardt:
> @toniengelhardt Would you open an issue with a reproduction on nuxt/nuxt?

I tried to make a reproduction and realized that the error is actually caused by `unocss/nuxt@0.61.0`. There seems to be a type error where `display: inline-block` is not valid, which breaks satori and as a result the pre-renderer.

I'll open an issue there 🙏🏽

PR #27600, opened 2024-06-13
perf(kit): various performance improvements

danielroe:
### 📚 Description

This splits out changes from https://github.com/nuxt/nuxt/pull/26114 into a single PR, with some further improvements.

### DIFF
--- a/packages/kit/src/template.ts
+++ b/packages/kit/src/template.ts
@@ -11,7 +11,7 @@ import { readPackageJSON } from 'pkg-types'
 import { tryResolveModule } from './internal/esm'
 import { getDirectory } from './module/install'
 import { tryUseNuxt, useNuxt } from './context'
-import { getModulePaths } from './internal/cjs'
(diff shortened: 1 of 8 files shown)


packages/kit/src/template.ts
```diff
@@ -113,18 +113,55 @@ export async function updateTemplates (options?: { filter?: (template: ResolvedN
...
-      .map(m => getDirectory(m.entryPath!)),
-  )
+  const include = new Set<string>([
+    './nuxt.d.ts',
+    join(relativeRootDir, '.config/nuxt.*'),
+    join(relativeRootDir, '**/*'),
+  ])
 
-  const isV4 = nuxt.options.future?.compatibilityVersion === 4
+  if (nuxt.options.srcDir !== nuxt.options.rootDir) {
```
GalacticHypernova, line 125: Just wanted to say good call on this one, I decided to not refsctor these 1 element spread operators yet because I was working on stuff that has more impact as we discusses, and then forgot to fix that after I was done. 👍

PR #27602, opened 2024-06-13
feat(nuxt): allow to disable usage of island payload

### LINKED ISSUE #27503: Add option to always re-render server components without a prop change
### Describe the feature

Currently if I add a server component to one of my pages, it will be rendered only **_once_**, even if I leave that page and visit it again. I see in the documentation, that the component will be re-rendered, if a value of some property on the server component will change, however I wish there was a prop to simply specify for it to re-render every time. Sending arbitary value, like for example `Math.random()` as a prop just for the component to re-render seems silly.

### Additional information

### Final checks

huang-julien:
### 🔗 Linked issue

resolve #27503

### 📚 Description

Hey :wave: this PR allow to disable the island component to use the payload. So every time it is mounted or if it needs to fetch the component, nuxt island will not look for any cached result


hdwebpros:
The `Math.random()` workaround refetches on every render and leaves a new payload entry each time (the cache key hashes props). There's no middle ground between never-refetch and always-refetch.

Proposal: a `maxAge` prop (seconds), matching Nitro's cache.maxAge naming:

```
<LatestReviews :max-age="30" />
```

Within the window, cached HTML is reused. After it elapses, the next prop change or re-mount refetches instead of reusing. undefined = current behavior, unchanged.

Spike tested it: 19 insertions / 2 deletions, 2 files. Stamps `renderedAt` in `setPayload` (same pattern as `prerenderedAt`) and makes the payload short-circuit age-aware. Island suite passes; no new test yet, would add one in a PR. Happy to open one if the shape looks right.

Related: once #36180 lands and islands carry request headers, a frozen personalized island (stale cart count) makes this sharper.

Issue #27627, opened 2024-06-14
Allow configuring vite.cacheDir


jdbdnz:
### Describe the feature

Presently, configuring `vite.cacheDir` ([docs](https://vitejs.dev/config/shared-options.html#cachedir)) is not possible as it is hard coded to `node_modules/.cache/vite` in https://github.com/nuxt/nuxt/blob/main/packages/vite/src/client.ts#L121 and https://github.com/nuxt/nuxt/blob/main/packages/vite/src/server.ts#L74.

I'll be making a PR, but as per the contribution guide, here is an associated issue.

Fore more context about why I would like this:
I am running instances of `nuxt dev` simultaneously in one directory. This is the only configuration I cannot currently make to prevent them from stepping on each others toes with build/cache files.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

PR #27628, opened 2024-06-15
fix(schema,vite): respect `vite.cacheDir` if defined

jdbdnz:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/27627

### 📚 Description

Vite's shared option `cacheDir` is not currently supported by `vite-builder`.  This PR allows setting it in Nuxt config `vite` object.

```
# nuxt.config.js

  ...
  vite: {
    cacheDir: "node_modules/.cache/vite_custom"
  }
}
```

#### 🤨 Why

I am running instances of `nuxt dev` simultaneously in one directory. When they both target the same `vite.cacheDir` there are conflicts and unexpected behaviors ensue (mostly, app doesn't load).

#### 😌 QA

Modified in Playground to confirm it modified Vite's cache directory.

image

### DIFF
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -110,7 +110,7 @@ export default defineUntypedSchema({
       },
     },
     cacheDir: {
-      $resolve: async (val, get) => val ?? (await get('cacheDir')),
+      $resolve: async (val, get) => val ?? resolve(await get('rootDir'), 'node_modules/.cache/vite'),
     },
   },
 })
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -109,6 +109,8 @@ export default defineUntypedSchema({
         },
       },
     },
-    cacheDir: await get('cacheDir'),
+    cacheDir: {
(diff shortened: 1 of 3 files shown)


dash-:
I'd love to see this merged. I could finally get rid of the dirty hacks I'm doing to avoid the same issue.

danielroe:
This seems good to me. Only - I wonder if we could do it automatically. I assume you also have a custom buildDir, right?

packages/schema/src/config/vite.ts
```diff
@@ -109,5 +109,8 @@ export default defineUntypedSchema({
         },
       },
     },
+    cacheDir: {
+      $resolve: async (val, get) => val ?? (await get('cacheDir')),
```
danielroe, line 113: I don't think we have a `nuxt.options.cacheDir` option set, so this should probably be something more like:

```suggestion
      $resolve: async (val, get) => val ?? resolve(await get('rootDir'), 'node_modules/.cache/vite'),
```

dash-:
Great to see this. An overridable default that works automatically when different build directories exist (using hashes of the build directories) also makes sense to me.

PR #27640, opened 2024-06-15
fix(nuxt): add missing script stubs

huang-julien:
### 📚 Description
This PR adds a missing stub for nuxt scripts


danielroe (approved):
nice spot!

Issue #11858, opened 2021-10-02
ESM improvements


pi0:
Following up nuxt/framework#539, some tasks remaining to (almost) remove dependency on CommonJS:

- [x] Avoid `createModule` as much as possible
- [x] Remove usage of `__dirname`/`__filename` and `require` 
  - [x] Docs and examples (#4381)
- [x] Prefix and resolve built-in imports with node (#4384)
- [x] Start deprecating CJS utils from kit 
- [x] Use `mlly.require.resolve` instead of `require.resolve` to apply import conditions
- [x] Upgrade estree-walker (needs ESM resolve from jiti w/loader for stub)

Upstream:
- [ ] Jiti ESM loader for stub matching final package

PR #27642, opened 2024-06-15
fix(nuxt): reduce usage of cjs utilities

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/11858

### 📚 Description

This:

* removes last use of `require.resolve` outside the deprecated kit cjs utils
* remove usage of __dirname and __filename in favour of directly using `import.meta.url`
* adds a note to remove `requireModule` when loading postcss plugins in vite + webpack (in a subsequent PR)

### DIFF
--- a/packages/nuxt/src/core/plugins/import-protection.ts
+++ b/packages/nuxt/src/core/plugins/import-protection.ts
@@ -1,4 +1,3 @@
-import { createRequire } from 'node:module'
 import { createUnplugin } from 'unplugin'
 import { logger, useNuxt } from '@nuxt/kit'
 import { resolvePath } from 'mlly'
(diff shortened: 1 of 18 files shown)


packages/nuxt/src/core/plugins/import-protection.ts
```diff
@@ -85,7 +88,7 @@ export const ImportProtectionPlugin = createUnplugin(function (options: ImportPr
         matched = true
       }
       if (matched) {
-        return _require.resolve('unenv/runtime/mock/proxy')
+        return proxy
```
DamianGlowala, line 91:
```suggestion
        return resolvePath('unenv/runtime/mock/proxy', { url: useNuxt().options.modulesDir })
```

packages/nuxt/src/core/plugins/import-protection.ts
```diff
@@ -58,6 +59,8 @@ export const nuxtImportProtections = (nuxt: { options: NuxtOptions }, options: {
 export const ImportProtectionPlugin = createUnplugin(function (options: ImportProtectionOptions) {
   const cache: Record<string, Map<string | RegExp, boolean>> = {}
   const importersToExclude = options?.exclude || []
+  const nuxt = useNuxt()
+  const proxy = resolvePath('unenv/runtime/mock/proxy', { url: nuxt.options.modulesDir })
```
DamianGlowala, line 63:
```suggestion
```

packages/nuxt/src/core/plugins/import-protection.ts
danielroe: I thought it might be more performant to do this once to rather than every time the id is resolved...

PR #27644, opened 2024-06-15
feat(nuxt): await custom `routes` function in `router.options`

### LINKED ISSUE #23678: Asynchronously adding routes at runtime doesn't work reliably
### Environment

Applies to any environment. Reproduction uses Nuxt 3.7.4.

### Reproduction

https://stackblitz.com/edit/github-wyjpmc?file=app.vue

## Setup
- Two pages exist: `pages/index.vue` with name `home` and `pages/foobar/index.vue` with name `foobar`.
- The plugin `plugins/alterRoutes.ts` adds an additional route using `router.addRoute()` with name `foobar-dynamic` and path `/foobar-dynamic`. The path is loaded async (e.g. from an API call)
- In `app.vue` a computed property returning the route name exists
- In `app.vue` a watcher on the computed route name logs the name to the console
- In `app.vue` the computed route name is logged to the console directly in setup()

## Reproduce bug
- Visit the front page `home`. Refresh. The name of the route is logged correctly to the console.
- Now click on `foobar-dynamic`. The name is logged correctly to the console.
- Now refresh the page.
- The name of the route logged to the console is `undefined`.

### Describe the bug

Only the initially available routes generated at build time are used for the inital router state. The reactive route from `useRoute()` does not contain the dynamically added route during setup.

The problem can be fixed with a workaround by calling router.replace() after adding the dynamic routes:

```typescript
router.addRoute({
  // ...
})
await router.replace({
  ...router.resolve(route.path),
  force: true
})
```

### Additional context

The closed issue #12823 seems related and also contains a few comments about adding/altering routes async. There is also #13467 which is open, but it doesn't specifically mention an async way of adding routes at runtime. There's also at least one open discussion about it.

Looking into it a bit this could be solved by allowing us to use an async method for the `routes` option in `app/router.options.ts`. That way the initial state of the router would already be correct and wouldn't require any additional plugins.

https://github.com/nuxt/nuxt/blob/5c7cca170a85b0584b4091e7e7e461b4800c0e36/packages/nuxt/src/pages/runtime/plugins/router.ts#L64C62-L64C62

This would be much more reliable, because it would also make sure that any other plugins (from third party modules or locally) won't have a stale/unmatched route. With the router.replace() workaround that has to be done in a custom plugin we additionally have to make sure that the order of the plugins is correct, which is another challenge in itself.

Alternatively a hook to alter the routes would also work.

A fix for this would really improve DX for a lot of us who build "headless CMS" sites where route paths can't possibly be known at build time and must be loaded at runtime, because they can be freely edited by content editors or are dynamically generated based on page titles (and therefore can change any time).

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/23678

### 📚 Description

This allows awaiting the `routes()` function from `router.options.ts` to allow routes to be added before the router initialises.

### DIFF
--- a/packages/nuxt/src/pages/build.d.ts
+++ b/packages/nuxt/src/pages/build.d.ts
@@ -1,5 +1,6 @@
 declare module '#build/router.options' {
   import type { RouterOptions } from '@nuxt/schema'
+
   const _default: RouterOptions
   export default _default
 }
--- a/packages/nuxt/src/pages/build.d.ts
+++ b/packages/nuxt/src/pages/build.d.ts
@@ -1,0 +1,5 @@
+declare module '#build/router.options' {
+  import type { RouterOptions } from '@nuxt/schema'
+  const _default: RouterOptions
+  export default _default
+}
--- a/packages/nuxt/src/pages/runtime/plugins/prerender.server.ts
+++ b/packages/nuxt/src/pages/runtime/plugins/prerender.server.ts
@@ -5,7 +5,6 @@ import { defineNuxtPlugin } from '#app/nuxt'
 import { prerenderRoutes } from '#app/composables/ssr'
 // @ts-expect-error virtual file
 import _routes from '#build/routes'
-// @ts-expect-error virtual file
 import routerOptions from '#build/router.options'
 
 let routes: string[]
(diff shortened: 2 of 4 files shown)


SmootherSpike:
Hi, is this also planned for 3.12.5, since this functionality existed in the old @nuxtjs/router module for nuxt2, so this is rather a bugfix than a feature?
This would mean not beeing forced to migrate/wait to v4

Thanks for getting back here @danielroe

danielroe:
@SmootherSpike It's not a bug fix as this has never existed in v3.

However, yes, you can expect to see it in v3.13.

Issue #26435, opened 2024-03-21
context.outputFileSystem.lstat is not a function


mark980828:
### Environment

System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    nuxt: ^2.17.3 => 2.17.3

### Reproduction

https://stackblitz.com/edit/nuxt-starter-vi9px5?file=pages%2Findex.vue

### Describe the bug

This issue is like a version issue with "webpack-dev-middleware".

### Logs

```shell-script
ERROR  context.outputFileSystem.lstat is not a function

  at size (node_modules/webpack-dev-middleware/dist/middleware.js:157:36)
  at new Promise (<anonymous>)
  at processRequest (node_modules/webpack-dev-middleware/dist/middleware.js:155:28)
  at ready (node_modules/webpack-dev-middleware/dist/utils/ready.js:17:5)
  at Object.middleware (node_modules/webpack-dev-middleware/dist/middleware.js:69:5)
  at eval (node_modules/pify/index.js:33:11)
  at new Promise (<anonymous>)
  at Object.eval (node_modules/pify/index.js:6:9)
  at Object.apply (node_modules/pify/index.js:80:20)
  at WebpackBundler.middleware (node_modules/@nuxt/webpack/dist/webpack.js:1917:37)
```

danielroe:
Likely latest patch in webpack-dev-middleware causes this - you can try pinning previous version.

https://github.com/nuxt/nuxt/pull/26427 does highlight some type changes with our implementation of mfs and we likely need to update.

mark980828:
> Likely latest patch in webpack-dev-middleware causes this - you can try pinning previous version.

Yes, it can run now.
"resolutions": {
    "webpack-dev-middleware": "5.0.0"
  }

bhrutledge:
Duplicate of https://github.com/nuxt/nuxt/issues/20564?

mark980828:
> Duplicate of #20564?

yeah

manniL:
[...]

PR #21804, opened 2023-06-26
fix(webpack): pin `webpack-dev-middleware` to 5.0.0

danielroe:
### 🔗 Linked issue

resolves #20564

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

See https://github.com/nuxt/nuxt/issues/20564#issuecomment-1536201468. This PR pins `webpack-dev-middleware` to 5.0.0.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #20564, opened 2023-04-28
Webpack error on video tag display with require on src


JordanQui:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v19.7.0
- Nuxt Version:     2.16.3
- Nitro Version:    -
- Package Manager:  npm@9.5.0
- Builder:          webpack
- User Config:      head, css, plugins, components, buildModules, modules, devMiddleware, build
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

load a video tag in a page with a video loaded with require on src

### Describe the bug

video don't load and a message error occur 

### Logs

```shell-script
ERROR  context.outputFileSystem.lstat is not a function

  at size (node_modules/webpack-dev-middleware/dist/middleware.js:157:36)
  at new Promise (<anonymous>)
  at processRequest (node_modules/webpack-dev-middleware/dist/middleware.js:155:28)
  at ready (node_modules/webpack-dev-middleware/dist/utils/ready.js:17:5)
  at Object.middleware (node_modules/webpack-dev-middleware/dist/middleware.js:69:5)
  at node_modules/pify/index.js:33:11
  at new Promise (<anonymous>)
  at Object.<anonymous> (node_modules/pify/index.js:6:9)
  at Object.apply (node_modules/pify/index.js:80:20)
  at WebpackBundler.middleware (node_modules/@nuxt/webpack/dist/webpack.js:1863:37)
  at node_modules/@nuxt/server/dist/server.js:495:14
  at call (node_modules/connect/index.js:239:7)
  at next (node_modules/connect/index.js:183:5)
  at SendStream.error (node_modules/serve-static/index.js:121:7)
  at SendStream.emit (node:events:512:28)
  at SendStream.error (node_modules/send/index.js:270:17)
  at SendStream.onStatError (node_modules/send/index.js:417:12)
  at onstat (node_modules/send/index.js:722:26)
  at FSReqCallback.oncomplete (node:fs:191:21)
```

danielroe:
Would you provide a [reproduction](https://nuxt.com/docs/community/reporting-bugs/#create-a-minimal-reproduction)? 🙏

JordanQui:
my mistake it's a nuxt 2.16 issue

JordanQui:
a clean instal and a try to load a video in a tag with require in src is sufficient to reproduce on my side

ClydeFroga:
I have the same problem

ian-bitts:
[...]

PR #27652, opened 2024-06-16
feat(webpack): migrate to `memfs`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/26435
related https://github.com/nuxt/nuxt/pull/21804, https://github.com/nuxt/nuxt/issues/20564

### 📚 Description

This migrates from 'memory-fs' to 'memfs'.

This is necessary because of a breaking change with 'webpack-dev-middleware' in 5.0.0+. But sticking to 5.0.0 would mean that we would have a known security vulnerability with path traversal: https://github.com/advisories/GHSA-wr3j-pwj9-hqq6.

### DIFF
--- a/packages/webpack/src/utils/async-mfs.js
+++ b/packages/webpack/src/utils/async-mfs.js
@@ -1,24 +1,54 @@
-import MFS from 'memory-fs'
-export default class AsyncMFS extends MFS {}
+import path from 'path'
+import mkdirp from 'mkdirp'
+import { createFsFromVolume, Volume } from 'memfs'
+import { dirname } from 'upath'
 
 const syncRegex = /Sync$/
 
-const propsToPromisify = Object.getOwnPropertyNames(MFS.prototype).filter(n => syncRegex.test(n))
+export default function createFS () {
+  const volume = new Volume()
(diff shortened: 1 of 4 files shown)


packages/webpack/src/utils/async-mfs.js
```diff
@@ -1,24 +1,54 @@
...
+  const writeFileSync = fs.writeFileSync
+
+  function ensureDirSync (...args) {
+    if (typeof args[0] === 'string') {
+      const dir = dirname(args[0])
+      fs.mkdirSync(dir, { recursive: true })
     }
+  }
+
+  fs.writeFileSync = function (...args) {
```
pi0, line 26: Do we need a permanent patch to `fs.writeFileSync`?

pi0 (approved):
~This is ancient history for me to comment haha, i would recommand to fork memfs or even inline it to avoid fs workaround~

packages/webpack/src/utils/async-mfs.js
danielroe: (This is only a patch to the `memfs.writeFileSync`. We are already patching all async props too.)

PR #27689, opened 2024-06-18
feat(nuxt): generate basic jsdoc for module config entry

antfu:
Before:

Screenshot 2024-06-18 at 14 06 30

After: 

Screenshot 2024-06-18 at 14 06 02

The duplication of tags are because we are extending both `@nuxt/schema` and `nuxt/schema` - I guess it won't hurt much, we could figure out it later, or we only generate jsdoc for one of them?


Atinux:
I like the `description` and `documentation` fields 👍

danielroe (approved):

danielroe:
Maybe we could add the jsdoc tags to only one of the augmented modules?

packages/nuxt/src/core/templates.ts
```diff
@@ -157,36 +157,44 @@ const adHocModules = ['router', 'pages', 'imports', 'meta', 'components', 'nuxt-
...
     }
     const moduleOptionsInterface = [
-      ...modules.map(([configKey, importName]) =>
+      ...modules.flatMap(([configKey, importName, mod]) => {
+        return [
+        `    /**`,
+        `     * Configuration for \`${importName}\``,
+        mod.meta?.rawPath
+          ? undefined
+          : `     * @see https://www.npmjs.com/package/${importName}`,
```
pi0, line 180: IMO this should be properly inferred from `reponsitory` field or module meta if we are going to generate docs. Many misleading cases can happen by assuming import name is always from npm registry...

antfu: Yeah. That's why I am bringing up the question above. I see currently those fields are "agnositic" to Nuxt and module authors - not everyone merging their package.json into the meta field, and it could probably be namespace collisions if we always ask module authors to do so. Have a few more explicit field might make this behaviour more clear IMO

antfu:
It should be good now; I assume `mod.meta` could be the data structured as `package.json`

danielroe (approved):
Lovely - thank you! ❤️ 

It's possible there are still some edge cases (like alternative registries) but we can always iterate if needed.

PR #27702, opened 2024-06-19
feat(kit,nuxt,schema,vite,webpack): migrate to nitropack v3 nightly

danielroe:
### 📚 Description

This migrates to a particular version of nitro v3 and h3 v2 nightly releases, for further testing.

- [x] https://github.com/unjs/nitro/pull/2546
- [x] consider using `nitro` rather than `nitropack` (at least internally)
- [x] reimplement nitro appConfig using vfs
- [x] resolve type issues (upstream?)
- [x] consider augmenting both `nitro/types` and `nitropack/types`

More significant breaking changes are coming in `h3`, along with the following further todos

- [ ] update Nuxt upgrade guide for nitro-related changes
- [ ] update nuxt/test-utils in a forward/backward compatible way
- [ ] remove internal dependencies on `event.node`
- [ ] remove db0 in Nitro upstream
- [ ] refactor runtimeConfig inside Nuxt


packages/schema/src/config/typescript.ts
```diff
@@ -34,7 +34,7 @@ export default defineUntypedSchema({
       $resolve: (val) => {
         const defaults = [
           // Nitro auto-imported/augmented dependencies
-          'nitropack',
+          // TODO: add 'nitro/types'
```
danielroe, line 37: need to add subpath export support for typescript.hoist

packages/nuxt/src/core/runtime/nitro/app-config.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (!event) {
+    return _sharedAppConfig
+  }
+  // Reuse cached app config from event context
+  if (event.context.nitro.appConfig) {
+    return event.context.nitro.appConfig
+  }
+  // Prepare app config for event context
+  const appConfig = klona(_inlineAppConfig)
+  event.context.nitro.appConfig = appConfig
```
pi0, line 20: Even for polyfill/backward compatibility, filling nitro context in event would be misleading we should store it in context.nuxt

packages/vite/src/server.ts
```diff
@@ -104,7 +104,7 @@ export async function buildServer (ctx: ViteBuildContext) {
   } satisfies vite.InlineConfig, ctx.nuxt.options.vite.$server || {}))
 
   if (!ctx.nuxt.options.dev) {
-    const nitroDependencies = await tryResolveModule('nitropack/package.json', ctx.nuxt.options.modulesDir)
+    const nitroDependencies = await tryResolveModule('nitro/package.json', ctx.nuxt.options.modulesDir)
```
pi0, line 107: This is available now via `nitro/runtime/meta`

packages/webpack/src/presets/base.ts
```diff
@@ -125,7 +125,7 @@ function baseAlias (ctx: WebpackConfigContext) {
     ...ctx.alias,
   }
   if (ctx.isClient) {
-    ctx.alias['#internal/nitro'] = resolve(ctx.nuxt.options.buildDir, 'nitro.client.mjs')
+    ctx.alias['nitro/runtime'] = resolve(ctx.nuxt.options.buildDir, 'nitro.client.mjs')
```
pi0, line 128: Just checking it was because of runtime config access right?

test/bundle.test.ts
```diff
@@ -86,6 +87,7 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
         "@unhead/dom",
         "@unhead/shared",
         "@unhead/ssr",
+        "db0",
```
pi0, line 90: It shouldn't be added by default in the prod bundle seems strange!

packages/nuxt/src/core/plugins/import-protection.ts
```diff
@@ -85,6 +85,7 @@ export const ImportProtectionPlugin = createUnplugin(function (options: ImportPr
         matched = true
       }
       if (matched) {
+        console.log(id, cache[id])
```
pi0, line 88: (i guess leftover?)

```suggestion
```

packages/webpack/src/presets/base.ts
danielroe: Exactly. This is so we can use isomorphic code that gets replaced with different runtime config code on client.

pi0: Okay! I have an idea to get rid of it!

packages/nuxt/src/core/plugins/import-protection.ts
danielroe: Thank you 🙏

packages/webpack/src/presets/base.ts
danielroe: 🙌

PR #27745, opened 2024-06-20
fix(nuxt)!: remove `__NUXT__` after hydration

### LINKED ISSUE #25336: Remove `window.__NUXT__` once app initialises
### Describe the feature

Apparently it only exists for backwards-compat according to code comments:

https://github.com/nuxt/nuxt/blob/648ef069931336b8ee0c0e48fef3ee700ebc3dbe/packages/nuxt/src/app/plugins/revive-payload.client.ts#L52-L53

Nuxt v4 might be a good point to remove it if that's the case.

PR: #19205

### Additional information

### Final checks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25336

### 📚 Description

This removes `window.__NUXT__` after app initialises. The payload can still be accessed with `useNuxtApp().payload`

### 👉 Migration

```diff
- console.log(window.__NUXT__)
+ console.log(useNuxtApp().payload)
```


packages/nuxt/src/app/plugins/revive-payload.client.ts
```diff
@@ -49,7 +49,6 @@ export default defineNuxtPlugin({
       definePayloadReviver(reviver, revivers[reviver as keyof typeof revivers])
     }
     Object.assign(nuxtApp.payload, await nuxtApp.runWithContext(getNuxtClientPayload))
-    // For backwards compatibility - TODO: remove later
-    window.__NUXT__ = nuxtApp.payload
+    delete window.__NUXT__
```
manniL, line 52: I wonder: Is there a specific reason why we wanted to remove it on the first place? 🤔

TheAlexLichter (approved):

atinux (approved):

packages/nuxt/src/app/plugins/revive-payload.client.ts
ivomarsan: @manniL I guess to hide the explicit bucket where has useful data. In my case, I pass token for integrations and there is no problem to be explicit, but window.__NUXT__ is very chamative

ivomarsan (approved):

packages/nuxt/src/app/plugins/revive-payload.client.ts
danielroe: to be clear, this is not about security - it is essential to treat all client side code as non privileged and fully accessible by end user.

In this case `window.useNuxtApp().payload` will return the _same information_.

PR #27768, opened 2024-06-21
docs: upgrade guide, remind people to upgrade third-party configs like tailwind.config.js

cheike569:
https://nuxt.com/docs/getting-started/upgrade

It should be noted to upgrade any third-party libraries that depend on the directory structure in the "_New Directory Structure_" Section. This may be Tailwind CSS in many circumstances.

### 📚 Description

The Nuxt 4 Upgrade Guide includes a significant change, which is the new Directory Structure.

Since many projects depend on third-party libraries and processors, a reminder to upgrade these configs in your project as well might be beneficial, namely `tailwind.config.js`


danielroe:
cc: @ineshbose - do you think we need to do anything else here?

ineshbose:
> cc: @ineshbose - do you think we need to do anything else here?

Thanks for pinging!

I tend to discourage users from specifying Nuxt directory paths within Tailwind config as the module already adds in their resolved paths, and if nuxt.options.dir will also consider the appDir automatically, the module should still work. Maybe the doc can remind them to review any content paths with an alert saying that `@nuxtjs/tailwindcss` automatically resolves the required paths for Nuxt directories.

danielroe (approved):

PR #27800, opened 2024-06-24
fix(nuxt): avoid name collisions with user-crafted lazy components

GalacticHypernova:
### 📚 Description

This PR removes name conflicts that happen with lazy prefixed components. If a developer were to create a component, (like \<LazyTest\>), using it in a template causes the "failed to resolve component" warning to show. That's because the current component loading strategy does not take into account components that could be prefixed with Lazy


danielroe (commented):
I'm not sure we want users to create components that have `Lazy` in the name, as I think it would add a lot of confusion.

Is there a reason you wanted to support this?

GalacticHypernova:
It's not a feature per se, we can add a warning not to do it, but it does take care of edge cases if such components are created, even accidentally.

danielroe:
I think instead we should just print a warning if a user provides any component that collides, if we don't already. It would be in the `scan` file rather than in the loader.

GalacticHypernova:
Sure thing!
I'll rework it :+1:

PR #27803, opened 2024-06-24
perf(kit,nuxt): use `fdir` to speed up fs scanning

danielroe:
### 📚 Description

This aims to provide faster glob performance when resolving files with Nuxt.

Discovered @antfu's thoughts in https://github.com/unocss/unocss/pull/3722#issuecomment-2053769072 too late, so would value his review. 🙏 

We might choose to revert the adoption of `fdir` here and later go with native `fs.glob`, keeping just the move to native `readdir`...


antfu:
I also continued the discussion with @43081j in e12e's discord - my main worry about this is the cost of API changes and implicit behavior changes. My suggestion would be we find a `fs.glob` shims with the exact API for Node.js prior that version. This way it would be more like the ultimate state (zero-deps, standardized) that the ecosystem would likely to agree to move to. And at one point when we drop the Node version that without the API, we can then remove the shims without migrating again

43081j:
just my two cents: only the API style will differ here. behaviour is likely the same if setup correctly

for that reason though, i'd likely still opt for `fs.glob` if you're able to (but its pretty new so doubt we can require that version yet). not sure how likely a shim/polyfill is

PR #27819, opened 2024-06-25
refactor(nuxt): remove _registeredComponents from ssrContext

huang-julien:
### 📚 Description
This PR remove the _registeredComponents getter and setter.


danielroe (approved):
I think this was probably backwards compatibility. Seems vue-loader doesn't use this any more: https://github.com/vuejs/vue-loader/issues/1887.

And `vue-bundle-renderer` prefers modules if it exists:

https://github.com/nuxt-contrib/vue-bundle-renderer/blob/0620829c4f4998939f5edb3f3423d4d644e15220/src/runtime.ts#L182

So I think it's safe to merge.

PR #27822, opened 2024-06-25
fix(nuxt): use URL to encode redirected URLs

### LINKED ISSUE #27554: header.location in Redirects is Double-Encoded in Nuxt v3.12.1
### Environment

- Operating System: Darwin
- Node Version:     v21.7.3
- Nuxt Version:     3.12.1
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.1.4
- Builder:          -
- User Config:      app, srcDir, devtools, modules, i18n, css, pinia, imports, nitro, runtimeConfig, vue, vite, build
- Runtime Modules:  @pinia/nuxt@0.5.1, @vant/nuxt@1.0.4, @nuxtjs/i18n@8.3.1, nuxt-swiper@1.2.2
- Build Modules:    -

### Reproduction

1. Set up a Nuxt project with the latest version.
2. Implement a redirection in your server middleware or a page, where the redirection URL contains another URL as a query parameter. For example:
```javascript
navigateTo('https://example.com/?redirect='+encodeURIComponent('https://anotherexample.com/?param=value'), {
	external: true,
});
```
3. Observe the redirection behavior in the browser or using a tool like curl.

### Describe the bug

In the latest version of Nuxt, it appears that the header.location in the redirects is being encoded using encodeURI. This causes issues when the redirection URL contains another URL as a query parameter. The nested URL gets encoded multiple times, resulting in a failed redirection.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/27554

### 📚 Description

This uses URL to ensure characters are ASCII-encoded as required (as well as perform other normalisation steps), without double encoding other segments of the URL.

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -266,7 +266,7 @@ export function resolveRouteObject (to: Exclude<RouteLocationRaw, string>) {
 /**
  * @internal
  */
-export function encodeURL(location: string, isExternalHost = false) {
+export function encodeURL (location: string, isExternalHost = false) {
   const url = new URL(location, 'http://localhost')
   if (!isExternalHost) {
     return url.toString().replace(/^http:\/\/localhost\//, '/')
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -269,7 +269,7 @@ export function resolveRouteObject (to: Exclude<RouteLocationRaw, string>) {
 export function encodeURL (location: string, isExternalHost = false) {
   const url = new URL(location, 'http://localhost')
   if (!isExternalHost) {
-    return url.toString().replace(/^http:\/\/localhost\//, '/')
+    return url.pathname + url.search + url.hash
   }
   if (location.startsWith('//')) {
     return url.toString().replace(url.protocol, '')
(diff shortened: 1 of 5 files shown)


packages/nuxt/src/app/composables/router.ts
```diff
@@ -259,3 +262,17 @@ export const setPageLayout = (layout: unknown extends PageMeta['layout'] ? strin
 export function resolveRouteObject (to: Exclude<RouteLocationRaw, string>) {
   return withQuery(to.path || '', to.query || {}) + (to.hash || '')
 }
+
+/**
+ * @internal
+ */
+export function encodeURL (location: string, isExternalHost = false) {
+  const url = new URL(location, 'http://localhost')
+  if (!isExternalHost) {
```
huang-julien, line 271: not with import.meta.dev ?

packages/nuxt/src/app/composables/router.ts
```diff
@@ -259,3 +262,17 @@ export const setPageLayout = (layout: unknown extends PageMeta['layout'] ? strin
...
+    return url.toString().replace(/^http:\/\/localhost\//, '/')
```
huang-julien, line 272: instead of using `replace`, is it more performant to construct the string with URL properties ?

danielroe: Like:
```suggestion
    return url.pathname + url.search + url.hash
```

huang-julien: like that yes :+1:

pi0 (approved):

PR #27823, opened 2024-06-25
fix(nuxt): store augmented pages with route path

### LINKED ISSUE #27814: `layout: false` does not remove the layout when multiple routes use the same page component
### Environment

- Operating System: Darwin
- Node Version:     v22.1.0
- Nuxt Version:     3.12.0
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools, runtimeConfig, vite, alias, modules, image, hooks, nitro, serverMiddleware, build, css
- Runtime Modules:  @pinia/nuxt@0.5.1, @tixel/ui-nuxt@0.0.7, @tixel/component-library@0.4.0, nuxt-jsonld@2.0.8, @pinia-plugin-persistedstate/nuxt@1.2.0, @nuxt/devtools@1.3.6, @nuxt/image@1.0.0-rc.3, @vueuse/nuxt@10.11.0, @nuxtjs/device@3.1.1, @nuxt/content@2.12.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-m1cmbw?file=pages%2Fexample-page.vue

### Describe the bug

On the reproduction URL

Go to `/example-page`

You will see that there is no default layout set, because the page component has:

```
definePageMeta({
  layout: false,
});
```

However, now go to `/example-page-2` and you will see that this page still includes the layout, despite using the same component as the other route.

Both pages should not render the layout.

See the routes are rendered in `routes.ts`

### Additional context

This only occurs in `v3.12.x`, rolling back to `v3.11.x` fixes the issue.

### Logs

_No response_

TheAlexLichter:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/27814

### 📚 Description

Before, when augmenting pages via scanPageMeta, we cached the already-augmented pages based on the file path. When to different routes used the same file, this would lead to not augmenting the second route. This PR caches them by using file path and route path as key

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -140,19 +140,19 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
   return prepareRoutes(routes)
 }
 
-export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPages = new Set<string>()) {
+export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPagesWithPaths = new Set<string>()) {
   for (const route of routes) {
-    if (route.file && !augmentedPages.has(route.file)) {
+    if (route.file && !augmentedPagesWithPaths.has(route.file)) {
       const fileContent = route.file in vfs ? vfs[route.file] : fs.readFileSync(await resolvePath(route.file), 'utf-8')
       Object.assign(route, await getRouteMeta(fileContent, route.file))
-      augmentedPages.add(route.file)
+      augmentedPagesWithPaths.add(`${route.file}_${route.path}`)
     }
 
     if (route.children && route.children.length > 0) {
       await augmentPages(route.children, vfs)
     }
   }
-  return augmentedPages
+  return augmentedPagesWithPaths
 }
 
 const SFC_SCRIPT_RE = /<script(?<attrs>[^>]*)>(?<content>[\s\S]*?)<\/script[^>]*>/i
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -142,10 +142,11 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
 
 export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPagesWithPaths = new Set<string>(), parentPath = '') {
   for (const route of routes) {
-    if (route.file && !augmentedPagesWithPaths.has(route.file)) {
+    const key = `${route.file}_${parentPath}${route.path}`
+    if (route.file && !augmentedPagesWithPaths.has(key)) {
       const fileContent = route.file in vfs ? vfs[route.file] : fs.readFileSync(await resolvePath(route.file), 'utf-8')
       Object.assign(route, await getRouteMeta(fileContent, route.file))
-      augmentedPagesWithPaths.add(`${route.file}_${parentPath}${route.path}`)
+      augmentedPagesWithPaths.add(key)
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/utils.ts
```diff
@@ -140,19 +140,19 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
...
 
-export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPages = new Set<string>()) {
+export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPagesWithPaths = new Set<string>()) {
   for (const route of routes) {
-    if (route.file && !augmentedPages.has(route.file)) {
+    if (route.file && !augmentedPagesWithPaths.has(route.file)) {
       const fileContent = route.file in vfs ? vfs[route.file] : fs.readFileSync(await resolvePath(route.file), 'utf-8')
       Object.assign(route, await getRouteMeta(fileContent, route.file))
-      augmentedPages.add(route.file)
+      augmentedPagesWithPaths.add(`${route.file}_${route.path}`)
```
danielroe, line 148: if we are doing this, we should probably build the path from parents as well, as a child path might otherwise cause a false positive match.

e.g. parent with path `/test` and a child with path `foo` would cause a match with parent with path `/other` and a child with path `foo`, if they were the same filename....

manniL: Good point

manniL: Added a `parentPath` param for recursive calls

danielroe (approved):

danielroe:
One complex issue here is how we should support users overriding config set in `definePageMeta` in `pages:extends`. At the moment this is possible fully. With this change, it no longer becomes possible if you modify the page path at all, meaning it is no longer possible to override the path a page sets in its `definePageMeta`. This might be an issue with `@nuxtjs/i18n` for example. (cc: @BobbieGoede)

(Previously it was possible by checking what the scanned page meta was for a page before adding it as a route somewhere else.)

BobbieGoede:
>With this change, it no longer becomes possible if you modify the page path at all, meaning it is no longer possible to override the path a page sets in its definePageMeta. This might be an issue with @nuxtjs/i18n for example. 

Yes, if I understand correctly, this would be a problem 😅 Would have to see if i18n tests fail with this change, I find it a bit hard to mentally step through the order of these hooks vs those in the module rn 🤔

packages/nuxt/src/pages/utils.ts
```diff
@@ -140,19 +140,19 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
   return prepareRoutes(routes)
 }
 
-export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPages = new Set<string>()) {
+export async function augmentPages (routes: NuxtPage[], vfs: Record<string, string>, augmentedPagesWithPaths = new Set<string>(), parentPath = '') {
   for (const route of routes) {
-    if (route.file && !augmentedPages.has(route.file)) {
+    if (route.file && !augmentedPagesWithPaths.has(route.file)) {
```
BobbieGoede, line 145: We're not checking if `augmentedPagesWithPaths` has `${route.file}_${parentPath}${route.path}` here, does this mean the route file will always be read?

manniL: Oh, my bad, yes. You are right. we should check that

danielroe:
I think we can close this in favour of the new v4 behaviour which is to scan for metadata after calling pages:extend, with pages:resolved following for users to amend/inspect the data.

PR #27838, opened 2024-06-26
feat(nuxt): warn when components have a name conflict

GalacticHypernova:
### 🔗 Linked issue

closes #27800

### 📚 Description

This PR warns the user when a component is found with name conflicts that use the reserved Lazy prefix


danielroe (commented):
Instead, if you update L96 as follows:

```diff
- const pascalName = pascalCase(componentNameSegments)
+ const pascalName = pascalCase(componentNameSegments).replace(/^Lazy(?=[A-Z])/, '')
```

then this is what will be logged:

```
 WARN  Two component files resolving to the same name Bob:                                                     11:49:28 AM

 - /Users/daniel/code/nuxt/nuxt/playground/components/LazyBob.vue
 - /Users/daniel/code/nuxt/nuxt/playground/components/Bob.vue
```

That feels like a good approach, and it's an even more minimal change.

GalacticHypernova:
How about files that don't have a a non-lazy counterpart?

danielroe:
Hm - yes, it's probably worth over-warning, but it's most helpful when we can point to the component it would collide with.

either way, it shouldn't just be about the component file name but the resolved component name. So `MyDirLazyComponent` would be fine (`~/components/my-dir/lazy-component.vue`). And `lazy-component.vue` wouldn't be, even though it doesn't start with `Lazy`. And `lazyn-component.vue` would be fine.

GalacticHypernova:
Ah, I see. Yea, that is true, the warning will also need to be modified for it

packages/nuxt/src/components/scan.ts
```diff
@@ -95,7 +95,11 @@ export async function scanComponents (dirs: ComponentsDir[], srcDir: string): Pr
       const suffix = (mode !== 'all' ? `-${mode}` : '')
       const componentNameSegments = resolveComponentNameSegments(fileName.replace(/["']/g, ''), prefixParts)
       const pascalName = pascalCase(componentNameSegments)
-
+      
+      if (pascalName.startsWith('Lazy')) {
```
danielroe, line 99: we should use a regexp that ensures it is a complete chunk. e.g. `/^Lazy(?=[A-Z])/`

GalacticHypernova: We can do that of course, but I can't seem to think of any examples of components that won't get correctly captured without the additional positive lookahead. Do you think it's necessary?

danielroe: I gave some examples here: https://github.com/nuxt/nuxt/pull/27838#issuecomment-2191302610.

GalacticHypernova: Yes, but would they not capture correctly with the beginning check? It already checks that it starts with Lazy regardless of what it has afterwards, as that's also the pattern in loader.ts

danielroe: I realise it's not very likely, but we should only warn if the first _segment_ is lazy, so `lazyn-component.vue` should not trigger this warning.

GalacticHypernova: I've looked at the loader pattern again, it just looks for Lazy and then more characters, so a component named Lazyn-component.vue would still capture the Lazy part and read it as n-component.vue, wouldn't it? That means it would still be modified because it has that name conflict, unless I'm missing something?

danielroe: It would capture it with the regexp but the check on the next line would fail as no component is found. So it is safe.

GalacticHypernova: I see, so what is the difference in scan.ts? If it begins with Lazy it means it will be modified in loader.ts which would resolve to no component. 

And sorry for all the questions, I'm just trying to understand, I don't mean for it to sound dismissive or anything

GalacticHypernova: Just checked:
https://stackblitz.com/edit/github-5spadh?file=components%2FLazyn.vue,app.vue

Even if a component only begins with Lazy it fails to resolve, meaning that chunk regex would actually not capture it and won't display the warning
image

danielroe: Good catch! We should update the regexp in the loader too then

GalacticHypernova: I mean, as long as the component doesn't begin with Lazy (at all, no matter how many chunks it consists of), it would not cause any issues, that's why it only errors when the component contains Lazy. The question is, does the regex need to be updated, or is it more so about enforcing that components will not begin with "lazy" which can be documented through the warning? Because by changing the loader regex, it's indirectly allowing components to contain lazy without it breaking the app, like I did in my other PR about this. There's also edge cases like LazyN.vue which would conform to the pattern but still break because it still has the Lazy part

danielroe: Yes, it's important to update the regexp; it fixes the bug you found with loading components like `LazynessThing`.

danielroe (approved):

packages/nuxt/src/components/scan.ts
GalacticHypernova: @danielroe I just noticed a possible regression with the regex:
What if the component is used like [a-z]*- as opposed to [A-Z][a-z]*?
(For example, how you can use lazy-something and LazySomething)

Issue #27003, opened 2024-04-30
more granular control of link prefetching


danielroe:
- ability to prefetch on hover (or other event)
- ability to reactively enable prefetching (by setting prop `prefetch` to true later on)
- expose composable to trigger a prefetch/preload of a page (`prefetchPage('/')`)

OrbisK:
@danielroe isnt this mostly solved?

danielroe:
nuxtathon closed @OrbisK

PR #27846, opened 2024-06-26
feat(nuxt): allow defining triggers for prefetching links

kewinzaq1:
### 🔗 Linked issue
more granular control of link prefetching #27003

### 📚 Description
Currently prefetching is not flexible. The default behavior prefetches page as soon as it is visible within the viewport.
The solution is the introduction of a new prop `prefetchOn` that allows control over prefetching. Possible options are `interaction`, and `visibility` (default)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -56,9 +95,15 @@ export interface NuxtLinkProps extends Omit<RouterLinkProps, 'to'> {
...
-   * When enabled will prefetch middleware, layouts and payloads of links in the viewport.
+   * When enabled, will prefetch middleware, layouts and payloads of links in the viewport.
    */
   prefetch?: boolean
+  /**
+   * Determines when to prefetch links. Defaults to `on-visible`. See
+   * `prefetchMode` for more details. This prop is only used when `prefetch` is
+   * enabled and `noPrefetch` is not set.
+   */
+  prefetchMode?: 'on-hover' | 'on-visible'
```
kewinzaq1, line 106: @danielroe I think adding a prop such as `prefetchMode` makes more sense than changing the current `prefetch` to a union type with diversification. What do you think about this?

danielroe: That does make sense I think, as we could also support setting prefetch mode globally or even creating a custom Nuxt link component.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -59,6 +59,11 @@ export interface NuxtLinkProps extends Omit<RouterLinkProps, 'to'> {
    * When enabled will prefetch middleware, layouts and payloads of links in the viewport.
    */
   prefetch?: boolean
+  /**
+   * Determines when to prefetch links. Defaults to `on-visible`. This prop is only used when `prefetch` is
+   * enabled (default) and `noPrefetch` is not set.
+   */
+  prefetchMode?: 'on-hover' | 'on-visible'
```
Mini-ghost, line 66: This is just a personal opinion, but would it be better to use `visible` and `hover`?

kewinzaq1: Thanks for the suggestion, I made it this way because it's more explicit from my viewpoint. `prefetchMode`: 'on-hover".  It sounds for me like "When? On hover".

Mini-ghost: I would like to share my thoughts on this matter. When we use native event listeners, we do it like this:

```ts
element.addEventListener('pointerenter', callbackHandler)
```

It does not start with `on`. I simply believe that using `visible` and `hover` is more in line with what I assume to be common practice.

However, this is just my personal opinion. Regardless of the API you ultimately choose, I am grateful for your and the core team's hard work.

danielroe: how about:

```suggestion
  prefetchMode?: 'hover' | 'visibility'
```

kewinzaq1: @danielroe makes sense

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -355,6 +373,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
             replace: props.replace,
             ariaCurrentValue: props.ariaCurrentValue,
             custom: props.custom,
+            onMouseenter: props.prefetchMode === 'hover' && checkShouldPrefetch() ? prefetch.bind(null, undefined) : undefined,
```
Mini-ghost, line 376: Would it be better to use `onPointerEnter`? As far as I know, `MouseEvent` is not supported on mobile devices, while `PointerEvent`, which inherits from `MouseEvent`, supports input methods such as mouse, stylus, and touch.

kewinzaq1: @Mini-ghost Sounds interesting, I did not take mobile into account because the common usage of web apps on mobile won't benefit from this since users will click immediately while on a desktop is required to hover before clicking so prefetching that way is handy. If we can support prefetching on hover while using a mouse on mobile devices I'd be happy to implement it in the proposed way. Thanks - I appreciate it!

kewinzaq1: Also, it makes me think a bit about users navigating across a page using a keyboard; maybe prefetching on focus would be a good option. What do you think @Mini-ghost @danielroe

danielroe: Agreed. What about 'interaction' as the option and whatever is most generic events we can listen to.

Mini-ghost: You are correct. I tested it, and unless intentionally triggered, `hover` is indeed difficult to activate on mobile devices. However, I still recommend using `onPointerenter` as it does not have any negative impact.

Additionally, having `interaction` as one of the options is fantastic! This seems to encompass both `onPointerenter` and `onFocus` (or `onFocusin`).

kewinzaq1: @danielroe @Mini-ghost The `interaction` options sounds fine, but then it made me think if having two options `hover` and `interaction` can be a bit confused for customers of the API, and would require a vivid description in docs describing what exactly does interaction mean since it's a broad term. Ultimately I think that API like 
```tsx
  prefetchTriggers?: Partial<{
    visibility: boolean
    focus: boolean
    hover: boolean
  }>
```
With defaults:
```tsx
  prefetchTriggers: {
    visibility: true
    focus: false
    hover: false
  }
```
might be a good option. What do you think about this shift?

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -299,10 +316,26 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
       const el = import.meta.server ? undefined : ref<HTMLElement | null>(null)
       const elRef = import.meta.server ? undefined : (ref: any) => { el!.value = props.custom ? ref?.$el?.nextElementSibling : ref?.$el }
 
+      function checkShouldPrefetch () {
+        const shouldPrefetch = props.prefetch !== false && props.noPrefetch !== true && props.target !== '_blank' && !isSlowConnection() && !prefetched.value
+        return shouldPrefetch
+      }
+
+      async function prefetch (nuxtApp = useNuxtApp()) {
```
quentinmcq, line 324:
```suggestion
      async function prefetch (nuxtApp = useNuxtApp()): Promise<void> {
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -299,10 +316,26 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
       const el = import.meta.server ? undefined : ref<HTMLElement | null>(null)
       const elRef = import.meta.server ? undefined : (ref: any) => { el!.value = props.custom ? ref?.$el?.nextElementSibling : ref?.$el }
 
+      function checkShouldPrefetch () {
+        const shouldPrefetch = props.prefetch !== false && props.noPrefetch !== true && props.target !== '_blank' && !isSlowConnection() && !prefetched.value
+        return shouldPrefetch
+      }
```
quentinmcq, line 322:
```suggestion
function shouldPrefetch (): boolean {
        return props.prefetch && !props.noPrefetch && props.target !== '_blank' && !isSlowConnection() && !prefetched.value
      }
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -299,10 +316,26 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+        ])
+        prefetched.value = true
+      }
+
       if (import.meta.client) {
         checkPropConflicts(props, 'prefetch', 'noPrefetch')
-        const shouldPrefetch = props.prefetch !== false && props.noPrefetch !== true && props.target !== '_blank' && !isSlowConnection()
-        if (shouldPrefetch) {
+        const shouldPrefetch = checkShouldPrefetch()
+        if (shouldPrefetch && props.prefetchTriggers?.visibility) {
```
quentinmcq, line 338:
```suggestion
        if (shouldPrefetch() && props.prefetchTriggers?.visibility) {
```

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -355,6 +380,8 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
             replace: props.replace,
             ariaCurrentValue: props.ariaCurrentValue,
             custom: props.custom,
+            onPointerenter: (props.prefetchTriggers?.focus || props.prefetchTriggers?.hover) && checkShouldPrefetch() ? prefetch.bind(null, undefined) : undefined,
+            onFocus: props.prefetchTriggers?.focus && checkShouldPrefetch() ? prefetch.bind(null, undefined) : undefined,
```
quentinmcq, line 384:
```suggestion
            onPointerenter: (props.prefetchTriggers?.focus || props.prefetchTriggers?.hover) && shouldPrefetch() ? prefetch.bind(null, undefined) : undefined,
            onFocus: props.prefetchTriggers?.focus && shouldPrefetch() ? prefetch.bind(null, undefined) : undefined,
```

packages/nuxt/src/app/components/nuxt-link.ts
kewinzaq1: @quentinmcq the `check` prefix is more explicit, `shouldPrefetch` as a function, does not resonate with me. `checkShouldPrefetch` is clearer and depict the fact that this is a function and calling it I'm checking if prefetch should be done, adding return type is useless, inline return without variable looks like something worth considering

packages/nuxt/src/app/components/nuxt-link.ts
kewinzaq1: @quentinmcq I don't see any value with declaring the return type, inferring return type is fine

packages/nuxt/src/app/components/nuxt-link.ts
quentinmcq: > @quentinmcq the check prefix is more explicit, shouldPrefetch as a function, does not resonate with me. checkShouldPrefetch is clearer and depict the fact that this is a function and calling it I'm checking if prefetch should be done, adding return type is useless, inline return without variable looks like something worth considering

I can't find this comment but to answer it, the fact is that the name of your function is supposed to concretely define what it does there. If you add a const behind just for a return line, your function is poorly named and in our case it is redundant. Your function returns a boolean regarding whether it should retrieve data or not, so "should prefetch" matches perfectly, no need for a prefix.

kewinzaq1: @quentinmcq yeah at first glance it was the case but then after thinking more I agree with you, so I removed the comment and implemented suggested change, thx

docs/3.api/1.components/4.nuxt-link.md
```diff
@@ -185,8 +186,13 @@ interface NuxtLinkOptions {
   externalRelAttribute?: string;
   activeClass?: string;
   exactActiveClass?: string;
-  prefetchedClass?: string;
   trailingSlash?: 'append' | 'remove'
+  prefetch?: boolean
+  prefetchedClass?: string
+  prefetchOn?: Partial<{
```
atinux, line 192:
```suggestion
  prefetchOn?: 'visibility' | 'interaction' | Partial<{
```

danielroe: This deliberately forces using object syntax with `defineNuxtLink` as it's already a more 'custom' API

atinux (approved):
Looking great!

PR #23703, opened 2023-10-16
fix: remove md4 patch now that `webpack` has it in core

danielroe:
### 🔗 Linked issue

https://github.com/webpack/webpack/pull/17628
https://github.com/nuxt/nuxt/pull/19108

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Now that webpack has support for md4 in latest node versions, this PR removes our monkey-patch.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #27865, opened 2024-06-27
fix(config): add back md4 monkey-patch if needed for wider ecosystem

danielroe:
This reverts commit https://github.com/nuxt/nuxt/pull/23703.

### 📚 Description

We previously patched the `crypto.createHash` function for use on webpack 4. Webpack has now added this in core and we removed our patch, but there are still other plugins in the webpack ecosystem that fail if this monkey patch is not present, so this PR aims to revert the change before EOL for Nuxt 2.

resolves https://github.com/nuxt/nuxt/issues/10844

### DIFF
--- a/packages/config/src/options.js
+++ b/packages/config/src/options.js
@@ -520,6 +520,17 @@ export function getNuxtConfig (_options) {
     options.build.indicator = false
   }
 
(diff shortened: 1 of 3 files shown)


dockleryxk:
@danielroe you're a real one for this, thank you.

PR #27889, opened 2024-06-28
perf(nuxt)!: return island earlier in renderer

huang-julien:
### 📚 Description

Hey :wave: This PR makes island returned earlier in the nitro renderer. `render:html` won't be called anymore for islands as we have `render:island` hook


danielroe (commented):
As a positive, what kind of performance difference does this make?

The potential negative is that if users rely on a `render:html` hook, things might stop working. For example, I guess this will now bypass [html validator](https://github.com/nuxt-modules/html-validator) - or other modules which aim to manipulate rendered HTML.

A potential solution might be a 'lite' call to `render:html`, but missing all the irrelevant things we won't use in rendering the island, like bodyAttrs, etc.

huang-julien:
Yes this is intended. `render:html` feels more adequate to only hook for pages while `render:island` hook for islands.

As an example, html validator would validate twice the island in SSR. 
There's also `@nuxt/devtools` that adds some additionnal scripts we don't want to be duplicated (by the page and the island). This is currently not going client-side but with a future PR for head manipulation with islands, we need to isolate a maximum islands to avoid having duplicated tags injected into the DOM

PR #27893, opened 2024-06-28
fix(nuxt): prompt to set compatibility date with latest nitro

danielroe:
### 📚 Description

This adds a prompt + default compatibility date, which will be relevant for latest nitro patch release.


packages/nuxt/src/core/nuxt.ts
```diff
@@ -82,6 +92,59 @@ async function initNuxt (nuxt: Nuxt) {
...
+      console.info(`Nuxt now supports pinning the behavior of provider and deployment presets with a compatibility date. We recommend you specify a \`compatibilityDate\` in your \`nuxt.config\` file.`)
+    }
+
+    // Prompt to update in dev mode
+    if (!warnedAboutCompatDate && nuxt.options.dev) {
+      const result = await consola.prompt(`Do you want to update your ${colorize('cyan', 'nuxt.config')} to set ${colorize('cyan', `compatibilityDate: '${todaysDate}'`)}?`, {
+        type: 'confirm',
+        default: true,
+      })
+      if (result) {
```
pi0, line 111:
```suggestion
      if (result === true) {
```

packages/nuxt/src/core/nuxt.ts
```diff
@@ -60,6 +66,9 @@ export function createNuxt (options: NuxtOptions): Nuxt {
   return nuxt
 }
 
+// TODO: update to nitro import
+const fallbackCompatibilityDate = '2024-04-03' as DateString
```
pi0, line 70: It is a fixed thing (lasted release of Nitro that didn't had the compat date probably you can add the comment simply)

danielroe: yes, I will update when merging to 3.x branch, but current nitro-nightly doesn't have it

packages/nuxt/src/core/nuxt.ts
danielroe: what else could it be?

pi0: Symbol(abort)

PR #27919, opened 2024-06-30
docs: `compatibilityVersion` is available in the latest release.

wattanx:
### 📚 Description
Hi! 
The first of the opt in steps in Nuxt 4 is the change to nightly release, but the `compatibilityVersion` is available in the latest version.
Therefore, we would like to clarify that it is also available in the latest version.


danielroe (approved):
❤️

danielroe:
In fact we should probably also indicate that the nightly version now has breaking changes without opt out. So nightly release would have to have 3x version... Maybe worth mentioning both here and on nightly release page.

wattanx:
Indeed, that's true.
I'll create a PR!

wattanx:
You have already fixed it! 
Thank you!

PR #27934, opened 2024-07-01
docs: more accurate async data warning (#27874)

### LINKED ISSUE #27874: improve asyncData warning
**Summary (generated)**:

Here is a summary of the text in English:

The author of the message is describing a warning they received while using the Nuxt framework. They are unsure why the warning is occurring, but believe that the issue may be with returning falsey values such as an empty string. They think that the error message should be more clear about the cause of the problem.

<hr>

### Describe the feature

This warning was happening to me: https://github.com/nuxt/nuxt/pull/27599/files
I am ignorant about fine-grained reasons for this warning, but I think clarity on the cause would help.

I think it should probably be more clear that the problem is with falsey values? In my case, I was returning empty string and wondering why I was getting this warning when I knew I was not null or undefined. The error message made it sound like the only problem values were null and undefined explicitly.

Because I am not sure what the real problem is here, I didn't do a simple PR to change the language. Is it actually the case that the problem is with falsey values? A more clear explanation is visible in the code comments, for example, here:
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L132C2-L132C73

Thanks for reading!

### Additional information

### Final checks

s-en-o:
### 🔗 Linked issue

resolves #27874 

### 📚 Description

More accurate warning from async data.

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -310,7 +310,7 @@ export function useAsyncData<
 
         if (import.meta.dev && import.meta.server && !result) {
           // @ts-expect-error private property
-          console.warn(`[nuxt] \`${options._functionName || 'useAsyncData'}\` should return a value that is not \`null\` or \`undefined\` or the request may be duplicated on the client side.`)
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
Thank you ❤️

PR #27946, opened 2024-07-01
refactor: replace requireModule with ESM imports in getPostcssConfig

izzygld:
### 📚 Description

This pull request includes the following changes:

- Removed use of `requireModule` in favor of dynamic ESM imports
- Updated `sortPlugins` function and plugin instantiation to be asynchronous
- Ensured compatibility with existing plugin loading mechanism

This change modernizes the codebase by utilizing ES module imports and improves compatibility with ESM environments. The asynchronous handling of plugin instantiation aligns with the dynamic import approach, ensuring a smooth and efficient loading process for PostCSS plugins.

Solves [this TODO Message in the Repo](https://github.com/nuxt/nuxt/blob/main/packages/webpack/src/utils/postcss.ts#L66)


danielroe (commented):
this will also need to be handled here:

https://github.com/nuxt/nuxt/blob/61c3a2a4cd0c35c5c8fdc4affc2d6ace10e1ab34/packages/webpack/src/presets/style.ts#L120

... and that will in turn require any calls of `createPostcssLoadersRule` to be awaited as well, and so forth.

We should also probably fall back to CJS loading with an warning message.

danielroe:
We also likely want to mirror any improvements here too:
https://github.com/nuxt/nuxt/blob/2de885bab5b86d5e20103cf8c8172019cec92617/packages/vite/src/css.ts#L23-L28

danielroe:
(I'll take a stab at implementing those suggestions.)

packages/webpack/src/utils/postcss.ts
```diff
@@ -55,20 +57,35 @@ export const getPostcssConfig = (nuxt: Nuxt) => {
...
+
+      const path = await tryResolveModule(pluginName, nuxt.options.modulesDir)
+
+      let pluginFn: (opts: Record<string, any>) => Plugin
+      if (path) {
+        pluginFn = await import(pathToFileURL(path).href).then(interopDefault)
+      } else {
+        console.warn(`[nuxt] could not import postcss plugin \`${pluginName}\` with ESM. Please report this as a bug.`)
+        // fall back to cjs
+        pluginFn = requireModule(pluginName, { paths: [cwd] })
```
pi0, line 80: Migrating to jiti v2, this can all be done automatically and it always prefers native ESM with fallback.

izzygld: Good point.

Will need to test to see if there are any Potential Side Effects and Breaking Changes.

than you

danielroe (approved):

Issue #27106, opened 2024-05-08
Degraded dev server start as project grows in size


martinszeltins:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.14.2
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    -
- Package Manager:  pnpm@8.6.12
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

N/A

### Describe the bug

I have noticed that as my project grows _(as more components, composables, modules, plugins are added)_ - the longer dev server startup takes. It took 27 seconds from the time I ran `pnpm dev` in console to first paint when the page opened. After that each page refresh was snappy - less than 1 sec. The issue is the dev server startup time to first paint. If my app grows 10 times - that means I will be waiting for 5 minutes on dev server start. I wanted to find out if there is a guilty module or code that was the reason behind this. I don't have a very fast computer plus I am developing in a Docker container (which adds a little bit of overhead but not much).

So I started removing layer by layer from my app, it's a mid size app _(900 files in total, 6 months in development)_. As I removed modules, plugins and files, I noticed that with each thing I remove the dev server start time decreases. Eventually after removing everything, I got down to < 1 second (as with a fresh Nuxt app). This proved that it wasn't one thing that caused the slow dev server start but many things together, each adding a little bit to it.

It looks to me like we have a scaling problem with Nuxt. With each module, plugin, component that we add - we add to dev server startup time. Eventually, if we would add 100x modules, the time for dev server to start would be extremely long. That is a bottleneck for our app to scale and grow indefinitely. I thought Vite was supposed to solve this issue by loading only what we need to render the current page _(in contrast to webpack bundling the whole app)_. Perhaps some of these modules could be initialized in the background or be delayed? My plugins were already set to initialize in parallel.

I also found it strange that I gained 4 seconds by removing all components, composables, utils even when they were not being used anywhere in the app. I guess the dev server still scanned them even when not being used anywhere. I literally had a page that said `Hello World` nothing else. And after removing all unused components, composables I gained 4 seconds.

Here are my findings, I cannot share the project since it is a company project _(NDA etc.)_. But these are the times that I observed as I removed code from my app bit by bit. Yes, I took a stopwatch and measured dev server startup time multiple times after each thing I removed.

* 27 sec - dev server start ... to first paint _(from `pnpm dev` to seeing something on the page)_
[...]

Issue #26211, opened 2024-03-12
Nuxt extremely slow on dev startup


besscroft:
### Environment

Nuxt project info:                                                                                                                                                                             12:44:36

------------------------------
- Operating System: Windows_NT
- Node Version:     v18.18.0
- Nuxt Version:     3.10.3
- CLI Version:      3.10.1
- Nitro Version:    2.8.1
- Package Manager:  pnpm@8.11.0
- Builder:          -
- User Config:      modules, vue, experimental, pinia, piniaPersistedstate, css, colorMode, nitro, app, devtools
- Runtime Modules:  @vueuse/nuxt@10.9.0, @pinia/nuxt@0.5.1, @nuxtjs/color-mode@3.3.2, @unocss/nuxt@0.58.5, @nuxtjs/device@3.1.1, @formkit/auto-animate/nuxt@0.8.1, @pinia-plugin-persistedstate/nuxt@1.2.0, @element-plus/nuxt@1.0.7, @nuxt/ui@2.14.2
- Build Modules:    -
------------------------------

### Reproduction

https://github.com/besscroft/kamera

### Describe the bug

pnpm run dev
![G2` HIKBZC(9@IKE8CWTS~H](https://github.com/nuxt/nuxt/assets/33775809/a6282aa5-8120-46dd-b88e-3aa4d8448f02)

hdqbasith:
Try to add 
```
"exclude": [
    "node_modules"
],
```

to your tsconfig.json

hdqbasith:
Vite monitors every file in the directory. Usually "node_modules" included. So exclude it. Try add it in "tsconfig,json", not ".nuxt\tsconffig.json"

```
"extends": "./.nuxt/tsconfig.json",
"exclude": [
    "node_modules"
],
```

Ashlayyy:
Is there any update on this issue? I have this problem the exact same way. 
I have added the node_modules to the exclude clause like said above, but this didn't seem to help at all.

Every time I reload my webpage I have to wait almost 40 to 60 seconds, sometimes even longer.
I have added a picture of how my network tab looks like during these slow reloads.
Here is more info about the project, although I also have this issue in a brand new project using `npx nuxi@latest init <project-name>`

Information about project:
- Operating System: Windows 11
- Node Version: v20.11.0
- Nuxt Version: 3.10.3
- Nuxi Version: 3.11.1
- Nitro Version: 2.8.1
- Package Manager: npm@10.2.1

[...]

PR #27963, opened 2024-07-02
perf(vite): start warmups after nitro build

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/26211
https://github.com/nuxt/nuxt/issues/27106

### 📚 Description

Investigating reports of slow startup time for Nuxt, I think there are a couple of ways we can improve things:

1. use Vite's own warmup mechanism.
2. ensure we do not block the nitro _build_ with the warmup's heavy fs read operations (it's not a technical block as they run in parallel, but I think in practice it delays it)

Before | After
-- | --
CleanShot 2024-07-02 at 13 51 20@2x | CleanShot 2024-07-02 at 13 54 32@2x

We definitely warm up fewer files this way (for example, in the test fixture I'm running, we don't warm up plugin files) and this may be a problem.

I'm also - related to https://github.com/nuxt/nuxt/issues/26211 - trying to track down an issue on Windows machines where vite warmup seems to be non-effective in the server build - even passing entry file to `server.warmup` doesn't yield transform results for key files. (Maybe a Vite bug - cc: @antfu?)

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -217,6 +217,7 @@ export async function buildClient (ctx: ViteBuildContext) {
     // Dev
     const viteServer = await vite.createServer(clientConfig)
     ctx.clientServer = viteServer
+    ctx.nuxt.hook('close', async () => viteServer.close())
     await ctx.nuxt.callHook('vite:serverCreated', viteServer, { isClient: true, isServer: false })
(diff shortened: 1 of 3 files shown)


antfu (approved):

antfu:
Looks very nice!

Re: Vite warmup: I have not been using Windows for a long while to tell if it's a Vite bug. I guess it would need a reproduction to make sure it's about vite (or maybe our path normalization?)

PR #27975, opened 2024-07-02
fix(schema): resolve public alias correctly

KazariEX:

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -421,7 +421,7 @@ export default defineUntypedSchema({
         '~~': rootDir,
         '@@': rootDir,
         [basename(assetsDir)]: join(srcDir, assetsDir),
-        [basename(publicDir)]: join(srcDir, publicDir),
+        [basename(publicDir)]: resolve(rootDir, publicDir),
         ...val,
       }
     },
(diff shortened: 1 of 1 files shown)


packages/schema/src/config/common.ts
```diff
@@ -421,7 +421,7 @@ export default defineUntypedSchema({
         '~~': rootDir,
         '@@': rootDir,
         [basename(assetsDir)]: join(srcDir, assetsDir),
-        [basename(publicDir)]: join(srcDir, publicDir),
+        [basename(publicDir)]: resolve(rootDir, publicDir),
```
danielroe, line 424: Nice find! I think:

```suggestion
        [basename(publicDir)]: resolve(srcDir, publicDir),
```

KazariEX: Shouldn't the public folder be resolved from the ``rootDir``?

danielroe: Previously `dir.public` was resolved relative to `srcDir`. For backwards compatibility `publicDir` is now fully resolved, so effectively the first argument to `resolve` is ignored. But in the case a user sets it to `some-dir` this will be respected as `<srcDir>/some-dir`.

KazariEX: I got it, thanks for explanation.

danielroe (approved):

jd1378:
from the looks of the fix, shouldn't line above do the same ? 
https://github.com/nuxt/nuxt/blob/228b9f6820ee63a13e95c8c7d3c1a66d2408de7f/packages/schema/src/config/common.ts#L425

danielroe:
@jd1378  yes, we could indeed!

PR #27987, opened 2024-07-02
feat(nuxt): head manipulation with islands

### LINKED ISSUE #27381: No control over <head> in server components
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools, experimental
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-zsncjd?file=pages%2Findex.server.vue

### Describe the bug

Usually, you can change Head subproperties either with template, useHead, useSeoMeta or useServerSeoMeta. This works fine - unless it is a .server. component. Neither of the described functions work there. If you rename the server component back to a normal one, it works as expected. 

### Additional context

This is required for SEO optimization.

### Logs

```shell-script
There are no error logs
```

huang-julien:
### 🔗 Linked issue
resolve #27381 

### 📚 Description

This PR allow head manipulation from islands


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -480,21 +478,32 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
...
+        if (h.resolvedInput) {
+          const resolvedInput = toValue(h.resolvedInput)
+          for (const key in resolvedInput) {
+            const input = resolvedInput[key as keyof MergeHead]
+            if (Array.isArray(input)) {
+              if (toValue(input).length === 0) {
+                delete resolvedInput[key as keyof MergeHead]
+              } else {
+                // @ts-expect-error todo fix this
+                input.forEach((i) => { if (typeof i === 'object') { i.key = 'island-key-' + hash(i) } })
```
harlan-zw, line 495: what's this for?

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -480,21 +478,32 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
...
-        islandHead.style.push({ key: 'island-style-' + hash(tag.innerHTML), innerHTML: tag.innerHTML })
-      }
-    }
+    // re-push inlined styles - head is reset by the island renderer
+    head.push({ style: inlinedStyles })
     const islandResponse: NuxtIslandResponse = {
       id: islandContext.id,
-      head: islandHead,
-      html: getServerComponentHTML(htmlContext.body),
+      head: head.headEntries().filter((h) => {
```
harlan-zw, line 485: if you're just trying to resolve the reactivity, you should use

```ts
import { resolveUnrefHeadInput } from '@unhead/vue'
// ...
for (const entry of ctx.entries)
  entry.resolvedInput = resolveUnrefHeadInput(entry.input)
```

can't we then just serialize this?

packages/nuxt/src/core/runtime/nitro/renderer.ts
huang-julien: This was implemented for tag deduplication I think, @danielroe can we drop the `key` property ?

huang-julien: see here https://github.com/nuxt/nuxt/blob/128edc76fc0879119525fccee63f504657b36d63/packages/nuxt/src/core/runtime/nitro/renderer.ts#L470-L477

packages/nuxt/src/core/runtime/nitro/renderer.ts
huang-julien: This is to remove empty arrays from the entries. Do you think we can instead do
`head.headEntries().filter(h => Array.isArray(h) ? h.length : true)` ?

packages/nuxt/src/core/runtime/nitro/renderer.ts
harlan-zw: Link and style tags can't dedupe themselves so I can see why this exists (if someone is rendering multiple of the same component).

packages/nuxt/src/core/runtime/nitro/renderer.ts
harlan-zw: Is this just to reduce the payload size? Are empty arrays getting in there through user input or something we're doing :thinking:

huang-julien: yes. I'm not sure where it comes from. I'll have to debug that

test/basic.test.ts
```diff
@@ -2239,41 +2254,63 @@ describe('component islands', () => {
...
+              }
+              return h
+            })
+          }
+        }
       }
-      result.head.link.sort((a, b) => b.href!.localeCompare(a.href!))
+
+      // todo : why do we need to sort ?
+      // result.head.filter(h => h.tag === 'link').sort((a, b) => b.props.href!.localeCompare(a.props.href!))
```
huang-julien, line 2271: @danielroe why do we needed to sort the links here ?

danielroe: only for test stability - I think I worried that in dev mode vs production they might be in a different order (or maybe it was a race condition?)

test/basic.test.ts
```diff
@@ -2287,41 +2289,60 @@ describe('component islands', () => {
...
+              },
+            ],
+          },
+          {
+            "style": [
+              {
+                "innerHTML": "pre[data-v-xxxxx]{color:blue}",
+              },
+            ],
+          },
```
danielroe, line 2323: need to investigate this

huang-julien: That's because we don't use keys anymore

danielroe (approved):

PR #28000, opened 2024-07-03
docs: update example to use `nuxtApp.runWithContext`

### LINKED ISSUE #27991: navigateTo in  $fetch  onResponseError is doesn't work
### Environment

Nuxt 3.12.3 with Nitro 2.9.7  

### Reproduction

see Describe

### Describe the bug

this is offical example，But it didn't work
https://nuxt.com/docs/guide/recipes/custom-usefetch

this is plugins，to injecct nuxt
```
    const router=useRouter()
    const api = $fetch.create({
        baseURL: 'http://127.0.0.1:4523/m1/4632367-4282747-default',
         async onResponseError({response}) {
            if (response.status === 401) {
                console.log("yes, is runing here,but navigateTo doesn't work")
                 await navigateTo('/login', {
                    redirectCode: 302,
                })
            }
        }
    })

    // Expose to useNuxtApp().$api
    return {
        provide: {
            api
        }
    }
})
```

The page does not change to login.vue when my http status is 401, navigateTo doesn't work
I'm pretty sure this is the official case

```
<script setup lang="ts">
let path="/user/1"
const { $api } = useNuxtApp()
const { data: modules } = await useAsyncData(path, () => $api(path))
</script>
```

### Additional context

_No response_

### Logs

_No response_

danielroe:
resolves https://github.com/nuxt/nuxt/issues/27991

### DIFF
--- a/docs/2.guide/4.recipes/3.custom-usefetch.md
+++ b/docs/2.guide/4.recipes/3.custom-usefetch.md
@@ -24,7 +24,7 @@ Let's pretend here that:
 - If the API responds with a `401` status code, we redirect the user to the `/login` page
 
 ```ts [plugins/api.ts]
-export default defineNuxtPlugin(() => {
+export default defineNuxtPlugin((nuxtApp) => {
   const { session } = useUserSession()
 
   const api = $fetch.create({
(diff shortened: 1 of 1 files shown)


EmilRiedeman:
[In this example in the docs](https://nuxt.com/docs/examples/advanced/use-custom-fetch-composable) the custom fetch should also use `nuxtApp.runWithContext`, which is not yet the case.

PR #27995, opened 2024-07-03
feat(deps): upgrade to latest versions of c12, jiti and unbuild

danielroe:
### 🔗 Linked issue

https://github.com/unjs/nitro/pull/2582
https://github.com/unjs/c12/pull/172
https://github.com/unjs/unbuild/pull/409

### 📚 Description

This updates to prerelease versions of unbuild, jiti + c12. It brings native ESM and top level await support 🔥

#### 🚧 TODO (possibly in subsequent PRs for clarity)

- [ ] replace/update CJS utils with new jiti functionality
- [ ] update to use `jiti.import` in more places
- [ ] consider replacing ESM utils as well - wdyt @pi0?

PR #28008, opened 2024-07-03
refactor(kit,nuxt,vite,webpack)!: remove legacy require utils in favour of`jiti`

danielroe:
### 📚 Description

This follows up on https://github.com/nuxt/nuxt/pull/27995 to remove legacy CJS utils and rely on jiti to resolve/import modules and postcss plugins. 

We use `resolvePath` from `mlly` in other places which we might leave in place as it has lower overhead.

👉 Modules should now support Nuxt aliases! 🙌

### 🚨 Breaking changes

Internal CJS utils (marked as deprecated) have now been removed. They were never meant for external use anyway. We could add an example to the upgrade guide.


packages/webpack/src/utils/postcss.ts
```diff
@@ -38,28 +36,25 @@ export async function getPostcssConfig (nuxt: Nuxt) {
...
-      // TODO: use jiti v2
-      if (path) {
-        pluginFn = await import(pathToFileURL(path).href).then(interopDefault)
-      } else {
-        console.warn(`[nuxt] could not import postcss plugin \`${pluginName}\` with ESM. Please report this as a bug.`)
-        // fall back to cjs
-        pluginFn = requireModule(pluginName, { paths: [cwd] })
-      }
+      const path = jiti.esmResolve(pluginName)
+      const pluginFn = (await jiti.import(path)) as (opts: Record<string, any>) => Plugin
```
pi0, line 53: jiti will do esmResolve i think in module.ts only reason was that we need to also reuse same `path` in resolved meta

PR #28013, opened 2024-07-03
fix(kit): add back requireModule and tryRequireModule

danielroe:
### 📚 Description

reverts the rest of https://github.com/nuxt/nuxt/pull/28008 and reimplements using jiti.

### DIFF
--- a/packages/kit/src/internal/esm.ts
+++ b/packages/kit/src/internal/esm.ts
@@ -1,5 +1,6 @@
 import { pathToFileURL } from 'node:url'
 import { interopDefault, resolvePath, resolvePathSync } from 'mlly'
+import { createJiti } from 'jiti'
 
 export interface ResolveModuleOptions {
   paths?: string | string[]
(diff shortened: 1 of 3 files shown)


packages/kit/src/internal/esm.ts
```diff
@@ -40,3 +41,31 @@ export function tryImportModule<T = unknown> (id: string, opts?: ImportModuleOpt
...
 }
+
+const warnings = new Set<string>()
+
+/**
+ * @deprecated Please use `importModule` instead.
+ */
+export function requireModule<T = unknown> (id: string, opts?: ImportModuleOptions) {
+  if (!warnings.has(id)) {
+    console.warn('[@nuxt/kit] `requireModule` is deprecated. Please use `importModule` instead.')
```
pi0, line 52: Just as idea, we might use `new Error().stack` to show the source of the warning. Most likely end-users won't have a clue of where is the origin...

pi0 (approved):

PR #28053, opened 2024-07-05
docs(nuxt): update 7.state-management.md

lirantal:
### 📚 Description

The website docs say not to use the practice of ref() outside of a component's setup. It's not clear what's an example of a bad practice of using it. Adding an [example shared by Daniel Roe](https://x.com/danielcroe/status/1808232524956455106) to clarify one use-case (that shouldn't be followed).

This update to the README help clarifies it.

### DIFF
--- a/docs/1.getting-started/7.state-management.md
+++ b/docs/1.getting-started/7.state-management.md
@@ -23,7 +23,7 @@ Read more about `useState` composable.
 ## Best Practices
 
 ::warning
-Never define `const state = ref()` outside of `<script setup>` or `setup()` function.
+Never define `const state = ref()` outside of `<script setup>` or `setup()` function.<br>
 For example, doing `export myState = ref({})` would result in state shared across requests on the server and can lead to memory leaks.
 ::
 
--- a/docs/1.getting-started/7.state-management.md
+++ b/docs/1.getting-started/7.state-management.md
@@ -24,8 +24,7 @@ Read more about `useState` composable.
 
 ::warning
 Never define `const state = ref()` outside of `<script setup>` or `setup()` function.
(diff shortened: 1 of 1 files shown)


docs/1.getting-started/7.state-management.md
```diff
@@ -23,7 +23,8 @@ Read more about `useState` composable.
 ## Best Practices
 
 ::warning
-Never define `const state = ref()` outside of `<script setup>` or `setup()` function.<br>
+Never define `const state = ref()` outside of `<script setup>` or `setup()` function.
+For example, doing `export myState = ref({})` would result in state shared across requests on the server.<br>
```
danielroe, line 27:
```suggestion
For example, doing `export myState = ref({})` would result in state shared across requests on the server and can lead to memory leaks.
```

docs/1.getting-started/7.state-management.md
```diff
@@ -23,7 +23,8 @@ Read more about `useState` composable.
...
 Such state will be shared across all users visiting your website and can lead to memory leaks!
```
danielroe, line 28:
```suggestions
```

docs/1.getting-started/7.state-management.md
```diff
@@ -23,8 +23,8 @@ Read more about `useState` composable.
 ## Best Practices
 
 ::warning
-Never define `const state = ref()` outside of `<script setup>` or `setup()` function.<br>
-Such state will be shared across all users visiting your website and can lead to memory leaks!
+Never define `const state = ref()` outside of `<script setup>` or `setup()` function.
```
danielroe, line 26:
```suggestion
Never define `const state = ref()` outside of `<script setup>` or `setup()` function.<br>
```

danielroe (approved):
Thank you ❤️

PR #28088, opened 2024-07-09
fix(nuxt): solve unhead module name error

### LINKED ISSUE #28031: `unhead` setting of nuxt.config not work
### Environment

- Operating System: Darwin
- Node Version:     v20.9.0
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.4.0
- Builder:          -
- User Config:      compatibilityDate, devtools, unhead
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

When we configure it this way in versions after 3.12.0:

```ts
  unhead: {
    renderSSRHeadOptions: {
      omitLineBreaks: true,
    },
  },
})
```

We expect that the `<head>` in the production environment should have no line breaks. However, there are still line breaks, and this setting does not seem to be working.

sourcecode

### Describe the bug

In the previous PR (https://github.com/nuxt/nuxt/pull/26989), we attempted to support setting `renderSSRHeadOptions` within `unhead` in `nuxt.config.ts`.

In commit 48fffa3c618084d7ab68cf06ef22d97f4b15b09a, `configKey: 'unhead'` was removed. This caused the functionality to stop working unless we use the `meta` setting.

```ts
  meta: {
    renderSSRHeadOptions: {
      omitLineBreaks: true,
    },
  },
})
```

What would be the best course of action? Should we restore the `configKey` or change `name: 'meta'` to `name: 'unhead'`?

Once it is clear which solution to adopt for the fix, I would be more than happy to promptly assist in resolving this issue.

### Additional context

_No response_

### Logs

_No response_

Mini-ghost:
### 🔗 Linked issue

resolves #28031

### 📚 Description

Change the unhead module name from `meta` to `unhead` to ensure the configuration in `nuxt.config.ts` aligns with the documentation.

### DIFF
--- a/packages/nuxt/src/head/module.ts
+++ b/packages/nuxt/src/head/module.ts
@@ -7,7 +7,8 @@ const components = ['NoScript', 'Link', 'Base', 'Title', 'Meta', 'Style', 'Head'
 
 export default defineNuxtModule<NuxtOptions['unhead']>({
   meta: {
-    name: 'unhead',
+    name: 'meta',
+    configKey: 'unhead',
   },
   async setup (options, nuxt) {
     const runtimeDir = resolve(distDir, 'head/runtime')
--- a/packages/nuxt/src/head/module.ts
+++ b/packages/nuxt/src/head/module.ts
@@ -7,7 +7,7 @@ const components = ['NoScript', 'Link', 'Base', 'Title', 'Meta', 'Style', 'Head'
 
 export default defineNuxtModule<NuxtOptions['unhead']>({
   meta: {
-    name: 'meta',
+    name: 'unhead',
   },
   async setup (options, nuxt) {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/head/module.ts
```diff
@@ -7,7 +7,7 @@ const components = ['NoScript', 'Link', 'Base', 'Title', 'Meta', 'Style', 'Head'
 
 export default defineNuxtModule<NuxtOptions['unhead']>({
   meta: {
-    name: 'meta',
+    name: 'unhead',
```
danielroe, line 10: Let's add a configKey property instead, and rename the module in a separate PR...

danielroe (approved):

danielroe (approved):

PR #28125, opened 2024-07-11
fix: `getNuxtVersion` returning `any`

BobbieGoede:
### 📚 Description
Currently the return type does not match the documentation: https://nuxt.com/docs/api/kit/compatibility#getnuxtversion

### DIFF
--- a/packages/kit/src/compatibility.ts
+++ b/packages/kit/src/compatibility.ts
@@ -108,7 +108,7 @@ export function isNuxt3 (nuxt: Nuxt = useNuxt()) {
  */
 export function getNuxtVersion (nuxt: Nuxt | any = useNuxt() /* TODO: LegacyNuxt */) {
   const rawVersion = nuxt?._version || nuxt?.version || nuxt?.constructor?.version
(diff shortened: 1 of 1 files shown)


packages/kit/src/compatibility.ts
```diff
@@ -108,7 +108,7 @@ export function isNuxt3 (nuxt: Nuxt = useNuxt()) {
  */
 export function getNuxtVersion (nuxt: Nuxt | any = useNuxt() /* TODO: LegacyNuxt */) {
   const rawVersion = nuxt?._version || nuxt?.version || nuxt?.constructor?.version
-  if (!rawVersion) {
+  if (!rawVersion || typeof rawVersion !== 'string') {
```
huang-julien, line 111:
```suggestion
  if (typeof rawVersion !== 'string') {
```

I think we can directly do this

BobbieGoede: I think you're right!

TheAlexLichter (approved):

huang-julien (approved):

PR #28138, opened 2024-07-12
docs: correct showError tip

huang-julien:
### 📚 Description

This PR correct the doc about `showError`


huang-julien:
closing in favor iof #28139

PR #20585, opened 2023-04-29
fix(nuxt): ensure `useError` is called with nuxt app context

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20578

### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This is a long standing issue that was uncaught until we merged https://github.com/nuxt/nuxt/pull/20511. Basically registering _any_ `app:error` hook would cause the nuxt context to be unset, and therefore `showError` would fail to correctly set the SSR error, triggering an unhandled error and a normal 200 response.

This PR does two things:

1. it no longer calls `app:error` on server twice (first, _immediately_ when the error occurs and then second, before rendering the error response). Instead, it will be called with the final resolved error before rendering the error response.
2. ~~For Node v16, we use `appendHeader` to add any new cookies set in the error handler rather than completely overriding `set-cookie` headers. This seems to be a node 16 bug that is fixed in node 18+. This now has the side effect of creating a 'double' cookie entry for anything set in both normal + error pages, but I think this is 'safe.' We can either adopt this workaround now (and revert once node 16 is eol in September)~~ or add a warning in the docs - I think it that might actually be the better strategy.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #28139, opened 2024-07-12
fix(nuxt): call `app:error` in ssr

huang-julien:
### 📚 Description

This PR reverts partially https://github.com/nuxt/nuxt/pull/20585 .
Currently throwing an error in a vue component does call `app:error` but not when calling `showError` . The issue mentionned does not happen anymore (app:error being called twice)

### DIFF
--- a/packages/nuxt/src/app/composables/error.ts
+++ b/packages/nuxt/src/app/composables/error.ts
@@ -24,9 +24,7 @@ export const showError = <DataT = unknown>(
     const nuxtApp = useNuxtApp()
     const error = useError()
 
-    if (import.meta.client) {
-      nuxtApp.hooks.callHook('app:error', nuxtError)
-    }
+    nuxtApp.hooks.callHook('app:error', nuxtError)
 
     error.value = error.value || nuxtError
   } catch {


danielroe:
Could you share a reproduction of the issue? 🙏 

We should be calling `app:error` here on SSR:

https://github.com/nuxt/nuxt/blob/0be6929523ddde35893a616d849e38305e40540c/packages/nuxt/src/core/runtime/nitro/renderer.ts#L349

huang-julien:
https://stackblitz.com/edit/nuxt-starter-elgyv8?file=pages%2Findex.vue,plugins%2Flog-error.ts,app.vue,pages%2Fshow-error.vue

huang-julien:
I did dig a bit more, seems like error handling with vue ssr and suspense is swallowed by renderToString.... 

I think we should just open the issue and mark it upstream except if there's a  workaround

huang-julien:
I think error handling are swallowed due to suspense ....

huang-julien:
closeing in favor of https://github.com/nuxt/nuxt/issues/28162

PR #28154, opened 2024-07-15
fix(nuxt): only warn when `useAsyncData` returns undefined

### LINKED ISSUE #28096: Should `useAsyncData` emit warn in dev mode when returning `false` or `''` values?
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-nhpzcs?file=app.vue

### Describe the bug

When returning `false` or `''` from `useAsyncData`, we get the following warn in dev mode: 
```bash
 WARN  [nuxt] useAsyncData must return a truthy value (for example, it should not be undefined, null or empty string) or the request may be duplicated on the client side.
```

The request is indeed duplicated on the client side when returning `null` or `undefined` but it is not when returning `false` or `''`, as you can see on the reproduction link. Is there anything else that's preventing `false`/`''` to be returned form `useAsyncData`? Should this warn be displayed only when returning `null`/`undefined`?

(I'll happily fix/update this behaviour if it's confirmed)

### Additional context

_No response_

### Logs

_No response_

xjccc:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/28096

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -250,16 +250,16 @@ export function useAsyncData<
   }
 
   // TODO: make more precise when v4 lands
-  const hasCachedData = () => options.getCachedData!(key, nuxtApp) != null
+  const hasCachedData = () => options.getCachedData!(key, nuxtApp) !== undefined
 
   // Create or use a shared asyncData entity
   if (!nuxtApp._asyncData[key] || !options.immediate) {
     nuxtApp.payload._errors[key] ??= undefined
 
     const _ref = options.deep ? ref : shallowRef
-
+    const cachedData = options.getCachedData!(key, nuxtApp)
     nuxtApp._asyncData[key] = {
-      data: _ref(options.getCachedData!(key, nuxtApp) ?? options.default!()),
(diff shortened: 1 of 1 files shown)


danielroe (approved):

Brain1024:
@danielroe Hi, I updated Nuxt to 3.12.4. When I use `useAsyncData` to return null, the following warning occurs:
`WARN  [nuxt] useAsyncData must return a value (it should not be undefined) or the request may be duplicated on the client side.
`
However, I also found that the request is still duplicated on the client side. It seems we only modified the warning content, but the behavior remains the same as in version 3.12.2.

danielroe:
@Brain1024 That is correct. The request will be duplicated if it is either undefined or null. That will not change until Nuxt 4.

Brain1024:
> @Brain1024 That is correct. The request will be duplicated if it is either undefined or null. That will not change until Nuxt 4.

ok, I got it, thanks.

PR #28156, opened 2024-07-15
fix(nuxt): hide unhandled error messages in prod

danielroe:
### 🔗 Linked issue

https://github.com/unjs/nitro/pull/2591

### 📚 Description

This applies a fix made in nitro to our forked version of error utility. From @pi0:

> We should hide error messages (to client) in case of unhandled errors because they can include internal server state.
>
> Thanks for @GalacticHypernova for discovering this! (issue was if a chunk is missing, we were showing full path but it can be extended to other causes)
>
> Note: This might be a minor behavior change but since we only hide unhandled errors, (new Error not createError) i guess should be safe for v2 as patch.


test/basic.test.ts
```diff
@@ -1051,7 +1051,7 @@ describe('errors', () => {
     const error = await res.json()
     delete error.stack
     expect(error).toMatchObject({
-      message: 'This is a custom error',
+      message: isDev() ? 'This is a custom error' : 'internal server error',
```
danielroe, line 1054: @pi0 This behaviour change seems like it might be unexpected. I would expect throwing as below:

```ts
throw createError({ statusCode: 422, fatal: true, statusMessage: 'This is a custom error' })
```

Would behave the same way in dev + production as far as the rendered error _message_.

pi0: In nitro security model, assume both unhandled (thrown unintentionally) and fatal (thrown intentionally) as internal errors. Doing this for nuxt can have security implications that only nuxt users will have. if Nuxt shows message of fatal errors in the client rendered HTML, i highly advice to use another flag to distinguish that and keep both secret. (server logs and hooks still get the right message)

danielroe: This is also for the JSON response from an error endpoint.

I'm more worried that this would be a breaking change for users and with no way for them to customise the `message` field of an error.

pi0: Can nuxt use a new flag to avoid this? (I'm not sure exactly the situation, if you can maybe give me a code snipped i might probably have better knowledge of it..) 

but regardless if you feel this might be a breaking change, i can wait for nitro 2.x release until we adopt a new flag in nuxt internals that avoids a hard breaking change yet prioritizes security (i know it is it -dx but +secuirty really makes sense also to hide internals as much as we can..)

danielroe: This PR is for Nuxt v4 so it shouldn't block Nitro v2 release. The 3.x branch does not have a custom fork of `normalizeError` so it will inherit the nitro change automatically.

I missed that in the test we're setting `fatal: true` deliberately, so it is indeed possible to have a custom `message`. But we do advise users to do this if they want the possibility of a full-screen error on client:

https://github.com/nuxt/nuxt/blob/2afc1fe2264ae6fd51853dd53f26f2e08812323e/docs/1.getting-started/8.error-handling.md#L158-L160

(This behaviour will now potentially cause issues.)

* first issue is in rendering JSON response. I think that it's a strange thing to suppress error messages that have been crafted by the end user + thrown with `createError`. I'm not sure we should handle this with `internal: false` or whatever as this is pure nitro/h3 behaviour...
* second issue is that now we are unexpectedly passing less detail to `error.vue` to render/handle the error. this would only be embedded in the page if the user chose to do so - but 🤷

Is it possible that we could differentiate somehow an error produced with `createError` and explicitly thrown by a user? (That is, only care about `unhandled` and not care about `fatal`?) I've just reverted the change where I implemented this, but it feels reasonable....

pi0: Ah right. In the future i think an `internal: boolean` flag can help. in meantime https://github.com/unjs/nitro/pull/2619

PR #28160, opened 2024-07-15
feat(nuxt): add a plugin that triggers an immediate reload on chunkError & manifest update

### LINKED ISSUE #23612: handle chunk loading errors that are not triggered by navigation
### Describe the feature

When a chunk has been given a new name during build as it has changed, some clients will try to load the old chunk name. This results in errors like:
- `TypeError: error loading dynamically imported module:  ....`
- `Importing a module script failed.`
- `Failed to fetch dynamically imported module: ...`

With pages this issue gets detected, and the page reloaded which is necessary.
https://github.com/nuxt/nuxt/blob/24c5af328f94bb3cba26121530406ee5ac124dfe/packages/nuxt/src/app/plugins/chunk-reload.client.ts#L21
However, when I other lazy components that aren't pages, they might fail the same way, but there's have no strategy in place for them, nor can I safely tell the page and another lazy component error apart.

Solutions:

1. Don't emit app:chunkError for pages as this issue is handled and there's nothing further to be done (This counts as a breaking change, I take it). This would mean the user is free to implement their own error handling strategy.
2.  Perhaps force reload the manifest file and attempt to load the chunk again.
3. Weave in an argument to the app:chunkError hook that signifies that the chunk is a page or not.

I don't think it would be wise to refresh the page on any and all chunk errors (as a lazy component might not be too critical and it's more important to have a smooth ux).

If nuxt cannot resolve this issue universally, perhaps we could come up with a strategy and some documentation to include in the docs.

Original Issue for this feature: https://github.com/nuxt/nuxt/issues/14594#issuecomment-1757162054

### Additional information

### Final checks

BracketJohn:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/23612

### 📚 Description

This PR:
- adds a plugin which performs an immediate reload if a `chunkError` happens
  - the plugin is enabled by setting `emitRouteChunkError: 'automatic-immediate'`
- adds docs to explain the new `emitRouteChunkError`-option
- improves docs by dropping the word `hard` from `hard reload` (see reasoning below)

**Long form**

This PR adds an alternative version of the already existing `chunk-reload` plugin: `chunk-reload-immediate`. The goal of this plugin is to help all users that encounter with lazy-component loading errors in their live-apps (e.g.: due to component-chunks being replaced by a deployment).

Compared to the `chunk-reload`-plugin it:
- triggers a `reloadNuxtApp`-call right when the `app:chunkError`-hook is triggered, instead of waiting for route-navigation
- triggers the `reloadNuxtApp`-call on the current route, instead of on the to-route (impossible, as there is no to-route in this case)

This PR is based on this proposal https://github.com/nuxt/nuxt/issues/23612#issuecomment-2225145683, given a first positive indication by @danielroe (❤️). I improved on the proposal with better naming, going with `-immediate` as a post-fix, to clarify the main difference: The reload happens right away instead of waiting for navigation.

In addition I noticed that the docs use `hard reload` when talking about this feature and the `reloadNuxtApp`-composable. However, AFAIK, the reload performed by Nuxt is not hard in the sense that it bypasses browser caching:
https://github.com/nuxt/nuxt/blob/c260eaac57916e0eb2e584437c4ab42f20386a9f/packages/nuxt/src/app/composables/chunk.ts#L56-L60

`window.location.reload()` performs a regular reload: [As per the MDN-docs](https://developer.mozilla.org/en-US/docs/Web/API/Location/reload) on `window.location.reload` only Firefox supports a hard reload that bypasses the cache, but for this you'd need to call `window.location.reload(true)`.

Let me know if I have a miss-understanding on the latter part and I'm happy to change it back. I decided against updating the code to `.reload(true)`, as it would result in inconsistent behavior between browsers.


docs/2.guide/3.going-further/1.experimental-features.md
```diff
@@ -59,14 +59,16 @@ This feature will likely be removed in a near future.
 
 ## emitRouteChunkError
 
-Emits `app:chunkError` hook when there is an error loading vite/webpack chunks. Default behavior is to perform a hard reload of the new route when a chunk fails to load.
+Emits `app:chunkError` hook when there is an error loading vite/webpack chunks. Default behavior is to perform a reload of the new route on navigation to a new route when a chunk fails to load.
+
+If you set this to `'automatic-immediate'` Nuxt will reload the current route immediatly, instead of waiting for a navigation. This is useful for chunk errors that are not triggered by navigation, e.g., when your Nuxt app fails to load a [lazy component](docs/guide/directory-structure/components#dynamic-imports). A potential downside of this behavior is undesired reloads, e.g., when your app does not need the chunk that caused the error.
```
BracketJohn, line 64:
```suggestion
If you set this to `'automatic-immediate'` Nuxt will reload the current route immediatly, instead of waiting for a navigation. This is useful for chunk errors that are not triggered by navigation, e.g., when your Nuxt app fails to load a [lazy component](/docs/guide/directory-structure/components#dynamic-imports). A potential downside of this behavior is undesired reloads, e.g., when your app does not need the chunk that caused the error.
```

BracketJohn:
@danielroe anything I can do to push this PR along? (:

Would be cool to land this in [`3.12.5`](https://github.com/nuxt/nuxt/pull/28217) or know what the plan with this PR is, but of course don't know when you want to land it at all & have no expectations here. Thanks!

packages/nuxt/src/app/plugins/chunk-reload-immediate.client.ts
```diff
@@ -0,0 +1,16 @@
...
+import { reloadNuxtApp } from '../composables/chunk'
+
+const reloadNuxtApp_ = () => { reloadNuxtApp({ persistState: true }) }
+
+// See https://github.com/nuxt/nuxt/issues/23612 for more context
+export default defineNuxtPlugin({
+  name: 'nuxt:chunk-reload-immediate',
+  setup (nuxtApp) {
+    // Reload when a `chunkError` is thrown
+    nuxtApp.hook('app:chunkError', reloadNuxtApp_)
```
danielroe, line 11: If we are reloading while navigating, we should probably reload at the new path rather than the current one.

noriyuki-shimizu: @danielroe ( @BracketJohn )
Can I ask you something? :)
The image of the source would be as follows?

```ts
    nuxtApp.hook('app:chunkError', () => {
      reloadNuxtApp({
        persistState: true,
        path: nuxtApp._route.path
      })
    })
```

danielroe: Yes, much like that! Except, I think, that we might need to track the potential route with a middleware in case the error happens whilst navigating, but before the route has updated.

noriyuki-shimizu: You're absolutely right. We should definitely account for the possibility of an error occurring during navigation before the route is updated. In that case, it might be a good idea to use addRouteMiddleware.
Here is a sample code.

```ts
    const state = useState<{ path: string | null }>('chunk-reload-immediate-path-state', () => ({
      path: null
    }))

    addRouteMiddleware((to) => {
      state.value.path = to.path
    })

    nuxtApp.hook('app:chunkError', () => {
      reloadNuxtApp({
        persistState: true,
        path: state.value.path ?? nuxtApp._route.path
      })
    })
```

I'm just unsure if the middleware's process is executed before `app:chunkError` . 🤔

BracketJohn: > If we are reloading while navigating, we should probably reload at the new path rather than the current one.

@danielroe good point. I've implemented an approach that remembers the last `currentlyNavigationTo = to.path` using a route middleware (similar to how @noriyuki-shimizu proposed above). If no `currentlyNavigationTo` is known, we fallback to `nuxtApp._route.path`.

BracketJohn:
@danielroe think you can have a look here and let me know what is the next step? I'm also happy for you to push this PR to a finish if it's easier / more efficient than iterating.

AlejandroAkbal:
Half of my Sentry errors are because of lazy loaded components

I'm so eager for this to be merged 🥹

test/bundle.test.ts
```diff
@@ -25,7 +25,7 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
     expect.soft(roundToKilobytes(clientStatsInlined!.totalBytes)).toMatchInlineSnapshot(`"114k"`)
 
     const files = new Set([...clientStats!.files, ...clientStatsInlined!.files].map(f => f.replace(/\..*\.js/, '.js')))
-
+
```
BracketJohn, line 28:
```suggestion

```

test/bundle.test.ts
```diff
@@ -25,7 +25,6 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
     expect.soft(roundToKilobytes(clientStatsInlined!.totalBytes)).toMatchInlineSnapshot(`"114k"`)
 
     const files = new Set([...clientStats!.files, ...clientStatsInlined!.files].map(f => f.replace(/\..*\.js/, '.js')))
-
     expect([...files]).toMatchInlineSnapshot(`
```
BracketJohn, line 28:
```suggestion

```

jexme:
@danielroe Is there any information on the progress of this MR?

Rigo-m:
@AlejandroAkbal  @jexme you can quickly implement it as a plugin in your nuxt application today without waiting for it to be merged (worked wonders in our production apps)

danielroe (approved):

MikeBellika:
> @danielroe I wonder if this should be a default in Nuxt 4 compared to `automatic` 🤔

@TheAlexLichter Did you ever come to a conclusion on this? I think it would be a great idea. I'm getting a ton of `Failed to fetch dynamically imported module` errors in sentry, even without using any lazy components in our app.

I can consistent reproduce it in our mono repo like this:
1. `pnpm generate`
2. `npx serve .output/public`
3. Go to a page that doesn't use a specific component from our internal UI library (This is another package in our mono repo). For example our index page doesn't have any `FormButton` elements
4. Make some changes to `FormButton`
5. `pnpm generate` while the app is still being served
6. Navigate to a page that has a `FormButton`

The impact is not consistent but in a lot of cases the app crashes.
Might seem like an edge case but we're seeing this error multiple times a day in Sentry. Every time we deploy while we have active users on the site, we run the risk of getting this error.

I think that an unneeded reload is much better UX than a crashing app.
I'd happily create the PR to make this default option and update the docs.

TheAlexLichter:
I think that wouldn't hurt! @danielroe any objections?

Issue #27660, opened 2024-06-17
Use useCookie to modify the value in one component, but the value in other components is not modified


thhary:
### Environment

- Operating System: Windows_NT
- Node Version:     v20.9.0
- Nuxt Version:     3.12.2
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  npm@10.1.0
- Builder:          -

### Reproduction

When running on my device, the value of useCookie can be modified correctly when I run 'npm run dev', only when I run 'npm run generate', the value of useCookie will not be updated.
But in this website, even if I just run 'npm run dev', the value of useCookie will not be updated
https://stackblitz.com/edit/github-jaqn2d?file=components%2FChild.vue&file=app.vue

### Describe the bug

![image](https://github.com/nuxt/nuxt/assets/74284785/f0c277b6-7f31-4cf7-9eb6-9b3ee2d7dfe9)

estige:
I also have this problem after updating to 3.12

vptill:
Same issue for me after updating to 3.12

yesworld:
Same issue for me after updating from 3.10 to 3.12.2. Have to downgrade to 3.10.. :(

rolfkoizumi:
Having the same issue on MacOS with nuxt@3.12.2. Last working version seems to be 3.11.2 for me.
`useCookie()` is not reactive anymore and won't update its value if the cookie is changed. I encounter this even with `npm run dev`.

Reproductions on Stackblitz where I use `useCookie()` in a component and add this component twice. Editing the cookie value in one component should update the value in the other component. With nuxt@3.11.2 it is working, with nuxt@3.12.2 not anymore:
- [useCookie with nuxt@3.11.2](https://stackblitz.com/edit/nuxt-starter-wmwvx2?file=package.json,components%2FEditCookie.vue,pages%2Findex.vue)
- [useCookie with nuxt@3.12.2](https://stackblitz.com/edit/nuxt-starter-hhkqfe?file=package.json,components%2FEditCookie.vue,pages%2Findex.vue)

manniL:
This is related to the now switched-on `cookieStore` feature. As a workaround, disabling it for now (`experimental.cookieStore` to `false`) should resolve the issue.

https://github.com/nuxt/nuxt/commit/f597ca59a

TheDutchCoder:
I think this is a much deeper issue.

[...]

PR #28211, opened 2024-07-18
fix(nuxt): prevent duplicate `set-cookie` headers

danielroe:
### 📚 Description

When `useCookie` is called more than once, it results in multiple `set-cookie` headers.

This avoids setting duplicate headers if the value is unchanged. (If it is changed, it will continue to previous behaviour and add a second header.)

In dev mode, it will warn if a different value is passed as this can indicate an issue that can trigger a hydration mismatch (e.g. https://github.com/nuxt/nuxt/issues/27660#issuecomment-2233048700).

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -126,6 +126,16 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?: 
     const nuxtApp = useNuxtApp()
     const writeFinalCookieValue = () => {
       if (opts.readonly || isEqual(cookie.value, cookies[name])) { return }
+      nuxtApp._cookies ||= {}
+      if (name in nuxtApp._cookies) {
+        // do not append a second `set-cookie` header
+        if (isEqual(cookie.value, nuxtApp._cookies[name])) { return }
+        // warn in dev mode
+        if (import.meta.dev) {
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -126,6 +126,16 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?:
     const nuxtApp = useNuxtApp()
     const writeFinalCookieValue = () => {
       if (opts.readonly || isEqual(cookie.value, cookies[name])) { return }
+      nuxtApp._cookies ||= {}
+      if (name in nuxtApp._cookies) {
+        // do not append a second `set-cookie` header
+        if (isEqual(cookie.value, nuxtApp._cookies[name])) { return }
```
pi0, line 132: Checking equality only allows dedup when exact same options are used. Cookies are distinct by name, domain, path, secure, httpOnly and sameSite.

Also parse from cookie-es, does not parse all `set-cookie` options

See https://github.com/unjs/h3/blob/b078181bd5c103dd7365bd33e263c492399afd67/src/utils/cookie.ts#L185, https://github.com/unjs/h3/blob/b078181bd5c103dd7365bd33e263c492399afd67/src/utils/cookie.ts#L44 and https://github.com/unjs/cookie-es/pull/43 for test.

We might do same thing in Nuxt composable or at least until adopting h3 v2 which properly do this.

PR #28215, opened 2024-07-19
perf(nuxt): call cookie `decode` function only for named cookie

### LINKED ISSUE #27246: Decode function for useCookie called for every cookie present
### Environment

- Operating System: Darwin
- Node Version:     v20.10.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Package Manager:  pnpm@9.0.6
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

1. Call `useCookie` more than once with unique cookie names. 
2. Add a decode function to at least one of the calls

### Describe the bug

The decode function is called with the values for all cookies present, while the expected behaviour is for the function to be called with the value for the cookie name the hook was registered with. 

image
Reproduction:

- [Repo](https://github.com/Eazash/nuxt-useCookie-repro)
- [Playground](https://stackblitz.com/~/github.com/Eazash/nuxt-useCookie-repro)

### Additional context

_No response_

### Logs

_No response_

yxw007:
…decoding (#27246)

### 🔗 Linked issue

Resolves #27246

### 📚 Description

 [Related Discussions](https://github.com/unjs/cookie-es/pull/32)


packages/nuxt/package.json
```diff
@@ -74,7 +74,7 @@
     "chokidar": "^3.6.0",
     "compatx": "^0.1.8",
     "consola": "^3.2.3",
-    "cookie-es": "^1.1.0",
+    "cookie-es": "^1.2.0",
```
danielroe, line 77:
```suggestion
    "cookie-es": "^1.1.0",
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -21,9 +21,10 @@ export interface CookieOptions<T = any> extends _CookieOptions {
   default?: () => T | Ref<T>
   watch?: boolean | 'shallow'
   readonly?: boolean
+  filter?: (key: string) => boolean
 }
 
-export interface CookieRef<T> extends Ref<T> {}
+export interface CookieRef<T> extends Ref<T> { }
```
danielroe, line 27:
```suggestion
export interface CookieRef<T> extends Ref<T> {}
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -38,7 +39,10 @@ const store = import.meta.client && cookieStore ? window.cookieStore : undefined
 export function useCookie<T = string | null | undefined> (name: string, _opts?: CookieOptions<T> & { readonly?: false }): CookieRef<T>
 export function useCookie<T = string | null | undefined> (name: string, _opts: CookieOptions<T> & { readonly: true }): Readonly<CookieRef<T>>
 export function useCookie<T = string | null | undefined> (name: string, _opts?: CookieOptions<T>): CookieRef<T> {
-  const opts = { ...CookieDefaults, ..._opts }
+  const filter = (key: string) => {
+    return key !== name
+  }
+  const opts = { ...CookieDefaults, ..._opts, filter }
```
danielroe, line 45: I think we should only add filter if user hasn't provided their own.

yxw007: yes, you are right

test/nuxt/composables.test.ts
```diff
@@ -691,6 +691,26 @@ describe('useCookie', () => {
     expect(computedVal.value).toBe(0)
   })
 
+  it('cookie decode function should be invoked once', () => {
```
danielroe, line 694: this test passes on main branch as well - would you look at this? 🙏

yxw007: > this test passes on main branch as well - would you look at this? 🙏This test passes on main branch as well - would you look at this? 🙏

After looking at your refactoring and simplification of my code, I found no problems. Is there anything else I need to cooperate with?

danielroe: the code is good - but the test doesn't test it, as it also passes on the `main` branch, even without this PR. would you update the test so it _fails_ on the main branch, but passes with this PR? 🙏

yxw007: OK, I found that the new version of cookie-es does not have the code I changed. I will ask what is happening with pi0, and then add test cases.

yxw007: Cookie filter test case added.
by the way, the judgment conditions of the filter have been revised.

test/nuxt/composables.test.ts
```diff
@@ -691,6 +692,56 @@ describe('useCookie', () => {
     expect(computedVal.value).toBe(0)
   })
 
+  it('preset test cookie', () => {
```
yxw007, line 695: I don't understand why, in the same test case, some cookies are preset but in subsequent test code, I can't get cookies even if I try to add delayed fetch." That's why I preset cookies in separate test cases in advance and then add test cases later (because that's the way to get to the preset cookies). I know this approach isn't very reliable, but after reviewing the project code, I can't think of a better solution. Is there a better solution?

test/nuxt/composables.test.ts
```diff
@@ -691,6 +692,56 @@ describe('useCookie', () => {
...
+    const foo = useCookie('foo', {
+      default: () => 'Foo',
+      maxAge: 60 * 60,
+    })
+    foo.value = 'Foo'
```
yxw007, line 700: Why does foo's cookie have to be modified before it is set into cookie ?

PR #28223, opened 2024-07-19
docs: fix issue in cookie passing example 

johankrijt:
Use H3's `splitCookiesString` function to split the set-cookie header without choking on commas that are within a single set-cookie field-value, such as in the Expires portion.

### DIFF
--- a/docs/1.getting-started/6.data-fetching.md
+++ b/docs/1.getting-started/6.data-fetching.md
@@ -466,13 +466,13 @@ Be very careful before proxying headers to an external API and just include head
  If you want to pass on/proxy cookies in the other direction, from an internal request back to the client, you will need to handle this yourself.
 
 ```ts [composables/fetch.ts]
-import { appendResponseHeader, H3Event } from 'h3'
+import { appendResponseHeader, H3Event, splitCookiesString } from 'h3'
 
 export const fetchWithCookie = async (event: H3Event, url: string) => {
   /* Get the response from the server endpoint */
   const res = await $fetch.raw(url)
   /* Get the cookies from the response */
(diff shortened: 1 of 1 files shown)


danielroe (approved):

docs/1.getting-started/6.data-fetching.md
```diff
@@ -466,13 +466,13 @@ Be very careful before proxying headers to an external API and just include head
...
 ```ts [composables/fetch.ts]
-import { appendResponseHeader, H3Event } from 'h3'
+import { appendResponseHeader, H3Event, splitCookiesString } from 'h3'
 
 export const fetchWithCookie = async (event: H3Event, url: string) => {
   /* Get the response from the server endpoint */
   const res = await $fetch.raw(url)
   /* Get the cookies from the response */
-  const cookies = (res.headers.get('set-cookie') || '').split(',')
+  const cookies = splitCookiesString(res.headers.get('set-cookie') || '')
```
pi0, line 475: We can instead also use native [headers.getSetCookie](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) that returns properly splitter cookies (please double check if works within current nitro v1 env)

danielroe:
Sorry, didn't spot your comments @pi0!

docs/1.getting-started/6.data-fetching.md
danielroe: works well!

johankrijt: Even nicer! Sorry I was to late to the party to safe you the work, as I see it's already fixed.

PR #28232, opened 2024-07-19
fix(nuxt): call close hook in dev

### LINKED ISSUE #28188: Nitro `close` hook not being called when stopping dev server
### Environment

- Operating System: Darwin
- Node Version:     v20.12.2
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Package Manager:  pnpm@9.5.0
- Builder:          -
- User Config:      ssr, srcDir, devtools, modules, app, runtimeConfig, components, tailwindcss, typescript, compatibilityDate
- Runtime Modules:  @pinia/nuxt@0.5.1, @vueuse/nuxt@10.11.0, @nuxt/eslint@0.3.13, @nuxt/icon@1.1.1, @nuxtjs/tailwindcss@6.12.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-ftkhvf?file=server%2Fplugins%2Fshutdown.ts

### Describe the bug

Even if I hook to the Nitro `"close"` lifecycle hook using a server plugin, the code inside the hook doesn't get called when stopping the dev server (`npm run dev` then `Ctrl + C`). However, if I run the production build and stop it, the hook does get called correctly.

I would expect that the close hook gets called so I can perform a graceful shutdown of my database connections and queues even on development, but for now it only seems to work on the production build.

For some reason, the next time I start the dev server, the logs from the close hook get called on startup but not on close...

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue

resolve #28188

### 📚 Description

This PR call the `close` hook when closing the process in dev

### DIFF
--- a/packages/nuxt/src/core/builder.ts
+++ b/packages/nuxt/src/core/builder.ts
@@ -18,6 +18,11 @@ export async function build (nuxt: Nuxt) {
 
   if (nuxt.options.dev) {
     watch(nuxt)
+
+    process.on('exit', () => {
+      nuxt.callHook('close', nuxt)
+    })
+
     nuxt.hook('builder:watch', async (event, relativePath) => {
       if (event === 'change') { return }
       const path = resolve(nuxt.options.srcDir, relativePath)


packages/nuxt/src/core/builder.ts
```diff
@@ -18,6 +18,11 @@ export async function build (nuxt: Nuxt) {
 
   if (nuxt.options.dev) {
     watch(nuxt)
+
+    process.on('exit', () => {
+      nuxt.callHook('close', nuxt)
+    })
```
danielroe, line 24: this doesn't seem to result in this hook being called:

```ts
  hooks: {
    close: () => {
      console.log('closing things')
    },
  },
})
```

huang-julien: oh indeed, it's weird, the exit listener is correctly called

huang-julien: hmmm the main issue is that callHook is async

huang-julien: tbh it is probably not possible at least, we can't support it on windows since SIGTERM directly terminate the process without any way to handle the exit. On linux tho, we can probably try the `beforeExit` hook.

WDYT ? Should we drop this ?

danielroe:
Yes, likely we should drop it in this form. Possibly something could be implemented in https://github.com/nuxt/cli instead.

What do you think @pi0?

PR #28246, opened 2024-07-21
fix(nuxt): ignore definePageMeta warning for client only pages in dev

huang-julien:
### 📚 Description

Hi :wave: this PR fix an issue with https://github.com/nuxt/nuxt/pull/27839
Since the wrapper is being rendered as the route component for client-pages, we need to mark clientonly page components so `definePageMeta`  can ignore it in dev


danielroe:
Is the issue here that page metadata isn't respected for client-only pages?

huang-julien:
The issue is that `component` is the wrapper returned from `createClientPage` while `definePageMeta` is called from the page component. So it means that conditions in the if statement will never be true for client only pages

danielroe (approved):

danielroe:
Oh, just wondering why you linked https://github.com/nuxt/nuxt/pull/27839.

PR #28249, opened 2024-07-21
fix(schema): fix srcDir being overriden by default value

### LINKED ISSUE #28242: Nuxt `serverDir` config not working since v3.12.4
### Environment

- Node Version:     v20.15.1
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Package Manager:  pnpm@9.5.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

1. Open  https://stackblitz.com/edit/github-cyl8mm?file=nuxt.config.ts
2. Access the endpoint `/api/test`

### Describe the bug

Hello! I want to report an issue I encountered since `v3.12.4`, the `serverDir` config in `nuxt.config.ts` is not working with the same exact config used in `v3.12.3`, I'm not sure if it has something to do with the change made in https://github.com/nuxt/nuxt/commit/6abaae7fa1c8656ccd2be0cf5ca80cacc598ad57

### Additional context

Inside the reproduction `nuxt.config.ts`, I'm trying to set `myServer` folder as server directory instead of `server` 

```js
  serverDir: './myServer',
});
```

It has an API endpoint `/api/test` it should work when I access this endpoint in the app but it doesn't

When I lock the dependency back to `"nuxt": "3.12.3"` it works.

### Logs

_No response_

huang-julien:
### 🔗 Linked issue

fix #28242 

### 📚 Description

this PR fix schema to stop overridinguser value for srcDir by the default value

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -158,7 +158,7 @@ export default defineUntypedSchema({
     $resolve: async (val: string | undefined, get): Promise<string> => {
       const isV4 = ((await get('future') as Record<string, unknown>).compatibilityVersion === 4)
 
-      return val ? val : resolve(await get('rootDir') as string, isV4 ? 'server' : resolve(await get('srcDir') as string, 'server'))
+      return resolve(isV4 ? await get('rootDir') as string : await get('srcDir') as string, val ?? 'server')
     },
   },
 
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -158,7 +158,7 @@ export default defineUntypedSchema({
     $resolve: async (val: string | undefined, get): Promise<string> => {
       const isV4 = ((await get('future') as Record<string, unknown>).compatibilityVersion === 4)
 
-      return resolve(await get('rootDir') as string, (val || isV4) ? 'server' : resolve(await get('srcDir') as string, 'server'))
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/common.ts
```diff
@@ -158,7 +158,7 @@ export default defineUntypedSchema({
     $resolve: async (val: string | undefined, get): Promise<string> => {
       const isV4 = ((await get('future') as Record<string, unknown>).compatibilityVersion === 4)
 
-      return resolve(await get('rootDir') as string, (val || isV4) ? 'server' : resolve(await get('srcDir') as string, 'server'))
+      return val ? val : resolve(await get('rootDir') as string, isV4 ? 'server' : resolve(await get('srcDir') as string, 'server'))
```
danielroe, line 161: We should still resolve val if it is provided.

danielroe, line 161: What about this?

```suggestion
      return resolve(await get('rootDir') as string, isV4 ? val ?? 'server' : resolve(await get('srcDir') as string, val ?? 'server'))
```

huang-julien: sure !

danielroe (approved):

PR #28269, opened 2024-07-23
docs: use `UserConfig` rather than `ViteConfig` in signature

SergkeiM:

### DIFF
--- a/docs/3.api/5.kit/14.builder.md
+++ b/docs/3.api/5.kit/14.builder.md
@@ -112,7 +112,7 @@ Extends the Vite configuration. Callback function can be called multiple times, 
 ### Type
 
 ```ts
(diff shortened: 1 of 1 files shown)


docs/3.api/5.kit/14.builder.md
```diff
@@ -112,7 +112,7 @@ Extends the Vite configuration. Callback function can be called multiple times,
 ### Type
 
 ```ts
-function extendViteConfig (callback: ((config: ViteConfig) => void), options?: ExtendViteConfigOptions): void
+function extendViteConfig (callback: ((config: UserConfig) => void), options?: ExtendViteConfigOptions): void
```
danielroe, line 115: Although technically correct (as in, it's the `UserConfig` export from `vite`), this is less intuitive for users who need to know that this configuration is _vite_ configuration. Can you think of a better way of communicating it?

PR #28285, opened 2024-07-23
refactor(nuxt): update to vue v3.5 + native `useId`

### LINKED ISSUE #15086: ClientOnly: using async methods breaks defineExpose
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.13.0`
- Nitro Version:    `0.5.4`
- Package Manager:  `npm@8.1.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/p/github/nuxt/starter/csb-zd6kiy/draft/nostalgic-murdock

### Describe the bug

Implementing any async operations in a Client Only component will break `defineExpose` and the component instance won't have its exposed method available to parent anymore. 

E.g.

ChildComponent.vue
```vue
<script setup>
  const { data, pending, error, refresh } = await useFetch(
    () => "https://api.nuxtjs.dev/mountains"
  )

  const childMethod = () => {
     // do something
  }

  defineExpose({ childMethod })
</script>
```

ParentComponent.vue
```vue
<template>
  <ClientOnly>
    <ChildComponent ref="childRef" />
  </ClientOnly>
</template>

<script setup>
  const childRef = ref()

  // ...
  childRef.value.childMethod()
  // ERROR: childRef.value.childMethod is not a function
</script>
```

### Additional context
If you remove the `useFetch` call on `childComponent`, the exposed child method is available in the parent again 

### Logs

_No response_

### LINKED ISSUE #18405: data fetched again on client when inside an async component
### Environment

- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@8.19.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-jyws2q

### Describe the bug

In my nuxt3 application (CMS) i notice that sometime  query are run twice on SSR
I make a simple repro with the structure of the application

When query are run in deep level module loaded dynamicaly useAsyncData of the module run twice 
Two query are simulated the pageData query in page and todoQuery in last component 'Block1'

I made 3 test pages in repro 
- index.vue is like in my application  => the todo query is run twice [BUG]
- dynamic.vue skip PageContent component and call TplPage component dynamicaly => the todo query is run twice [BUG]
- direct.vue skip PageContent component and call TplPage component directly  => the todo query is run once

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #25131: useFetch returns null when used within layout and wrapped with ClientOnly
### Environment

- Operating System: Darwin
- Node Version:     v20.9.0
- Nuxt Version:     3.9.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1

### Reproduction

Not working (returns null): https://stackblitz.com/edit/github-uranmz?file=app.vue,layouts%2Fdefault.vue,components%2FTest.vue
Working (without layout): https://stackblitz.com/edit/github-uranmz-j92ydc?file=app.vue,components%2FTest.vue

### Describe the bug

There is a component that fetches data and it needs to be rendered on server or as client only. It depends on scenario.

Component is using useFetch composable to fetch data and data needs to be processed once fetched before returned to template. Everything works fine if data is fetched on server. But it returns null if component is wrapped with <ClientOnly> and used in layout.

If used without layout, it works fine.
The same problem exists when useFetch is used in onMounted hook. It returns null if layout is used.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #27378: `useId` hydration mismatch when used in a dynamic global component
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools, components
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-hqhma3?file=app.vue

### Describe the bug

When rendering a globally registered component like this: `<Component :is="resolveComponent('ComponentWithId')" />`, `useId` inside that component returns a different id on the server and client, causing a hydration mismatch. The difference is in the separator, `-` changes to `_`.

When rendering that same component normally like this `<ComponentWithId />`, there is no mismatch.

When rendering it like this, it also works:

```vue
<template>
  <Component :is="ComponentWithId" />
</template>

<script setup lang="ts">
const ComponentWithId = resolveComponent('ComponentWithId');
</script>

```

But when the component name is a variable, it breaks again:

```vue
<template>
  <Component :is="ComponentWithId" />
</template>

<script setup lang="ts">
const componentName = 'ComponentWithId';
const ComponentWithId = resolveComponent(componentName);
</script>

```

### Additional context

_No response_

### Logs

```shell-script
[Vue warn]: Hydration text content mismatch on <div data-v-inspector=​"components/​ComponentWithId.vue:​2:​3" data-n-ids=​"{"nQ7nnZaq0gB-0":​"nQ7nnZaq0gB-0"}​">​nQ7nnZaq0gB_0​</div>​ 
  - rendered on server: nQ7nnZaq0gB-0
  - expected on client: nQ7nnZaq0gB_0 
  at <ComponentWithId> 
  at <AsyncComponentWrapper> 
  at <App key=3 > 
  at <NuxtRoot>
```

danielroe:
resolves https://github.com/nuxt/nuxt/issues/15086
resolves https://github.com/nuxt/nuxt/issues/18405
resolves https://github.com/nuxt/nuxt/issues/25131
resolves https://github.com/nuxt/nuxt/issues/27378

### 📚 Description

This updates to the latest v3.5.0 alpha with a built-in first-party `useId`: https://github.com/vuejs/core/pull/11404.

It currently highlights two issues:

- [x] ~~this does not seem to be working within unit tests~~
   this is likely because each `mount` creates a fresh vue app and the id is reset - so we are simply removing our unit tests from the fixture
- [x] ~~this doesn't seem to work when hydrating async components, likely because of https://github.com/vuejs/core/issues/6638 - which is a significant bug which affects us quite a lot.~~ resolved! ✅


DamianGlowala:
@danielroe small note: we currently declare a type in Nuxt which can be now imported from `vue` and kept in sync, see https://github.com/vuejs/core/pull/9563

packages/nuxt/src/app/composables/id.ts
```diff
@@ -1,60 +1,3 @@
...
-
-    if (import.meta.dev && instance.vnode.type && typeof instance.vnode.type === 'object' && 'inheritAttrs' in instance.vnode.type && instance.vnode.type.inheritAttrs === false) {
-      console.warn('[nuxt] `useId` might not work correctly with components that have `inheritAttrs: false`.')
-    }
-  }
-
-  // pure client-side ids, avoiding potential collision with server-side ids
-  return key + '_' + nuxtApp._genId++
-}
+export const useId = _useId
```
huang-julien, line 3: Why don't we just delete the file and add the auto import from vue ?

HigherOrderLogic: I think it's for backward compatiblity. Some may have disabled auto import and do `import { useId } from '#imports';`

huang-julien: I mean, we could completely remove it from nuxt. And add useId to auto imports and #imports. This shouldn't be a breaking change since the behaviour wouldn't change

danielroe: I did switch the auto imports in this PR. The reason the export remains is in case someone directly imports from #app/composables/id.

dr46ul:
@danielroe is there a plan to support the `app.config.idPrefix` in Nuxt as well? I didn't see any mention to that in the docs.

PR #28333, opened 2024-07-27
fix(nuxt): include correct type declaration for `vue-router`

### LINKED ISSUE #28218: Wrong auto import types of `vue-router` API
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

```ts
// auto import wrong type of onBeforeRouteUpdate
onBeforeRouteUpdate((to) => {
  console.log(to);
});
```

https://stackblitz.com/edit/github-bjyhxs?file=app.vue,pages/index.vue,package.json,tsconfig.json

```sh
npm run typecheck
```

### Describe the bug

```
pages/index.vue:2:22 - error TS7006: Parameter 'to' implicitly has an 'any' type.
2 onBeforeRouteUpdate((to) => {
```

The reason is the below `vue-router` API inside `.nuxt/types/imports.d.ts` imported from `../../node_modules/vue-router/auto-routes` which does not exports the `useLink`, `onBeforeRouteLeave`, `onBeforeRouteUpdate`

```ts
readonly useLink: UnwrapRef<typeof import('../../node_modules/vue-router/auto-routes')['useLink']>
readonly onBeforeRouteLeave: UnwrapRef<typeof import('../../node_modules/vue-router/auto-routes')['onBeforeRouteLeave']>
readonly onBeforeRouteUpdate: UnwrapRef<typeof import('../../node_modules/vue-router/auto-routes')['onBeforeRouteUpdate']>
```

### Additional context

in Nuxt v3.12.2 has no issue and the below is the `imports.d.ts` of it.
```ts
const onBeforeRouteLeave: typeof import('../vue-router-stub')['onBeforeRouteLeave']
const onBeforeRouteUpdate: typeof import('../vue-router-stub')['onBeforeRouteUpdate']
const useLink: typeof import('../vue-router-stub')['useLink']
```

#### Temporary fix

Add `vue-router` imports manually
```diff
+ import { onBeforeRouteUpdate } from 'vue-router'
onBeforeRouteUpdate((to) => {
  console.log(to);
});
```

### Logs

_No response_

babs20:
### 🔗 Linked issue

resolves #28218 

### 📚 Description

This updates the module specifier to be more specific when importing the types for `vue-router`. Currently it is resolving `../../node_modules/vue-router/auto-routes` instead of `../../node_modules/vue-router`. This happens because `mlly` finds the `./auto-routes` export first and uses that instead of the `./` which is [last in the exports object](https://github.com/vuejs/router/blob/77d7217be5295a31cf57771a0b05cd363b30233d/packages/router/package.json#L9-L60).

This fixes the type imports for:
- `onBeforeRouteLeave`
- `onBeforeRouteUpdate`
- `useLink`

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -233,7 +233,7 @@ export default defineNuxtModule({
     nuxt.hook('imports:sources', (sources) => {
       const routerImports = sources.find(s => s.from === '#app/composables/router' && s.imports.includes('onBeforeRouteLeave'))
       if (routerImports) {
-        routerImports.from = 'vue-router'
+        routerImports.from = 'vue-router/dist/vue-router.d.ts'
       }
     })
 
@@ -336,7 +336,7 @@ export default defineNuxtModule({
     nuxt.hook('imports:extend', (imports) => {
       imports.push(
         { name: 'definePageMeta', as: 'definePageMeta', from: resolve(runtimeDir, 'composables') },
-        { name: 'useLink', as: 'useLink', from: 'vue-router' },
+        { name: 'useLink', as: 'useLink', from: 'vue-router/dist/vue-router.d.ts' },
       )
       if (nuxt.options.experimental.inlineRouteRules) {
         imports.push({ name: 'defineRouteRules', as: 'defineRouteRules', from: resolve(runtimeDir, 'composables') })


packages/nuxt/src/pages/module.ts
```diff
@@ -233,7 +233,7 @@ export default defineNuxtModule({
     nuxt.hook('imports:sources', (sources) => {
       const routerImports = sources.find(s => s.from === '#app/composables/router' && s.imports.includes('onBeforeRouteLeave'))
       if (routerImports) {
-        routerImports.from = 'vue-router'
+        routerImports.from = 'vue-router/dist/vue-router.d.ts'
```
danielroe, line 236: This is the actual place that these imports should be resolved from, not the declaration file - and will break the usage of these composables.

Instead, the issue is likely with how we are handling `typescript.hoist` elsewhere in the codebase.

babs20: Sounds good, I'll check how that is being handled!

serkodev:
Besides updating Nuxt, I think `vue-router` should reorder the `export` in `packages.json` because the key order is significant. After reordering the `export`, Nuxt works as expected too.

Created PR including details: https://github.com/vuejs/router/pull/2323

babs20:
@serkodev Thanks for making that PR on vue-router. As you said, that will fix the issue we are seeing here within Nuxt once it is published.

After looking into the typescript hoisting, everything there looks correct, as far as I can tell. I think we can probably close this PR and wait for the vue-router changes.

danielroe:
Thank you for your quick action and investigation ❤️ 

This might indicate a bug in `mlly` in that case if it is resolving the wrong subpath... Do you think you could create a reproduction using just `mlly` and `vue-router`? 🙏

babs20:
Yeah definitely! [Here is one](https://stackblitz.com/edit/stackblitz-starters-w3d5xq?file=index.js) comparing `vue-router@4.4.2` and `vue-router@4.4.3` using `mlly`.

There is [this issue](https://github.com/unjs/mlly/issues/218) on `mlly` which is closely related to the behavior we are seeing here. I was planning on looking into that issue further and possibly contributing to see what adding the `conditions` option would look like in practice.

PR #28334, opened 2024-07-27
perf(nuxt): prevent converting client only component setup to Promise

### LINKED ISSUE #27999: Client only component not mounted after nextTick
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-bf56vz

### Describe the bug

According to the [documentation](https://nuxt.com/docs/guide/directory-structure/components#client-components), we can access the rendered template after `await nextTick()` in the callback of the `onMounted` hook. But I found if the parent component of the client-only component is mounted after setting `v-if` as `true`. We can no longer get the ref template after the first `await nextTick()`.
```vue
<script setup>
const mc = ref()

onMounted(async () => {
  await nextTick();
  console.log(mc.value.mounted$);
  await nextTick();
  console.log(mc.value.mounted$);
  await nextTick();
  console.log(mc.value.mounted$);
  await nextTick();
  console.log(mc.value.mounted$);
});
</script>
<template>
    <MyClientComponent ref="mc" />
</template>
```
will log as the following:
```
undefined
undefined
false
true
```

I don't know whether it is an expected behavior. If so, how do we know when we can get the ref template of the client-only component in the `onMounted` hook?

### Additional context

1. Enter the reproduction
2. Press the `Click to display` button
3. See the log in console

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
closes #27999

### 📚 Description
Hey :wave: 

This PR improve components from `createClientComponent` if they are mounted after hydration. This bypass all code run specifically for hydration. Meaning the component will directly be returned if nuxt has already been hydrated.

This PR also treeshake code that is not required server side


danielroe (commented):
is there a way to add a regression test for this one?

huang-julien:
@danielroe so with this fix, we still need to wait exactly 3 ticks until the client component is mounted.  Should we instead advise users to use `@vue:mounted` instead ?

danielroe (approved):
❤️

PR #28349, opened 2024-07-30
fix(nuxt): use static import for hmr `updateAppConfig`

### LINKED ISSUE #698: fix: DevTools cannot be showed in Chrome and Firefox on Nuxt v4
### 🐛 The bug

I am using the nightly version of Nuxt ([4.0.0-28693174.8aefb554](https://www.npmjs.com/package/nuxt-nightly/v/4.0.0-28693174.8aefb554)). I found Nuxt DevTools (no matter the nightly or `@nuxt/devtools@1.3.9`) does not work on Chrome 126 and Firefox 129.0b6 (Attachment 2). Notably, it works on Safari 17.5 (Attachment 1). It used to work on Nuxt v3.

image
image

When adding breakpoints to the code related to Nuxt DevTools, I found Firefox (and Chrome) indeed ran `import("./view/client")` but `setupDevToolsClient` was never called, as the Attachment 3 shown. Safari ran as expected (Attachment 4). Not sure why.

https://github.com/user-attachments/assets/df0c3b68-3e77-4791-ba55-f837bfe0efb5

https://github.com/user-attachments/assets/0ad4a9cd-7570-4cb4-b62f-df250b996a32

### 🛠️ To reproduce

https://github.com/pan93412/nuxt-4-devtools-regression

### 🌈 Expected behavior

The Nuxt DevTools should be able to show on Chrome, Firefox, and Safari.

### ℹ️ Additional context

I am using macOS 14.5 (23F79) if this matters.

### LINKED ISSUE #28348: Circular import in #app/config makes some plugins failed to load
### Environment

```
- Operating System: Darwin
- Node Version:     v20.14.0
- Nuxt Version:     4.0.0-28696289.ed54884c
- CLI Version:      3.13.0-1719571044.bca75ab
- Nitro Version:    -
- Package Manager:  pnpm@9.4.0
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

> For details, check the parent issue: https://github.com/nuxt/devtools/issues/698

**Reproducible repository:** https://github.com/pan93412/nuxt-4-devtools-regression

**Steps:**

1. Clone the repository
2. `pnpm install && pnpm dev`
3. Open the page in Chrome or Firefox, the DevTools cannot show up.

### Describe the bug

> Check this parent issue for details: https://github.com/nuxt/devtools/issues/698#issuecomment-2244969473

After the final investigation, I can conclude this bug is a due to the circular import.

`#app/config` contains this import [^1]:

```ts
import { reactive } from 'vue'
import { klona } from 'klona'
import type { AppConfig } from 'nuxt/schema'
import { useNuxtApp } from './nuxt'
import __appConfig from '#build/app.config.mjs'  // oops
```

`#build/app.config.mjs` is basically the content I mentioned above [^2]:

```js
import { defuFn } from 'defu'

const inlineConfig = ${JSON.stringify(nuxt.options.appConfig, null, 2)}

/** client **/
// Vite - webpack is handled directly in #app/config
if (import.meta.dev && !import.meta.nitro && import.meta.hot) {
  const { updateAppConfig } = await import('#app/config')  // oops
  import.meta.hot.accept((newModule) => {
    updateAppConfig(newModule.default)
  })
}
/** client-end **/
```

[^1]: telegram-cloud-photo-size-5-6125390261027717598-y
[^2]: telegram-cloud-photo-size-5-6125390261027717599-y

### Additional context

My original fix idea[^1] does not seem like a good approach. It runs `import('#app/config')` regardless of whether we are in a development environment or not. Instead of refactoring the entire module, the best idea would be to add a counter to prevent circular imports.

[^1]: I can open a PR to revert the await import('#app/config') to import { updateAppConfig } from '#app/config'. [src](https://github.com/nuxt/devtools/issues/698#issuecomment-2244969473)

### Logs

_No response_

pan93412:
### 🔗 Linked issue

Fix #28348
Fix https://github.com/nuxt/devtools/issues/698

### 📚 Description

Importing `#app/config` introduces a circular import (as mentioned in #28348). Since we only use the client part of [`updateAppConfig`](https://github.com/nuxt/nuxt/blob/68b720538c46dfffbd62b18727b91c21f663a34f/packages/nuxt/src/app/config.ts#L39-L55), we can write a simple client-side `updateAppConfig` in `#build/config` to prevent the dependency on `#app/config`.

The entire code block should only run on the client side (wrapped in `/** client **/`), so I have removed the server-side part and simplified the app config logic.

```js
export function useAppConfig (): AppConfig {
  const nuxtApp = useNuxtApp()
  if (!nuxtApp._appConfig) {
    // klona is unneeded in client side
    nuxtApp._appConfig = (import.meta.server ? klona(__appConfig) : reactive(__appConfig)) as AppConfig
  }
  return nuxtApp._appConfig
}

export function updateAppConfig (appConfig: DeepPartial<AppConfig>) {
  const _appConfig = useAppConfig()
  deepAssign(_appConfig, appConfig)
}
```

`deepAssign` might be the subset of [defuFn](https://github.com/unjs/defu), so I leverage it directly.

This PR makes Nuxt DevTools work again: https://screen.pan93.com/m6NL4mj8 (not sure why I cannot upload picture to GitHub)


packages/nuxt/src/core/templates.ts
```diff
@@ -343,9 +343,19 @@ const inlineConfig = ${JSON.stringify(nuxt.options.appConfig, null, 2)}
 /** client **/
 // Vite - webpack is handled directly in #app/config
 if (import.meta.dev && !import.meta.nitro && import.meta.hot) {
-  const { updateAppConfig } = await import('#app/config')
+  const { reactive } = await import('vue')
```
danielroe, line 346: is there a reason you're using a dynamic import here from `vue`? I think you can directly import above and have it tree-shaken out in production mode...

pan93412: I tried to match my code's behavior to that of the original. I’d be happy to make it a static import if there are no side effects from this Vue import.

pan93412:
Not sure if there is a good way to test this PR.

Seems like `@nuxt/devtools` did not contain a E2E test.

danielroe (approved):

danielroe (approved):
I've taken the liberty of making a simpler fix, which seems to work for me - let me know if you think it doesn't resolve the issue 🙏

pan93412:
> I've taken the liberty of making a simpler fix, which seems to work for me - let me know if you think it doesn't resolve the issue 🙏 

I am not sure if the circular import is a good idea. This approach was working on my patched Nuxt though, but I afraid it may introduce other issues later since the root cause still exists.

danielroe:
It was working previously and I think there's no issue with circular _static_ imports as they can be resolved correctly by the browser.

PR #28389, opened 2024-08-02
docs: fix `options` type in custom useFetch recipe

MikeBellika:
### 📚 Description

When using the existing recipe, the return type of data doesn't include `null` (or `undefined` in 4.0)
```typescript
import type { UseFetchOptions } from "nuxt/app"

export function useAPI<T>(
  url: string | (() => string),
  options?: Omit<UseFetchOptions<T>, "default"> & { default: () => T | Ref<T> },
) {
  return useFetch(url, {
    ...options,
    $fetch: useNuxtApp().$api,
  })
}

type MyReturnType = { data: string }
const { data: a } = await useFetch<MyReturnType>("/api/url")
//                    ^? globalThis.Ref<MyReturnType | null>
const { data: b } = await useAPI<MyReturnType>("/api/url")
//                    ^? globalThis.Ref<MyReturnType>
```
With this PR, the return type will be the same as for `useFetch`.
See also for some discussion https://discord.com/channels/473401852243869706/1042756264205230081/threads/1268846009270800415


danielroe:
I'm not sure this is in error. As written, the type _requires_ a `default` function, and with a default function the return value should not be `null`/`undefined`.

MikeBellika:
> I'm not sure this is in error. As written, the type _requires_ a `default` function, and with a default function the return value should not be `null`/`undefined`.

Is there a reason why you would want to require a `default` function, when implementing your own `useFetch`? In my case I just wanted to wrap `useFetch` and add some error handling. I'm not interested in always supplying a `default` function.

Also a bit further down in the recipe, it's used like this:
```typescript
<script setup>
const { data: modules } = await useAPI('/modules')
</script>
```
So I think we should either update the `useAPI` function or the example usage.

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -79,7 +79,7 @@ import type { UseFetchOptions } from 'nuxt/app'
 
 export function useAPI<T>(
   url: string | (() => string),
-  options: Omit<UseFetchOptions<T>, 'default'> & { default: () => T | Ref<T> },
+  options: UseFetchOptions<T>,
```
danielroe, line 82:
```suggestion
  options?: UseFetchOptions<T>,
```

danielroe (approved):

danielroe:
Thank you! ❤️

PR #28402, opened 2024-08-04
docs: fix TypeScript errors for examples

AuroraTea:
### 📚 Description

1. `RegExp.text()` needs a `string` but `NuxtPage.file` is a `string | undefined`, and `undefined` will convert to `"undefined"`.
2. Explicitly specify the imports.
3. Add type defining.
4. Change language tags of fenced code blocks to ts.


danielroe:
What was the issue?

PR #28403, opened 2024-08-04
docs: fix TypeScript errors for examples

AuroraTea:
### 📚 Description

1. `RegExp.text()` needs a `string` but `NuxtPage.file` is a `string | undefined`, and `undefined` will convert to `"undefined"`.
2. Explicitly specify the imports.
3. Add type defining.
4. Change language tags of fenced code blocks to ts.

### DIFF
--- a/docs/2.guide/4.recipes/1.custom-routing.md
+++ b/docs/2.guide/4.recipes/1.custom-routing.md
@@ -55,12 +55,10 @@ export default defineNuxtConfig({
       function removePagesMatching (pattern: RegExp, pages: NuxtPage[] = []) {
         const pagesToRemove: NuxtPage[] = []
         for (const page of pages) {
-          if (page.file !== undefined) {
-            if (page.file && pattern.test(page.file)) {
-              pagesToRemove.push(page)
-            } else {
-              removePagesMatching(pattern, page.children)
-            }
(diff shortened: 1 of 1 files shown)


docs/2.guide/4.recipes/1.custom-routing.md
```diff
@@ -51,12 +53,14 @@ export default defineNuxtConfig({
...
       function removePagesMatching (pattern: RegExp, pages: NuxtPage[] = []) {
-        const pagesToRemove = []
+        const pagesToRemove: NuxtPage[] = []
         for (const page of pages) {
-          if (pattern.test(page.file)) {
-            pagesToRemove.push(page)
-          } else {
-            removePagesMatching(pattern, page.children)
+          if (page.file !== undefined) {
+            if (pattern.test(page.file)) {
```
danielroe, line 59: I think better to do something like:
```suggestion
            if (page.file && pattern.test(page.file)) {
```

As we still want to run the else block, even if file is undefined

AuroraTea: This is a recursion in else; Can't believe I didn't notice yesterday. 😂

danielroe (approved):
Thank you! ❤️

PR #28446, opened 2024-08-07
fix(nuxt): augment `@vue/runtime-core` and `@vue/runtime-dom`

### LINKED ISSUE #28373: Property '$route' does not exist on type
### Environment

- Operating System: Darwin
- Node Version:     v20.12.1
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    -
- Package Manager:  pnpm@9.1.4
- Builder:          -
- User Config:      compatibilityDate, devServer, devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/pdsuwwz/nuxt3-minimal-reproduction

### Describe the bug

Within the template of a Vue component, using `$route`, `$router` will report a type error.

I tried restarting vscode and the extension and still got an error.

image

image

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #28440: error with vue-tsc 2.0.29
### Environment

- Operating System: Linux
- Node Version:     v18.17.0
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Package Manager:  npm@9.6.6
- Builder:          -
- User Config:      devtools, typescript, vue, modules, i18n, quasar, site, nitro, routeRules, sitemap, runtimeConfig, css, security
- Runtime Modules:  @nuxtjs/i18n@8.3.1, @pinia/nuxt@0.5.1, @pinia-plugin-persistedstate/nuxt@1.2.0, @nuxtjs/seo@2.0.0-rc.10, nuxt-quasar-ui@2.1.2, nuxt-security@2.0.0-rc.9
- Build Modules:    -

### Reproduction

it looks like auto import is not working

### Describe the bug

if i upgrade vue-tsc version,

i can't npm run build or dev

because I encounter errors with composable... etc.

`components/recruit/Table.vue:29:11 - error TS2339: Property 'convertUnixtoYMD' does not exist on type 'CreateComponentPublicInstance<Readonly<ExtractPropTypes<__VLS_TypePropsToOption<{ searchStr: string; data: RecruitList[]; }>>>, {}, unknown, ... 16 more ..., {}>'.

29           convertUnixtoYMD(item.endTime)
             ~~~~~~~~~~~~~~~~
`
`components/solution/Intro.vue:40:21 - error TS2339: Property 'navigateTo' does not exist on type 'CreateComponentPublicInstance<Readonly<ExtractPropTypes<__VLS_TypePropsToOption<{ data: Res_Solution; }>>>, { t: ComposerTranslation<{ [x: string]: LocaleMessage<VueMessageType>; }, ... 4 more ..., string>; }, ... 17 more ..., {}>'.

40             @click="navigateTo({ path: '/customer/brochure' })"
                       ~~~~~~~~~~
`

it looks like auto import is not working


### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/28373, resolves https://github.com/nuxt/nuxt/issues/28440

### 📚 Description

Currently we need to augment `vue`, `@vue/runtime-dom` and `@vue/runtime-core` - or it will be possible that a library which augments only one of these will 'break' type augmentation.

However, it might be we have an opportunity across the ecosystem to migrate to augmenting only one of these (ideally just `vue`: https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties). Removing these duplicate augmentations would _also_ fix the issue in a minimal install of Nuxt but likely needs to be updated in other libraries/projects.

cc: @johnsoncodehk, @posva

### DIFF
--- a/packages/nuxt/src/app/types/augments.d.ts
+++ b/packages/nuxt/src/app/types/augments.d.ts
@@ -51,3 +51,23 @@ declare module 'vue' {
     head?(nuxtApp: NuxtApp): UseHeadInput
   }
 }
+
+declare module '@vue/runtime-core' {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface App<HostElement> {
+    $nuxt: NuxtApp
+  }
+  interface ComponentCustomProperties {
+    $nuxt: NuxtApp
+  }
+}
+
+declare module '@vue/runtime-dom' {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface App<HostElement> {
+    $nuxt: NuxtApp
+  }
+  interface ComponentCustomProperties {
+    $nuxt: NuxtApp
+  }
+}
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -130,6 +130,14 @@ declare module '#app' {
   }
 }
 
+declare module '@vue/runtime-core' {
+  interface ComponentCustomProperties extends NuxtAppInjections { }
+}
+
+declare module '@vue/runtime-dom' {
+  interface ComponentCustomProperties extends NuxtAppInjections { }
+}
+
 declare module 'vue' {
   interface ComponentCustomProperties extends NuxtAppInjections { }
 }


posva:
In the long run, it will be easier for library maintainers if they can just `declare module 'vue'` rather than using three different module augmentations that are just copy-paste.

I don't know if this also affects end users: does their code break if they use the currently documented solution of `declare module 'vue'` in a project that has a library installed that uses a different module augmentation?

PerryProjects (approved):
I tested it local in VSCode and in Stackblitz and it seams to work for now.

Anoesj:
Found this because of #28373. I see this is merged into main, great to see a solution has been found, thanks! When something is merged into main, does that mean this is being released with the next release, so 3.12.5 / 3.13.0? I don't know about the branching/versioning strategy, so probably asking a stupid question here 😇

daniandl:
> Found this because of #28373. I see this is merged into main, great to see a solution has been found, thanks! When something is merged into main, does that mean this is being released with the next release, so 3.12.5 / 3.13.0? I don't know about the branching/versioning strategy, so probably asking a stupid question here 😇

Daniel confirmed on Discord it would be coming in 3.x

image

dargmuesli:
Yes, it'll be part of the next release. You can also use the [Nightly Release Channel](https://nuxt.com/docs/guide/going-further/nightly-release-channel) to have the fix included right now.

tjbp:
I've just tested this with the Nightly Release Channel (`npm:nuxt-nightly@3x`) and unfortunately still had the type error for auto-imports used in the template (problem doesn't occur if the auto-import is referenced in the script block).

danielroe:
@tjbp  would you provide a reproduction? This can also be caused by other libraries.

tjbp:
@danielroe I've been trying to but have failed - so it's probably as you say. Sorry don't want to derail a closed PR but would I probably be looking for other libraries that also augment Vue's types in a similar way?

danielroe:
Exactly!

I've raised a handful of PRs (all linked to https://github.com/nuxt/nuxt/pull/28542) that swap out `declare module '@vue/runtime-core'` with `declare module 'vue'`. If you spot more out there, I think it would be very helpful to submit PRs...

PR #28449, opened 2024-08-07
fix(nuxt): updated renamed stub composables from `nuxt/scripts`

Archetipo95:
Renamed `useElementScriptTrigger` to `useScriptTriggerElement` as it changed in https://github.com/nuxt/scripts/releases/tag/v0.6.0


danielroe (approved):
Thank you ❤️

PR #28472, opened 2024-08-08
perf(nuxt): avoid multiple calls to `getCachedData`

### LINKED ISSUE #27200: getCachedData causing repeated requests
### Environment

- Operating System: Darwin
- Node Version:     v20.11.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Package Manager:  yarn@1.22.21
- Builder:          -
- User Config:      app, devtools, experimental, typescript, modules, css, runtimeConfig, components, googleFonts, i18n, alias, nitro, build
- Runtime Modules:  @pinia/nuxt@0.4.11, @nuxtjs/tailwindcss@6.12.0, @nuxtjs/google-fonts@3.2.0, @nuxt/image@1.5.0, @vueuse/nuxt@10.9.0, @nuxtjs/i18n@8.3.0, dayjs-nuxt@2.1.9
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-fu2dhl?file=composables%2FapiFetch.ts

### Describe the bug

I've created a composable around `useFetch` to simplify the usage of `getCachedData`. However I've noticed that there was a repeated request sent if  getCachedData is passed as the options for `useFetch`. 

If the getCachedData is not passed as an option, the request is only sent once. 

### Additional context

I initially stumble upon this bug after noticing that a weird behavior whereby if i refresh the page it would display my data as usual and the data request is fetched on the client once again causing a flicker when using v-if / v-else

### Logs

```shell-script
------------------
With getCachedData
------------------

hit apiFetch
Cache disabled || invalidateCache /api/test false false
Cache disabled || invalidateCache /api/test false false
Cache disabled || invalidateCache /api/test false false
hit server api
hit apiFetch
Cache disabled || invalidateCache /api/test false false
Cache disabled || invalidateCache /api/test false false
Cache disabled || invalidateCache /api/test false false
hit server api
hit server api


---------------------
Without getCachedData
---------------------
hit apiFetch
hit server api
```

Martichou:
### 🔗 Linked issue

Resolves #27200

### 📚 Description

Issue is that getCachedData is called multiple time each time (it's clearly visible in the code). So this PR attempt to resolve that issue.

Instead of having `hasCachedData` calling `options.getCachedData!` and checking if it's `undefined` or not, we do it once and sort of cache to be reused when needed. But we don't forget that when execute or refresh is called, we should check the cache again to avoid returning a value not present anymore EXCEPT when doing the `_initial` call.


danielroe:
@Martichou Updated to use `nuxtApp._asyncData[key]` instead of a local variable, but otherwise largely the same. Does it look good to you?

Martichou:
@danielroe looks good to me, thanks for taking a look at it :)

danielroe (approved):

Issue #28483, opened 2024-08-09
Hydration mismatch when using `Teleport` in combination with `useHead` scripts at `'bodyOpen'` position


CernyMatej:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      compatibilityDate, devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-lkzbwy?file=components%2FModal.vue,app.vue

### Describe the bug

When adding elements like script tags using `useHead` with the `tagPosition: 'bodyOpen'` option, `<Teleport to="body">` results in a hydration mismatch.

<img width="379" alt="image" src="https://github.com/user-attachments/assets/bd4da22b-65d3-4e80-aa69-730df35d1c2a">

danielroe:
We likely need to update the docs, but would you try now targeting it to `#teleports` and seeing if that works for you?

CernyMatej:
@danielroe 
Thank you! Seems to have done the trick.

I made the following PRs to update the docs. Let me know if there's anything I need to fix:
https://github.com/nuxt/examples/pull/82
https://github.com/nuxt/nuxt/pull/28489
https://github.com/nuxt/nuxt/pull/28490

PR #28490, opened 2024-08-09
test: use '#teleports' target instead of 'body'

cernymatej:
### 🔗 Linked issue
https://github.com/nuxt/nuxt/issues/28483

### 📚 Description

Since the recommended teleport target is now `'#teleports'`, I figured it would be a good idea to update this test fixture as well. Not sure if it won't break something, though 🙈

### DIFF
--- a/test/fixtures/basic/pages/teleport.vue
+++ b/test/fixtures/basic/pages/teleport.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <teleport to="body">
+    <teleport to="#teleports">
       <div>Teleport</div>
     </teleport>
     <h1>Normal content</h1>


danielroe:
We likely also need to update the test assertions.

CernyMatej:
Yeah

I'm trying to figure out how to get rid of a playwright error to run the tests locally so that I can verify it after updating them.
image

HigherOrderLogic:
You need to run `npx playwright install`.

CernyMatej:
@danielroe 
So, for the basic tests, the `teleportId` is set to `'nuxt-teleport'`:
https://github.com/nuxt/nuxt/blob/52faf52cdef63835d02e6da6b30c34c4f6ad64cd/test/fixtures/basic/nuxt.config.ts#L19

There is already a test case for it:
https://github.com/nuxt/nuxt/blob/52faf52cdef63835d02e6da6b30c34c4f6ad64cd/test/basic.test.ts#L2700

Is it worth keeping the `<teleport to="body">` test, given that this is a use case that is not recommended?  
Should the default case be tested as well? If so, should I test it under a separate Nuxt instance, since the one in the basic tests has a custom `teleportId` & `teleportTag` ?

PR #28508, opened 2024-08-12
fix(vite): log transformation result when a PARSE_ERROR happens

### LINKED ISSUE #25525: Log the whole file when a transformation issue happens
### Describe the feature

related to https://github.com/nuxt/nuxt/issues/25513

When a transform issue happens within a virtual file, it's almost impossible to get the cause of the issue. We should fidn a way to provide a log of the whole content of the file or allow the user to acces to the virtual file

### Additional information

### Final checks

huang-julien:
### 🔗 Linked issue

fix #25525 

### 📚 Description

This PR log the transform result of a file if there is a JS parsing issue following all the trasformation of a file. This will greatly help developers when an issue happens due to transfrormation plugins in vite since their own file could be 100% fine.

### DIFF
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -141,7 +141,7 @@ function createViteNodeApp (ctx: ViteBuildContext, invalidates: Set<string> = ne
         web: [],
       },
     })
-    
+
     const isExternal = createIsExternal(viteServer, ctx.nuxt.options.rootDir, ctx.nuxt.options.modulesDir)
     node.shouldExternalize = async (id: string) => {
       const result = await isExternal(id)
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -15,6 +15,8 @@ import { isCSS } from './utils'
 import { createIsExternal } from './utils/external'
 import { transpile } from './utils/transpile'
 
+const NEED_FRAME_ERRORS = ['PARSE_ERROR']
+
 // TODO: Remove this in favor of registerViteNodeMiddleware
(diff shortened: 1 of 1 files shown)


packages/vite/src/vite-node.ts
```diff
@@ -156,13 +159,18 @@ function createViteNodeApp (ctx: ViteBuildContext, invalidates: Set<string> = ne
       if (isAbsolute(moduleId) && !isFileServingAllowed(moduleId, viteServer)) {
         throw createError({ statusCode: 403 /* Restricted */ })
       }
-      const module = await node.fetchModule(moduleId).catch((err) => {
+      const module = await node.fetchModule(moduleId).catch(async (err) => {
         const errorData = {
           code: 'VITE_ERROR',
           id: moduleId,
           stack: '',
+          message: err.message,
```
danielroe, line 167: does this make a difference?

packages/vite/src/vite-node.ts
```diff
@@ -156,13 +159,18 @@ function createViteNodeApp (ctx: ViteBuildContext, invalidates: Set<string> = ne
...
           ...err,
         }
+
+        if (!errorData.frame && NEED_FRAME_ERRORS.includes(errorData.code)) {
```
danielroe, line 171: I think we can directly compare for a slight perf benefit (and refactor to Set if we need to check more error codes)

packages/vite/src/vite-node.ts
huang-julien: oops, that's something i need to remove

danielroe (approved):

PR #28517, opened 2024-08-13
fix(nuxt): add reason when aborting request in `useFetch`

MikeBellika:
### 📚 Description
At the moment aborted requests show up in error tracking tools as `AbortError: signal is aborted without reason`.
This PR will make the reason more clear, helping people decide if they need to take action.


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -152,7 +152,7 @@ export function useFetch<
   let controller: AbortController
 
   const asyncData = useAsyncData<_ResT, ErrorT, DataT, PickKeys, DefaultT>(key, () => {
-    controller?.abort?.()
+    controller?.abort?.('Duplicate request detected and aborted.')
```
danielroe, line 155:
```suggestion
    controller?.abort?.('Request aborted as another request to the same endpoint was initiated.')
```

danielroe (approved):

danielroe:
Thank you! ❤️

PR #28520, opened 2024-08-13
docs: improve typing of default exports

typed-sigterm:
### 📚 Description

`satisfies` ensures type conformity without altering the object's type, better than `<T>` casting.


danielroe (approved):
Thanks! 🙏

PR #28521, opened 2024-08-13
feat(vite): preserve vite sourcemaps for nitro build

### LINKED ISSUE #28506: Useless sourcemaps generated during production build
### Environment

- Operating System: Darwin
- Node Version:     v20.14.0
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/dovca/nuxt-sourcemaps-test

```
npm install
npm build
node --enable-source-maps .output/server/index.mjs
```

Visit `http://localhost:3000`, the page should throw an error and log it in the server console.

### Describe the bug

During production build, the server sourcemaps are not pointing to the original `.vue` or `.ts` source files. Instead, error stack traces show the generated server chunks.

### Additional context

This is a sourcemap file (`.output/server/chunks/build/index-O15BBwZ3.mjs.map`):

```
{"version":3,"file":"index-O15BBwZ3.mjs","sources":["../../../../.nuxt/dist/server/_nuxt/index-O15BBwZ3.js"],"sourcesContent":null,"names":[],"mappings":" ... "}
```

In the `sources` array, we can see the generated nuxt server code being referenced, not the actual `index.vue` source file.

### Logs

```shell-script
Server error being logged: 


H3Error: TEST
    at throwTest (/Users/david/webs/sandbox/nuxt-sourcemap-test/.nuxt/dist/server/_nuxt/index-O15BBwZ3.js:3:9)
    ... 8 lines matching cause stack trace ...
    at renderComponentVNode (/Users/david/webs/sandbox/nuxt-sourcemap-test/.output/server/node_modules/@vue/server-renderer/dist/server-renderer.cjs.prod.js:386:12) {
  cause: Error: TEST
      at throwTest (/Users/david/webs/sandbox/nuxt-sourcemap-test/.nuxt/dist/server/_nuxt/index-O15BBwZ3.js:3:9)
      at setup (/Users/david/webs/sandbox/nuxt-sourcemap-test/.nuxt/dist/server/_nuxt/index-O15BBwZ3.js:9:15)

... and more unrelated rows

```

I would expect the location of `throwTest` to be `/Users/david/webs/sandbox/nuxt-sourcemap-test/throwTest.ts`.
```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/28506

### 📚 Description

We have two builds for the Nuxt server - a Vite/webpack build and then a Nitro build. This means that server sourcemaps end up pointing to the intermediate Vite build rather than the original source file. As both Vite + Nitro are based on rollup at the moment, we can preserve sourcemaps between the builds in the form of JSON, without needing to use something like [rollup-plugin-sourcemaps](https://github.com/maxdavidson/rollup-plugin-sourcemaps) (which is both outdated and has a greater performance cost - see https://github.com/rollup/rollup/issues/4532).

We write the sourcemaps to disk rather than persisting them in memory to avoid excessive memory consumption.

### DIFF
--- a/packages/vite/src/plugins/nitro-sourcemap.ts
+++ b/packages/vite/src/plugins/nitro-sourcemap.ts
@@ -1,5 +1,5 @@
 import { mkdir, readFile, writeFile } from 'node:fs/promises'
-import { dirname, join, resolve } from 'pathe'
+import { dirname, resolve } from 'pathe'
 
 import type { Plugin as RollupPlugin } from 'rollup'
 import type { Plugin as VitePlugin } from 'vite'
@@ -17,7 +17,7 @@ export const createSourcemapPreserver = () => {
       for (const chunk of Object.values(bundle)) {
         if (chunk.type !== 'chunk' || !chunk.map) { continue }
 
-        const id = join(outputDir, chunk.fileName)
+        const id = resolve(outputDir, chunk.fileName)
         ids.add(id)
         const dest = id + '.map.json'
         await mkdir(dirname(dest), { recursive: true })
--- a/packages/vite/src/plugins/nitro-sourcemap.ts
+++ b/packages/vite/src/plugins/nitro-sourcemap.ts
@@ -1,0 +1,63 @@
+import { mkdir, readFile, writeFile } from 'node:fs/promises'
+import { dirname, join, resolve } from 'pathe'
+
+import type { Plugin as RollupPlugin } from 'rollup'
+import type { Plugin as VitePlugin } from 'vite'
+
+export const createSourcemapPreserver = () => {
+  let outputDir: string
+  const ids = new Set<string>()
+
+  const vitePlugin = {
+    name: 'nuxt:sourcemap-export',
+    configResolved (config) {
+      outputDir = config.build.outDir
+    },
+    async writeBundle (_options, bundle) {
+      for (const chunk of Object.values(bundle)) {
+        if (chunk.type !== 'chunk' || !chunk.map) { continue }
+
+        const id = join(outputDir, chunk.fileName)
+        ids.add(id)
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Nice!

Only i'm concerned about permornace overhead we are imposing by reading and writing sourcemaps to the disk. Random disk io is quite expensive even for faster storage drivers but more painfully for users with older hardware. 

I think we could do some simple checks for sourcemap size and use a ceiling to keep in memory or persist (or drop, nitro already drops sourcemap chunks including wasm files because of this. i think vendor sourcemap is also not that high quality to preserve. while nitro can do it later, it would be too much logic in higher level to involve nitro and also too late because the write part at least already happened).

danielroe:
Thank you for thinking about performance! These are excellent points for further implementation.

I think we can safely merge as the IO load from this PR is limited to the number of files output from the Vite server build - these are written 1x and read in 1x, making it a small proportion of total IO. (We could even later prevent writing the `.map` files, preferring our own `.map.json` files, reducing that IO still further.)

For further implementation, I think we could consider skipping nullifying sourcemaps (for nitro as well as both vite builds) for vendored files or massive WASM chunks if we want to do this, but I'd suggest doing it consistently across vite server + client rather than specifically in this plugin.

pi0:
> these are written 1x and read in 1x, making it a small proportion of total

Additional (number of chunks) x (serialize+write+read+parse). each need disk seek time and extra CPU time. 

I let you decide when to make further optimizations but I hope to happen in the same stable release. Client-generated maps are only consumed on demand but server chunks immediately in the next build step.

PR #28542, opened 2024-08-14
fix(nuxt): only augment `vue`, not sub-packages

### LINKED ISSUE #28561: Module type augmentation conflict breaks types for `$config` property on `ComponentCustomProperties` on nightly release
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.13.0-28726360.b1358220
- CLI Version:      3.13.0-1719571044.bca75ab
- Nitro Version:    2.10.0-28726471.af42c19d
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

1. https://stackblitz.com/edit/github-dz3tm4?file=pages%2Findex.vue,nuxt.config.ts,app.vue
2. after installation run `npm run typecheck`
3. observer type error `pages/index.vue:5:17 - error TS18046: '__VLS_ctx.$config' is of type 'unknown'.`

### Describe the bug

This is caused by the changes in https://github.com/nuxt/nuxt/pull/28446 (or more accurately https://github.com/vuejs/router/pull/2295) and actually demonstrates the issue its trying to solve, the mixing of module augmentation causes issues and will likely do the same for modules downstream.

### Additional context

If we want to continue the route of https://github.com/nuxt/nuxt/pull/28446 then I have [a PR](https://github.com/nuxt/nuxt/pull/28559) that would fix the issue described here. 

I'm in favor to instead only augmenting `vue` to be in line with Vue's guidelines (which https://github.com/nuxt/nuxt/pull/28542 does) but I'm pretty sure this will also cause issues downstream.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/28561

### 📚 Description

This removes augmentations of `@vue/runtime-core` in favour of only augmenting vue, which should fix issues when other packages (like `vue-router`) are only augmenting vue ([release notes](https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#bug-fixes-2)).

This may cause issues until the ecosystem moves away from other augments.

### DIFF
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -227,7 +227,7 @@ declare module '#app' {
   }
 }
 
-declare module 'vue' {
+declare module '@vue/runtime-core' {
   interface ComponentCustomProperties {
     $hello (msg: string): string
   }
--- a/docs/2.guide/2.directory-structure/1.plugins.md
+++ b/docs/2.guide/2.directory-structure/1.plugins.md
@@ -227,7 +227,7 @@ declare module '#app' {
   }
 }
 
-declare module '@vue/runtime-core' {
+declare module 'vue' {
   interface ComponentCustomProperties {
     $hello (msg: string): string
   }
--- a/docs/3.api/2.composables/use-runtime-config.md
+++ b/docs/3.api/2.composables/use-runtime-config.md
@@ -4,7 +4,7 @@ description: 'Access runtime config variables with the useRuntimeConfig composab
 links:
   - label: Source
     icon: i-simple-icons-github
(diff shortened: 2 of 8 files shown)


BobbieGoede:
So just to be sure, we're sticking with augmenting `vue`? It will take some time for modules to change augments, as long as any of these augment runtime-core it will cause Nuxt's types to break (the module types will work instead).

There may be a way for users to still have working types while modules are catching up, see https://stackblitz.com/edit/github-dz3tm4-brftas?file=pages%2Findex.vue,nuxt.config.ts,shim.d.ts. Maybe this causes other issues, I haven't tested it thoroughly 🤷‍♂️

Issue #23392, opened 2023-09-25
Optimize build time speed with a cache (or other settings)


polarathene:
### Describe the feature

Presently Nuxt seems to lack a build cache? (_[Next has one](https://nextjs.org/docs/pages/building-your-application/deploying/ci-build-caching) that can be cached for speeding up builds_)

In a real Nuxt project, I am observering `nuxt build` taking roughly 38 seconds (_8s vite client, 11s vite server, 19s nitro server_). This does not improve when running `nuxt build` again, even though nothing has changed.

Is it possible to optimize any of the build steps involved?
- Do the build tools used support re-use of a cache and Nuxt just needs to expose a setting (_or document one_) for enabling this support?
- If not, are there any opportunities to implement this where it'd be worthwhile?

Perhaps you could use [BLAKE3](https://github.com/BLAKE3-team/BLAKE3#readme) via [`@noble/hashes`](https://www.npmjs.com/package/@noble/hashes#blake3) as a speedy way to get a checksum on build inputs? (_could be an opt-in flag_). This is perhaps a naive approach, and depending on overhead may not always help reduce the build time 😅 

---

### Reproduction / Comparison

Nuxt example:

```console
# Reproduction environment:
$ docker run --rm -it --workdir /reproduction node:20-bookworm-slim bash
$ corepack enable

# An example project (possibly too small for measurable difference):
$ npx -y nuxi init --packageManager pnpm --gitInit false .
# Repeat this step, no improvement to build time:
$ time pnpm build
```

Next.js example:

```console
# Reproduction environment:
$ docker run --rm -it --workdir /reproduction node:20-bookworm-slim bash
$ corepack enable

# An example project:
$ npx -y create-next-app@latest --use-pnpm --example hello-world .
# Repeat this step, 1st build 10s while builds after are 5s:
$ time pnpm build
```

---

### Related

I've looked through Nuxt and Nitro docs, nothing really stands out cache wise for a CI to optimize reducing future builds with. Ideally a small change (_or none at all_) should not require the full build time.

> Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

First five pages of each for keywords searches of: `build time`, `build speed`, `build cache`:

- Disabling GZIP in builds for some gain (_`sourcemaps` is another potential option_): https://github.com/nuxt/nuxt/issues/23352#issuecomment-1732026846
[...]

Issue #14607, opened 2022-08-16
Split dev and production output inside `buildDir`


pi0:
We have a special directory in nuxt called `buildDir` and by default is `${rootDir}/.nuxt` used for 
generated files such as templates, types, caches and build output.

Currently, we combine both into the same top-level directory structure which causes issues if at the same time build and start the development server they override each other. There are also directories that can leverage caching if left intact and separated including `.nuxt/cache` and `.nuxt/dist` (see nuxt/nuxt.js#14605 useful for smart rebuild and project hashes) 

This is a semi-breaking change if modules were implicitly depending on filesystem structure **with full path** of buildDir and hardcoded dirs (while they shouldn't) but probably worth doing this before final 3.0.0 release to improve long term caching of this directory.

ruslankonev:
How about to implement this feature, because in some situations references from dev env like `import from 'file://'` get into the production builds.

Issue #14605, opened 2022-08-16
Create project manifest in buildDir


pi0:
Context: In nuxt 2.13, after the introduction of full-static mode, we introduced a new build-time optimization to reduce CI time (static generation) by avoiding a rebuild if project hashes were not different but only a new prerendering was needed (https://github.com/nuxt/nuxt.js/pull/7712 - [lastest impl](https://github.com/nuxt/nuxt.js/blob/86570cf6ab82cd7f58fb7973bd4c43a55307175b/packages/cli/src/utils/generate.js#L167))

Use cases for Nuxt 3:
- [ ] Smart rebuild for `nuxi generate`
- [x] Auto cleanup cache directories on hash changes (nuxt/framework#6668, nuxt/framework#6672)
- [ ] Use for test-utils setup
- [x] https://github.com/nuxt/nuxt/issues/23392

gabrielsze:
Curious if this could potentially include improving the `nuxt generate` step by only generating pages that have changed since the last generate step. This will be quite helpful for big sites that generate (tens of) thousands of pages on every build/generate step (i.e. only generate the marginal routes that are new or have changed since last generate)

danielroe:
That’s exactly the idea!

surgiie:
Is there any update or roadmap for this feature? It would be great to be able to build only single/pages needed.

SoaresMG:
Is there an option in the roadmap that allows `build` and `generate` separately?

Imagining an image that contains a build (uses `nuxi build`) and the entrypoint only runs `nuxi generate`.
This would be tremendous for website generation times.

alancolant:
Does this PR enable static generation using an existing .output directory without triggering a full rebuild?

My use case: the Nuxt source code remains unchanged between deployments — only the API data changes. I want to avoid rebuilding the entire app each time.

Would this PR allow running:
```bash
nuxi build        # once
npx nuxi generate # later, using the existing .output --target-dir=dist/current-date
```
...to regenerate static pages without recompiling the whole project?

That would significantly improve deployment time for data-driven static sites.

danielroe:
this has been implemented 🙏

PR #28594, opened 2024-08-19
feat(schema): split dev/prod build directories

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/14607

### 📚 Description

This adds support for splitting dev + production build directories. Because type-support is based on having a `.nuxt/tsconfig.json` in place, we can't _always_ split out the build directory.

I've opted to use `buildId` for the production build directory to allow parallel production builds - but this implementation detail will likely change when we implement https://github.com/nuxt/nuxt/issues/23392 (next on the list!)

This should enable running parallel builds alongside a dev server (or even multiple builds in parallel at the same time).

### Next steps
- [ ] consider whether `nuxi` should run prepare step on a separate directory before building (?)
- [x] https://github.com/nuxt/nuxt/issues/14605
- [x] https://github.com/nuxt/nuxt/issues/23392

### DIFF
--- a/packages/schema/src/config/common.ts
+++ b/packages/schema/src/config/common.ts
@@ -189,16 +189,14 @@ export default defineUntypedSchema({
         return defaultBuildDir
       }
 
-      const buildId = await get('buildId') as string
-
       // TODO: nuxi CLI should ensure .nuxt dir exists
       if (!existsSync(defaultBuildDir)) {
         // This is to ensure that types continue to work for CI builds
         return defaultBuildDir
       }
 
-      // TODO: handle build caching
-      return resolve(rootDir, 'node_modules/.cache/nuxt/builds', buildId)
+      // TODO: handle build caching + using buildId in directory
+      return resolve(rootDir, 'node_modules/.cache/nuxt/builds', 'production')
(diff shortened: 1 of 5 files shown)


packages/schema/src/config/common.ts
```diff
@@ -175,7 +175,31 @@ export default defineUntypedSchema({
...
+      const buildId = await get('buildId') as string
+
+      // TODO: nuxi CLI should ensure .nuxt dir exists
+      if (!existsSync(defaultBuildDir)) {
+        // This is to ensure that types continue to work for CI builds
+        return defaultBuildDir
+      }
+
+      // TODO: handle build caching
+      return resolve(rootDir, 'node_modules/.cache/nuxt/builds', buildId)
```
pi0, line 201: (maybe in later pr) we probably need to take care about GC. Size of this dir can be so big over the time. One fast start might be to keep `.cache/nuxt/builds.json` with id + date (so we can gc)

danielroe: Agreed!

pi0 (approved):

PR #28682, opened 2024-08-23
feat(schema,nuxt): add `shared` folder to alias and auto-imports

### LINKED ISSUE #28675: Add `#shared` alias
### Describe the feature

With the new directory structure, it makes sense to have a `shared/` directory for sharing constants, types and other utils between the `app/` and server/` directory.

After discussing with the core team, it makes sense to create a `#shared` alias.

```ts
// before
import { formatNumber } from '~~/shared/utils/dates'
// after
import { formatNumber } from '#shared/utils/dates'
```

It would be nice to support auto-import for `#shared/types` and `#shared/utils` as well ✨ 

### Additional information

### Final checks

Barbapapazes:
### 🔗 Linked issue

fix #28675

### 📚 Description

Hello 👋,

This PR adds:

- `#shared` alias to both `tsconfig.json` and `tsconfig.server.json`
- Enable auto imports from `#shared/types` and `#shared/utils` in `app` and `server` directories

~_The second bullet point is still in progress because I'm still trying to understand the usage of unimport within the project._~

So, after talking with @danielroe, this `#shared` folder is more complicated than expected. Here a list of expected behavior (that needs to be implemented)

- [ ] `./app` folder can import (and auto-import) from the `./shared` folder
- [ ] `./server` folder can import (and auto-import) from the `./shared` folder
- [ ] Inside the `./shared` folder, you **should not** be able to import stuff from both `./app` and `./server` (could be achieved by excluding the file from the `tsconfig.json` and a single chunk could be created)
- [ ] Inside the`./shared` folder, you could (we need to decide if we want or not this behavior) import and auto-import stuff from the `./shared` folder.

At the end, the shared folder is just for pure JS (or TS) stuff. Nothing have to be related to Vue or Nitro.

## Release Notes

- **New Features**
  - Enhanced import protection mechanisms with context-aware handling for shared directories.
  - Conditional inclusion of additional directories for Nuxt 4 compatibility.

- **Bug Fixes**
  - Improved error handling and reporting for module fetching in the Vite node server.

- **Documentation**
  - Updated configuration schema to include a shared directory for better resource organization.

- **Refactor**
  - Restructured Nitro configuration process for enhanced flexibility across different Nuxt versions.


danielroe (commented):
This looks great!

The next step is to configure import protection so that we do not use certain protected imports in that folder...

Barbapapazes:
> import protection

What do you mean? I never heard this term before. Do you have an example?

danielroe:
You can add patterns for the shared directory here:

https://github.com/nuxt/nuxt/blob/77e36ee27497cb5b9765a0cdc61e11f810272e1a/packages/nuxt/src/core/plugins/import-protection.ts#L12-L52

Or let me know, and I'll happily implement. 🙏

Barbapapazes:
> You can add patterns for the shared directory here:
> 
> https://github.com/nuxt/nuxt/blob/77e36ee27497cb5b9765a0cdc61e11f810272e1a/packages/nuxt/src/core/plugins/import-protection.ts#L12-L52
> 
> Or let me know, and I'll happily implement. 🙏

I do not know to know when the folder is the shared folder 🤔

Barbapapazes:
I'm facing another issue. Everything in the `shared` folder is available from `imports` folder because of this, people would want to import inside the shared folder, stuff from the shared folder but at the same time, I do not want people to import Vue stuff (from `imports`) inside the shared folder.

manniL:
> I'm facing another issue. Everything in the `shared` folder is available from `imports` folder because of this, people would want to import inside the shared folder, stuff from the shared folder but at the same time, I do not want people to import Vue stuff (from `imports`) inside the shared folder.

Is there anything in `imports` which is neither server nor client-dependent?

manniL:
@Barbapapazes I think adding the import protection for folders is a good first step. (cc @danielroe if you have ideas how to protect from `#imports`)

TheAlexLichter (approved):

Rednas83:
Why not share the `utils` on default with the server?

**Something like**
```
utils/capitalize.ts // Shared with the server on default
utils/capitalize.client.ts // Client only
utils/capitalize.server.ts // Server only
```

**Advantages**
* All `utils` in a single folder
* Same ux that is already used for `components`, `pages` and `plugins`
* No need to introduce  a new `shared` folder

danielroe:
we now have an `app/` folder for vue server+client, and the `.server` and `.client` suffix only applies to server vs client on the vue side of the app, not to the nitro or `server/` folder, which would make using that convention somewhat confusing.

Rednas83:
Idk, seems like a bit more complicated ux to me. 

**Something like**
```
shared/utils/capitalize.ts // Shared with the server 
[app]/utils/capitalize.ts // Client only
server/utils/capitalize.ts // Server only
```
**Versus in one folder**
```
[app]/utils/capitalize.ts // Shared with the server on default
[app]/utils/capitalize.client.ts // Client only
[app]/utils/capitalize.server.ts // Server only
```
or perhaps create a new root folder which contains all the utils to form a universal library that can be used on server and client level
**Universal library**
```
library/capitalize.ts // Shared with the server on default
library/capitalize.client.ts // Client only
library/capitalize.server.ts // Server only
```

Barbapapazes:
I think you're missing the point that the app run both on the server and on the client while the server is more like an API.

Because of this, using `.client` and `.server` in the app folder have a different meaning (see how it works for component).

Rednas83:
I know `server` and `app` are 2 different entities, but to have as little friction as possible I think a universal `utils`/`library` folder could help a lot.

Would also be cool to be able to call server utils directly from the client without the need of writing api endpoints

Perhaps something like🤔
```
library/capitalize.ts // Shared with the server on default
library/capitalize.client.ts // Client only
library/capitalize.server.ts // Server only
library/login.remote.ts // Remote call of a server function
```

For inspiration
[https://github.com/antfu/nuxt-server-fn](https://github.com/antfu/nuxt-server-fn)
[https://trpc-nuxt.vercel.app/](https://trpc-nuxt.vercel.app/)

packages/schema/src/config/common.ts
```diff
@@ -424,12 +429,13 @@ export default defineUntypedSchema({
...
   alias: {
     $resolve: async (val: Record<string, string>, get): Promise<Record<string, string>> => {
-      const [srcDir, rootDir, assetsDir, publicDir, buildDir] = await Promise.all([get('srcDir'), get('rootDir'), get('dir.assets'), get('dir.public'), get('buildDir')]) as [string, string, string, string, string]
+      const [srcDir, rootDir, assetsDir, publicDir, buildDir, sharedDir] = await Promise.all([get('srcDir'), get('rootDir'), get('dir.assets'), get('dir.public'), get('buildDir'), get('dir.shared')]) as [string, string, string, string, string, string]
       return {
         '~': srcDir,
         '@': srcDir,
         '~~': rootDir,
         '@@': rootDir,
+        '#shared': resolve(rootDir, sharedDir),
```
ineshbose, line 438: sorry I know its been merged, but I was just reviewing the changes.

I was wondering, do these alias resolution keys need to be static, especially in this case, where, say, I configure the shared dir to be `common`, so my imports would still be `#shared` (that's fine, I suppose I don't make the rules! but maybe its a little friction in that dev experience), or would we rather make it align to the dir name:

```suggestion
        [`#${sanitisedSharedDirName}`]: resolve(rootDir, sharedDir),
```

CernyMatej: I'd say thst it is better to keep them static for consistency. You can always define a custom alias if you want.

ineshbose: fair enough!

KevinOustry:
Hi there,
Nice idea of adding this `shared` folder.
But how it will works with Vue-Email ? 
Indeed we need to import specific Components from `app` into the `server` directory in order to render them with the data.
Thanks !

s1gr1d:
It seems like this breaks the Nuxt build when using the Sentry Nuxt SDK where Sentry is adding a Rollup plugin that wraps the server entry with a dynamic `import()`. We at Sentry have to do this, so OpenTelemetry works and users don't have to add the `--import` flag to the node command.

This issue was filed by a user: https://github.com/getsentry/sentry-javascript/issues/14179
I also checked this by running the default build and got this error:
```
RollupError: [plugin impound] This module cannot be imported in server runtime. [importing node_modules/nitropack/dist/presets/node/runtime/node-server.mjs from node_modules/nitropack/dist/presets/node/runtime/node-server.mjs?sentry-query-wrapped-entry?sentry-query-reexported-functions=defaultSENTRY-QUERY-END]
```

What the Sentry Nuxt Rollup plugin does is similar to the polyfill code example in the Rollup docs: https://rollupjs.org/plugin-development/#resolveid

This is the Rollup plugin in the Sentry Nuxt SDK: https://github.com/getsentry/sentry-javascript/blob/develop/packages/nuxt/src/vite/addServerConfig.ts#L116

Tell me if you need a reproduction :)

ThimoDEV:
Just tried the new shared folder, but found that auto import breaks for subfolders inside the ./shared/utils folder. I tried to specify a config folder in ./shared/utils/config/myConfig.ts and have mutiple files there but these are not auto imported

danielroe:
@ThimoDEV The auto-imports is only top level. For files nested more deeply, you can import from `#shared/your-folder`

ThimoDEV:
@danielroe Makes sense that its not possible like with the components. If we would like to configure more folders for auto import besides the types and utils folder, we would need to add that in nuxt config imports setting, right?

manniL:
> Makes sense that its not possible like with the components. If we would like to configure more folders for auto import besides the types and utils folder, we would need to add that in nuxt config imports setting, right?

Correct 👍

matt-clegg:
Is there a reason why only the top level files in the `shared/utils` and `shared/types` folders are auto-imported, and not nested folders like `shared/foo` or `shared/utils/bar`?

Barbapapazes:
> Is there a reason why only the top level files in the `shared/utils` and `shared/types` folders are auto-imported, and not nested folders like `shared/foo` or `shared/utils/bar`?

Because it is the same behavior for composables, utils, modules, plugins

vasilistotskas:
Is this recommended practice if i need to auto import all nested folders of the /shared directory? (it seems to work)

```
  imports: {
    autoImport: true,
    dirs: [
      'stores/**',
      '../shared/constants/**',
      '../shared/schemas/**',
      '../shared/types/**',
      '../shared/utils/**',
    ],
  },
    nitro: {
    imports: {
      dirs: [
        'shared/constants/**',
        'shared/schemas/**',
        'shared/types/**',
        'shared/utils/**',
      ],
    },
})
```

manniL:
you can just do `shared/**` to get all subfolders 👍🏻
The others are as you pointed out. Not sure if you could even use `#shared/**` as alias to avoid the difference between nitro and nuxt auto imports.

BlueskyFR:
Hi everyone! It seems no documentation has been added along with this PR, is it planned?
For example `shared`, `shared/types` and `shared/utils` are not present in the directory structure tab.
Also the rest of the documentation does not yet refer to the `shared` folder when specifying which directories are provided with autoimport capabilities.

BlueskyFR:
Also it seems that this feature requires Nuxt V4 although the [blog post](https://nuxt.com/blog/v3-14#shared-folder-for-code-and-types-shared-with-clientserver) pretends it is also available for V3. Maybe either the PR or the blog post should be updated accordingly?

BlueskyFR:
Just noticed a bug (after enabling V4 😋): type aliases are not created if the filenames are `*.d.ts`, they **SHOULD** be `*.ts` it seems.
Is this a desired behavior?

BlueskyFR:
Another one... Should I open up separate issues for these?

Having a `shared/types/test.ts` file with that:
```ts
export const hey = 2;
export type hey = string;
```

In `app.vue` I have the following problems:
image
image

manniL:
@BlueskyFR Please raise an extra issue if you encounter problems 👍

PR #28720, opened 2024-08-27
feat(schema): enable devtools by default

danielroe:
### 📚 Description

This moves to make `devtools` enabled by default for v4+

(Let me know if you have any concerns at all about this approach.)

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -827,9 +827,11 @@ function createPortalProperties (sourceValue: any, options: NuxtOptions, paths: 
   }
 }
 
-const _getDefaultNuxtConfig = () => /* js */
-  `// https://nuxt.com/docs/api/configuration/nuxt-config
+const _getDefaultNuxtConfig = () => {
+  const todaysDate = formatDate(new Date())
+  return /* js */ `// https://nuxt.com/docs/api/configuration/nuxt-config
 export default defineNuxtConfig({
-  devtools: { enabled: true }
+  compatibilityDate: '${todaysDate}'
 })
 `
+}
--- a/packages/schema/src/config/adhoc.ts
+++ b/packages/schema/src/config/adhoc.ts
@@ -64,5 +64,7 @@ export default defineUntypedSchema({
    * @see  [Nuxt DevTools](https://devtools.nuxt.com/) for more information.
    * @type { { enabled: boolean, [key: string]: any } }
    */
-  devtools: {},
+  devtools: {
+    enabled: true,
+  },
 })


atinux (approved):

antfu:
Originally I made that way (`undefined` as value) mostly because I don't want to force every one to see the toggle (which can be annoying in some cases), but let users to opt-in by keyboard shortcuts or config, keeping the default behaviour clean (while DevTools still enables). To me, honestly, I don't feel we **have to** enable the toggle explicitly by default. Would you share more about your point of views?

danielroe:
Mostly, for me, this is about cleaning up the default configuration, and thinking about what the best out-of-the-box experience might be.

with a toggle, I feel people are more likely to 'discover' the devtools. But I could be wrong. Let's do a poll!

alexcroox:
I personally disabled the dev tools because the toggle button was often in the way of UI elements, especially when testing a bottom mobile nav bar for example. I'd prefer a keyboard shortcut, or maybe a way to trigger from the browser Vue dev tools which would already be open?

SCR-20240828-iuvd

kricsleo:
I personally disable it when developing, maybe it's because I prefer a clean interface and nothing disturbs me in the UI. 
The devtools is awesome, and I hope everyone knows about it. But I don't use it all the time, only in some cases, so the default `not open` suits me better.

NamesMT:
+1 for a first time tour for the keyboard shortcut without the toggler in the UI by default.

danielroe:
Here's the result from the [Twitter poll](https://x.com/danielcroe/status/1828710815974752608):

CleanShot 2024-08-30 at 09 48 20@2x

I would observe that we're polling users who already know about devtools and in many cases are experienced users. Yet the need for greater devtools visibility is mostly for totally new users. 

So, on reflection, my feeling is that the best behaviour is:
* omitting any configuration from `nuxt.config`
* displaying toggle in browser by default
* maybe the first time you load a nuxt project on your computer, devtools starts _open_ with a little 'welcome to nuxt' tour?
* allow user to disable toggle being displayed by default for project _and all future projects_ from devtools itself, either modifying `.nuxtrc` in user directory, or (lighter) setting a local storage config key

Quite a lot of that would require implementation in devtools itself. What do you think @antfu?

antfu:
Yeah, I am up for it! And I agree it would be better and easier to implement in the DevTools itself. In that regard, do you want this behavior of DevTools only to apply to Nuxt 4? Or would it be enrolled to all Nuxt users with the latest DevTools?

danielroe:
I would be happy for that to apply to all users of devtools, unless you think it's a breaking behaviour change

antfu:
Created an issue to track it https://github.com/nuxt/devtools/issues/717

PR #28724, opened 2024-08-27
docs: fix typo

iiio2:
### 📚 Description

service workers are not written with hypen (-) service-workers. it is always written with two different words as far as I have seen. I think it should not be service-workers, rather it should be two different words, `Service Workers` like that.

https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers
https://web.dev/learn/pwa/service-workers/

### DIFF
--- a/docs/2.guide/1.concepts/4.server-engine.md
+++ b/docs/2.guide/1.concepts/4.server-engine.md
@@ -7,7 +7,7 @@ While building Nuxt 3, we created a new server engine: [Nitro](https://nitro.unj
 
 It is shipped with many features:
 
-- Cross-platform support for Node.js, Browsers, service-workers and more.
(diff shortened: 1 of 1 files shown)


docs/2.guide/1.concepts/4.server-engine.md
```diff
@@ -7,7 +7,7 @@ While building Nuxt 3, we created a new server engine: [Nitro](https://nitro.unj
 
 It is shipped with many features:
 
-- Cross-platform support for Node.js, Browsers, service-workers and more.
+- Cross-platform support for Node.js, Browsers, Service Workers and more.
```
danielroe, line 10:
```suggestion
- Cross-platform support for Node.js, browsers, service workers and more.
```

danielroe (approved):

PR #28743, opened 2024-08-28
docs: add missing `query` returned value from `useRoute()`

ismailsabet:


atinux (approved):

docs/3.api/2.composables/use-route.md
```diff
@@ -38,6 +38,7 @@ Apart from dynamic parameters and query parameters, `useRoute()` also provides t
 
 - `fullPath`: encoded URL associated with the current route that contains path, query and hash
 - `hash`: decoded hash section of the URL that starts with a #
+- `query`: access the route query parameters
```
danielroe, line 41:
```suggestion
- `query`: access route query parameters
```

PR #28745, opened 2024-08-28
feat(nuxt): normalise component names to match nuxt pattern

huang-julien:
### 📚 Description

This PR correct the name of auto-imported components. 

Vue plugin is defining a component's name based on the file name if it is not defined by the user. This PR replace it by the component's full name in pascal case.

Here's an exemple of the improved component name on the left side.

image

⚠️ this may be a breaking change for users that uses `@nuxt/test-utils` with `findComponent()` as instead of looking for `World`, users will have to look for `HelloWorld`


packages/nuxt/src/components/loader.ts
```diff
@@ -74,10 +75,13 @@ export const loaderPlugin = createUnplugin((options: LoaderOptions) => {
           if (lazy) {
             imports.add(genImport('vue', [{ name: 'defineAsyncComponent', as: '__defineAsyncComponent' }]))
             identifier += '_lazy'
-            imports.add(`const ${identifier} = __defineAsyncComponent(${genDynamicImport(component.filePath, { interopDefault: false })}.then(c => c.${component.export ?? 'default'} || c)${isClientOnly ? '.then(c => createClientOnly(c))' : ''})`)
+            imports.add(`const ${identifier} = __defineAsyncComponent(${genDynamicImport(component.filePath, { interopDefault: false })}.then(c => ${options.isDev ? `({...(c.${component.export ?? 'default'} || c), name: ${component.pascalName}})` : `c.${component.export ?? 'default'} || c`})${isClientOnly ? '.then(c => createClientOnly(c))' : ''})`)
```
danielroe, line 78:
```suggestion
            imports.add(`const ${identifier} = __defineAsyncComponent(${genDynamicImport(component.filePath, { interopDefault: false })}.then(c => ${options.isDev ? `({...(c.${component.export ?? 'default'} || c), name: '${component.pascalName}'})` : `c.${component.export ?? 'default'} || c`})${isClientOnly ? '.then(c => createClientOnly(c))' : ''})`)
```

packages/nuxt/src/components/loader.ts
```diff
@@ -74,10 +75,13 @@ export const loaderPlugin = createUnplugin((options: LoaderOptions) => {
...
           } else {
             imports.add(genImport(component.filePath, [{ name: component._raw ? 'default' : component.export, as: identifier }]))
-
+            if (options.isDev) {
+              imports.add(`const ${identifier}_dev = Object.assign(${identifier}, { name: '${component.pascalName}' })`)
```
danielroe, line 82:
```suggestion
              imports.add(`const ${identifier}_dev = Object.assign(${identifier}, { name: ${JSON.stringify(component.pascalName)} })`)
```

packages/nuxt/src/components/loader.ts
```diff
@@ -15,6 +15,7 @@ interface LoaderOptions {
   sourcemap?: boolean
   transform?: ComponentsOptions['transform']
   experimentalComponentIslands?: boolean
+  isDev: boolean
```
huang-julien, line 18: @danielroe We could probably also do it with nuxt.options.test ?

danielroe: makes sense! 👍

danielroe (commented):
the tricky thing is that I _think_ we will be overriding the name _even_ when the user has provided that name themselves.

do you know of a way to tell?

danielroe (commented):
What about directly modifying the transformed component rather than handling this at the loader level?

huang-julien:
We could :+1: . it's probably a bit more complex, do you think pre or post vue plugin ?

danielroe:
It should be more straightforward. We just replace the generated `__name` attribute that is generated by the vue compiler.

```ts
const __sfc__ = /* #__PURE__ */Object.assign({
  name: 'MyName', // this will override if user provides their own...
}, {
  __name: 'App',
  setup (__props, { expose: __expose }) {
    // code
  },
})
```

(We could also likely override `__name` instead of `name` in the existing approach, but it would be slightly less performant than doing this at build-time.)

huang-julien:
It will. Do you still prefer to move it to another plugin and modify the component or to keep the current implementation ?

packages/nuxt/src/components/module.ts
```diff
@@ -42,6 +43,12 @@ export default defineNuxtModule<ComponentsOptions>({
         : context.components
     }
 
+    if (nuxt.options.dev || nuxt.options.test) {
```
danielroe, line 46: We should probably override component names even in build, so KeepAlive will work consistently in this case.

huang-julien: In that case, this would slightly increase the final bundle 🤔 . Do you think this would solve a bug with KeepAlive ?

danielroe (approved):

PR #28746, opened 2024-08-28
docs: updated link for source code for nuxt hooks at build time

onmax:


danielroe (approved):

pi0:
Cross sharing from https://github.com/unjs/automd/issues/77

Here is an alternative method:

https://github.com/nuxt/nuxt/blob/main/packages/schema/src/types/hooks.ts#:~:text=export%20interface%20NuxtHooks

PR #28760, opened 2024-08-30
fix(nuxt): cookie composable does not propogate null to all refs

### LINKED ISSUE #28343: Deleting cookies `useCookie` does not work in components if you pass `null`
### Environment

- Operating System: Darwin
- Node Version:     v20.11.1
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/yesworld/reproduce-issue-cookie-test

uncomment and comment out these lines..
- https://github.com/yesworld/reproduce-issue-cookie-test/blob/main/composables/useBanner.ts#L7
- https://github.com/yesworld/reproduce-issue-cookie-test/blob/main/composables/useBanner.ts#L16

### Describe the bug

If you delete cookies using `null`, nothing changes in the components `<CompA />` and `<CompB />`.
https://github.com/yesworld/reproduce-issue-cookie-test/blob/main/composables/useBanner.ts#L7
```js
    token.value = null // doesn't work
```

If you delete using an empty string `''`, everything works fine.
https://github.com/yesworld/reproduce-issue-cookie-test/blob/main/composables/useBanner.ts#L8
```js
    token.value = '' // it works!
```

### Additional context

_No response_

### Logs

_No response_

Eckhardt-D:
### 🔗 Linked issue

resolves #28343

### 📚 Description

When `useCookie` is in a composable and used in different layers of pages/components. Calling `.value = null` from a parent layer does not unset the cookie in the children via the event listener.

The reason is that, in the event listener, the null'ed cookie is not included in the array of changes and therefor never updated in the ref, because it is not found by the `.find` call.

## Solution

Check in the `event.deleted` if the cookie was not removed or updated with a time in the past:

https://developer.mozilla.org/en-US/docs/Web/API/CookieChangeEvent/deleted

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -104,8 +104,13 @@ export function useCookie<T = string | null | undefined> (name: string, _opts?: 
 
     if (store) {
       const changeHandler = (event: any) => {
+        // The cookieStore API does not include the cookies that have
+        // been set to null, so we need to check the cookies object
         const cookie = event.changed.find((c: any) => c.name === name)
-        if (cookie) { handleChange({ value: cookie.value }) }
+
+        if (cookie || cookies[name] !== null) {
+          handleChange({ value: cookie?.value ?? null })
+        }
       }
       store.addEventListener('change', changeHandler)
       if (hasScope) {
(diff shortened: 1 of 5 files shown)


Eckhardt-D:
Hmm, same tests passed locally. Perhaps a flaky test. Any recommendations on the best way to test this @danielroe? FWIW I also cloned the repro in the issue and patched the Nuxt module with the same changes in node_modules which fixed the issue.

danielroe (approved):
Thank you! ❤️

yesworld:
Thanks!

PR #28801, opened 2024-09-03
docs: fix postcss codeblock typo

Hebilicious:
### 📚 Description

Welcome back everyone!

Following google search I stumbled onto this https://github.com/nuxt/docs

Any way we could remove this from the google search, and I guess we could upgrade the readme there.

I think the right UX would be to have a link at the bottom or top of each doc page to the source markdown file for easier collaborating for newcomers.


danielroe (approved):

danielroe:
thank you! and archived that repo... 🤞

PR #28814, opened 2024-09-03
feat: improve `isNuxtError()` type inference

### LINKED ISSUE #28813: Improve `isNuxtError()` type inference
### Describe the feature

Here are some use cases with types that can be improved: https://stackblitz.com/edit/isnuxterror-type?file=app.vue

Currently, `isNuxtError()` accept an _optional_ param of type `string | object`.

I see 3 areas of improvements:
- Since the runtime code already supports every type, **`string | object` is unnecessarily strict** and forces type casting to avoid TS errors.
- When you're forced to cast it (`as any` or `as NuxtError`), you'll **lose its type inference** (`error is NuxtError<DataT>`) and the param will keep having the same previous type, even after asserting it inside a condition.
- The param is currently optional, but maybe it makes more sense to make it **mandatory** (while also accept values like `undefined | null`).

~I'll follow up with a simple PR for this~ [Done](https://github.com/nuxt/nuxt/pull/28814)!

### Additional information

### Final checks

gianlucadifrancesco:
### 🔗 Linked issue

resolves #28813

### 📚 Description

This PR will improve the type inference of `isNuxtError()`, **without changing its implementation at all**.

The change is a one-liner, very simple, but it offers several advantages:
- **Wider param type:** the runtime code is already resilient to all parameter types, so its type is set to `unknown` to improve DX and avoid casting.
- **Always infer type:** in this way, the type inference (`error is NuxtError<DataT>`) will always be applied after being asserted.
- **Mandatory param:** I thought it was good to avoid using `isNuxtError()` without a param, as it's not particularly meaningful on its own.

### DIFF
--- a/packages/nuxt/src/app/composables/error.ts
+++ b/packages/nuxt/src/app/composables/error.ts
@@ -53,7 +53,7 @@ export const clearError = async (options: { redirect?: string } = {}) => {
 
 /** @since 3.0.0 */
 export const isNuxtError = <DataT = unknown>(
-  error?: string | object,
+  error: unknown,
 ): error is NuxtError<DataT> => !!error && typeof error === 'object' && NUXT_ERROR_SIGNATURE in error
 
 /** @since 3.0.0 */


danielroe (approved):

danielroe:
Thank you!

PR #28873, opened 2024-09-06
feat(nuxt,schema): allow configuring vue `AppConfig`

DamianGlowala:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/28285#issuecomment-2333986417

### 📚 Description

This allows for configuring the Vue's [`app.config.idPrefix`](https://vuejs.org/api/application.html#app-config-idprefix) through `vue.idPrefix` in Nuxt config.

Thoughts on the name? Shall we consider `vue.useIdPrefix`?


huang-julien (commented):
Isn't this something configurable with `app:created` hook ?

DamianGlowala:
Good point, it is. The question is whether we should recommend creating a plugin like so:
```
  nuxtApp.hook('app:created', (app) => {
    app.config.idPrefix = 'foo'
  })
})
```
... or prefer to set it in the Nuxt config. (I don't have a strong opinion on this matter and happy to close the PR if using a hook is the preferred way.)

huang-julien:
I'm in favor of letting users create a plugin. But our documentation may require a change if this wasn't clear enough that we can set runtime config of vue within a plugin :eyes:

danielroe:
I can see a benefit to allowing users to define the _serialisable_ bits of the [vue app config](https://vuejs.org/api/application.html#app-config) in their nuxt.config file.

could we generalise this a bit, maybe with `vue.config`?

DamianGlowala:
@danielroe updated 👌

danielroe (approved):

PR #28876, opened 2024-09-07
feat(nuxt): warn if dups is deteced in layouts & pages

OsamaHaikal:
This PR adds warning logs when duplicate page or layout files are detected in different Nuxt layers. It enhances DX to help identify and resolve conflicts caused by files with the same relative paths.

Sample output logs:

image


danielroe (commented):
Part of the idea of layers is to _allow_ overriding by higher level layers. I'm worried this might be too much noise - and also suggest there is an issue, when in fact things are working as intended? Maybe we could show this info only if `nuxt.options.debug` is enabled?

OsamaHaikal:
I got you, I may have rushed to find a solution. but if the overriding was a goal from the developer then I agree it will be noisy

we might display this info in the nuxt devtools instead

PR #28878, opened 2024-09-07
Fix bun script for init new project

naruko-hstk:
The correct init script should use "bunx" not "bun x".


danielroe:
bunx is an alias for bun x

https://bun.sh/docs/cli/bunx

PR #28912, opened 2024-09-10
perf(nuxt): remove interop default for dynamic components

danielroe:
### 📚 Description

since https://github.com/nuxt/nuxt/pull/28509 (released in v3.13.0), we no longer need an interop default for dynamic vue components as Vue + Vue router already do this check themselves, which also will slightly reduce bundle size.

By removing the interopDefault in the case of vue-router, we unlock possibilities for vue-router to do things with named exports (such as data loading 👀).

### DIFF
--- a/docs/2.guide/4.recipes/1.custom-routing.md
+++ b/docs/2.guide/4.recipes/1.custom-routing.md
@@ -22,7 +22,7 @@ export default {
     {
       name: 'home',
       path: '/',
-      component: () => import('~/pages/home.vue').then(r => r.default || r)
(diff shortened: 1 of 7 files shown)


mattmoredev:
Since updating nuxt from 3.13.1 to 3.13.2 I am now having an issue in one of my components when I import a library called [Jexl](https://github.com/TomFrost/Jexl) and try to initialize it I am seeing the following error:

`_interopRequireDefault2 is not a function
`

I am guessing this is related to this particular update. Is there a workaround for this?

PR #28944, opened 2024-09-11
fix(nuxt): ensure component imports are injected last

danielroe:
### 📚 Description

detected in https://github.com/elk-zone/elk/pull/2946.

otherwise if `.vue` files were not already transformed the injections were inserted at the beginning of the file, not in the `<script>` block.

cc: @userquin


userquin (approved):
local test and patch in elk.zone pr working

PR #28964, opened 2024-09-12
docs: Added additional instructions for server types

victorgarciaesgi:
### 📚 Description

I managed to keep `nuxi typecheck` working by adding this lines to my  root`tsconfig.json`.

### DIFF
--- a/docs/2.guide/2.directory-structure/1.server.md
+++ b/docs/2.guide/2.directory-structure/1.server.md
@@ -142,7 +142,17 @@ To improve clarity within your IDE between the auto-imports from 'nitro' and 'vu
 }
 ```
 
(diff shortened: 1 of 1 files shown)


danielroe (commented):
If this is a better solution we should update the generated tsconfig instead. (It's already something I'm considering.)

victorgarciaesgi:
Oh yeah I see, that's maybe better than to do it manuelly yeah! Otherwise on server plugin I get typecheck error with vue-tsc

danielroe:
When you do this, does api inference work - e.g. `$fetch('/api/route')` returns correctly value?

danielroe:
> When you do this, does api inference work - e.g. `$fetch('/api/route')` returns correctly value?

checking, this breaks API type inference, so I wouldn't want to recommend it.

But improving the type support for server vs app is an important priority for me now that we have a directory structure in nuxt v4 that is more conducive to it...

PR #28967, opened 2024-09-12
test: add regression test for using `route` in template

danielroe:
### 📚 Description

We have a weird issue with latest `vue-router` now that we don't de-default routes (https://github.com/nuxt/nuxt/pull/28912). When using `route` in template, it throws a hard error:

```
Cannot convert object to primitive value

  at String (<anonymous>)
  at Object.replacer (@vue/shared#cjs:488:12)
  at JSON.stringify (<anonymous>)
  at Module.toDisplayString (@vue/shared#cjs:466:196)
  at setup (pages/index.vue#cjs:18:41)
  at _sfc_main.setup (pages/index.vue#cjs:37:23)
  at callWithErrorHandling (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js#cjs:200:19)
  at setupStatefulComponent (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js#cjs:7757:25)
  at setupComponent (node_modules/@vue/runtime-core/dist/runtime-core.cjs.js#cjs:7718:36)
  at renderComponentVNode (node_modules/@vue/server-renderer/dist/server-renderer.cjs.js#cjs:645:15)
```

> https://stackblitz.com/edit/github-38rsyf-1zb3ja

To be fair, this is not something users should ever do on a live site, but it's common in development or debugging and the error is very confusing.

The reason occurs because, since `vue-router@4.4.4`, the entire imported module is stored on the route object for use with data loaders (https://github.com/vuejs/router/commit/5ff2cd09d656a6f5ff38517c87d48b89f41ff49b).

That is fine, except that `toDisplayString` from vue (used to stringify objects in template) basically now can't handle displaying a route in the template. That is because `toDisplayString` handles stringifying more complex objects like Modules like this:

```ts
if (isObject(val) && !isArray(val) && !isPlainObject(val)) {
  // native elements
  return String(val)
}
```

> https://github.com/vuejs/vue-next/blob/0126cfff9d93bcec70e5745519f6378e3cd3f39c/packages/shared/src/toDisplayString.ts#L57-L60

A workaround is to partially revert https://github.com/nuxt/nuxt/pull/28912 and de-default route components for now, until this is resolved either in `vue` (by adding the ability to log Modules) or in `vue-router`.


posva:
This will probbaly not fix the issue you mentioned, it had to be fixed in vue router. I just released a patch version

posva:
I will have to add the toDisplayString test to Vue router 😅

thanks for finding out about it

PR #28999, opened 2024-09-14
docs: update links to flowchart in 4.contribution.md

giveget-ee:
Links to flowchart lead to empty page if they contain a token.
Deleting the token fixes the problem.

Update: It actually loads, but only after a page refresh. No need to delete the token.


giveget-ee:
My bad - just checked, that without token flowchart is displayed, but it's a wrong chart.
Then it's an issue - not a PR. 
Please replace the token to correct one.

danielroe:
This link isn't correct - try viewing it in an incognito tab.

The hash contains the entire mermaid diagram and is required to display it.

I'd ideally like to include/render the mermaid diagram inline rather than linking out to another service - PR would be welcome, but might require needing to do some work in https://github.com/nuxt/nuxt.com to make that possible.

PR #29010, opened 2024-09-15
docs: improve explanation of ssr + data fetching

cybercoder-naj:
### 🔗 Linked issue

This [reddit](https://www.reddit.com/r/Nuxt/comments/1ffsl21/what_confuses_you_the_most_in_nuxt/) post from @manniL highlighted many users having trouble understanding how SSR works and the differences between the 3 ways to fetch the data.

### 📚 Description

I have made some changes to the relevant sections of the documentation that perhaps explain the concepts more ergonomically. Here is a rundown of the why I made these changes:

**Getting Started > Data Fetching**
1. The first content of the page should be more direct. My changes attempt to simplify the "nutshell" to provide foundational knowledge, building from `$fetch` being the simplest way to `useAsyncData` being abstractions over `$fetch`.
2. I added a simple example, demonstrating where `$fetch` is more suitable. 

**Guide > Rendering**
1. The content for Universal Rendering has major restructuring. This [thread](https://www.reddit.com/r/Frontend/comments/1bgm6zu/what_is_the_difference_between_server_side/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) displayed confusions between SSR and Nuxt's Universal Rendering, which is why the first line explains that they are **similar** to PHP's and Ruby's rendering mode.
2. Although the concept was explained somewhat properly, the phrases seem disconnected from one another. I've attempted to make the sentences flow from one to the other using a "story" philosophy. This "story" begins in the server where I establish what the server does and what is returned. 
3.  I altered the second paragraph completely, trying my best not to change the meaning of the earlier content. I pick up the "story" from the previous paragraph where the stage now belongs to the browser. I explain what is happening on the browser, which gives an opportunity to hint towards Hydration.
4. I removed all mentions of CSR till the 2nd paragraph to avoid confusion. It is explained clearly in the following paragraphs and other portions on the same page.
5. Lastly, with an example, I briefly mention which portions of a vue file are executed on the server and/or the client.

---

I hope you like these changes. I am happy to discuss them. Two queries come to mind with these changes:
1. Now that the "nutshell" pointers in "Data Fetching" follows the order of `$fetch`, `useFetch` and `useAsyncData`, should this order change in the page too? I could not find a fitting way to do this. (**Edit**: have now done this)
2. The section **`What's server-rendered and what's client-rendered?`** seems like it should have a page of its own. We can talk about the order of execution and the lifecycle hooks of Vue.js. I see a potential where a separate page for this topic could benefit greatly.


docs/1.getting-started/6.data-fetching.md
```diff
@@ -8,9 +8,9 @@ Nuxt comes with two composables and a built-in library to perform data-fetching
 
 In a nutshell:
 
-- [`useFetch`](/docs/api/composables/use-fetch) is the most straightforward way to handle data fetching in a component setup function.
-- [`$fetch`](/docs/api/utils/dollarfetch) is great to make network requests based on user interaction.
-- [`useAsyncData`](/docs/api/composables/use-async-data), combined with `$fetch`, offers more fine-grained control.
+- [`$fetch`](/docs/api/utils/dollarfetch) is the simplest way to make a network request.
+- [`useFetch`](/docs/api/composables/use-fetch) is the SSR-friendly version of `$fetch` that prevent hydration mismatches.
```
manniL, line 12: Might be good to add that it also will not fetch data twice (on server and on client) but just once

cybercoder-naj: Thanks for the review, a good point! I've amended that line and it is clearer now.

cybercoder-naj:
Additional changes in **Getting Started > Data Fetching**

1. Changed the heading to *The need for `useFetch` and `useAsyncData`* to force an emphasis on the contents.
i. Removed the subheading "Network calls duplication" since we mentioned it a couple of times and it is not fitting to have this subtitle in the revised changes.
2. Changed the order of `$fetch` and `useFetch` to follow the same order of the nutshell pointers.

Let me know what you think.

docs/1.getting-started/6.data-fetching.md
```diff
@@ -8,63 +8,63 @@ Nuxt comes with two composables and a built-in library to perform data-fetching
 
 In a nutshell:
 
-- [`useFetch`](/docs/api/composables/use-fetch) is the most straightforward way to handle data fetching in a component setup function.
-- [`$fetch`](/docs/api/utils/dollarfetch) is great to make network requests based on user interaction.
-- [`useAsyncData`](/docs/api/composables/use-async-data), combined with `$fetch`, offers more fine-grained control.
+- [`$fetch`](/docs/api/utils/dollarfetch) is the simplest way to make a network request.
+- [`useFetch`](/docs/api/composables/use-fetch) is wrapper around `$fetch` that fetches data only once in [universal rendering](/docs/guide/concepts/rendering#universal-rendering).
+- [`useAsyncData`](/docs/api/composables/use-async-data) is similar to `useFetch` but offers more fine-grained control.
```
huang-julien, line 13: How about switching those two ? 
Since `useFetch` is a "wrapper" around `$fetch` and `useLazyAsyncData`, it's probably easier to understand if we explain in this order: `$fetch` +  `useAsyncData` ->`useFetch`

cybercoder-naj: That change did cross my mind. Although you're right, in a new Nuxt dev's point of view, understanding `$fetch` and `useFetch` first is more fundamentally important. `useAsyncData` is a step towards more customisability which comes later.

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -11,18 +11,41 @@ By default, Nuxt uses **universal rendering** to provide better user experience,
...
+  <div>
+    <p>Count: {{ counter }}</p>
+    <button @click="handleClick">Increment</button>
+  </div>
+</template>
+```
+
+On the initial request, the `counter` ref is initialized in the server since it is rendered inside the `<p>` tag. The contents of `handleClick` is never executed here. During hydration in the browser, the `counter` ref is re-initialized. The `handleClick` finally binds itself to the button; Therefore it is reasonable to deduce that the body of `handleClick` will always run in a browser environment.
+
+[Middlewares](/docs/2.guide/2.directory-structure/1.middleware.md) and [pages](/docs/2.guide/2.directory-structure/1.pages.md) run in the server and on the client during hydration. [Plugins](/docs/2.guide/2.directory-structure/1.plugins.md) can be rendered on the server or client or both. [Components](/docs/2.guide/2.directory-structure/1.components.md) can be forced to run on the client only as well. [Composables](/docs/2.guide/2.directory-structure/1.composables.md) and [utilities](/docs/2.guide/2.directory-structure/1.utils.md) are rendered based on the context of their usage.
```
xjccc, line 45: [Middlewares](/docs/guide/directory-structure/middleware)
[pages](/docs/guide/directory-structure/pages)
[Plugins](/docs/guide/directory-structure/plugins)
[Components](/docs/guide/directory-structure/components)
[Composables](/docs/guide/directory-structure/composables)
[Utils](/docs/guide/directory-structure/utils)
Is this right?

danielroe: yes, the numbers are used for ordering but not in URLs

cybercoder-naj: Thanks, just fixed it. 🚀

docs/2.guide/1.concepts/3.rendering.md
```diff
@@ -11,18 +11,41 @@ By default, Nuxt uses **universal rendering** to provide better user experience,
...
 
-To not lose the benefits of the client-side rendering method, such as dynamic interfaces and pages transitions, the Client (browser) loads the JavaScript code that runs on the Server in the background once the HTML document has been downloaded. The browser interprets it again (hence **Universal rendering**) and Vue.js takes control of the document and enables interactivity.
-
-Making a static page interactive in the browser is called "Hydration".
+Once the HTML document has been downloaded, The browser interprets this and Vue.js takes control of the document. The same JavaScript code that once ran on the server runs on the client (browser) **again** in the background now enabling interactivity (hence **Universal rendering**) by binding its listeners to the HTML. This is called **Hydration**. When hydration is complete, the page can enjoy benefits such as dynamic interfaces and page transitions.
```
HigherOrderLogic, line 16:
```suggestion
Once the HTML document has been downloaded, the browser interprets this and Vue.js takes control of the document. The same JavaScript code that once ran on the server runs on the client (browser) **again** in the background now enabling interactivity (hence **Universal rendering**) by binding its listeners to the HTML. This is called **Hydration**. When hydration is complete, the page can enjoy benefits such as dynamic interfaces and page transitions.
```

danielroe (approved):
thank you!

PR #29011, opened 2024-09-15
docs: add tip about using defineNuxtComponent instead of defineComponent

### LINKED ISSUE #29003: `useFetch` twice "A composable that requires access to the Nuxt instance was called outside of a plugin..."
### Environment

- Operating System: Linux
- Node Version:     v20.15.1
- Nuxt Version:     3.13.1
- CLI Version:      3.13.1
- Nitro Version:    2.9.7
- Package Manager:  bun@1.1.28
- Builder:          -
- User Config:      compatibilityDate
ui, tailwindcss, css, vite, nitro, typescript
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-jdafkm?file=app.tsx

### Describe the bug

Calling `useFetch` / `useLazyFetch` with await returns:
`[nuxt] A composable that requires access to the Nuxt instance was called outside of a plugin, Nuxt hook, Nuxt middleware, or Vue setup function. This is probably not a Nuxt bug. Find out more at `https://nuxt.com/docs/guide/concepts/auto-imports#vue-and-nuxt-composables`.`

When I remove the `await` the page loads but I'm not sure is the correct way of using `useFetch`.

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
fix #29003 

### 📚 Description

@danielroe i'm not sure whether this is just something missing in the documentation or a regression.

### DIFF
--- a/docs/2.guide/1.concepts/1.auto-imports.md
+++ b/docs/2.guide/1.concepts/1.auto-imports.md
@@ -64,6 +64,10 @@ If you get an error message like `Nuxt instance is unavailable` then it probably
 Watch a video from Alexander Lichter about handling async code in composables and fixing `Nuxt instance is unavailable` in your app.
 ::
 
+::tip
+When using a composable that requires the Nuxt context inside a non-SFC component, you need to wrap your component with `defineNuxtComponent` instead of `defineComponent`
+::
+
 ::read-more{to="/docs/guide/going-further/experimental-features#asynccontext" icon="i-ph-star-duotone"}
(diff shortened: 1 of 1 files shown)


danielroe:
Maybe the unctx transform isn't running for jsx files.

huang-julien:
🤔Well this seems to be also not done in nuxt 3.1.0. There's probably a reason we didn't do it

danielroe:
I wouldn't dismiss a simple oversight!

danielroe (approved):

PR #29023, opened 2024-09-16
fix(nuxt): simplify plugin InjectionType template conditional

### LINKED ISSUE #29021: Plugin type `IsAny` check broken broken in Typescript 5.5 & above
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.17.0`
- Nuxt Version:     `3.13.2`
- CLI Version:      `3.13.2`
- Nitro Version:    `-`
- Package Manager:  `pnpm@9.9.0`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `modules`, `scripts`
- Runtime Modules:  `@nuxt/scripts@0.8.5`
- Build Modules:    `-`

### Reproduction

[https://github.com/Eazash/nuxt-plugin-type-reproduction](https://github.com/Eazash/nuxt-plugin-type-reproduction)

### Describe the bug

In the event that a registry script is enabled from module config, nuxt scripts calls `extendTypes` to insert its plugin registration for registry scripts (https://github.com/nuxt/scripts/blob/84c6c4c039523485e84e058f50ef66272ecec45a/src/module.ts#L160). This adds `InjectionType<typeof import("../modules/@nuxt-scripts.js").default>`. 

The generated `.nuxt/modules/@nuxt-scripts.d.ts` file however does not have a default export & `IsAny<typeof import("../modules/@nuxt-scripts.js").default>` resolves to `any` as of typescript 5.5. This in turn means that `InjectionType<typeof import("../modules/@nuxt-scripts.js").default>` resolves to `any` and breaks plugin type injections for all plugins, including custom plugins. 

### Additional context

The above reproduction uses `@nuxt/scripts` to recreate the issue but I believe the bug is widespread. The root cause is that as of microsoft/typescript#58610, the "error" `any` behaves differently from regular `any`. This somehow causes `IsAny<typeof import("../modules/@nuxt-scripts.js").default>` to resolve to `any` instead of `true`

Related Issues
- microsoft/typescript#59981
- microsoft/typescript#58960

### Logs

_No response_

Eazash:
### 🔗 Linked issue

Resolves #29021
### 📚 Description

This PR simplifies the generated `InjectionType` types used for plugins by removing the type `IsAny` and instead relying on directly inferring the default export of plugins. The reasoning behind this is that `IsAny` was not a consistent check and could be inlined.

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -117,10 +117,9 @@ import type { Plugin } from '#app'
 
 type Decorate<T extends Record<string, any>> = { [K in keyof T as K extends string ? \`$\${K}\` : never]: T[K] }
 
-type IsAny<T> = 0 extends 1 & T ? true : false
-type InjectionType<A extends Plugin> = IsAny<A> extends true ? unknown : A extends Plugin<infer T> ? Decorate<T> : unknown
+type InjectionType<A extends Plugin> = A extends {default: Plugin<infer T>} ? Decorate<T> : unknown
 
-type NuxtAppInjections = \n  ${tsImports.map(p => `InjectionType<typeof ${genDynamicImport(p, { wrapper: false })}.default>`).join(' &\n  ')}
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
Thank you! ❤️

PR #29025, opened 2024-09-16
docs(nuxt): edit `url type` to preserve `autocompletion`

marcel-mpunkt:
### 📚 Description

When wrapping `useFetch` the types for `parameters`,e.g. for the url, are not inferred automatically. They could also not be imported easily.


MarcelMpunkt:
I guess it is not the cleanest solution, but should come in handy for other devs.
Alternative could be to make the `type` from `useFetch` exported

danielroe (commented):
We don't want to preserve the auto complete as the fetch has a different base URL.

MarcelMpunkt:
🙊 sorry then...keep up the good work...thx for your review though

PR #29052, opened 2024-09-18
perf(nuxt): use native `fetch` to get build manifest

InvincibleJun:
### 📚 Description
fix: fetch manifest.json in chrome 63. Request method has no default value at ofetch lib, it will returen 400 Bad reuqest.


danielroe:
Would you provide a reproduction? I was able to run `fetch` in chrome 63 without passing `method` to it without any issues.

Also, Chrome 63 was released in 2017 - seven years ago. That's a bit long to be supporting this in Nuxt core, so I think it might make more sense for you to use `patch-package` or `pnpm patch` to update this line if you really need it.

InvincibleJun:
I am do something with low chrome, But I think nuxt3 is a very nice framework.So, I do something to use nuxt3 in low browser, But I find doubtful questions.
when i fetchManifest, i will return 400 Bad Request. I debugger it then find use window fetch to request with method undefined.
if update patch-package will causing chain problem. I don't how many lib use it. And I think fetch manifest method only be 'get', it will not causing any problem.

```
// it will return 400 Bad request
fetch('/test', { method: undefined })
```

```
// it will return 200 Ok
fetch('/test', {})
```

InvincibleJun:
image

step in any nuxt3 project
```
// dev
pnpm run dev

// also build 
pnpm run build 
node .output/server/index.mjs
```

packages/nuxt/src/app/composables/manifest.ts
```diff
@@ -24,9 +24,7 @@ function fetchManifest () {
   if (!isAppManifestEnabled) {
     throw new Error('[nuxt] app manifest should be enabled with `experimental.appManifest`')
   }
-  manifest = $fetch<NuxtAppManifest>(buildAssetsURL(`builds/meta/${useRuntimeConfig().app.buildId}.json`), {
-    responseType: 'json',
-  })
+  manifest = fetch(buildAssetsURL(`builds/meta/${useRuntimeConfig().app.buildId}.json`)).then(r => r.json() as Promise<NuxtAppManifest>)
```
pi0, line 27: One immediate downside: There won’t be any retry mechanisms anymore. 

Does it really improve performance or any compatibility ?

danielroe: oh, that's a very good point ...

it slightly improves performance simply by reducing the number of functions called (as `fetch` underlies `ofetch` anyway) - and removes possible edge cases that might impair the initial load

of course, I would prefer to see this 'fixed' in `ofetch` if there is in fact any bug (such as passing `method: undefined` as the issue suggests) ...

danielroe: yes, that is my impression

InvincibleJun:
thanks

PR #29054, opened 2024-09-18
fix(nuxt): avoid throwing 404 error before middleware finishes

### LINKED ISSUE #27903: 404 error when using navigateTo in a middleware
### Environment

- Node Version:     v21.6.2
- Nuxt Version:     3.12.2
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  bun@1.1.15
- Builder:          -
- User Config:      telemetry, devtools, typescript, modules, ssr, i18n, app, image, security, pwa, googleFonts, ui, colorMode, vcalendar, css, vite, nitro, runtimeConfig, eslint, tailwindcss      
- Runtime Modules:  nuxt-security@1.4.3, @vite-pwa/nuxt@0.8.1, @nuxtjs/i18n@8.3.1, @nuxt/ui@2.17.0, @nuxtjs/google-fonts@3.2.0, @pinia/nuxt@0.5.1, @nuxt/image@1.7.0, @samk-dev/nuxt-vcalendar@1.0.3,
 @vueuse/nuxt@10.11.0, nuxt-applicationinsights@0.3.4, @nuxt/eslint@0.3.13
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-mz7brn?file=package.json

Run the app, navigate to `/auth/login`, you're automatically redirected to `/fr/auth/login`
Now change the version in package.json to Nuxt 3.12+ (3.12.2 for example), do the same thing and you get a 404 error.

### Describe the bug

After upgrading to Nuxt 3.12, Nuxt throws a 404 error when trying to redirect from a page that doesn't exist in the router using a middleware.
SSR is disabled on some specific routes, through Nitro config, which seems to be the source of the issue.
```
  nitro: {
    routeRules: {
      '/': {
        ssr: false,
      },
      '/**': {
        ssr: false,
      },
    },
  },
```

EDIT: this can even be simplified. Just having `ssr: false` in the Nuxt config causes the issue.

### Additional context

_No response_

### Logs

> [nuxt] error caught during app initialization Error: Page not found: /auth/login

### LINKED ISSUE #29051: 404 error screen appears after navigating to a non-existing page even when there is a global middleware that redirects to an existing page
### Environment

- Node Version:     v20.17.0
- Nuxt Version:     3.13.2
- CLI Version:      3.13.2
- Nitro Version:    2.9.7
- Builder:          -
- User Config:      compatibilityDate, ssr
- Runtime Modules:  -
- Build Modules:    -
The bug does not depend on OS because it was reproduced in an app deployed via GitHub Actions

### Reproduction

Demo repo: https://github.com/nicky1038/nuxt-404-demo
The app is deployed via GitHub Actions: https://nicky1038.github.io/nuxt-404-demo

Reproduction is simple:
* go to https://nicky1038.github.io/nuxt-404-demo/non-existing (or http://localhost:3000/non-existing if you launched the dev server)

404 screen will appear but it shouldn't, since there should be a redirect to `/existing-page`

It is noteworthy that:
* nuxt 3.12.0 is the first version with this bug
* this bug appears only when `ssr: false` is set

### Describe the bug

Since nuxt 3.12, when using `ssr: false` option, there is a bug.

Before 3.12, a global middleware could be used to redirect users from non-existing pages to existing ones, without the need to create dummies for that non-existing pages.

Now, even if there is such a global middleware with redirects to existing pages, there is a 404 error, when the page that user visits at first does not exist. 

### Additional context

_No response_

### Logs

_No response_

wattanx:
### 🔗 Linked issue
Fixes: https://github.com/nuxt/nuxt/issues/29051
Fixes: https://github.com/nuxt/nuxt/issues/27903

### 📚 Description

It seems that since Nuxt 3.12.0, there are cases where a 404 error occurs before a global middleware redirect.
This is because the 404 is thrown before the middleware is executed.
ref: https://github.com/nuxt/nuxt/pull/27016


danielroe:
I think maybe these lines prevent rendering HTML that won't be returned. Or the following would occur:

- we know there's a 404 but go ahead and render `app.vue` -> this needs to be avoided
- we then render `error.vue`

If that's the case we could guard with `if (import.meta.server)`.

We also need to verify that removing them will still show a 404 page correctly if rendering with `ssr: false`.

wattanx:
@danielroe 
Thank you! I'm thinking of adding tests, but are there any tests written for `ssr: false` somewhere?

danielroe:
We use `routeRules` to test it within the basic fixture.

wattanx:
Due to this fix, a 404 was no longer displayed. The reason the test is not failing is that if `[...slug].vue` exists and the result of validate is false, a 404 is displayed.

packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -252,6 +244,19 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
       await nuxtApp.callHook('page:loading:end')
     })
 
+    router.afterEach(async (to, _from) => {
+      if (to.matched.length === 0) {
+        await nuxtApp.runWithContext(() => showError(createError({
```
wattanx, line 249: If this process is registered before `beforeEach`, it will be executed before the global middleware. Therefore, it needs to be registered after `beforeEach`.

danielroe: is it sufficient to move it?

wattanx: Yes, simply moving this was sufficient. If there is no middleware that performs a redirect, a 404 is displayed. 
If there is middleware that performs a redirect, I have confirmed that the issue is resolved with the redirect.

This issue wasn't detected in the tests because [...slug].vue exists, so I'm considering fixing the tests, but I plan to do it in a separate PR.

packages/nuxt/src/pages/runtime/plugins/router.ts
```diff
@@ -148,16 +148,8 @@ const plugin: Plugin<{ router: Router }> = defineNuxtPlugin({
...
-            fatal: false,
-            statusMessage: `Page not found: ${to.fullPath}`,
-            data: {
-              path: to.fullPath,
-            },
-          })))
-        } else if (import.meta.server && to.redirectedFrom && to.fullPath !== initialURL) {
+
+        if (import.meta.server && to.redirectedFrom && to.fullPath !== initialURL) {
           await nuxtApp.runWithContext(() => navigateTo(to.fullPath || '/'))
```
danielroe, line 153: should this also be moved?

danielroe (approved):

PR #29100, opened 2024-09-20
docs: move custom directories note to the correct place

cernymatej:
### 📚 Description

This PR moves the note related to Custom Directories to the appropriate section, instead of being under npm Packages.


danielroe (approved):
Thank you ❤️

PR #29118, opened 2024-09-23
docs: update to new `ofetch` headers for interceptors

atinux:
I guess we should wait until we force `ofetch@^1.4.0`

Related PR: https://github.com/unjs/ofetch/pull/436


pi0 (dismissed):
LGTM thanks ❤️ (i also suggest to wait, because while new change is backward compatible, some users might be using < 1.4.0 and input can be of type HeadersInit sometimes)

danielroe (approved):

PR #29165, opened 2024-09-25
refactor(nuxt,schema,vite,webpack): use unplugin for vfs

danielroe:
### 🔗 Linked issue

context: https://github.com/nuxt/nuxt/pull/29142

### 📚 Description

this migrates away from `webpack-virtual-modules` in favour of using unplugin to provide the nuxt vfs. it means we can use a consistent implementation that is bundler-agnostic.

Some key differences:
- internal vfs paths now _always_ include file extension (previously `#build/<filename>` paths stripped it - this should not make a difference to resolution but might if manually accessing `nuxt.vfs`
- we now respect nuxt extensions
- we now add two different vfs plugins - one for server + one for client - this will likely _address_ potential bugs
- we now support a convention of `.client` and `.server` suffixes in the vfs for conditional resolution in server vs client
- `#build` is now in `nuxt.options.alias` again
- we have added a null byte - `\0virtual:nuxt:` is the prefix - in line with vite convention and other plugins

### 🚧 TODO

- [x] tailwindcss
- [ ] i18n module


danielroe:
current test failures:
- [nuxt/examples](https://github.com/nuxt/ecosystem-ci/actions/runs/11048285101/job/30692159281)
- [nuxt-modules/i18n](https://github.com/nuxt/ecosystem-ci/actions/runs/11048285101/job/30692165553)
> https://github.com/nuxt/ecosystem-ci/actions/runs/11048285101

packages/nuxt/src/core/plugins/virtual.ts
```diff
@@ -0,0 +1,74 @@
...
+import { createUnplugin } from 'unplugin'
+
+const PREFIX = '\0virtual:nuxt:'
+
+interface VirtualFSPluginOptions {
+  mode: 'client' | 'server'
+  alias?: Record<string, string>
+}
+
+export const VirtualFSPlugin = (nuxt = useNuxt(), options: VirtualFSPluginOptions) => createUnplugin(() => {
```
HigherOrderLogic, line 12:
```suggestion
export const VirtualFSPlugin = (nuxt, options: VirtualFSPluginOptions) => createUnplugin(() => {
```

Isn't this redundant? Since you don't normalize the options down the line inside the function, you'll always have to pass the `options` argument, which require you to pass the `nuxt` argument first.

Issue #22661, opened 2023-08-16
routeRules prerender not work


a1mersnow:
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.20.1`
- Nuxt Version:     `3.6.5`
- Nitro Version:    `2.5.2`
- Package Manager:  `pnpm@8.6.12`
- Builder:          `vite`
- User Config:      `app`, `ui`, `devtools`, `typescript`, `vue`, `modules`, `content`, `colorMode`, `vite`
- Runtime Modules:  `@nuxt/content@2.7.2`, `@nuxthq/ui@2.7.0-28200741.b8b7d74`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-2mnq8w-9arzti?file=nuxt.config.ts

### Describe the bug

There are no `.html` files in `.output` dir.

danielroe:
Check `.output/public/index.html`.

a1mersnow:
@danielroe But, there is not any `.html` file.

a1mersnow:
@danielroe I found the problem! The config below doesn't generate any html files:
```js
routeRules: {
  '/**': { prerender: true },
},
```

So I guess the `prerender` doesn't support glob pattern, right?

So is there any solution for this(using glob pattern for prerender)?

danielroe:
If you pass a glob pattern, then it will prerender any routes _matching_ that glob pattern that are discovered whilst crawling your HTML pages. But it doesn't know to prerender the root path. So additionally you will need to set `nitro.prerender.routes` to `['/']`.

richgcook:
Sorry to bring this back up (but better than making a new issue, right?) but the following routeRules doesn't prerender any routes – is this sort of approach not possible?

```
routeRules: {
	// Prerender all routes
	'/**': { prerender: true },
	// Exclude API routes from prerendering
	'/api/**': { prerender: false },
	// API routes that fetch Are.na content use SWR (cached for 1 hour)
	'/api/arena/**': { swr: 3600 },
}
```

danielroe:
@richgcook You still need to pass a starting URL. Does my comment above make sense?

richgcook:
@danielroe Thanks very much for the reply. Yeah, I thought this would've solved it but it only prerenders that one route.

```js
nitro: {
	prerender: {
		routes: ['/']
	}
},

routeRules: {
	// Prerender all routes
	'/**': { prerender: true },
	// Exclude API routes from prerendering
[...]

PR #29172, opened 2024-09-26
fix(nuxt): use `routeRules` to hint pages to prerender

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/22661#issuecomment-2376237042

### 📚 Description

This addresses an overlooked area - when a user is building a site and wants to prerender some routes but has just specified the route rule with a wildcard _and not_ added `prerender.routes`.

This won't cover every area (such as routes changed with `app/router.options.ts`) but it should be safe as it requires an overlap between `pages/` and `routeRules`.

If you have a `pages/bing/foo.vue`, then the following config will ensure it is prerendered when running `nuxi build`:

```ts
  routeRules: {
    '/bing/**': { prerender: true },
  },
})
```

However, note that this still won't help with non-deterministic pages, so we still can't know the paths to prerender if you have `pages/bing/[slug].vue` - you'll still need to provide a starting point in this case.

### 🚧 TODO

- [x] add a similar route-rule-aware approach to the runtime hinting.


richgcook:
> However, note that this still won't help with non-deterministic pages, so we still can't know the paths to prerender if you have `pages/bing/[slug].vue` - you'll still need to provide a starting point in this case.

Good point. When we use `nitro.prerender.crawlLinks: true` it does the following (from the docs):

```
1. Load the HTML of your application's root route (/), any non-dynamic pages in your ~/pages directory, and any other routes in the nitro.prerender.routes array.
2. Save the HTML and payload.json to the ~/.output/public/ directory to be served statically.
3.Find all anchor tags (<a href="...">) in the HTML to navigate to other routes.
4. Repeat steps 1-3 for each anchor tag found until there are no more anchor tags to crawl.
```

Is there the capability for something like `'/bing/**': { prerender: true },` to use the same workings as `crawlLinks` so it would run through these steps but just for that route so it would pick up on the non-deterministic pages?

Ultimately if I keep `nitro.prerender.crawlLinks` set to true, which is fine, but any `routeRules` set to make one of those routes ssr or swr then it shouldn't prerender those overrides, right?

richgcook:
I have a query about whether this fix impacts the following:

```
nitro: {
	prerender: {
		crawlLinks: true
	}
},

routeRules: {

	'/making-public/reading-room-library': { ssr: true },

	// Exclude API routes from prerendering
	'/api/**': { prerender: false },
	// API routes that fetch Are.na content use SWR (cached for 1 hour)
	'/api/arena/**': { swr: 3600 }, // 3600 seconds = 1 hour

	// Redirects
	'/program/labs': { 
		redirect: {
			to: '/program',
			statusCode: 301 
		},
	},
	'/community/profiles': {
		redirect: {
			to: '/community',
			statusCode: 301
		},
	},

},
```

When I run npm run build it prerenders the '/making-public/reading-room-library' route rather than skipping and running it as ssr. Any thoughts?

danielroe:
ssr doesn't mean not to pre-render. it means that when prerendered, it will have the full html of the page present, not an empty div

richgcook:
Thanks, @danielroe

Can I confirm that if I need a route to always be up to date (always fetching "live" data) what's the best route rule to use? prerender: false? 

I think I'm slightly confused because if you were to run nuxi build without any route rules then you'd have no static prerendered files, right? That's ssr out of the box. So when setting ssr: true I'd assume it would yield the same results and therefore on said route it wouldn't produce a .html file. Am I missing something? Am I mad? 😂

 The /server/api route that I'm fetching would be set to swr: 3600 so it's cached every hour(ish)

richgcook:
If we set `nitro.prerender.crawlLinks: true` then setting a route (via routeRules) to prerender: false is essentially making the route server side rendered. Correct? However, you're suggesting there seems to be an "in between state" where setting `ssr: true` with `crawlLinks: true` on a route prerenders that route but then what's server side rendered?

danielroe:
`ssr: true` is default
`prerender: false` is default

If you set `crawlLinks: true` then it's like enabling `prerender: true` for every route.

SSR refers to whether full HTML is rendered by the server, at whatever time that render happens. It's not related to whether that render happens at build time or per-request.

richgcook:
Thanks. That totally clears things up. Now I also see why `/**` didn't work without `crawlLinks`.

This is an example of what I ended up using with `nitro.prerender.crawlLinks: true`

```js
routeRules: {

    // Exclude specific routes from pre-rendering (fresh data on each request)
    '/events': { prerender: false },
    '/page-that-uses-api-for-example': { prerender: false },
    ...
    
    // Disable pre-rendering for all API routes
    '/api/**': { prerender: false },
    
    // Apply SWR caching (1 hour) for API routes
    '/api/some-api/**': { swr: 3600 },
    
    // Redirects
    ...

},
```

PR #29180, opened 2024-09-26
test: add tests to check 404 handling in client-side rendering

wattanx:
### 📚 Description
Hi ✋ 
Before adding https://github.com/nuxt/nuxt/pull/29054, I would like to add tests to check whether 404 handling in client-side rendering behaves as intended.

I would actually like to add tests for the case where there are no catch-all routes, but it was difficult because there are tests that depend on catch-all routes.

### DIFF
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -249,6 +249,16 @@ describe('pages', () => {
     await serverPage.close()
   })
 
+  it('should render 404 page even in spa mode', async () => {
+    const { page } = await renderPage('/route-rules/spa')
+
+    await page.getByText('should throw a 404 error').click()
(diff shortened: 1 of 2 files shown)


wattanx:
If it's okay, I can create a fixtures directory without catch-all routes.

wattanx:
It seems that the test cases already exist 🙈 
https://github.com/nuxt/nuxt/blob/main/test/basic.test.ts#L195-L211

danielroe:
we probably still need to cover the ssr: false example from the issue your PR resolves (maybe use a route rule to force the ssr: false?)

also maybe we could update the fixture so the catch all was only present/used for a subpath - because you are right that it conceals errors at the moment

wattanx:
I'm thinking of using a subpath! However, I'm currently investigating because other tests are breaking.

PR #29280, opened 2024-10-05
refactor: update jiti

pi0:
v3: #29281

Update jiti to 2.3 and other deps that use jiti v2 (c12, unbuild, untyped)

To avoid behavior changes, new `jiti.import(id, { default: true }` is used that explicitly asks for "default export" (rather than interop proxy).


pi0:
waiting for CI but locally seems good

danielroe (approved):
thank you! ❤️

PR #29287, opened 2024-10-05
fix(nuxt): error on build when required module is missing

### LINKED ISSUE #29283: Auto install `@nuxt/image` prompts hangs `nuxt build`
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.17.0`
- Nuxt Version:     `3.13.2`
- CLI Version:      `3.14.0`
- Nitro Version:    `2.9.7`
- Package Manager:  `bun@1.1.29`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

- https://github.com/unjs/h3/tree/7324eeec854eecc37422074ef9f2aec8a5e4a816/docs
- `bun install` > `bun run build`

### Describe the bug

Even though no `<NuxtImg>` is used, Nuxt seems wrongly detecting usage it and prompts to install (which is another issue)

When running build command, interactive prompt is asked but also not even awaited (IMO we should avoid any prompt) and build will be stuck

Image

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/29283

### 📚 Description

This prevents a hanging prompt in build when stubbed components are being used. Instead an error will be thrown.

### DIFF
--- a/packages/nuxt/src/components/plugins/loader.ts
+++ b/packages/nuxt/src/components/plugins/loader.ts
@@ -2,7 +2,7 @@ import { createUnplugin } from 'unplugin'
 import { genDynamicImport, genImport } from 'knitwork'
 import MagicString from 'magic-string'
 import { pascalCase } from 'scule'
-import { resolve } from 'pathe'
+import { relative, resolve } from 'pathe'
 import type { Component, ComponentsOptions } from 'nuxt/schema'
 
 import { logger, tryUseNuxt } from '@nuxt/kit'
@@ -21,6 +21,7 @@ export const LoaderPlugin = (options: LoaderOptions) => createUnplugin(() => {
   const exclude = options.transform?.exclude || []
   const include = options.transform?.include || []
   const serverComponentRuntime = resolve(distDir, 'components/runtime/server-component')
+  const nuxt = tryUseNuxt()
 
   return {
     name: 'nuxt:components-loader',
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/components/plugins/loader.ts
```diff
@@ -46,10 +47,14 @@ export const LoaderPlugin = (options: LoaderOptions) => createUnplugin(() => {
...
-          // @ts-expect-error TODO: refactor to nuxi
-          if (component._internal_install && tryUseNuxt()?.options.test === false) {
-            // @ts-expect-error TODO: refactor to nuxi
-            import('../../core/features').then(({ installNuxtModule }) => installNuxtModule(component._internal_install))
+          // TODO: refactor to nuxi
+          const internalInstall = ((component as any)._internal_install) as string
+          if (internalInstall && nuxt?.options.test === false) {
+            if (!nuxt.options.dev) {
+              const relativePath = relative(nuxt.options.rootDir, id)
+              throw new Error(`[nuxt] \`~/${relativePath}\` is using \`${component.pascalName}\` which requires \`${internalInstall}\``)
```
pi0, line 55: What if the component is registred at runtime (in a plugin). Would we still strictly fail builds?

danielroe: Yes. Because the transform has already taken place and injected `<NuxtImg>` pointing to the stub component provided by Nuxt, and if used it would throw an error:

https://github.com/nuxt/nuxt/blob/f209158352b09d1986aa320e29ff36353b91c358/packages/nuxt/src/app/components/nuxt-stubs.ts#L3-L17

pi0: Ah i see it is really there!

PR #29292, opened 2024-10-06
docs: consistent directory structure

AcelisWeaven:
### 🔗 Linked issue

None

### 📚 Description

While reading the docs, I've noticed some inconsistencies in the way directory structures were shown. This MR attempts to use the same convention everywhere.

I tried to use the most common one through the docs:

```bash [Directory Structure]
-| some-app/
---| nuxt.config.ts
---| pages
-----| app-page.vue
-| nuxt.config.ts
```

### DIFF
--- a/docs/1.getting-started/5.routing.md
+++ b/docs/1.getting-started/5.routing.md
@@ -15,7 +15,7 @@ This file system routing uses naming conventions to create dynamic and nested ro
 ::code-group
 
 ```bash [Directory Structure]
-| pages/
+-| pages/
 ---| about.vue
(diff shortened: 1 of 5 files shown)


AcelisWeaven:
I've noticed I missed some places, so I'll get back to it 🙂

AcelisWeaven:
Hi @danielroe, it's now ready to review 🙏 I missed a trailing slash in the very example I used in the first place. I double checked and it seems it was the only place missing one.

danielroe (approved):
thank you!

Issue #24332, opened 2023-11-16
Re-use `getCachedData` when `refresh`ing `asyncData`


TheAlexLichter:
### Describe the feature

Hey 👋🏻

With the introduction of `getCachedData`, `useFetch` and `useAsyncData` have an elegant way to re-use existing data easily.

There is a downside though, as @broox [mentioned in this YouTube comment](https://www.youtube.com/watch?v=aQPR0xn-MMk&lc=UgyH9sQQrLwAG9NGoEh4AaABAg):

When **refreshing the data**, either via calling `refresh` manually or implicitly using it through the `watch` property or by defining reactive parameters like `query` in the `useFetch` options, there is **no option to re-use the cached data**.

Reason behind it is that we only take cached data into account when nuxt is hydrating *or* on the initial request of the composable. https://github.com/nuxt/nuxt/blob/a087b87d3ff7f64c779199f0b4f1a866625671fa/packages/nuxt/src/app/composables/asyncData.ts#L182-L184

Spontaneous idea to resolve this issue:

1. Add a `force` parameter for `refresh`, so manual `refresh()` calls can bypass the `getCachedData` function if necessary. By default, this can be set to `true` if we want to avoid breaking changes.
3. Set `force` to `false` by default for `watch`. This is a behavior change, but when implementing an own `getCachedData` function you can take this into account easily in there. Projects not using the `getCachedData` function should not be impacted.

More than happy for other ideas or feedback ☺️

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

broox:
in my opinion, it should be default behavior that all requests, even those made by reactive changes, use the cache if your definition of `getCachedData()` resolves cached data.

manniL:
> i.e. the definition of getCachedData() should be the "forcing" function to use the cache.

I think that sounds fair, if the function is specified and returns data, it should be used instead of calling again by default. 

But it'd be still nice to have a chance to bypass the function "on demand" (which would be equivalent to a hard reload or similar). I wonder if we need the `force` parameter for that though or if that could be implemented in userland (which I guess it could be easily. 🤔

broox:
[...]

PR #29300, opened 2024-10-07
docs: update use-async-data.md

TimGuendel:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/24332

### 📚 Description

Note that getCachedData does not work with watch and execute/refresh (for now)


danielroe (commented):
I would like to fix this issue rather than document the bug, ideally 🙏

TimGuendel:
Unless you are close to fixing it, I would have appreciated this behaviour to be documented. It would have saved some time 🥲

danielroe:
If it is documented, then changing it becomes a breaking change.

TimGuendel:
All right ✅

PR #29351, opened 2024-10-09
docs: use text fragments for permalinks

danielroe:
### 🔗 Linked issue

https://github.com/unjs/automd/issues/77

### 📚 Description

thanks to a suggestion from @pi0 we can improve links to github source code that is a bit more resilient to changes in source files ...

### DIFF
--- a/docs/1.getting-started/12.upgrade.md
+++ b/docs/1.getting-started/12.upgrade.md
@@ -589,9 +589,9 @@ These options have been set to their current values for some time and we do not 
 
 ##### Migration Steps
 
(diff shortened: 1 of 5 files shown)


manniL:
Do you think the trade-off is worth it (possibly "older" file but with line reference, vs. newer file but no line reference)?

pi0:
Glad you liked the idea! I deleted my comment because found it is supported in Chromium-based browsers but it seems a reasonable compromise as is always up to date.

onmax:
Hey! I am implemeting the new generator in automd as we speak :)


If it is ok with the team, I would like to also implement the generator in the docs in the Nuxt repo if that's ok. Thank you!

PR #29366, opened 2024-10-10
feat(nuxt): add `isolate` page meta

### LINKED ISSUE #29365: Allow injecting pages without inheriting from `app.vue`
### Describe the feature

Hello!

I'm currently working on implementing devtools for Nuxt UI (PR: [#2196](https://github.com/nuxt/ui/pull/2196)). As part of this, I need to inject a page within the user's app that allows our devtools to render components based on the user's Nuxt UI configuration.

The challenge I'm encountering is that I could not find a way to inject this page without inheriting from the user's `app.vue` (using [`extendPages`](https://nuxt.com/docs/api/kit/pages#extendpages)) , which could cause unintended side effects.

To address this, I thought we could introduce a way to inject pages without inheriting from `app.vue`, similar to what’s done in for the [test utils](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/components/nuxt-root.vue#L43), by rendering the nuxt page directly if it matches a specific condition.

### Additional information

### Final checks

romhml:
### 🔗 Linked issue

resolves #29365 

### 📚 Description
This PR introduces a new page meta `isolate` to render pages without inheriting from the `app.vue` component.   

## Release Notes

- **New Features**
  - Introduced the `isolate` property in the `PageMeta` interface, allowing pages to render independently from the default layout.
  - Added new components: `isolated.vue` and `not-isolated.vue`, demonstrating the use of the `isolate` property.
  - Implemented a new template for isolated pages to enhance rendering control.

- **Bug Fixes**
  - Improved rendering logic to ensure isolated pages do not inherit settings from the main app.

- **Tests**
  - Added a test case to verify rendering behavior for isolated and non-isolated pages.

- **Documentation**
  - Updated documentation for the `definePageMeta` function to include the new `isolate` property.

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-root.vue
+++ b/packages/nuxt/src/app/components/nuxt-root.vue
@@ -13,6 +13,7 @@
       :is="SingleRenderer"
       v-else-if="SingleRenderer"
     />
+    <NuxtPage v-else-if="route?.meta?.isolate" />
     <AppComponent v-else />
   </Suspense>
 </template>
@@ -43,8 +44,9 @@ const url = import.meta.server ? nuxtApp.ssrContext.url : window.location.pathna
 const SingleRenderer = import.meta.test && import.meta.dev && import.meta.server && url.startsWith('/__nuxt_component_test__/') && defineAsyncComponent(() => import('#build/test-component-wrapper.mjs')
   .then(r => r.default(import.meta.server ? url : window.location.href)))
 
+const route = useRoute()
 // Inject default route (outside of pages) as active route
-provide(PageRouteSymbol, useRoute())
+provide(PageRouteSymbol, route)
 
 // vue:setup hook
 const results = nuxtApp.hooks.callHookWith(hooks => hooks.map(hook => hook()), 'vue:setup')
--- a/packages/nuxt/src/app/components/nuxt-root.vue
+++ b/packages/nuxt/src/app/components/nuxt-root.vue
@@ -13,7 +13,10 @@
       :is="SingleRenderer"
       v-else-if="SingleRenderer"
     />
-    <NuxtPage v-else-if="route?.meta?.isolate" />
+    <component
+      :is="IsolatedPage"
+      v-else-if="IsolatedPage && route?.meta?.isolate"
+    />
     <AppComponent v-else />
   </Suspense>
 </template>
@@ -47,6 +50,8 @@ const SingleRenderer = import.meta.test && import.meta.dev && import.meta.server
 const route = useRoute()
 // Inject default route (outside of pages) as active route
 provide(PageRouteSymbol, route)
+
+const IsolatedPage = defineAsyncComponent(() => import('#build/isolated-page.mjs'))
 
 // vue:setup hook
 const results = nuxtApp.hooks.callHookWith(hooks => hooks.map(hook => hook()), 'vue:setup')
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -10,7 +10,6 @@ import { filename } from 'pathe/utils'
 import type { NuxtTemplate } from 'nuxt/schema'
 import type { Nitro } from 'nitro/types'
 
-import { distDir } from '../dirs'
 import { annotatePlugins, checkForCircularDependencies } from './app'
 import { EXTENSION_RE } from './utils'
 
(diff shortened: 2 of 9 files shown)


packages/nuxt/src/app/components/nuxt-root.vue
```diff
@@ -13,6 +13,7 @@
       :is="SingleRenderer"
       v-else-if="SingleRenderer"
     />
+    <NuxtPage v-else-if="route?.meta?.isolate" />
```
huang-julien, line 16: It's probably better to use the lazy one

romhml: Do you mean to use `LazyNuxtPage` instead? Doing this breaks a few more tests on the bundle:

```bash
Error: Snapshot `minimal nuxt application > default client bundle size 1` mismatched

Expected: ""116k""
Received: ""119k""
```

Is this kind of thing expected? Should I simply update these tests?

HigherOrderLogic: For that test, just push and the CI will simply fix it for you.

packages/nuxt/src/app/components/nuxt-root.vue
```diff
@@ -13,6 +13,7 @@
       :is="SingleRenderer"
       v-else-if="SingleRenderer"
     />
+    <LazyNuxtPage v-else-if="route.meta?.isolate" />
```
danielroe, line 16: this implementation will result in always bundling `<NuxtPage>`, even if the `vue-router` integration isn't enabled.

we should probably protect it so that if no routes have the `isolate` metadata then there will be no changes to the bundle at all

packages/nuxt/src/core/templates.ts
```diff
@@ -553,3 +554,11 @@ export const buildTypeTemplate: NuxtTemplate = {
     return declarations
   },
 }
+
+export const isolatedPageTemplate: NuxtTemplate = {
+  filename: 'isolated-page.mjs',
+  getContents (ctx) {
+    const hasPages = ctx.nuxt.options.dev || ctx.app.pages?.length
```
romhml, line 561: I initially planned on using `ctx.app.pages?.some((page) => page.meta?.isolate))`   but the metas are not completely resolved at this stage. I'm checking if the app has some pages instead which should be fine since this means the vue-router integration is enabled and the NuxtPage component is already bundled.

danielroe:
I'm thinking one way to do this would be

- update SingleRenderer to render any component that has `renderMode: 'isolated'` (or equivalent) in the component options (in production too)
- update the `fetchComponent` implementation in nuxt/test-utils to pass a component name and not file path - and allow specifying opt-in components in test setup options (or possibly setting to `components: true`)

Would this be acceptable? It would basically hook into existing functionality (but make it more secure by requiring a defined list of components that could be rendered this way).

It would also mean you would access this component at `/__nuxt_component/` or equivalent, which wouldn't be great for end-user DX, but would probably be fine for a devtools integration.

packages/nuxt/src/app/components/nuxt-root.vue
```diff
@@ -13,6 +13,10 @@
       :is="SingleRenderer"
       v-else-if="SingleRenderer"
     />
+    <component
+      :is="IsolatedPage"
+      v-else-if="IsolatedPage && route?.meta?.isolate"
```
huang-julien, line 18: Isolated page will always be truthy because it is an async loader

romhml:
Thanks for the suggestion @danielroe! This should work perfectly for my use-case, but I'm thinking it might be a bit specific. The current implementation is a bit more flexible and could be reused by other module authors to inject pages (for devtools, auth, etc.). What's bothering you with the current solution?

I'll look into implementing your suggestion so we can compare.

danielroe:
`NuxtPage` is not available unless you have a `pages/` directory, so using it (even on a conditional basis) will not work. Even if it did work, it would increase the bundle size even for people not using isolated rendering.

Although it might seem less flexible to implement `/__nuxt_component`, the idea would be to implement a limited functionality and then later build on top of it (for example, by adding an integration in the pages module that extends the simpler first step). Also, we do not currently have an idea how many people would need/use isolated rendering, so starting simple and adding more functionality down the line makes more sense to me.

romhml:
Thanks, I understand. We can close this PR then, I'll work on implementing it using `/__nuxt_component` in another one.

danielroe:
No, I don't think we should use the `components.json` template. It's not always generated. But you could create a new template if needed:

https://github.com/nuxt/nuxt/blob/0f897a056e8a4ae2565e2cf96f39de878f5cad44/packages/nuxt/src/components/module.ts#L134-L137

PR #29594, opened 2024-10-18
docs: update 3.custom-usefetch.md

jkpeyi:
The options.headers.set  has some limitations as the type of options.headers vary. So i suggest a new method of passing the headers to the request.


danielroe:
This was a deliberate change and relies on `ofetch>= 1.4.0`.

If it's not working with that version, please let me know 🙏 

The change in this PR would result in a change of behaviour as the header would be fixed when `$fetch.create` is called, and not up-to-date at the time of request.

PR #29617, opened 2024-10-20
fix: use component html string inside payload

### LINKED ISSUE #29345: NuxtIsland: Failed to locate Teleport target with selector
### Environment

Nuxt project info:                                                                                                                                                   10:52:59 AM

- Node Version:     v20.15.0
- Nuxt Version:     3.13.2
- CLI Version:      3.14.0
- Nitro Version:    2.9.7
- Package Manager:  pnpm@9.9.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-d9qil3

Be sure to open in full page, and open the dev console

### Describe the bug

`<SomeComponent>.server.vue` actually still works, but if it contains a **selectiveClient** / `nuxt-client` element, this part is now broken, until you refresh or navigate to another page, using links and not the browser buttons (well refresh does work)

Whether or not the server component contains a **selectiveClient** is irrelevant to the error message in the console, it will still throw that error

However if it does contain a **selectiveClient**, then that part of the server component, is now broken

### Additional context

When using normal links (either a href, or <NuxtLink>) this does not happen

It only happens when hitting the back button in the browser (or shortcut on mouse)

EDIT:
Actually it does happen with NuxtLink's, the href is simply reloading the whole page, that's why it's not happening there

### Logs

runtime-core.esm-bundler.js?v=c0e84aa4:50 [Vue warn]: Failed to locate Teleport target with selector "div[data-island-uid='1'][data-island-component="CounterButton-hb1RAsRv3M"]". Note the target element must exist before the component is mounted - i.e. the target cannot be rendered by the component itself, and ideally should be outside of the entire Vue component tree. 
  at <NuxtIsland name="HelloIsland" > 
  at <Index onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< Proxy(Object) {__v_skip: true} > > 
  at <Anonymous key="/" vnode= {__v_isVNode: true, __v_skip: true, type: {…}, props: {…}, key: null, …} route= {fullPath: '/', path: '/', query: {…}, hash: '', name: 'index', …}  ... > 
  at <RouterView name=undefined route=undefined > 
  at <NuxtPage> 
  at <SideMenu ref=Ref< Proxy(Object) {…} > > 
  at <LayoutLoader key="side-menu" layoutProps= {ref: RefImpl} name="side-menu" > 
  at <NuxtLayoutProvider layoutProps= {ref: RefImpl} key="side-menu" name="side-menu"  ... > 
  at <NuxtLayout> 
  at <App key=4 > 
  at <NuxtRoot>

### LINKED ISSUE #29616: Nuxt Island <slot/> fails to render inside teleport target
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     ^3.13.2
- CLI Version:      3.14.0
- Nitro Version:    -
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-qvc8ul?file=pages%2Findex.vue

1. Run the reproduction
2. Navigate to /other
3. Go back to /
4. Error is thrown:  `Failed to locate Teleport target with selector "div[data-island-uid="1"][data-island-slot="default"]"`

### Describe the bug

Inside a NuxtIsland component, the `<slot />` isn't being rendered when navigating to another page, then navigating back. The issue seems to stem from the fact that the `ssrHTML` ref remains empty upon navigation because the `instance.vnode.el` is still null, which prevents the SSR `uid` from being captured correctly:

https://github.com/nuxt/nuxt/blob/49242b25008d5593d1a279f030dac8936712cfe9/packages/nuxt/src/app/components/nuxt-island.ts#L129-L136

I've currently fixed the issue by simply setting the `ssrHTML` ref to be `nuxtApp.payload.data[key]?.html || ''` upon initialization. I'm happy to provide a PR if that seems like a viable solution to you.

### Additional context

_No response_

### Logs

_No response_

yassilah:
### 🔗 Linked issue

resolves #29616 
resolves https://github.com/nuxt/nuxt/issues/29345

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -126,11 +126,11 @@ export default defineComponent({
       }
     }
 
-    const ssrHTML = ref<string>('')
+    const key = `${props.name}_${hashId.value}`
+    const ssrHTML = ref<string>(nuxtApp.payload.data[key]?.html || '')
 
-    if (import.meta.client && instance.vnode?.el) {
+    if (import.meta.client && instance.vnode?.el && !ssrHTML.value) {
       ssrHTML.value = getFragmentHTML(instance.vnode.el, true)?.join('') || ''
-      const key = `${props.name}_${hashId.value}`
       nuxtApp.payload.data[key] ||= {}
       nuxtApp.payload.data[key].html = ssrHTML.value
     }
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -126,13 +126,13 @@ export default defineComponent({
       }
     }
 
-    const key = `${props.name}_${hashId.value}`
-    const ssrHTML = ref<string>(nuxtApp.payload.data[key]?.html || '')
+    const cmpKey = `${props.name}_${hashId.value}`
+    const ssrHTML = ref<string>(nuxtApp.payload.data[cmpKey]?.html || '')
 
     if (import.meta.client && instance.vnode?.el && !ssrHTML.value) {
       ssrHTML.value = getFragmentHTML(instance.vnode.el, true)?.join('') || ''
-      nuxtApp.payload.data[key] ||= {}
-      nuxtApp.payload.data[key].html = ssrHTML.value
+      nuxtApp.payload.data[cmpKey] ||= {}
+      nuxtApp.payload.data[cmpKey].html = ssrHTML.value
     }
 
     const uid = ref<string>(ssrHTML.value.match(SSR_UID_RE)?.[1] ?? getId())


danielroe:
Would you have a look at the tests?

yassilah:
Hi! So I just took a look and there's an issue at this line:

https://github.com/yassilah/nuxt/blob/825b0bf2e38cea121c9635321772c4c0404e029d/test/basic.test.ts#L1881C5-L1881C10

The issue is that there are actually two elements that match `locator('.box').getByText('"number": 101,')` and not just one. But when looking at the page content, I would assume that is what should be expected since there's twice the same component?

https://github.com/yassilah/nuxt/blob/825b0bf2e38cea121c9635321772c4c0404e029d/test/fixtures/basic/pages/islands.vue#L18C1-L31C11

So I wonder if I should actually just update the test to:

`expect(await page.locator('.box').getByText('"number": 101,').all()).toHaveLength(2)`

Or if I should update my PR to prevent the `ssrHtml` ref from being initialized with the received payload when wrapped inside `<ClientOnly />`. If so I'd be curious to know why? Also, I believe that this specific case of having the same component being rendered in SSR and then the same component rendered inside a `<ClientOnly />` on the same page is very unlikely in a real project.

Please let me know what you think and I'm happy to do the necessary changes :)

Thanks again a million for your amazing work!

bernhardberger:
> Also, I believe that this specific case of having the same component being rendered in SSR and then the same component rendered inside a <ClientOnly /> on the same page is very unlikely in a real project.

Not entirelyimpossiblet hough. I exactly have this scenario in one of my projects (the same component is once used for displaying data from a CMS and on another place on the same page with data from localStorage).

bernhardberger:
PS: I think this also resolves https://github.com/nuxt/nuxt/issues/29345

huang-julien (approved):
TY and LGTM. Can you add a regression tests please 🙏 ?

danielroe:
We also need to handle the case of having duplicate components 🙏  as @bernhardberger points out.

Marking as draft so it's easier to see the state - let me/julien know if you need any help with these changes.

PR #29619, opened 2024-10-20
chore(vite): avoid redundant postcss plugins overwrite

GalacticHypernova:
### 📚 Description

Since the plugins array is initialized as empty, there is no need to make it empty again


danielroe (approved):

danielroe:
Would you be able to fix the type issue? 🙏

GalacticHypernova:
Sure thing!

danielroe (approved):

PR #29621, opened 2024-10-20
perf: use fixed-length arrays for known result sizes

GalacticHypernova:
### 📚 Description

Using fixed length arrays with direct assignment is very useful for performance when the output size is known. It allows for a constant O(1) as opposed to amortized O(1) like that of `Array.prototype.push`, which in hot code paths could make quite a big difference given it will not resize itself dynamically internally . This also optimizes its memory efficiency because the engine does not need to allocate more objects than necessary, and the GC has less work to do.


GalacticHypernova:
Once again, the ui-templates directory has been left untouched

packages/nuxt/src/app/components/nuxt-teleport-island-slot.ts
```diff
@@ -35,23 +35,23 @@ export default defineComponent({
     }
 
     return () => {
-      const vnodes: VNode[] = []
+      const vnodes: VNode[] = new Array(1)
```
danielroe, line 38: this array might have either 1 or 2 members - is it worth optimising it for 1?

GalacticHypernova: From the very scarce resources about the internal operations in arrays, in v8, JSC, and Bun, using `push` immediately allocates space for ~8-16 spaces, and then doubles capacity on each push where it exceeds the array length

If it ends up with 1 element, then it avoids those extra 7-15 spaces. If it ends up with 2 elements, it would allocate another single place, which means it would be left with exactly 2 spots for 2 elements, once again avoiding the extra space

packages/nuxt/src/components/templates.ts
```diff
@@ -118,20 +118,27 @@ export const componentsTypeTemplate = {
...
+      componentExports[i] = `${pascalName}: ${type}`
+    }
     return `
 import type { DefineComponent, SlotsType } from 'vue'
 ${nuxt.options.experimental.componentIslands ? islandType : ''}
 interface _GlobalComponents {
-  ${componentTypes.map(([pascalName, type]) => `    '${pascalName}': ${type}`).join('\n')}
-  ${componentTypes.map(([pascalName, type]) => `    'Lazy${pascalName}': ${type}`).join('\n')}
+    '${components.join('\n    \'')}
+    'Lazy${components.join('\n    \'Lazy')}
```
danielroe, line 133: I'm pretty sure there's no benefit to switching from `.map` as this already is able to optimise by creating an array with the correct number of elements.

GalacticHypernova: That's exactly the problem, it makes 4 total arrays and 4 total iterations.

Let's take the Vuetify Nuxt module for example, given that they're working on migrating from a custom Vite plugin to Nuxt's component system:

They have, from a rough estimate of mine, 76 components.

The current approach makes a total of 304 elements in 4 arrays (76 elements in each array) and 4 iterations over them (so 304 counts in total).

The new approach makes a total of 152 elements in 2 arrays (76 elements in each array) and 1 iteration only (so 76 counts in total).

All of this calculation does not even take into consideration the delayed hydration PR, which will introduce LazyIdle, LazyVisible, LazyPromise, LazyTime, LazyMedia, LazyEvent, and  LazyNever, which will make it 9 types in total (76 * 9 = 684 components in total)

PR #29638, opened 2024-10-22
fix(nuxt): change definePageMeta and vite handleHotUpdate to `full-reload` file

### LINKED ISSUE #24794: Add/fix HMR support for `definePageMeta`
### Environment

```
- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.8.2
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/github-bybsea?file=pages%2Findex.vue

### Describe the bug

Not sure if this is a known/intended limitation but HMR does not work for changes made to `definePageMeta`'s parameters. For example, if I change `layout: 'default'` to `layout: 'blue'` in the reproduction, the page does not update until I do a manual reload.

There are some DX scenarios where having HMR for `definePageMeta` parameters could be helpful (e.g. experimenting with different transition classes for page/layout transitions).

### Additional context

_No response_

### Logs

_No response_

xjccc:
### 🔗 Linked issue

resolves #24794

### 📚 Description

1. Remove unnecessary `module.ts` types
2. When change `definePageMeta`, vite `handleHotUpdate ` to `full-reload` file.

Is it a simple way for this? And there are some anthor way for update `meta` or other data? Thank you.


packages/nuxt/src/pages/runtime/composables.ts
```diff
@@ -59,17 +59,28 @@ const warnRuntimeUsage = (method: string) => {
...
+      const route = useRoute()
+      const isRouteComponent = component && route.matched.some(p => Object.values(p.components || {}).includes(component))
+      // update route metadata with the provided meta (for HMR support)
+      if (isRouteComponent) {
+        for (const key in route.meta) {
+          route.meta[key] = meta[key]
+        }
+        for (const key in meta) {
+          route.meta[key] = meta[key]
+        }
```
danielroe, line 75: what about this implementation?

`definePageMeta` isn't removed from the code in dev mode, so we can directly update the route meta (which is a reactive object)

danielroe:
looks like there might be some issues with nested routes - might need to update the implementation ....

xjccc:
emm.. I have a question for this.

```
<template>
  <div class="container">
    index - {{ $route.name }}
  </div>
</template>

<script setup lang="ts">
definePageMeta({
  name: 'index123',
})
const route = useRoute()
console.log(route)
</script>
```

When changing `definePageMeta.name`,  then `command + s`,  the `route` log is also the old value.

PR #29660, opened 2024-10-23
perf: avoid array iterators over objects

GalacticHypernova:
### 📚 Description

Object iterators like `Object.keys`, `Object.values`, and `Object.entries`, when used for iteration, create a temporary intermediate array by internally making another iteration, essentially doubling the memory usage and execution time. Using object iterators will avoid that intermediate array and the extra internal iteration.


packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -34,7 +34,8 @@ async function loadComponents (source = appBaseURL, paths: NuxtIslandResponse['c
 
   const promises: Array<Promise<void>> = []
 
-  for (const [component, item] of Object.entries(paths)) {
```
danielroe, line 37: This is deliberate, in many of these cases, for greater type safety - it avoids needing to use the (unsafe) non-null assertion.

If you have another solution that preserves type safety I'd be very grateful.

GalacticHypernova: I figured it was most likely because of this, but there is a clear downside here.

TypeScript has an issue with nullability with for in loops because it inherently infers the keys as string, regardless of the object's type.

But until that's fixed on TypeScript's end, it would be an inevitable side effect.

However, I personally don't think the performance and memory impacts are a proportional tradeoff to type safety in the loop declaration, especially with these stats:

```ts
const largeObject = {}
for (let i=0; i<100000;i++) {
    largeObject[i] = i
}

console.time("array method")
for (const num of Object.keys(largeObject)) {
    // It doesn't matter what happens within the loop as the main slowdown is the keys iterator
}
console.timeEnd("array method")

console.time("obj method")
for (const num in largeObject) {
    // It doesn't matter what happens within the loop as the main slowdown is the keys iterator
}
console.timeEnd("obj method")
```

Performance results:
image
image

Memory results:
image
image

It's also worth noting that the values inside the loop are correctly typed, it's just that TS is struggling to verify it exists because of their flawed type inferring logic.

What do you think?

danielroe:
It looks like these changes have a negligible impact on performance (with the benchmarks we've created). Can you think of some better benchmarks to create that might show an impact?

GalacticHypernova:
Hey @danielroe ! Apologies for the late response.
The benchmarks primarily test for performance, which indeed might not show tangible results unless tested against datasets of thousands of elements. However, there is the memory efficiency aspect which this PR addresses, as array iterators essentially duplicate the object (or only keys/values, depending on the method), with no real benefit for the iteration purpose

PR #29586, opened 2024-10-17
feat(nuxt): add definePageMeta props

xjccc:
### 🔗 Linked issue

resolves #29585

### 📚 Description

add `definePageMeta.props` passing params to components props

PR #29683, opened 2024-10-25
fix(nuxt): definePageMeta props set true

xjccc:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/29586

### 📚 Description

Sorry for this, I find that, in this pull request, this judgment condition has some effect.

```
// /pages/[params]/[id].vue
// not work
definePageMeta({
   props: true,
})
```

```
if (property.value.type !== 'Literal' || (typeof property.value.value !== 'string' && typeof property.value.value !== 'boolean')) {
```

Please recheck this. @danielroe 

Thank you very much❤

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -273,7 +273,7 @@ export async function getRouteMeta (contents: string, absolutePath: string): Pro
             continue
           }
 
-          if (property.value.type !== 'Literal' || typeof property.value.value !== 'string') {
+          if (property.value.type !== 'Literal' || (typeof property.value.value !== 'string' && typeof property.value.value !== 'boolean')) {
             console.debug(`[nuxt] Skipping extraction of \`${key}\` metadata as it is not a string literal or array of string literals (reading \`${absolutePath}\`).`)
             dynamicProperties.add(key)
             continue
@@ -571,6 +571,10 @@ async function createClientPage(loader) {
 
       if (route.children) {
         metaRoute.children = route.children
+      }
+
+      if (route.props) {
+        metaRoute.props = route.props
       }
 
       if (route.meta) {
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -574,7 +574,7 @@ async function createClientPage(loader) {
       }
 
       if (route.props) {
-        metaRoute.props = route.props
+        metaRoute.props = serializeRouteValue(route.props)
       }
 
       if (route.meta) {
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -273,7 +273,7 @@ export async function getRouteMeta (contents: string, absolutePath: string): Pro
             continue
           }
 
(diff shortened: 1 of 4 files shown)


xjccc:
image

I find this always `false`.. then the `props` is deleted. When set `props: true`

packages/nuxt/src/pages/utils.ts
```diff
@@ -573,6 +573,10 @@ async function createClientPage(loader) {
         metaRoute.children = route.children
       }
 
+      if (route.props) {
+        metaRoute.props = route.props
+      }
```
danielroe, line 578: nice catch.

I think instead we need to set props value on the generated route meta.

does this work for you?

xjccc: Sorry for late. I will try this next day..

xjccc: Do you think this is right ? Add `props: serializeRouteValue(page.props)`

packages/nuxt/src/pages/utils.ts
```diff
@@ -574,7 +575,7 @@ async function createClientPage(loader) {
       }
 
       if (route.props) {
-        metaRoute.props = serializeRouteValue(route.props)
+        metaRoute.props = route.props
```
danielroe, line 578: I think we still need this in case an object is passed in

packages/nuxt/src/pages/utils.ts
danielroe: perfect!

packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
```diff
@@ -36,7 +36,7 @@
       "meta": "mockMeta || {}",
       "name": "mockMeta?.name ?? "page-with-props"",
       "path": "mockMeta?.path ?? "/page-with-props"",
-      "props": "mockMeta?.props ?? false",
+      "props": "true",
```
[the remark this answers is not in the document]
xjccc: I have update this. What do you think about?

packages/nuxt/src/pages/utils.ts
```diff
@@ -542,7 +543,7 @@ export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> =
       const metaRoute: NormalizedRoute = {
         name: `${metaImportName}?.name ?? ${route.name}`,
         path: `${metaImportName}?.path ?? ${route.path}`,
-        props: `${metaImportName}?.props ?? false`,
+        props: `${metaImportName}?.props ?? ${route.props ?? false}`,
```
xjccc, line 546: Is this need to set `false`?

danielroe: hm... I think either is fine. false is probably more minifiable (!0)

danielroe (approved):

packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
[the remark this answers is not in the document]
danielroe: looks good!

PR #29719, opened 2024-10-29
Update nuxt-loading-indicator.ts

haradhansharma:
Opacity controlling is unnecessary here to avoid hydration issue.


danielroe:
Would you provide a bug report/reproduction?

(Or a test for this PR?)

PR #29755, opened 2024-11-01
docs: remove tip comparing `useFetch` to `useAsyncData(url, () => $fetch(url))`

moshetanzer:
imo this is deceiving, they are not similar at all in the way that $fetch doesn't pass cookies server side. This `tip` doesn't really add anything, and if anything may confuse the user (as useRequestFetch is used behind useFetch)

### 📚 Description

imo this tip is deceiving, `useFetch` and `useAsyncData(url, () => $fetch(url))` are not similar at all in the way that $fetch doesn't pass cookies server side. 
This `tip` doesn't really add anything, and if anything may confuse the user (as useRequestFetch is used behind useFetch, which does pass the cookies). 

## Summary by CodeRabbit

- **Documentation**
	- Enhanced explanations of data-fetching composables in Nuxt, including `useFetch`, `useAsyncData`, and `$fetch`.
	- Expanded sections to address double-fetching and hydration issues in isomorphic applications.
	- Introduced a new section on "Passing Headers and Cookies" for managing client headers during server-side rendering.
	- Clarified the role of `$fetch` and its limitations regarding network call de-duplication.
	- Added examples for handling loading states, lazy fetching, and client-only fetching.
	- Included practical recipes for consuming Server-Sent Events (SSE) via POST requests.
	- Discussed custom serialization functions for data between server and client.


danielroe:
I would rather add a clarifying comment if it is in fact unclear

moshetanzer:
Hey @danielroe as getting to know nuxt better - based on the example I was under the impression that $fetch passes cookies by default in server context. Where in fact it doesn't as userequestfetch is used which does pass the cookies. Not sure what this tip adds exactly? Feel free to close pr.

danielroe:
i have an idea - i'll push

danielroe (approved):
Hope this adds extra clarity! 🙏 

Feel free to open another PR if you have additional ideas about how to make it better.

moshetanzer:
Think this gives enough of an idea that there is more there ❤️❤️

cyrilf:
Hey,
I read through the docs and I find the comparison between `useFetch` and `useAsyncData` less clear now..

> useFetch(url) is nearly equivalent to useAsyncData(url, () => event.$fetch(url))
> It's developer experience sugar for the most common use case. (You can find out more about event.fetch at [useRequestFetch](https://nuxt.com/docs/api/composables/use-request-fetch).)

- I don't understand where is the `event` variable coming from?
- On the following sentence it's mentioned `event.fetch` without the `$` dollar sign. Is that on purpose?

Thanks for the clarification :+1:  @danielroe  @moshetanzer

PR #29776, opened 2024-11-04
fix(nuxt,schema): allow showing spa loader til after hydration

### LINKED ISSUE #21721: Keep showing the spa-loading-template until page:finish
### Describe the feature

It would be nice to show the spa loading template until the 'page:finish' event. 

When using Nuxt with SSR turned off and using the spa-loading-template to show a loading screen, the loading screen disappears once the app mounts but the app isn't rendered yet. I believe because nuxt lazy loads the components that need to be rendered. This leads to a blank page once the app mounts and until the components are loaded. 

Other than that I love the new SPA loading template feature!

### Additional information

### Final checks

RokeAlvo:
### 🔗 Linked issue

Resolves #21721

### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality)
- [] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Currently, the SPA loader is displayed only until the application is mounted. After that, the user sees a white screen until suspense:resolved occurs.

Changes:
  - Show SPA loading template outside of app root element
  - Add configuration options for SPA loading template element (app.spaLoaderTag, app.spaLoaderAttrs)
  - Add flag to config.experimental.spaPreloaderOutside (default value is false)

### 📝 Checklist

## Release Notes

- **New Features**
  - Introduced `spaLoadingTemplateLocation` for customizable loading screens in client-only pages.
  - Added `spaLoaderTag` and `spaLoaderAttrs` for enhanced SPA loader configuration.
  - New Vue component `app.vue` with asynchronous data handling.
  - New HTML element for SPA loading indication.

- **Bug Fixes**
  - Improved handling of SPA loader visibility based on configuration settings.

- **Documentation**
  - Updated upgrade guide with instructions for transitioning to Nuxt 4.
  - Enhanced documentation on experimental features, including `spaLoadingTemplateLocation`.

- **Tests**
  - Added tests to validate SPA loader behavior based on configuration settings.


danielroe (commented):
This looks like a nice fix to me but will need to defer to v4 as it's might be a breaking change (CSS might be targeted assuming the loading element is within the root app).

Alternatively we could take an experimental option that 'switches on' this behaviour, which is enabled by default in v4 mode, and merge sooner.

RokeAlvo:
> Alternatively we could take an experimental option that 'switches on' this behaviour, which is enabled by default in v4 mode, and merge sooner.

add `experimental.spaLoaderOutside` flag to nuxt.config?

danielroe:
that makes sense. and you can also use `routeRules` with `ssr: false` to turn on spa loading, even within the basic fixture, which might be a simpler approach...

RokeAlvo:
> that makes sense. and you can also use `routeRules` with `ssr: false` to turn on spa loading, even within the basic fixture, which might be a simpler approach...

- added experimenal flag
- changed tests

test/spa-loader/spa-preloader-outside-disabled.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  it('spa-loader does not appear while the app is mounting', async () => {
+    const browser = await getBrowser()
+    const page = await browser.newPage({})
+    await page.goto(url('/spa'), { waitUntil: 'domcontentloaded' })
+
+    const loader = page.getByTestId('loader')
+    expect(await loader.isHidden()).toBeTruthy()
+
+    await page.close()
+  }, 60_000)
```
danielroe, line 40: what is the desired behaviour we're testing here?

(or is this just proof of the current bug?)

test/spa-loader/spa-preloader-outside-enabled.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const browser = await getBrowser()
+    const page = await browser.newPage({})
+    await page.goto(url('/ssr'), { waitUntil: 'domcontentloaded' })
+
+    const loader = page.getByTestId('loader')
+    expect(await loader.isHidden()).toBeTruthy()
+
+    const content = page.getByTestId('content')
+    await content.waitFor({ state: 'visible' })
+    expect(await loader.isHidden()).toBeTruthy()
```
danielroe, line 48: maybe we need a slightly different assertion? or modify the implementation so both are not present on the page at the same moment?

test/spa-loader/spa-preloader-outside-disabled.test.ts
RokeAlvo: Correct. Since we maintain the current behavior with the value of within, it needs to be established.

test/spa-loader/spa-preloader-outside-enabled.test.ts
RokeAlvo: This is a test for the /ssr page. We should not see the SPA loader at any time:

- neither during document loading (line 44)
- nor while resolving promises. (line 48)

RokeAlvo: Hmm... the first test is broken...

packages/schema/src/config/app.ts
```diff
@@ -274,14 +274,7 @@ export default defineUntypedSchema({
      * Customize Nuxt Nuxt SpaLoader element attributes.
      * @type {typeof import('@unhead/schema').HtmlAttributes}
      */
-    spaLoaderAttrs: {
-      $resolve: async (val: undefined | null | Record<string, unknown>, get) => {
-        const spaLoaderId = await get('app.spaLoaderId')
-        return defu(val, {
-          id: spaLoaderId === false ? undefined : (spaLoaderId || '__nuxt-spa-loader'),
```
RokeAlvo: @danielroe We need to ensure that the element's ID is set. These changes break the preloader.

danielroe: ahh, of course. I was just intending to get rid of the reference to the nonexistent spaPreloaderId.

danielroe: ... and what do you mean it breaks the preloader?

packages/schema/src/config/app.ts
```diff
@@ -274,7 +274,9 @@ export default defineUntypedSchema({
      * Customize Nuxt Nuxt SpaLoader element attributes.
      * @type {typeof import('@unhead/schema').HtmlAttributes}
      */
-    spaLoaderAttrs: {},
+    spaLoaderAttrs: {
```
RokeAlvo, line 277: maybe, better?
```
    spaLoaderAttrs: {
<<<<<<< HEAD
      $resolve: (val: undefined | null | Record<string, unknown>) => {
        return defu(val, {
          id: val?.id ?? '__nuxt-spa-loader',
        })
      },
=======
      id: '__nuxt-loader',
>>>>>>> f4106a79061f41753ee9caff028d8755272634af
    },
  },
```

danielroe: not sure I understand

RokeAlvo: So, if I pass

spaLoaderAttrs: { class: "some-class" }
it will ultimately result in

{ class: "some-class", id: "__nuxt-loader" }
right?

test/spa-loader/spa-preloader-outside-disabled.test.ts
RokeAlvo, line 1: I don't know how to fix it.

PR #29823, opened 2024-11-06
fix(rspack): rspack builder usage without installing webpack

### LINKED ISSUE #29811: webpack is required for rspack builder
### Environment

```bash
- Node Version:     v22.7.0
- Nuxt Version:     3.14.0
- CLI Version:      3.15.0
- Nitro Version:    2.10.3
- Package Manager:  yarn@4.5.1
- Builder:          rspack
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

A minimal Nuxt app is enough.

`package.json`:
```json
{
    "devDependencies": {
        "@nuxt/rspack-builder": "^3.12.2",
        "nuxt": "^3.14.0"
    },
    "scripts": {
        "build": "nuxt build"
    },
    "packageManager": "yarn@4.5.1"
}
```

`nuxt.config.ts`:
```ts
    builder: 'rspack'
})
```

`app.vue`:
```vue
<template>
    <NuxtLayout>
        <NuxtPage />
    </NuxtLayout>
</template>
```

`pages/index.vue`:
```vue
<template>
    <div>
        Hello!
    </div>
</template>
```

Note that I dont have webpack installed on my machine.

### Describe the bug

When running `nuxt build`, the process throws error, say that it cant import webpack in webpackbar's file.

I have tried patching the import to `@rspack/core` but no luck.

### Additional context

N/A.

### Logs

```bash
Nuxt Build Error: Cannot find package 'webpack' imported from D:\[...]\node_modules\webpackbar\dist\index.mjs  

  at packageResolve (node:internal/modules/esm/resolve:839:9)
  at moduleResolve (node:internal/modules/esm/resolve:908:18)
  at defaultResolve (node:internal/modules/esm/resolve:1038:11)
  at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:554:12)
  at ModuleLoader.resolve (node:internal/modules/esm/loader:523:25)
  at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:246:38)
  at ModuleJob._link (node:internal/modules/esm/module_job:126:49)
```

HigherOrderLogic:
### 🔗 Linked issue

Resolves #29811.

### 📚 Description

Checking what's the type of builder using and importing it to avoid modules not found problem.

## Summary by CodeRabbit

- **New Features**
	- Introduced `WebpackBarPlugin` for improved build progress reporting and error handling.
  
- **Dependency Updates**
	- Upgraded `webpackbar` dependency from version `^6.0.1` to `^7.0.0` in both `@nuxt/rspack-builder` and `@nuxt/webpack-builder` packages.

- **Bug Fixes**
	- Enhanced error state evaluation and progress reporting within the Webpack preset configuration.

### DIFF
--- a/packages/webpack/builder.d.ts
+++ b/packages/webpack/builder.d.ts
@@ -1,8 +1,10 @@
 declare module '#builder' {
   import type Webpack from 'webpack'
+  import type WebpackBarPlugin from 'webpackbar'
   import type MiniCssExtractPlugin from 'mini-css-extract-plugin'
 
   export const webpack: typeof Webpack
+  export const WebpackBarPlugin: typeof WebpackBarPlugin
   export const MiniCssExtractPlugin: typeof MiniCssExtractPlugin
   export const builder: 'webpack' | 'rspack'
 }
--- a/packages/webpack/builder.d.ts
+++ b/packages/webpack/builder.d.ts
@@ -1,10 +1,9 @@
 declare module '#builder' {
   import type Webpack from 'webpack'
-  import type WebpackBarPlugin from 'webpackbar'
   import type MiniCssExtractPlugin from 'mini-css-extract-plugin'
 
   export const webpack: typeof Webpack
-  export const WebpackBarPlugin: typeof WebpackBarPlugin
+  export const WebpackBarPlugin: typeof import('webpackbar').default
   export const MiniCssExtractPlugin: typeof MiniCssExtractPlugin
   export const builder: 'webpack' | 'rspack'
 }
--- a/packages/webpack/src/presets/base.ts
+++ b/packages/webpack/src/presets/base.ts
@@ -104,14 +104,15 @@ function basePlugins (ctx: WebpackConfigContext) {
             if (stats.hasErrors()) {
               ctx.nuxt.callHook(`${builder}:error`)
             } else {
-              // @ts-expect-error TODO: this should be valid
+              // @ts-expect-error TODO
               logger.success(`${stats.name} ${stats.message}`)
             }
           },
           allDone: () => {
             ctx.nuxt.callHook(`${builder}:done`)
           },
-          progress (_, { statesArray }) {
+          // @ts-expect-error TODO: there is NO statesArray on type ProgressPlugin, this error is correct
+          progress ({ statesArray }) {
             ctx.nuxt.callHook(`${builder}:progress`, statesArray)
           },
         },
--- a/packages/webpack/src/presets/base.ts
+++ b/packages/webpack/src/presets/base.ts
@@ -94,12 +94,12 @@ function basePlugins (ctx: WebpackConfigContext) {
       stats: !ctx.isDev,
       reporter: {
         reporter: {
-          change: (_, { shortPath }) => {
+          change: (_: any, { shortPath }: any) => {
             if (!ctx.isServer) {
               ctx.nuxt.callHook(`${builder}:change`, shortPath)
             }
           },
-          done: ({ state }) => {
+          done: ({ state }: any) => {
             if (state.hasErrors) {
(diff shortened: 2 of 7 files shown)


packages/webpack/src/presets/base.ts
```diff
@@ -1,7 +1,9 @@
 import { normalize, resolve } from 'pathe'
 // @ts-expect-error missing types
 import TimeFixPlugin from 'time-fix-plugin'
+import type { WebpackBarOptions } from 'webpackbar'
```
pi0, line 4: This line requires webpack as dependency. We need to use dynamic async import

danielroe (commented):
the way I've done it so far is to add exports from `builder.mjs` in each package's folder

that way we can have direct imports that are different in webpack vs rspack

HigherOrderLogic:
Then how should I declare the type for the `#builder` module? Just import from the `webpackbar` as usual?

danielroe:
yes - the types aren't included in the built project

HigherOrderLogic:
> the way I've done it so far is to add exports from builder.mjs in each package's folder

Done, not sure about the typecheck part tho.

If there is any way to remove the `: any` part to retain the strict typecheck, I'm happy to do it.

packages/webpack/src/presets/base.ts
```diff
@@ -88,20 +87,19 @@ function basePlugins (ctx: WebpackConfigContext) {
...
+    ctx.config.plugins.push(new WebpackBarPlugin({
       name: ctx.name,
       color: colors[ctx.name as keyof typeof colors],
       reporters: ['stats'],
-      // @ts-expect-error TODO: this is a valid option for Webpack.ProgressPlugin and needs to be declared for WebpackBar
       stats: !ctx.isDev,
       reporter: {
         reporter: {
-          change: (_, { shortPath }) => {
+          change: (_: any, { shortPath }: any) => {
```
danielroe, line 97: any reason we don't have types for these?

HigherOrderLogic: I'm not sure, after changing to importing from `#builder`, the typecheck just throw error in those place.

My Vscode's intellisense is also showing the `WebpackBarPlugin` as `any` type.

image

packages/webpack/builder.d.ts
```diff
@@ -3,6 +3,7 @@ declare module '#builder' {
   import type MiniCssExtractPlugin from 'mini-css-extract-plugin'
 
   export const webpack: typeof Webpack
+  export const WebpackBarPlugin: typeof import('webpackbar').default
```
danielroe, line 6: can you add an `import type` above like with `MiniCssExtractPlugin`? might that resolve the issue?

HigherOrderLogic: That was what I did previously, the type was `any`.

danielroe: if so then that is a bug in the types for `webpackbar`, probably

packages/webpack/src/presets/base.ts
```diff
@@ -101,16 +100,18 @@ function basePlugins (ctx: WebpackConfigContext) {
...
             } else {
-              logger.success(`${state.name} ${state.message}`)
+              // @ts-expect-error TODO
+              logger.success(`${stats.name} ${stats.message}`)
             }
           },
           allDone: () => {
             ctx.nuxt.callHook(`${builder}:done`)
           },
+          // @ts-expect-error TODO: there is NO statesArray on type ProgressPlugin, this error is correct
```
danielroe, line 114: in that case we probably need to update the hook signature + pass what we have

HigherOrderLogic: This has been fixed, the only left is the `stats` one.

danielroe (approved):
thank you

PR #29871, opened 2024-11-10
feat(schema): add snow effect on loading screen in winter

lihbr:
https://github.com/user-attachments/assets/cac06ab3-2490-470f-9b2a-13f3ddbc5318

### 📚 Description

This PR adds a seasonal (temporary) snow effect on Nuxt 3 loading screen. It's just an idea I wanted to play with after a discussion during our last meeting. If we don't like it, we can trash it, or see how to make it better ☺️

I tried to keep the code as minimal as possible. It also respects the user's preferences regarding animations.

For credits, the snow effect is inspired by [Boris Šehovac previous WebGL work on Codepen](https://codepen.io/bsehovac/pen/GPwXxq), but written using the Canvas's 2d context for conciseness here.

### DIFF
--- a/packages/ui-templates/templates/loading/index.html
+++ b/packages/ui-templates/templates/loading/index.html
@@ -189,9 +189,9 @@
       if ([0, 10, 11].includes(new Date().getMonth())) {
         // Create canvas
         const $canvas = document.createElement('canvas')
-        $canvas.classList.add('fixed', 'inset-0', '-z-10', 'pointer-events-none', 'opacity-0', 'transition-opacity', 'duration-[2s]', 'ease-in', 'blur-[4px]')
+        $canvas.style = 'position: fixed; inset: 0; z-index: -10; pointer-events: none; opacity: 0; transition: opacity 2s ease-in; filter: blur(4px);'
         document.body.appendChild($canvas)
-        $canvas.classList.remove('opacity-0')
+        $canvas.style.opacity = 1
 
         // Animation options
         const density = 0.00025
--- a/packages/ui-templates/templates/loading/index.html
+++ b/packages/ui-templates/templates/loading/index.html
@@ -184,5 +184,91 @@
         check()
       }
     </script>
+    <script>
+      // Create canvas
+      const $canvas = document.createElement('canvas')
+      $canvas.classList.add('fixed', 'inset-0', '-z-10', 'pointer-events-none', 'opacity-0', 'transition-opacity', 'duration-[2s]', 'ease-in', 'blur-[4px]')
+      document.body.appendChild($canvas)
+      $canvas.classList.remove('opacity-0')
+
+      // Animation options
+      const density = 0.00025
+      const wind = { current: 0, maxCurrent: 4, force: 0.1, target: 0.1, min: 0.1, max: 0.4, easing: 0.01 }
+      const gravity = 1.25
+
+      // Auto resize canvas
+      let recreateParticles = false
+      function resize() {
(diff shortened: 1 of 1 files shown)


lihbr:
Glad you like it ☺️

I believe it's enough on the loading template, and that having it everywhere would be a bit overkill, at least for now while we experiment with seasonal effects.

If it's something the community likes, maybe we could later on just load instead a hosted script from nuxt.com, this way we can change it/remove it without prompting a package release. This would also allow us to change it more often and let the community contribute to it, a bit like with [Google's doodles](https://doodles.google)

danielroe:
For some reason I'm getting this:

CleanShot 2024-12-03 at 10 07 32@2x

Can investigate more.

lihbr:
I'll investigate and make it Nov - December - January only if that works :)

lihbr:
Just curious what do you run to get the above result? Looks like Tailwind as a whole struggle to load as even the Nuxt logo is not centered 🤔

danielroe:
I ran:

```
pnpm i
pnpm dev:prepare

pnpm nuxi dev test/fixtures/basic
```

and then opened localhost:3000 in the browser.

lihbr:
OK thank you so much! This should be good ☺️

💅 Style fixes
📆 Winter-bound

danielroe (approved):

typed-sigterm:
The first time I felt that not having DST was not a good thing🤔Maybe we can add a button to toggle it manually?
??

lihbr:
Thanks for your feedback @typed-sigterm, you have a toggle animation button at the top right of the loading screen :)

M-GRV:
> Thanks for your feedback @typed-sigterm, you have a toggle animation button at the top right of the loading screen :)

congrats on the merge! I love this!

seanchatmangpt:
How can I disable this? It is making me motion sick. You might want to consider accessibility or a clear way to disable in the future.

atomflunder:
@seanchatmangpt Theres a button on the top right of the loading screen to toggle the effect.  
You only see it when you hover over it.

memochou1993:
Some countries, like Taiwan, don’t use daylight saving time (DST), so the timezone offset stays the same year-round.

😞 

image

danielroe:
Improvements are welcome 🙏

hendriksenremco:
Is there a way to disable this effect?

danielroe:
Yes, as Lucie commented above, there's a link in the top right of the loading screen.

https://volta.net/nuxt/nuxt/pull/29871#comment-4872944

danielroe:
Agreed. As I commented above - improvements are welcome!

hendriksenremco:
> (also, no need to delete your comment!)

Yes, sorry about that. My actions were faster then my mind.

typed-sigterm:
@danielroe I made [this](https://github.com/typed-sigterm/which-hemisphere), perhaps it can help.

finnbuster:
Hello, just reporting from Australia where we don't get snow in winter ;)

danielroe:
@finnbuster it _tries_ to detect hemisphere. is that not working?

finnbuster:
@danielroe it's definitely getting the hemisphere right and is displaying snow in our winter as defined in the code but the thing is that it doesn't snow in most of the souther hemisphere :) https://earthobservatory.nasa.gov/global-maps/MOD10C1_M_SNOW#:~:text=Snow%20and%20ice%20cover%20most,snow%20cover%20almost%20year%20round.

danielroe:
😱

PR #29895, opened 2024-11-12
fix: #29894 addTemplate should consider the entire path

### LINKED ISSUE #29894: addTemplate filter ignoring wrong templates - breaking change from nuxt2
### Environment

- Operating System: Linux
- Node Version:     v22.6.0
- Nuxt Version:     3.14.159
- CLI Version:      3.15.0
- Nitro Version:    2.10.3
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-uuytf8-be6znm?file=generated%2Fsubfolder%2Findex.js

### Describe the bug

addTemplate function is wrongfully filtering the templates by removing reused filenames, possibly to avoid duplication. In nuxt 2 it worked as intended, you could use the same filename in multiple templates under different subfolders and every file was written under the correct subfolder.

The problem is at @nuxt/kit/index.mjs@addTemplate

Image

I think it would be best to consider the entire dst param if possible, or at least not fail silently because using addTemplate is already hard enough :(

### Additional context

_No response_

### Logs

```shell-script

```

guska8:
### 🔗 Linked issue

resolves #29894

### 📚 Description

When using addTemplate with the same filename but different subfolders only the last template is considered, the ones before it are silently ignored.

It is a breaking change in nuxt 3 from nuxt 2 because on the previous version all files are created as expected.

Feel free to edit or request changes if needed.

-----
<a href="https://stackblitz.com/~/github.com/guska8/nuxt/tree/guska8/issue29894">Review PR in StackBlitz</a> _Submitted with [StackBlitz](https://stackblitz.com/~/github.com/guska8/nuxt/tree/guska8/issue29894)._

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved template management by ensuring that duplicates are identified based on destination paths instead of filenames, enhancing the overall template handling process.

### DIFF
--- a/packages/kit/src/template.ts
+++ b/packages/kit/src/template.ts
@@ -22,9 +22,9 @@ export function addTemplate<T> (_template: NuxtTemplate<T> | string) {
   // Normalize template
   const template = normalizeTemplate(_template)
 
-  // Remove any existing template with the same filename
+  // Remove any existing template with the same destination path
   nuxt.options.build.templates = nuxt.options.build.templates
-    .filter(p => normalizeTemplate(p).filename !== template.filename)
+    .filter(p => normalizeTemplate(p).dst !== template.dst)
 
   // Add to templates array
   nuxt.options.build.templates.push(template)


guska8:
> **Actionable comments posted: 0**
> 
> 🧹 Outside diff range and nitpick comments (1)
> > packages/kit/src/template.ts (1)> `25-27`: **Consider adding documentation and warning for this behavior change.**
> > > While this change restores the expected Nuxt 2 behavior, it modifies how template deduplication works. Consider:
> > > 
> > > 1. Adding documentation about this behavior change
> > > 2. Adding a warning when multiple templates with the same filename are detected, to help users understand why their templates are now being processed differently
> > > 
> > > Here's a suggested enhancement to make the change more transparent:
> > > ```diff
> > >  // Remove any existing template with the same destination path
> > > +const existingTemplate = nuxt.options.build.templates
> > > +  .find(p => normalizeTemplate(p).filename === template.filename && normalizeTemplate(p).dst !== template.dst)
> > > +if (existingTemplate) {
> > > +  console.warn(`Multiple templates with filename "${template.filename}" detected. ` +
> > > +    `Both templates will be processed as they have different destination paths. ` +
> > > +    `This is a change from previous behavior where only the last template would be kept.`)
> > > +}
> > >  nuxt.options.build.templates = nuxt.options.build.templates
> > >    .filter(p => normalizeTemplate(p).dst !== template.dst)
> > > ```
> 
> 📜 Review details

I think the code suggestion from coderabbit is irrelevant because the dst variable already includes the filename.

The part about the warning i don't know how you usually do

danielroe (approved):
This seems good to me - thank you ❤️ 

I think there's some more optimisation we could do here for a future PR. (Do we need to normalise each template at every iteration? 🤔)

PR #29957, opened 2024-11-17
refactor(nuxt): define layouts as async vue components

danielroe:
### 📚 Description

this follows up on https://github.com/vuejs/core/issues/6638 to mark layouts as explicit vue components

it should also be slightly more performant 🙏 

## Summary by CodeRabbit

- **New Features**
	- Simplified layout loading process by switching from asynchronous to synchronous calls, which may enhance performance.
	- Enhanced functionality of layout and middleware templates for better component and middleware handling.

- **Bug Fixes**
	- Maintained error handling for invalid layouts with console warnings.

- **Documentation**
	- Updated comments and added TODOs for future improvements in the codebase.

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-layout.ts
+++ b/packages/nuxt/src/app/components/nuxt-layout.ts
@@ -16,7 +16,6 @@ import layouts from '#build/layouts'
 // @ts-expect-error virtual file
 import { appLayoutTransition as defaultLayoutTransition } from '#build/nuxt.config.mjs'
 
-// TODO: revert back to defineAsyncComponent when https://github.com/vuejs/core/issues/6638 is resolved
 const LayoutLoader = defineComponent({
   name: 'LayoutLoader',
   inheritAttrs: false,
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/components/nuxt-layout.ts
```diff
@@ -16,21 +16,17 @@ import layouts from '#build/layouts'
...
   },
-  async setup (props, context) {
+  setup (props, context) {
     // This is a deliberate hack - this component must always be called with an explicit key to ensure
     // that setup reruns when the name changes.
-
-    const LayoutComponent = await layouts[props.name]().then((r: any) => r.default || r)
-
-    return () => h(LayoutComponent, props.layoutProps, context.slots)
+    return () => h(layouts[props.name], props.layoutProps, context.slots)
```
adamdehaven, line 29: @danielroe is there any chance that the removal of `await` before calling `layouts[props.name]` here (or another change in this PR) could cause a hydration error in some scenarios? 

After upgrading to `v3.14.1592` (the only change in my project) I'm now seeing `Hydration completed but contains mismatches` on a page that only contains a NuxtLayout and a paragraph tag in the default slot.

hydration error


It doesn't occur when the dev server is running; only when building and running the build (e.g. with `wrangler`).

I've tried putting together a reproduction but (of course) it doesn't happen there as of yet. 

Do you have any suggestions on how to further triage? Typically I'd start disabling modules, etc. but in a larger project, that isn't the easiest task since so many things depend on others 😬

adamdehaven: It actually also seems to break styles injected during SSR via `useHead` 🤔

adamdehaven: Put together an issue; I'm not confident this PR is the root of the issue, but it's my best guess lol
https://github.com/nuxt/nuxt/issues/30188

Issue #29258, opened 2024-10-03
Better in-terminal reporting for reserved cases


ramijames:
### Describe the feature

[As discussed on Reddit here](https://www.reddit.com/r/Nuxt/comments/1fubamh/usestate_error/).

This was a new-user issue for me, as I didn't realize that useState was reserved. The terminal error that it spit out was: 

WARN  Duplicated imports "useState", the one from "#app/composables/state" has been ignored and "/Users/ramijames/_CODE/vewrite/composables/useState.js" is used

I'd like to suggest that there be a more specific error for reserved cases that explicitly states that they are reserved.

@danielroe suggested that we can add some code for this case in packages/nuxt/src/imports/module.ts

If you want, I am happy to create a PR for this.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

danielroe:
Thank you! 🙏

sim-codes:
@ramijames are you working on this?
If not, I would love to work on this.

PR #29971, opened 2024-11-19
feat(nuxt): warn if core auto-import presets are overridden

OrbisK:
### 🔗 Linked issue

#29258

### 📚 Description

This pr adds a console.warn if the /composables or /utils override some imports from the defined presets.

It implements this in packages/nuxt/src/imports/module.ts (as suggested in #29258). This is not perfect. This pr does not cover/warn about imports from https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/module.ts#L355, for example.

The text of the warning can of course still be adapted/updated.

I have done my best to test it. I tried to stubGlobal('console') but had no luck so far.

output if adding "useState" at `/composables`

```
[15:39:33]  WARN  [imports] "useState" is already defined and auto imported from "#app/composables/state" within nuxt itself. Please consider renaming "useState" at /home/robin/WebstormProjects/nuxt/playground/composables/state.ts.

ℹ Running with compatibility version 4                                                                                                                                                                                   15:39:33

 WARN  Duplicated imports "useState", the one from "#app/composables/state" has been ignored and "/home/robin/WebstormProjects/nuxt/playground/composables/state.ts" is used                                              15:39:33


[15:39:33]  WARN  [imports] "useState" is already defined and auto imported from "#app/composables/state" within nuxt itself. Please consider renaming "useState" at /home/robin/WebstormProjects/nuxt/playground/composables/state.ts.

✔ Vite client built in 24ms                                                                                                                                                                                              15:39:33

 WARN  Duplicated imports "useState", the one from "#app/composables/state" has been ignored and "/home/robin/WebstormProjects/nuxt/playground/composables/state.ts" is used     
``` 

## Summary by CodeRabbit

- **New Features**
	- Enhanced import handling with a new `presetMap` for better tracking of imported presets.
	- Added a warning mechanism for naming conflicts between composables and presets.
	- Improved type declaration generation based on import context.

- **Bug Fixes**
	- Refined logic for scanning directories to prioritize imports based on their source.

- **Chores**
	- Optimized path resolution by skipping relative import paths for explicitly installed node modules.


packages/nuxt/src/imports/module.ts
```diff
@@ -117,7 +117,25 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
           })
+
+          const presetMap = new Map<string, string>()
+
+          for (const preset of presets) {
+            preset.imports = preset.imports ?? []
+            for (const i of preset.imports) {
+              presetMap.set(i, preset.from)
+            }
+            presetMap.set(preset.as, preset.from)
```
OrbisK, line 128: the previous implementation has type issues. types are weired on the presets. I tried to solve it like this, but seems kinda hacky
```suggestion
            for (const i of preset.imports) {
              if(typeof i === "string") {
                presetMap.set(i, preset.from)
                continue
              }
              let name, as, from
              if(Array.isArray(i)) {
                [name, as, from] = i
              }
              else {
                name = i?.name ?? undefined;
                as = i?.as ?? undefined;
                from = preset.from
              }
              if(!name || !from) {
                continue
              }
              presetMap.set(as ?? name, from)
            }
```

danielroe (commented):
I think what we want is:

- not to warn if using an import from a layer
- only warn if overriding with an _equal_ `priority` - this way modules can intentionally override without warnings

OrbisK:
> I think what we want is:
> 
> * not to warn if using an import from a layer
> * only warn if overriding with an _equal_ `priority` - this way modules can intentionally override without warnings

I can try. This shouldn't be possible in src/imports/module, should it?

Any advise?

danielroe (approved):

danielroe:
sorry for directly pushing - I tried to move as much of the calculation out of `regenerateImports` for performance reasons. I've also moved it after `imports:extend` so we catch all changes.

modules can always remove nuxt imports in `imports:sources`, and we won't warn in this case.

OrbisK:
> sorry for directly pushing - I tried to move as much of the calculation out of `regenerateImports` for performance reasons. I've also moved it after `imports:extend` so we catch all changes.
> 
> modules can always remove nuxt imports in `imports:sources`, and we won't warn in this case.

Nice. I actually missed that it is regenerating 😮. 

Why is the priority stuff no longer important?

packages/nuxt/src/imports/module.ts
```diff
@@ -147,9 +147,9 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
         for (const i of imports) {
           if (!defaultImportSources.has(i.from)) {
             const value = i.as || i.name
-            if (defaultImports.has(value)) {
-              const relativePath = relative(nuxt.options.srcDir, i.from)
-              logger.error(`\`${value}\` is an auto-imported function that is in use by Nuxt. Overriding it will likely cause issues. Please consider renaming \`${value}\` at \`~/${relativePath}\`.`)
+            if (defaultImports.has(value) && (!i.priority || i.priority >= 0 /* default priority */)) {
```
OrbisK, line 150: i cant suggest my own pr, but like this:

```ts
if (defaultImports.has(value) && ((i.priority ?? -1) >= 0)) {
```
we may not need the extra comment. (maybe just preference/nit)

Issue #29850, opened 2024-11-08
build manifest can be preloaded or embedded in html


danielroe:

Norbiros:
@danielroe I think this should be marked as completed

danielroe:
I had in mind that we wouldn't need an external file at all - have it just in the initial html

danielroe:
this is implemented indeed - we already preload if needed, and in fact in other PRs we've improved things so the manifest isn't needed at all for hydration/initial render, only for onward navigation (and it's prefetched in this case)

PR #30017, opened 2024-11-22
perf(nuxt): preload app manifest

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/29850

### 📚 Description

Certain sites require the use of the app manifest immediately as they either need route rules or some data on the page being prerendered.

In these cases, Nuxt will load the file when it hydrates. This leads to a delayed time to interactive web vital as we're waiting for the manifest to download to fully finish hydrating our app. While this download should be fast as it's a static file, we're still adding network latency to the hydration beyond our initial js files.

To resolve this we can track the usage of the manifest in an SSR environment so we can determine if it needs to be preloaded, potentially allowing the file to be fetched immediately.

The default behavior for sites not immediately using the app manifest, Nuxt will load the manifest 1 second after the network is idle (if they have decent internet), we do not care about preloading for these sites.

While working on this PR I fixed a couple of bugs related `isPrerendered()`:
- Calling `isPrerendered()` client-side would fetch the app manifest and cause a 404 error, by importing the server module for the manifest we can get around this
- If we got to the `getRouteRules` within the function it would throw an error as the Nuxt context would be lost after the `await getAppManifest()` 

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of app manifest based on execution context (server vs. client).
	- Introduced optional property `_preloadManifest` to indicate manifest preloading in the SSR context.
	- Conditional preloading of the app manifest JSON file based on its enabled status.

- **Bug Fixes**
	- Improved clarity and readability of existing functions without altering functionality.

- **Documentation**
	- Updated comments and documentation for better context and understanding throughout the codebase.

- **Chores**
	- Updated expected snapshot for server bundle size test to reflect changes in the build process.


harlan-zw:
@danielroe the logic isn't exactly clear but it seems like maybe in a server context we should never be accessing the app manifest? Given we should have all the data we need from the ssr context. 

Not too sure how to fix the tests otherwise

danielroe (approved):
This looks great! Ultimately we may want to revisit requiring the route rules so early in the app startup:

https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/runtime/plugins/router.ts

PR #30028, opened 2024-11-24
refactor(nuxt): prefix all core modules with `nuxt:`

### LINKED ISSUE #29965: Inadvertently overriding a built-in nuxt module should warn the user
### Environment

Warn users if they create a module the same as a built-in Nuxt module.

### Reproduction

1. Create a module
2. Name it `components`
3. It will override the built-in `components` module

### Describe the bug

Overriding the default built-in module (something like `components`) should be allowed; however, it's not obvious that this will likely completely break the app. 

Nuxt should at least warn in the logs that a built-in is being overridden, etc.

### Additional context

This was demonstrated (accidentally) at Vue Toronto

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue
resolves #29965

### 📚 Description

This PR prefix all core modules name with `nuxt:` to harmonize then naming of core modules with core plugins.

This is kinda a breaking change if users were overwriting core modules ? 

## Release Notes

- **New Features**
	- Updated module naming conventions for clarity (e.g., `nuxt:components`, `nuxt:imports`, `nuxt:pages`).
	- Enhanced handling of component directories and router options for improved functionality.
	- Introduced new type templates for middleware and layouts to enhance type safety.
	- Improved dynamic updates for imports and better handling of component paths.

- **Bug Fixes**
	- Improved error handling and logging for schema loading and route parsing.

- **Documentation**
	- Enhanced comments and structure to clarify the integration of modules within the Nuxt framework. 

These changes aim to improve the overall organization, error handling, and functionality of the Nuxt framework for developers.

### DIFF
--- a/packages/nuxt/src/core/schema.ts
+++ b/packages/nuxt/src/core/schema.ts
@@ -16,7 +16,7 @@ import { createJiti } from 'jiti'
 
 export default defineNuxtModule({
   meta: {
-    name: 'nuxt-config-schema',
+    name: 'nuxt:nuxt-config-schema',
   },
   async setup (_, nuxt) {
     const resolver = createResolver(import.meta.url)
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -24,6 +24,7 @@ const OPTIONAL_PARAM_RE = /^\/?:.*(?:\?|\(\.\*\)\*)$/
 export default defineNuxtModule({
   meta: {
     name: 'nuxt:pages',
+    configKey: 'pages',
   },
   async setup (_options, nuxt) {
     const useExperimentalTypedPages = nuxt.options.experimental.typedPages
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -23,7 +23,7 @@ const OPTIONAL_PARAM_RE = /^\/?:.*(?:\?|\(\.\*\)\*)$/
 
 export default defineNuxtModule({
   meta: {
-    name: 'pages',
+    name: 'nuxt:pages',
   },
   async setup (_options, nuxt) {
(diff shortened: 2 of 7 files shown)


packages/nuxt/src/pages/module.ts
```diff
@@ -23,7 +23,7 @@ const OPTIONAL_PARAM_RE = /^\/?:.*(?:\?|\(\.\*\)\*)$/
 
 export default defineNuxtModule({
   meta: {
-    name: 'pages',
+    name: 'nuxt:pages',
```
danielroe, line 26:
```suggestion
    name: 'nuxt:pages',
    configKey: 'pages',
```

packages/nuxt/src/core/schema.ts
```diff
@@ -16,7 +16,7 @@ import { createJiti } from 'jiti'
 
 export default defineNuxtModule({
   meta: {
-    name: 'nuxt-config-schema',
+    name: 'nuxt:nuxt-config-schema',
```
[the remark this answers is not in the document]
danielroe: cc: @huang-julien - this is a good spot from 🐇 - we need to update the logic there, not just for this one module name but for all of them (could even simplify to a `startsWith`)

packages/nuxt/src/utils.ts
```diff
@@ -8,3 +8,5 @@ export function toArray<T> (value: T | T[]): T[] {
 export async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
+
+export const NUXT_PREFIX = 'nuxt:'
```
huang-julien, line 12: for better naming consistency, do you think we should use it everywhere (plugins + modules) to define names ?
`name: NUXT_PREFIX + 'page'`

danielroe: I think just using the string should be ok

danielroe (approved):

danielroe (approved):

PR #30038, opened 2024-11-26
fix(nuxt): nuxt link observer is null

Akryum:
### 📚 Description

This fixes a crash of the app if there is a "conflict" between the `unobserve` from the `onMounted`/`onNuxtReady` and from the `onBeforeUnmount` -- meaning the "unobserve" function (return of `observe()`) is called twice.

image

image

Then `observer` is already null and this causes the crash.

image
image

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved safety of the `nuxt-link` functionality by enhancing the handling of the `IntersectionObserver`, reducing potential runtime errors.
  
- **Style**
	- Minor adjustments made to comments and formatting for better clarity.


danielroe (approved):
thank you!

PR #27672, opened 2024-06-17
feat(kit,nuxt,schema): support experimental decorators syntax

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14126

### 📚 Description

This PR enables enabling decorator syntax across your entire Nuxt/Nitro app, powered by [esbuild](https://github.com/evanw/esbuild/releases/tag/v0.21.3). It also allows passing and customising esbuild options across nitro, webpack, and vite build/optimizeDeps configurations via a new top-level `esbuild` option in your `nuxt.config` file.

#### Usage

```ts
// nuxt.config.ts
export default defineNuxtConfig({
  experimental: {
    decorators: true,
  },
})
```

```ts
// app.vue
function something (_method: () => unknown) {
  return () => 'decorated'
}

class SomeClass {
  @something
  public someMethod () {
    return 'initial'
  }
}

const value = new SomeClass().someMethod()
// this will return 'decorated'
```

#### Background

For a long time, TypeScript has had support for decorators via `compilerOptions.experimentalDecorators`. This implementation predated the TC39 standardization process. Now, decorators are a [Stage 3 Proposal](https://github.com/tc39/proposal-decorators), and supported without special configuration in TS 5.0+ (see https://github.com/microsoft/TypeScript/pull/52582 and https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#decorators).

> [!CAUTION]
> Note that there may be changes before this finally lands in the JS standard.

> [!CAUTION]
> If you have previously used `experimentalDecorators` from TypeScript, the implementation and types are different. Please read [the proposal](https://github.com/tc39/proposal-decorators) before enabling.

TODO:
- [x] tests
- [x] update other usages of `esbuild`

PR #30066, opened 2024-11-27
feat(nuxt): use `oxc-parser` instead of esbuild + acorn

### LINKED ISSUE #29790: definePageMeta try to process with ancorn ts code. Vue composables don't do so.
### Environment

latest nuxt

### Reproduction

```
async function getRouteMeta(contents, absolutePath) {
  if (!(absolutePath in pageContentsCache) || pageContentsCache[absolutePath] !== contents) {
    pageContentsCache[absolutePath] = contents;
    delete metaCache$1[absolutePath];
  }
  if (absolutePath in metaCache$1 && metaCache$1[absolutePath]) {
    return metaCache$1[absolutePath];
  }
  const loader = getLoader(absolutePath);
  const scriptBlocks = !loader ? null : loader === "vue" ? extractScriptContent(contents) : [{ code: contents, loader }];
  if (!scriptBlocks) {
    metaCache$1[absolutePath] = {};
    return {};
  }
  const extractedMeta = {};
  for (const script of scriptBlocks) {
    if (!PAGE_META_RE.test(script.code)) {
      continue;
    }
    const js = await transform(script.code, { loader: script.loader });
    throw new Error(js.code)
    const ast = parse(js.code, {
      sourceType: "module",
      ecmaVersion: "latest",
      ranges: true
    });
    const dynamicProperties = /* @__PURE__ */ new Set();
    let foundMeta = false;
    walk(ast, {
      enter(node) {
        if (foundMeta) {
          return;
        }
        if (node.type !== "ExpressionStatement" || node.expression.type !== "CallExpression" || node.expression.callee.type !== "Identifier" || node.expression.callee.name !== "definePageMeta") {
          return;
        }
        foundMeta = true;
        const pageMetaArgument = node.expression.arguments[0];
        for (const key of extractionKeys) {
          const property = pageMetaArgument.properties.find((property2) => property2.type === "Property" && property2.key.type === "Identifier" && property2.key.name === key);
          if (!property) {
            continue;
          }
          if (property.value.type === "ObjectExpression") {
            const valueString = js.code.slice(property.value.range[0], property.value.range[1]);
            try {
              extractedMeta[key] = JSON.parse(runInNewContext(`JSON.stringify(${valueString})`, {}));
            } catch {
              console.debug(`[nuxt] Skipping extraction of \`${key}\` metadata as it is not JSON-serializable (reading \`${absolutePath}\`).`);
              dynamicProperties.add(key);
              continue;
            }
          }
          if (property.value.type === "ArrayExpression") {
            const values = [];
            for (const element of property.value.elements) {
              if (!element) {
                continue;
              }
              if (element.type !== "Literal" || typeof element.value !== "string") {
                console.debug(`[nuxt] Skipping extraction of \`${key}\` metadata as it is not an array of string literals (reading \`${absolutePath}\`).`);
                dynamicProperties.add(key);
                continue;
              }
              values.push(element.value);
            }
            extractedMeta[key] = values;
            continue;
          }
          if (property.value.type !== "Literal" || typeof property.value.value !== "string" && typeof property.value.value !== "boolean") {
            console.debug(`[nuxt] Skipping extraction of \`${key}\` metadata as it is not a string literal or array of string literals (reading \`${absolutePath}\`).`);
            dynamicProperties.add(key);
            continue;
          }
          extractedMeta[key] = property.value.value;
        }
        for (const property of pageMetaArgument.properties) {
          if (property.type !== "Property") {
            continue;
          }
          const isIdentifierOrLiteral = property.key.type === "Literal" || property.key.type === "Identifier";
          if (!isIdentifierOrLiteral) {
            continue;
          }
          const name = property.key.type === "Identifier" ? property.key.name : String(property.value);
          if (name) {
            dynamicProperties.add("meta");
            break;
          }
        }
        if (dynamicProperties.size) {
          extractedMeta.meta ?? (extractedMeta.meta = {});
          extractedMeta.meta[DYNAMIC_META_KEY] = dynamicProperties;
        }
      }
    });
  }
  metaCache$1[absolutePath] = extractedMeta;
  return extractedMeta;
}
```

### Describe the bug

Ts code with decorators is passed to acorn.parse which can parse only js code. I see some plugins for ancorn to support ts!?

```
 const ast = parse(js.code, {
      sourceType: "module",
      ecmaVersion: "latest",
      ranges: true
    });
```

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/29790
closes https://github.com/nuxt/nuxt/pull/30025

### 📚 Description

this moves from `acorn` + `esbuild` to using `oxc-parser` directly to parse code in our plugins. The main benefits are:

1. ~~⚡️ speed - this is faster and has the potential to be faster still - need to create some benchmarks to measure~~
2. ~~🎨 decorator support - this is natively supported by `oxc-parser` - we'll need to update https://github.com/nuxt/nuxt/pull/27672 to apply only to nitro~~
3. 📐 alignment with vite move to oxc/rolldown toolchain
4. native support for parsing TS/TSX where before we only supported JS (see https://github.com/nuxt/nuxt/issues/29790)

This PR also includes additional tests for the transformation plugins and fixes a few bugs I found along the way, including removing type assertions ~~and relying on the auto-generated AST types from `oxc`~~.

### 🚧 TODO

- [x] fix upstream issue in `oxc-parser` with unicode: https://github.com/oxc-project/oxc/issues/7508


packages/nuxt/src/core/utils/parse.ts
```diff
@@ -0,0 +1,31 @@
+import { parseSync } from 'oxc-parser'
+import type { CatchClause, ClassBody, Declaration, Expression, MethodDefinition, ModuleDeclaration, ObjectProperty, Pattern, PrivateIdentifier, Program, PropertyDefinition, SpreadElement, Statement, Super, SwitchCase, TemplateElement } from 'oxc-parser'
+import { walk as _walk } from 'estree-walker'
+import type { SyncHandler } from 'estree-walker'
+import type { Node as ESTreeNode, Program as ESTreeProgram, ModuleSpecifier } from 'estree'
+
+/** estree also has AssignmentProperty, Identifier and Literal as possible node types */
+export type Node = Declaration | Expression | ClassBody | CatchClause | MethodDefinition | ModuleDeclaration | ModuleSpecifier | Pattern | PrivateIdentifier | Program | SpreadElement | Statement | Super | SwitchCase | TemplateElement | ObjectProperty | PropertyDefinition
+
+type InferThis<T extends (...args: any[]) => any> = T extends (this: infer U, ...args: infer A) => any ? U : unknown
```
snewell92, line 10: Can `InferThis` be replaced with `ThisParameterType` [handbook link](https://www.typescriptlang.org/docs/handbook/utility-types.html#thisparametertypetype)

Courtesy of [Dillon](https://bsky.app/profile/dillonis.online/post/3lbwtwbf4ck2b), thanks @dmmulroy

danielroe: perfect! TIL.

harlan-zw:
This is cool, love the utils. 

I was wondering if we could export these within @nuxt/kit as many modules have the same requirements and the util API seems fairly stable.

danielroe:
Great idea! I've created [`oxc-walker`](https://github.com/danielroe/oxc-walker) as a separate util for now. Happy to retire it or donate the name to the oxc team if they want (cc: @boshen).

(We could also create a workaround for the incorrect start/end positions with unicode text within that library.)

Boshen:
The latest release (v0.40.0) contains many useful utilities for parser plugins, see API example in https://www.npmjs.com/package/oxc-parser

danielroe:
@Boshen This looks amazing!

Right now we are missing the `hasChanged()` utility, which is a slight performance improvement for rollup - we can return undefined from a transform hook if no replacements are made.

The other key utility is `generateMap()`, which I'm sure you're aware of - it's one of the key reasons we use `magic-string`.

Finally, an edge case is that sometimes (once/twice) we use `magic-string` without parsing the code, and then parse the code _if necessary_. There may be no solution for this.

Boshen:
@danielroe Added `.hasChanged()` and `.generateMap()` in v0.47.1. Please let me know if there are more missing APIs.

PR #30122, opened 2024-12-02
docs: add giget limitation in nuxt layers documentation

### LINKED ISSUE #30103: Can't use dependencies inherited from remote layer inside app
### Environment

- Operating System: Darwin
- Node Version:     v20.18.0
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Package Manager:  pnpm@9.14.2
- Builder:          -
- User Config:      default
- Runtime Modules:  @nuxt/eslint
- Build Modules:    -

### Reproduction

Layer: https://stackblitz.com/~/github.com/mrleblanc101/repro-dependencies-layer
App: https://stackblitz.com/~/github.com/mrleblanc101/repro-dependencies-app

I don't think we can test in StackBlitz because of `c12 / giget` not working there, so here are the Github repo
Layer: https://github.com/mrleblanc101/repro-dependencies-layer
App: https://github.com/mrleblanc101/repro-dependencies-app

### Describe the bug

In the app, when running `pnpm install && pnpm run dev`, the app dependencies are installed and the layers dependencies are also installed.
See tailwind-merge is installed properly from the layer:
Image

But in VS Code, the dependencies installed by the layer are not recognized by `ts-plugin`, but the `twMerge` package still work: 
Image

I have a similar problem with other dependencies.
See eslint-config-prettier is installed properly from the layer:
Image

Also, there is an error in the `pnpm run dev`, but it doesn't crash and the app still render:
Image

I think this is because `tailwind-merge` and `eslint-config-prettier` are installed inside `node_modules/.c12/layer_name/node_modules` and not at the root of the `node_modules`.
I have tried with and without `shamefully-hoist=true` in `.npmrc` inside the app and the layer.
I also made sure that the packages are installed inside `dependencies` and not `devDependencies` because the later would then not be installed inside the app in the postinstall `nuxi prepare`
Image

### Additional context

_No response_

### Logs

```shell-script

```

mrleblanc101:
### 🔗 Linked issue

fix #30103

### 📚 Description

Improve the documentation by adding details about giget limitation.

## Summary by CodeRabbit

- **Documentation**
	- Enhanced explanations of Nuxt layers for better understanding and usability.
	- Clarified the required structure and essential files for Nuxt layers.
	- Added a warning regarding dependencies when using Giget with Nuxt layers.
	- Expanded guidance on publishing layers and using npm packages.
	- Provided important notes on relative paths and multi-layer support for Nuxt modules.

### DIFF
--- a/docs/2.guide/3.going-further/7.layers.md
+++ b/docs/2.guide/3.going-further/7.layers.md
@@ -104,6 +104,11 @@ If you want to extend a private remote source, you need to add the environment v
 If you want to extend a remote source from a self-hosted GitHub or GitLab instance, you need to supply its URL with the `GIGET_GITHUB_URL=<url>` or `GIGET_GITLAB_URL=<url>` environment variable - or directly configure it with [the `auth` option](https://github.com/unjs/c12#extending-config-layer-from-remote-sources) in your `nuxt.config`.
 ::
 
+::warning
+Nuxt Layers that are installed with Giget should only requires Nuxt Modules, since dependencies that are not Nuxt modules cannot be resolved.
+If you wish to install dependencies in your layer to be shared with the Nuxt app, please install your Nuxt Layer using a package manager.
+::
+
 ::note
 When using git remote sources, if a layer has npm dependencies and you wish to install them, you can do so by specifying `install: true` in your layer options.
 
--- a/docs/2.guide/3.going-further/7.layers.md
+++ b/docs/2.guide/3.going-further/7.layers.md
@@ -105,8 +105,8 @@ If you want to extend a remote source from a self-hosted GitHub or GitLab instan
 ::
 
 ::warning
-Nuxt Layers that are installed with Giget should only requires Nuxt Modules, since dependencies that are not Nuxt modules cannot be resolved.
-If you wish to install dependencies in your layer to be shared with the Nuxt app, please install your Nuxt Layer using a package manager.
++Nuxt Layers installed with Giget should only require Nuxt modules. Other dependencies cannot be resolved because they are installed in a nested structure (node_modules/.c12/layer_name/node_modules) rather than at the root level.
++To share dependencies between your layer and the Nuxt app, install your Nuxt Layer using a package manager instead of Giget.
 ::
 
 ::note
--- a/docs/2.guide/3.going-further/7.layers.md
+++ b/docs/2.guide/3.going-further/7.layers.md
@@ -105,8 +105,7 @@ If you want to extend a remote source from a self-hosted GitHub or GitLab instan
 ::
 
 ::warning
-Nuxt Layers installed with Giget should only require Nuxt modules. Other dependencies cannot be resolved because they are installed in a nested structure (node_modules/.c12/layer_name/node_modules) rather than at the root level.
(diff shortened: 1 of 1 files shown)


docs/2.guide/3.going-further/7.layers.md
```diff
@@ -104,6 +104,11 @@ If you want to extend a private remote source, you need to add the environment v
 If you want to extend a remote source from a self-hosted GitHub or GitLab instance, you need to supply its URL with the `GIGET_GITHUB_URL=<url>` or `GIGET_GITLAB_URL=<url>` environment variable - or directly configure it with [the `auth` option](https://github.com/unjs/c12#extending-config-layer-from-remote-sources) in your `nuxt.config`.
 ::
 
+::warning
++Nuxt Layers installed with Giget should only require Nuxt modules. Other dependencies cannot be resolved because they are installed in a nested structure (node_modules/.c12/layer_name/node_modules) rather than at the root level.
++To share dependencies between your layer and the Nuxt app, install your Nuxt Layer using a package manager instead of Giget.
```
danielroe: you should be able to share dependencies between your layer and the nuxt app.

I thought your issue was about sharing dependencies between eslint and the layer?

mrleblanc101: Oh, I think I misunderstood you.
This is quite complex !

I guess I thought the dependencies wouldn't work because of this error in VS Code, but now I see that this is probably also an ESLint error, since as stated previously, the `twMerge` function works even with the error.
391588409-572f6ca4-90dc-4530-b5ba-b895384a6567

So if I understand properly, dependencies installed by giget will work if they are required by the nuxt app, but won't work if they are only required outside the nuxt app because Nuxt do some magic behind the scene with the include/exclude (For exemple, `.nuxt/tsconfig.json`), but can't do it for dependencies that are required outside of Nuxt ?

docs/2.guide/3.going-further/7.layers.md
```diff
@@ -104,6 +104,11 @@ If you want to extend a private remote source, you need to add the environment v
 If you want to extend a remote source from a self-hosted GitHub or GitLab instance, you need to supply its URL with the `GIGET_GITHUB_URL=<url>` or `GIGET_GITLAB_URL=<url>` environment variable - or directly configure it with [the `auth` option](https://github.com/unjs/c12#extending-config-layer-from-remote-sources) in your `nuxt.config`.
 ::
 
+::warning
+Nuxt Layers installed with Giget should only require Nuxt modules. Other dependencies cannot be resolved because they are installed in a nested structure (node_modules/.c12/layer_name/node_modules) rather than at the root level.
+To share dependencies between your layer and the Nuxt app, install your Nuxt Layer using a package manager instead of Giget.
```
danielroe, line 109:
```suggestion
Bear in mind that if you are extending a remote source as a layer, you will not be able to access its dependencies outside of Nuxt. For example, if the remote layer depends on an eslint plugin, this will not be usable in your eslint config. That is because these dependencies will be located in a special location (`node_modules/.c12/layer_name/node_modules/`) that is not accessible to your package manager.
```

danielroe (approved):

danielroe:
Thank you for this ❤️

PR #30171, opened 2024-12-06
docs: added note about awaiting `useLazyFetch`

### LINKED ISSUE #29931: Documentation: Consistent examples for useFetchLazy
At the moment the documentation uses useLazyFetch (and useFetch with lazy option) without await (https://nuxt.com/docs/getting-started/data-fetching#lazy) as well as with await (https://nuxt.com/docs/api/composables/use-lazy-fetch). Which example is correct? Or is there any difference? But I think it would be good to agree on one way, to avoid confusion.

bjoaquinc:
Updated use-lazy-fetch.md by adding a note that await only ensures that the call is initialized, not that the data is available.

### 🔗 Linked issue

resolves #29931 

### 📚 Description

As suggested, I added a note that addresses the confusion about the functionality of `await` on `useLazyFetch()`

## Summary by CodeRabbit

- **Documentation**
	- Updated the documentation for `useLazyFetch` to clarify its functionality and usage.
	- Added a note regarding the behavior of data fetching and the importance of handling 'pending' and 'error' states.
	- Included a warning against naming custom functions with the same name as `useLazyFetch`.

### DIFF
--- a/docs/3.api/2.composables/use-lazy-fetch.md
+++ b/docs/3.api/2.composables/use-lazy-fetch.md
@@ -17,7 +17,7 @@ By default, [`useFetch`](/docs/api/composables/use-fetch) blocks navigation unti
 ::
 
 ::note
-Awaiting `useLazyFetch` in this mode only ensures the call is initialized; data will be fetched client-side after navigation and may not be immediately available.
+Awaiting `useLazyFetch` in this mode only ensures the call is initialized. On client-side navigation, data may not be immediately available, and you should make sure to handle the pending state in your app.
 ::
 
 :read-more{to="/docs/api/composables/use-fetch"}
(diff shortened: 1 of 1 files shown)


docs/3.api/2.composables/use-lazy-fetch.md
```diff
@@ -16,6 +16,10 @@ By default, [`useFetch`](/docs/api/composables/use-fetch) blocks navigation unti
 `useLazyFetch` has the same signature as [`useFetch`](/docs/api/composables/use-fetch).
 ::
 
+::note
+Awaiting `useLazyFetch` in this mode only ensures the call is initialized; data will be fetched client-side after navigation and may not be immediately available.
```
danielroe, line 20:
```suggestion
Awaiting `useLazyFetch` in this mode only ensures the call is initialized. On client-side navigation, data may not be immediately available, and you should make sure to handle the pending state in your app.
```

danielroe (approved):
thanks - nice addition!

PR #30190, opened 2024-12-07
fix(nuxt): if href in nuxt-link starts with hash, render a tag directly

### LINKED ISSUE #30074: `NuxtLink` is not drop-in replacement for `<a href="?query">` or `<a href="#hash">`
### Environment

- Node Version:     v22.10.0
- Nuxt Version:     3.14.159
- CLI Version:      3.15.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.9.0
- Builder:          -
- User Config:      default
- Runtime Modules:  @nuxt/eslint@0.6.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-zxu7oo?file=pages%2Findex.vue

### Describe the bug

When `NuxtLink`'s `href` is set to a string starting with `?` or `#`, the `a` tag's rendered `href` value will contain not only that string but also the current route's path prepended to it. This is contrary to the vanilla HTML behavior: `a` tags are allowed to have an `href` value starting with `?` or `#` with no path preceding.

In both cases, this can break CSS selectors that filter on the `href` attribute. In the case of `#`, prepending the path can break the link since it excludes the query. Simply rendering the `href` the way it was originally set, as expected, would fix this. In fact, when the `href` starts with `#`, it could be rendered as a vanilla `a` tag rather than a router link. No event handler is needed for that.

Reproduction steps:
1. Inspect the "Query" link. You'll find `href="/?query=1"`, and its text color is red.
2. Click the "Query" link. The current route changes to `/?query=1`.
3. Inspect the "Hash" link. You'll find `href="/#hash"`, and its text color is red.
4. Click the "Hash" link. The current route changes to `/#hash`.

Expected:
- All links should be colored blue, not red.
- Inspecting the "Query" link should show `href="?query=1"`, as expected from the source code.
- Inspecting the "Hash" link should show `href="#hash"`, as expected from the source code.
- Clicking "Query" and then "Hash" should result in the current route `/?query=1#hash`.

### Additional context

When `NuxtLink` is used in `app.vue` instead of in a page, it correctly mirrors the value set in the `href` rather than prepending a path. This issue only affects `NuxtLink`s used in pages.

vmrjnvc:
### 🔗 Linked issue

resolves #30074 

### 📚 Description

When a NuxtLink component's href starts with # or ?, the rendered <a> tag's href includes the current route's path prepended to it. This issue only occurs when the component is located within the pages folder.  

To address this issue, I followed the suggestion from the comments section of the open issue. I have implemented a solution specifically for cases where the href starts with '#'. In such cases, the component will render an <a> tag directly instead of using RouterLink.

Currently, this PR focuses on fixing the issue with #. I am open to feedback and to extend the solution to handle cases where the href starts with '?', provided there's guidance on the best approach for tackling that problem.  

### Additional note

This is my first PR here, so while I’ve done my best to follow the contribution guidelines, I welcome any suggestions for improvement.


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -176,7 +182,9 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
 
     // Resolves `to` value if it's a route location object
     const href = computed(() => {
-      if (!to.value || isAbsoluteUrl.value) { return to.value as string }
+      if ((!to.value || isAbsoluteUrl.value) || isHashLinkWithoutHashMode(to.value as string, hashMode)) {
```
danielroe, line 185:
```suggestion
      if (!to.value || isAbsoluteUrl.value || isHashLinkWithoutHashMode(to.value as string, hashMode)) {
```

danielroe (commented):
thank you - this looks good! would you add a test in `packages/nuxt/test/nuxt-link.test.ts`?

vmrjnvc:
@danielroe  I've added test cases for that and made the changes you suggested. Just let me know if they need any changes, and I'll update them

danielroe:
(no need to merge in the latest main branch unless there are conflicts)

vmrjnvc:
@danielroe I've made several changes to resolve the issues that were causing tests to fail.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -110,6 +110,10 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
     }
   }
 
+  function isHashLinkWithoutHashMode (link: unknown, hashMode: boolean): boolean {
+    return typeof link === 'string' && link.startsWith('#') && !hashMode
```
danielroe, line 114:
```suggestion
    return !hashMode && typeof link === 'string' && link.startsWith('#')
```

TheAlexLichter:
I wonder if this would cause any unexpected gotcha's with transitions between links. But as all links where just the hash changes should only move to certain anchors on the same page, this seems fine to me.

For query parameters, proper handling might be needed (but happy to be proven wrong otherwise).

danielroe:
thank you very much @vmrjnvc ❤️

PR #30192, opened 2024-12-07
fix(schema): fix router-options type

wzc520pyfm:
### 📚 Description

The [nuxt documentation](https://nuxt.com/docs/guide/recipes/custom-routing#custom-history-advanced) states that `history` can return `null` or `undefined`, but this will cause a type error in the code.

image
image

## Summary by CodeRabbit

- **New Features**
	- Enhanced flexibility of the `history` function to return `null` or `undefined`, allowing for better handling of various scenarios.


danielroe (approved):
thank you!

PR #30195, opened 2024-12-07
perf(nuxt): prefer structuredClone over stringify/parse

huang-julien:
### 📚 Description

`structuredClone` is faster than the good old `JSON.stringify(JSON.parse())` 🫡
image

## Summary by CodeRabbit

- **New Features**
	- Introduced a new function for generating type templates for imports, enhancing type declaration management.
	- Improved the cloning process for import presets, ensuring better handling of object structures.
	- Added a caching mechanism for import paths to optimize path resolution.

- **Bug Fixes**
	- Enhanced path resolution logic to prevent issues with shared references in import presets.

### DIFF
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -33,7 +33,7 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
   }),
   async setup (options, nuxt) {
     // TODO: fix sharing of defaults between invocations of modules
-    const presets = JSON.parse(JSON.stringify(options.presets)) as ImportPresetWithDeprecation[]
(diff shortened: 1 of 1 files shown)


danielroe:
Hm. I thought it was much slower.

Let's see if I can run a benchmark.

danielroe:
Testing on our default presets with https://jsbench.me:

CleanShot 2024-12-09 at 10 59 10@2x

it seems `structuredClone` is still 25% slower.

pi0:
yep it is slower. might worth to bench against [klona](https://www.npmjs.com/package/klona) too btw might be faster (if also this is really a perf point for us ofc!)

PR #30199, opened 2024-12-08
docs: update `nuxi` command pages

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/cli/pull/592
* https://github.com/nuxt/module-builder/issues/356

### 📚 Description
This changes the command pages description to more closely match the actual help/usage output of nuxt cli, unfortunately lots of the default values shown originally are lost, those values aren't exactly defaults but fall back values (the argument/option is unset and its value is based on environment variables, upstream cli fallbacks or nuxt configuration).

The values are extracted from the command definitions with a (very unreadable and wacky) script I wrote, I'm going to see if I can clean it up and perhaps find a way to ease syncing the commands docs with their actual config definitions.


BobbieGoede:
I don't think CI is failing due to changes made in this PR 🤔

Also, the marker comments in the markdown will be removed at some point, I'm using these to find and replace the content extracted from the nuxt cli command definitions.

Deeyuude:
Great work 😊 a question please : could we use the --channel with nuxi init ?

BobbieGoede:
> Great work 😊 a question please : could we use the --channel with nuxi init ?

Thanks! The `init` command currently does not support the `--channel` flag, no. This PR only updates the documentation for the cli commands, you can open a feature request (or a PR) on https://github.com/nuxt/cli if you would like this functionality to be added 🙏

danielroe:
@BobbieGoede any chance you could update this following the latest changes in https://github.com/nuxt/cli/pull/592 ?

Also, if there's a way we can add (even a WIP) of your tool to the Nuxt repo so we could auto-update the docs as things change in the CLI, that would be genuinely amazing.

danielroe (approved):

BobbieGoede:
@danielroe 

Actually the updated descriptions in this PR are not really the same as they are on the nuxt/cli repo, it depends on more changes (https://github.com/nuxt/cli/pull/601).

>Also, if there's a way we can add (even a WIP) of your tool to the Nuxt repo so we could auto-update the docs as things change in the CLI, that would be genuinely amazing.

I actually have changes across nuxt/cli and citty (different defaults, `negativeDescriptions`) locally which makes it a bit more difficult to share something that will have the same output. 

The script is honestly really messy, I'll share it as a gist in a bit so you can decide whether to include it here or not 😂 I was thinking to generalize it, remove the file writing, and publish it as a small package that just returns usage/help data for a given citty command when I have time.

BobbieGoede:
@danielroe 

Here is the code I used https://gist.github.com/BobbieGoede/b58803b8285cd3075f54ff03f33e45a8, use at your own risk :face_with_peeking_eye:

PR #30241, opened 2024-12-12
feat(nuxt): support local functions in definePageMeta

### LINKED ISSUE #30197: `__name is not defined` when using `definePageMeta.validate`
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.9.0`
- Nuxt Version:     `3.14.1592`
- CLI Version:      `3.16.0`
- Nitro Version:    `2.10.4`
- Package Manager:  `pnpm@9.14.4`
- Builder:          `-`
- User Config:      `default`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-frp6zf8s?file=pages%2Fdynamic%2F[id].vue,nuxt.config.ts

### Describe the bug

I've noticed that the dev server fails to start when using [route validation](https://nuxt.com/docs/getting-started/routing#route-validation) in `definePageMeta`, and `vite.esbuild.keepNames` is set to `true` in the `nuxt.config.ts`.

The following error message appears in the console:
```
 ERROR  [nuxt] [request error] [unhandled] [500] __name is not defined
```

```ts
definePageMeta({
  // removing this check makes the error disappear
  validate: ({ params }) => {
    return !!(params.id && !isNaN(Number(params.id)));
  },
})
```

I believe this is a regression, but I cannot pinpoint the exact version where it started.
Could it be an upstram issue maybe? Or have something to do with https://github.com/nuxt/nuxt/pull/29869 ?

### Additional context

_No response_

### Logs

```shell-script

```

cernymatej:
### 🔗 Linked issue

resolves #30197

### 📚 Description

This PR adds support for locally defined functions in `definePageMeta`. A scope tracker was implemented to keep track of function's dependencies.

This fixes a problem with esbuild defining helper functions when using the `vide.esbuild.keepNames = true` option, for example:
```ts
var __name = (target, value) => __defProp(target, "name", { value, configurable: true });
// ...

definePageMeta({
  validate: /* @__PURE__ */ __name(({ params }) => {
    return !!(params.id && !isNaN(Number(params.id)));
  }, "validate")
})
```

but also maybe a slightly more common use case, where the user provides a named function, regardless of the `esbuild.keepNames` setting:
```ts
function validateIdParam(route) {
  return !!(route.params.id && !isNaN(Number(route.params.id)))
}

definePageMeta({
  validate: validateIdParam,
})
```

📝 TODO
- [ ] ~Detect impure functions~
- [ ] ~Show a readable error message when an impure function is detected~
- [x] Consider supporting imported functions
- [x] Add tests
- [x] Use the new Scope Tracker in `nuxt/packages/nuxt/src/core/plugins/composable-keys.ts` too
- [x] Update the docs


packages/nuxt/src/core/utils/parse.ts
```diff
@@ -31,3 +31,173 @@ export function parseAndWalk (code: string, _sourceFilename: string, callback: {
 export function withLocations<T> (node: T): WithLocations<T> {
   return node as WithLocations<T>
 }
+
+export function createScopeTracker<T> () {
```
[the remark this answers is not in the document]
cernymatej: I'll have a look at that as well :)

cernymatej:
Initially, I wanted to extract only the functions that don’t reference anything outside their block scope and throw an error otherwise. However, I quickly noticed this approach wouldn’t work, as it treated JavaScript globals like `Number`, `Object`, etc., as undeclared references. Do you have any ideas on how we could exclude these from the detection? Or should we just let it be?

Another feature I wanted to add support for was the use of imported functions. I found out that this is already supported, but we’re not currently checking for any side effects in this situation either. Do you think it’s worth exploring further, or should we leave it as is, like the previous point?

---

Btw. the autofix ci seems to be doing some nasty stuff 😬😄 Example: https://github.com/nuxt/nuxt/pull/30241/commits/e55682b7e78300c702b63093e8693fce788d52a5

cernymatej:
The new scope tracker should also address some issues with key injection. Consider the following example:
```ts
function foo(useAsyncData: (key?: any) => void) {
    useAsyncData()
}

foo((key) => {
    console.log(key)
})
```
In the current version of Nuxt, the key would be injected here because the original tracker didn’t account for function parameters, among other things.

That said, I might have missed something too, so I’d be very grateful for a review. 🙏

docs/2.guide/2.directory-structure/1.pages.md
```diff
@@ -266,17 +266,24 @@ console.log(route.meta.title) // My home page
...
 <script setup lang="ts">
 import { someData } from '~/utils/example'
 
-const title = ref('')
+function validateIdParam(route) {
+  return route.params.id && !isNaN(Number(route.params.id))
+}
 
 definePageMeta({
-  title,  // This will create an error
```
danielroe, line 278: I think it's worth preserving, as this example is a warning example...

cernymatej: It won't create an error now, though. The ref will be extracted too
```ts
const title = ref("");
const __nuxt_page_meta = {
      title
    }
export default __nuxt_page_meta
```

It doesn't really make sense to use a ref in this way, since it will be a different ref than the one in the component - hence the warning "Make sure not to reference any reactive data"

I've added it back with a different comment

danielroe: ah, of course.

thank you 🙏

danielroe (approved):

danielroe:
Amazing work - thank you so much ❤️

PR #30277, opened 2024-12-16
fix(nuxt): ensure `getRouteRules` works with nitro signature

danielroe:
### 📚 Description

reported in https://github.com/nitrojs/nitro/issues/2939

we need to make sure the type is compatible with the one exposed by Nitro, to avoid type errors

cc: @pi0

### DIFF
--- a/packages/nuxt/src/app/composables/manifest.ts
+++ b/packages/nuxt/src/app/composables/manifest.ts
@@ -1,6 +1,8 @@
 import type { MatcherExport, RouteMatcher } from 'radix3'
 import { createMatcherFromExport, createRouter as createRadixRouter, toRouteMatcher } from 'radix3'
 import { defu } from 'defu'
+import type { H3Event } from 'h3'
+import type { NitroRouteRules } from 'nitro/types'
 import { useNuxtApp, useRuntimeConfig } from '../nuxt'
 // @ts-expect-error virtual file
 import { appManifest as isAppManifestEnabled } from '#build/nuxt.config.mjs'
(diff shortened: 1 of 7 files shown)


packages/nuxt/src/app/composables/manifest.ts
```diff
@@ -52,7 +54,10 @@ export function getAppManifest (): Promise<NuxtAppManifest> {
 }
 
 /** @since 3.7.4 */
-export async function getRouteRules (url: string) {
+export async function getRouteRules (url: string): Promise<Record<string, any>>
+export async function getRouteRules (event: H3Event): Promise<NitroRouteRules>
+export async function getRouteRules (url: string | H3Event) {
```
pi0, line 59: I wish we never overlapped but perhaps would be safer to deprecate `string` version and support `{ path: string }` interface?

danielroe: the key thing that's missing from the nitro implementation is the ability to get the route rules for an arbitrary path...

might it be possible for nitro to handle a path rather than just the current event?

pi0: Possible, yes but I'm afraid to do this because it imposes a new limit that route rules cannot be aware of event context anymore. In meanwhile, Nuxt impl can be API-compatible with event interface.

pi0 (approved):
❤️

moshetanzer:
Thanks for all the work @danielroe and @pi0. Love the outcome!

PR #30283, opened 2024-12-17
fix(nuxt): respect `replace` in middleware with `navigateTo`

### LINKED ISSUE #30282: navigateTo doesn't use NavigateToOptions in middleware
### Environment

- Node Version:     v22.12.0
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.9.2
- Builder:          -
- User Config:      default
- Runtime Modules:  @nuxt/eslint@0.7.3, @nuxtjs/tailwindcss@6.12.2, @primevue/nuxt-module@4.2.5, @pinia/nuxt@0.9.0, @vueuse/nuxt@12.0.0, @vee-validate/nuxt@4.14.7, @nuxtjs/i18n@9.1.1
- Build Modules:    -

### Reproduction

Just read my description its not that hard to understand.

### Describe the bug

navigateTo doesn't use NavigateToOptions in middleware.

For example:
```ts
export default defineNuxtRouteMiddleware(async (to) => {
    return navigateTo("/login", { replace: true });
});
```
The `{ replace: true }` doesn't work.

This line returns a string: https://github.com/nuxt/nuxt/blob/48e50c030ab75ef6e50769c61aec2a02f61b6126/packages/nuxt/src/app/composables/router.ts#L155
This line uses router.beforeEach that returns above string: https://github.com/nuxt/nuxt/blob/48e50c030ab75ef6e50769c61aec2a02f61b6126/packages/nuxt/src/pages/runtime/plugins/router.ts#L225

So what ends up happening is,

This:
```ts
export default defineNuxtRouteMiddleware(async (to) => {
    return navigateTo("/login", { replace: true });
});
```

Gets turned into:
```ts
router.beforeEach(() => {
  return "/login";
});
```

Which means NavigateToOptions is not even used at all. 

### Additional context

_No response_

### Logs

```shell-script

```

murshex:
### 🔗 Linked issue

resolves #30282

### 📚 Description

navigateTo doesn't use NavigateToOptions in middleware. This fixes it.

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -153,7 +153,7 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
     const replace = options?.replace ?? false
-    return typeof to === 'string' ? { path: to, replace  } : { ...to, replace }
+    return typeof to === 'string' ? { path: to, replace } : { ...to, replace }
   }
 
   const router = useRouter()
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -152,7 +152,7 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
 
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
-    return to
+    return { path: to, replace: options?.replace ?? false }
   }
 
   const router = useRouter()
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -152,8 +152,10 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
 
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
-    const replace = options?.replace ?? false
-    return typeof to === 'string' ? { path: to, replace } : { ...to, replace }
+    if (options?.replace) {
+      return typeof to === 'string' ? { path: to, replace: true } : { ...to, replace: true }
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/composables/router.ts
```diff
@@ -152,7 +152,7 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
 
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
-    return to
+    return { path: to, replace: options?.replace ?? false }
```
danielroe, line 155: `to` might be an object - would you be able to handle that? 🙏

murshex: https://github.com/nuxt/nuxt/pull/30283/commits/3053972c4b625e0d60a184ed76c54d4949874aea
Should fix it

packages/nuxt/src/app/composables/router.ts
```diff
@@ -152,7 +152,8 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
 
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
-    return to
+    const replace = options?.replace ?? false
+    return typeof to === 'string' ? { path: to, replace } : { ...to, replace }
```
danielroe, line 156:
```suggestion
    if (options?.replace) {
      return typeof to === 'string' ? { path: to, replace: true } : { ...to, replace: true }
    }
    return to
```

danielroe (approved):
thank you very much for this ❤️

nicky1038:
Hi everyone!
This solution seem to work wrong when `to` contains query params, e. g. `/some/path?foo=bar`.
in nuxt >=3.15.0 passing such an url to the `navigateTo` function along with `replace: true` option leads to losing query params.

PR #30292, opened 2024-12-18
docs: update auto-imports to advertise the scan feature

skoenfaelt:
- advertise imports.scan feature
- name limitations in layer systems

### 📚 Description

The source came from the video of alexander lichter "Auto Imports in Vue and Nuxt - The Good and the Bad" https://youtu.be/YHdDraNNGCk?si=PFNmmI6o2JSXj9KQ&t=910

I saw, that this is not covered in the docs. As I tried it in my project I faced issues with the layer system.
Please review the "### Partially Disabling Auto-imports" Section and the caution I mentioned. Let me know if I should remove that  caution part from the docs.

### DIFF
--- a/docs/2.guide/1.concepts/1.auto-imports.md
+++ b/docs/2.guide/1.concepts/1.auto-imports.md
@@ -140,8 +140,29 @@ export default defineNuxtConfig({
   }
 })
 ```
-
 This will disable auto-imports completely but it's still possible to use [explicit imports](#explicit-imports) from `#imports`.
+
+### Partially Disabling Auto-imports
+
+If you want framework-specific functions like `ref` to remain auto-imported but wish to disable auto-imports for your own code (e.g., custom composables), you can set the `imports.scan` option to `false` in your `nuxt.config.ts` file:
+
+```ts
+export default defineNuxtConfig({
+  imports: {
+    scan: false
+  }
+})
+```
+
+With this configuration:
+- Framework functions like `ref`, `computed`, or `watch` will still work without needing manual imports.
+- Custom code, such as composables, will need to be manually imported in your files.
+
+::warning
+**Caution:** This setup has certain limitations:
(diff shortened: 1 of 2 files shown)


derHodrig:
> https://nuxt.com/docs/api/nuxt-config#imports
> 
> Is `nuxt-config` also need to update?

Yes you are right, nuxt.config should also update to show this setting. I updated it, please have a look, if its done correctly.

docs/2.guide/1.concepts/1.auto-imports.md
```diff
@@ -140,9 +140,30 @@ export default defineNuxtConfig({
   }
 })
 ```
-
 This will disable auto-imports completely but it's still possible to use [explicit imports](#explicit-imports) from `#imports`.
```
danielroe, line 143:
```suggestion

This will disable auto-imports completely but it's still possible to use [explicit imports](#explicit-imports) from `#imports`.
```

docs/2.guide/1.concepts/1.auto-imports.md
```diff
@@ -140,9 +140,30 @@ export default defineNuxtConfig({
...
+})
+```
+
+With this configuration:
+- Framework functions like `ref`, `computed`, or `watch` will still work without needing manual imports.
+- Custom code, such as composables, will need to be manually imported in your files.
+
+::warning
+**Caution:** This setup has certain limitations:
+- If you structure your project to include an "abstract layer" which extends from a "core layer" (which is an npm package), you’ll need to import composables from the npm package similarly to third-party packages (e.g., `vue-i18n`) into your "abstract layer".
```
danielroe, line 163:
```suggestion
- If you structure your project with layers, you will need to explicitly import the composables from each layer, rather than relying on auto-imports.
```

danielroe (approved):
thank you ❤️

PR #30305, opened 2024-12-19
perf(schema): only inline styles for vue components

danielroe:
### 📚 Description

I think this is possibly a better default for 4.x. global CSS can be cached in a `.css` file and probably it's better not to include in the page HTML in this case, reserving inlined styles in HTML for _component-level_ CSS, which is unique to vue components.

### DIFF
--- a/packages/schema/src/config/experimental.ts
+++ b/packages/schema/src/config/experimental.ts
@@ -59,8 +59,8 @@ export default defineUntypedSchema({
         if (val === false || (await get('dev')) || (await get('ssr')) === false || (await get('builder')) === '@nuxt/webpack-builder') {
           return false
         }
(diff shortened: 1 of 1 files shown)


harlan-zw (approved):

florian-strasser:
I know this change was introduced way earlier, but within my tests after nuxt 4 release I see that most of my managed sites don't benefit from this change. Most have slightly worse Lighthouse results. We are talking about a drop of 3 - 4% at already top page speed sites. All my sites are generated statically, maybe that is the exclamation why the performance is not better for me.

I also know that I can still use this within nuxt config and get back to the "old" way, but I guess it should be standard when nuxt generate is used:
```
features: {
    inlineStyles: true,
  },
```

PR #30322, opened 2024-12-20
docs: add bluesky

iiio2:
I have added Nuxt's Bluesky profile in README.md. Actually I do not have the svg image. I think If we add the svg image, then this would be complete.


danielroe (approved):
thank you!

PR #30328, opened 2024-12-22
perf(nuxt): remove useId from composable key plugin

huang-julien:
### 📚 Description

During some `useId` debugging, i noticed it was keyed which is not necessary anymore with `useId` from vue


danielroe (approved):
👌

PR #30331, opened 2024-12-22
docs: add spacing

iiio2:

### DIFF
--- a/docs/2.guide/1.concepts/9.code-style.md
+++ b/docs/2.guide/1.concepts/9.code-style.md
@@ -1,5 +1,5 @@
 ---
-title: 'Code Style'
+title: "Code Style"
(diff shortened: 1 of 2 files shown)


danielroe (commented):
'middleware' can be both singular and plural and we have chosen to use it throughout the docs for both singular and plural

iiio2:
> 'middleware' can be both singular and plural and we have chosen to use it throughout the docs for both singular and plural

really ? I didn't know that. Thank you.

iiio2:
Untitled
@danielroe , In the picture, see, `if` is very close to ESLintv9. I think there should be a just one space. I think I have tried fo fix it.

danielroe (approved):

Issue #30289, opened 2024-12-17
Id from `useId()` changes on hydration.


lutejka:
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-bra69axh?file=app.vue&file=components%2FComp.vue

### Describe the bug

Once I use `useAsyncData` in `app.vue` or some layout, `useId()` within child components return different Id's after hydration.  As soon as I remove `useAsyncData`, everything works as intended. 

The error from the reproduction looks like this: 
```
Vue warn]: Hydration text content mismatch on 
<div data-v-inspector="components/Comp.vue:2:3"> 
  - rendered on server: v-0-0
  - expected on client: v-1 
  at <Comp> 
  at <App key=4 > 
  at <NuxtRoot>
```

### Additional context

Maybe relates to this issue: https://github.com/nuxt/nuxt/issues/27378

### Logs

```shell-script

```

huang-julien:
Hello, you forgot to await useAsyncDFata

cernymatej:
@huang-julien you shouldn't need to await it
or am i missing something?

huang-julien:
We do need to await it. Because this will be transformed by unctx to bind the context to the handler of `useAsyncData`. 
... Is there a place in the doc where we say to not await it ?

cernymatej:
I would think that the option to await it is there only so that you can access the data further down in the script setup if you need to. Even Daniel (cc @danielroe) often recommends not awaiting it when the data is only needed in the template, to avoid creating an unnecessary async setup function.

There is even a test case to verify that the data is available directly on the Promise itself:
https://github.com/nuxt/nuxt/blob/1f09e69335426c221617994aaeec556671c23aa4/test/nuxt/composables.test.ts#L128-L146

lutejka:
The documentation states neither that it is needed nor that it is not needed. But this note at least implies that it is optional.

https://github.com/nuxt/nuxt/blob/1f09e69335426c221617994aaeec556671c23aa4/docs/3.api/2.composables/use-async-data.md?plain=1#L102-L104

huang-julien:
Yeah, I get what both of you means.
[...]

PR #30343, opened 2024-12-23
fix(nuxt): mark instance with empty sp array in useAsyncData

huang-julien:
### 🔗 Linked issue

https://github.com/vuejs/core/issues/12591
https://github.com/nuxt/nuxt/issues/30289

### 📚 Description

This PR add an empty array to the instance.sp within `useAsyncData` to avoid having hydartion issues with `useId`.

Not a big fan of this solution as this feels like a huge workaround.

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -350,8 +350,10 @@ export function useAsyncData<
   if (import.meta.client) {
     // Setup hook callbacks once per instance
     const instance = getCurrentInstance()
-    // @ts-expect-error - internal vue property. This force vue to mark the component as async boundary client-side to avoid useId hydration issue since we treeshake onServerPrefetch
-    instance.sp = instance.sp || []
+    if (instance && fetchOnServer && options.immediate) {
+      // @ts-expect-error - internal vue property. This force vue to mark the component as async boundary client-side to avoid useId hydration issue since we treeshake onServerPrefetch
+      instance.sp = instance.sp || []
+    }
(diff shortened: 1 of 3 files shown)


danielroe:
can we call markAsyncBoundary directly instead?

huang-julien:
let's try but tests seems to fail

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -350,6 +350,8 @@ export function useAsyncData<
   if (import.meta.client) {
     // Setup hook callbacks once per instance
     const instance = getCurrentInstance()
+    // @ts-expect-error - internal vue property. This force vue to mark the component as async boundary client-side to avoid useId hydration issue since we treeshake onServerPrefetch
+    instance.sp = instance.sp || []
```
danielroe, line 354:
```suggestion
    if (instance) {  
      instance.sp = instance.sp || []
    }
```

danielroe, line 354:
```suggestion
    if (instance && fetchOnServer && options.immediate) {
      // @ts-expect-error - internal vue property. This force vue to mark the component as async boundary client-side to avoid useId hydration issue since we treeshake onServerPrefetch
      instance.sp = instance.sp || []
    }
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -350,6 +350,10 @@ export function useAsyncData<
   if (import.meta.client) {
     // Setup hook callbacks once per instance
     const instance = getCurrentInstance()
+    if (instance && fetchOnServer && options.immediate) {
```
edison1105, line 353:
```js
if (instance && fetchOnServer && options.immediate && !instance.sp) {
	instance.sp = []
}
```
This avoids unnecessary assignment

danielroe (approved):

PR #30349, opened 2024-12-23
fix(docs): AsyncData typings #23114

### LINKED ISSUE #23114: AsyncDataOptions type in docs is incorrect
### Environment

N/A

### Reproduction

N/A

### Describe the bug

`AsyncDataOptions` type snippet [in docs](https://nuxt.com/docs/api/composables/use-async-data#type) doesn't match the actual type from [the code](https://github.com/nuxt/nuxt/blob/a07b34a6edb2ab900c7397fcf0965b1a468d910a/packages/nuxt/src/app/composables/asyncData.ts#L34). Old type doesn't seem to have correct `transform` definition, instead making it a weird function that takes value in and then returns the same value back, which is not true, and not how `transform` is actually used.

### Additional context

This has already led to confusion in [johannschopplich/nuxt-api-party#49](https://github.com/johannschopplich/nuxt-api-party/issues/49#issuecomment-1712895777).

I wonder if it is even worth including type signatures in docs. After all, it's a hassle to extract and keep them up-to-date, and probably doesn't provide a lot of value to the reader. If they need to, they can just access the code. It's also possible to include GitHub search links instead, like this: [view AsyncDataOptions on GitHub](https://github.com/search?q=repo%3Anuxt%2Fnuxt+path%3Apackages%2Fnuxt%2Fsrc%2Fapp%2Fcomposables%2FasyncData.ts+symbol%3AAsyncDataOptions&type=code).

### Logs

_No response_

RokeAlvo:
### 🔗 Linked issue

resolve #23114

### 📚 Description

The typing of AsyncData in the documentation does not match reality. The type descriptions in the documentation have been replaced with a link to GitHub

### DIFF
--- a/docs/3.api/2.composables/use-async-data.md
+++ b/docs/3.api/2.composables/use-async-data.md
@@ -105,44 +105,7 @@ If you have not fetched data on the server (for example, with `server: false`), 
 
 ## Type
 
-```ts [Signature]
-function useAsyncData<DataT, DataE>(
-  handler: (nuxtApp?: NuxtApp) => Promise<DataT>,
-  options?: AsyncDataOptions<DataT>
-): AsyncData<DataT, DataE>
-function useAsyncData<DataT, DataE>(
-  key: string,
-  handler: (nuxtApp?: NuxtApp) => Promise<DataT>,
-  options?: AsyncDataOptions<DataT>
-): Promise<AsyncData<DataT, DataE>>
-
-type AsyncDataOptions<DataT> = {
-  server?: boolean
-  lazy?: boolean
-  immediate?: boolean
-  deep?: boolean
-  dedupe?: 'cancel' | 'defer'
-  default?: () => DataT | Ref<DataT> | null
-  transform?: (input: DataT) => DataT | Promise<DataT>
-  pick?: string[]
-  watch?: WatchSource[]
-  getCachedData?: (key: string, nuxtApp: NuxtApp) => DataT
-}
-
-type AsyncData<DataT, ErrorT> = {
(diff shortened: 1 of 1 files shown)


cernymatej:
I think it is helpful to see the types directly in the docs, rather than having to go to GitHub 🙏

docs/3.api/2.composables/use-async-data.md
```diff
@@ -105,44 +105,7 @@ If you have not fetched data on the server (for example, with `server: false`),
...
-};
-
-interface AsyncDataExecuteOptions {
-  dedupe?: 'cancel' | 'defer'
-}
-
-type AsyncDataRequestStatus = 'idle' | 'pending' | 'success' | 'error'
-```
+::note
+[view AsyncData types on GitHub](https://github.com/nuxt/nuxt/blob/32d9e2fe740d6d8c41e729db57aa46ebad387710/packages/nuxt/src/app/composables/asyncData.ts)
```
huang-julien, line 109: This is targeting to a specific commit

huang-julien (changes requested):

docs/3.api/2.composables/use-async-data.md
RokeAlvo: yes, you are right
fixed

RokeAlvo:
> I think it is helpful to see the types directly in the docs, rather than having to go to GitHub 🙏

then we just need to close the issue

danielroe:
I'd be up for improving this or using `twoslash` to render the types, but I think we should not remove them entirely.

PR #30355, opened 2024-12-24
fix(nuxt): remove div on createClientPage

### LINKED ISSUE #30344: Unwanted `div` created for client only pages using `NuxtLayout` + `NuxtPage`
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.14.1592
- CLI Version:      3.17.1
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

### Describe the bug

### Additional context

Image

Image

huang-julien:
### 🔗 Linked issue
#30344 

### 📚 Description
This pr removes the div on clientonly page.
Not sure why this was here

### DIFF
--- a/packages/nuxt/src/components/runtime/client-component.ts
+++ b/packages/nuxt/src/components/runtime/client-component.ts
@@ -18,11 +18,9 @@ export const createClientPage = (loader: AsyncComponentLoader) => {
     setup (_, { attrs }) {
       const nuxtApp = useNuxtApp()
       if (import.meta.server || nuxtApp.isHydrating) {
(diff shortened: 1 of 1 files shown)


logotip4ik:
If I'm not mistaken, it was added because when wrapped in `Transition`, vue was throwing errors like:

> <Transition> renders non-element root node that cannot be animated.

huang-julien:
Yes I remember now thanks !

PR #30359, opened 2024-12-24
chore(nuxt): document div wrapper in client-only page

huang-julien:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/30355


danielroe (approved):
good! I always forget these things otherwise 

might be a good idea to add some regression tests for these sorts of issues too

logotip4ik:
Maybe we also can add `display: contents` to div wrapping the page, so it is transparent to css styles ?

PR #30360, opened 2024-12-24
fix(nuxt): use withAsyncContext from Vue on the client side

### LINKED ISSUE #30356: `asyncContext` does not see the new route after navigation
### Environment

````
- Operating System: Linux (Ubuntu 24.04 LTS)
- Node Version:     v22
- Nuxt Version:     3.14.159
- CLI Version:      3.15.0
- Nitro Version:    -
- Package Manager:  npm@9.5.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
````

### Reproduction

https://stackblitz.com/~/github.com/martinszeltins/async-context-test

### Describe the bug

I noticed that if I enable `experimental.asyncContext` and I navigate to a new route, then `useRoute()` still sees the old route. It seems to happen only when using layers (EDIT: As @cernymatej  pointed out in https://github.com/nuxt/nuxt/issues/30356#issuecomment-2561060219 it is reproducable also without layers). I have a simple app with a page and there is a component inside the page. This component uses a composable and the composable uses `useRoute()` to get the current route. When I navigate to a new route, the composable still sees the old route.

But it is also interesting that this happens only on client-side navigation. If I refresh the page, then the composable sees the correct route.

Please see the reproduction and I also made a screencast to make it easier to see the issue. As you can see, when I enter the `/offices/123` route, the `useRoute()` does not see the new route.

**Screencast**:
https://github.com/user-attachments/assets/9e82e0c5-24c6-4611-95e3-2311d628ef87

### Additional context

_No response_

### Logs

```shell-script

```

cernymatej:
### 🔗 Linked issue
Fix #30356

### 📚 Description

This PR fixes a bug with the Nuxt implementation of `withAsyncContext`, where the component scope was being lost on the client side.

### DIFF
--- a/packages/nuxt/src/app/composables/asyncContext.ts
+++ b/packages/nuxt/src/app/composables/asyncContext.ts
@@ -3,10 +3,6 @@ import { getCurrentInstance, withAsyncContext as withVueAsyncContext } from 'vue
 
 /** @since 3.8.0 */
 export function withAsyncContext (fn: () => PromiseLike<unknown>) {
-  if (import.meta.client) {
-    return withVueAsyncContext(fn)
-  }
-
   return withVueAsyncContext(() => {
     const nuxtApp = getCurrentInstance()?.appContext.app.$nuxt
     return nuxtApp ? nuxtApp.runWithContext(fn) : fn()
--- a/packages/nuxt/src/app/composables/asyncContext.ts
+++ b/packages/nuxt/src/app/composables/asyncContext.ts
@@ -3,6 +3,10 @@ import { getCurrentInstance, withAsyncContext as withVueAsyncContext } from 'vue
 
 /** @since 3.8.0 */
 export function withAsyncContext (fn: () => PromiseLike<unknown>) {
+  if (import.meta.client) {
+    return withVueAsyncContext(fn)
+  }
+
   return withVueAsyncContext(() => {
     const nuxtApp = getCurrentInstance()?.appContext.app.$nuxt
     return nuxtApp ? nuxtApp.runWithContext(fn) : fn()
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -499,12 +499,6 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp |
   } else {
     // In client side we could assume nuxt app is singleton
     nuxtAppCtx.set(nuxt as NuxtApp)
-
-    // assume that the Nuxt context is available in components
-    // (enables injecting data for components in the callback)
-    if (getCurrentInstance()) {
-      return fn()
-    }
     return nuxt.vueApp.runWithContext(fn)
   }
 }
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -499,6 +499,12 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp |
   } else {
     // In client side we could assume nuxt app is singleton
     nuxtAppCtx.set(nuxt as NuxtApp)
+
+    // assume that the Nuxt context is available in components
+    // (enables injecting data for components in the callback)
+    if (getCurrentInstance()) {
+      return fn()
+    }
     return nuxt.vueApp.runWithContext(fn)
   }
 }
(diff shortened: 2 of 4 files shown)


danielroe:
It's a very tricky thing. Basically while the new page is loading, there are two suspense forks. We control which route object is injected into each of them as long as someone is using `$route` in template, or `useRoute()` (from nuxt) in setup. Outside a page context, the route object is always updated _after_ the route transition or you get rerendering before (from the user's perspective) the route has actually finished changing. This is particularly relevant for layouts, for example.

cernymatej:
Yeah, I've just realized that.
It would make sense to always provide the current route to the `withAsyncContext` branch, wouldn't it? Perhaps we could inject it in `callWithNuxt` ?

cernymatej:
Basic Idea:
- Provide a symbol at the root (`RouterView` ?) to determine whether we are within its context
- If the symbol is unavailable, assume that `useRoute` is not being called in pages / layouts and instead use the reactive value, which we could maybe provide globally through the `vueApp` ?

danielroe:
that's basically what we currently do

(minus the specific withAsyncContext handling)

the key thing is that the route available outside is 'pre' the navigation

cernymatej:
Ah, I see. I thought, that layouts were in the `RouterView` context, but not in the `RouteProvider`. (That's how I wanted to keep the pre-navigation route in layouts and the reactive one elsewhere)

What would you recommend to differentiate the `useRoute` call in `withAsyncContext` then? 🙏

We can provide the reactive route (not the synced one) in `withAsyncContext` without causing the rerendering, can't we?

martinszeltins:
> It's a very tricky thing. Basically while the new page is loading, there are two suspense forks. We control which route object is injected into each of them as long as someone is using `$route` in template, or `useRoute()` (from nuxt) in setup. Outside a page context, the route object is always updated _after_ the route transition or you get rerendering before (from the user's perspective) the route has actually finished changing. This is particularly relevant for layouts, for example.

But this is happening only if `experimental.asyncContext` is turned on. I'm curious why it works without `asyncContext` but not with it?

danielroe:
Turning on asyncContext _should_ not affect client side behaviour at all - perhaps that's something to investigate?

cernymatej:
> But this is happening only if `experimental.asyncContext` is turned on. I'm curious why it works without `asyncContext` but not with it?

It's because when you're using `experimental.asyncContext`, the `withAsyncContext` function from Vue gets replaced with the implementation from Nuxt, which, if I understand it correctly, uses the global vueApp context istead of the one in the component

It calls:
```ts
runWithContext <T>(fn: () => T) {
  if (nuxtApp._scope.active && !getCurrentScope()) {
    return nuxtApp._scope.run(() => callWithNuxt(nuxtApp, fn))
  }
  return callWithNuxt(nuxtApp, fn)
},
```

which calls
```ts
callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
  const fn: () => ReturnType<T> = () => args ? setup(...args as Parameters<T>) : setup()
  const nuxtAppCtx = getNuxtAppCtx(nuxt._id)
  if (import.meta.server) {
    return nuxt.vueApp.runWithContext(() => nuxtAppCtx.callAsync(nuxt as NuxtApp, fn))
  } else {
    // In client side we could assume nuxt app is singleton
    nuxtAppCtx.set(nuxt as NuxtApp)
    return nuxt.vueApp.runWithContext(fn) // <--- Isn't this the problem?
  }
}
```

Why couldn't we assume that a component scope (checking with `getCurrentInstance()` is also a Nuxt scope?

martinszeltins:
> Turning on asyncContext _should_ not affect client side behaviour at all - perhaps that's something to investigate?

@danielroe That is exactly what [this issue](https://github.com/nuxt/nuxt/issues/30356) is reporting. Without asyncContext it works perfectly fine. But as soon as asyncContext is turned on then it does not work anymore.

danielroe:
(I think we should probably only add that build plugin for server mode....)

cernymatej:
```ts
export function withAsyncContext (fn: () => PromiseLike<unknown>) {
  if (import.meta.client) {
    return withVueAsyncContext(fn)
  }

  return withVueAsyncContext(() => {
    const nuxtApp = getCurrentInstance()?.appContext.app.$nuxt
    return nuxtApp ? nuxtApp.runWithContext(fn) : fn()
  })
}
```
Would this do the trick?

danielroe (approved):
thank you! ❤️

daniluk4000:
Fun fact: memory usage dropped from 95mb to 65mb. Will also test on my main work later. Great catch!

PR #30380, opened 2024-12-26
chore: add configuration for JetBrains IDEs

cernymatej:
### 📚 Description
This PR turns on Eslint on by default in JetBrains IDEs and excludes some of the generated directories from being indexed.


danielroe (approved):
❤️

PR #30381, opened 2024-12-26
fix(nuxt): ignore non-reference identifiers when extracting pageMeta

### LINKED ISSUE #30378: Pre-transform error when pinia store used in inline middleware
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.0
- CLI Version:      3.17.2
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      default
- Runtime Modules:  @pinia/nuxt@0.9.0
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-zvdjtwqu

1. Run project in dev mode

### Describe the bug

After upgrading to 3.15, we started getting errors when running the project in dev mode on pages where pinia is used in inline middleware.

### Additional context

This is not reproduced in version 3.14.1592

### Logs

```shell-script
ERROR Pre-transform error: Maximum call stack size exceeded
Plugin: nuxt:pages-macros-transform
File: /home/projects/nuxt-starter-zvdjtwqu/pages/index.vue?macro=true
```

cernymatej:
### 🔗 Linked issue
Fix #30378

### 📚 Description
This fixes an issue where we didn't check if an identifier was a reference one or not while extracting pageMeta.
This could lead to an infinite recursion loop when encountering a member expression, for example:
```ts
const bar = { bar: 'bar' }.bar  // <- this `bar` property access
                                // would trigger `processDeclaration()` for `bar` again

definePageMeta({
  middleware: [
    () => {
      const a = bar
    },
  ],
})
```

I added a regression test too!

### DIFF
--- a/packages/nuxt/src/core/utils/parse.ts
+++ b/packages/nuxt/src/core/utils/parse.ts
@@ -476,7 +476,7 @@ function getPatternIdentifiers (pattern: WithLocations<Node>) {
   return identifiers
 }
 
-function isNotReferencePosition (node: WithLocations<Node>, parent: WithLocations<Node> | null) {
+export function isNotReferencePosition (node: WithLocations<Node>, parent: WithLocations<Node> | null) {
   if (!parent || node.type !== 'Identifier') { return false }
 
   switch (parent.type) {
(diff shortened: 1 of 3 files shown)


danielroe (approved):
❤️

PR #30390, opened 2024-12-27
chore: only allow pnpm

iiio2:
I think may be we may add `preinstall` because here we have pnpm-lock.yaml that means we should install deps with pnpm only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "license": "MIT",
   "type": "module",
   "scripts": {
+    "preinstall": "npx only-allow pnpm",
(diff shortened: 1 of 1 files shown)


danielroe:
I don't think we need to add this as we already have packageManager set, and we tell people to run corepack enable before development

pi0:
i think corepack already gives enough guards. it can break ability to test runtimes such as bun/deno even if without lockfile they are compatibile.

iiio2:
Oh, Thanks a lot @danielroe and @pi0 . I feel so lucky that you both comment on my PR 😄

PR #30396, opened 2024-12-27
fix(nuxt): warn about ignored char while parsing route segment

### LINKED ISSUE #30392: [zh:translated] Dynamic Routes: dynamic params not kebabcase
### Environment

- Operating System: Darwin
- Node Version:     v22.11.0
- Nuxt Version:     3.15.0
- CLI Version:      3.17.2
- Nitro Version:    2.10.4
- Package Manager:  pnpm@9.12.2
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://github.com/Lionad-Morotar/reproduce-nuxt-dynamic-route-kebabcase

### Describe the bug

我创建了一个`[my-id].vue`文件，我以为能通过 `params['my-id']` 取得这个参数，但不行。在路由参数中只能通过`params.myid`去获取。

Image

不确定这是一个缺陷或者是功能，也许应该在文档标注一下？

### Additional context

_No response_

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

fix #30392 

### 📚 Description

This could go into a patch version IMO.

This PR add a warnign (not sure why this was removed) when a a character is ignored when building route path


danielroe (commented):
I think we might need to tweak the message to make more sense - maybe log a warning after the segment is totally extracted, explaining what the extracted param is vs what the user might have expected?

huang-julien:
I see, maybe should we add a documentation link to avoid having a long warning ?


 I may have another idea in addition to this.
How about storing all build warnings somewhere and let the devtools display them, show the stack trace and why this happened ?
IMO, the devtools is way more UX friendly than the old CLI

danielroe (approved):

danielroe:
This was resolved in https://github.com/nuxt/nuxt/pull/30619 and will be in the next patch release.

PR #30407, opened 2024-12-28
perf(vite): remove client entry from server warmup

userquin:
### 📚 Description

Starting dev server with client entry added to the vite server warmup is too slow (10x): this PR removes the server warmup in the vite client module included in this PR https://github.com/nuxt/nuxt/pull/27963

This should be ported to 3.* version (3.15.1).

You can start the playground in nuxt repo: on i18n nuxt module playground, it takes 20sc or even more on my laptop.

image
_Nuxt playground with this PR_

image
_Nuxt playground without this PR_

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -152,9 +152,6 @@ export async function buildClient (ctx: ViteBuildContext) {
     ],
     appType: 'custom',
     server: {
-      warmup: {
-        clientFiles: [ctx.entry],
-      },
       middlewareMode: true,
     },
   } satisfies vite.InlineConfig, ctx.nuxt.options.vite.$client || {}))


danielroe:
this will certainly speed up the dev server start

how does it affect the first load speed in the browser?

userquin:
Using nuxt i18n module ~5sc  (pnpm nuxt dev playground --open) vs 20sc (3.15.0):

image

In the playground here ~2sc vs 12sc:

image

BobbieGoede:
We're not sure why the playground in the i18n module is so slow after updating to `3.15.0`, for me `build:done` takes 40+ seconds. With the changes in this PR it's down to 700ms, I haven't been able to reproduce the slowdown in a clean nuxt starter project though..

danielroe:
@BobbieGoede Is that reproducible if you pin vite to v5 with Nuxt v3.15?

BobbieGoede:
@danielroe 
Yep that seems to the trick as well (without the changes in this patch), takes the `build:done` duration back to 922ms.

danielroe:
cc: @antfu

antfu:
For context, you were talking about this playground (https://github.com/nuxt-modules/i18n/tree/main/playground) that is experiencing a performance downgrade, right? I am looking into it

antfu:
I have managed to reproduce that. It indeed seems like a Vite issue. I did a CPU profile, here are the profile files in case anyone is interested in investigating as well:

- [3.14.cpuprofile](https://github.com/user-attachments/files/18280036/3.14.cpuprofile)
- [3.15.cpuprofile](https://github.com/user-attachments/files/18280037/3.15.cpuprofile)

I'll keep tracing down

The Vite part of computation:

CleanShot 2024-12-31 at 15 42 52@2x

antfu:
Upstream PR to fix this:
- https://github.com/vitejs/vite/pull/19107

PR #30408, opened 2024-12-29
refactor(nuxt): use consola with nuxt tag instead of console

huang-julien:
### 📚 Description

this PR refactors console to consola


danielroe (approved):

danielroe:
i also think we could use consola tags and maybe some nice colours for some of our logs

danielroe (approved):

PR #30417, opened 2024-12-30
fix(nuxt): added a class on wrapper div on createClientPage

### LINKED ISSUE #30344: Unwanted `div` created for client only pages using `NuxtLayout` + `NuxtPage`

JJPSeno:
### 🔗 Linked issue

Resolves #30344

### 📚 Description

There is a limitation to client-only pages where they generate a classless and id-less div which is tricky to target with CSS. 
I added a 'client-page-wrapper' class to the generated wrapper to more easily target this div with CSS.

I've checked the effects via running the playground with the corresponding issue's reproduction steps:
nuxt-client-page-wrapper

### DIFF
--- a/packages/nuxt/src/components/runtime/client-component.ts
+++ b/packages/nuxt/src/components/runtime/client-component.ts
@@ -20,7 +20,7 @@ export const createClientPage = (loader: AsyncComponentLoader) => {
       if (import.meta.server || nuxtApp.isHydrating) {
         // wrapped with div to avoid Transition issues
         // @see https://github.com/nuxt/nuxt/pull/25037#issuecomment-1877423894
-        return () => h('div', [
+        return () => h('div', { class: 'client-page-wrapper' }, [
           h(ClientOnly, undefined, {
             default: () => h(page, attrs),
           }),


cernymatej:
I think we should probably make it configurable and allow specifying other attributes as well. (similar to the bodyAttrs in useHead)

huang-julien:
I'm currently exploring another solution which could remove the `div` wrapper around client pages 🤔

JJPSeno:
Hi, I was able to add the customizability by adding this to the app.ts file and importing it to the client-component file
image

Despite the fact that it seems to be working as intended, I am getting this editor error when I do add my custom value to the nuxt config file:
image

```
Object literal may only specify known properties, and 'clientPageAttrs' does not exist in type '{ baseURL?: string | undefined; buildAssetsDir?: string | undefined; cdnURL?: string | undefined; head?: { title?: string | { textContent?: string | false | null | undefined; tagPriority?: number | ... 7 more ... | undefined; } | null | undefined; ... 11 more ...; viewport?: string | undefined; } | undefined; ... 11...'.ts(2353)
config.d.ts(121, 2): The expected type comes from property 'app' which is declared here on type 'InputConfig<NuxtConfig, ConfigLayerMeta>'
(property) clientPageAttrs: {
    class: string;
}
```

image
I'm not really sure how to squash this error. Maybe I missed a type declaration file I should have added the new option to?

huang-julien:
That's tricky, i'm not sure what do we want here.
- client page wrapper attributes for all client page at the app level ?
- client page wrapper attributes defined in at the page level (maybe with definePageMeta ? )

cernymatej:
Do we need to keep the div, though? I though that this resolves it: https://github.com/nuxt/nuxt/pull/30425

huang-julien:
I'll let Daniel decide about that. If #30425 should go in 4.x only or also in 3.x . If it goes into 3.x then yes this issue and PR will be closed

JJPSeno:
> Do we need to keep the div, though? I though that this resolves it: #30425

It was mentioned as a possible workaround https://github.com/nuxt/nuxt/issues/30344#issuecomment-2564855758
But yeah I am also not sure if their decision will be to remove the div or if we can go through with just adding some control on it.

> That's tricky, i'm not sure what do we want here.
client page wrapper attributes for all client page at the app level ?
client page wrapper attributes defined in at the page level (maybe with definePageMeta ? )

Oh, I'd like to know some thoughts on this as well.

danielroe:
let's go ahead with removing the `div` entirely - I think that's more predicable, and I'll aim to land it in v3.16.

thank you for raising this PR however - it would also have been a good solution ❤️

PR #30425, opened 2024-12-30
fix(nuxt): remove `div` wrapper in client-only pages

### LINKED ISSUE #30344: Unwanted `div` created for client only pages using `NuxtLayout` + `NuxtPage`

huang-julien:
### 🔗 Linked issue

close #30417
close #30344 

### 📚 Description

The reason we needed the `div` wrapper was because we were passing slots to `ClientOnly`. And slots in Vue are `Fragments` which `Transition` can't apply animations even if there's only one element.

This PR removes the div wrapper on client-only pages by cloning the component and making it client-only. This is almost identical to `createClientOnly()`

The reason `createClientOnly` isn't used is that it reads the current el html and returns a StaticVNode which `<Transition>` can't modify. So the remaining solution is to create a simplified `createClientOnly` that is suitable for page components.


huang-julien:
BTW i'm planning to do some light refactors in client-only components to slightly reduce the amount of code.

danielroe (approved):
This looks good to me as a bug fix, though we should probably wait until v3.16 rather than ship it in a patch.

Ideally it'd be nice to have a navigation + hydration test to confirm things work with ssr + then subsequent page transitions.

PR #30447, opened 2025-01-03
fix(nuxt): use node `location` instead of `range` for route meta property extraction

### LINKED ISSUE #30445: [3.15.0]  ERROR  [unhandledRejection] Cannot read properties of undefined (reading '0')
### Environment

After npx nuxi@latest upgrade --force, receiving the following error.

Works fine on "nuxt": "3.14.1592"

```
 ERROR  [unhandledRejection] Cannot read properties of undefined (reading '0')                                                                                                                                                                                                                                                                                                                                      10:55:13 AM

    at Object.<anonymous> (node_modules/nuxt/dist/index.mjs:711:65)
    at Object.enter (node_modules/nuxt/dist/index.mjs:75:23)
    at SyncWalker.visit (node_modules/estree-walker/src/sync.js:65:16)
    at SyncWalker.visit (node_modules/estree-walker/src/sync.js:100:19)
    at walk (node_modules/estree-walker/src/index.js:20:18)
    at walk (node_modules/nuxt/dist/index.mjs:72:10)
    at parseAndWalk (node_modules/nuxt/dist/index.mjs:85:3)
    at getRouteMeta (node_modules/nuxt/dist/index.mjs:693:5)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async augmentPages (node_modules/nuxt/dist/index.mjs:639:25)

ℹ Restarting Nuxt due to error: TypeError: Cannot read properties of undefined (reading '0')    
```

### Reproduction

It is a large project and reproduction is very hard to provide.

### Describe the bug

After 3.15.0 update the dev server is not starting. Also the project is not building.

### Additional context

_No response_

### Logs

```shell-script

```

cernymatej:
### 🔗 Linked issue
fix #30445

### 📚 Description
This PR fixes an issue introduced in #30088, where the `range` property of the ESTree node was used to get the character locations, but `ranges: true` was no longer present in the `parse` options in the `parseAndWalk` function, with which the original `parse` function had been replaced.

The `start` and `end` properties should represent the same values, quote:
> ranges: Nodes have their start and end characters offsets recorded in start and end properties (directly on the node, rather than the loc object, which holds line/column data. To also add a [semi-standardized](https://bugzilla.mozilla.org/show_bug.cgi?id=745678) range property **holding a [start, end] array with the same numbers**, set the ranges optionto true.
>
> _source: https://github.com/acornjs/acorn/blob/master/acorn/README.md_


cernymatej:
Should we consider changing `withLocations` to perform a runtime check instead of just type assert?

danielroe (approved):

PR #30456, opened 2025-01-04
perf(nuxt): drop `pathe` browser dep for deep server components

danielroe:
### 📚 Description

noticed we were using `join` from `pathe` in a runtime context - we probably want `joinURL` instead (unless there's a reason I'm unaware of)


huang-julien (approved):
We can drop it, I probably did that out of habit of using Pathé 😅

PR #30479, opened 2025-01-06
docs: add context for useAsyncData side effects

### LINKED ISSUE #30149: useAsyncData does not work with pinia storage async actions in generated project
### Environment

- Node Version:     v20.18.1
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Package Manager:  pnpm@9.14.4
- Builder:          -
- User Config:      default
- Runtime Modules:  @nuxtjs/i18n@9.1.0, @pinia/nuxt@0.8.0, @sidebase/nuxt-auth@0.9.4, @vueuse/nuxt@12.0.0, pinia-plugin-persistedstate/nuxt, nuxt-swiper@1.1.1, nuxt-icons@3.2.1, nuxt-proxy@0.4.1, vue-yandex-maps/nuxt
- Build Modules:    -

### Reproduction

Commercial project, I can't show the project code, but I will describe my problem in detail with a minimal example:

[git@github.com:wokalek/issue-nuxt-prerender-asyncdata-store-problem.git](https://github.com/wokalek/issue-nuxt-prerender-asyncdata-store-problem)

#### ! The project is completely pre-generated

When navigating to a page from other page (after all dependencies are downloaded), the data from the payload does not display the page in the same way as with a direct jump to page.

Template like this:

```vue
<template>
  <div>
    <pre>{{ item.tags }}</pre>
            <!-- ↑ here it will crash with an error when going to the page indirectly  -->
  </div>
</template>

<script setup lang="ts">
const catalogStore = useCatalogStore()

const item = computed(() => catalogStore.item)

await useAsyncData(() => catalogStore.getProduct(route.params.slug))
</script>
```

Image

At the same time, the page payload is actually loaded at the address, I see this in devtools, for example:

```
http://localhost:3000/product/bazalit-venti-n/_payload.json?e31404b9-477f-4833-a64b-8683b8d219fe
```

So it feels like

```ts
await useAsyncData(() => catalogStore.getProduct(route.params.slug))
```

isn't running asynchronously for payload and isn't waiting for its payload to load before it starts rendering the page, which I expect.

Is this normal behavior for Nuxt when generating? Or am I missing something?

### Additional context

_No response_

### Logs

```shell-script

```

Kamsou:
# Hi, it's my first contribution to Nuxt 🙃

### 🔗 Linked issue

resolves #30149

### 📚 Description

The issue arises because `useAsyncData` is designed for data fetching, not for triggering side effects such as calling Pinia actions. Using it in this way can lead to multiple problems:

- Unintended Side Effects: Pinia actions can modify the store or trigger additional behaviors, which goes beyond the intended use of useAsyncData.

- Re-execution on Nullish Results: When a Pinia action returns undefined or a nullish value, useAsyncData may re-execute, causing repeated calls to the same action and potentially introducing unintended behavior.

- Separation of Concerns: Calling Pinia actions directly within useAsyncData blurs the line between data fetching and store management, making the code harder to understand and maintain.

### ❓ One question 

I hesitated to put an alternative. I think it might overload the doc, what do you think?

### 👁 Overview

Capture d’écran 2025-01-06 à 19 17 23

Thanks 💚


cernymatej:
wouldn't it be better to put it somewhere here? maybe also wrapped in the `::warning` component?
https://github.com/nuxt/nuxt/blob/728b7e1c25f0b4641014bcb51095a4b3ac412adf/docs/3.api/2.composables/use-async-data.md?plain=1#L15-L17

Kamsou:
> wouldn't it be better to put it somewhere here? maybe also wrapped in the `::warning` component?
> 
> https://github.com/nuxt/nuxt/blob/728b7e1c25f0b4641014bcb51095a4b3ac412adf/docs/3.api/2.composables/use-async-data.md?plain=1#L15-L17

Yes, I admit I hesitated. Would you put the warning without the example then @cernymatej? 

With the addition of the example code, I thought it would fit in well with the narrative of the “Data Fetching” page.

cernymatej:
Maybe we can keep it in both places?

The warning on the `useAsyncData` page could perhaps be a little more generic - only mentioning that the fetch function should be side-effect free. 🤔

Kamsou:
> Maybe we can keep it in both places?
> 
> The warning on the `useAsyncData` page could perhaps be a little more generic - only mentioning that the fetch function should be side-effect free. 🤔

Let's go.  I suggest putting it in the Params section, under handler, because that's where the fetch function is defined. What do you think?

Capture d’écran 2025-01-06 à 21 13 01

danielroe (approved):
This looks great! I added a link to `callOnce` docs. How does it look to you?

Kamsou:
> This looks great! I added a link to `callOnce` docs. How does it look to you?

Perfect ! Thanks @danielroe 🫰

PR #30480, opened 2025-01-06
fix(nuxt): omit the body from useFetch for GET requests

### LINKED ISSUE #20667: useFetch GET method with body is not working
### Environment

- Operating System: Darwin
- Node Version:     v18.15.0
- Nuxt Version:     3.4.2
- Nitro Version:    2.3.3
- Package Manager:  pnpm@8.1.1
- Builder:          vite
- User Config:      app, components, typescript, modules, extends, css, runtimeConfig, imports, pinia, linkChecker, vite
- Runtime Modules:  @pinia/nuxt@0.4.9, @element-plus/nuxt@1.0.4, @nuxtjs/tailwindcss@6.6.6, @nuxtjs/eslint-module@4.0.2, @pinia-plugin-persistedstate/nuxt@1.1.1, nuxt-lodash@2.4.1, nuxt-icons@3.2.1
- Build Modules:    -

### Reproduction

```
await useFetch<WXData>('/signature', {
  method: 'GET',
  baseURL: window.location.origin,
  body: {
    path: window.location.pathname,
  },
})
  .then((res) => {
    console.log('res', res)
  })
  .catch((err) => {
    console.log('err', err)
  })
  ```

### Describe the bug

If I set 'body', it does not appear in the network tab.  
If I remove 'body', it will appear in the network tab.  
I know using the GET method here is not the best practice, but the backend is reluctant to change.    
It's working on postman.

### Additional context

_No response_

### Logs

_No response_

### PR DESCRIPTION
**Pull request by Dobefu**
### 🔗 Linked issue

Resolves #20667

### 📚 Description

This PR will ever so slightly improve DX by extending the useFetch type to omit the body when the method is "GET".

### 📝 Code example

The code below is what I've used to test the functionality. I've written the code in `./playground/server/api/test.ts`.

```typescript
import { useFetch } from '../../../packages/nuxt/src/app/composables/fetch'

export default eventHandler(async (_event) => {
  const { data: it } = await useFetch('https://jsonplaceholder.typicode.com/todos/1', {
    method: 'GET',
    body: { test: 1 }, // This should create a linter error.
  })

  const { data: just } = await useFetch('https://jsonplaceholder.typicode.com/todos/1', {
    method: 'HEAD',
    body: { test: 1 }, // This should create a linter error.
  })

  const { data: works } = await useFetch('https://jsonplaceholder.typicode.com/todos/1', {
    method: 'POST',
    body: { test: 1 }, // This is fine!
  })

  return { it, just, works }
})
```

Of course, please reach out to me if there's any questions or issues!


Dobefu:
Ah yes, the cute little rabbit is right. I should probably add the HEAD method as well, for completeness' sake

Dobefu:
@coderabbitai review

HigherOrderLogic:
I dont think this should be the default. There's probably still some APIs around that doesnt follow the HTTP spec.

Dobefu:
True, but a type can easily be added by someone to override this. It will not prevent the code from running, but is simply a type of linting error.

Additionally, I personally think that the same logic applies to many other types. But it will, in my opionion, harm DX to make types too broad to accomodate for projects and APIs that do not follow the official specification.

@danielroe What do you think?

danielroe:
Thank you! ❤️  This is a really nice idea for improving DX.

I think if we adopt it should probably be upstream in [`ofetch`](https://github.com/unjs/ofetch), and then inherit/replicate in Nuxt, but I would observe that it might also pose an issue when method is a string, and not strongly typed as a constant. (As I _think_ I tried to do this very thing previously but encountered issues.) Might be worth adding a test in `basic/types.ts` to ensure that kind of behaviour still works.

Dobefu:
Hi @danielroe!

Thank you for your feedback, and your kind words! 

Adding tests is definitely a good idea! I've not realized that types could be unit tested dynamically like that. I'll look into adding some tests to it.

As for upstreaming it in `ofetch`, that's also a good shout. I'm just not entirely certain what you mean by issues when method is not strongly typed as a constant. The `method` right now - though not strongly typed as a constant - has a string type with specific allowed values. Could you please explain what issues you're expecting that I did not think of?

danielroe:
let's resolve upstream in `ofetch` first and then port back to Nuxt when that's done.

linking https://github.com/nitrojs/nitro/issues/2758 to track.

PR #30500, opened 2025-01-08
fix(nuxt): use `fullPath` instead of empty string in router hmr

### LINKED ISSUE #30469: The route parameters are lost during HMR in Nuxt 3.15
### Environment

- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.15.1
- CLI Version:      3.17.2
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/devbox/condescending-star-cd6z2n

### Describe the bug

After updating to Nuxt 3.15, in the development environment, if the route contains parameters, when I make changes and save on the current page, the page undergoes a hot reload and causes the route parameters to be lost. I did not encounter this issue in previous versions of Nuxt. Can you help me resolve this problem?

### Additional context

_No response_

### Logs

```shell-script

```

vmrjnvc:
### 🔗 Linked issue

resolves #30469

### 📚 Description

When HMR occurs, it uses router.replace("") with an empty string as the argument. In my tests with Vue, this method removes query parameters from the URL. To address this, I tried using fullPath, and it seems to work.

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -642,7 +642,7 @@ if (import.meta.hot) {
       for (const route of routes) {
         router.addRoute(route)
       }
-      router.replace('')
+      router.replace(router.currentRoute.value.fullPath)
     }
     if (routes && 'then' in routes) {
       routes.then(addRoutes)


danielroe (approved):

danielroe:
thank you!

PR #30510, opened 2025-01-09
feat(kit,nuxt): allow multiple nuxts to run in one process

### LINKED ISSUE #537: Context conflict when running multiple projects with nuxt-vitest
I have a monorepo with a nuxt 3 app and multiple nuxt modules. The nuxt 3 app already has nuxt-vitest tests and it works great with the new RootDir option. But when I try to add nuxt-vitest to a module in the monorepo, it fails with this error:
```
> vitest --ui --coverage

failed to load config from /home/jordy/repositories/my_repo/packages/my_package/vitest.config.ts
Error: Context conflict
 ❯ checkConflict node_modules/unctx/dist/index.mjs:6:13
 ❯ Object.set node_modules/unctx/dist/index.mjs:40:9
 ❯ initNuxt node_modules/nuxt/dist/index.mjs:2689:11
 ❯ Object.ready node_modules/nuxt/dist/index.mjs:2680:18
 ❯ loadNuxt node_modules/nuxt/dist/index.mjs:2989:16
 ❯ loadNuxt node_modules/@nuxt/kit/dist/index.mjs:534:19
 ❯ startNuxtAndGetViteConfig node_modules/nuxt-vitest/dist/config.mjs:8:16
 ❯ getVitestConfigFromNuxt node_modules/nuxt-vitest/dist/config.mjs:48:15
```

On their own it works, it's only when running the global vitest workspace, which has all the coverage and UI in one place.

Not sure how to fix this.

### LINKED ISSUE #664: Vitest workspaces support
### Describe the feature

When using vitest in a monorepo setting, one is likely to use the vitest workspaces feature: https://vitest.dev/guide/workspace

However, in this setting one would use `defineProject` instead of `defineConfig`, to reuse the global vitest instance. 

It would be cool to be able to do this also for nuxt apps, in order to run nuxt tests seamlessly along the ones of other packages

### Additional information

### Final checks

### LINKED ISSUE #1034: Can't load config files in workspace since vitest 2.1.5
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.14.1592
- CLI Version:      3.16.0
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      default
- Runtime Modules:  ../../packages/package-one/src/module, ../../packages/package-two/src/module
- Build Modules:    -

### Reproduction

You get the error when using vitest 2.1.8 (the error appears since 2.1.5): https://stackblitz.com/edit/github-n797ja-hxhseq

It works in 2.1.4: https://stackblitz.com/edit/github-n797ja-2rtuvw

### Describe the bug

In a workspace with multiple projects, it is no longer possible to call defineVitestConfig because of unctx conflicts:

```
Error: Context conflict
    at checkConflict (file:///home/projects/github-n797ja-hxhseq/node_modules/unctx/dist/index.mjs:25:13)
    at set (file:///home/projects/github-n797ja-hxhseq/node_modules/unctx/dist/index.mjs:59:22)
    at initNuxt (file:///home/projects/github-n797ja-hxhseq/node_modules/nuxt/dist/index.mjs:4404:21)
    at ready (file:///home/projects/github-n797ja-hxhseq/node_modules/nuxt/dist/index.mjs:4306:26)
    at loadNuxt (file:///home/projects/github-n797ja-hxhseq/node_modules/nuxt/dist/index.mjs:4871:16)
```

Could be related to https://github.com/nuxt/test-utils/issues/537

### Additional context

_No response_

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

unblocked by https://github.com/nitrojs/nitro/pull/2983 :heart: 
fix https://github.com/nuxt/test-utils/issues/537
fix https://github.com/nuxt/test-utils/issues/1034
resolve https://github.com/nuxt/test-utils/issues/664
fix https://github.com/nuxt/nuxt/issues/22565

### 📚 Description

This PR allows to create multiple Nuxt instance within a single process.

We were previously setting a Nuxt Singleton with unctx. This means that we can only run a single Nuxt instance in a single process.

By using asyncLocalStorage, we'll be able to track an unique id for each `buildNuxt` and `initNuxt` -> allowing to run these in parallel and all `useNuxt()` within it. It also convert `nuxtCtx` to be a getter so it retrieves the correct namespace with `ASyncLocalStorage`

For backward compatibility, we'll keep a fallback global singleton of the first Nuxt instance that is being set. This behavior CAN  be removed in V5. We can also add a `.run` on the Nuxt build instance later.

next step: merge all nuxt test within a single job


huang-julien:
https://github.com/nuxt/nuxt/pull/30537 ecosystem ci for nuxt 3

packages/kit/src/context.ts
```diff
@@ -36,5 +45,10 @@ export function useNuxt (): Nuxt {
  * ```
  */
 export function tryUseNuxt (): Nuxt | null {
-  return nuxtCtx.tryUse()
+  const nuxt = nuxtCtx().tryUse()
+  if (!nuxt) {
+    logger.warn('Using fallback global Nuxt instance. You may be using a @nuxt/kit composable outside of a Nuxt context, this behavior is deprecated and will be removed in v4.')
```
huang-julien, line 50: I'm not sure how to rephrase this to make it sound a bit gentler 😅. Any suggestions?

cc @danielroe

huang-julien: this also isn't going to be remove d in V4, maybe in V5 🤔

packages/nuxt/src/core/nuxt.ts
```diff
@@ -22,6 +22,7 @@ import { gt, satisfies } from 'semver'
 import { hasTTY, isCI } from 'std-env'
 import { genImport } from 'knitwork'
 
+import { randomUUID } from 'uncrypto'
```
danielroe, line 25:
```suggestion
import { randomUUID } from 'node:crypto'
```

packages/kit/src/index.ts
```diff
@@ -18,7 +18,7 @@ export type { ExtendConfigOptions, ExtendViteConfigOptions, ExtendWebpackConfigO
 export { assertNuxtCompatibility, checkNuxtCompatibility, getNuxtVersion, hasNuxtCompatibility, isNuxtMajorVersion, normalizeSemanticVersion, isNuxt2, isNuxt3 } from './compatibility'
 export { addComponent, addComponentsDir } from './components'
 export type { AddComponentOptions } from './components'
-export { nuxtCtx, tryUseNuxt, useNuxt } from './context'
+export { getNuxtCtx as nuxtCtx, tryUseNuxt, useNuxt, globalNuxtCtx } from './context'
```
danielroe, line 21: we should probably not change the type of `nuxtCtx` to a function. Rather, we should output `globalNuxtCtx` as `nuxtCtx` so we're backwards compatible.

packages/nuxt/src/core/nuxt.ts
```diff
@@ -50,18 +51,19 @@ import { VirtualFSPlugin } from './plugins/virtual'
 
 export function createNuxt (options: NuxtOptions): Nuxt {
   const hooks = createHooks<NuxtHooks>()
```
danielroe, line 53: we probably need to ensure that hooks are called in a way that passes the name of the Nuxt concerned.... maybe via a custom caller, or by overriding `callHook`?

huang-julien: hooks is assigned to the Nuxt instance.
If users wants to know which nuxt is concerned, they probably knows it when calling `useNuxt` or `nuxt.hook` <-- means they have stored it in a variable somewhere

What are you thinking of :eyes: ?

danielroe: I mean we want this to work:

```ts
nuxt.hook('modules:done', () => {
  addTemplate(someTemplate)
})
```

huang-julien: How about 
```ts
  const nuxt: Nuxt = {
    _version: version,
    options,
    hooks,
    callHook: hooks.callHook,
    addHooks: hooks.addHooks,
    hook: (name, fn, opts) => asyncNameStorage.run(name, () => hooks.hook(name, fn, opts)),
    ready: () => asyncNameStorage.run(name, () => initNuxt(nuxt)),
    close: () => hooks.callHook('close', nuxt),
    vfs: {},
    apps: {},
    __name: name,
  }
```

this would only work in hooks tho.

huang-julien: I think we'd still need to add a `nuxt.run` in the future

danielroe: It's also possible to acess `nuxt.hooks` directly, and in fact we use `addHooks` to add user hooks, for example.

we can have `nuxt.run`, of course, but the value of most of kit utilities is that they shouldn't _require_ it by having to pass `nuxt` around. at least reducing the surface area of possible places where users can call it would help a lot.

huang-julien: > It's also possible to acess nuxt.hooks directly, and in fact we use addHooks to add user hooks, for example.

Indeed so do we need to override it directly in `hooks` ? 

> we can have nuxt.run, of course, but the value of most of kit utilities is that they shouldn't require it by having to pass nuxt around. at least reducing the surface area of possible places where users can call it would help a lot.

It's only for cases when users wants to run it outside of `loadNuxt` or `buildNuxt`, I don't really see a lot of cases 🤔

danielroe (approved):

packages/kit/src/context.ts
```diff
@@ -1,9 +1,20 @@
...
 
-/** Direct access to the Nuxt context - see https://github.com/unjs/unctx. */
+/**
+ * Direct access to the Nuxt global context - see https://github.com/unjs/unctx.
+ * @deprecated Use `getNuxtCtx` instead
+ */
 export const nuxtCtx = getContext<Nuxt>('nuxt')
 
+/** async local storage for the name of the current nuxt instance */
+const asyncNameStorage = new AsyncLocalStorage<string>()
```
pi0, line 13: We could still leverage unctx..

danielroe: cc: @huang-julien - did you try this?

huang-julien: I did try but got into a `Context is not available`. I think I know what i did wrong

packages/kit/src/context.ts
```diff
@@ -1,9 +1,22 @@
...
 export const nuxtCtx = getContext<Nuxt>('nuxt')
 
+/** async local storage for the name of the current nuxt instance */
+const asyncNameStorage = createContext({
+  asyncContext: true,
+  AsyncLocalStorage,
+})
+
+/** Direct access to the Nuxt context with asyncLocalStorage - see https://github.com/unjs/unctx. */
+export const getNuxtCtx = () => getContext<Nuxt>(asyncNameStorage.tryUse())
```
huang-julien, line 18: @pi0 is it okay to use `getContext` with an undefined value with `tryUse()`?

pi0: It seems you are accessing by name (why?)

nuxt context instances are themselves unique. I should be something like this in impl:

- `const nuxtCtx = useContext('nuxt', { .. })`
- `getNuxtCtx = () => nuxtCtx.use()
- `runWithNuxtContext = (nuxt, fn) => { nuxtCtx.call(nuxt, fn) }`

huang-julien: I think i did taht to allow users to use `getContext('nuxt_name')` outside of any instance, just in case (maybe the storybook module will need this).

Do you prefer to have the instance of nuxt being stored ?

huang-julien: right :+1: , let's go for the nuxt instance being stored. We can open another PR to add this feature

huang-julien:
FYI we need a patch release for nitro v2 and v3 with https://github.com/nitrojs/nitro/pull/2983 to allow running multiple nuxt builders

danielroe:
If everything looks good we can merge it into `main` branch and wait for nitro release before merging into `3.x`. Alternatively - is it broken with current nitro, or does running multiple nuxts just not yet work?

huang-julien:
This only breaks when running multiple Nuxt. Users will only have to update nitro to the next patch

Deeyuude:
Thanks for the amazing work :) Any docs or examples for this one ?

Deeyuude:
> I'm not sure what doc you're looking for ? 🤔 This is a fix which unblock some use cases. Like having 2 nuxt ^roject in a monorepo which break Vitest Extension on Vscode for example

Ow i see ! Sorry but i remember Daniel was talking about Multi-App Support i thought that is something without monorepos sorry again @huang-julien

huang-julien:
This PR is meant for build time.
Previously Nuxt was a singleton at build time, so running multiple `loadNuxt` would result in a context conflict.

PR #30555, opened 2025-01-13
feat(nuxt): support tracking changes to nuxt options by modules

antfu:
This PR adds the proxy to the `nuxt` instance for every module to track how each module mutates the `nuxt.options` object, when `debug: true`. 

With this, it allows DevTools to present the data like (https://github.com/nuxt/devtools/pull/770):

CleanShot 2025-01-13 at 15 41 14@2x

I am not sure if this deserved to be a separate option, or `debug` should accept an object to allow granular controls over what to debug. Bring this up first for discussion.


huang-julien:
Do you think we could also show the stacktrace ? that could be very usefull :eyes:

danielroe (commented):
I really like this as an idea.

I agree that we probably need a more granular control than just `debug`. For now we could ship it as an experimental option for testing: `experimental.debugModuleOptionMutations` or similar. (So setting both `debug` and that option would record.)

Note that this will not record any changes when `useNuxt()` is used within the module, which is a common pattern. Using async local storage might be the way to handle this.

antfu:
Yeah you are right. I did miss the `useNuxt()` part. I have updated the implementation to use AsyncLocalStorage, and a getter to `nuxt.options`, so we don't wrap `nuxt` but only `nuxt.options`. I see this approach captures more changes that were previously missing:

CleanShot 2025-01-14 at 14 03 35@2x

danielroe:
I really love this ❤️

atinux:
This is awesome ❤️

danielroe:
Updated so there's no behaviour change unless enabled. Note that enabling this feature delayed the basic test fixture startup by 1s, which was detected by codspeed:

|     | Benchmark | `BASE` | `HEAD` | Change |
| --- | --------- | ----------------------- | ------------------- | ------ |
| ❌ | `` initial dev server build in the basic test fixture `` | 6 s | 6.9 s | -12.72% |
| ❌ | `` loadNuxt in the basic test fixture `` | 2.5 s | 2.7 s | -10.14% |

danielroe (approved):

PR #30570, opened 2025-01-13
docs: adjust example and additional instructions of `useNuxtData`

Mini-ghost:
### 🔗 Linked issue

N/A

### 📚 Description

This PR adds additional details about `useNuxtData()`.

#### Updated Example Content  
```diff
- const { id } = useRoute().param
+ const route = useRoute()
```

While this implementation works in certain scenarios, it will not respond to changes in `param.id` when route reactivity is required. This might be unintuitive for beginners.

#### Updated Optimistic Updates Example  

- The original example used `useFetch()` for data updates, which is not a common use case, as `useFetch()` is primarily designed for data fetching scenarios.  
- The implementation of Optimistic Updates has been revised to fix a reference error in the original example and updated to provide examples compatible with the default settings of both `v3` and `v4`.  

  In `v3`, the original implementation worked. However, in `v4`, the default behavior of `data` in `asyncData` changes to `shallowRef()`. Unless users are aware of this difference and explicitly set `options.deep`, the original example will no longer function as expected.

  ```ts
  const { data: todos } = useNuxtData('todos')

  // Works in `v3`, but in `v4`, the UI may not reavtie to changes by default.
  todos.value.push(newTodo)

  // Works in both `v3` and `v4`.
  todos.value = [...todos.value, newTodo]
  ```

If there are any issues I may have overlooked, please let me know. Thank you!

### DIFF
--- a/docs/3.api/2.composables/use-nuxt-data.md
+++ b/docs/3.api/2.composables/use-nuxt-data.md
@@ -9,10 +9,26 @@ links:
 ---
 
 ::note
-`useNuxtData` gives you access to the current cached value of [`useAsyncData`](/docs/api/composables/use-async-data) , `useLazyAsyncData`, [`useFetch`](/docs/api/composables/use-fetch) and [`useLazyFetch`](/docs/api/composables/use-lazy-fetch) with explicitly provided key.
+`useNuxtData` gives you access to the current cached value of [`useAsyncData`](/docs/api/composables/use-async-data) , [`useLazyAsyncData`](/docs/api/composables/use-lazy-async-data), [`useFetch`](/docs/api/composables/use-fetch) and [`useLazyFetch`](/docs/api/composables/use-lazy-fetch) with explicitly provided key.
 ::
 
 ## Usage
+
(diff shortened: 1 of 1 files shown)


danielroe (approved):
❤️

docs/3.api/2.composables/use-nuxt-data.md
```diff
@@ -52,28 +72,34 @@ const { data } = await useAsyncData('todos', () => $fetch('/api/todos'))
...
-    todos.value.push(newTodo.value) // Optimistically update the todos.
-  },
-  onRequestError () {
-    todos.value = previousTodos.value // Rollback the data if the request failed.
-  },
-  async onResponse () {
-    await refreshNuxtData('todos') // Invalidate todos in the background if the request succeeded.
-  }
-})
+const addTodo = async () => {
```
atinux, line 80: As a not, I prefer using the function declaration when possible for code clarity when scanning what are variables vs functions:

```diff
- const addTodo = async () => {
+ async function addTodo () {
```

cc @Mini-ghost 

Beautiful PR btw ❤️

PR #30579, opened 2025-01-14
feat(nuxt): update to `@unhead/vue` v2

harlan-zw:
### 🔗 Linked issue

https://github.com/unjs/unhead/issues/395

### 📚 Description

The goal of Unhead v2 was to remove deprecations and remove the implicit context implementation.

- :heavy_check_mark: For v3 Nuxt users we ship a legacy Unhead v2 compatibility build that aims to avoid any breaking changes.
- :warning: For v4 Nuxt the removed deprecations may cause issues. End users will need to follow the migration guide (WIP)
  -  https://github.com/unjs/unhead/pull/445
  -  https://github.com/unjs/unhead/pull/447
  -  https://github.com/unjs/unhead/pull/440

### Implicit Context

The implicit context implementation would keep a global instance of Unhead available so that you could use Unhead in a non-Vue context.

The context concern is now something put on frameworks implementing Unhead. For Nuxt, we were already handling this with the hacky solution of `setHeadInjectionHandler()` for implementing the core composables in Nuxt itself so we can resolve straight from the Nuxt context. We move away from this as it's leading to false-positive [memory leaks](https://github.com/unjs/unhead/issues/281) (memory isn't freed post request).

```ts
// How did this know about the Unhead instance ?
useHead({
  title: 'hello world'
})
// In Nuxt we an equivalent of:
setHeadInjectionHandler(() => useNuxtApp().head)
```

With v2 this is dropped in favour of Nuxt itself re-exporting the composables while resolving the context.

```ts
export function injectHead (nuxtApp?: NuxtApp): VueHeadClient<MergeHead> {
  //  implement our composable to resolve the context
  const nuxt = nuxtApp || useNuxtApp()
  return  nuxt?.ssrContext?.head || nuxt?.runWithContext(inject)||  inject()
}

export function useHead<T extends MergeHead> (input: UseHeadInput<T>, options: NuxtUseHeadOptions = {}): ActiveHeadEntry<UseHeadInput<T>> | void {
  const unhead = injectHead(options.nuxt) // <-- use nuxt specific context
  return head(input, { head: unhead, ...options }) as ActiveHeadEntry<UseHeadInput<T>>
}
```

These changes introduce some potentially breaking changes:

#### :warning: Composable types changed

In Nuxt v3 it's possible for `useHead()` to return a `void` if there's no instance available as prior to this no error was thrown. In Nuxt v4 as we now can guarantee this works or an error is thrown the types are updated.

```diff
-function useHead(...): ActiveHeadEntry | void
+function useHead(...): ActiveHeadEntry
```

To solve this we need to export v3 and v4 composables. This is done and accomplished through a simple alias.

```ts
nuxt.options.alias['#unhead/composables'] = resolve(runtimeDir, 'composables', isNuxtV4 ? 'v4' : 'v3')
```

#### :warning: **Async Context lost when importing from `@unhead/vue`**

Many Nuxt apps may be directly importing from `@unhead/vue` instead of using `#imports`. These imports will **not** work outside of the Vue context (plugins, middleware).

```ts
import { useHead } from '@unhead/vue'

export default defineNuxtPlugin(async () => {
  await new Promise(resolve => setTimeout(resolve, 100))
  useHead({
    title: `title after async`,
  })
})
```

This will throw an error:

> useHead() was called without provide context, ensure you call it through the setup() function.

The solution is twofold: 
1. We need an easy way for end users to import the head utils when auto-imports are disabled.
    -  The import is `#app/composables/head`
2. We need to replace or warn about `@unhead/vue` imports.
    - This is being handled by an Unplugin plugin

We avoid any extra aliasing as it creates more issues than it solves.


danielroe:
Do you think we could ship the migration guide for v4 in `docs/1.getting-started/12.upgrade.md` as part of this PR?

packages/nuxt/src/head/plugins/unhead-imports.ts
```diff
@@ -0,0 +1,77 @@
...
+      const importsFromHead = importsToAdd.filter(specifier => !unheadVueComposablesImports[UnheadVue].includes((specifier.imported as Identifier)?.name))
+      if (importsFromUnhead.length) {
+        // warn if user has imported from @unhead/vue themselves
+        if (!id.includes('node_modules')) {
+          logger.warn(`You are importing from \`${UnheadVue}\` in \`./${relative(options.rootDir, id)}\`. Please import from \`#app\` instead for full type safety.`)
+        }
+        s.prepend(`import { ${toImports(importsFromUnhead).join(', ')} } from "#app/composables/head"\n`)
+      }
+      if (importsFromHead.length) {
+        s.prepend(`import { ${toImports(importsFromHead).join(', ')} } from "${UnheadVue}"\n`)
```
HigherOrderLogic, line 64: How about using `genImport()` from [knitwork](https://github.com/unjs/knitwork) here?

harlan-zw: Good idea, have implemented

docs/1.getting-started/12.upgrade.md
```diff
@@ -245,6 +245,52 @@ export default defineNuxtConfig({
 })
 ```
 
+#### Unhead V2
```
danielroe, line 248:
```suggestion
#### Unhead v2
```

danielroe (approved):
This looks 👌 to me.

harlan-zw:
I've pushed up changes for latest Unhead alpha which includes several performance improvements. This is the extent of things I wanted to include in Unhead v2 for Nuxt.

The only remaining change before the beta would be possibly moving `useScript()` back to the core which won't affect Nuxt.

packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -132,7 +132,7 @@ export function onPrehydrate (callback: string | ((el: HTMLElement) => void), ke
 
   useServerHead({
     script: [{
-      key: vm && key ? key : code,
+      key: vm && key ? key : undefined,
```
harlan-zw, line 135: As Unhead v2 does not create a hash from the provided `key`, this would set the `data-hid` to the script inner content. I don't think we need a key in these instances as:
- Unhead will dedupe based on inner content when no script is provided
- This is server-only so the key may not be needed at all (key is used to hydrate without causing duplicates in this instances)

test/fixtures/basic/plugins/my-plugin.ts
```diff
@@ -1,12 +1,17 @@
...
-    titleTemplate: '%s - Fixture',
-  })
-  const path = useRoute()?.path
-  return {
-    provide: {
-      myPlugin: () => 'Injected by my-plugin',
-      path: () => path,
-    },
-  }
+export default defineNuxtPlugin({
```
harlan-zw, line 1: I don't think `useHead()` was working from plugins for server components previously :thinking: 

We inadvertently fixed this in this PR but then it caused the test snapshots to break, moving this to not run within a server component was the easiest fix.

harlan-zw:
Any ideas on debugging these failed windows tests or can we assume it's just being flakey?

test/fixtures/basic/plugins/my-plugin.ts
danielroe: cc: @huang-julien

danielroe:
I've run in several times and it doesn't seem to be flaky tests (and it's all the windows + webpack/rspack tests). Do you have access to a windows machine? Want me to have a look?

harlan-zw:
Claude seems to have got it before I could pull out my old Windows laptop :smile: 

Btw the link is failing while DNS for new docs is being sorted

harlan-zw:
Going to remake this PR as it's a detached from the 3.x merge progress.

PR #30606, opened 2025-01-15
Feat/better dev err

danielphilipjohnson:
### 📚 Description

- Added Youch v4.1.0-beta.4 as a dependency for enhanced error pages

- Port error handling improvements from Pooya's PR, moving src/runtime/internal/error/dev.ts to packages/nuxt/src/core/runtime/nitro/error.ts in the nuxt/nuxt codebase.

### Possible enhancements

Screenshot from 2025-01-15 13-58-40

Current nuxt error page.

Screenshot from 2025-01-15 13-58-11

Current youch error page.

Current Nuxt error page shows the status code prominently at the top. We could maintain this in the Youch implementation by setting:

```js

title: ${errorObject.statusCode}
```

## Dependencies

- Added youch@4.1.0-beta.4

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/error.ts
@@ -80,15 +80,16 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error, 
     return send(event, template(errorObject))
   }
 
+  if (event.handled) { return }
+
   const html = await youch.toHTML(error, {
+    title: `${errorObject.statusCode}`,
     request: {
       url,
       method: event.method,
       headers: getRequestHeaders(event),
     },
   })
-
-  if (event.handled) { return }
 
   for (const [header, value] of res.headers.entries()) {
     setResponseHeader(event, header, value)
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -74,15 +80,22 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
...
-  const html = await res.text()
+  const html = await youch.toHTML(error, {
+    request: {
+      url,
+      method: event.method,
+      headers: getRequestHeaders(event),
+    },
+  })
+
   if (event.handled) { return }
```
HigherOrderLogic, line 92:
```suggestion
  if (event.handled) { return }
  
  const html = await youch.toHTML(error, {
    request: {
      url,
      method: event.method,
      headers: getRequestHeaders(event),
    },
  })
```

danielphilipjohnson: Sorry for the lateness will push this change shortly :pray:

danielroe:
thank you ❤️ 

linking also https://github.com/nuxt/nuxt/pull/30624 - and see comment in https://github.com/nuxt/nuxt/pull/30624#issuecomment-2594995699

danielroe:
marking as draft until implementation in nitro is finalised. ❤️

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/31230

PR #30624, opened 2025-01-16
feat(nuxt): better formatting errors

Saeid-Za:
### 📚 Description

This is an attempt to implement similar functionality in this [nitro](https://github.com/nitrojs/nitro/pull/3002) PR to provide better error formatting on the server.

I've tested this in two different scenarios: 
1. Direct error throwing before mounting the component
```vue app.vue
<script setup lang="ts">
throw new Error("Some Error")
</script>
```
2. Error throwing in `server/api` using raw errors or `createError`:
```ts test.ts
export default eventHandler((_event) => {
  throw createError("asd")
})
```

I tried to use the termial output for client side errors, but unfortunately the underlying library `youch` is tightly coupled to Node APIs.


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -1,93 +1,216 @@
...
+      console.error(`${tags} [${event.method}] ${event.path}\n\n`, errorToLog)
     }
+
     if (event.handled) { return }
-    setResponseHeader(event, 'Content-Type', 'text/html;charset=UTF-8')
-    return send(event, template(errorObject))
+
+    // Send response
+    setResponseStatus(event, (statusCode !== 200 && statusCode) as any as number || 500, statusMessage)
+    setSecurityHeaders(event, true /* allow js */)
```
Saeid-Za, line 84: There is an issue with csp headers, I suspect that this line is causing this issue, but can't confirm if this line should be deleted or not.

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/31230

PR #30660, opened 2025-01-19
chore: allow disabling snow effect using env variable

onmax:
### 🔗 Linked issue

Multiple people comment's in #29871.

### 📚 Description

In v3.15 a new Snow effect for winter has been introduced. I would like to keep the gradient animation but also would like to disable the snow effect completely.

I made this PR to allow configure the snow effect using an env variable that I will have in my `~/.bashrc`

Please let me know if there is a better approach :)

### DIFF
--- a/packages/schema/src/config/dev.ts
+++ b/packages/schema/src/config/dev.ts
@@ -38,6 +38,12 @@ export default defineUntypedSchema({
      * Template to show a loading screen
      * @type {(data: { loading?: string }) => string}
      */
-    loadingTemplate,
+    loadingTemplate: {
+      $resolve (data: Parameters<typeof loadingTemplate>[0], get) {
+        const template = loadingTemplate(data)
+        template.replace('{{ env.NUXT_TEMPLATE_DISABLE_SNOW_EFFECT }}', get('NUXT_TEMPLATE_DISABLE_SNOW_EFFECT') || 'false')
+        return template
+      },
+    },
   },
 })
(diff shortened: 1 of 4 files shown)


onmax:
Not sure why the errors in CI. Any idea?

danielroe:
you could dynamically set the loading template function by configuring the schema here using a `$resolve` function:

https://github.com/nuxt/nuxt/blob/64287dbd5c150775c1fe94a6d6dcd8ec7e805e40/packages/schema/src/config/dev.ts#L37-L42

packages/schema/src/config/dev.ts
```diff
@@ -38,6 +38,12 @@ export default defineUntypedSchema({
      * Template to show a loading screen
      * @type {(data: { loading?: string }) => string}
      */
-    loadingTemplate,
+    loadingTemplate: {
+      $resolve (data: Parameters<typeof loadingTemplate>[0], get) {
```
onmax, line 42: How do I test the changes? Where should I look?

I think the current implementation is wrong as I am not sure how to get the value from the `env`?
Should I use `useRuntimeConfig` somehow?

Sorry for the questions, but I am a bit lost. Thanks for any help!

danielroe:
I think we can close this as there is no snow in v4 ....

packages/schema/src/config/dev.ts
danielroe: it is definitely tricky. we would have needed to support a custom message/prop (maybe custom handling in packages/ui-templates/lib/render.ts). we'd then need to pass the env variable (or option) -> props/messages, which maybe we could do via a wrapped loadingTemplate in schema.

but ... surprisingly difficult

Issue #29263, opened 2024-10-03
`vue-tsc` overwrite server imports with app imports


Barbapapazes:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v20.17.0
- Nuxt Version:     3.13.2
- CLI Version:      3.14.0
- Nitro Version:    2.9.7
- Package Manager:  pnpm@9.12.0
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://github.com/Barbapapazes/nuxt-authorization/issues/12

Or

Clone the module https://github.com/Barbapapazes/nuxt-authorization, install deps, cd to the playground and run `npm run typecheck`

### Describe the bug

This module provides utilities with the same name on both the app and the server.

This means that it's possible to import `authorize` from `#imports` in both the app and the server but this function does not have the same type definition (the server one use `event`. When running `vue-tsc`, it raises an error because it resolves server imports using app imports.

<details><summary>`vue-tsc` error</summary>
<p>

```sh
nuxt-authorization/playground on  main [✘!?] via ⬢ v20.17.0 
➜ nr typecheck 

> nuxt-authorization-playground@ typecheck /Users/esoub/dev/p/nuxt-authorization/playground
> nuxi typecheck

ℹ Running with compatibility version 4                                                               3:02:34 PM
server/api/products/[id].patch.ts:11:19 - error TS2345: Argument of type 'H3Event<EventHandlerRequest>' is not assignable to parameter of type 'BouncerAbility<any>'.
  Type 'H3Event<EventHandlerRequest>' is missing the following properties from type 'BouncerAbility<any>': original, execute, allowGuest

11   await authorize(event, editProduct, product)
                     ~~~~~

server/api/products/index.get.ts:4:19 - error TS2345: Argument of type 'H3Event<EventHandlerRequest>' is not assignable to parameter of type 'BouncerAbility<any>'.

4   await authorize(event, listProducts)
                    ~~~~~

server/api/products/index.post.ts:4:19 - error TS2345: Argument of type 'H3Event<EventHandlerRequest>' is not assignable to parameter of type 'BouncerAbility<any>'.

4   await authorize(event, createProduct)
                    ~~~~~


Found 3 errors in 3 files.

Errors  Files
     1  server/api/products/[id].patch.ts:11
     1  server/api/products/index.get.ts:4
     1  server/api/products/index.post.ts:4

[...]

Issue #24087, opened 2023-10-27
fix: typecheck fails on imports which exist for nuxt and nitro


dargmuesli:
Run `npm exec nuxi typecheck` here: https://stackblitz.com/edit/github-dnls7g?file=server%2Futils%2Futil.ts

You'll see an error because a nuxt util has the same name as a server util but a different signature. I'd expect no error to be thrown as no error is shown to me in both files which make use of their respective utils.

---

I see that `.nuxt/tsconfig.json` does not `exclude` the `server` directory and that `.nuxt/tsconfig.server.json` does include `../**/*` by default. Is that meant to be this way? I'd expect `tsconfig.json` to exclude `server` and `tsconfig.server.json` to only include `server` and not the rest.

Alternatively, there should be an error shown in vscode matching the one given by the typecheck, in case the typecheck error is correct.

abbytec:
In addition, when i run `vue-tsc --no-emmit` or [type-checking](https://nuxt.com/docs/guide/concepts/typescript#type-checking) they use tsconfig.json. 

Which following [tsconfig.json config](https://nuxt.com/docs/guide/directory-structure/tsconfig) does not allow to import things like useStorage() (which are on "./.nuxt/tsconfig.server.json")

extending both, leads in overwritting their configs and depending on the order I put those:

-  useStorage() type is well imported but other things like useHead() don't
or
-  useHead() type is well imported but other things like useStorage() don't

dargmuesli:
I suspect the Nuxt v4 directory structure and likely also https://github.com/nuxt/nuxt/issues/28675 to resolve this issue, haven't checked yet though.

Barbapapazes:
I have a similar issue on my module https://github.com/Barbapapazes/nuxt-authorization/issues/12

dargmuesli:
This indeed seems to be resolved when using Nuxt v4's directory structure.

PR #30665, opened 2025-01-19
feat(kit,nuxt,schema): separate ts projects for node/app/nitro

### LINKED ISSUE #22572: separate type contexts for nitro, nuxt & modules/nuxt.config
### Environment

- Operating System: Darwin
- Node Version:     v18.16.1
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Package Manager:  pnpm@8.5.0
- Builder:          vite
- User Config:      devtools, nitro
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-22572?file=nuxt.config.ts

1. Open the repo and `nuxt.config.ts`
2. See no TS errors
3. See errors when running as `defineEventHandler` is not defined but gives false type safety.

### Describe the bug

In the `nuxt.config.ts`, imports like `defineEventHandler` or `$fetch` are wrongly "available" as auto-imports on the type-level, while using them without importing them would throw.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #31989: With `typeCheck: true`, `vue-tsc` checks `/server` dir with the wrong `tsconfig` file
### Environment

Any environment

### Reproduction

https://stackblitz.com/edit/github-epjdwiy5?file=server%2Fapi%2Ftest.get.ts

### Describe the bug

Nuxt's TypeScript setup is split up in two contexts:
- `/server` (`<rootDir>/server/tsconfig.json`, which extends `<rootDir>/.nuxt/tsconfig.server.json`)
- and everything else (`<rootDir>/tsconfig.json`, which extends `<rootDir>/.nuxt/tsconfig.json`).

When you modify those contexts manually through `nuxt.config.ts`, those two contexts can become quite different. For example, one might want to define an alias for the `/server` directory only and not make it available elsewhere too (see repro). Maybe you want to add certain files to `include` for the `/server` dir, but not for the rest of the application. Or perhaps you even want to manually add certain auto-imports for the app, but not for the `/server` dir.

When `typescript.typeCheck` is set to `true`, Nuxt seems to use a custom Vite plugin that transforms and prepends  files that need to be type checked with a comment that tells `vite-plugin-checker` to check those files. That in turn runs `vue-tsc` and uses its output to display it in the CLI and in the dev server UI. Currently, that doesn't take those two TS contexts into account. It always runs `vue-tsc` on all files that need to be type checked with the general/app `<rootDir>/tsconfig.json`. That means the `/server` directory is now typechecked with the wrong TS config file. When adding a `console.log(userConfig)` in `vite-plugin-checker/dist/main.js` in the `checker` function, this is confirmed by its output:
```
{ vueTsc: { tsconfigPath: '/var/www/experiments/nuxt-typecheck-wrong-tsconfig/tsconfig.json' } }
```

So to recap: in the repro, I wanted an alias called `#alias-test` only to work in the server directory. It does, but `vue-tsc` tells me I have a TS error, because that alias does not exist. That's because the alias is not registered in the main TS config (`<rootDir>/tsconfig.json`) and `vue-tsc` wrongfully type checks the `/server` dir with **that** TS config file:

Image

Why is this a problem? Well, in the runtime, the written code from the repro works perfectly, but once you deploy your app and your CI/CD fails a build, because you have TS errors, this starts to suck.

Ideally, Nuxt would somehow run `vite-plugin-checker` twice, with different TS config files and scanning different files. Even more ideally, the whole TypeScript config setup might need a rework, because there are other issues, like `/shared` using the general/app tsconfig file (`<rootDir>/tsconfig.json`), while its files could also run in the server TS context. I won't dive into that here though.

### Additional context

**Note:** In StackBlitz, the TS error that `vue-tsc` outputs in the CLI is also displayed in the "IDE" editor itself. In VS Code (with the Vue (Official) extension), this is not the case:

Image

### Logs

```shell-script

```

### LINKED ISSUE #32343: separate type contexts for nuxt, nitro and node
https://github.com/nuxt/nuxt/issues/22572
https://github.com/nuxt/nuxt/issues/31989

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/cli/issues/323
* https://github.com/nuxt/nuxt/issues/24087
* https://github.com/nuxt/nuxt/issues/29263

resolves https://github.com/nuxt/nuxt/issues/31989
resolves https://github.com/nuxt/nuxt/issues/32343
resolves https://github.com/nuxt/nuxt/issues/22572

### 📚 Description
Not sure but I can imagine these changes may break other functionality.

* `skipLibCheck` needs to be enabled to get around type errors when `tsconfig.server.json` is used for typecheck. 
~~* Removes `./types/nitro.d.ts` from `nuxt.d.ts` since that bypasses the `exclude` entries by having explicit imports to server files (https://github.com/nuxt/cli/issues/323#issuecomment-2600912950).~~
* `tsconfig.server.json` has an `../**/*` entry in the `include` array, having this adds will also include client files, this is added based on Nitro's `rootDir`, so as a hack/workaround I assign the `rootDir` to `srcDir` (which I think is the `serverDir`) which replaces/removes the entry.

---

Okay I think I'm starting to understand the underlying issue.

The `nitro.d.ts` reference is required to have types for nitro routes/endpoints from within the app/client code, this causes issues when those types rely on different aliases/imports/types isolated to the nitro typescript config (as it does in https://github.com/nuxt/nuxt/issues/29263). This is also the source of the nitro auto import autocompletion in app/client code.


danielroe:
we might need a hook for nitro like `prepare:types` for nuxt that lets us modify the generated file

also, I'm thinking of moving to this pattern for the tsconfig:

https://github.com/danielroe/roe.dev/blob/main/tsconfig.json

which would allow a single pass for type checking rather than multiple calls.

docs/1.getting-started/18.upgrade.md
```diff
@@ -1047,7 +1047,7 @@ However, to take advantage of improved type checking, you can opt in to the new
    export default defineNuxtConfig({
      typescript: {
        // Customize app/server TypeScript config
-       tsConfig: {
+       TSConfig: {
```
danielroe: 🤔

danielroe (approved):

PR #30670, opened 2025-01-20
docs: explain how to use ClientOnly with onMounted hook

zakshyman:
### 📚 Description

Added to <ClientOnly> documentation page explanation of how to access that template. Made similar to this https://nuxt.com/docs/guide/directory-structure/components#client-components  

::important
Components inside `<ClientOnly>` are rendered only after being mounted. To access the rendered template using `onMounted()`, add `await nextTick()` in the callback of the `onMounted()` hook.
::


danielroe (commented):
we could alternatively advise watching the template ref, as this will be less fragile to internal changes in rendering

Mini-ghost:
Not sure if you agree, but could this be a more reliable approach?  

```vue
<script setup lang="ts">
const nuxtWelcomRef = ref()

// Use `watchEffect`
watchEffect(() => {
  if (!nuxtWelcomRef.value) return
  console.log('<NuxtWelcome /> mounted')
})

// Use watch
watch(nuxtWelcomRef, () => {
  console.log('<NuxtWelcome /> mounted')
})
</script>

<template>
  <ClientOnly>
    <NuxtWelcome ref="nuxtWelcomRef" />
  </ClientOnly>
</template>
```

danielroe:
I would advise `{ once: true }`, and `watch` rather than `watchEffect`. What do you think?

Mini-ghost (approved):

PR #25850, opened 2024-02-18
feat: implement `force` option for refreshing and `triggeredBy` for `getCachedData`

TheAlexLichter:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/24332.

And linking https://github.com/nuxt/nuxt/issues/24271 for prep work

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds a `force` option to to `refresh` function (from `useAsyncData`) and a force option to `refreshNuxtData`. It allows to bypass a possibly existing `getCachedData` function and forces to refetch.

As `force` cannot be set when watching params, e.g. through `useFetch` or manually in `useAsyncData`, `getCachedData` now receives the "context" that triggered the function and requests the cached data, allowing more granular results.

Happy for feedback on the implementation.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have added tests (if possible).
- [ ] I have updated the documentation accordingly.

PR #30678, opened 2025-01-21
`useCache` option for `useFetch` and `useAsyncData`

### LINKED ISSUE #24332: Re-use `getCachedData` when `refresh`ing `asyncData`

### LINKED ISSUE #30663: `useFetch` updates in `dev` and is not changing after `generate`
### Environment

Nuxt v3.15.12 (Windows and Stackblitz)

### Reproduction

https://stackblitz.com/edit/nuxt-starter-sf8je8kp?file=pages%2Findex.vue

### Describe the bug

Launch app in `dev` mode.

When you click on link "Page" and "Index" you see data changing on `useFetch` field and "Payload".
This is the expected behaviour: there is no static data and both `useFetch` and payload are updated when page is changed.

---

Now run `npm run generate` and then `npx serve .output/public`.

Immediately you see "Static" is empty, while it should not be empty.
And when the page is changed for the first time, "Static" data suddenly appears and changes correctly but, what is more confusing, both `useFetch` and "Payload" fields are not changing!

If you reload the page, it will actually get the correct value from `_payload.json`.
But again it will not update when page is changed.

In my understanding when the page is changed lets say from "Index" to "Page" it should see, that "Page" page (lol) has a different payload value for `apiData` property and update it. Instead, the value of  `apiData` stays the same.

---

I really don't understand what is happening, what is expected and what is not here. Is `dev` version correct? Or `generate` version? Why they behave differently?

There are some options to workaround this but they all have disadvantages:

1. Store each useFetch result under unique payload name. **The problem** is obvious — I am duplicating data that already exists in `_payload.json`. What if I have a ton of data? Duplicate it too?
2. `prerenderRoutes(...)` for both URLs and use `$fetch`. **The problem** — it is impossible to preprender routes with query parameters

Please help me, all this `useFetch` and `useAsyncData` stuff is killing me)

In my project I managed to bypass and go around using these composables but this time I can't go around because I need to `useFetch` server with query parameters and that is why I cant use `prerenderRoutes`.

In simple words — how do I get  payload data after `generate` which is present both in `_payload.json` for each route and in each `<name>.html` file?

Gwynerva:
### 🔗 Linked issue

Resolves #24332 and partially resolves #30663

### 📚 Description

Right now Nuxt calls `getCachedData` only once — on initial `useFetch` of `useAsyncData` call. Using `refresh` or `execute` functions or just changing the ref its is watching will skip looking for cached data and tries to fetch.

This completely messes things up for SSG sites created via `nuxi generate` and requires building workarounds for manually checking for static payload first.

So I added an additional boolean option `useCache` in both `useFetch` and `useAsyncData` composables that when `true` will always make them `getCachedData` first.


Gwynerva:
Does not work of `useFetch`. Will fix in a moment.

Gwynerva:
Seems to work now but non-reactive `key` property (#21532) blocks `useCache` full pontential.

Gwynerva:
I don't know why but it fails to return updated `getCachedData` values.

This test fails:

```ts
it('should always call getCachedData when `useCache` is true', async () => {
  let cacheValue = 1;
  const getCachedData = vi.fn(() => ({ val: cacheValue++ }))
  const { data, refresh } = await useAsyncData(() => Promise.resolve({ val: 'fetch' }), { useCache: true, getCachedData })

  for (let i = 0; i < 3; i++) {
    await refresh()
  }

  expect(data.value).toMatchInlineSnapshot(`
    {
      "val": 4,
    }
  `)
  expect(getCachedData).toHaveBeenCalledTimes(4)
})
```

Data value equals `1`, not `4` although `getCachedData` is called four times...

Gwynerva:
Now `data` ref always gets the most recent cache value.

danielroe:
linking https://github.com/nuxt/nuxt/pull/25850 - @TheAlexLichter had started some work on this and may have some ideas on the best way to implement

Gwynerva:
@danielroe Not directly related to this but I think we need some handy tools to manage payload directly and also correct the docs.

The docs and official Nuxt Dev Tools says "data/payload" is related to `useState`, `useFetch` and `useAsyncData` composablses but it is actually so much more than this. It is a unified space for storing application state.

image


It is never said anywhere, that setting `nuxtApp.static.data` is actually the data that will be saved in `_payload.json` when prerendering pages.

This can be used to precisely set the data that is needed to be prerendered.
For example, this is how I save table of contents data for each article page without using any of Nuxt data fetching composables:

```ts
export async function getToc(location: Location)
{
    const nuxtApp = useNuxtApp();
    const strLocation = stringifyLocation(location);
    const tocData = nuxtApp.static.data['toc'] ||= nuxtApp.payload.data['toc'] ||= {};
    return tocData[strLocation] ||= await $fetch(`/api/bitran/toc/${encodeLocation(strLocation)}`);
}
```

With this code, all TOCs are saved under on payload key `toc`, not polluting payload namespace.
This is impossible to do with stardart composables.

image

PR #30634, opened 2025-01-16
refactor(vite): drop `externality` and use vite internal config

danielroe:
### 📚 Description

this drops `externality` which should clean up the dependency tree a _bit_ ([graph](https://npmgraph.js.org/?q=externality)).

this will need testing to ensure behaviour is still the same in dev - you can use the packages published in [this comment](https://github.com/nuxt/nuxt/pull/30635#issuecomment-2596163109) to do so.

PR #30690, opened 2025-01-21
fix(vite): inline shared folder in dev mode

danielroe:
### 📚 Description

this fixes a regression from https://github.com/nuxt/nuxt/pull/30634.

we inline `#shared` folder in dev mode only

(there may be a better solution)


danielroe:
effectively what that does is the same as this, I believe - it gives the shared folder higher priority than the `external` setting.

Mini-ghost (approved):

PR #30704, opened 2025-01-22
feat(nuxt): add type hints for `NuxtPage`

Mini-ghost:
### 📚 Description

Add `props` and `slots` type to `<NuxtPage>` to enhance the developer experience. The type descriptions were adapted from the documentation.

截圖 2025-01-22 晚上7 07 25

If you have any suggestions, please let me know. Thank you!


packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -14,6 +14,23 @@ import { LayoutMetaSymbol, PageRouteSymbol } from '#app/components/injections'
...
+  /**
+   * Control state preservation of pages rendered with the `NuxtPage` component.
+   */
+  keepalive?: boolean | KeepAliveProps
+
+  /**
+   * Control when the `NuxtPage` component is re-rendered.
+   */
+  pageKey?: string | ((route: RouteLocationNormalizedLoaded) => string)
+}
```
Mini-ghost, line 32: It seems we can’t export `NuxtPageProps` like `NuxtLink`. Any suggestions on how we could handle this?

danielroe: We could re-export, as we do here:

- https://github.com/nuxt/nuxt/blob/f209158352b09d1986aa320e29ff36353b91c358/packages/nuxt/src/app/types.ts#L1
- https://github.com/nuxt/nuxt/blob/33789bec056b6409bd99cdf520db53d7bee62076/packages/nuxt/src/app/index.ts#L11

Mini-ghost: Thank you for your guidance.

danielroe (approved):
❤️

PR #30708, opened 2025-01-22
feat(nuxt): extract `middleware` when scanning page metadata

### LINKED ISSUE #30689: support scanning `middleware` page metadata
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.2
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      compatibilityDate, devtools, hooks
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-tp1ghhcz?file=nuxt.config.ts

### Describe the bug

The nuxt documentation states that the "pages:resolved" hook is 
>[Called after page routes have been augmented with scanned metadata.](https://nuxt.com/docs/api/advanced/hooks#nuxt-hooks-build-time:~:text=the%20file%20system.-,pages%3Aresolved,-pages)

Yet the page.meta object seems to be empty. This prevents the ability to add middleware instead of completely replacing the middleware to a page.

### Additional context

_No response_

### Logs

```shell-script

```

Saeid-Za:
### 🔗 Linked issue

resolves #30689 

### 📚 Description

Adding `middleware` key to default extracted keys in `scanPageMeta`


danielroe:
@Saeid-Za Did you have a look at the failing tests?

Saeid-Za:
For some unknown reason to me, whenever middleware is added to the default extraction keys, every middleware fails to function, as if the middleware code snippet was removed.

By comparing the output of the previous commit to this PR, I found out that middleware should always have a meta key as dynamic property.

This is a workaround, as I couldn't find the root cause that enforced this patch.

danielroe (approved):

PR #30717, opened 2025-01-23
fix(nuxt): use klona to deep clone extractedMeta

### LINKED ISSUE #30695: experimental.scanPageMeta 'after-resolve' combines metas from same-file routes
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.2
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      compatibilityDate, devtools, experimental
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

Set experimental.scanPageMeta = `after-resolve`
Add routes to existing pages via pages:extended hook.
Add meta data to some of those routes

https://stackblitz.com/edit/nuxt-starter-tpklcoio

### Describe the bug

When adding multiple routes to the same file via pages:extended hook, meta tags are combined top-down to each route when they should be separate as they are different routes.

Image

### Additional context

`experimental.scanPageMeta: true` = good
`experimental.scanPageMeta: false` = good

### Logs

```shell-script

```

xjccc:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/30695

### 📚 Description

I think add a unique cacheKey with `hash(route)` could resolve this?

If has any other idea, please tell me. Thank you very much.


packages/nuxt/src/pages/utils.ts
```diff
@@ -313,7 +316,7 @@ export async function getRouteMeta (contents: string, absolutePath: string, extr
     })
   }
 
-  metaCache[absolutePath] = extractedMeta
+  metaCache[cacheKey] = extractedMeta
   return extractedMeta
```
danielroe, line 320: would it resolve it just to do something like:

```suggestion
  return JSON.parse(JSON.stringify(extractedMeta))
```

xjccc:
```
  if (absolutePath in metaCache && metaCache[absolutePath]) {
    return metaCache[absolutePath]
  }
```

I think it's the reason here? Has `key of absolutePath` cache, `return the first cache`?

Because the file is the same by `page:extends`?

danielroe: yes, we'd also have to update it there too

xjccc: Yes, you are right. Thank you very much. I will update it.

xjccc: Oh.. `extraxtedMeta` has `Set` properties. So, `JSON.stringify` has some problem..

danielroe: we could use klona

xjccc: I have changed this. Please check it out. Thank you very much. ❤

danielroe: you had some nice tests in the previous commit…. is it worth keeping them?

packages/nuxt/test/page-metadata.test.ts
```diff
@@ -63,8 +63,9 @@ definePageMeta({ name: 'bar' })
 
   it('should use and invalidate cache', async () => {
```
danielroe, line 64: To make this test work, I think we probably have to mock `klona` so it doesn't modify the object.

Otherwise it doesn't in fact test whether the cache is used or not.

xjccc: Thank you. I will look at it next day. 🫡

danielroe: pushed a quick test update so I can merge for the patch later today

thank you for your work on this 🙏

danielroe (approved):

PR #30720, opened 2025-01-23
perf(nuxt): enable `Transition` component only on client side

Mini-ghost:
### 📚 Description

- Referenced the approach used in `wrapInKeepAlive`, ensuring that the `<Transition>` components inside `<NuxtPage>` and `<NuxtLayout>` are initialized only on the client side.  
- Separated the code executed on the server from the code executed on the client, which may improve clarity when reading.


packages/nuxt/src/pages/runtime/utils.ts
```diff
@@ -25,6 +25,10 @@ export const wrapInKeepAlive = (props: any, children: any) => {
   return { default: () => import.meta.client && props ? h(KeepAlive, props === true ? {} : props, children) : children }
 }
 
+export const wrapInTransition = (props: any, children: any) => {
+  return { default: () => import.meta.client && props ? h(Transition, props === true ? {} : props, children) : children.default?.() }
+}
```
danielroe, line 30: because this utility function is needed by nuxt core (not just the pages module), we should probably keep it in `packages/nuxt/src/app/components/utils.ts` (unless there's a reason I haven't spotted that it needs to be moved)

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -101,8 +100,30 @@ export default defineComponent({
             nuxtApp.callHook('page:loading:end')
             pageLoadingEndHookAlreadyCalled = true
           }
+
           previousPageKey = key
 
+          if (import.meta.server) {
```
danielroe, line 106: looking at the vue implementation of `Transition` for SSR, is there any performance benefit or reason to have a separate implementation in nuxt? (I'm guessing yes, but just want to confirm.)

Mini-ghost: I looked into the implementations of `<KeepAlive>` and `<Transition>`.

It seems that `<KeepAlive>` in SSR scenarios returns the default slots early on the server side without continuing to execute its internal implementation.  

https://github.com/vuejs/core/blob/29216853d4eb2bc72c284bcc5234afbca9c81854/packages/runtime-core/src/components/KeepAlive.ts#L79-L109

For `<Transition>`, I couldn’t find any related SSR-specific implementations in `BaseTransition`.  

https://github.com/vuejs/core/blob/29216853d4eb2bc72c284bcc5234afbca9c81854/packages/runtime-core/src/components/BaseTransition.ts#L145-L267

The reason for opening this PR is that `<Transition>` involves DOM operations, so even if the component is initialized on the server side, its functionality wouldn’t be utilized. Additionally, `<Transition>` does not generate extra DOM nodes, so skipping its execution on the server side will not cause hydration errors on the client side. Based on these considerations, I’ve made adjustments in this PR.  

Separating the server-side and client-side implementations also helps to clearly show the differences between the two, rather than relying on numerous conditional checks to understand the logic.  

If there are any adjustments or considerations I might have missed, please let me know. I truly appreciate your feedback. Thank you very much!

danielroe (approved):

PR #30722, opened 2025-01-23
chore: provide version in `package.json`

iiio2:
Here in the package.json file, version is empty. That is why I thought we can give the current version of Nuxt.


danielroe:
Thank you! But I don't think we need to add a version to the repository root. It's not a published package.

iiio2:
Oh, ok @danielroe. Thank you.

PR #30743, opened 2025-01-24
fix(nuxt): change default assets filename

huang-julien:
### 📚 Description

This PR change the default assets filenames. It seems some end-users can still experience CORS issues with css chunks on chromium based browsers due to a bug. 

The workaround in vite was to change the generated hash. Changing the default asset filename would ensure at a 100% that end-users would not encouter this bug anymore.


danielroe (commented):
is there a way we can force the hash to be updated without changing the naming pattern?

huang-julien:
maybe with augmentHash ?  
But this was supposed to be fixed by Vite. Maybe it doesn't apply to assets files ? 
This would also make our pattern align with rollup's default pattern `assets/[name]-[hash][extname]`

danielroe:
I'm very up for changing the asset filenames but let's do this to align with Vite for v4/v4 compat (as it will be a breaking change for some).

danielroe:
I'm closing but if you still think we need to do this, just reopen 🙏

PR #30765, opened 2025-01-26
feat: add import.meta.envName

maximepvrt:
### 📚 Description

Currently, the environment name is defined only by `process.env.NODE_ENV` and is not documented.

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -57,6 +57,7 @@ export async function buildClient (ctx: ViteBuildContext) {
       'import.meta.browser': true,
       'import.meta.nitro': false,
       'import.meta.prerender': false,
+      'import.meta.envName': JSON.stringify(ctx.config.mode),
       'module.hot': false,
       ...nodeCompat.define,
     },
--- a/docs/3.api/6.advanced/2.import-meta.md
+++ b/docs/3.api/6.advanced/2.import-meta.md
@@ -27,7 +27,7 @@ Property | Type | Description
 `import.meta.dev` | boolean | True when running the Nuxt dev server.
 `import.meta.test` | boolean | True when running in a test context.
 `import.meta.prerender` | boolean | True when rendering HTML on the server in the prerender stage of your build.
-`import.meta.envName` | string | Specifies the current environment name (e.g., 'development', 'production', …).
+`import.meta.envName` | string | Specifies the current environment name (e.g., 'development', 'production', 'staging', …).
 
 ## Builder Properties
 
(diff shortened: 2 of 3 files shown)


packages/vite/src/client.ts
```diff
@@ -57,6 +57,7 @@ export async function buildClient (ctx: ViteBuildContext) {
       'import.meta.browser': true,
       'import.meta.nitro': false,
       'import.meta.prerender': false,
+      'import.meta.envName': JSON.stringify(ctx.config.mode),
```
danielroe, line 60: thank you for this!

the options here for are `production`, `development` or `test`, and they can be accessed in `import.meta.dev` and `import.meta.test`.

in general adding `envName` is more verbose - `production` vs a simple boolean, so I'm less inclined to add this feature.

what are you thinking this adds?

maximepvrt: @danielroe Thanks for your response!

I understand your point about the verbosity of adding an `envName`. However, as per the [Nuxt documentation](https://nuxt.com/docs/api/commands/build#options), it’s now possible to define a custom `envName`, like `staging`, which offers a more powerful way to manage configurations across different environments, as outlined [here](https://nuxt.com/docs/getting-started/configuration#environment-overrides). 

The challenge, however, is that there is currently no documented way to access the `envName` within the application. For example, I might want to pass a different logo color for the `staging` environment. I was hoping to have a documented variable to handle this, especially in the context of this [Sentry issue](https://github.com/getsentry/sentry-javascript/issues/15147).

danielroe: The `envName` is not accessible even in this PR; it is not the same thing as `mode`.

maximepvrt: @danielroe Thanks for the clarification!

Given that envName is a build-time concept, would it be possible for the build to generate a variable that could be accessed at runtime? This could be really useful for modules (like Sentry) and would also help with the example I mentioned earlier.

danielroe: That's not a bad idea!

But the sentry issue should be fixed by checking `nuxt.options.dev`.

maximepvrt:
https://github.com/nuxt/nuxt/issues/30776

packages/vite/src/client.ts
HigherOrderLogic: @danielroe 

Can I take this issue, following the path of generating a runtime variable like stated above?

danielroe: if @maximepvrt is happy to carry on with it, I think it makes sense for him to update the PR...

PR #30771, opened 2025-01-26
fix(nuxt): allow restarting nuxt on paths outside `srcDir`

danielroe:
### 📚 Description

With upcoming change to move local `modules/` out of the `srcDir`, changes to local modules no longer trigger restarts

this PR allows watching selected paths outside of `srcDir` (adding them individually to the existing watcher).

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -414,7 +414,6 @@ async function initNuxt (nuxt: Nuxt) {
   await nuxt.callHook('modules:before')
   const modulesToInstall = new Map<string | NuxtModule, Record<string, any>>()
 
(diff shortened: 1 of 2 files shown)


Akryum (approved):

packages/nuxt/src/core/nuxt.ts
```diff
@@ -660,10 +659,6 @@ export default defineNuxtPlugin({
 
   nuxt.hooks.hook('builder:watch', (event, relativePath) => {
     const path = resolve(nuxt.options.srcDir, relativePath)
-    // Local module patterns
-    if (watchedPaths.has(path)) {
-      return nuxt.callHook('restart', { hard: true })
```
danielroe, line 665: note to self - we may have lost the hard restart - will need to confirm

PR #30774, opened 2025-01-26
fix(kit): load `@nuxt/schema` from `nuxt` package dir

danielroe:
### 📚 Description

this addresses a potential issue we had - that every nuxt kit brought its own version of the nuxt defaults, whereas we only want to load the version of nuxt's defaults that matches the version of nuxt.

more changes/refactors will be made to this part of the code, but this is a separate fix we can either merge in now, or keep pending until we move more of the defaults into nuxt itself

### DIFF
--- a/packages/kit/package.json
+++ b/packages/kit/package.json
@@ -27,7 +27,6 @@
     "test:attw": "attw --pack"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


huang-julien (approved):
:+1:

PR #30781, opened 2025-01-27
docs: add additional information about `NuxtPage`

Mini-ghost:
### 📚 Description

This PR includes some additions and adjustments to the `<NuxtPage>` documentation:

- Added a structural diagram of the components used internally within `<NuxtPage>`, which may help users better understand this component.  
- Included a note that when `<Transition>` is enabled, the page component must have a single root element.  
- Revised the example for Custom Props to make it clearer for users on how to use them.  

I would greatly appreciate your feedback on these updates. Thank you!


atinux:
I really like the idea of explaning how NuxtPage works under the hood 💯

docs/3.api/1.components/2.nuxt-page.md
```diff
@@ -100,14 +119,32 @@ defineExpose({
...
+We can access the `foobar` prop in the page component:
+
+```vue [pages/page.vue]
+<script setup lang="ts">
+const props = defineProps<{ foobar: number }>()
+
+console.log(props.foobar) // Outputs: 123
+```
+
+If without `defineProps`, you can access the `foobar` prop directly:
```
danielroe, line 141:
```suggestion
If you have not defined the prop with `defineProps`, any props passed down to `NuxtPage` can still be accessed directly from the page `attrs`:
```

docs/3.api/1.components/2.nuxt-page.md
```diff
@@ -11,34 +11,53 @@ links:
...
+  - type: `boolean` or [`KeepAliveProps`](https://vuejs.org/api/built-in-components#keepalive)
 
 ::tip
 Nuxt automatically resolves the `name` and `route` by scanning and rendering all Vue component files found in the `/pages` directory.
 ::
 
 ## Example
 
-For example, passing `static` key, `NuxtPage` component is rendered only once when it is mounted.
+For example, passing `static` key, `<NuxtPage>` component is rendered only once when it is mounted.
```
danielroe, line 60:
```suggestion
For example, if you pass a key that never changes, the `<NuxtPage>` component will be rendered only once - when it is first mounted.
```

danielroe (approved):

PR #30784, opened 2025-01-27
feat(nitro,nuxt): add support for payload extraction in dev

### LINKED ISSUE #30679: payload extraction support in development mode
Nuxt has full support for payload extraction in production, when run with `nuxi generate` or `nuxi build --prerender` but there's no way to experience this in development (or identify possible bugs).

It would be good to support payload extraction when explicitly enabled by setting `nitro.static` to `true` (or by enabling `experimental.payloadExtraction` directly).

We might also consider moving `payloadExtraction` out of experimental namespace and/or renaming it.

Saeid-Za:
## 🔗 Linked Issue

Resolves #30679

## 📚 Description

This pull request introduces support for development mode in payload extraction.

To validate this implementation, I added  `prerender` route rule to the playground page and confirmed that `useNuxtApp().static` successfully extracts the data as intended.
I've also verified that `_payload` endpoint, adds a `prerenderedAt` timestamp to the response.

Please let me know if this aligns with your expectations, Thanks!


Saeid-Za:
This is a windows-only failing test and unfortunately, I cannot reproduce the issue on my `windows 11`.

I'm sorry for overlooking the test results, but since there are always some false-positives, I couldn't determine if the change in code is breaking things or not.

danielroe (commented):
we probably need to add some extra conditions for when payload extraction will be enabled.

If `nitro: { static: true }` is set in the config, then payload extraction should be enabled for every page. Otherwise, we probably need a route rule for `prerender: true` to trigger payload extraction in dev mode.

Finally (and maybe not in this PR, but just making a note of it) we need a way to enable payload extraction for routes that are not prerendered.

Saeid-Za:
From my understanding,  `nitro: { static: true }` is a `Nitro` config not accessible in `NitroApp` or `NuxtApp`, one solution would be to add it to `manifest.json`.
Is this the desired way?

danielroe (commented):
It's not enough to add the payload plugin - we also have to enable this `_PAYLOAD_EXTRACTION` variable - either when `nitro.static` is set, or route rules specify the page will be prerendered:

https://github.com/nuxt/nuxt/blob/a7b9df11c3f17a9d320905d97f1275d4f554966f/packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts#L119-L121

danielroe (approved):
thank you! ❤️

Issue #28893, opened 2024-09-09
warning about 'Calling 'useRoute' within middleware may lead to misleading results' when it is in the '<script setup>' of the SFC component


n0099:
### Environment

- Operating System: Linux
- Node Version:     v22.8.0
- Nuxt Version:     3.13.1
- CLI Version:      3.13.1
- Nitro Version:    2.9.7
- Package Manager:  yarn@4.4.1
- Builder:          -
- User Config:      compatibilityDate, devServer, devtools, srcDir, imports, modules, pinia, eslint, vueQuery, site, sitemap, ogImage, schemaOrg, features, sourcemap, build, vite, experimental, runtimeConfig
- Runtime Modules:  @nuxt/eslint@0.5.6, @pinia/nuxt@0.5.4, @vueuse/nuxt@11.0.3, @ant-design-vue/nuxt@1.4.2, @hebilicious/vue-query-nuxt@0.3.0, @vesp/nuxt-fontawesome@1.1.0, @nuxtjs/seo@2.0.0-rc.21
- Build Modules:    -

### Reproduction

```bash
git clone https://github.com/n0099/open-tbm && cd open-tbm
git checkout e7c516a2ac5bad2e726074fe9155c9b93c940df0
cd fe
yarn
cat <<'ENV' > .env
NUXT_APP_BASE_URL=/tbm
NUXT_SITE_URL=https://localhost
NUXT_PUBLIC_BE_URL=https://n0099.net/tbm/be
ENV
yarn dev
```
access https://localhost/tbm/posts/fid/228500/cursor/vNcpHQI,bLKIZg,lMJmGCM,mKqzZg,W3eEFSM,kuKoZg in browser and open devtools
scroll to the bottom of page and click the button with text `下一页`
you can see many `console.warn()` with callstack at https://github.com/n0099/open-tbm/blob/e7c516a2ac5bad2e726074fe9155c9b93c940df0/fe/src/components/post/badge/Common.vue#L49
![image](https://github.com/user-attachments/assets/d7dc73c4-7f11-4671-a890-a3d807230ed6)

### Describe the bug

https://github.com/nuxt/nuxt/blob/d5cf409ddca078266bd71ca011b9bc6ed6e58e05/packages/nuxt/src/app/composables/router.ts#L21-L22
https://github.com/nuxt/nuxt/blob/d5cf409ddca078266bd71ca011b9bc6ed6e58e05/packages/nuxt/src/app/composables/router.ts#L81-L84
[...]

PR #30788, opened 2025-01-27
fix(nuxt): don't warn about calling `useRoute` in SFC setup

Gwynerva:
### 🔗 Linked issue

Related [#28893](https://github.com/nuxt/nuxt/issues/28893#issuecomment-2615825977)

### 📚 Description

Provides more info about the problem of calling `useRoute` in the process of page change.
Current warn does not cover cases where an SFC component that calls `useRoute` in `setup` is created during page changing.

Also, if there is a way to tell if `useRoute` is called inside setup function of component, I think it would be better to check on this and leave the old warn message.

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -19,7 +19,7 @@ export const useRouter: typeof _useRouter = () => {
 /** @since 3.0.0 */
 export const useRoute: typeof _useRoute = () => {
   if (import.meta.dev && isProcessingMiddleware()) {
-    console.warn('[nuxt] Calling `useRoute` within middleware may lead to misleading results. Instead, use the (to, from) arguments passed to the middleware to access the new and old routes.')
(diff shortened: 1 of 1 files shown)


danielroe:
perhaps we can use `getCurrentInstance` and skip the warning if there's one?

Gwynerva:
@danielroe Yes, your suggestion is correct, thank you.

Also, I was only able to reproduce the problem when calling `useRoute` after a top-level await. God, this `await` stuff kills me. It breaks things so unintuitively sometimes :(

Still, I have added a check of being inside SFC and the warn is not logging in such cases anymore!

packages/nuxt/src/app/composables/router.ts
```diff
@@ -18,7 +18,8 @@ export const useRouter: typeof _useRouter = () => {
 
 /** @since 3.0.0 */
 export const useRoute: typeof _useRoute = () => {
-  if (import.meta.dev && isProcessingMiddleware()) {
+  const notInsideSFC = !getCurrentInstance()
+  if (import.meta.dev && isProcessingMiddleware() && notInsideSFC) {
```
danielroe, line 22:
```suggestion
  if (import.meta.dev && isProcessingMiddleware() && !getCurrentInstance()) {
```

danielroe (approved):

PR #30854, opened 2025-02-04
Fix/allow ignore options

### LINKED ISSUE #7632: path should be a path.relative()d string, but got "../../store/users.js"
### Version

[v2.13.2](https://github.com/nuxt.js/releases/tag/v2.13.2)

### Reproduction link

[https://codesandbox.io/s/github/zabavnikov/nuxt-store](https://codesandbox.io/s/github/zabavnikov/nuxt-store)

### Steps to reproduce

Run codesandbox

### What is expected ?

Nuxt will find store files

path should be a path.relative()d string, but got "../../store/users.js"

### Additional comments?

Prior to version 2.13.0, such a path to the store worked

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/nuxt.js/issues/c10868">#c10868</a>)</em></sub></div>

lgcardeas:
### 🔗 Linked issue

"resolves #7632"

### 📚 Description

In a monolithic app, were we share common features, like layout, middleware, components, etc.
If those features are out of the srcDir, it will fail with `path should be a path.relative()d string, but got "../......"`

`@nuxt/builder` uses Ignore class, which already accepts `ignoreOptions` however, these options are never passed down to that class or even accepted. 

from ignore package https://www.npmjs.com/package/ignore:
```
options.allowRelativePaths?: boolean since 5.2.0
This option brings backward compatibility with projects which based on ignore@4.x. If options.allowRelativePaths is true, ignore will not check whether the given path to be tested is [path.relative()d](https://www.npmjs.com/package/ignore#pathname-conventions).
```

Reproduced issue: https://github.com/lgcardeas/Nuxt_Relative_Path_Not_Working

### DIFF
--- a/packages/builder/src/builder.js
+++ b/packages/builder/src/builder.js
@@ -80,7 +80,8 @@ export default class Builder {
 
     this.ignore = new Ignore({
       rootDir: this.options.srcDir,
-      ignoreArray: this.options.ignore
+      ignoreArray: this.options.ignore,
+      ignoreOptions: this.options.ignoreOptions
     })
   }
 
--- a/packages/builder/test/builder.ignore.test.js
+++ b/packages/builder/test/builder.ignore.test.js
@@ -8,7 +8,7 @@ import Builder from '../src/builder' // Adjust path as needed to import your loc
 /**
  * Write a minimal nuxt.config.js to the test directory
  */
-function writeNuxtConfig(dir, config) {
+function writeNuxtConfig (dir, config) {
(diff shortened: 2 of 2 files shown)


danielroe:
This looks like a good PR, but Nuxt v2 is EOL and we do not plan to merge PRs or make any future releases.

If you are still using it, I would advise patching locally using `pnpm patch` or https://github.com/ds300/patch-package.

PR #30876, opened 2025-02-05
docs: improve `navigateTo` docs with clearer structure and examples

### LINKED ISSUE #29104: navigateTo doesn't work inside plugins called inside middleware
### Environment

Environment
Node Version: v20.15.0
Nuxt Version: ^3.12.3
CLI Version: 3.12.0
Package Manager: npm@10.7.0
The issue does not depend on OS

### Reproduction

Link: https://github.com/WillRy/nuxt3-middleware-issues

Just access the home page with **npm run dev**

### Describe the bug

First of all, thank you very much for the work that the entire team has been doing.

When creating a 'custom fetch' as ​​indicated in the documentation, to automatically redirect when receiving a 401 status code, we will have a problem where navigationTo is ignored in middlewares when we made a API call

**This is the custom fetch instance registered as a plugin**
```javascript

    const nuxtApp = useNuxtApp();
    const $customFetch = $fetch.create({
      baseURL: '/api',
      headers: {
        Accept: 'application/json',
      },
      async onRequest({ request, options, error }) {
      },
      async onResponse({ response }) {
        if (response.status == 401) {
          await nuxtApp.runWithContext(() => navigateTo('/login?logout=1'));
        }
      },
    });
    // Expose to useNuxtApp().$customFetch
    return {
      provide: {
        customFetch: $customFetch,
      },
    };
  });
```

**This is the custom fetch COMPOSABLE**

```javascript
import type { UseFetchOptions } from "nuxt/app";
export function $serverFetch<T>(
  url: string | (() => string),
  options: UseFetchOptions<T> = {},
) {
  return useFetch(url, {
    ...options,
    $fetch: useNuxtApp().$customFetch,
  });
}
```

And inside a global middleware, **auth.global.ts**:
```javascript

export default defineNuxtRouteMiddleware(async (to, from) => {
  
  const haveLogoutQueryString = to.fullPath.includes("logout");

  if (haveLogoutQueryString) {
    return;
  }

  const { error } = await $serverFetch("/user");

  console.log(error);
});

```

**The API call made within the middleware will return 401 and according to the 'custom fetch' interceptors, a navigateTo should be made, which is ignored.**

Thanks!

### Additional context

_No response_

### Logs

_No response_

Mini-ghost:
### 🔗 Linked issue

Resolve #29104

### 📚 Description

In this PR, I have made some improvements to the documentation for `navigateTo`:  

1. Moved the note **"Make sure to always use `await` or `return` on the result of `navigateTo` when calling it"** to a more prominent position.  
2. Added an explanation of why `navigateTo` should be returned when used inside route middleware, along with an example of incorrect usage. This might help with #29104.  
3. Provided more detailed type declarations.  
4. Adjusted the formatting of the **Parameters** section for better readability.  

    **Before**
	Before

    **After**
    After

I hope these changes make the `navigateTo` documentation clearer and easier to read. If there’s anything I might have overlooked or if you have any suggestions, please feel free to let me know. I truly appreciate your feedback.


Mini-ghost:
There might be a minor issue, the `target` option in `options.open` is mentioned in the documentation as having a default value of `_blank`, but in the type annotations, it is marked as required. Could this be a bit confusing for users?

https://github.com/nuxt/nuxt/blob/55d0e5c97c739ad525e6f4380a78ce7c15a88630/docs/3.api/3.utils/navigate-to.md?plain=1#L183-L189

https://github.com/nuxt/nuxt/blob/55d0e5c97c739ad525e6f4380a78ce7c15a88630/packages/nuxt/src/app/composables/router.ts#L105-L108

docs/3.api/3.utils/navigate-to.md
```diff
@@ -144,82 +174,58 @@ await navigateTo({ name: 'product', params: { id: 1 } })
...
+      | `width` or `innerWidth`  | `number`  | Specifies the content area's width (minimum 100 pixels), including scrollbars. |
+      | `height` or `innerHeight` | `number`  | Specifies the content area's height (minimum 100 pixels), including scrollbars. |
+      | `left` or `screenX`   | `number`  | Sets the horizontal position of the new window relative to the left edge of the screen. |
+      | `top` or `screenY`   | `number`  | Sets the vertical position of the new window relative to the top edge of the screen. |
+      | `noopener` | `boolean` | Prevents the new window from accessing the originating window via `window.opener`. |
+      | `noreferrer` | `boolean` | Prevents the Referer header from being sent and implicitly enables `noopener`. |
 
-    Refer to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) for more detailed information on the **windowFeatures** properties.
+      Refer to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) for more detailed information on the **windowFeatures** properties.
```
xjccc, line 231: Is it better change this link to `https://developer.mozilla.org/en-US/docs/Web/API/Window/open#windowfeatures`

docs/3.api/3.utils/navigate-to.md
```diff
@@ -8,16 +8,16 @@ links:
...
-::
-
 ## Usage
 
 `navigateTo` is available on both server side and client side. It can be used within the [Nuxt context](/docs/guide/going-further/nuxt-app#the-nuxt-context), or directly, to perform page navigation.
 
-::tip
-To send a redirect from a server endpoint, use [`sendRedirect`](https://h3.unjs.io/utils/response#sendredirectevent-location-code) instead.
+::warning
+Make sure to always use `await` or `return` on result of `navigateTo` when calling it.
```
huang-julien, line 16: IMO this literally pinpoint a DX issue. Users shouldn't think about "should i use `await` or `return` here ?" .

@danielroe should we do something about this in V4 ?

danielroe: we probably could (in v3) if we _require_ `defineNuxtRouteMiddleware` wrapper - because we could 'catch' the value and return it.

(basically vue-router middleware redirects via a 'returned value' but elsewhere `navigateTo` navigates via a push/server redirect.)

docs/3.api/3.utils/navigate-to.md
```diff
@@ -81,12 +97,13 @@ await navigateTo('https://nuxt.com', {
...
 ```
 
-### Using open()
+### Opening a Page in a New Tab
 
 ```vue
 <script setup lang="ts">
 // will open 'https://nuxt.com' in a new tab
 await navigateTo('https://nuxt.com', {
+  external: true,
```
danielroe, line 106: this isn't needed when used with `open`.

https://github.com/nuxt/nuxt/blob/c01050f7ba931b557ebf66c08331969152942aca/packages/nuxt/src/app/composables/router.ts#L127-L137

```suggestion
```

danielroe (approved):
beautiful! ❤️

huang-julien (approved):
:heart:

PR #30889, opened 2025-02-06
feat: upgrade `@nuxt/devtools` to v2

antfu:
https://github.com/nuxt/devtools/releases/tag/v2.0.0


danielroe (approved):

martinszeltins:
This looks lovely! @antfu Does devtools v2 support customizing the devtools using a config file? Right now all settings are erased / cleared when cookies are deleted (things like autoclose, click outside, dark mode etc.)

PR #30895, opened 2025-02-06
feat: treeshake client pages from server bundle

### LINKED ISSUE #30786: ssr: false routes are inluded in both client and server output bundle
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.3
- CLI Version:      3.20.0
- Nitro Version:    2.10.4
- Builder:          -
- User Config:      compatibilityDate, routeRules, devtools, sourcemap
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-otbzpuzh

1. Build the project using `npm run build`
2. Inspect the `.nuxt/dist/server/_nuxt/spa.js` file which includes the `<TextEditor>` component

### Describe the bug

When setting `ssr: false` for specific routes in `routeRules`, the expected behavior is for these routes to be excluded from the server bundle and executed entirely on the client side. However, the routes are still bundled into the server, defeating the purpose of ssr: false. This forces the use of `<ClientOnly>` or similar workarounds, which is not desirable.

### Additional context

My use case involves an SSR Nuxt app that includes a large number of `/admin` SPA pages. These routes are intended to run entirely on the client. While setting `ssr: false` for `/admin/**` routes works as expected in terms of rendering, Nuxt still includes all their code in the server bundle.
Currently, I am forced to wrap every SPA route's content in a `<ClientOnly>` component or use conditional checks like `if (import.meta.client)` for tree-shaking.
Also making page components as `.client.vue` didn't work for me

### Logs

```shell-script

```

Saeid-Za:
### 🔗 Linked issue

Resolves #30786

### 📚 Description

Component tree shaking was limited to `ClientOnly` usage. This PR attempts to resolve `routeRules` and, based on the final state, determines which page components should be completely removed from the server bundle.

To achieve this functionality, access to the updated `nitro.routeRules` is necessary. However, in `experimental.inlineRouteRules` mode, this configuration is [changed ](https://github.com/nuxt/nuxt/blob/4be52d341f5f6dacd8056e4a2407986bcc596234/packages/nuxt/src/pages/module.ts#L391)after the `nitro:init` hook but the change is not emitted for other listeners to be notified. As a workaround, a new (internal?) Nuxt hook has been introduced to always emit the latest available `routeRules` during the build stage.

### More Context

I believe access to the most updated state of `nitro:routeRules` is a missing feature in Nuxt, especially when authoring modules that need to be aware of their context.

Please let me know if any changes are required. Thank you in advance!


packages/schema/src/types/hooks.ts
```diff
@@ -263,6 +263,12 @@ export interface NuxtHooks {
    * @returns Promise
    */
   'nitro:config': (nitroConfig: NitroConfig) => HookResult
+  /**
+   * Called after route rules are resolved
+   * @param routeRules The route rules config to be read
+   * @returns Promise
+   */
+  'nitro:routeRules': (routeRules: { [path: string]: NitroRouteConfig }) => HookResult
```
HigherOrderLogic, line 271: I think this shouldnt be prefixed with `nitro:`, since it's not specific to Nitro, but also the Nuxt app.

Also, dont you need to add this hook to the docs also?

Saeid-Za: Thanks for the feedback 🙏
Yes, the prefix could be changed to something more relevant, do you have something in mind?
About documentation, I'm not in a position to introduce a new hook to nuxt, as the purpose of this PR was something else and this a byproduct of the solution.
This hook seems to be an internal only hook, but if you and other members think this would be a suitable addition, let me know !

danielroe (commented):
I think we don't need to do any tree-shaking of these routes in development mode, which might simplify this somewhat.

I'd also like to see if we can avoid any additional hooks in this PR.

We should also not add the build plugin in a hook, but we can pass it a context object which we update when the rules are updated.

danielroe:
why do we need an event system?

Saeid-Za:
I have updated the PR,
1. ✅️ No nested hook calling, the vite plugin is added outside any hook.
2. ✅️ Non-dev only tree-shaking.
3. ❌ This PR still is dependent on `nitro:routeRules`

## More Context

### Non-dev tree-shaking
The non-dev only checking might be implemented like this:
```ts
addBuildPlugin(TreeShakeTemplatePlugin({ sourcemap: !!nuxt.options.sourcemap.server, getComponents, routeData }), { client: false, dev: false })
```
If there is no need to tree-shake on `dev` mode, this might be a better solution, which seems a valid option IMO.

### Why I think a new hook is still needed
`updateRouteConfig` in `pages/module.ts` is responsible for updating the route-rules, which is being called asynchronously when `nuxt.options.experimental.inlineRouteRules` is activated. This means that an async communication channel is needed to wait/listen for/on it's value.
This could be the hook introduced in this PR or any other async communication channel.

Please let me know if you need any further modifications !

packages/nuxt/src/components/module.ts
```diff
@@ -210,37 +199,31 @@ export default defineNuxtModule<ComponentsOptions>({
...
+      pages: [] as NuxtPage[],
+      routeRules: {} as { [path: string]: NitroRouteConfig },
+    }
 
-    const clientDelayedComponentRuntime = await findPath(join(distDir, 'components/runtime/lazy-hydrated-component')) ?? join(distDir, 'components/runtime/lazy-hydrated-component')
+    nuxt.hook('pages:resolved', (pages) => {
+      routeData.pages = pages
+    })
+
+    nuxt.hook('nitro:routeRules', (routeRules) => {
```
harlan-zw, line 211: this is pretty much just for HMR right? as we should be able to reference the nitro route rule config from here 

if so, we may consider just using the `rollup:reload` nitro hook, which runs when the route rules are updated, it cuts out some of the complexity of adding a new hook (maintenance / docs)

packages/nuxt/src/components/plugins/tree-shake.ts
```diff
@@ -20,6 +29,8 @@ const CLIENT_ONLY_NAME_RE = /^(?:_unref\()?(?:_component_)?(?:Lazy|lazy_)?(?:cli
 
 export const TreeShakeTemplatePlugin = (options: TreeShakeTemplatePluginOptions) => createUnplugin(() => {
   const regexpMap = new WeakMap<Component[], [RegExp, RegExp, string[]]>()
+  const ruleMatcher = !import.meta.dev ? toRouteMatcher(createRadixRouter({ routes: options.routeData.routeRules })) : null
```
harlan-zw, line 32: We should probably add a nuxt/kit util for this at some point with dedupe and with the rollup:reload hook, probably out of scope here

packages/nuxt/src/components/plugins/tree-shake.ts
```diff
@@ -40,6 +51,21 @@ export const TreeShakeTemplatePlugin = (options: TreeShakeTemplatePluginOptions)
...
+      const page = !import.meta.dev && options.routeData.pages.find(p => p.file === id)
+      // A page component is being tree shaken
+      if (page) {
+        const matchedRules = ruleMatcher!.matchAll(page.path).reverse()
+        const resolvedRoute = defu({} as Record<string, any>, ...matchedRules) as NitroRouteConfig
+        if (resolvedRoute) {
+          // page is SPA
+          if (resolvedRoute.ssr === false) {
+            return {
+              code: '',
```
harlan-zw, line 63: - do we need to keep source maps for debugging?
- emitting an empty Vue SFC seems a bit strange, it feels like this should be happening at the router level, but non-blocking if its working

PR #30910, opened 2025-02-07
fix(kit): ensure nuxt is loaded from cwd rather than parent dir

### LINKED ISSUE #25019: @nuxt/kit loadNuxt function has wrong logic to find nuxt pkg in monorepo
### Environment

wsl: 2
linux: Ubuntu 20.04.6 LTS
node: v18.19.0
npm: 10.2.3
nuxt: 3.9.0
nuxi: 3.10.0
@nuxt/kit: 3.9.0
pkg-types: 1.0.3
mlly: 1.4.2

### Reproduction

I created a example project for you:
https://stackblitz.com/edit/stackblitz-starters-rzj56g?file=index.js

### Describe the bug

In the example project, the test project in the root directory is experiencing dependency confusion. When testing with `npm run dev`, `defineNuxtConfig` is not defined. I found that it can't locate the `test/node_modules/nuxt` when starting, but instead it's looking in `root/node_modules/nuxt`. This issue arose when I used a debugger in your `nuxt/kit/dits/index.mjs` `loadNuxt` function. Upon logging the `nearestNuxtPkg` variable, it showed the root project's Nuxt path.
*Seems like `resolvePackageJSON` function has something wrong.*

### Additional context

Even though I use `import`, it still encounters issues due to conflicting dependencies. However, when I remove the `root/node_modules`, the issue is resolved. I believe the dependency resolution mechanism should follow the default behavior of Node.js.

### Logs

_No response_

cjpearson:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25019

### 📚 Description

When resolving a package using a base URL with no trailing slash, it searches relative to the parent of the final path component. Since `cwd` is a directory, adding a trailing slash will treat it like a directory and search for Nuxt there first. 

There's a discussion on this topic in `mlly` (https://github.com/unjs/mlly/issues/158), but since this is the way `import.meta.resolve` works in Node, it makes sense that `mlly` has the same behavior.

(I've tested this with a Nuxt 3 app, but I'm not sure if this is the proper way to target a PR for Nuxt 3. The lines have changed, but same fix should be able to be applied to `main`)


danielroe:
Any chance you could provide a reproduction of the issue? 🙏

I tried with this unit test, placed in `packages/kit/test/load-nuxt.spec.ts`, but was unable to replicate:

(there's a `nuxt.config.ts` in `playground` but also in repo root)

```ts
import { fileURLToPath } from 'node:url'
import { describe, expect, it } from 'vitest'
import { join, normalize } from 'pathe'
import { withoutTrailingSlash } from 'ufo'
import { loadNuxt } from '../src'

const repoRoot = withoutTrailingSlash(normalize(fileURLToPath(new URL('../../../', import.meta.url))))

describe('loadNuxt', () => {
  it('respects correct directory', async () => {
    const nuxt = await loadNuxt({
      cwd: join(repoRoot, 'playground'),
    })
    expect(nuxt.options.rootDir).toBe(join(repoRoot, 'playground'))
    await nuxt.close()
  })
})
```

cjpearson:
I can try looking into a unit test. Here's a reproduction repository which shows the issue. It has a workspace with a couple of node module packages, then an application which is not part of the workspace.

https://github.com/cjpearson/nuxt-monorepo-test

When you build the application at `templates/nuxt-app`, it uses the nuxt from the root `node_modules`.

danielroe (approved):

danielroe:
Many thanks for this! Probably a lot of weird issues/edge cases could have stemmed from it.

I think we can also do something similar in a lot of other places in the code base where we pass arrays of directories to `paths` or `url` options, which might even produce a performance improvement as it could remove an iteration `mlly` has to perform.

pi0:
probably for places explicitly passing path/url/parent(s), would be best to pass a `file://` URL. OS paths in ESM resolution are ambiguous (for a URL, there is no difference between "file" and "directory")

danielroe:
@pi0 If passing a file URL, the bug reoccurs.

pi0:
Hotfix looks good (and seems only option here at least with CWD), adding arbitrary `_index.js` is also good idea, but in general I think right root cause fix is that all places of usage should pass a `file://` URL because this way mlly can worry less and be use it is really a path to file (which it's dir is relative base). without, behavior will remain undeterministic or at least less efficient (since a fs stat is needed)

danielroe:
Can `mlly` trust it if we pass URLs? (looking for best possible perf?)

pi0:
(yes, i have to check but if it doesn’t thats a bug IMO. We should trust file:// as file)

danielroe:
Okay, I'll merge this and then create a further-reaching PR that passes file URLs everywhere instead.

PR #30921, opened 2025-02-08
feat: `useMaybeAbsolutePath`

### LINKED ISSUE #30850: `useMaybeAbsolutePath` combosable for automatic `baseURL` treatment
## Problem

Nuxt 3.15 out of the box can transform `<NuxtLink>` and inline urls starting with `/` into absolute urls that include app's `baseURL`.

```
// In SFC Template
<img src="/images/foo.png" />

// In dev and production when baseURL = '/subfolder/'
<img src="/subfolder/images/foo.png />
```

The automatic transformation is not applied when using variables in template.

```
const imgSrc = '/images/foo.png';

// In SFC Template
<img :src="imgSrc" />

// In dev and production when baseURL = '/subfolder/'
<img src="/images/foo.png />
// No transformation applied!
``` 

## Workaround

One way to handle it is to manually retrieve `baseURL` from runtime config and add it to url:

```
const runtimeConfig = useRuntimeConfig();
const imgSrc = runtimeConfig.app.baseURL + 'images/foo.png';

<img :src="imgSrc" />
```

## Solution

To improve DX I think Nuxt should provide a composable `useMaybeAbsolutePath`, which returns a function that automatically adds `baseURL` to all paths that start with `/` and does not touch any other paths:

```ts
export default useMaybeAbsolutePath()
{
    const runtimeConfig = useRuntimeConfig();
    return (path: string) => {
        if (path.startsWith('/')) return runtimeConfig.app.baseURL + path.substring(1);
        else return path; // Do not edit relative paths and external links
    }
}
```

Usage:

```vue
<script lang="ts" setup>
const maybeAbsolutePath = useMaybeAbsolutePath();
const imgSrc = '/images/foo.png';
</script>

<template>
    <img :src="maybeAbsolutePath(imgSrc)" />
</template>
```

### Additional information

### Final checks

Gwynerva:
### 🔗 Linked issue

Resolves #30850 

### 📚 Description

Nuxt 3.15 out of the box can transform `<NuxtLink>` and inline urls in attributes starting with `/` into absolute urls that include app's `baseURL`. However, the automatic transformation is not applied when directly using string variables in template.

```
const imgSrc = '/images/foo.png';

// Base url won't be applied to 'src' attribute
<img :src="imgSrc" />
```

This PR adds `useMaybeAbsolutePath` for automatic `baseURL` treatment both in `setup` and for variables in `template` sections of SFC.

Usage:

```vue
<script lang="ts" setup>
// Consider the `baseURL` is `/foo/bar/`
const maybeAbsolute = useMaybeAbsolutePath();

const path1 = maybeAbsolute('qux'); // qux
const path2 = maybeAbsolute('/qux'); // /foo/bar/qux
</script>

<template>
  <div>
    Not changed:
    <ul>
      <li>{{ maybeAbsolute('/foo/bar/baz') }}</li>
      <li>{{ maybeAbsolute('baz/qux') }}</li>
      <li>{{ maybeAbsolute('https://www.google.com') }}</li>
    </ul>
  </div>
  <div>
    Changed:
    <ul>
      <li>{{ maybeAbsolute('/baz/qux') }}</li>
      <!-- /foo/bar/baz/qux -->
    </ul>
  </div>
</template>
```

Also I think it can be renamed to something less confusing like `useBaseUrlPath` and used like this:

```ts
const baseUrlPath = useBaseUrlPath();
console.log(baseUrlPath('/foo/bar'));
```


danielroe:
Thank you for this!

I added a comment in https://github.com/nuxt/nuxt/issues/30850#issuecomment-2653032092. What do you think?

PR #30935, opened 2025-02-10
docs: fix typo and improve data fetching examples

marr:
### 🔗 Linked issue

none

### 📚 Description

This change fixes a typo on the `useRequestFetch` example. The `.value` is incorrect as the `useRequestHeaders` value is not a ref. 

I only discovered the fact that _useFetch_ already proxies headers to SSR event handlers after scrolling and re-reading the documentation. A small change to the data fetching section about sending cookies helps users know they may not need to reach for `useRequestFetch` if they are already using `useFetch`.


danielroe (approved):
👌 

thank you!

PR #30948, opened 2025-02-11
feat(kit): add named layer aliases

### LINKED ISSUE #30941: Named layer aliases
it may be beneficial to allow creating aliases for layers to reference themselves - or referencing a path within a layer from the 'consuming' final Nuxt app.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/30941

### 📚 Description

This implements named layer aliases.

Auto-scanned layers (from your `~~/layers` directory) will automatically create aliases. For example, you can access your `~~/layers/test` layer via `#layers/test`.

If you want to create named layer aliases for other layers, you can specify a name in the configuration of the layer.

```ts [nuxt.config.ts]
  $meta: {
    name: 'example-layer',
  },
})
```

This will produce an alias of `#layers/example-layer` which points to your layer.

~~(This PR will fail unit tests until c12@2.0.2 is updated in Nuxt main branch.)~~


pi0:
Wondering if would be better idea to use `#layers/{name}`, as it is 1) can be more explicitly identified  and 2) avoid conflict possibilities with internals and plugins using `#.*` aliases.

atinux:
I love it 🚀

danielroe:
this issue with be resolved in v3.16.1

PR #30973, opened 2025-02-12
fix(nuxt): decode URI components in cache driver methods

### LINKED ISSUE #27722: Error "ENAMETOOLONG: name too long" occurs during pre-rendering on versions 3.11 and 3.12
### Environment

The issue persists in both versions 3.11 and 3.12.

### Version 3.11
- Operating System: `Darwin`
- Node Version:     `v18.19.0`
- Nuxt Version:     `3.11.2`
- CLI Version:      `3.12.0`
- Nitro Version:    `2.9.6`
- Package Manager:  `npm@10.2.3`
- Builder:          `-`
- User Config:      `devtools`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Version 3.12
- Operating System: `Darwin`
- Node Version:     `v18.19.0`
- Nuxt Version:     `3.12.2`
- CLI Version:      `3.12.0`
- Nitro Version:    `2.9.6`
- Package Manager:  `npm@10.2.3`
- Builder:          `-`
- User Config:      `devtools`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Steps to reproduce:

1. Create an empty Nuxt project using versions 3.11 or 3.12 (the issue does not exist in v3.10).
2. Create a `pages` directory.
3. Inside the `pages` directory, create an `index.vue` file and add content to it (it can be something simple like `<p>Test</p>`).
4. Create a `աաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաաա.vue` file in the `pages` directory and add content to it (again, it can be simple, like `<p>Test</p>`).
5. Run npm run generate.

You should receive this error during the pre-rendering process.

Screenshot 2024-06-19 at 21 01 16

### Describe the bug

I manage a multilingual blog that is integrated with a headless CMS, using slugs derived from the titles for routing to different blog pages. 

Since the content is available in multiple languages, the slugs are also multilingual, including Armenian characters. It seems that during the pre-rendering process, Nuxt attempts to store the payload files locally for caching inside the `<rootDir>/.nuxt/cache/nitro/prerender` directory. However, before storing, it encodes the file names using US-ASCII characters.

This encoding process significantly increases the length of the file names, causing them to exceed the 255-character limit for file names and I get the "ENAMETOOLONG: name too long" error.

image

This issue did not occur in versions 3.10, 3.9, and 3.8. I have not used any versions older than these.
Unfortunately, this issue has become a blocker preventing me from upgrading beyond version 3.10.

### Additional context

_No response_

### Logs

_No response_

vzkn:
### 🔗 Linked issue

Resolves #27722

### 📚 Description

When using Nuxt version 3.11 or above, pre-rendering pages with multilingual slugs (including non-ASCII characters) causes filenames to be encoded to US-ASCII. This encoding significantly increases the filename length, potentially exceeding the file system's 255-character limit, and ultimately triggers an "ENAMETOOLONG" error. This issue did not occur in version 3.10 and earlier.

To address this, I reviewed all the commits between https://github.com/nuxt/nuxt/compare/v3.10.3...v3.11.0. One commit in particular stood out: in https://github.com/nuxt/nuxt/commit/bc44dfc48404fd3aee58a662099952568ed71f58, @danielroe updated the `nitro.options._config.storage` configuration for the pre-render cache (`internal:nuxt:prerender`).

**Before:**
```ts
nitro.options._config.storage['internal:nuxt:prerender'] = { driver: 'memory' }
```

**After:**
```ts
nitro.options._config.storage = defu(nitro.options._config.storage, {
  'internal:nuxt:prerender': {
    driver: pathToFileURL(await resolvePath(join(distDir, 'core/runtime/nitro/cache-driver'))).href,
    base: resolve(nuxt.options.buildDir, 'cache/nitro/prerender')
  }
})
```

Previously, the pre-render cache was stored in memory. With the updated configuration, the cache is now stored using a `cache-driver` in the `<rootDir>/.nuxt/cache/nitro/prerender` folder.

Upon further inspection of the `cache-driver` code, I discovered that when a pathname includes non-ASCII characters, it is used directly as a filename. This can lead to the "ENAMETOOLONG" error when the filename exceeds the 255-character limit. To resolve this, I added a call to `decodeURIComponent` to decode the pathnames before they are used, ensuring that filenames remain within acceptable length limits.

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/cache-driver.js
+++ b/packages/nuxt/src/core/runtime/nitro/cache-driver.js
@@ -7,7 +7,7 @@ import lruCache from 'unstorage/drivers/lru-cache'
 /**
  * @param {string} item
  */
-const normalizeFsKey = item => item.replaceAll(':', '_')
+const normalizeFsKey = item => decodeURIComponent(item.replaceAll(':', '_'))
 
 /**
  * @param {{ base: string }} opts
@@ -20,15 +20,15 @@ export default defineDriver((opts) => {
     ...fs, // fall back to file system - only the bottom three methods are used in renderer
     async setItem (key, value, opts) {
       await Promise.all([
-        fs.setItem?.(normalizeFsKey(decodeURIComponent(key)), value, opts),
+        fs.setItem?.(normalizeFsKey(key), value, opts),
         lru.setItem?.(key, value, opts),
       ])
     },
     async hasItem (key, opts) {
-      return await lru.hasItem(key, opts) || await fs.hasItem(normalizeFsKey(decodeURIComponent(key)), opts)
+      return await lru.hasItem(key, opts) || await fs.hasItem(normalizeFsKey(key), opts)
     },
     async getItem (key, opts) {
-      return await lru.getItem(key, opts) || await fs.getItem(normalizeFsKey(decodeURIComponent(key)), opts)
+      return await lru.getItem(key, opts) || await fs.getItem(normalizeFsKey(key), opts)
     },
   }
 })
--- a/packages/nuxt/src/core/runtime/nitro/cache-driver.js
+++ b/packages/nuxt/src/core/runtime/nitro/cache-driver.js
@@ -20,15 +20,15 @@ export default defineDriver((opts) => {
     ...fs, // fall back to file system - only the bottom three methods are used in renderer
     async setItem (key, value, opts) {
       await Promise.all([
-        fs.setItem?.(normalizeFsKey(key), value, opts),
+        fs.setItem?.(normalizeFsKey(decodeURIComponent(key)), value, opts),
         lru.setItem?.(key, value, opts),
       ])
     },
     async hasItem (key, opts) {
-      return await lru.hasItem(key, opts) || await fs.hasItem(normalizeFsKey(key), opts)
+      return await lru.hasItem(key, opts) || await fs.hasItem(normalizeFsKey(decodeURIComponent(key)), opts)
     },
     async getItem (key, opts) {
-      return await lru.getItem(key, opts) || await fs.getItem(normalizeFsKey(key), opts)
+      return await lru.getItem(key, opts) || await fs.getItem(normalizeFsKey(decodeURIComponent(key)), opts)
     },
   }
 })


packages/nuxt/src/core/runtime/nitro/cache-driver.js
```diff
@@ -20,15 +20,15 @@ export default defineDriver((opts) => {
     ...fs, // fall back to file system - only the bottom three methods are used in renderer
     async setItem (key, value, opts) {
       await Promise.all([
-        fs.setItem?.(normalizeFsKey(key), value, opts),
+        fs.setItem?.(normalizeFsKey(decodeURIComponent(key)), value, opts),
```
danielroe, line 23: could this be moved to `normalizeFsKey` function?

packages/nuxt/src/core/runtime/nitro/cache-driver.js
```diff
@@ -7,7 +7,7 @@ import lruCache from 'unstorage/drivers/lru-cache'
 /**
  * @param {string} item
  */
-const normalizeFsKey = item => item.replaceAll(':', '_')
+const normalizeFsKey = item => decodeURIComponent(item.replaceAll(':', '_'))
```
danielroe, line 10: my only worry is whether windows/mac/linux filesystems will all support non-ASCII characters. (maybe, if not, we need an additional sanitisation step)

Vahagn-Zaqaryan: That's a really good point, thanks for the feedback!

I’ve given this a lot of thought, and here’s my reasoning:

1. **Initial Approach:**  
   I initially considered that users might supply virtually any string for the article slug. Since we can’t predict which characters they might use, I thought about normalizing the input by applying a hashing function (e.g., SHA-256 or a lighter alternative). This would ensure that the output is consistent and only contains allowed characters.

2. **Reevaluation:**  
   After looking at the bigger picture, I realized that the files created by the `cache-driver` during the `nuxt generate` command are temporary, they’re only used for generation. What really matters is that in the final output, under the `.output` folder, users see a sub-folder with the original name they expect.

For example:

- **During Generation:**  
  `<rootDir>/.nuxt/cache/nitro/prerender`  
  Cache Folder

- **Final Output:**  
  `<rootDir>/.output/public`  
  Output Folder

Let me know if you see any issues with this approach or if there’s another angle I might have missed.

danielroe: what approach are you suggesting?

Vahagn-Zaqaryan: I suggest we stick with the `decodeURIComponent` approach to avoid any performance overhead from hashing. Since the temporary cache file and the final output folder share the same name, any system that doesn't support non-ASCII characters will fail to create both. If I implement a solution for the temporary cache file, then it will fail when creating the folder.

danielroe (approved):

PR #30986, opened 2025-02-13
feat(kit,nuxt,vite): `directoryToURL` to normalise paths

danielroe:
### 📚 Description

this should improve perf with import path resolution because we can avoid an extra fs stat call and accurately communicate that paths like `nuxt.options.modulesDir` are directories and not paths.

### DIFF
--- a/packages/kit/src/internal/esm.ts
+++ b/packages/kit/src/internal/esm.ts
@@ -33,7 +33,7 @@ export async function tryResolveModule (id: string, url: string | string[] | URL
 }
 
 export function resolveModule (id: string, options?: ResolveModuleOptions) {
-  return resolvePathSync(id, { url: options?.url ?? options?.url ?? [importMetaURL] })
+  return resolvePathSync(id, { url: options?.url ?? options?.paths ?? [importMetaURL] })
 }
 
 export interface ImportModuleOptions extends ResolveModuleOptions {
--- a/packages/kit/src/internal/esm.ts
+++ b/packages/kit/src/internal/esm.ts
@@ -2,6 +2,7 @@ import { fileURLToPath, pathToFileURL } from 'node:url'
 import { interopDefault, resolvePath, resolvePathSync } from 'mlly'
 import { createJiti } from 'jiti'
 import { captureStackTrace } from 'errx'
+import { normalize } from 'pathe'
 
 export interface ResolveModuleOptions {
   /** @deprecated use `url` with URLs pointing at a file - never a directory */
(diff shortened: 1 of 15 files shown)


packages/kit/src/internal/esm.ts
```diff
@@ -4,16 +4,27 @@ import { createJiti } from 'jiti'
 import { captureStackTrace } from 'errx'
 
 export interface ResolveModuleOptions {
+  /** @deprecated use `url` with URLs pointing at a file - never a directory */
   paths?: string | string[]
+  url?: URL | URL[]
 }
 
+export function directoryToParentURL (directory: string) {
+  return new URL('./index.js', pathToFileURL(directory.replace(/\/?$/, '/')))
```
pi0: I think this could be simplified. We don't need a `URL` instance but simple string starting with `file://`.

If this util can also trust input is a dir path, we can use `filePathToURL(join(dir, '_index.js'))` (which returns a string).

packages/kit/src/internal/esm.ts
```diff
@@ -2,18 +2,30 @@ import { fileURLToPath, pathToFileURL } from 'node:url'
...
+  /** @deprecated use `url` with URLs pointing at a file - never a directory */
   paths?: string | string[]
+  url?: URL | URL[]
 }
 
+export function directoryToParentURL (dir: string): URL {
+  return pathToFileURL(join(dir, '_index.js'))
+}
+
+const importMetaURL = new URL(import.meta.url)
```
pi0, line 17: I think this can be simplified too

danielroe: I was sort of imagining `mlly` could short circuit something if it was a URL instance, but if it doesn't matter....

pi0: it does by checking `startsWith('file://')` (also lots of old logic there, `URL` constructor is overhead most of the times)

pi0:
please hold this on little bit. (it seems was my bad, file:// can mean also directory if ending with trailing slash. trying some changes on mlly)

packages/kit/src/internal/esm.ts
```diff
@@ -2,9 +2,16 @@ import { fileURLToPath, pathToFileURL } from 'node:url'
...
 import { captureStackTrace } from 'errx'
+import { join } from 'pathe'
 
 export interface ResolveModuleOptions {
+  /** @deprecated use `url` with URLs pointing at a file - never a directory */
   paths?: string | string[]
+  url?: URL | URL[]
+}
+
+export function directoryToParentURL (dir: string): URL {
```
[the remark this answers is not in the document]
pi0: sure we can add it later

PR #31005, opened 2025-02-16
docs: update nuxt context section in 6.nuxt-app.md

idorenyinudoh:
### 📚 Description

specify the exact middleware types that the nuxt context is accessible in (global route middleware and named route middleware).

i think this is necessary to note because attempting to use a composable in an anonymous middleware throws the following error — A composable that requires access to the Nuxt instance was called outside of a plugin, Nuxt hook, Nuxt middleware, or Vue setup function.

hence, the types of middleware where the nuxt context is accessible are the global route middleware and named route middleware.


danielroe:
If you wrap it in `defineNuxtRouteMiddleware` it should also work.

docs/2.guide/3.going-further/6.nuxt-app.md
```diff
@@ -23,7 +23,7 @@ Jump over the `NuxtApp` interface documentation.
 
 Many composables and utilities, both built-in and user-made, may require access to the Nuxt instance. This doesn't exist everywhere on your application, because a fresh instance is created on every request.
 
-Currently, the Nuxt context is only accessible in [plugins](/docs/guide/directory-structure/plugins), [Nuxt hooks](/docs/guide/going-further/hooks), [Nuxt middleware](/docs/guide/directory-structure/middleware), and [setup functions](https://vuejs.org/api/composition-api-setup.html) (in pages and components).
+Currently, the Nuxt context is only accessible in [plugins](/docs/guide/directory-structure/plugins), [Nuxt hooks](/docs/guide/going-further/hooks), [Nuxt middleware](/docs/guide/directory-structure/middleware) (in named and global route middleware), and [setup functions](https://vuejs.org/api/composition-api-setup.html) (in pages and components).
```
danielroe, line 26:
```suggestion
Currently, the Nuxt context is only accessible in [plugins](/docs/guide/directory-structure/plugins), [Nuxt hooks](/docs/guide/going-further/hooks), [Nuxt middleware](/docs/guide/directory-structure/middleware) (if wrapped in `defineNuxtRouteMiddleware`), and [setup functions](https://vuejs.org/api/composition-api-setup.html) (in pages and components).
```

danielroe (approved):
Thank you! ❤️

idorenyinudoh:
> If you wrap it in `defineNuxtRouteMiddleware` it should also work.

just to be sure, are you saying i can have my anonymous middleware like this?
```vue
<script setup lang="ts">
definePageMeta({
  middleware: [
    function (to, from) {
      defineNuxtRouteMiddleware (
        // Custom inline middleware
      )
    },
  ],
});
</script>

PR #31035, opened 2025-02-18
fix(kit): expose Nuxt module dependency errors

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt-modules/og-image/issues/313,
https://github.com/harlan-zw/nuxt-seo/issues/390

Possibly: https://github.com/nuxt-modules/i18n/issues/3289, https://github.com/nuxt-modules/i18n/issues/3318

### 📚 Description

When Nuxt loads a module instance it attempts several paths, these paths predictably will throw errors as they may not exist. These errors are hidden from end users as they're just noise, however, the logic to hide the errors is too aggressive and is stopping some errors from within the module itself from bubbling up.

This makes it difficult for module authors to debug issues, specifically when the issues popping up relate to using specific package managers on specific environments (i.e NPM on Mac).

This issue can be replicated by adding a dependency (e.g. `nuxt-og-image`) and adding a missing dependency import to the `module.mjs` entry

```ts
import missingDep from 'missing-dep'
```

### DIFF
--- a/packages/kit/src/module/install.ts
+++ b/packages/kit/src/module/install.ts
@@ -129,6 +129,7 @@ export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, n
           // i.e module is importing a module that is missing
           if (!module || module.includes(nuxtModule as string)) {
             continue
+          }
         }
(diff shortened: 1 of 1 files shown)


danielroe (approved):
❤️

PR #31037, opened 2025-02-18
perf(kit): optimize Nuxt module resolving

harlan-zw:
### 📚 Description

When Nuxt loads modules using the `modules` config key in `nuxt.config`, it's quite forgiving. It resolves aliases, and relative paths and will check subpath exports of `/nuxt` and `/module` from the string path.

This loading strategy is good DX but is not optimized for how most Nuxt modules are published and loaded.

The [nuxt-module-builder](https://github.com/nuxt/module-builder) does not suggest subpath exports, instead, it only adds a main entry. Meaning that modules published the official way, are being loaded inefficiently as subpath exports are being checked first.

This results in an excessive waste of time trying to resolve paths that don't exist.

To solve this is quite simple, we just reorder the path resolve order. We can also avoid passing non-existent modules to jiti to import.

**:fire: Benchmarks**

https://github.com/harlan-zw/unhead.unjs.io

Across 5 runs, an average of 4,961.4 ms was wasted trying to resolve non-existent module paths. Switching the path order in this PR wasted 0ms.

This resulted in:
- `nuxi prepare` 8 seconds -> 3 seconds (62% faster)
- `nuxi dev` 20 seconds -> 15 seconds (25% faster)


packages/kit/src/module/install.ts
```diff
@@ -94,9 +96,9 @@ export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, n
       nuxtModule = resolve(nuxt.options.rootDir, nuxtModule)
     }
 
-    paths.add(join(nuxtModule, 'nuxt'))
-    paths.add(join(nuxtModule, 'module'))
     paths.add(nuxtModule)
+    paths.add(join(nuxtModule, 'module'))
```
harlan-zw, line 100: `module` subpath export seems more likely than `nuxt`

packages/kit/src/module/install.ts
```diff
@@ -106,9 +108,17 @@ export async function loadNuxtModuleInstance (nuxtModule: string | NuxtModule, n
             url: nuxt.options.modulesDir.map(m => directoryToURL(m.replace(/\/node_modules\/?$/, '/'))),
             extensions: nuxt.options.extensions,
           })
-
-        nuxtModule = await jiti.import(src, { default: true }) as NuxtModule
         resolvedModulePath = fileURLToPath(new URL(src))
+        if (!existsSync(resolvedModulePath)) {
```
harlan-zw, line 112: i'm presuming we save some performance here by not passing in a non-existent file to `jiti.import`

danielroe (approved):

PR #31039, opened 2025-02-19
feat(nuxt): nuxt link add whitelist

### LINKED ISSUE #20500: Add a whitelist for NuxtLink
### Describe the feature

Hello,

I'm currently building 2 websites with Nuxt and Nuxt Content.

The first one will be with the domain [le-classement.fr](https://le-classement.fr) and the second one with the domain [annuaire.le-classement.fr](https://annuaire.le-classement.fr).

I'm adding many links from the first one to the second one but automatically, NuxtLink add `noopener and noreferrer`. My issue is that I want to know about the referrer but I don't want to change this behavior one link by one.

I was thinking about a white list (configured from the `nuxt.config.ts` or from the `app.config.ts`). If the link contains a domain from the whitelist, the `noreferrer` rel is not set.

This could be useful also for Studio in order to allow content creator manage referrer on markdown links.

### Additional information

### Final checks

xjccc:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/20500

### 📚 Description

Add `NuxtLinkOptions.whitelist` to remove default rel `noreferrer`

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -487,6 +487,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
             isExternal: isExternal.value || hasTarget.value,
             isActive: false,
             isExactActive: false,
+            isInWhitelist: isInWhitelist.value,
           })
         }
 
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -104,7 +104,7 @@ export interface NuxtLinkOptions extends
    * A white list of domain that should be not set noreferrer.
    * @example ['nuxt.com', 'nuxtjs.org']
    */
-  whiteList?: string[]
+  whitelist?: string[]
 }
 
 /* @__NO_SIDE_EFFECTS__ */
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
TheAlexLichter, line 1: Could we possibly expose a computed property indicating if the URL is part of the whitelist or not to the scoped slot to ensure full flexibility in terms of customization?

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -421,6 +426,20 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+            return whitelist.some(domain => urlObj.hostname.endsWith(domain))
+          } catch (e) {
+            console.error(e)
+            return false
+          }
+        }
+        let defaultExternalRel = 'noopener noreferrer'
+        if (isUrlInWhitelist(href.value, options.whitelist)) {
+          defaultExternalRel = 'noopener'
+        }
```
Mini-ghost, line 442: Would using `computed` to cache the calculated result be a more performance-friendly approach?

xjccc: Is it right for this?❤

danielroe (commented):
I feel this might be a feature we don't need in the built-in `<NuxtLink>`. Is there any way this can be added in a composable way? I imagine not all users would need this...

xjccc:
> I feel this might be a feature we don't need in the built-in `<NuxtLink>`. Is there any way this can be added in a composable way? I imagine not all users would need this...

Okay. It's a good idea. I will close this. And do you have some use case? I could try for this. Thank you ❤

PR #31081, opened 2025-02-21
fix(nuxt): showError on island page error

### LINKED ISSUE #30938: `showError` doesn't work properly in server pages/components
### Environment

```
- Operating System: Linux
- Node Version:     v22.5.1
- Nuxt Version:     3.15.4
- CLI Version:      3.21.1
- Nitro Version:    2.10.4
- Package Manager:  pnpm@9.15.4
- Builder:          -
- User Config:      modules, site, runtimeConfig, app, devtools, css, future, compatibilityDate, eslint, openFetch, fonts, colorMode, components
- Runtime Modules:  @nuxt/ui@3.0.0-alpha.12, @nuxt/eslint@1.0.1, @nuxt/test-utils/module@3.15.4, @nuxtjs/seo@2.1.1, @nuxt/scripts@0.9.5, @formkit/auto-animate/nuxt@0.8.2, nuxt-open-fetch@0.9.5, @nuxt/image@1.9.0
- Build Modules:    -
```

But based on my tests I can reproduce this bug in pure nuxt.

### Reproduction

https://stackblitz.com/edit/github-pej9nmvf?file=pages%2Ferror-server.server.vue

It has `error-client.vue` - example of displaying error in normal page and `error-server.server.vue` - example that doesn't work

### Describe the bug

Using [`showError`](https://nuxt.com/docs/api/utils/show-error) or [`throw createError`](https://nuxt.com/docs/api/utils/create-error) doesn't properly work in Nuxt's server components (or just pages). Instead of properly displaying `error.vue` page, I get `hydration error`.

I'm not sure if this is a bug or just a missing feature, but I couldn't find any information about it.

### Additional context

Thanks to @kubaryt for noticing this issue at first.

### Logs

```shell-script
[Vue warn]: Hydration node mismatch:
- rendered on server: 
<div>
 <empty string> 
- expected on client: Symbol("v-fgt") 
  at <NuxtIsland name="page:error-server" lazy=false ref=Ref< null >  ... > 
  at <ErrorServer onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< undefined > > 
  at <Anonymous key="/error-server" vnode= 
Object { __v_isVNode: true, __v_skip: true, type: {…}, props: {…}, key: null, ref: {…}, scopeId: null, slotScopeIds: null, children: null, component: null, … }
 route= 
Object { fullPath: "/error-server", hash: "", query: {}, name: "error-server", path: "/error-server", params: {}, matched: (1) […], meta: Proxy, redirectedFrom: undefined, href: "/error-server" }
  ... > 
  at <RouterView name=undefined route=undefined > 
  at <NuxtPage > 
  at <App key=4 > 
  at <NuxtRoot> runtime-core.esm-bundler.js:51:12
Hydration completed but contains mismatches.
```

huang-julien:
### 🔗 Linked issue

fix #30938 

### 📚 Description

This PR makes island pages call `showError` when an error happens


packages/nuxt/src/components/runtime/server-component.ts
```diff
@@ -46,17 +48,19 @@ export const createIslandPage = (name: string) => {
...
-
       return () => {
         return h('div', [
           h(NuxtIsland, {
             name: `page:${name}`,
             lazy: props.lazy,
             ref: islandRef,
             context: { url: path },
+            onError: (e) => {
+              nuxtApp.runWithContext(() => showError(e))
```
danielroe, line 62: should we do this only on the client-side?

huang-julien: we need to do it on both sides, otherwise a blank page will show up

danielroe: and just to be clear, it doesn't render an error page within the 'island' slot - it fully shows the error, right?

danielroe (approved):

packages/nuxt/src/components/runtime/server-component.ts
huang-julien: Yes, do you prefer to throw the error server side instead ?

danielroe (approved):

PR #31090, opened 2025-02-22
feat(nuxt): allow specifying glob patterns for `pages`

### LINKED ISSUE #31071: Make page scan pattern configurable
Context: https://github.com/nuxt/nuxt/issues/28894#issuecomment-2674247626

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31071

### 📚 Description

This adds configurability for the pattern scanned for pages in the `pages/` directory, for example:

```ts
  pages: {
    pattern: ['**/*.vue', '!**/*.spec.*'],
  },
})
```

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -4,7 +4,7 @@ import { addBuildPlugin, addComponent, addPlugin, addTemplate, addTypeTemplate, 
 import { dirname, join, relative, resolve } from 'pathe'
 import { genImport, genObjectFromRawEntries, genString } from 'knitwork'
 import { joinURL } from 'ufo'
-import type { NuxtPage } from 'nuxt/schema'
+import type { Nuxt, NuxtPage } from 'nuxt/schema'
 import { createRoutesContext } from 'unplugin-vue-router'
 import { resolveOptions } from 'unplugin-vue-router/options'
 import type { EditableTreeNode, Options as TypedRouterOptions } from 'unplugin-vue-router'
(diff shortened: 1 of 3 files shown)


TheAlexLichter (approved):

packages/nuxt/src/pages/module.ts
```diff
@@ -86,9 +94,9 @@ export default defineNuxtModule({
 
       return false
     }
-    nuxt.options.pages = await isPagesEnabled()
```
BobbieGoede, line 89: Not setting `nuxt.options.pages` leaves it `undefined` 🤔 is that intended?

danielroe: you shouldn't need to set it.

undefined should mean it auto detects

will check - is this throwing an error?

BobbieGoede: In the i18n module we're using it to check whether routing/pages are set up during the build step, but I suppose that's not really the intended use of `nuxt.options.pages`.

Will check if we need it or if there's a different way to detect it, not sure if there are other modules using it in the same way.

TheAlexLichter: Now that it isn't a plain bool you can check `nuxt.options.pages` ("legacy" if manually set) and `nuxt.options.pages.enabled` I suppose 🤔

BobbieGoede: > Now that it isn't a plain bool you can check `nuxt.options.pages` ("legacy" if manually set) and `nuxt.options.pages.enabled` I suppose 🤔

The change in this PR leaves the `nuxt.options.pages` untouched, previously it would be overwritten with the value returned by `isPagesEnabled()`. So modules will only know if/when users have manually set the pages options, not whether a pages dir has been detected and enabled router setup.

danielroe: this pr should set options.pages.enabled

TheAlexLichter: image

It should be detectable through the `enabled` property 🤔

BobbieGoede: Check out https://github.com/nuxt/nuxt/pull/31101, I tried to make it more consistent with the previous behavior and added tests to check the `nuxt.options.pages` value.

PR #31092, opened 2025-02-23
feat(nuxt): support `appLayout` in route rules

### LINKED ISSUE #26615: option to set Layout in routeRules
### Describe the feature

Hey, it's my first issue, so hope it's fine, if something is off please let me know

The middleware setting in routeRules on Nuxt 3.11 was awesome, so why not doing the same thing with layouts?
Let's pretend we want the same layout for a /app path and all it's children, as of now this would probably result in quite some repeated definePageMeta in each subpage from my understanding. Why not allowing the same thing in routeRules?
It would allow nice grouping for layouts to multiple pages.

I would also be willing to implement it, or help implementing it, mysefl, but I struggle navigating the Nuxt/Nitro (I don't know where exactly the implementation would be).

### Additional information

### Final checks

mukundshah:
Here's the refined PR description reflecting the suggested middleware workaround and your custom module approach:  

---

### 🔗 Linked issue  
closes #26615  

### 📚 Description  

This PR adds support for setting layouts via route rules using the `appLayout` property. Developers can now define layouts based on route patterns directly in the Nuxt config, offering a more centralized and declarative way to manage layouts across the application.  

#### Example usage:  
```ts
  routeRules: {
    '/admin': { appLayout: 'admin' },
    '/dashboard/**': { appLayout: 'dashboard' }
  }
})
```

#### Prior Workaround:  
Before this, the suggested workaround was to use a middleware that dynamically assigned layouts based on the route. However, I opted for a more structured approach using a custom Nuxt module. 

```ts
import type { RouterContext } from 'rou3'

import { defu } from 'defu'
import { defineNuxtModule } from 'nuxt/kit'
import { createRouter as _createRouter, addRoute, findAllRoutes } from 'rou3'

const declaration = `
import type { LayoutKey } from '#build/types/layouts'
declare module "nitropack" {
  interface NitroRouteConfig {
    layout?: LayoutKey | false
  }
}`

const createRouter = <T extends Record<string, any> = Record<string, string>>(routes: string[] | Record<string, T>): RouterContext<T> => {
  const router = _createRouter<T>()
  if (Array.isArray(routes)) {
    for (const route of routes) {
      addRoute(router, 'GET', route, { path: route } as unknown as T)
    }
  } else {
    for (const [route, data] of Object.entries(routes)) {
      addRoute(router, 'GET', route, data)
    }
  }
  return router
}

  meta: {
    name: 'route-rules-layout',
  },
  async setup(options, nuxt) {
    const router = createRouter(nuxt.options.routeRules ?? {})

    const getRules = (url: string) => {
      return defu(
        {},
        ...findAllRoutes(router, 'GET', url)
          .map(route => route.data)
          .reverse(),
      ) as Record<string, any>
    }

    nuxt.hook('prepare:types', ({ declarations }) => {
      declarations.push(declaration)
    })

    nuxt.hook('pages:resolved', (pages) => {
      pages.forEach((page) => {
        const rules = getRules(page.path)
        if (rules?.layout) {
          page.meta ||= {}
          page.meta.layout ??= rules.layout
        }
      })
    })
  },
})
```
With this PR, `appLayout` eliminates the need for either approach, providing native support for layout assignment through route rules.


packages/nuxt/src/pages/module.ts
```diff
@@ -464,6 +464,19 @@ export default defineNuxtModule({
...
+      const routeRulesMatcher = toRouteMatcher(createRadixRouter({ routes: nuxt.options.routeRules }))
+
+      pages.forEach((page) => {
+        const rules = defu({}, ...routeRulesMatcher.matchAll(page.path).reverse()) as Record<string, any>
+        if (rules?.appLayout) {
+          page.meta ||= {}
+          page.meta.layout ??= rules.appLayout
+        }
+      })
+    })
```
danielroe: we probably need to do this at runtime, like this:

https://github.com/nuxt/nuxt/blob/1505fb5449f69960d90d78bdae841af719ac8036/packages/nuxt/src/pages/runtime/plugins/router.ts#L201-L213

mukundshah: Why though? What am I missing here? 😅

danielroe: only because routeRules can be more specific. for example, we could have a catchall or dynamic route and only apply a layout to some slugs

there are also aliases to consider

but if all appLayout rules are specific enough to apply to just a single page/route then doing it at build time is preferable

(but if not we also need some runtime code)

maybe a hybrid approach

mukundshah: I am still confused. 😅

Right now, this is in my `nuxt.config`

```ts
routeRules: {
  '/**': { layout: 'app', robots: false, auth: { required: true } },
  '/auth/login': { robots: true },
  '/auth/**': { layout: 'auth', robots: false, auth: { required: false, redirectIfLoggedIn: '/dashboard' } },
  '/': { layout: 'site', robots: true, auth: { required: false } },
  '/terms': { layout: 'site', robots: true, auth: { required: false } },
  '/privacy': { layout: 'site', robots: true, auth: { required: false } },
  '/contact': { layout: 'site', robots: true, auth: { required: false } },
  '/about': { layout: 'site', robots: true, auth: { required: false } },
},
```

`layout` and `auth` is from my local modules.`robots` is from `@nuxtjs/seo`. And both of my local module injects the meta by matching the path of page against the matcher created from route rules.

Note: In the PR, i used `appLayout` to keep it consistent with `appMiddleware`.

Can you please help me out here? When this approach can be insufficient? Probably with an example.

danielroe: Imagine you have a page called `/[slug].vue` and a route rule which is `/auth` and another which is `/about`. It is not sufficient to set the layout in the meta field of `/:slug()` which could be used to render either of those routes.

mukundshah: Oh, that means, it will conflict with each other, if not done at runtime.

I will update it accordingly now.

mukundshah: @danielroe I need a little bit of help here. I tried doing it from `router.beforeEach`, but 

```ts
if (isAppManifestEnabled) {
  const routeRules = await nuxtApp.runWithContext(() => getRouteRules({ path: to.path }))

  console.log('routeRules', routeRules)
  
  to.meta.layout = routeRules.appLayout
}
```

`routeRules` is empty on client side. It works on SSR, but then the layout changes again on hydration, and I get `Hydration Mismatch`.

What am I missing here?

Also, what does this part do:

https://github.com/nuxt/nuxt/blob/ce727bc6fa40af199b7ebb53b5fc2e813e3d7c0a/packages/nuxt/src/pages/runtime/plugins/router.ts#L185-L187

mukundshah:
```ts
if (isAppManifestEnabled) {
  const routeRules = await nuxtApp.runWithContext(() => getRouteRules({ path: to.path }))

  console.log('routeRules', routeRules)

  if(routeRules.appLayout !== undefined) {
    await nuxtApp.runWithContext(() => setPageLayout(routeRules.appLayout))
  }
}
```
Using `setPageLayout` fixed the hydration mismatch. 

But, why the `routeRules` come empty on client side navigation?

danielroe:
thank you very much for your work on this ❤️

PR #31094, opened 2025-02-23
perf(nuxt): return island response earlier

huang-julien:
### 📚 Description

This PR makes islands reponses in the renderer returned just a bit earlier when possible

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -430,51 +430,7 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
     }
   }
 
-  if (!NO_SCRIPTS && !isRenderingIsland) {
-    // 3. Resource Hints
-    // TODO: add priorities based on Capo
-    head.push({
-      link: getPreloadLinks(ssrContext, renderer.rendererContext) as Link[],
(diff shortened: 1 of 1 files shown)


danielroe (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -517,6 +473,50 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
...
+        // if we are rendering script tag payloads that import an async payload
+        // we need to ensure this resolves before executing the Nuxt entry
+        tagPosition: (_PAYLOAD_EXTRACTION && !process.env.NUXT_JSON_PAYLOADS) ? 'bodyClose' : 'head',
+        crossorigin: '',
+      })),
+    }, headEntryOptions)
+  }
+
+  // remove certain tags for nuxt islands
+  const { headTags, bodyTags, bodyTagsOpen, htmlAttrs, bodyAttrs } = await renderSSRHead(head, renderSSRHeadOptions)
```
danielroe, line 518: @huang-julien is there a chance we handled this incorrectly? are the tags correctly rendered for islands?

huang-julien: what are you thinking about ? :eyes: Are you talking about the comment ?

PR #31101, opened 2025-02-24
fix(nuxt): set `nuxt.options.pages` to detected configuration

BobbieGoede:
### 🔗 Linked issue
* https://github.com/nuxt/nuxt/pull/31090

### 📚 Description
Adds tests to check whether the `nuxt.options.pages` property is overwritten with detected values. I'm sure there is a better way to test this, but this was the quickest and easiest way for me to add them.

This has a change to overwrite the `nuxt.options.pages` property, the tests still fail with (and without) this change, please check if the test cases cover the expected behavior.

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -47,11 +47,12 @@ export default defineNuxtModule({
     configKey: 'pages',
   },
   defaults: nuxt => ({
-    enabled: undefined as undefined | boolean,
+    enabled: typeof nuxt.options.pages === 'boolean' ? nuxt.options.pages : undefined as undefined | boolean,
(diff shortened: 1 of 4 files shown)


danielroe:
You might need to set `ready: true` when running `loadNuxt`.

packages/nuxt/src/pages/module.ts
```diff
@@ -95,6 +95,8 @@ export default defineNuxtModule({
       return false
     }
     options.enabled = await isPagesEnabled()
+    // to be reviewed
+    nuxt.options.pages = options
```
danielroe, line 99: if we assign the whole object (with array of patterns), we might find that defu keeps inflating the array with duplicate patterns - we can deduplicate above.

PR #31104, opened 2025-02-24
fix(nuxt): ensure `NuxtLink` `default` slot is typed

### LINKED ISSUE #31053: NuxtLink: slot typing missing
### Environment

- Operating System: Linux
- Node Version:     v22.12.0
- Nuxt Version:     3.15.4
- CLI Version:      3.22.2
- Nitro Version:    2.10.4
- Package Manager:  npm@10.9.0
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

NuxtLink: slot typing missing 

Image

Image

### Describe the bug

-

### Additional context

_No response_

### Logs

```shell-script

```

94726:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31053

### 📚 Description

This adds the `default` slot type of `<RouterLink>` to `<NuxtLink>`.
Other than that I'm not really sure how to add a test here as vue's `h` surprisingly doesn't type slots 🤷
And if there's an easier way to type the slots, I'd love to know :)


danielroe (commented):
this looks great - but note that we add additional props:

https://github.com/nuxt/nuxt/blob/082565b6c11ac233e1acab63336d4d5947cbfaf1/packages/nuxt/src/app/components/nuxt-link.ts#L444-L468

94726:
> this looks great - but note that we add additional props:
> 
> 
> 
> https://github.com/nuxt/nuxt/blob/082565b6c11ac233e1acab63336d4d5947cbfaf1/packages/nuxt/src/app/components/nuxt-link.ts#L444-L468

I see, but this isn't always the case right? At https://github.com/nuxt/nuxt/blob/082565b6c11ac233e1acab63336d4d5947cbfaf1/packages/nuxt/src/app/components/nuxt-link.ts#L414-L418 no extra props are added.

As the extra props are only there if custom is set, I could try making it generic, only adding the props on that condition. Would that make sense?

danielroe:
That would be great!

danielroe (approved):
very nice 👌

PR #31114, opened 2025-02-25
docs: explain why headers not forwarded when using `$fetch` on the server

### LINKED ISSUE #27996: Defining `$fetch` within `useFetch()` doesnt pass cookies on SSR
### Environment

- Operating System: Darwin
- Node Version:     v21.6.1
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.6
- Package Manager:  bun@1.1.13
- Builder:          -
- User Config:      future, extends, modules, runtimeConfig, routeRules, i18n
- Runtime Modules:  @nuxt/ui@^2.17.0
- Build Modules:    -

### Reproduction

https://github.com/MickL/nuxt-bug-use-fetch-custom-fetch

- Run `bun dev` to start Nuxt app on port 3000 and Nitro app on port 3001
- Press button "set cookie"
- Reload page 1: Cookie is not set
- Navigate to page 2, back to page 1: Cookie is set

### Describe the bug

When setting `$fetch` param within `useFetch()` options Nuxt does not forward cookies on SSR:

```ts
await useFetch(url, {
    ...options,
    $fetch, // Or `$fetch: useNuxtApp().$customFetch`
  });
```

On the server (separate Nitro / H3 app) the cookie is undefined:

```ts
export default defineEventHandler(async (event) => {
   console.log(getCookie(event, 'auth_session'));
});
```

This happens only on SSR. When the Nuxt app is already loaded in the browser and navigating between pages everything works as expected.

### Additional context

Potentially this is also an issue of using routeRules proxy?

Mini-ghost:
### 🔗 Linked issue

Resolve #27996, 

### 📚 Description

Many users find it confusing that `$fetch` does not forward headers, even for internal requests. The reason for this has already been explained in [Data Fetching #Passing Headers and Cookies](https://nuxt.com/docs/getting-started/data-fetching#passing-headers-and-cookies).  

However, based on my experience, when encountering this issue, users are more likely to check the documentation for [useAsyncData](https://nuxt.com/docs/api/composables/use-async-data) and [$fetch](https://nuxt.com/docs/api/utils/dollarfetch) instead. Therefore, this PR adds a note about the limitation that `$fetch` does not forward headers on the server to these pages.  

### Checklist:  
- [x] Data Fetching
- [x] `$fetch`
- [ ] ~`useAsyncData`~

Additionally, the current documentation explains that this behavior is due to `$fetch` being used for internal server requests. However, in various issues, it has been mentioned that this is a security-related decision. So in this PR, I’ve adjusted the explanation to emphasize that the restriction is primarily for security reasons.  

I’m not entirely sure if this adjustment is appropriate, please let me know your thoughts. I’d really appreciate your feedback!

### DIFF
--- a/docs/1.getting-started/6.data-fetching.md
+++ b/docs/1.getting-started/6.data-fetching.md
@@ -515,9 +515,9 @@ For finer control, the `status` variable can be:
 
 When we call `$fetch` in the browser, user headers like `cookie` will be directly sent to the API.
 
-Normally, during server-side-rendering, since the `$fetch` request takes place 'internally' within the server, it wouldn't include the user's browser cookies, nor pass on cookies from the fetch response.
+Normally, during server-side-rendering, due to security considerations, the `$fetch` wouldn't include the user's browser cookies, nor pass on cookies from the fetch response.
 
-However, when calling `useFetch` on the server, Nuxt will use [`useRequestFetch`](/docs/api/composables/use-request-fetch) to proxy headers and cookies (with the exception of headers not meant to be forwarded, like `host`).
+However, when calling `useFetch` with a relative URL on the server, Nuxt will use [`useRequestFetch`](/docs/api/composables/use-request-fetch) to proxy headers and cookies (with the exception of headers not meant to be forwarded, like `host`).
 
 ### Pass Cookies From Server-side API Calls on SSR Response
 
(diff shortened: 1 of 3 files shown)


Mini-ghost:
I'm not sure if this needs to be mentioned in `useAsyncData`, as it is not directly related to it.

danielroe (approved):

danielroe:
This looks good - thank you!

PR #31131, opened 2025-02-25
chore: re-organize runtime/nitro directory

huang-julien:
### 📚 Description

As the project grows, the nitro part of the repository contains more and more files.

This PR gives a try to a re-organization of `src/runtime/nitro` in Nuxt. This would make contributions easier to make, understand and review.

````bash
|- nitro
  |- handlers --> contains all handlers of nuxt-nitro
  |- middleware
  |- plugins --> contains all plugins for nitro
  |- utils --> Some are added to auto-importsa
````

This is a first step as `renderer.ts` needs some refactor too.

### CODE REVIEW

File: packages/nuxt/src/core/runtime/nitro/utils/cache.ts [hunk 1]
```diff
@@ -0,0 +1,24 @@
+import { useStorage } from 'nitro/runtime'
```
huang-julien, line 1: should we move this file in a subdir `prerender` with `cache-driver` ?

packages/nuxt/src/core/runtime/nitro/utils/build-files.ts
```diff
@@ -0,0 +1,119 @@
+import {
```
huang-julien, line 1: not such a huge fan of the name i gave to this file. Any suggestions ?

danielroe (approved):

PR #31142, opened 2025-02-26
docs: add shared config recipe to module authors guide

### LINKED ISSUE #18762: Allow using `useAppConfig` in modules
### Describe the feature

Hi :wave: , 
It would be nice to be able to use `useAppConfig` within modules since this can be used by modules to get non-stringifyable data in plugins. 

Currently there's like a dependency loop when we import an `app.config.ts` with 

````ts
import { defineAppConfig } from "nuxt/app"
export default defineAppConfig({ /* ..... */  })
````
if we don't specify the import, the import will fail since `app.config.ts` has not been transformed.

using `useAppConfig()` in modules triggers the same issue since it need to be imported from `nuxt/app`

### Additional information

### Final checks

IsraelOrtuno:
### 🔗 Linked issue

Resolves #18762 

### 📚 Description

Create a recipe about how to create a shared configuration to be accessed from build and runtime from `app.config`.

### DIFF
--- a/docs/2.guide/3.going-further/3.modules.md
+++ b/docs/2.guide/3.going-further/3.modules.md
@@ -630,6 +630,29 @@ nuxt.hook('builder:watch', async (event, path) => {
 })
 ```
 
+#### Build And Runtime Config
+
+You might want to access your [`app.config`](/docs/guide/directory-structure/app-config) via `useAppConfig` from your module. This config is meant to be used at runtime and it is not available at build time.
+
+If you want to access `app.config` from your module you can create a shared configuration file like this:
+
+```js [~/config/shared.config.mjs]
+export default {
+  theme: 'default'
+}
+```
+
+And then include this configuration in your `app.config`:
+
+```ts [app.config.ts]
+defineAppConfig({
+  ...sharedConfig
+})
+```
+
+You should now be able to import `~~/config/shared.config` and access your shared configuration from your module.
+
+
 ### Testing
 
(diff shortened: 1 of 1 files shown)


atinux:
Thanks! Wondering if it's not better to use the `shared/` directory instead? See https://nuxt.com/docs/guide/directory-structure/shared

Also you can reference it with `#shared/config` for example

IsraelOrtuno:
> Thanks! Wondering if it's not better to use the `shared/` directory instead? See https://nuxt.com/docs/guide/directory-structure/shared
> 
> Also you can reference it with `#shared/config` for example

Actually it makes more sense to me, updated!

danielroe (commented):
The `~~/shared` folder is also runtime, not build-time. And why access code within it through `useAppConfig` rather than directly importing it?

If the use case here is setting some configuration with a module at build-time, and reading it at runtime, then you could do that by setting `nuxt.options.appConfig` in your module, and then reading it with `useRuntimeConfig` in your runtime code.

IsraelOrtuno:
> The `~~/shared` folder is also runtime, not build-time. And why access code within it through `useAppConfig` rather than directly importing it?
> 
> If the use case here is setting some configuration with a module at build-time, and reading it at runtime, then you could do that by setting `nuxt.options.appConfig` in your module, and then reading it with `useRuntimeConfig` in your runtime code.

Now that you mention it makes sense that way, sometimes we make things so complex.

PR #31144, opened 2025-02-27
feat(nuxt): add useAsyncData dev warning

### LINKED ISSUE #31064: useAsyncData must return a value (it should not be undefined) - add warning location for better DX
### Describe the feature

For a dev who sees `WARN  [nuxt] useAsyncData must return a value (it should not be undefined) or the request may be duplicated on the client side.` and has tens of places fetching data to check, adding info of where the warning comes from would provide better DX.

### Additional information

### Final checks

xjccc:
### 🔗 Linked issue
resolves https://github.com/nuxt/nuxt/issues/31064#issuecomment-2677885668

### 📚 Description

Show user file in warning.

Is what I'm doing too simple? If you have some suggestion, please tell me, thank you.

**before:** 

**after:**


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -301,8 +301,13 @@ export function useAsyncData<
         }
 
         if (import.meta.dev && import.meta.server && typeof result === 'undefined') {
+          const { captureStackTrace } = await import('errx')
+          const { fileURLToPath } = await import('node:url')
+          const stack = captureStackTrace()
+          const caller = stack[stack.length - 1]
```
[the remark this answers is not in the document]
xjccc: Thank you

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -301,8 +301,13 @@ export function useAsyncData<
         }
 
         if (import.meta.dev && import.meta.server && typeof result === 'undefined') {
+          const { captureStackTrace } = await import('errx')
+          const { fileURLToPath } = await import('node:url')
```
ShayanBits, line 305: From my understanding, the issue here is that the modules are fetched every time the function is being called. I know node will cache them after the first fetch, but still it is an overhead.
I suggest to have a more generic approach for also similar cases in the future, namely import them once:

```
let fileURLToPath: typeof import('node:url').fileURLToPath
let captureStackTrace: typeof import('errx').captureStackTrace

async function loadServerModules () {
    const { fileURLToPath: _fileURLToPath } = await import('node:url')
    const { captureStackTrace: _captureStackTrace } = await import('errx')

    fileURLToPath = _fileURLToPath
    captureStackTrace = _captureStackTrace
}

if (import.meta.server) {
  loadServerModules()
}
```

This way it is guaranteed the modules are imported once. The downside of it is that the import is being trigered regardless of whether the modules will be used at all. But imho it is a better compromise than fetch them every time there is a warning. By the way the typing in my suggestions can be implemented better.

ShayanBits (commented):
Thank you!
I left two comment, one for better readability, and another one for a better import strategy.

packages/nuxt/src/app/composables/asyncData.ts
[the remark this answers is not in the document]
ShayanBits: you're welcome! 🙏

packages/nuxt/src/app/composables/asyncData.ts
xjccc: It's in dev mode. And I'd want another way for this? Because it’s none in build.

ShayanBits: The arguments from above are still valid for dev env too and the condition could be extended with `import.met.dev`.

danielroe (approved):
nice!

PR #31175, opened 2025-03-01
feat(kit,nuxt): resolve template imports from originating module

danielroe:
### 📚 Description

this adds a fast-path return for when we can resolve a module from the path that called `addTemplate` or `addPluginTemplate`, which allows different major versions of certain libraries used in templates (e.g. `ohash` to coexist simultaneously).

if you experience any issues, you can disable it with:

```ts
  experimental: {
    templateImportResolution: false
  }
})
```

### DIFF
--- a/packages/kit/src/template.ts
+++ b/packages/kit/src/template.ts
@@ -1,4 +1,5 @@
 import { existsSync, promises as fsp } from 'node:fs'
+import { fileURLToPath } from 'node:url'
 import { basename, isAbsolute, join, parse, relative, resolve } from 'pathe'
 import { hash } from 'ohash'
 import type { Nuxt, NuxtServerTemplate, NuxtTemplate, NuxtTypeTemplate, ResolvedNuxtTemplate, TSReference } from '@nuxt/schema'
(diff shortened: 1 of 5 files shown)


packages/kit/src/template.ts
```diff
@@ -27,6 +29,18 @@ export function addTemplate<T> (_template: NuxtTemplate<T> | string) {
   // Remove any existing template with the same destination path
   filterInPlace(nuxt.options.build.templates, p => normalizeTemplate(p).dst !== template.dst)
 
+  try {
+    const { source } = captureStackTrace().find(e => e.source && !e.source.startsWith('file://' + distDir)) ?? {}
+    if (source) {
+      const path = fileURLToPath(source)
+      if (existsSync(path)) {
+        template._path = path
```
pi0, line 37: just precaution here windows paths are not normalized

packages/kit/src/template.ts
```diff
@@ -27,6 +29,18 @@ export function addTemplate<T> (_template: NuxtTemplate<T> | string) {
   // Remove any existing template with the same destination path
   filterInPlace(nuxt.options.build.templates, p => normalizeTemplate(p).dst !== template.dst)
 
+  try {
+    const { source } = captureStackTrace().find(e => e.source && !e.source.startsWith('file://' + distDir)) ?? {}
```
pi0, line 33: Since we are using url comparation, wondering if would be better to use `distDirURL` (and avoid `fileURLToPath` which is also platform dependent in utils.ts)

packages/kit/src/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { fileURLToPath } from 'node:url'
```
danielroe, line 1:
```suggestion
```

packages/kit/src/utils.ts
```diff
@@ -1,3 +1,5 @@
...
+
```
danielroe, line 2:
```suggestion
```

PR #31190, opened 2025-03-03
fix(nuxt): fall back to wasm if oxc native bindings are missing

danielroe:


packages/nuxt/src/core/utils/parse.ts
```diff
@@ -38,14 +37,22 @@ export function walk (ast: Program | Node, callback: Partial<WalkOptions>) {
...
 export async function initParser () {
-  parseSync = await (provider === 'stackblitz' ? import('@oxc-parser/wasm') as unknown as Promise<typeof import('oxc-parser')> : import('oxc-parser'))
-    .then(r => r.parseSync)
+  try {
+    parseSync = await import('oxc-parser').then(r => r.parseSync)
+  } catch {
+    // this can fail on stackblitz so we fall back to wasm build
+    const { parseSync: wasmParse } = await import('@oxc-parser/wasm')
+    parseSync = (sourceFilename, code, options) => wasmParse(code, {
+      sourceFilename: sourceFilename.replace(/\?.*$/, '') + `.${options?.lang || 'ts'}`,
```
BobbieGoede, line 46: Was about to share something like this, but I see we found the same workaround 😂

BobbieGoede: that would be nice, I initially thought the signatures were the same.

btw, you're not passing `options` to `wasmParse`, might be unintended?

danielroe: it doesn't take any more options if the type signature can be trusted, and we only use it in the function below

PR #31192, opened 2025-03-03
feat(nuxt): support lazy hydration macros

### LINKED ISSUE #31663: resolveComponent and Lazy Hydration in Nuxt 3
### Environment

Environment
Operating System: MacOS
Node Version: v22.13.0
Nuxt Version: 3.16.2
CLI Version: 3.23.1
Nitro Version: 2.11.8
Package Manager: pnpm@9.15.4
Builder: -
User Config: compatibilityDate, devtools, experimental
Runtime Modules: -
Build Modules: -

### Reproduction

When trying to dynamically resolve a component with resolveComponent("LazyComponentName"), lazy hydration doesn't seem to work correctly. However, when using the component directly in the template, lazy hydration functions as expected.

1. Register the component dynamically using resolveComponent:
```
const Components = {
  ComponentSectionsHeader: resolveComponent("LazySectionsHeader"),
};
```
2. Use it in the template:
<component :is="Components.ComponentSectionsHeader" />
3. Observe that lazy hydration is not applied.
4. Now, test it by adding the component directly in the template:

`<LazySectionsHeader hydrate-never />`
5.  Notice that lazy hydration works correctly in this case.

### Describe the bug

When I try to use resolveComponent with the Lazy prefix, lazy hydration doesn't work correctly:

`const Components = {
  ComponentSectionsHeader: resolveComponent("LazySectionsHeader"),
};`

For testing, I added the component directly like this:

 <LazySectionsHeader hydrate-never />

And in this case, it works as expected.

Is there a solution for this?

### Additional context

_No response_

### Logs

```shell-script

```

Mini-ghost:
### 📚 Description

This PR introduces a few macros to provide an explicit import solution for lazy hydration.  

resolves https://github.com/nuxt/nuxt/issues/31663

#### TODO

- [x] Macro plugin (`LazyHydrationMacroTransformPlugin`)
- [x] Type support
- [x] Documentation
- [x] Tests

#### Visibility Strategy

Visibility strategy enables deferring the hydration of a component until the element(s) become visible in the viewport.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'visible',
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!-- 
      Hydration will be triggered when
      the element(s) is 100px away from entering the viewport.
    -->
    <LazyHydrationMyComponent :hydrate-on-visible="{ rootMargin: '100px' }" />
  </div>
</template>
```

#### Idle Strategy

Idle strategy enables deferring the hydration of a component until the browser is idle, ensuring that critical rendering tasks are prioritized first.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'idle',
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!-- Hydration will be triggered when the browser is idle or after 2000ms. -->
    <LazyHydrationMyComponent :hydrate-on-idle="2000" />
  </div>
</template>
```

#### Interaction Strategy

Interaction strategy enables delaying the hydration of a component until a user interacts with it (e.g., click, mouseover).

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'interaction',
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!--
      Hydration will be triggered when
      the element(s) is hovered over by the pointer.
    -->
    <LazyHydrationMyComponent hydrate-on-interaction="mouseover" />
  </div>
</template>
```

#### Media Query Strategy

Media query strategy enables deferring the hydration of a component until a specified media query matches.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'mediaQuery',
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!--
      Hydration will be triggered when
      the window width is greater than or equal to 768px.
    -->
    <LazyHydrationMyComponent hydrate-on-media-query="(min-width: 768px)" />
  </div>
</template>
```

#### If Strategy

If strategy enables delaying the hydration of a component until a boolean condition is met.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'if',
  () => import('./components/MyComponent.vue')
)
const isReady = ref(false)
function myFunction() {
  // Trigger custom hydration strategy...
  isReady.value = true
}
</script>
<template>
  <div>
    <!-- Hydration is triggered when isReady becomes true. -->
    <LazyHydrationMyComponent :hydrate-when="isReady" />
  </div>
</template>
```

#### Time Strategy

Time strategy enables delaying the hydration of a component until a specified time (in milliseconds) has passed.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'time', 
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!-- Hydration is triggered after 1000ms. -->
    <LazyHydrationMyComponent :hydrate-after="1000" />
  </div>
</template>
```

#### Never Hydrate

Never strategy prevents the component from being hydrated by Vue. This means that the component remains static and does not require Vue's hydration process.

```vue
<script setup lang="ts">
const LazyHydrationMyComponent = defineLazyHydrationComponent(
  'never',
  () => import('./components/MyComponent.vue')
)
</script>
<template>
  <div>
    <!-- This component will never be hydrated by Vue. -->
    <LazyHydrationMyComponent />
  </div>
</template>
```

If there's anything I may have overlooked or if you have any suggestions, please feel free to let me know. I truly appreciate your feedback.


danielroe (commented):
I think it's important to align with the hydration props, for clarity.

~~We could use: `defineHydrateAfterComponent`, `defineHydrateOnIdleComponent`, `defineHydrateWhenComponent`, `defineHydrateOnInteractionComponent`, `defineHydrateOnMediaQueryComponent`, `defineHydrateNeverComponent`, `defineHydrateOnVisibleComponent`,~~

Let's use a single `defineLazyHydrationComponent(hydrationStrategy, () => import(''))` for better DX - I think it would be just as doable to use as a macro. And it simplifies the documentation too...

Mini-ghost:
Thank you for your feedback! I’ll take some time to explore how to improve this further.

I’m not sure if defining `hydrationStrategy` as a static string would provide a better DX, and if it would also make analysis easier. What do you think?

danielroe:
Yes, we would have to be very clear users would have to pass a string literal. On the other hand, we can strongly type the arguments, and I think it reduces the cognitive load if the string exactly matches the prop set in the template. Also, we only have one page in the docs to explain how it works, rather than seven.

Mini-ghost:
I’ve noticed an issue:

```vue
<script setup lang="ts">
const LazyMyComponent = defineLazyHydrationComponent('time', () => import('~/components/MyComponent.vue'))
</script>

<template>
  <div>
    <LazyMyComponent
      :hydrate-after="500"
    />
  </div>
</template>
```

This causes a conflict between auto-import and macro usage:

```
Uncaught (in promise) SyntaxError: Identifier 'createLazyTimeComponent' has already been declared
```

I may need to spend some time investigating how to resolve this issue!  
If you have any suggestions or ideas, I’d really appreciate your guidance.

danielroe:
I think we could add support in the html-transform plugin for detecting if `defineLazyHydrationComponent` is used?

Mini-ghost:
I'm not entirely sure if this is the correct approach, but to address this issue, perhaps we should check for existing variable declarations during the `nuxt:components-loader-pre` transformation.  

For example, in the following scenario, we might want to skip the transformation:

```vue
<script setup lang="ts">
const LazyMyComponent = defineLazyHydrationComponent('time', () => import('~/components/MyComponent.vue'))
</script>

<template>
  <div>
    <LazyMyComponent :hydrate-after="500" />
  </div>
</template>
```

Currently, however, `<LazyMyComponent :hydrate-after="500" />` gets transformed into `<LazyTimeMyComponent :hydrate-after="500" />`, which leads to the issue previously mentioned.

截圖 2025-04-07 凌晨1 14 04

Would love to hear your thoughts on this approach, in case there's a better solution I might have missed.

danielroe:
yes, exactly! 👌

jakubm95:
@Mini-ghost do you know when it will be merged?

packages/nuxt/src/components/plugins/lazy-hydration-transform.ts
```diff
@@ -41,6 +54,20 @@ export const LazyHydrationTransformPlugin = (options: LoaderOptions) => createUn
...
+      const variables = new Set<string>()
+      if (codeWithoutComments !== code) {
+        for (const regex of EXCLUDE_RE) {
+          for (const match of codeWithoutComments.matchAll(regex)) {
+            if (!match[1]) { continue }
+            variables.add(match[1])
+          }
+        }
+      }
+
```
danielroe, line 70: I'm not sure this is quite right. Do we only want to scan variables if the code has comments? And I think we should respect 'scope' rather than looking for any variable anywhere in the code.

In general I'd prefer to avoid using regexp for this. I feel a parsing-based scope-tracking approach would be sounder (but we can make that performant by only parsing the code if it code contains some of the 'reserved' names - if not we have a 'fast path' without parsing)

Mini-ghost: Thank you for pointing that out. I realized we probably shouldn’t restrict variable scanning to cases where comments are present, so I’ve submitted a corrected version accordingly.

By the way, just to clarify, were you suggesting that using a JavaScript AST parser such as `acorn` might be a more reliable approach compared to using regular expressions?

packages/nuxt/src/components/plugins/loader.ts
```diff
@@ -90,43 +90,57 @@ export const LoaderPlugin = (options: LoaderOptions) => createUnplugin(() => {
               switch (modifier) {
                 case 'Visible':
                 case 'visible-':
-                  imports.add(genImport(options.clientDelayedComponentRuntime, [{ name: 'createLazyVisibleComponent' }]))
+                  if (!code.includes('createLazyVisibleComponent')) {
```
danielroe, line 93: do we need this? (or, maybe we could run `findImports` instead of matching across the whole code)?

Mini-ghost: This is to prevent potential conflicts when users mix both auto-imports and explicit imports within a single SFC. I’ll take a closer look and evaluate whether `findImports` or the scope tracker would be a better fit for this case.

danielroe: `createLazyVisibleComponent` is not exposed to end users - it's a purely internal utility

packages/nuxt/src/components/plugins/lazy-hydration-macro-transform.ts
```diff
@@ -0,0 +1,115 @@
...
+  sourcemap?: boolean
+  transform?: ComponentsOptions['transform']
+  clientDelayedComponentRuntime: string
+}
+
+const LAZY_HYDRATION_MACRO_RE = /(?:\b(?:const|let|var)\s+(\w+)\s*=\s*)?defineLazyHydrationComponent\(\s*['"]([^'"]+)['"]\s*,\s*\(\s*\)\s*=>\s*import\s*\(\s*['"]([^'"]+)['"]\s*\)\s*\)/g
+const COMPONENT_NAME = /import\(["'].*\/([^\\/]+?)\.\w+["']\)/
+const HYDRATION_STRATEGY = ['visible', 'idle', 'interaction', 'mediaQuery', 'if', 'time', 'never']
+
+export const LAZY_HYDRATION_MAGIC_COMMENT = '/* #__LAZY_HYDRATION_MACRO__ */'
```
danielroe, line 22: I think we can actually avoid the need for this by importing with a prefix, e.g. `import { createLazyInteractionComponent as _createLazyInteractionComponent }`. Rollup will deduplicate these with any imports added in the component loader. And we avoid the costliness/flakiness of checking for magic comments + parsing AST.

jakubm95:
@Mini-ghost 
Do you know when it will be finally implemented?

Mini-ghost:
@jakubm95 If everything goes smoothly, I believe it might be included in the 3.18 release.

danielroe (approved):

danielroe:
Amazing work - thank you ❤️

Issue #31195, opened 2025-03-03
Sometimes "page:loading:end" hook is not called


madeofsun:
### Environment

❯ npx nuxi info
[ nuxi  17:54:41] Working directory: /home/projects/github-eqmukpad
[ nuxi  17:54:41] Nuxt project info: 

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.4
- CLI Version:      3.22.4
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      compatibilityDate, devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-eqmukpad

In order to reproduce:
- click `Clear`
- click `abc`  - page:loading:end is present
- click `abc?1` - page:loading:end is present
- click `qwe` - page:loading:end **is missing**

### Describe the bug

If a custom key is used for the page and next location have the same key (`abc` -> `abc?1`), then further navigation to location with different key (`abc?1` -> `qwe`) is missing "page:loading:end" event.

I suspect that behaviour is caused by that [flag](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/runtime/page.ts#L122).
The flag is set when key is not changed and is reset only after `page:finish` - which does not happen, because the key is not changed. Since flag is not reset, on the next navigation the hook is skipped.
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/pages/runtime/page.ts#L162

### Logs

```shell-script

```

danielroe:
resolved in https://github.com/nuxt/nuxt/pull/31504.

test added in 55af0f7d98f00a17847cb55b8411788098182a16

PR #31201, opened 2025-03-04
fix: Loading indicator gets stuck when NuxtPage CustomProps changed

### LINKED ISSUE #31099: NuxtLoadingIndicator gets stuck when custom props change
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.15.4
- CLI Version:      3.22.2
- Nitro Version:    2.10.4
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-3hjrp6na

### Describe the bug

Basically, any render function call of a child NuxtPage will suppress the next page:loading:end hook. In a situation where non-static custom props are passed to NuxtPage, NuxtLoadingIndicator will get stuck (page:loading:end won't be triggered, but page:finish will), while there won't be a next URL change.

### Additional context

It looks like a regression after https://github.com/nuxt/nuxt/pull/29009.

oimo23:
### 🔗 Linked issue
resolves  https://github.com/nuxt/nuxt/issues/31099
also https://github.com/nuxt/nuxt/issues/31195

### 📚 Description
fix: Loading indicator gets stuck after NuxtPage CustomProps changed.

**Cause**  
When `NuxtPage` custom props change, re-render happens and falls into that `if` statement. 
On the next navigation, `pageLoadingEndHookAlreadyCalled` remains `true`, causing the loading indicator to get stuck.

```
if (!nuxtApp.isHydrating && !hasChildrenRoutes(forkRoute, routeProps.route, routeProps.Component) && previousPageKey === key) {
  nuxtApp.callHook('page:loading:end')
  pageLoadingEndHookAlreadyCalled = true
}
```

But when custom props change, `page:loading:start` isn’t even called, so there’s no need to explicitly invoke `page:loading:end`. 
In other words, this `if` statement is getting triggered unnecessarily.

This `if` statement should only be triggered when `page:loading:start` has fired but Suspense never resolves.

**What I Did**  
- I removed the `pageLoadingEndHookAlreadyCalled` flag because it introduced unnecessary complexity and bug like this issue.

- Although `page:loading:start` always fires when the route changes, `page:loading:end` isn’t guaranteed—particularly if the route itself doesn’t change but only the query or fragment does, meaning Suspense never resolves. This was the scenario covered by that `if` statement.

```
const isFinalRender = !nuxtApp.isHydrating && !hasChildrenRoutes(forkRoute, routeProps.route, routeProps.Component)
const isSamePageButDifferentQueryOrFragment = key === previousPageKey && previousRouteFullPath !== routeProps.route.fullPath

if (isFinalRender && isSamePageButDifferentQueryOrFragment) {
  nuxtApp.callHook('page:loading:end')
}
```


packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -117,11 +117,17 @@ export default defineComponent({
...
+          const isSamePageButDifferentQueryOrFragment = key === previousPageKey && previousRouteFullPath !== routeProps.route.fullPath
+          
+          // When navigating within the same page and only changing the query or fragment, 
+          // Suspense doesn't resolve, so we need to manually call `page:loading:end`
+          if (isFinalRender && isSamePageButDifferentQueryOrFragment) {
             nuxtApp.callHook('page:loading:end')
-            pageLoadingEndHookAlreadyCalled = true
           }
 
+          previousRouteFullPath = routeProps.route.fullPath
```
oimo23, line 130: It seemed feasible to unify `previousPageKey` and `previousRouteFullPath` into one value. However, doing so would require altering the `generateRouteKey` implementation, which could impact other parts of the code. Therefore, I decided not to make that change.

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -117,11 +117,17 @@ export default defineComponent({
...
-          if (!nuxtApp.isHydrating && !hasChildrenRoutes(forkRoute, routeProps.route, routeProps.Component) && previousPageKey === key) {
+
+          const isFinalRender = !nuxtApp.isHydrating && !hasChildrenRoutes(forkRoute, routeProps.route, routeProps.Component)
+          const isSamePageButDifferentQueryOrFragment = key === previousPageKey && previousRouteFullPath !== routeProps.route.fullPath
+
+          // When navigating within the same page and only changing the query or fragment,
+          // Suspense doesn't resolve, so we need to manually call `page:loading:end`
+          if (isFinalRender && isSamePageButDifferentQueryOrFragment) {
             nuxtApp.callHook('page:loading:end')
-            pageLoadingEndHookAlreadyCalled = true
```
oimo23, line 122: call 'page:loading:end' is ok.

but "pageLoadingEndHookAlreadyCalled" flag change into true and remains until next navigation even if when suspense doesn't resolves. 
(customProps change, url anchor or fragment change)

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -117,11 +117,17 @@ export default defineComponent({
           }
 
           const key = generateRouteKey(routeProps, props.pageKey)
-          if (!nuxtApp.isHydrating && !hasChildrenRoutes(forkRoute, routeProps.route, routeProps.Component) && previousPageKey === key) {
```
oimo23, line 120: This if statement originally added in this PR.
https://github.com/nuxt/nuxt/issues/24928

oimo23:
@danielroe 
Sorry to bother you, but when you have some time could you review the code?

danielroe:
my apologies for the delay in merging this. I'm closing it as I think it ended up resolved in https://github.com/nuxt/nuxt/pull/31504, but please do take a look and see if I've missed anything.

I added a test here: https://github.com/nuxt/nuxt/commit/965b4098e82a7d1bf7a4df6e0c5dfcc3421868af

thank you for your work on this ❤️

oimo23:
@danielroe 
Thank you, daniel.
I noticed one thing, this doc description is not accurate.

スクリーンショット 2025-03-26 8 11 51

I think this description accurately reflects the actual behavior.
What do you think?
```
Called after page:finish (if triggered). Always called after page:loading:start, even if page:finish is skipped (e.g., on query-only changes).
```

PR #31209, opened 2025-03-05
fix(nuxt): consolidate head component context

harlan-zw:
### 📚 Description

With Unhead v2 meta tags are deduped intuitively based on the entry they are registered in. With this change it no longer dedupes based on `key` or `id` and only on the meta name (or property) and the entry.

This means doing something like this: https://github.com/nuxt-modules/i18n/blob/main/specs/fixtures/basic_usage/layouts/default.vue, no longer works.

```html
<Meta id="theme-color-1" name="theme-color" content="cyan" media="(prefers-color-scheme: light)"> <!-- renders -->
<Meta id="theme-color-2" name="theme-color" content="black" media="(prefers-color-scheme: dark)"> <!-- does not -->
```

To mitigate any potential breaking changes related to this in Nuxt 3.16 we can collect a shared context when rendering the head components inside of a `<Html>` or `<Head>`. Meaning we only insert one Unhead entry for each  `<Html>`  or `<Head>` if they exist. This also has a slight performance improvement as we have fewer entries that we're tracking between head DOM renders.

With this fix, we also make it clear that using the head components is not recommended and only exists for personal preference reasons. As we may end up deprecating these in the future we should nudge people towards not using them.

This PR also includes some fixes around the props:
- `body` is deprecated and they should use `tagPosition` if needed
- `renderPriority` never actually did anything, have deprecated for `tagPriority`
- `<Meta property="...">` will now work

### DIFF
--- a/docs/1.getting-started/5.seo-meta.md
+++ b/docs/1.getting-started/5.seo-meta.md
@@ -100,9 +100,10 @@ useSeoMeta({
 
 ## Components
 
-Nuxt provides `<Title>`, `<Base>`, `<NoScript>`, `<Style>`, `<Meta>`, `<Link>`, `<Body>`, `<Html>` and `<Head>` components so that you can interact directly with your metadata within your component's template.
+While using [`useHead`](/docs/api/composables/use-head) is recommended in all cases, you may have a personal preference for defining your head tags in your template using components.
 
-Because these component names match native HTML elements, they must be capitalized in the template.
+Nuxt provides the following components for this purpose: `<Title>`, `<Base>`, `<NoScript>`, `<Style>`, `<Meta>`, `<Link>`, `<Body>`, `<Html>` and `<Head>`. Note
+the capitalization of these components ensuring we don't use invalid native HTML tags.
 
 `<Head>` and `<Body>` can accept nested meta tags (for aesthetic reasons) but this does not affect _where_ the nested meta tags are rendered in the final HTML.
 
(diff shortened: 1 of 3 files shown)


danielroe (approved):

BobbieGoede:
>This means doing something like this: [nuxt-modules/i18n@main/specs/fixtures/basic_usage/layouts/default.vue](https://github.com/nuxt-modules/i18n/blob/main/specs/fixtures/basic_usage/layouts/default.vue?rgh-link-date=2025-03-05T05%3A58%3A01Z), no longer works.

@harlan-zw 
Are you referring to the `<Link>` and `<Meta>` components which have their `id` set?

I came across this PR since nuxt-i18n tests in ecosystem-ci are failing, it seems like `<Title>` no longer reactively updates to changes in [the same fixture you referenced](https://github.com/nuxt-modules/i18n/blob/main/specs/fixtures/basic_usage/layouts/default.vue#L15) ([test logs](https://github.com/nuxt/ecosystem-ci/actions/runs/13700617272/job/38313243477#step:7:7097) and [test checking page title after navigation](https://github.com/nuxt-modules/i18n/blob/main/specs/basic_usage.spec.ts#L384)), not sure if that is what you meant, I haven't checked yet if the changes in this PR are the source of this though.

Also, feel free to tag me next time when referencing nuxt-i18n with possible changes to recommendations 🙏

PR #3085, opened 2018-03-22
feat: passing props to route components on extendRoutes

NicoPennec:
Missing part on vue-router support with `extendRoutes`: https://router.vuejs.org/en/essentials/passing-props.html

According the following mention from the doc: https://nuxtjs.org/api/configuration-router/#extendroutes

> The schema of the route should respect the vue-router schema.

Related to https://github.com/nuxt/nuxt.js/issues/1591

PR #31230, opened 2025-03-05
feat(nuxt): align nuxt error handling with nitro

danielroe:
Nitro has released a very nice error integration using Youch to render errors both to console + as a lovely HTML page (https://github.com/nitrojs/nitro/pull/3002).

This PR aligns the Nuxt error handling with Nitro.

1. Where possible we defer to Nitro (specifically, in the case where previously Nuxt rendered a JSON response for errors, such as in response to a `curl` or API call). This is made possible by the new 'stacked' error handler support in Nitro ([#3085](https://github.com/nitrojs/nitro/pull/3085)).

1. We now redirect 404 errors that occur because they are hitting a `baseURL` that Nitro isn't serving.

1. The format of the error object returned from API routes by Nitro is a little different - the `url` is now a full URL, not just a path, and the `stack` (in development) is an array of lines rather than a string.

1. We continue to render HTML for requests made by browsers, even for API routes. (When Nuxt launched we had hard-coded JSON responses for these routes but we had to revert due to user issues and so it's important we continue to render HTML in this case.)

1. We do not render the beautiful Youch error page for Nuxt errors. We plan to show a button on the Nuxt development error page that allows toggling this view. This will ultimately be implemented by Nitro (probably via a hook) to allow us to merge the two error pages together.

1. (We do however display full-colour errors with stacktrace in the terminal, using Youch!)

   CleanShot 2025-03-05 at 23 42 17@2x

1. Finally, when rendering an error page the following headers are set:

   ```
   // Prevent browser from guessing the MIME types of resources.
   'X-Content-Type-Options': 'nosniff'
   // Prevent error page from being embedded in an iframe
   'X-Frame-Options': 'DENY'
   // Prevent browsers from sending the Referer header
   'Referrer-Policy': 'no-referrer'
   ```

We're likely going to see a lot of ongoing development here. Ultimately the aim is for Nuxt to align as much as possible to Nitro for error handling, so user experience is consistent regardless of whether debugging server routes or Vue app pages.

Many thanks to @Saeid-Za and @danielphilipjohnson for their PRs previously implementing this.

closes https://github.com/nuxt/nuxt/pull/30624, closes https://github.com/nuxt/nuxt/pull/30606


packages/nuxt/src/core/runtime/nitro/handlers/error.ts
```diff
@@ -1,54 +1,44 @@
...
-    console.error(tags, (error.message || error.toString() || 'internal server error') + '\n' + stack.map(l => '  ' + l.text).join('  \n'))
+export default <NitroErrorHandler> async function errorhandler (error, event, { defaultHandler }) {
+  if (isJsonRequest(event)) {
+    // let Nitro handle JSON errors
+    return
   }
+  // invoke default Nitro error handler (which will log appropriately if required)
+  const defaultError = await defaultHandler(error, event, { json: true })
 
   if (event.handled) { return }
```
pi0, line 18: defaultHandler does not handle response but only return one, it might be a misleading leftover check(also I suggest you rename `defaultError` to `response` or *response)

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
```diff
@@ -1,54 +1,44 @@
...
 
-  // Set response code and message
-  setResponseStatus(event, (errorObject.statusCode !== 200 && errorObject.statusCode) as any as number || 500, errorObject.statusMessage)
+  // let Nitro handle redirect if appropriate
+  const statusCode = error.statusCode || 500
+  if (statusCode === 404 && defaultError.status === 302) {
+    return defaultError
```
pi0, line 23: Terror handler is fully responsible for sending response, it cannot be retuerned ([example](https://github.com/nitrojs/nitro/blob/v2/src/runtime/internal/error/prod.ts#L14))

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
```diff
@@ -1,54 +1,44 @@
...
+  }
 
-  // JSON response
-  if (isJsonRequest(event)) {
-    setResponseHeader(event, 'Content-Type', 'application/json')
-    return send(event, JSON.stringify(errorObject))
+  if (import.meta.dev && typeof defaultError.body !== 'string' && Array.isArray(defaultError.body.stack)) {
+    // normalize to string format expected by nuxt `error.vue`
+    defaultError.body.stack = defaultError.body.stack.join('\n')
```
pi0, line 28: couldn't we at same time update error.vue to adopt stack as array? (avoids different nuxt behavior)

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
```diff
@@ -1,54 +1,44 @@
...
   }
 
+  const errorObject = defaultError.body as Pick<NonNullable<NuxtPayload['error']>, 'error' | 'statusCode' | 'statusMessage' | 'message' | 'stack'> & { url: string, data: any }
+  errorObject.message ||= 'Server Error'
+
+  setResponseHeaders(event, {
+    'x-content-type-options': defaultError.headers['x-content-type-options'],
```
pi0, line 35: We are discarding any possible future headers from default handler. could be `...defaultError.headers`

packages/nuxt/src/core/runtime/nitro/utils/error.ts
```diff
@@ -0,0 +1,27 @@
...
+  // If the client specifically requests HTML, then avoid classifying as JSON.
+  if (hasReqHeader(event, 'accept', 'text/html')) {
+    return false
+  }
+  return (
+    hasReqHeader(event, 'accept', 'application/json') ||
+    hasReqHeader(event, 'user-agent', 'curl/') ||
+    hasReqHeader(event, 'user-agent', 'httpie/') ||
+    hasReqHeader(event, 'sec-fetch-mode', 'cors') ||
+    event.path.startsWith('/api/') ||
```
pi0, line 17: FWIW it is a negative and confusing DX (i made a change in nitro based on lots of feedbacks heard from Nuxt users). At least for development we could make new behavior.

danielroe: I'm planning to update for nuxt v4

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
danielroe: I was wanting to avoid fragility in case you changed case of headers or added some additional ones that would break JS-enabled error pages (as we render in Nuxt).

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
danielroe: Not really. Because the user may provide their own `error.vue` which already handles a string.

packages/nuxt/src/core/runtime/nitro/handlers/error.ts
danielroe: `Cache-Control` doesn't seem to be lowercase, fyi

pi0: https://github.com/nitrojs/nitro/commit/e3866d04130b6ac9deb712a1f92515e85e880c43

Issue #31219, opened 2025-03-05
Configure ignore for nitro's dev filesystem mounts


cjpearson:
### Describe the feature

Related to https://github.com/nuxt/nuxt/issues/30481

I was able to workaround this issue by switching the parcel watcher, and tweaking nitro's [development filesystem mounts](https://nitro.build/guide/storage#development-only-mount-points). My solution was to use `fs-lite` for the root mount and avoid chokidar completely on this mount. 

But since the [fs driver](https://unstorage.unjs.io/drivers/fs) supports an `ignore` option, perhaps Nuxt could configure the driver with `nuxt._ignore`? My hope is this could reduce how many file handles are created during dev mode. 

### Additional information

- [x] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [x] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [x] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

cjpearson:
Done in https://github.com/nuxt/nuxt/issues/31219

Issue #30481, opened 2025-01-06
"EMFILE: too many open files, watch" with Nuxt 3, Vite with SST


claylevering:
### Environment

Hi team - I believe this is related to #29744 and a request to re-open the issue.  This is mostly because I believe it was unable to be reproduced previously however I now have a clear and easy reproduction process.

# Environment

- Operating System: Darwin
- Node Version: v22.11.0
- Nuxt Version: 3.15.1
- CLI Version: 3.15.1
- Nitro Version: default
- Package Manager: npm@10.9.0
- Builder: -
- User Config: default
- Runtime Modules: -
- Build Modules: -

### Reproduction

1. Initialize a new Nuxt 3 project with Vite using npm / install packages.  (_**NOTE**_: This DOES NOT install 3.15.1, that is a later step):
```shell
npx nuxi@latest init aws-nuxt
cd aws-nuxt
npm ci
```
3. Verify Nuxt functions:
```shell
npm run dev
```
4. Initialize SST (use default options):
```shell
npx sst@latest init
```
5. Update Nuxt config to use SST / Lambda:
```typescript
// https://nuxt.com/docs/api/configuration/nuxt-config
export default defineNuxtConfig({
  compatibilityDate: '2024-11-01',
  devtools: { enabled: true },
  nitro: {
    preset: "aws-lambda",
  },
})
```
6. Start SST dev:
```shell
npx sst dev
```

This will initiate SST as a lambda function in your AWS account.  Attached are screenshots of the SST console for review.  The error seen is:

```
ERROR  [unhandledRejection] EMFILE: too many open files, watch                                                                   at FSWatcher._handle.onchange (node:internal/fs/watchers:207:21)

Restarting Nuxt due to error: Error: EMFILE: too many open files, watch
```

Downgrading Nuxt to a locked `3.13.0` works fine:
```json
{
  "name": "nuxt-app",
  "private": true,
  "type": "module",
  "scripts": {
    "build": "nuxt build",
    "dev": "nuxt dev",
    "generate": "nuxt generate",
    "postinstall": "nuxt prepare",
    "preview": "nuxt preview"
  },
  "dependencies": {
    "nuxt": "3.13.0",
    "sst": "3.4.56",
    "vue": "latest",
    "vue-router": "latest"
  },
  "overrides": {
    "nitropack": "npm:nitropack-nightly@latest"
  }
}
```

```shell
rm -rf node_modules package-lock.json
npm install
```

[...]

PR #31233, opened 2025-03-06
fix(nuxt): apply ignore rules to nitro `devStorage`

cjpearson:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/31219

### 📚 Description

This change is an attempt to mitigate issues like https://github.com/nuxt/nuxt/issues/30481 by reducing the amount of files watched. It does this by applying the Nuxt ignore function to the development-only filesystem mounts created by Nitro. 

In my experience the threshold for the `EMFILE` error seems to vary, so this is no guarantee of eliminating it, but it may help. When I encountered https://github.com/nuxt/nuxt/issues/30481 the combination of switching to the parcel watcher and using fs-lite for nitro's root mount was sufficient to overcome the issue. Either alone was not enough.

For a rough estimate of the open file handles, I ran `lsof | grep '^node' | wc -l` while running in dev mode. Here are the results in my application:

Without changes: ~1500
Using fs-lite for the root mount (so no watching): ~650
Adding nuxt ignore (this PR):  ~1300

So in my case I have about 200 files in the application directory that should be ignored. This would vary for different codebases though and I'm not sure what's typical.

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -6,7 +6,7 @@ import { createRouter as createRadixRouter, exportMatcher, toRouteMatcher } from
 import { joinURL, withTrailingSlash } from 'ufo'
 import { build, copyPublicAssets, createDevServer, createNitro, prepare, prerender, scanHandlers, writeTypes } from 'nitropack'
 import type { Nitro, NitroConfig, NitroOptions } from 'nitropack'
-import { findPath, logger, resolveAlias, resolveIgnorePatterns, resolveNuxtModule } from '@nuxt/kit'
+import { createIsIgnored, findPath, logger, resolveAlias, resolveIgnorePatterns, resolveNuxtModule } from '@nuxt/kit'
 import escapeRE from 'escape-string-regexp'
 import { defu } from 'defu'
 import { dynamicEventHandler } from 'h3'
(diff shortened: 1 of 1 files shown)


pi0:
If a fix needs for Nitro, please do it in Nitro!!

Watcher is initiated by devtools btw, so at least it could there!

pi0:
storage watcher is initiated by Nuxt devtools (not nuxt core, not nitro core) -- even if patch needs to be here it should be in there.

This PR is **overriding** nitro default mountpoints also (not extending)

If a generic ignore pattern can be improved in Nitro, please do consider moving to upstream instead of divering nitro behavior in Nuxt.

cjpearson:
Hi @pi0, thanks for the feedback!

I was also interested in seeing if something could be done on the nitro side. One thought was to mount `root` using fs-lite instead. This is essentially what I'm doing through my nuxt.config now, but that would make `root` unwatchable and I wasn't sure if that would be a problem. (If it's not a problem this is even better in terms of limiting unnecessary watchers)

I guess another idea would be to add an ignore option to nitro which it would then use for all the fs mounts. This would allow Nuxt's ignore rules to be used as in this PR, but without overriding the mountpoints. Would you be open to a feature like this?

While devtools is the trigger for the watcher in my case, any other module could also watch these storage mounts so I think it makes sense to tackle at the nuxt or nitro level.

pi0:
Thanks for the explanations. It makes sense that nuxt ignores patterns respected for default mounts as well, but I'm worried that we should do it consistently (there are root, src, build, and cache default mount points) and these can be changed. 

+ if nuxt users use `nitro: { devStorage: {} }` we are completely overriding it in current impl. Similar to Nitro we should at least extend previous ones.

pi0:
creating an issue in devtools to see if it can use it’s explicit watched mount would be nice next step imo.

danielroe:
ultimately I would like to avoid hard coding this, so when we can pass an `isIgnored` to nitro to use in the appropriate places that would be ideal

but thank you so much for your work on this!

danielroe (approved):

PR #31234, opened 2025-03-06
docs: mention possibility of prerendering api routes

Gwynerva:
### 📚 Description

It is not clear that `prerenderRoutes` allows to prerender API routes as well, allowing to `$fetch` data on SSG sites as if live server is available. This allows a way to get data without polluting payload.

Also, I added warning about headers that are lost when prerendering.

### DIFF
--- a/docs/3.api/3.utils/prerender-routes.md
+++ b/docs/3.api/3.utils/prerender-routes.md
@@ -32,7 +32,7 @@ In the browser, or if called outside prerendering, `prerenderRoutes` will have n
 You can even prerender API routes which is particularly useful for full statically generated sites (SSG) because you can then `$fetch` data as if you have an available server!
 
 ```js
-prerenderRoutes('/api/content/article/name-of-article');
+prerenderRoutes('/api/content/article/name-of-article')
 
 // Somewhere later in App
 const articleContent = await $fetch('/api/content/article/name-of-article', {
@@ -41,6 +41,6 @@ const articleContent = await $fetch('/api/content/article/name-of-article', {
 ```
 
 ::warning
-Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type. In this case, you may have to specify an explicit `responseType`.
-Always manually set `responseType` when fetching prerenderered API routes!
+Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type.
+Always manually set `responseType` when fetching prerenderered API routes.
 ::
--- a/docs/3.api/3.utils/prerender-routes.md
+++ b/docs/3.api/3.utils/prerender-routes.md
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
 ::note
 In the browser, or if called outside prerendering, `prerenderRoutes` will have no effect.
 ::
+
+You can even prerender API routes which is particularly useful for full Statically Generated Sites (SSG) because you can then `$fetch` data as if you have an available server!
+
+```js
+prerenderRoutes('/api/content/article/name-of-article');
+
+// Somewhere later in App
+const articleContent = await $fetch('/api/content/article/name-of-article', {
+  responseType: 'json',
+});
+```
+
+::warning
+Fetching prerendered API routes on production looses all headers you set up in `defineEventHandler` on server side! This means the data you fetch will most likely have `application/octet-stream` content type even if it is just a JSON and therefore break the application in unpredictable ways.
+Always manually set `responseType` when fetching prerenderered API routes!
+::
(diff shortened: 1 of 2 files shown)


docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
 ::note
 In the browser, or if called outside prerendering, `prerenderRoutes` will have no effect.
 ::
+
+You can even prerender API routes which is particularly useful for full Statically Generated Sites (SSG) because you can then `$fetch` data as if you have an available server!
```
danielroe, line 32:
```suggestion
You can even prerender API routes which is particularly useful for full statically generated sites (SSG) because you can then `$fetch` data as if you have an available server!
```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
...
+
+```js
+prerenderRoutes('/api/content/article/name-of-article');
+
+// Somewhere later in App
+const articleContent = await $fetch('/api/content/article/name-of-article', {
+  responseType: 'json',
+});
```
danielroe, line 40:
```suggestion
})
```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
...
+```
+
+::warning
+Fetching prerendered API routes on production looses all headers you set up in `defineEventHandler` on server side! This means the data you fetch will most likely have `application/octet-stream` content type even if it is just a JSON and therefore break the application in unpredictable ways.
```
danielroe, line 44:
```suggestion
Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type. In this case, you may have to specify an explicit `responseType`.
```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
 ::note
 In the browser, or if called outside prerendering, `prerenderRoutes` will have no effect.
 ::
+
+You can even prerender API routes which is particularly useful for full statically generated sites (SSG) because you can then `$fetch` data as if you have an available server!
+
+```js
+prerenderRoutes('/api/content/article/name-of-article');
```
danielroe, line 35:
```suggestion
prerenderRoutes('/api/content/article/name-of-article')
```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
...
+
+// Somewhere later in App
+const articleContent = await $fetch('/api/content/article/name-of-article', {
+  responseType: 'json',
+})
+```
+
+::warning
+Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type. In this case, you may have to specify an explicit `responseType`.
```
danielroe, line 44:
```suggestion
Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type.
```

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
...
+Always manually set `responseType` when fetching prerenderered API routes!
```
danielroe, line 45:
```suggestion
Always manually set `responseType` when fetching prerenderered API routes.
```

danielroe (approved):

docs/3.api/3.utils/prerender-routes.md
```diff
@@ -28,3 +28,19 @@ prerenderRoutes(['/', '/about'])
...
+
+// Somewhere later in App
+const articleContent = await $fetch('/api/content/article/name-of-article', {
+  responseType: 'json',
+})
+```
+
+::warning
+Prerendered API routes in production may not return the expected response headers, depending on the provider you deploy to. For example, a JSON response might be served with an `application/octet-stream` content type.
+Always manually set `responseType` when fetching prerenderered API routes.
```
danielroe, line 45:
```suggestion
Always manually set `responseType` when fetching prerendered API routes.
```

danielroe:
thank you!

Gwynerva:
@danielroe I think the part about breaking the app should be readded. Because this is how I initially stumbled with all this missing headers stuff. `npx serve` does automatically recognise JSON content, but for example GitHub Pages is not. And I got a broken app with zero errors in console (+ hydration mismatch). Spent arount 5-6 hours tearing my hair until found the real cause of the problem.

So the solutions are basically: 1. Set a manual extension to api routes `/api/article/my-article.json` or set `responseType`.

What do you think?

danielroe:
It's a provider-specific limitation so I muted the warning somewhat - it's not a general issue with prerendering routes. But it does still say that you should add `responseType`.

I'd be up for improving the wording.

cc: @pi0 if there's a better way we can talk about this.

pi0:
There is an open issue (https://github.com/nitrojs/nitro/issues/2119) to add generic way of persisting prerender headers.

PR #31244, opened 2025-03-06
fix(nuxt): navigateToOptions replace not keeping query/hash

### LINKED ISSUE #31236: Calls to `navigateTo` with `replace: true` option in middleware lead to losing query parameters (if the path was passed via a string)
### Environment

nuxt >=3.15.0

### Reproduction

In an empty project add such a global middleware:

```typescript
export default defineNuxtRouteMiddleware(async (to) => {
    if (to.path !== '/some-url') {
        return navigateTo('/some-url?a=b&c=d', { replace: true })
    }
})
```

Also add an empty page `/some-url`.

Open the root page. You will be redirected to `/some-url`, while `/some-url?a=b&c=d` is expected.

### Describe the bug

The bug appeared in the PR https://github.com/nuxt/nuxt/pull/30283 and in nuxt 3.15.0.

Internally, before performing navigation, nuxt transforms the string which we pass to `navigateTo` and which contains a **full path**.

The string is being transformed to an object with `path` property, which is not intended to store an information about query parameters (and, perhaps, a hash).

murshex:
### 🔗 Linked issue

resolves #31236 

### 📚 Description

`navigateTo` arg1 (`RouteLocationRaw`) might be a string, so we have to parse it to extract the query/hash, and then send it off to vue router.

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -152,7 +152,7 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
   if (import.meta.client && !isExternal && inMiddleware) {
     if (options?.replace) {
       if (typeof to === 'string') {
-        const { pathname, search, hash } = parseURL(fullPath)
+        const { pathname, search, hash } = parseURL(to)
         return { path: pathname, query: parseQuery: (search), hash, replace: true }
       }
       return { ...to, replace: true }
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/composables/router.ts
```diff
@@ -151,7 +151,11 @@ export const navigateTo = (to: RouteLocationRaw | undefined | null, options?: Na
   // Early redirect on client-side
   if (import.meta.client && !isExternal && inMiddleware) {
     if (options?.replace) {
-      return typeof to === 'string' ? { path: to, replace: true } : { ...to, replace: true }
+      if (typeof to === 'string') {
+        const { pathname, search, hash } = parseURL(to)
+        return { path: pathname, query: parseQuery(search), hash, replace: true }
```
danielroe, line 156: we should probably omit hash/query if they are not present

murshex: Ok done

danielroe (approved):

Issue #31259, opened 2025-03-07
Wrong warn is given during `useFetch` or `useLazyFetch` execution on the server-side.


WonderBerry:
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.17.0`
- Nuxt Version:     `4.0.0-0`
- CLI Version:      `3.22.4`
- Nitro Version:    `-`
- Package Manager:  `pnpm@10.5.2`
- Builder:          `-`
- User Config:      `modules`, `pages`, `dir`, `typescript`
- Runtime Modules:  `()`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-pmvrs6vw?file=app.vue

### Describe the bug

Wrong warn is given during `useFetch` or `useLazyFetch` execution on the server-side.
It was mentioned here https://github.com/nuxt/nuxt/issues/31211 but without fix in useLazyFetch.

### Logs

```shell-script
WARN  [nuxt] useAsyncData must return a value (it should not be undefined) or the request may be duplicated on the client side.
```

PR #31213, opened 2025-03-05
fix: useFetch in middleware that returns undefined gives useAsyncData warning

moshetanzer:
### 🔗 Linked issue
resolves https://github.com/nuxt/nuxt/issues/31211
resolves https://github.com/nuxt/nuxt/issues/31259
closes https://github.com/nuxt/nuxt/pull/31260

### 📚 Description
Should fix above issue not sure what perf cost is

PR #31260, opened 2025-03-07
fix: wrong warn is given during `useFetch` or `useLazyFetch` execution on the server-side.

WonderBerry:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/31259

### 📚 Description

Fix warning message during useLazyFetch/useFetch execution


danielroe:
this is addressed also by https://github.com/nuxt/nuxt/pull/31213, but you caught a case that one didn't. (merging into that PR)

PR #31273, opened 2025-03-07
test: migrate spa preloader tests to playwright

### LINKED ISSUE #30867: flaky test for spa-preloader
needs to be investigated - probably not a terribly complex fix

also `test client-side navigation` for server islands

Kornil:
### 🔗 Linked issue

Resolves #30867 

### 📚 Description

I was investigating this issue, and I believe the flakiness comes from expecting the loading state 

```js
    await page.goto(url('/spa'), { waitUntil: 'domcontentloaded' })

>>    await page.getByTestId('loader').waitFor({ state: 'visible' })
>>    expect(await page.getByTestId('content').isHidden()).toBeTruthy()

    await page.getByTestId('content').waitFor({ state: 'visible' })
    expect(await page.getByTestId('loader').isHidden()).toBeTruthy()
```

What I've noticed in our webapp is that the loading can happen fast enough that tests will flake waiting for it while it already passed undetected. 

There is similar code in both 
`test/spa-loader/spa-preloader-outside-disabled.test.ts`
`test/spa-loader/spa-preloader-outside-enabled.test.ts`

The first file already has a `loader` check so it should not be a problem, should it be implemented on the second file too?

Removing it does not seem to fail anything on `pnpm test`.

Let me know if you have feedback.


danielroe (commented):
i understand this makes the tests pass but the purpose of this test is to assert that a loader is displayed and (except in the case of ‘within’) doesn’t disappear before the content loads)

maybe we could replace it with an assertion about the content of the html rendered with $fetch?

Kornil:
Thank you for the feedback.

- I addressed the `$fetch` test in 2522804ee118ab67025974ede48f5f1f264685a2 so we can see loading is correctly displayed (and content isn't) on page load, and later when content is displayed, there's no more loading.
- I also addressed a flake I didn't notice on ab2285068a2599c3f159151778619c6df2b248e0, it was checking content twice expecting to not be visible and then visible after, now it should not flake anymore.

Let me know if I addressed everything.

danielroe:
I agree we should migrate these (and other) tests to playwright.

However I'm still not sure about these changes. the tests are meant to verify that the loader does display and that it displays up until the moment that the content displays - ie. no white flash while hydration takes place. probably the test needs to be rewritten, but we shouldn't just get rid of the assertions meant to cover these possibilities.

danielroe:
I've deferred the finishing of hydration until a synthetic event is dispatched, which lets us inspect the intermediate state. this seems more stable.

danielroe:
the expect assertion has built-in anti-flake so it will _wait_ until that state is true. that's great for most test situations, but we want to assert the state at a moment in time. if we wait at all we might end up getting a false positive because it 'skips' over an intervening state that we do not desire to happen at all.

Kornil:
I see what you mean now, we want the "unsafe" behavior for this edge case. Thanks for the clarification.

Is there anything else that needs to be added to this PR?

danielroe (approved):

PR #31295, opened 2025-03-08
fix(nuxt): use new `mocked-exports`

danielroe:
### 📚 Description

this helps avoid issues with hoisting on unenv


packages/vite/src/client.ts
```diff
@@ -8,9 +8,12 @@ import { logger } from '@nuxt/kit'
 import { getPort } from 'get-port-please'
 import { joinURL, withoutLeadingSlash } from 'ufo'
 import { defu } from 'defu'
-import { env, nodeless } from 'unenv'
+import { defineEnv } from 'unenv'
+import { resolveModulePath } from 'exsolve'
 import { defineEventHandler, handleCors, setHeader } from 'h3'
 import type { ViteConfig } from '@nuxt/schema'
+import { resolveModulePath } from 'exsolve'
```
danielroe, line 15:
```suggestion
```

PR #31306, opened 2025-03-09
fix(nuxt): strip out extractedKeys from page meta

### LINKED ISSUE #31194: Route `meta` inconsistencies when used with `extraPageMetaExtractionKeys`
### Environment

- Operating System: `Darwin`
- Node Version:     `v22.13.1`
- Nuxt Version:     `4.0.0-0`
- CLI Version:      `3.22.2`
- Nitro Version:    `-`
- Package Manager:  `pnpm@10.5.0`
- Builder:          `-`
- User Config:      `modules`, `pages`, `dir`, `typescript`
- Runtime Modules:  `()`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-jrkfltvz

This reproduction sets 

```ts
{
  experimental: {
    extraPageMetaExtractionKeys: ['custom'],
  },
}
```

https://github.com/user-attachments/assets/374149af-374a-4f01-b3b4-18978cde6cd6

### Describe the bug

The meta data available on the VueRouter route of a page is different, depending on the keys present in `definePageMeta` and `extraPageMetaExtractionKeys` configuration.

If only "unknown" keys are used (`foo` in the reproduction) that are not listed in `extraPageMetaExtractionKeys`, those keys are available in the `meta` of the VueRouter route. If a mix of "unknown" keys and keys listed as extra extraction keys (`foo` and `custom` in the reproduction) are used, all keys show up in the runtime meta. 
However, if only keys from `extraPageMetaExtractionKeys` are used they don't show up in the runtime meta. 

Is this intended behavior? 

I was able to find this commit that explicitly excludes extraction keys from the route meta (https://github.com/nuxt/nuxt/commit/ae9f42f4aaf4204597f2b4eeb5f3d9022c0d1911), which happens to coincide with this PR (https://github.com/nuxt/nuxt/pull/30015) implementing support for the `extraPageMetaExtractionKeys`. Before this PR `extractionKeys` in `getRouteMeta` only contained the default Nuxt extraction keys, afterwards it also includes user defined. So I'm unable to figure out if this is an intentional omission of *all* extraction keys, or an oversight, because those two changes happened to coincide.

Nonetheless, the inconsistency between sometimes including the key, because it is part of the dynamic meta import and sometimes omitting it feels like a potential problem.

If this is an unintended omission, and the extra keys by the user should be included in the runtime route meta, I can contribute a PR for this issue.

### Additional context

_No response_

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

fix #31194 

### 📚 Description

This PR process the meta object to remove extracted keys


danielroe:
Just thinking aloud - we removed the extracted keys from the meta object if they were going to be available on the route object (e.g. alias, etc.). So they shouldn't _also_ be accessible on `route.meta`.

But I think it makes sense for the other keys to be accessible on route.meta at runtime (and also modifiable if changed in `pages:resolved`). They should still likely be stripped from the virtual module (as you do here) and then added back in `routes.mjs` so they are accessible at runtime from `route.meta`.

We should add a test to cover that. What do you think?

huang-julien:
Do you mean we should add a test to verify whether non-extracted keys are available in route meta within the virtual module imported by route.mjs ?

danielroe:
Even after this PR (not a criticism, as I think some of this is broken in main branch too)

- currently non-serialisable values still seem to be stripped out - e.g. functions
- values do not remain accessible on route.meta - this should likely be added back in [normalizeRoutes](https://github.com/nuxt/nuxt/blob/dfc66beeffe9b39ddcb9883d00e6b9d7f727ee30/packages/nuxt/src/pages/utils.ts#L505-L625) and merged with any of the meta that couldn't be extracted

danielroe (approved):

PR #31317, opened 2025-03-10
refactor(nuxt): simplify nuxt-layout return vnode

xjccc:
### 📚 Description

Is it correct to simplify the code? If it's not necessary, i will close this. Thank you❤

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-layout.ts
+++ b/packages/nuxt/src/app/components/nuxt-layout.ts
@@ -132,7 +132,7 @@ const LayoutProvider = defineComponent({
     if (import.meta.dev && import.meta.client) {
       onMounted(() => {
         nextTick(() => {
-          if (['#comment', '#text'].includes(vnode?.el?.nodeName)) {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-layout.ts
```diff
@@ -151,22 +151,12 @@ const LayoutProvider = defineComponent({
...
-        vnode = h(
-          LayoutLoader,
-          { key: name, layoutProps: props.layoutProps, name },
-          context.slots,
-        )
-
-        return vnode
-      }
-
-      return h(
```
danielroe, line 165: sorry for taking so long to review!

the reason I did this initially was to avoid memory leaks in production. is there any issue this is causing?

xjccc: Oh... there's nothing wrong with this. Just look at this by accident and just edit it for my opinion. No matter this, if unnecessary, close it please. Thank you for your feedback.❤

danielroe: okay, I'll close it then, but thank you ❤️

Issue #28710, opened 2024-08-26
`useRoute.matched` is not returning all matching routes


Aietes:
### Environment

- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.13.0
- CLI Version:      3.13.0
- Nitro Version:    2.9.7
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      compatibilityDate, future, devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-ezruhh?file=app%2Fapp.vue

### Describe the bug

With `future: { compatibilityVersion: 4:}` set in `nuxt.config.ts`, `useRouter.matched` is returning only 1 item on dynamic routes, e.g. `/app/pages/[id].vue`. This is unexpected behavior, as `$route.matched` should be an array of normalized matched routes with current route location, in this case it should contain two items. This bug breaks components depending on this property, e.g. NuxtLink activeClass stops working for dynamic routes.

danielroe:
This doesn't seem to be related to `future.compatibilityVersion`. Did you have a version of Nuxt + vue-router that behaves as you expect?

updated repro to show same behaviour on previous version of nuxt/vue-router:

https://stackblitz.com/edit/github-ezruhh-g2prlc

Aietes:
It worked before we changed to future, but as your repro shows, it doesn't seem to be the root cause. We noticed NuxtLink is not working as expected, as activeClass is not triggered on dynamic subroutes like it used to.

Aietes:
We found out why it worked before, it had nothing to do with versions. We had an additional `page1.vue` in pages, which rendered the `index.vue` as `NuxtPage`, something like this:

```
-| pages/
---| page1/
------| [id].vue
------| index.vue
---| page1.vue
```
> `page1.vue`
```html
<template>
  <NuxtPage />
</template>
```

With that, `$route.matched` contains two objects, and NuxtLink works. It's an accidental workaround. However, this is still a bug, as actually there should be 3 matches when looking at the routes:

![image](https://github.com/user-attachments/assets/5a16a585-8dec-4c86-b230-0751d17bb42c)

yapengsoft:
I encountered the same issue

Aietes:
Is there any update on this? It's still in triage, I think this should be confirmed as a bug?

andreclemente:
[...]

PR #31331, opened 2025-03-11
fix(nuxt): prevent typed pages param duplication in `RouteNamedMap`

Anoesj:
## 🔗 Linked issue
No issue yet, I ran into this myself and skipped right to finding a solution.

## 📚 Description
This is a bit of preparative work for making `useRoute()` automatically typed correctly in Nuxt pages when `typedPages: true`. When a route overrides a page's path using `definePageMeta()`, that new path is passed on to `unplugin-vue-router` directly, instead of passing the original path first and then updating the route's `path` afterwards, which `unplugin-vue-router` stores as an 'override'. This is more in line with how `definePage` works in `unplugin-vue-router`.

#### Before these changes:
Duplicate params in the route tree cause the `RouteRecordInfo` object for route `param-id-view-custom` in `typed-router.d.ts` to contain the `id` param twice.
```ts
/* eslint-disable */
/* prettier-ignore */
// @ts-nocheck
// Generated by unplugin-vue-router. ‼️ DO NOT MODIFY THIS FILE ‼️
// It's recommended to commit this file.
// Make sure to add this file to your tsconfig.json file as an "includes" or "files" entry.

declare module 'vue-router/auto-routes' {
  import type {
    RouteRecordInfo,
    ParamValue,
    ParamValueOneOrMore,
    ParamValueZeroOrMore,
    ParamValueZeroOrOne,
  } from 'vue-router'

  /**
   * Route name map generated by unplugin-vue-router
   */
  export interface RouteNamedMap {
    'param-id': RouteRecordInfo<'param-id', '/param/:id()', { id: ParamValue<true> }, { id: ParamValue<false> }>,
    'param-id-view-custom': RouteRecordInfo<'param-id-view-custom', '/param/:id()/view-custom', { id: ParamValue<true>, id: ParamValue<true> }, { id: ParamValue<false>, id: ParamValue<false> }>,
  }
}
```

#### After these changes:
No more duplication.
```ts
/* eslint-disable */
/* prettier-ignore */
// @ts-nocheck
// Generated by unplugin-vue-router. ‼️ DO NOT MODIFY THIS FILE ‼️
// It's recommended to commit this file.
// Make sure to add this file to your tsconfig.json file as an "includes" or "files" entry.

declare module 'vue-router/auto-routes' {
  import type {
    RouteRecordInfo,
    ParamValue,
    ParamValueOneOrMore,
    ParamValueZeroOrMore,
    ParamValueZeroOrOne,
  } from 'vue-router'

  /**
   * Route name map generated by unplugin-vue-router
   */
  export interface RouteNamedMap {
    'param-id': RouteRecordInfo<'param-id', '/param/:id()', { id: ParamValue<true> }, { id: ParamValue<false> }>,
    'param-id-view-custom': RouteRecordInfo<'param-id-view-custom', '/param/:id()/view-custom', { id: ParamValue<true> }, { id: ParamValue<false> }>,
  }
}
```

### Related PRs:
- `vue-router`: https://github.com/vuejs/router/pull/2475
- `unplugin-vue-router`:
  - https://github.com/posva/unplugin-vue-router/pull/602
  - https://github.com/posva/unplugin-vue-router/pull/601

#### Adjust tests
We'll need to adjust the tests in this PR if the PRs above get merged and new versions of `unplugin-vue-router` and `vue-router` are published.

#### Add type-level route file path ↔ route name map
Either append a type-level map of route file paths mapped to their corresponding route names to the generated DTS file (`.nuxt/types/typed-router.d.ts`) or write to a different `.d.ts` file after aforementioned DTS file is generated by `unplugin-vue-router`.

#### Add plugin for Vue Language Tools
Create a Vue Compiler plugin that augments `useRoute()` in the "virtual code" of Vue Language Tools and add that plugin to `.nuxt/tsconfig.json` if `typedPages: true`:
```json
{
  "vueCompilerOptions": {
    "plugins": ["/path-to-plugin.js"]
  }
}
```

This is why I think we need above-mentioned type-level route file path ↔ route name map. We need to know the route name of the currently edited Nuxt page component in order to make `useRoute()` typed properly in the virtual code. I'm probably going to need a bit of help with this part.

### Sidenote
This is one of my first contributions for Nuxt, so I probably did some ugly things here and there. The added test works, but it's probably in the wrong place?


Anoesj:
@danielroe I see some tests are failing, which makes sense, because those tests don't expect the presence of the `_pathBeforeAugmented` property. Instead of saving the original paths to `_pathBeforeAugmented`, could we maybe save the pre-augment paths to the `NuxtApp` instance, or some other context object instead? The output would be a bit cleaner, but I don't know much about the lifecycle of a `NuxtApp` instance, so I could use some advice first.

Anoesj:
Not sure if the `Nuxt` interface is the right place, but now tests shouldn't fail anymore? This also opens up the possibility to alter `nuxt._augmentPagesContext` in a `pages:extend` hook if needed.

Anoesj:
**Sidenote:** this is a bit of preparative work for making `useRoute()` automatically typed correctly in Nuxt pages when `typedPages: true`. 

#### Related (future) PRs:
- `vue-router`: https://github.com/vuejs/router/pull/2475
- `unplugin-vue-router`:
  - adjust DTS file generation (TBA)
  - https://github.com/posva/unplugin-vue-router/pull/601

#### Future steps:
- Either append a type-level map of route file paths mapped to their corresponding route names to the generated DTS file (`.nuxt/types/typed-router.d.ts`) or write to a different `.d.ts` file after aforementioned DTS file is generated by `unplugin-vue-router`.
- We'll need to adjust the tests in this PR if the PRs above get merged and new versions of `unplugin-vue-router` and `vue-router` are published.

danielroe:
I think there's probably a better way to ensure that we don't add duplicate params in the route tree. Perhaps the issue is that we're not properly sorting routes in the tree (see https://github.com/nuxt/nuxt/issues/28710).

packages/nuxt/src/pages/module.ts
```diff
@@ -202,9 +202,17 @@ export default defineNuxtModule({
...
 
-            // @ts-expect-error TODO: either fix types upstream or figure out another
             // way to add a route without a file, which must be possible
-            const route = addedPagePaths.has(absolutePagePath) ? parent : parent.insert(page.path, page.file)
+            const route = addedPagePaths.has(absolutePagePath)
+              ? parent
+              : /^\//.test(page.path)
+                // @ts-expect-error TODO: either fix types upstream or figure out another
+                // way to add a route without a file, which must be possible
+                ? rootPage.insert(page.path, page.file)
```
danielroe, line 211: this is the key thing - by inserting the page at the root, uvr is able to correctly map the route to the right place in the tree.

not sure if this is a bug in uvr or not, but it does resolve the issue for nuxt

danielroe (approved):

packages/nuxt/src/pages/module.ts
Anoesj: Interesting. I'll check it out locally in a bit.

Slightly related: I created this issue for the `@ts-expect-error` comments: https://github.com/posva/unplugin-vue-router/pull/607

packages/nuxt/src/pages/module.ts
```diff
@@ -185,56 +185,44 @@ export default defineNuxtModule({
...
+                // way to add a route without a file, which must be possible
+                : parent.insert(page.path, page.file)
 
             addedPagePaths.add(absolutePagePath)
-
             if (page.meta) {
               route.addToMeta(page.meta)
             }
-            if (page.alias !== undefined) {
+            if (page.alias) {
```
Anoesj: Possible edge case: what if you don't have an `index.vue`, but some other page has `alias` set to `''`?

danielroe: that seems like a different PR, perhaps? Shouldn't alias be set to `/` rather than an empty string?

Anoesj: Oh that's right 👍🏼

Magkoooh (approved):

PR #31353, opened 2025-03-13
feat(docs): publish raw markdown/yaml docs as `@nuxt/docs`

antfu:
This is a proposal that we should publish our docs together with the monorepo as an additional resource to be consumed by other integrations, namely LLM agents (https://github.com/antfu/nuxt-mcp), but potentially other tools like Nuxt DevTools to present offline docs, etc.

The benefit is that the version of the docs is explicitly matched to the runtime versions. Eliminate the problem of version mismatch between installed and online docs. It also allows consuming those data offline without API requests.

I feel that all the projects with docs should adapt to this (if we agree on this, I'd also do the same for other packages like VueUse, UnoCSS, etc.)


docs/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@nuxt/docs",
+  "version": "4.0.0-0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nuxt/nuxt.git",
+    "directory": "packages/docs"
```
HigherOrderLogic, line 7:
```suggestion
    "directory": "docs"
```

Shouldnt this be `docs`?

danielroe:
do you think it would be possible to run twoslash to augment docs at build time?

at the moment two slash often type errors because the nuxt/nuxt.com environment or nuxt version is different, and that will be increased when we support displaying different docs versions on the site

antfu:
That's an interesting idea! It's definitely possible, let me see how it can be built! I created https://github.com/twoslashes/twoslash/issues/67 to track on it

danielroe:
I agree this is an excellent change, and I think we can also migrate from the `nuxt.config` markdown currently present in `@nuxt/schema` to include this in `@nuxt/docs` instead.

after first publish, we should also update nuxt/nuxt.com to pull from the new `@nuxt/docs` package (cc: @HugoRCD @atinux)

HugoRCD:
> I agree this is an excellent change, and I think we can also migrate from the `nuxt.config` markdown currently present in `@nuxt/schema` to include this in `@nuxt/docs` instead.
> 
> after first publish, we should also update nuxt/nuxt.com to pull from the new `@nuxt/docs` package (cc: @HugoRCD @atinux)

Really cool idea, I had already added [nuxt-llms](https://github.com/nuxtlabs/nuxt-llms) to the new version of the doc but this would clearly be even better!

atinux:
This is now I believe it's also the right time to stop generating the nuxt configuration file on nuxt.com (see https://github.com/nuxt/nuxt.com/blob/main/server/plugins/docs-config.ts) as it will make sense to have it directly part of the `@nuxt/docs` package

danielroe:
Excellent - sounds like we are agreed.

One more important comment for a future PR (cc: @HugoRCD):

right now the docs are dependent on components in nuxt/nuxt.com, including components from nuxt/ui-pro. that means that even if we publish separately it's still not truly consumable anywhere but nuxt/nuxt.com

I think it would be good to consider making the docs 'pure' (or bundling all components used inside them?) so that they are truly independent. added to that, they would not need to be updated if/when nuxt/ui or nuxt/ui-pro is updated, meaning we wouldn't need to coordinate updates to the docs markdown source and the website (e.g. https://github.com/nuxt/nuxt/pull/31150).

any transforms that use components from the nuxt/nuxt.com codebase could be applied in a plugin or module in nuxt/nuxt.com itself.

atinux:
Agree! I think we can share the same Undocs content transformers so the `@nuxt/docs` can use as less as possible the MDC syntax. 

https://undocs.pages.dev/guide/content-transformation

danielroe (approved):

PR #31356, opened 2025-03-13
chore(deps): bump nitro nightly version

danielroe:
### 📚 Description

migrate to the updated v3 branch 🚀


package.json
```diff
@@ -56,10 +56,10 @@
     "@vue/compiler-dom": "3.5.13",
     "@vue/shared": "3.5.13",
     "c12": "3.0.2",
-    "h3": "npm:h3-nightly@1.15.1-20250222-111608-d1c00fc",
+    "h3": "1.15.1",
```
productdevbook, line 59: nitro v3 used h3-v2 ? `[2.0.0-20250312-144942-506a98f](https://www.npmjs.com/package/h3-nightly/v/2.0.0-20250312-144942-506a98f)` or no ?

danielroe: not yet

danielroe:
cc: @antfu @pi0 - I've [disabled](https://github.com/nuxt/nuxt/pull/31356/commits/1d39766a9822e644bd66be8aac195c937f46b97f) twoslash testing on nitro plugin example code because this is set in a different 'context' - these types are not available to nuxt. (in fact, I'm not sure how it worked previously). this is probably therefore a type fix....

though ... we might try to figure out how we can run _some_ nuxt examples in a nuxt context, and _some_ in a nitro context - can we select what `tsconfig.json` to apply to each one? or somehow respect the file path of the code example?

PR #31365, opened 2025-03-14
feat(nuxt): warn server components need root element

### LINKED ISSUE #31354: Standalone server component fails with "500 clone.querySelectorAll is not a function"
### Environment

```
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.16.0
- CLI Version:      3.22.5
- Nitro Version:    2.11.6
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/github-t9znupwu

### Describe the bug

Steps to reproduce
1. Create `components/MyComponent.server.vue`
2. Include it in `app.vue`

### Additional context

Image

### Logs

```shell-script

```

xjccc:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31354

### 📚 Description

In docs tip

image

Add a simple warn for this. And if @huang-julien has some best deals with this. I will close this. Thank you❤

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -1,4 +1,4 @@
-import type { Component, PropType, VNode } from 'vue'
+import type { Component, PropType, RendererNode, VNode } from 'vue'
 import { Fragment, Teleport, computed, createStaticVNode, createVNode, defineComponent, getCurrentInstance, h, nextTick, onBeforeUnmount, onMounted, ref, toRaw, watch, withMemo } from 'vue'
 import { debounce } from 'perfect-debounce'
 import { hash } from 'ohash'
@@ -12,7 +12,7 @@ import type { NuxtIslandResponse } from '../types'
 import { useNuxtApp, useRuntimeConfig } from '../nuxt'
 import { prerenderRoutes, useRequestEvent } from '../composables/ssr'
 import { injectHead } from '../composables/head'
-import { getFragmentHTML, isStartFragment } from './utils'
+import { getFragmentHTML, isEndFragment, isStartFragment } from './utils'
 
 // @ts-expect-error virtual file
 import { appBaseURL, remoteComponentIslands, selectiveClient } from '#build/nuxt.config.mjs'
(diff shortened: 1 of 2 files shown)


huang-julien (changes requested):
getFragmentChildren isn't only used by server components.

We can either do it at build time while transforming islands SFC or at runtime... 
maybe in the renderer server side 🤔

danielroe (dismissed):

xjccc:
> getFragmentChildren isn't only used by server components.
> 
> We can either do it at build time while transforming islands SFC or at runtime... maybe in the renderer server side 🤔

Emm ..I don't know if I'm right ? Thank you for your feedback

huang-julien:
I think I get the reason of you PR.
https://github.com/nuxt/nuxt/issues/31354 doesn't need a fix as it is not a bug.

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -130,6 +130,20 @@ export default defineComponent({
     const ssrHTML = ref<string>('')
 
     if (import.meta.client && instance.vnode?.el) {
+      if (import.meta.dev) {
+        let currentEl = instance.vnode.el
+        while (currentEl) {
+          // find first non-comment node
+          if (!isStartFragment(currentEl)) {
+            // if don't have root element
+            if (currentEl.nodeType !== 1) {
```
huang-julien, line 139: This won't verify cases like `<div></div><div></div>` which is also invalid for islands 🤔

huang-julien (changes requested):

packages/nuxt/src/app/components/nuxt-island.ts
xjccc: 🤔Do you have any idea for this?

xjccc: I have changed the code. Compare startEl with endEl's previousSibling as the same. What do you think about it?

huang-julien (approved):
Thank you for your PR :)

PR #23993, opened 2023-10-27
test(nuxt): add failing test for shared `asyncData`

or2e:
### 🔗 Linked issue
#23522

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description
Valid test cases have been added, which fail due to regression since version 3.7, more details in [issue](https://github.com/nuxt/nuxt/issues/23522)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #15438, opened 2022-11-10
Cleanup payload in memory cache when not needed


pi0:
When we are sure payload data is not going to be reused, we can drop it. It can both be done for `nuxt._asyncData` (like nuxt/framework#7265) and `nuxt.payload.data`.

PR #31373, opened 2025-03-14
fix(nuxt): use single synced asyncdata instance per key

### LINKED ISSUE #21532: useFetch & useAsyncData reactive key

### LINKED ISSUE #22348: PayloadExtraction/useFetch persists data when it shouldn't
### Environment

- Operating System: Linux
- Node Version:     v16.20.0
- Nuxt Version:     3.6.5
- Nitro Version:    2.5.2
- Builder:          vite
- User Config:      devtools, experimental, nitro
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-4ym9i1?file=pages%2F%5Bid%5D%2Findex.vue

### Describe the bug

To reproduce:
1. yarn run build
2. node .output/server/index.mjs
3. Go to a product and select a size
4. Go back
5. Go to same product as in step 3

You'll then see the size is not the default size.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #23522: useAsyncData - undocumented behavioral change
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

[3.6.5](https://stackblitz.com/edit/nuxt-starter-vetznc)
[3.7](https://stackblitz.com/edit/nuxt-starter-n6ggsy)

### Describe the bug

When upgrading from `3.6.5` to `3.7.x`, we found a change in the behavior of the `useAsyncData` composable

Behavior has changed for `useAsyncData` with options `server: false, immediate: false` 
We want to load some data only on the client side in the `onMounted` lifecycle hook (e.g. session-dependent data like the shopping cart)

Code examples you can see in the reproduction section

So, we load data into `app.vue`, and in another component we want to use `pending` to show to show skeleton-loading
In `3.6.5` this method worked, starting from `3.7.0` it stopped working. 
Most likely it is related to [this edit](https://github.com/nuxt/nuxt/pull/20980/files#diff-6b54d0385d537fe28bf4bca3e034a7de05f5a2fef365d3c1052a119df884313bR148)

So a few questions

1/ There is another way to know the `pending` state of a specific `useAsyncData`
2/ Is this a bug or a feature? If so, then such tasks will have to be solved with several utilitarian `useState`, e.g.:
```
useCartData = () => useState('key1', () => []))
useCartPending = () => useState('key2', () => true)
useLoadCartData = async () => (useCartData().value = await $fetch(....))
```

### LINKED ISSUE #24332: Re-use `getCachedData` when `refresh`ing `asyncData`

### LINKED ISSUE #26733: When using useFetch[post], response data will be bound to the same one.
### Environment

- Operating System: Darwin
- Node Version:     v20.9.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-57k4ba-aqkyrc?file=app.vue
- Different response when initializing.
- Click button then update param to 'new'
- Both components' responses are updated to 'new'.

### Describe the bug

The useFetch[post] within the components is parameterized based on props. While both components pass different parameters through props, when updating the props, the response for both components is updated to the same one.

### Additional context

It's strange that useFetch[get] doesn't exhibit the same behavior. Here's the link: https://stackblitz.com/edit/github-57k4ba-ddzfhp?file=components%2FFetchGet.vue.

### Logs

_No response_

### LINKED ISSUE #27204: Resetting async data on scope dispose
### Describe the feature

Currently, async data isn't reset on scope dispose, which I believe should be the case (same way watchers get disabled). This might have some unwanted consequences I am currently unaware of.

### Additional information

### Final checks

### LINKED ISSUE #27552: status of useLazyAsyncData for cached requests is success on router enter
### Environment

- Operating System: Darwin
- Node Version:     v20.11.0
- Nuxt Version:     3.11.2
- CLI Version:      3.11.1
- Nitro Version:    2.9.6
- Package Manager:  npm@10.2.4
- Builder:          -
- User Config:      srcDir, experimental, runtimeConfig, app, components, imports, devtools, typescript, modules, eslint, fontMetrics, i18n, tailwindcss, postcss, css, alias, vite, build, sourcemap, hooks
- Runtime Modules:  @nuxtjs/critters@0.7.1, @nuxtjs/fontaine@0.4.2, @nuxtjs/i18n@8.3.1, @nuxtjs/tailwindcss@6.12.0, @pinia/nuxt@0.5.1, @vueuse/nuxt@10.10.0, nuxt-lazy-hydrate@1.0.0, nuxt-viewport@2.1.5, @nuxt/eslint@0.3.13
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-k9azxd

Steps to reproduce:
1. Load index page
2. Click on link "different page" 
    watcher for "status" show states in order of "idle -> pending -> success" 
3. Click on link "index"
4. Click on link "differen page"
    watcher for "status" now shows  "success -> pending -> success"
    
Same is for pending which is initially "false" on step 4.     

### Describe the bug

On second visit of a route with  top level `useLazyAsyncData` in script setup, `status` and `pending` state seem to be incorrect. I would expect the initial state for status to be `idle` and for pending to be `true`. 

As this is only the case for the second visit it might related to the caching and is intended behavior?

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/21532
resolves https://github.com/nuxt/nuxt/issues/24332 and therefore closes https://github.com/nuxt/nuxt/pull/25850
resolves https://github.com/nuxt/nuxt/issues/22348
resolves https://github.com/nuxt/nuxt/issues/27552
resolves https://github.com/nuxt/nuxt/issues/23522 and therefore closes https://github.com/nuxt/nuxt/pull/23993
resolves https://github.com/nuxt/nuxt/issues/27204
resolves https://github.com/nuxt/nuxt/issues/26733

partly implements https://github.com/nuxt/nuxt/issues/15438

### 📚 Description

This PR is a major reorganisation of the data fetching layer in Nuxt, providing performance/memory improvements + increased consistency.

#### 1. `getCachedData` behavior change

The `getCachedData` option now:

- Always gets called before refetching data, even when called by `watch` or `refreshNuxtData`
- It receives a context object with additional information in a `cause` property
- Can make more granular decisions about serving cached data

```ts
// Before:
useAsyncData('users', fetchUsers, {
  getCachedData: (key, nuxtApp) => {
    return nuxtApp.isHydrating 
      ? nuxtApp.payload.data[key] 
      : nuxtApp.static.data[key]
  }
})

// Now:
useAsyncData('users', fetchUsers, {
  getCachedData: (key, nuxtApp, ctx) => {
    // ctx.cause can be 'initial' | 'refresh:hook' | 'refresh:manual' | 'watch'
    
    // Example: Don't use cache on manual refresh
    if (ctx.cause === 'refresh:manual') return undefined
    
    return cachedData[key]
  }
})
```

#### 1. Shared refs for the same key

All calls to `useAsyncData` or `useFetch` with the same key will now share not just the underlying data but also `data`, `error`, and `status` refs. This ensures consistency across components but may affect code that expected isolated instances.

```ts
// Before: These would be independent instances with separate refs
const { data: users1, status: status1 } = useAsyncData('users', () => $fetch('/api/users'))
const { data: users2, status: status2 } = useAsyncData('users', () => $fetch('/api/users'))

// Now: Both reference the same underlying state
// Any changes to one will affect the other
```

#### 2. Warnings for inconsistent options

Multiple calls to `useAsyncData` with the same key but different options will now trigger development warnings. The following options **must be consistent** across all calls with the same key:

- `handler` function
- `deep` option
- `transform` function
- `pick` array
- `getCachedData` function
- `default` value

The following options **can differ** without triggering warnings:

- `server`
- `lazy`
- `immediate`
- `dedupe`
- `watch`

```ts
// ❌ This will trigger a warning
const { data: users1 } = useAsyncData('users', () => $fetch('/api/users'), { deep: false })
const { data: users2 } = useAsyncData('users', () => $fetch('/api/users'), { deep: true })

// ✅ This is allowed
const { data: users1 } = useAsyncData('users', () => $fetch('/api/users'), { immediate: true })
const { data: users2 } = useAsyncData('users', () => $fetch('/api/users'), { immediate: false })
```

#### 1. Reactive keys

You can now use computed refs, plain refs or getter functions as keys, allowing for dynamic data fetching that automatically updates when dependencies change:

```ts
// Using a computed property as a key
const userId = ref('123')
const { data: user } = useAsyncData(
  computed(() => `user-${userId.value}`),
  () => fetchUser(userId.value)
)

// When userId changes, the data will be automatically refetched
// and the old data will be cleaned up if no other components use it
userId.value = '456'
```

#### 2. Deduped watch calls

Multiple components watching the same data source (like route changes) will now trigger only a single data refetch:

```ts
// In ComponentA.vue
const { data: users } = useAsyncData(
  'users', 
  () => $fetch(`/api/users?page=${route.query.page}`),
  watch: [() => route.query.page]
)

// In ComponentB.vue
const { data: users } = useAsyncData(
  'users', 
  () => $fetch(`/api/users?page=${route.query.page}`),
  watch: [() => route.query.page]
)

// When route.query.page changes, only one fetch will be performed
```

#### For getCachedData users

If you were using `getCachedData`, update your implementation to handle the new context parameter:

```ts
// Update your getCachedData implementation
useAsyncData('key', fetchFunction, {
  getCachedData: (key, ctx) => {
    // Handle the context object
    if (ctx.cause === 'refresh:manual') {
      // Skip cache on manual refresh
      return
    }
    
    return yourCacheImplementation.get(key)
  }
})
```

Alternatively, for now, you can disable this behaviour with:

```ts twoslash [nuxt.config.ts]
  experimental: {
    granularCachedData: false,
    purgeCachedData: false
  }
})
```

#### For duplicate key users

If you were intentionally using the same key in multiple places:

1. Use unique keys if you need independent behavior
2. Or, ensure consistent options across all calls with the same key (see the list of options that must be consistent above)
3. Consider extracting your data fetching into a composable to ensure consistency:

```ts
// composables/useUserData.ts
export function useUserData(userId) {
  return useAsyncData(
    `user-${userId}`,
    () => fetchUser(userId),
    { 
      deep: true,
      transform: (user) => ({ ...user, lastAccessed: new Date() })
    }
  )
}

// Now use this composable everywhere instead of direct useAsyncData calls
```

### 🧪 Testing Recommendations

When updating to this version:

1. Test components that use the same key in multiple places
2. Verify that reactive UI updates correctly when using shared keys
3. Test manual refresh behavior with any custom cache implementations
4. Check that reactive keys work as expected when their dependencies change
5. Run your app in development mode to catch any warnings about inconsistent options

### 🚧 TODO 

- [ ] consider caching utils to make it easier for people to migrate from `getCachedData`


docs/1.getting-started/6.data-fetching.md
```diff
@@ -350,12 +350,68 @@ Both `pick` and `transform` don't prevent the unwanted data from being fetched i
...
-- [`useAsyncData`](/docs/api/composables/use-async-data) uses its first argument as a key if it is a string. If the first argument is the handler function that performs the query, then a key that is unique to the file name and line number of the instance of `useAsyncData` will be generated for you.
+- [`useAsyncData`](/docs/api/composables/use-async-data) uses its first argument as a key if it is a string. If the first argument is the handler function that performs the query, then a key that is unique to the file name and line number of the instance of `useAsyncData` will be generated for you.
 
 ::tip
 To get the cached data by key, you can use [`useNuxtData`](/docs/api/composables/use-nuxt-data)
 ::
 
+#### Shared State and Option Consistency
+
+When multiple components use the same key with `useAsyncData` or `useFetch`, they will share the same data, error, and status refs. This ensures consistency across components but requires some options to be consistent.
```
DamianGlowala, line 361:
```suggestion
When multiple components use the same key with `useAsyncData` or `useFetch`, they will share the same `data`, `error` and `status` refs. This ensures consistency across components but requires some options to be consistent.
```

docs/3.api/2.composables/use-async-data.md
```diff
@@ -98,6 +115,35 @@ You can use `useLazyAsyncData` to have the same behavior as `lazy: true` with `u
 Learn how to use `transform` and `getCachedData` to avoid superfluous calls to an API and cache data for visitors on the client.
 ::
 
+### Shared State and Option Consistency
+
+When using the same key for multiple `useAsyncData` calls, they will share the same data, error, and status refs. This ensures consistency across components but requires option consistency.
```
DamianGlowala, line 120:
```suggestion
When using the same key for multiple `useAsyncData` calls, they will share the same `data`, `error` and `status` refs. This ensures consistency across components but requires option consistency.
```

docs/3.api/2.composables/use-fetch.md
```diff
@@ -66,6 +66,22 @@ const { data, status, error, refresh, clear } = await useFetch('/api/auth/login'
...
+<script setup lang="ts">
+const route = useRoute()
+const id = computed(() => route.params.id)
+
+// When the route changes and id updates, the data will be automatically refetched
+const { data: post } = await useFetch(() => `/api/posts/${id.value}`)
+</script>
+```
+
+When using `useFetch` with the same URL and options in multiple components, they will share the same data, error, and status refs. This ensures consistency across components.
```
DamianGlowala, line 83:
```suggestion
When using `useFetch` with the same URL and options in multiple components, they will share the same `data`, `error` and `status` refs. This ensures consistency across components.
```

DamianGlowala (commented):
For consistency with https://github.com/nuxt/nuxt/pull/31373/files#diff-878d2d71c55fa28451daeee665ad5c0c8260ffb517b827cf79106350d268e440R219

danielroe:
to investigate: https://stackblitz.com/edit/github-cpgxhnuj?file=package.json

ljcremer:
Will this fix the issue where useFetch does not work with a reactive key as in the issue
https://github.com/nuxt/nuxt/issues/21532

adamdehaven:
@danielroe I'm pulling in the `3.17.0` release into my app and am noticing separate tests utilizing `renderSuspended` are now failing as the data appears to persist between tests (e.g. utilizing a computed `fetchKey` that is based on the `route.path` in the component) as long as the `route` in the test is the same.

- Is this expected? 
- Is there a regular pattern for clearing Nuxt data between tests I should be using?

## Example

```ts
const fetchKey = computed((): string => `page-${route.path.replace(/\//g, '-')}`)
const { data } = await useFetch('/api/v3/pages/my-page', {
  key: fetchKey,
})
```

adamdehaven:
> Or call `clearNuxtData` after/before each test. Or maybe we can look at doing that automatically in `@nuxt/test-utils`. What do you think?

I actually tried calling `clearNuxtData` in a before/after hook and it had zero impact 😬 -- Yes, I think doing this automatically in `@nuxt/test-utils` would be ideal to totally isolate tests

metkm:
@danielroe Shouldn't this feature also apply to `useNuxtData`? It looks like we can't give reactive key to useNuxtData yet?

samirmhsnv:
@danielroe I believe there is a breaking change regarding FormData handling in fetch.js due to the newly added statement:

It originates from ohash. That library cannot hash File objects.

````js
if (opts.body) {
  segments.push(hash(toValue(opts.body)))
}
````

Reference:
https://github.com/nuxt/nuxt/pull/31373/files#diff-3ace8ffd6a6f0f4eb7979e99189c489e15fa917b74b48c26ee7e12ad13dfb866R270-R272

When I attempt to upload a FormData containing a File in the body, I encounter the following error:
"Cannot serialize File"

````js
const formData = new FormData();

formData.append("image", new_avatar.value);

await useAPI('/company', {
  method: 'POST',
  body: formData,
})
````

My current solution is to downgrade v3.16.2

**Comment by Tofandel** [comment]
There seems to be a breaking change in there as well, on route change without remounting components (via `<NuxtPage :page-key="(route: RouteLocationResolved) =>(route.matched[0].name || route.path)"></NuxtPage>`) then all the previously fetched data is cleared. This was not the case before

I'm trying to dig what's causing this

**Comment by GerryWilko** [comment]
@danielroe Thank you so much for your work!

I'm a little confused by some of the changes to `useAsyncData` here so I thought I would add my thoughts here and hopefully you can clarify when you get the time.

When initially reading about this upgrade to `useAsyncData` I initially thought this would resolve the typical issues we would get where using `useAsyncData` in multiple places with the same key could cause multiple triggers of the handler to be fired and therefore fire unnecessary requests.

This would typically lead me to the discussions around the fact that `useAsyncData` does not cache data and instead caching needs to be handled independently however we previously couldn't stop the handler from firing on initial call of `useAsyncData` even if the key provided matched an existing `useAsyncData` call.

This is where I presumed these changes came in attempting to unify the instance returned by `useAsyncData` and therefore allowing for the `getCachedData` function to be respected on subsequent `useAsyncData` calls.

However from my testing I'm struggling to understand how this works I have a simple unit test here that illustrates the problem: 

```ts
import type { NuxtApp } from '#app'
import type { AsyncDataRefreshCause } from '#app/composables/asyncData'
import { expect, it, vi } from 'vitest'

const handler = vi.fn(async () => Promise.resolve('hello'))

const getCachedData = vi.fn((key: string, nuxtApp: NuxtApp, ctx: { cause: AsyncDataRefreshCause }) => {
  if (nuxtApp.isHydrating) {
    return nuxtApp.payload.data[key]
  }

  const { data } = useNuxtData(key)
  if (ctx.cause !== 'refresh:manual' && ctx.cause !== 'refresh:hook' && data.value) {
    return data.value
  }
})

// eslint-disable-next-line ts/promise-function-async
function testAsyncData() {
  return useAsyncData('test-key', handler, {
    getCachedData,
  })
}

it('test duplicate calls are not made after first call has finished', async () => {
  const { status, data } = await testAsyncData()
  expect(status.value).toBe('success') // pass
  expect(data.value).toBe('hello') // pass
  expect(handler).toHaveBeenCalledTimes(1) // pass
  const { status: status2, data: data2 } = testAsyncData()
  expect.soft(handler).toHaveBeenCalledTimes(1) // fail - called twice
  expect.soft(getCachedData).toHaveBeenCalledTimes(2)
  expect.soft(data.value).toBe('hello') // pass
  expect.soft(data2.value).toBe('hello') // pass
  expect.soft(status.value).toBe('success') // fail - value is 'pending'
  expect.soft(status2.value).toBe('success') // fail - value is 'pending'
})
```

Effectively what I am trying to do here is to adjust the default `getCachedData` but instead always respect the data we have unless you explicitly try to refresh it.

However what actually happens is that whilst the `getCachedData` is called twice with cause `initial` the second call is effectively discarded as it is passed only to new `useAsyncData` instances as the `initialCachedData`.

https://github.com/nuxt/nuxt/blob/ebff42f83dfcf9d04c6e399088cde4b8ec14ff81/packages/nuxt/src/app/composables/asyncData.ts#L636-L645

It is then used here but it only ever uses the `initialCachedData` value even though the most recent evaluation of `getCachedData` returned a value. I presume there are some security considerations here that is the concern? Could this perhaps be adjusted to be that on client we always respect the current cached value (I think this should always be safe on client)?

At present it seems to be that we cannot stop the `handler` from being triggered on each call to `useAsyncData` (I believe some deduping occurs whilst the handler is in flight).

PR #31376, opened 2025-03-14
ci: monitor all jobs permissions

huang-julien:
### 📚 Description
monitor actions
Should we monitor every single jobs in our CIs ?

### DIFF
--- a/.github/workflows/autofix-docs.yml
+++ b/.github/workflows/autofix-docs.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@babd69bc8d78e6cdece903dfdcfb72d4e1a4f00d
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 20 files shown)


danielroe:
> The typical scenario is to include the Monitor action in every job of the workflow that doesn't specify permissions explicitly, collect the recommendations from several workflow runs, apply the recommended minimal permissions, and then remove the Monitor action.

so sure, go ahead and add it to all the jobs currently in the repo (in this pr) and we can see if the explicit permissions need to be updated at all

then we can remove it again - it might be we don’t need to make any changes

danielroe:
looks like we're good at the moment, right?

most jobs are running without any permissions.

huang-julien:
wait .... something's wrong here...
````yaml
  codeql:
    runs-on: ubuntu-latest
    timeout-minutes: 10
    strategy:
      matrix:
        language: ['javascript-typescript', 'actions']
    permissions:
      actions: read
      contents: read
      security-events: write
```

![image](https://github.com/user-attachments/assets/ca7d9543-5be2-4682-9efc-f37d3c74b8af)

I think CodeQL do need to read actions and contents. Is that a bug within  actions-permissions ?

danielroe:
yes, that was my take too

i think all our actions have the correct minimal permissions already

danielroe:
closing as I think there's nothing to do here but do reopen it if I'm wrong!

PR #31379, opened 2025-03-15
perf(nuxt): use browser cache for payloads

### LINKED ISSUE #15438: Cleanup payload in memory cache when not needed

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15438 (together with https://github.com/nuxt/nuxt/pull/31373)

### 📚 Description

this addresses a number of issues:

1. previously `preloadPayload` was entirely broken with JSON payloads (it was not internally used and seems not to have been used externally either as there was no bug report) and did not set the crossorigin attribute properly
2. previously the payload cache was never cleared and just increased in memory
3. previously on prefetch the payload was already loaded into memory, not just fetched

Instead, we now ~~hint to browser to prefetch a URL~~ we still import the URL but do not set it in memory (we will prefetch URL when we can also prefetch server components too). when we actually load it we force using browser cache (the query param busts this cache so it's safe to do). this allows the browser to safely clean up the cache when required; nuxt will refetch payloads in this case if needed.

this should _dramatically_ improve memory usage on large or content-driven websites.

### DIFF
--- a/packages/nuxt/src/app/composables/payload.ts
+++ b/packages/nuxt/src/app/composables/payload.ts
@@ -27,15 +27,35 @@ export async function loadPayload (url: string, opts: LoadPayloadOptions = {}): 
   const payloadURL = await _getPayloadURL(url, opts)
   return await _importPayload(payloadURL) || null
 }
+let linkRelType: string | undefined
+function detectLinkRelType () {
+  if (import.meta.server) { return 'preload' }
+  if (linkRelType) { return linkRelType }
+  const relList = document.createElement('link').relList
+  linkRelType = relList && relList.supports && relList.supports('prefetch') ? 'prefetch' : 'preload'
+  return linkRelType
+}
 /** @since 3.0.0 */
 export function preloadPayload (url: string, opts: LoadPayloadOptions = {}): Promise<void> {
   const nuxtApp = useNuxtApp()
   const promise = _getPayloadURL(url, opts).then((payloadURL) => {
-    nuxtApp.runWithContext(() => useHead({
-      link: renderJsonPayloads
-        ? [{ rel: 'preload', as: 'fetch', crossorigin: 'anonymous', href: payloadURL }]
-        : [{ rel: 'modulepreload', crossorigin: '', href: payloadURL }],
-    }))
+    const link = renderJsonPayloads
+      ? { rel: detectLinkRelType(), as: 'fetch', crossorigin: 'anonymous', href: payloadURL } as const
+      : { rel: 'modulepreload', crossorigin: '', href: payloadURL } as const
+
+    if (import.meta.server) {
+      nuxtApp.runWithContext(() => useHead({ link: [link] }))
+    } else {
+      const linkEl = document.createElement('link')
+      for (const key of Object.keys(link) as Array<keyof typeof link>) {
+        linkEl[key === 'crossorigin' ? 'crossOrigin' : key] = link[key]!
+      }
+      document.head.appendChild(linkEl)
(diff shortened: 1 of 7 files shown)


packages/nuxt/src/app/composables/payload.ts
```diff
@@ -19,34 +19,22 @@ interface LoadPayloadOptions {
 /** @since 3.0.0 */
 export async function loadPayload (url: string, opts: LoadPayloadOptions = {}): Promise<Record<string, any> | null> {
   if (import.meta.server || !payloadExtraction) { return null }
-  const payloadURL = await _getPayloadURL(url, opts)
-  const nuxtApp = useNuxtApp()
-  const cache = nuxtApp._payloadCache ||= {}
-  if (payloadURL in cache) {
-    return cache[payloadURL] || null
+  // TODO: allow payload extraction for non-prerendered URLs
```
TheAlexLichter, line 22: Will this be part of a separate PR?

danielroe: yes.

it's an existing todo that i forgot to write in previously (ie it's not changed by this pr)

packages/nuxt/src/app/composables/payload.ts
```diff
@@ -19,35 +19,43 @@ interface LoadPayloadOptions {
...
-      delete cache[payloadURL]
-      return null
-    })
-  })
-  return cache[payloadURL]
+  const payloadURL = await _getPayloadURL(url, opts)
+  return await _importPayload(payloadURL) || null
+}
+let linkRelType: string | undefined
+function detectLinkRelType () {
```
huang-julien, line 31: can this be run server side too ?

danielroe: yes. preloadPayload has server handling

huang-julien: oh my bad, didn't see the `import.meta.server` flag

TheAlexLichter (approved):

SoaresMG:
When doing some performance analysis this previous cache variable used to be a hotspot for stuff hanging around forever.

However, I'm not understanding how this new behaviour will work with `sharedPrerenderData` which if I'm not mistaken is the experimental feature that shares payloads between requests.

We are now using `force-cache` but if I'm using SSG, how will this work out? Since node doesn't have that concept (or does it? 🤔)

If I have 100 pages using the same URL in a `useAsyncData` during SSG, will it share the same payload and prevent a request?

danielroe:
this does not affect the sharedPrerenderData feature

SoaresMG:
Can confirm, works wonders!

On 24 websites using SSG with some reaching 3k pages (and most are pretty payload heavy) we had a huge memory decrease with 3.16.1 and the flag active.

On average it was around 2gb less, heaviest ones going from 9-10 to 7-8 while smaller ones also registering a 1-2gb decrease.

The only downside I noticed is that while memory decreased, generation time for all websites also increase by a max of 2-3 minutes, but it's weird because it's not correlated with website size (amount of routes), smaller websites had the same increase as some bigger ones, I need to investigate this further.

Overall great improvement!

mertgundoganx:
if i dont want to use this with useAsyncData what should i do? any global settings?

danielroe:
@mertgundoganx it is not configurable. 

if you don't want this, perhaps you could open a new issue and explain why not? do you understand what is happening and what happened before this pr?

PR #31386, opened 2025-03-16
refactor(nuxt): move island renderer into its own event handler

huang-julien:
### 📚 Description

This PR moves nuxt island into its own endpoint for better readability of the render handler. 
It also extract all shared logic within `utils/renderer`.


packages/nuxt/src/core/nuxt.ts
```diff
@@ -663,6 +663,11 @@ async function initNuxt (nuxt: Nuxt) {
       filePath: resolve(nuxt.options.appDir, 'components/nuxt-island'),
     })
 
+    addServerHandler({
+      route: '/__nuxt_island/**',
+      handler: resolve(distDir, 'core/runtime/nitro/handlers/__nuxt_island'),
+    })
```
danielroe, line 669: we might be able to add this conditionally - e.g. not add it if the user has disabled `experimental.componentIslands` entirely.

test/bundle.test.ts
```diff
@@ -58,10 +58,10 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
     const serverDir = join(rootDir, '.output/server')
 
     const serverStats = await analyzeSizes(['**/*.mjs', '!node_modules'], serverDir)
-    expect.soft(roundToKilobytes(serverStats.totalBytes)).toMatchInlineSnapshot(`"204k"`)
+    expect.soft(roundToKilobytes(serverStats.totalBytes)).toMatchInlineSnapshot(`"210k"`)
```
danielroe, line 61: we should investigate why the server bundle size is changing

huang-julien: that's probably because there's still a bit of slightly duplicated content between the renderer and islands

huang-julien:
wat
image

danielroe (approved):
very nice

packages/nuxt/src/core/nuxt.ts
```diff
@@ -625,6 +625,14 @@ async function initNuxt (nuxt: Nuxt) {
       filePath: resolve(nuxt.options.appDir, 'components/nuxt-island'),
     })
 
+    // sync conditions with /packages/nuxt/src/core/templates.ts#L539
+    if (nuxt.options.dev || nuxt.options.experimental.componentIslands !== 'auto' || nuxt.apps.default?.pages?.some(p => p.mode === 'server') || nuxt.apps.default?.components?.some(c => c.mode === 'server' && !nuxt.apps.default?.components.some(other => other.pascalName === c.pascalName && other.mode === 'client'))) {
```
danielroe, line 629: actually, we may need to move the addition of this event handler later as `nuxt.apps.default` isn't yet initialised.

PR #31448, opened 2025-03-19
docs: improve `refreshNuxtData` docs

Mini-ghost:
### 🔗 Linked issue

Resolve #26803

### 📚 Description

Update examples and clarification for `refreshNuxtData`.


danielroe (commented):
when we support a single asyncData instance then we probably will not want this behaviour, as when components are unmounted we automatically stop watching 

and I think this would mean, even in the current implementation, that deactivate - refresh - reactivate never refreshes

Mini-ghost:
> when we support a single asyncData instance then we probably will not want this behaviour, as when components are unmounted we automatically stop watching
> 
> and I think this would mean, even in the current implementation, that deactivate - refresh - reactivate never refreshes

This PR addresses an issue where, when a component is cached by `<KeepAlive>`, calling `refreshNuxtData()` still refreshes the `asyncData` inside those cached components.  

To prevent this, I’ve attempted to check whether the component is cached before refresh. This is necessary because cached components are not unmounted.  

I’ve also tested this issue with PR #31373 (a single `asyncData` instance) in the following minimal reproduction:  
https://stackblitz.com/edit/github-q2qypp-13axgcou

If there’s anything I may have overlooked, please let me know. I truly appreciate your feedback!

danielroe:
but it probably should refresh at _least_ before/on reactivation, right? perhaps we could 'store' the need to refresh and run it on reactivation?

what do you think @TheAlexLichter? would you consider the linked issue an actual bug?

danielroe:
Yes, I think that's a good idea!

docs/3.api/3.utils/refresh-nuxt-data.md
```diff
@@ -1,39 +1,42 @@
...
     to: https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts
     size: xs
 ---
 
-::note
-`refreshNuxtData` re-fetches all data from the server and updates the page as well as invalidates the cache of [`useAsyncData`](/docs/api/composables/use-async-data) , `useLazyAsyncData`, [`useFetch`](/docs/api/composables/use-fetch) and `useLazyFetch`.
-::
+`refreshNuxtData` is used to refetch all or specific `asyncData` instances, including those from [`useAsyncData`](/docs/api/composables/use-async-data), [`useLazyAsyncData`](/docs/api/composables/use-lazy-async-data), [`useFetch`](/docs/api/composables/use-fetch), and [`useLazyFetch`](/docs/api/composables/use-lazy-fetch).  
+
+If your component is cached by `<KeepAlive>` and enters a deactivated state, the `asyncData` inside the component will still be refetched until the component is unmounted.
```
danielroe, line 13:
```suggestion
::note
If your component is cached by `<KeepAlive>` and enters a deactivated state, the `asyncData` inside the component will still be refetched until the component is unmounted.
::
```

docs/3.api/3.utils/refresh-nuxt-data.md
```diff
@@ -1,39 +1,42 @@
...
 ## Type
 
 ```ts
 refreshNuxtData(keys?: string | string[])
 ```
 
-**Parameters:**
+## Parameters
+
+* `keys`: A single string or an array of strings as `keys` that are used to fetch the data. This parameter is **optional**. All [`useAsyncData`](/docs/api/composables/use-async-data) and [`useFetch`](/docs/api/composables/use-fetch) are re-fetched when no `keys` are specified.
```
danielroe, line 23:
```suggestion
* `keys`: A single string or an array of strings as `keys` that are used to fetch the data. This parameter is **optional**. All [`useAsyncData`](/docs/api/composables/use-async-data) and [`useFetch`](/docs/api/composables/use-fetch) keys are re-fetched when no `keys` are explicitly specified.
```

danielroe (approved):
❤️

PR #31450, opened 2025-03-19
fix(nuxt): use more performant router catchall pattern

danielroe:
### 📚 Description

this improves catchall pattern to avoid polynomial/exponential backtracking when used in the middle of a route pattern (e.g. `~/pages/[...id]/other`) which _in any case_ is an anti-pattern I would strongly advise avoiding.


vsergiu93:
Hi @danielroe , 

**TL;DR**: After upgrading from Nuxt 3.15.4 to 3.17.5, nested catch-all routes are no longer being grouped as children but are instead being flattened into separate top-level routes, breaking our existing URL structure.

I'm not directly opening a new issue, because I first want to understand if this is really an issue that needs to be fixed on the Nuxt side.


We have the following structure:

pages/
├── [...layout].vue                
├── [...layout]/
│   ├── index.vue                      
│   └── page-[page].vue   


This would have generated something like this in `3.15.4` : 

```javascript
[
  {
    "children": [
      {
        "children": [],
        "file": "/[...layout]/index.vue",
        "name": "layout",
        "path": ""
      },
      {
        "children": [],
        "file": "/[...layout]/page-[page].vue",
        "name": "layout-page-page",
        "path": "page-:page()"
      }
    ],
    "file": "pages/[...layout].vue",
    "path": "/:layout(.*)*"
  }
]
```

Now in `3.17.5` it generates the following: 

```javascript
[
  {
    "children": [
      {
        "children": [],
        "file": "/[...layout]/index.vue",
        "name": "layout",
        "path": ""
      }
    ],
    "file": "/[...layout].vue",
    "path": "/:layout(.*)*"
  },
  {
    "children": [],
    "file": "/[...layout]/page-[page].vue",
    "meta": undefined,
    "name": "layout-page-page",
    "path": "/:layout([^/]*)*/page-:page()"
  }
]
```
As you can see, the `page-[page].vue` is not added as a child of `[...layout].vue` anymore. What I'm not really sure about is if this change was intentional or if only the regex pattern was supposed to be modified.

**The issue**: This breaks our existing route structure where `/some-category/page-20` should match the catch-all parent with the page child, but now it tries to match the flattened route pattern instead.

Basically, as it stands now I can't upgrade. Changing the URL structure isn't possible since our URLs are dynamically controlled by a CMS where admins can configure arbitrary layout/page combinations. We rely on the flexible nested catch-all routing structure to handle these dynamic CMS-generated routes.


Please let me know if I should open an issue or if this was intended, thanks.

danielroe:
you are right, this is a bug

i believe there is an existing issue but it did not identify the cause, so it might be difficult to find

vsergiu93:
@danielroe  I think the issue is [here](https://github.com/nuxt/nuxt/blob/9e1d0dbc1c77cf46cdb75189c0db821b0ddba30a/packages/nuxt/src/pages/utils.ts#L142), I cloned the repo and I'm playing around, basically I added this lines to pages.test.ts: 

```js
[
  { path: `${pagesDir}/[...sub].vue` },
  { path: `${pagesDir}/[...sub]/index.vue` },
  { path: `${pagesDir}/[...sub]/route-[slug].vue` }
]
```
Then I run the test in debug mode, I allow the first two entries to be processed which seems to correctly group them as parent child routes

image

After a few steps we get here:

image

**The Problem:**
When processing `[...sub]/route-[slug].vue`, the algorithm tries to find a parent route to attach it to. However, it's looking for an exact path match between:
- Expected parent path: `/:sub([^/]*)*` 
- Actual parent path: `/:sub(.*)*`

Since these don't match exactly, the algorithm fails to recognize the parent-child relationship and creates the route as a separate top-level route instead of nesting it under the catch-all parent.

I have no idea how to actually fix it though.

Thanks

vsergiu93:
Hello @danielroe ,

Let me know if you want me to create a new issue (I searched through the currently open issues, but I was unable to find one where I can reference my comments). 

Thanks

danielroe:
@vsergiu93  please do 🙏

PR #31460, opened 2025-03-19
docs: remove invalid url

iiio2:
Actually here the url is invalid. I think instead of giving invalid url, we can give url like that.


danielroe (commented):
not sure I understand. it's an example of accessing an external api

PR #31467, opened 2025-03-20
chore(docs): add dedupe flag

TheAlexLichter:
### 📚 Description

The docs were missing the new flag

### DIFF
--- a/docs/3.api/4.commands/upgrade.md
+++ b/docs/3.api/4.commands/upgrade.md
@@ -10,7 +10,7 @@ links:
 
 <!--upgrade-cmd-->
 ```bash [Terminal]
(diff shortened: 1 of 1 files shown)


danielroe (approved):

docs/3.api/4.commands/upgrade.md
```diff
@@ -31,6 +31,7 @@ Option | Default | Description
 --- | --- | ---
 `--cwd=<directory>` |  | Specify the working directory, this takes precedence over ROOTDIR (default: `.`)
 `--logLevel=<silent\|info\|verbose>` |  | Specify build-time log level
+`--dedupe` |  | Will deduplicate dependencies but not recreate the lockfile
```
OrbisK, line 34: Not sure if this is relevant, but it depends on the package manager. It will dedupe if possible, and fall back to lockfile recreation if not supported (e.g. for bun). https://github.com/unjs/nypm/blob/dfb09efd82735733e2a766a58a98ee4f2a062705/src/api.ts#L247

PR #31484, opened 2025-03-20
perf(nuxt): remove `oxc-parser` manual wasm fallback logic

BobbieGoede:
### 📚 Description
Maybe I misunderstand, but based on https://stackblitz.com/edit/oxc-parser it looks like we can now rely on wasm being used without handling it ourselves, as long as `@oxc-parser/binding-wasm32-wasi` is installed as dependency? 👀


TheAlexLichter:
Yes! AFAIK StackBlitz set up package replacements that should remove the need for a manual workaround.

BobbieGoede:
I'm trying to remove the fallbback in nuxt-i18n as well, maybe `@oxc-parser/binding-wasm32-wasi` needs to be added as an optional dependency? (related https://github.com/nuxt-modules/i18n/pull/3428#issuecomment-2742829297)

BobbieGoede:
Moving `@oxc-parser/binding-wasm32-wasi` to optional deps is not enough to make it work in browser environment 🤔

BobbieGoede:
FYI - setting to draft since these changes will break browser environments, tests do not cover this, do not merge.

BobbieGoede:
Works as expected for the i18n module in StackBlitz (https://github.com/nuxt-modules/i18n/pull/3413#issuecomment-2825662048).

I haven't tried this in other web environments such as CodeSandbox (not familiar with it myself), it would be worth checking if it works there too before merging this.

danielroe (approved):

PR #31507, opened 2025-03-23
fix(nuxt): ensure layer array-type config is merged in order

### LINKED ISSUE #25719: Incorrect order of modules: modules from project root are loaded before the ones in layer
### Environment

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.10.1
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Builder:          -
- User Config:      extends, modules
- Runtime Modules:  ()
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-vg4ayj

You can see in the console that the module from the root folder is loaded first.

### Describe the bug

I'd expect modules from any layers to be loaded before the ones in project root.

### Additional context

_No response_

### Logs

_No response_

aadamcik:
### 🔗 Linked issue

- #25719

### 📚 Description

This PR fixes the the incorrect order of array values in config when using layers.

- Resolves #25719


packages/nuxt/src/core/nuxt.ts
```diff
@@ -655,6 +657,9 @@ async function initNuxt (nuxt: Nuxt) {
 
   await nuxt.callHook('modules:done')
 
+  // remove duplicate css after modules are done
+  nuxt.options.css = unique(nuxt.options.css)
```
danielroe, line 661: is there a reason you removed L819-832? these were necessary to ensure not just that the css was de-duplicated, but that it was added in the correct order.

packages/kit/src/loader/config.ts
```diff
@@ -17,22 +17,38 @@ export interface LoadNuxtConfigOptions extends Omit<LoadConfigOptions<NuxtConfig
...
+  if (Array.isArray(obj[key])) {
+    if (Array.isArray(value)) {
+      obj[key] = obj[key].concat(value)
+      return true
+    }
+
+    if (typeof value === 'function') {
+      obj[key] = value(obj[key])
+      return true
+    }
```
danielroe, line 30: I am not sure this is always safe as some options actually take functions, for example in `vite.build`.

packages/nuxt/src/core/nuxt.ts
danielroe: it was correct before (but only for CSS) - the issue that was being addressed by that code is that CSS added explicitly in a project needs to be later in the array so it has prioirty.

it wouldn't be true for modules, though

packages/kit/src/loader/config.ts
danielroe: let's split this functionality out into a new one with some unit tests to ensure we cover all cases. I definitely think it's worth allowing this kind of thing but we need to think about how to do it.

we also need to update the types so nuxt config accepts these functions directly passed into the configuration.

The only problematic example that immediately occurs to me is something like rollup's external option which can be either an array or a function. (And we do try to discourage non-serialisable options being passed in to nuxt configuration.) But I can't rule out that there might be other such situations, particularly in configuration that we don't 'control', like nitro, vite, esbuild, rollup, etc., which is why hitherto we've restricted custom array merging to `app.config`.

packages/nuxt/src/core/nuxt.ts
adamcikado: Updated

packages/kit/src/loader/config.ts
adamcikado: Updated

packages/nuxt/test/load-nuxt.test.ts
```diff
@@ -107,6 +107,23 @@ describe('loadNuxt', () => {
...
+  it('ensures layer modules remains in order', async () => {
+    const layerFixtureDir = withoutTrailingSlash(normalize(fileURLToPath(new URL('./layers-fixture', import.meta.url))))
+    const nuxt = await loadNuxt({ cwd: layerFixtureDir })
+    await nuxt.close()
+
+    const modules = nuxt.options.modules.map(item => typeof item === 'function' ? item.name : item)
+
+    expect(modules).toMatchInlineSnapshot(`
+      [
+        "layer",
```
danielroe, line 120: I think auto-registered layers should be last because they are in the user's project.

so while we are reordering, we should probably ensure they are 'after' layers in `extends`, etc.

adamcikado: I updated the logic, so that auto-registered layers are loaded after the layers in `extends`.

This also affects 2 things:

1. Order of CSS, see updated test.
Before:
```
"auto.css", // auto-registered layer
"custom.css", // layer from extends
```
Now:
```
"custom.css", // layer from extends
"auto.css", // auto-registered layer
```

2. Order of imports, see updated test.

danielroe: thank you!

danielroe (approved):

danielroe:
Thanks for this nice PR ❤️ 

We rely on the path to generate types but I think there's an opportunity to split out the types from path resolution, for a slight perf enhancement, in another PR, if you're interested.

danielroe:
as it's possible people are relying on the incorrect order (flagged up in an ecosystem-ci run on nuxt-modules/i18n), we're not including this in v3.16.2 release until we have a better signal on how likely it is this would affect people.

PR #31511, opened 2025-03-23
docs: mention autoimport resolveComponent only works with literal strings

### LINKED ISSUE #31503: `resolveComponent()` doesn't accept template string
### Environment

- Node Version:     `v22.14.0`
- Nuxt Version:     `3.16.1`
- CLI Version:      `3.23.1`
- Nitro Version:    `2.11.7`
- Package Manager:  `pnpm@9.15.4`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `experimental`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/gxres042/nuxt-reproduction

### Describe the bug

`resolveComponent()` can only resolve normal strings except template strings.

In my opinion, template strings is a type of *string* and it can be resolved by `resolveComponent()`.

### Additional context

_No response_

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

fix #31503

### 📚 Description

Docs are maybe not clear enough that autoimports only accept literal strings

### DIFF
--- a/docs/2.guide/2.directory-structure/1.components.md
+++ b/docs/2.guide/2.directory-structure/1.components.md
@@ -79,8 +79,7 @@ const MyButton = resolveComponent('MyButton')
 ```
 
 ::important
-If you are using `resolveComponent` to handle dynamic components, make sure not to insert anything but the name of the component, which must be a literal string (template strings are not valid) and not a variable.
-The string is statically analyzed at compilation step.
+If you are using `resolveComponent` to handle dynamic components, make sure not to insert anything but the name of the component, which must be a literal string and not be or contain a variable. The string is statically analyzed at compilation step.
 ::
 
 ::tip{icon="i-lucide-video" to="https://www.youtube.com/watch?v=4kq8E5IUM2U" target="\_blank"}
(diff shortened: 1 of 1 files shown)


danielroe:
i think we could probably support string templates without interpolation

maybe update this to indicate that the issue is that the string must be statically analysed?

and i can create a pr to support templates

huang-julien:
I don't really see the point of template string without any interpolation 🤔. Wouldn't handling it make the regex more complex (to read) ?

docs/2.guide/2.directory-structure/1.components.md
```diff
@@ -79,7 +79,8 @@ const MyButton = resolveComponent('MyButton')
 ```
 
 ::important
-If you are using `resolveComponent` to handle dynamic components, make sure not to insert anything but the name of the component, which must be a string and not a variable.
+If you are using `resolveComponent` to handle dynamic components, make sure not to insert anything but the name of the component, which must be a literal string (template strings are not valid) and not a variable.
+The string is statically analyzed at compilation step.
```
danielroe, line 83:
```suggestion
If you are using `resolveComponent` to handle dynamic components, make sure not to insert anything but the name of the component, which must be a literal string and not be or contain a variable. The string is statically analyzed at compilation step.
```

danielroe (approved):

PR #31528, opened 2025-03-25
fix(nuxt): use greedy catchall when `/index` is the last segment

### LINKED ISSUE #3464: Incorrect Route Resolution with Dynamic Routes
### Environment

- Node Version:     `v20.11.0`
- Nuxt Version:     `3.16.1`
- CLI Version:      `3.23.1`
- Nitro Version:    `2.11.7`
- Package Manager:  `npm@10.8.1`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `modules`, `i18n`
- Runtime Modules:  `@nuxtjs/i18n@9.4.0`
- Build Modules:    `-`

### Steps to Reproduce
1. Create a basic Nuxt project with the above pages structure
2. Verify that routing works correctly without i18n:
   - `/` routes to index page
   - `/category-name` routes to category page
3. Install the latest version of `@nuxt/i18n`
4. Configure basic i18n setup with English locale and `strategy: 'prefix'`
5. Try accessing the root path with locale (`/en`)
6. Observe that instead of showing the index page, it triggers the `[...category]` route

### Minimal Reproduction
https://stackblitz.com/edit/bobbiegoede-nuxt-i18n-starter-uebz17bh?file=pages%2Findex.vue

### Description
After installing the latest version of `@nuxt/i18n` module, there's an issue with route resolution where attempting to access the root path with locale (e.g., `/en`) incorrectly redirects to the dynamic category page (`[...category]`) instead of the index page.

### Current Behavior
When accessing the root path with locale (e.g., `/en`):
- The application incorrectly routes to the dynamic category page
- The `[...category]` catch-all route is triggered instead of the index route

### Expected Behavior
- Accessing `/en` should route to the index page
- The dynamic category routes should only be triggered when explicitly accessing category paths (e.g., `/en/category-name`)

### Additional Context
- The issue only occurs after installing the i18n module
- The routing behavior works correctly without the i18n module
- This appears to be a regression in the latest version of the module

### Logs

```shell

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt-modules/i18n/issues/3464

### 📚 Description

we don't want to apply the more narrow regexp when there is no following segment, but we neglected to check for `/index` as a segment (which we ignore) meaning the path matcher we created had too low of a priority

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -139,7 +139,7 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
       route.name += (route.name && '/') + segmentName
 
       // ex: parent.vue + parent/child.vue
-      const routePath = getRoutePath(tokens, segments[i + 1] !== undefined)
+      const routePath = getRoutePath(tokens, segments[i + 1] !== undefined && segments[i + 1] !== 'index')
       const path = withLeadingSlash(joinURL(route.path, routePath.replace(INDEX_PAGE_RE, '/')))
       const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path)
 
--- a/packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
+++ b/packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
@@ -594,6 +594,15 @@
   "should use more performant regexp when catchall is used in middle of path": [
     {
       "alias": "mockMeta?.alias || []",
+      "component": "() => import("pages/[...id]/index.vue")",
+      "meta": "mockMeta || {}",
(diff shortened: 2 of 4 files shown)


BobbieGoede (approved):

john-psina:
@danielroe Thank you for the fix! I have a couple of questions regarding the availability of these changes:

1. Will these changes be released only with Nuxt 4, or can we expect them in the next Nuxt 3 release?

2. Is there any way to use this fix before the official release? I'm wondering if it's possible to install a pre-release version with this fix from the main branch for local development until the changes are officially released.

This would be really helpful for our current development process, as we're actively working with dynamic routes and i18n.

danielroe:
you can expect a patch release this weekend, and before then (once this is cherry picked to 3.x) you can use the nightly channel: https://nuxt.com/docs/guide/going-further/nightly-release-channel

PR #31545, opened 2025-03-26
fix(nuxt): improve default scroll behaviour

### LINKED ISSUE #24941: Various issues with scroll position on history back / forward, hash-link change, initial load with hash links, etc.
### Environment

[11:42:49] Working directory: /home/projects/qgnimawow.github
[11:42:49] Nuxt project info: 

- Operating System: Linux
- Node Version:     v18.18.0
- Nuxt Version:     3.9.0
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  npm@9.4.2
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-t79jgw?file=app%2Frouter.options.ts

### Describe the bug

I'm creating this as a collective ticket tracking various issues that all start in `router.options.ts`. I've spent the last day tracking down and partially fixing those issues (PR will follow shortly). Most of them are small logical issues in the code (checking for the wrong values. wrong timing, ignoring first page load, upstream bug in isChangingPage etc.). Due to time restrictions on my part I can't create separate issues and reproductions for each one of them individually:

- [x] history back / history forward discards `savedPosition` and overwrites it with `{left: 0, top: 0}`
- [x] hash-link when changing page doesn't get scrolled to
- [x] hash-link is overwritten on first page load / page refresh
- [x] `isChangingPage()` returns a false positive (https://github.com/nuxt/nuxt/issues/24934)
- [x] ~~when navigating to a hash link `scroll-padding-top` is ignored~~
- [x] when refreshing the page (F5) for some reason we get a saved position and the hash in the url is ignored. it should scroll to it however

I will provide a pull request shortly that imho solve all those navigation issues without regressions or breaking changes.

### Additional context

This is really evident on a site that has a mix of hash-links and non-hash links and relies on browser back button to get to where you came from. Page transitions make it even worse.

### Logs

_No response_

dennisadriaans:
### 🔗 Linked issue
closes #24960

resolves https://github.com/nuxt/nuxt/issues/24941

### 📚 Description

Improved the default `router.options.scrollBehavior` to handle edge cases and provide more consistent scroll restoration behavior on route navigation.


packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -43,12 +43,8 @@ export default <RouterConfig> {
     const hasTransition = (route: RouteLocationNormalized) => !!(route.meta.pageTransition ?? defaultPageTransition)
     const hookToWait = (hasTransition(from) && hasTransition(to)) ? 'page:transition:finish' : 'page:finish'
```
danielroe, line 44: I think this might be a better default as it handles nested routes

```suggestion
    const hookToWait = (hasTransition(from) && hasTransition(to)) ? 'page:transition:finish' : 'page:loading:end'
```

dennisadriaans:
@danielroe, we refined the changes from yesterday to isolate the necessary changes for fixing the scroll behavior. We aimed to keep the changes as close to the current Nuxt code as possible.

We did manual testing:

- Refresh page - remember scroll position.
- Refresh page (hard reload) - remember scroll position.
- Navigate to the same URL - reset scroll to top.
- Prevent scroll to top on navigate to previous route with keep-alive (original bug).
- We have inspected route changes with Chrome performance check on 100ms precision with a 20x CPU slowdown to throttle performance. So, we are pretty confident that it works now.

danielroe:
thank you @dennisadriaans - this looks like great work ❤️

I tested it and it looks great - deployed it tentatively on https://nuxt.com with my small change ...

danielroe (approved):

rylanharper:
This is planned for the next `3.16.2` release?

mrleblanc101:
I think scroll behavior still has quirks.
I opened a new issue about the scroll behavior of nested pages.

https://github.com/nuxt/nuxt/issues/31638

packages/nuxt/src/pages/runtime/router.options.ts
matsn0w: @dennisadriaans you missed the comment above this line ;)

dennisadriaans: @matsn0w Did I? The code checks the transition before selecting the actual hook on line 44, right?

florian-strasser:
For me this broke navigation from one page to another with hash links.

Had to setup app/router.options.ts with following code to get everything working again:
`import type { RouterConfig } from "@nuxt/schema";

export default <RouterConfig>{
  scrollBehavior(to, from, savedPosition) {
    if(savedPosition)
    {
      return savedPosition
    }
    else{
      if (to.hash && to.path == from.path) {
        return { el: to.hash }
      }
      else if(to.hash && to.path != from.path) {
        return new Promise((resolve, reject) => {
          setTimeout(() => {
            resolve({ el: to.hash })
          }, 400)
        })
      }
      else{
        if(to)
        {
          return new Promise((resolve, reject) => {
            setTimeout(() => {
              resolve({ top:0, left:0 })
            }, 400)
          })
        }
        else
        {
          return { top: 0, left: 0 }
        }
      }
    }
  }
};
`

dennisadriaans:
@florian-strasser Could you provide a minimal reproduction so I can debug your exact use case?

kalnode:
> For me this broke navigation from one page to another with hash links.
> 
> Had to setup `app/router.options.ts` with following code to get everything working again:
> 
> ```
> import type { RouterConfig } from "@nuxt/schema";
> 
> export default <RouterConfig>{
>   scrollBehavior(to, from, savedPosition) {
>     if(savedPosition)
>     {
>       return savedPosition
>     }
>     else{
>       if (to.hash && to.path == from.path) {
>         return { el: to.hash }
>       }
>       else if(to.hash && to.path != from.path) {
>         return new Promise((resolve, reject) => {
>           setTimeout(() => {
>             resolve({ el: to.hash })
>           }, 400)
>         })
>       }
>       else{
>         if(to)
>         {
>           return new Promise((resolve, reject) => {
>             setTimeout(() => {
>               resolve({ top:0, left:0 })
>             }, 400)
>           })
>         }
>         else
>         {
>           return { top: 0, left: 0 }
>         }
>       }
>     }
>   }
> };
> ```

I've been doing similar; thanks for an additional edge case.
400ms is the magic number I've been using too, however even with that there's still some apparent jankyness for a split second.

jeffreyroulston:
> @florian-strasser Could you provide a minimal reproduction so I can debug your exact use case?

Had the same issue today, this occurs when the onLeave transition has a duration or a delay before the done callback function is called.

It will immediately snap to the top of the page. Example here


```
const handlePageLeave = (done: () => void) => {
    setTimeout(() => {
        done()
    }, 2000)
}

<NuxtPage 
    :transition="{
        name: 'page',
        mode: 'out-in',
        onLeave: (el, done) => {
            // wait 1 second 
            handlePageLeave(done)
        },
    }"
/>
```

PR #31571, opened 2025-03-28
fix(nuxt): parse `error.data` before rendering `error.vue`

### LINKED ISSUE #29746: When the route does not match, the data property in useError() is a string, not an object.
### Environment

- Node Version:     v20.10.0
- Nuxt Version:     3.13.2
- CLI Version:      3.14.0
- Nitro Version:    2.9.7
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-dcaplg?file=error.vue

### Describe the bug

Access a non-existent page, then capture the error in error.vue, where the data in the error is a string, why isn't it an object?

### Additional context

_No response_

### Logs

 WARN  [Vue Router warn]: No match found for location with path "/xxxx"
 WARN  [Vue Router warn]: No match found for location with path "/xxxx"
 WARN  [Vue Router warn]: No match found for location with path "/xxxx"
Not found: {"path":"/xxxx"}(string)
WARN  [Vue Router warn]: No match found for location with path "/xxxx"

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/29746

### 📚 Description

currently we convert object-format `error.data` to a string when rendering an error page, but this is inconvenient as users have to handle parsing it themselves

it would probably be better to parse it for them, though this might be considered a breaking change (and if so we can defer to v4)


packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts
```diff
@@ -134,17 +134,24 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
...
   if (ssrError && !('__unenv__' in event.node.req) /* allow internal fetch */) {
     throw createError({
       statusCode: 404,
       statusMessage: 'Page Not Found: /__nuxt_error',
     })
   }
 
+  if (ssrError) {
+    ssrError.statusCode &&= Number.parseInt(ssrError.statusCode as any)
+    if (typeof ssrError.data === 'string' && ssrError.data.startsWith('{')) {
```
DamianGlowala, line 146: WDYT about extending this condition to include arrays (valid JSON), or remove it (if not necessary)?
```suggestion
    if (typeof ssrError.data === 'string') {
```

DamianGlowala: Well, coderabbitai was first 😄

packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts
```diff
@@ -134,17 +134,24 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
...
+      try {
+        ssrError.data = JSON.parse(ssrError.data)
```
danielroe, line 148:
```suggestion
        ssrError.data = destr(ssrError.data)
```

PR #31584, opened 2025-03-30
fix: use `unhead` v2 compatible API in `ui-templates`

BobbieGoede:
### 🔗 Linked issue
* [Issue was described in Discord](https://discord.com/channels/473401852243869706/473406506579263488/1355853733094752366) 

### 📚 Description
The NuxtWelcome component lost its styling with compat v4 enabled as the generated code still used `unhead` v1 API, I believe this change should be backwards compatible with v1 too?

/cc @harlan-zw


BobbieGoede (commented):
These automated fixes don't look so good

TheAlexLichter:
> I believe this change should be backwards compatible with v1 too?

If I remember correctly backwards-compatibility is only enabled in Nuxt 3 (and not with the Nuxt 4 compat mode anymore) 🤔

BobbieGoede:
> > I believe this change should be backwards compatible with v1 too?
> 
> If I remember correctly backwards-compatibility is only enabled in Nuxt 3 (and not with the Nuxt 4 compat mode anymore) 🤔

Oh I was referring to the property change from `children` to `innerHTML`, it looks like this was already supported in unhead v1 https://v1.unhead.unjs.io/usage/guides/inner-content. so this change should work with both unhead v1 and v2.

Are you referring to something else? 😅

danielroe (approved):

harlan-zw:
In Nuxt v3 we continue to support `children` in Nuxt v4 it breaks, so PR LGTM. 

Worth noting that for styles we should try and use `textContent` over `innerHTML`, it's only supported as there were some reported encoding issues around using `textContent` with styles which I probably need to look into again

PR #31609, opened 2025-03-31
docs: add best practices section

huang-julien:
### 📚 Description

cc @Baroshem


Baroshem:
Awesome, thanks!

I will contribute to it. Should I fork the repo and create a PR to this branch?

I will add docs about Performance and Security :)

huang-julien:
We can probably iterate through multiple PRs. :+1:
This PR was created to lay the foundation for the entire "best practices" section.

docs/2.guide/0.index.md
```diff
@@ -19,4 +19,7 @@ surround: false
   ::card{icon="i-lucide-book-open" title="Recipes" to="/docs/guide/recipes"}
   Find solutions to common problems and learn how to implement them in your Nuxt project.
   ::
+  ::card{icon="i-lucide-square-check" title="Best Practices" to="/docs/guide/best-practices"}
+  Learn about Best Practices when developing with Nuxt
```
danielroe, line 23:
```suggestion
  Learn about best practices when developing with Nuxt
```

docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,305 @@
...
+)
+// queue events to be sent when ga loads
+proxy.gtag('config', 'UA-123456789-1')
+// or wait until ga is loaded
+onLoaded((gtag) => {
+  // script loaded
+})
+```
+
+:read-more{title="Nuxt Scripts" to="[/docs/api/components/nuxt-link](https://scripts.nuxt.com/scripts)"}
```
xjccc, line 211: Is there something wrong?

Baroshem: Good catch! 

@huang-julien can you fix it? I dont have access to the PR so I would need to create another PR to fix this small typo issue :)

docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,305 @@
...
+})
+```
+
+Nuxt server will automatically register corresponding middleware and wrap routes with cache handlers using Nitro caching layer.
+
+:read-more{title="Hybrid rendering" to="/docs/guide/concepts/rendering#hybrid-rendering"}
+
+### Lazy Loading Components
+
+To dynamically import a component (also known as lazy-loading a component) all you need to do is add the Lazy prefix to the component's name. This is useful if the component is not always needed.
```
divagueame, line 75: I found it confusing when using this in the docs. It seems adding the 'Lazy' prefix to the componet's name is not enough, the component still needs to use v-if with the ref, otherwise it won't be lazy loaded. 
Just doing this won't lazy load the component:
` <LazyMountainsList />`

I wrote about this here as personal notes:
https://nuxt-loading-guide.vercel.app/components

divagueame (approved):
so cool. Loved it. Very useful summary.

huang-julien:
@danielroe @atinux WDYT about merging this then we could iterate on it ?

danielroe:
always up for that!

danielroe (approved):

martinszeltins:
Regarding Data Fetching, I think another best practice is to avoid calling `await useFetch` multiple times one after another during SSR but to use `Promise.all` to avoid waterfall effect.

Baroshem:
@martinszeltins great idea! Would you be interested in contributing to the docs with this example? :)

martinszeltins:
> @martinszeltins great idea! Would you be interested in contributing to the docs with this example? :)

Yes, I could open a PR regarding this. I found this nice article that explains it very well: https://www.debugbear.com/blog/nuxt-ssr-performance#optimize-useasyncdata-and-usefetch-for-efficient-data-fetching

### Fetch data in parallel to reduce latency.

When requests don't rely on each other, we can make them in parallel with Promise.all() to boost performance

const { data } = await useAsyncData(() => {
  return Promise.all([$fetch("/api/products/"), $fetch("/api/category/shoes")]);
});

PR #31619, opened 2025-04-01
test(nuxt): add customizable test api for pages tests

OrbisK:
### 📚 Description

Watching @danielroe [suffer while commenting out tests](https://www.twitch.tv/videos/2409771304?t=0h13m10s) to skip them, I tried to update the test cases so that you can pass a custom test api (which defaults to `it`).

Now you can pass `testApi: it.fails` to mark tests as failing, `it.skip` or `it.only` to skip tests.

Example

```ts
 const tests: Array<TestAPI> = [
    { 
        description: 'should handle overrides', 
        testApi: it.fails, // or it.only 
        override: 'key', 
        route: getRouteProps(), 
        output: 'key' 
    },
]    
```

Daniel mentioned that it should be possible to run individual tests with `it.each`, but from the docs I do not see any way to run partial `it.each` tests.


danielroe (approved):
thank you! ❤️

packages/nuxt/test/pages.test.ts
```diff
@@ -22,6 +24,7 @@ describe('pages:generateRoutesFromFiles', () => {
     description: string
     files?: Array<{ path: string, template?: string, meta?: Record<string, any> }>
     output?: NuxtPage[]
+    focus?: boolean
```
OrbisK, line 27: Can you explain your thinking behind Focus instead of Flexible Test Variants? I thought it might be good to allow committing failing or todo tests as well.

danielroe: ahhh! I was only imagining a local workflow of skipping them.

danielroe: reverting back!

PR #31624, opened 2025-04-01
docs: add performance best practices section

Baroshem:
This pull request adds a comprehensive guide on performance best practices for Nuxt applications to the documentation. The guide covers built-in solutions, plug-in solutions, profiling tools, common problems, and useful resources to help developers optimize the performance of their Nuxt apps.

Key additions include:

### Built-in Solutions
* Explanation of `<NuxtLink>` for smart prefetching and server components for server-rendering individual components.
* Details on hybrid rendering, lazy loading components, and lazy hydration to optimize JavaScript bundle size and performance.
* Usage of `useFetch` and `useAsyncData` composables to avoid redundant data fetching on the client and server.

### Plug-in Solutions
* Introduction of `Nuxt Image` for image optimization, `Nuxt Fonts` for font optimization, and `Nuxt Scripts` for efficient loading of third-party scripts.

### Profiling Tools
* Overview of tools like `Nuxi Analyze`, `Nuxt DevTools`, Chrome DevTools, PageSpeed Insights, and WebPageTest to measure and analyze performance.

### Common Problems
* Identification and solutions for common performance issues such as overusing plugins, unused code, not using Vue performance tips, not following patterns, and loading everything at once.

### DIFF
--- a/docs/2.guide/5.best-practices/performance.md
+++ b/docs/2.guide/5.best-practices/performance.md
@@ -6,11 +6,11 @@ description: Best practices for improving performance of Nuxt apps
 
 # Performance
 
-Nuxt comes with several built-in optimizations designed to improve your application's performance and contribute to better [Core Web Vitals](https://web.dev/articles/vitals) score. Additionally, there are also multiple plug-in solutions such as modules that will help improve performance and User Experience even more. This guide outlines best practices to optimize performance of your Nuxt application.
+Nuxt comes with several built-in features designed to improve your application's performance and contribute to better [Core Web Vitals](https://web.dev/articles/vitals). There are also multiple Nuxt core modules that assist in improving performance in specific areas. This guide outlines best practices to optimize performance of your Nuxt application.
 
-## Built-in Solutions
+## Built-in Features
 
-Nuxt offers several built-in solutions that help you optimize performance of your website. Understanding how this works and how you can further customize it to suit your needs is crucial for achieving blazingly fast performance.
+Nuxt offers several built-in features that help you optimize performance of your website. Understanding how these features work is crucial for achieving blazingly-fast performance.
 
 ### Links
 
@@ -27,7 +27,21 @@ Nuxt offers several built-in solutions that help you optimize performance of you
 
 Nuxt automatically includes smart prefetching. That means it detects when a link is visible (by default), either in the viewport or when scrolling and prefetches the JavaScript for those pages so that they are ready when the user clicks the link.
 
-Read more about it [here](https://nuxt.com/docs/api/components/nuxt-link).
+You can also opt for prefetching on interaction instead:
+
+```ts
+export default defineNuxtConfig({
+  experimental: {
+    defaults: {
+      nuxtLink: {
+        prefetchOn: 'interaction',
+      },
+    }
+  }
+})
+```
+
+:read-more{title="NuxtLink" to="/docs/api/components/nuxt-link"}
 
 ### Server Components
 
@@ -44,7 +58,7 @@ Server components are currently experimental and in order to use them, you need 
 </template>
 ```
 
-Read more about it [here](https://nuxt.com/docs/guide/directory-structure/components#server-components).
+:read-more{title="Server components" to="/docs/guide/directory-structure/components#server-components"}
 
 ### Hybrid Rendering
 
@@ -73,11 +87,12 @@ export default defineNuxtConfig({
 
 Nuxt server will automatically register corresponding middleware and wrap routes with cache handlers using Nitro caching layer.
 
-Read more about it [here](https://nuxt.com/docs/guide/concepts/rendering#hybrid-rendering).
+
+:read-more{title="Hybrid rendering" to="/docs/guide/concepts/rendering#hybrid-rendering"}
 
 ### Lazy Loading Components
 
-To dynamically import a component (also known as lazy-loading a component) all you need to do is add the Lazy prefix to the component's name. This is particularly useful if the component is not always needed.
+To dynamically import a component (also known as lazy-loading a component) all you need to do is add the Lazy prefix to the component's name. This is useful if the component is not always needed.
 
 ```html
 <script setup lang="ts">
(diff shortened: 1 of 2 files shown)


docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,346 @@
...
+
+```ts
+<script setup lang="ts">
+const { data } = await useAsyncData('users', () => fetchUsers())
+</script>
+```
+
+The first argument of `useAsyncData` is a unique key used to cache the response of the second argument, the querying function.
+
+Data fetching composables will wait for the resolution of their asynchronous function before navigating to a new page by using Vue’s Suspense. This feature can be ignored on client-side navigation with the `lazy` option but requires manual handling of loading state:
```
cernymatej, line 128: maybe we could mention the difference of using `await` for `useFetch` and `useAsyncData` here?
meaning that on the server, it will still wait for the data to be fetched before rendering the template and on the client, it will behave like the `lazy` version?

would there be a meaningful performance difference if the component's setup function wasn't async?

docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,346 @@
...
+## Common problems
+
+When building more complex Nuxt applications, you will probably encounter some of the problems listed below. Understanding these problems and fixing them will help you improve performance of your website.
+
+### Overusing plugins
+
+**Problem**: A large number of plugins can cause performance issues, especially if they require expensive computations or take too long to initialize. Since plugins run during the hydration phase, inefficient setups can block rendering and degrade the user experience.
+
+**Solution**: Inspect your plugins and see if some of them could be implemented rather as a composable or utility function instead.
+
```
cernymatej, line 315: since, as far as I'm aware, plugins are still sequential by default at the moment, I think it would be a good idea to recommend to use the object syntax for plugins and set the `parallel: true` option with the possibility to set plugin dependencies using `dependsOn:`

Baroshem: I think this is covered in the plugins file created by @huang-julien in the main branch.

But you are correct :) 

Maybe we could sdd a link to the page created by Julien?

docs/2.guide/5.best-practices/performance.md
Baroshem: Good question that I think I dont have an answer for. But maybe @danielroe or @TheAlexLichter will be able to answer?

docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,346 @@
...
+  <NuxtLink to="/about">About page</NuxtLink>
+</template>
+
+<!-- Which will render to with Vue Router & Smart Prefetching -->
+<a href="/about">About page</a>
+```
+
+Nuxt automatically includes smart prefetching. That means it detects when a link is visible (by default), either in the viewport or when scrolling and prefetches the JavaScript for those pages so that they are ready when the user clicks the link.
+
+Read more about it [here](https://nuxt.com/docs/api/components/nuxt-link).
```
danielroe, line 30: (and also throughout the rest of the page, we should omit the protocol/host)

```suggestion
Read more about it [here](/docs/api/components/nuxt-link).
```

docs/2.guide/5.best-practices/performance.md
danielroe: This will only have an effect if plugins are asynchronous. I think I would first advise avoiding any async activity in a plugin, and then fall back to making plugins `parallel`.

danielroe (commented):
This is a great start! Thank you ❤️

docs/2.guide/5.best-practices/performance.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Server Components
+
+Server components allow server-rendering individual components within your client-side apps. It's possible to use server components within Nuxt, even if you are generating a static site. That makes it possible to build complex sites that mix dynamic components, server-rendered HTML and even static chunks of markup.
+
+Server components are currently experimental and in order to use them, you need to enable the 'component islands' feature in your `nuxt.config.ts`, name the file like `HighlightedMarkdown.server.vue`, and use it like following:
```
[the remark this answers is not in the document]
Mini-ghost: Would it generally be more recommended to use a Lazy Hydration Component with `hydrate-never` rather than using an Islands Component (Server Component) in most cases?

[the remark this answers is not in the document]
Baroshem: @huang-julien so you would not include server components at all in this best practices guide?

docs/2.guide/5.best-practices/performance.md
Baroshem: @danielroe 

Would you like me to include this info here in the performance best practices or should it be included in the pkugins page by Julien?

docs/2.guide/5.best-practices/performance.md
Baroshem: @danielroe you marked this as resolved. Shoukd I change it accordingly to your request or leave it as it it? :)

docs/2.guide/5.best-practices/performance.md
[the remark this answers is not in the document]
huang-julien: let's not add it first :+1:

docs/2.guide/5.best-practices/performance.md
huang-julien: I think the performance section could be split into several pages maybe ?

docs/2.guide/5.best-practices/performance.md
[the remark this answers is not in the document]
danielroe: lazy hydration with `hydrate-never` will still hydrate things on client-side navigation unless it was part of the initial render.

docs/2.guide/5.best-practices/performance.md
Baroshem: Ah, you have done it already. Nevermind :)

docs/2.guide/5.best-practices/performance.md
Baroshem: I have been thinking about it but maybe we could do it after merging that? Let's treat it as a MVP to see how these docs look like and if they will grow, we can split this page into separate ones :)

docs/2.guide/5.best-practices/performance.md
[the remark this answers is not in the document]
Baroshem: @danielroe what would be your opinion on that? Should we not add it to the performance best practices?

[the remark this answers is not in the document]
danielroe: if Julien says we should omit server components for now, let's do omit that section for now.

It's still worth mentioning lazy hydration.

docs/2.guide/5.best-practices/performance.md
danielroe: let's start with a single page. we can update the plugins PR from Julien to augment this page + later see if we need to split.

docs/2.guide/5.best-practices/performance.md
danielroe: let's avoid mentioning this for the moment - @TheAlexLichter is working on explaining this simply (and may have found a bug/behaviour that's worth documenting on the data fetching page)

docs/2.guide/0.index.md
```diff
@@ -19,4 +19,7 @@ surround: false
   ::card{icon="i-lucide-book-open" title="Recipes" to="/docs/guide/recipes"}
   Find solutions to common problems and learn how to implement them in your Nuxt project.
   ::
+  ::card{icon="i-lucide-square-check" title="Best Practices" to="/docs/guide/best-practices"}
+  Learn about Best Practices when developing with Nuxt
```
danielroe, line 23:
```suggestion
  Learn about best practices when developing with Nuxt
```

danielroe (approved):

docs/2.guide/5.best-practices/performance.md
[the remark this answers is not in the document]
Baroshem: Ok, will remove it then but keep the hydration part :)

atinux (approved):

PR #31634, opened 2025-04-01
fix(nuxt): throw when multiple definePageMeta in a file

huang-julien:
### 📚 Description

This PR makes the page meta plugin throw an error when a page has multiple `definePageMeta`. This can happen when an user is refactoring their code.


danielroe (commented):
This isn't a safe check as it isn't parsing the code, just doing a 'contains' check - for example, this would likely fail on a page which has a commented out section with an old `definePageMeta`.

(We can safely use it to do an early return as if the string _isn't_ present we know we don't have page meta.)

I think it should be possible to do the check as we walk the AST, however.

danielroe (approved):

PR #31646, opened 2025-04-02
fix(vite): prevent 404 for existent dev server handlers under `_nuxt/` paths in development

### LINKED ISSUE #31639: Adding a dev server handler under `/_nuxt` results in a 404 error
### Environment

```
- Operating System: Linux
- Node Version:     v20.19.0
- Nuxt Version:     3.16.2
- CLI Version:      3.24.0
- Nitro Version:    2.11.8
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
```

## Working on 3.15.4
- https://stackblitz.com/edit/github-6tiz2dme
- Open /_nuxt/demo

## Not working on 3.16.2
- https://stackblitz.com/edit/github-6tiz2dme-ngbwxbv6
- Open /_nuxt/demo

### Describe the bug

When adding a dev server handler in a module using `addDevServerHandler` that uses the `_nuxt` prefix results in a 404 on a vanilla Nuxt installation:

```ts
addDevServerHandler({
  handler: defineEventHandler(() => {
    return 'Foobar /_nuxt/demo';
  }),
  route: `/_nuxt/demo`,
});
```

More specifically, I am using `nuxt.options.app.buildAssetsDir` to derive the prefix without hardcoding `/_nuxt` (I saw this being done in Nuxt's code too, so I took it over).

### Additional context

Obviously I could use any other prefix to fix the issue, but since this is done as part of a public module I'd ideally like to make it work with any Nuxt setup. For example, in most of my projects, Nuxt is proxied using nginx, where some path prefixes are routed to Nuxt and others to a backend. Introducing a new prefix in Nuxt would require changing the rewrites/proxy pass in nginx too. Using the build dir is the safest, because it's guaranteed to be already routed to Nuxt.

### Logs

```shell-script

```

Yizack:
### 🔗 Linked issue

Fixes #31639

### 📚 Description

This PR resolves an issue where dev server handlers added using `addDevServerHandler` with routes under the `_nuxt` prefix (or the configured buildAssetsDir) would result in a 404 error in development

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -248,10 +248,10 @@ export async function buildClient (ctx: ViteBuildContext) {
       }
     }
 
-    const devHandlerBases: RegExp[] = []
+    const devHandlerRegexes: RegExp[] = []
     for (const handler of ctx.nuxt.options.devServerHandlers) {
       if (handler.route && handler.route !== '/' && handler.route.startsWith(ctx.nuxt.options.app.buildAssetsDir)) {
-        devHandlerBases.push(new RegExp(
+        devHandlerRegexes.push(new RegExp(
           `^${handler.route
             .replace(/:[^/]+/g, '[^/]+') // dynamic segments (:param)
             .replace(/\*\*/g, '.*') // double wildcard (**) to match any path
@@ -289,7 +289,7 @@ export async function buildClient (ctx: ViteBuildContext) {
       })
 
       // if vite has not handled the request, we want to send a 404 for paths which are not in any static base or dev server handlers
-      if (!event.handled && event.path.startsWith(ctx.nuxt.options.app.buildAssetsDir) && !staticBases.some(baseURL => event.path.startsWith(baseURL)) && !devHandlerBases.some(baseRegex => baseRegex.test(event.path))) {
+      if (!event.handled && event.path.startsWith(ctx.nuxt.options.app.buildAssetsDir) && !staticBases.some(baseURL => event.path.startsWith(baseURL)) && !devHandlerRegexes.some(regex => regex.test(event.path))) {
         throw createError({
           statusCode: 404,
         })
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -276,8 +276,8 @@ export async function buildClient (ctx: ViteBuildContext) {
         })
       })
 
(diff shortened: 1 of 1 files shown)


danielroe (commented):
this is a good fix but here are some things that would make it better:

- extracting the possible bases ahead-of-time, as we do with `staticBases`
- converting them to regexp (as it's not just a string-to-string comparison as with staticBases), as they can include wildcard/dynamic segments.

Yizack:
Hello, I added the regexp conversion this way, according to https://h3.unjs.io/guide/router to support `:`, `*` and `**`

Dev server routes no longer error in nitro

---

Just as a note, I noticed that when a dynamic route is set, for example:

```ts
addDevServerHandler({
  handler: defineEventHandler(() => {
    return 'Foobar /_nuxt/demo';
  }),
  route: `/_nuxt/demo/**`,
});
```

when accessing `/_nuxt/demo/foo` the `h3` handler function is not resolved in either `3.15.4` or in this commit, only when accessing directly `/_nuxt/demo/foo` the handler resolves, so that might be a matter for another issue?

If you want to take a look at this, here's a reproduction in `3.15.4`:
- https://stackblitz.com/edit/github-6tiz2dme-yh4lqtbn?file=modules%2Fdemo%2Fmodule.ts

packages/vite/src/client.ts
```diff
@@ -248,6 +248,18 @@ export async function buildClient (ctx: ViteBuildContext) {
       }
     }
 
+    const devHandlerRegexes: RegExp[] = []
+    for (const handler of ctx.nuxt.options.devServerHandlers) {
+      if (handler.route && handler.route !== '/' && handler.route.startsWith(ctx.nuxt.options.app.buildAssetsDir)) {
+        devHandlerRegexes.push(new RegExp(
```
danielroe, line 254: we might need to escape special characters in the path, right? (like `.`?)

packages/vite/src/client.ts
```diff
@@ -248,6 +248,18 @@ export async function buildClient (ctx: ViteBuildContext) {
...
+          `^${handler.route
```
Yizack, line 255: You are right

packages/vite/src/client.ts
Yizack: you are right

danielroe (approved):

PR #31657, opened 2025-04-02
feat(kit,schema): new `enforceModuleCompatibility` feature

huang-julien:
### 📚 Description

with https://github.com/nuxt/cli/pull/818 , we can freely throw errors when Nuxt loads. 

This PR improves DX by throwing an error when a module is not compatible with the current version of Nuxt. We've seen issues with  people being surprised by modules (nuxt/scripts) being disabled due to compatiblity.


packages/schema/src/config/experimental.ts
```diff
@@ -589,5 +589,10 @@ export default defineResolvers({
         return typeof val === 'boolean' ? val : true
       },
     },
+
+    /**
+     * Whether Nuxt should stop if a Nuxt module is incompatible.
+     */
+    enforceModuleCompatibility: false,
```
huang-julien, line 596: Maybe can this be moved into `future` instead and be enabled by default for v4 ?

danielroe: Let's enable it by default for v4 (in the experimental namespace), and document in the migration guide. This can be done in a patch release if necessary.

danielroe (approved):

PR #31676, opened 2025-04-04
fix(nuxt): sync error types for `useLazyAsyncData`

### LINKED ISSUE #31669: Different types of errors in useLazyAsyncData and useAsyncData with lazy: true.
Different types of error in useLazyAsyncData and useAsyncData with lazy: true.

useLazyAsyncData: 

Image

useAsyncData with lazy true: 

Image

In the first case, the error is of type Error, while in the second, it's of type NuxtError, but the statusCode property exists on the error object in both cases. 

Image

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31669

### 📚 Description

It seems when we improved types for `useAsyncData`, we neglected to do the same for `useLazyAsyncData`.

This PR updates it and adds JSdocs for `useLazyFetch`.


artem-kostylev:
Tnx 👍

PR #31681, opened 2025-04-04
feat: stop showing compatibility date prompt after first time

HigherOrderLogic:
### 🔗 Linked issue

N/A.

### 📚 Description

It's quite annoying for Nuxt to keep asking whether I want to set the compatibility date every time after starting up the dev server, so I made it remember the user's choice.


pi0:
couldn't nuxt CLI always add it by default when initializing project?

HigherOrderLogic:
> couldn't nuxt CLI always add it by default when initializing project?

I prefer to not set it in the config file so the compatibility date will always be the day that I `dev`/`build` the project.

danielroe:
that means you are opting into potentially breaking changes from upstream providers

HigherOrderLogic:
> that means you want to opt into potentially breaking changes from upstream providers?

Yes.

> also, if you don't set it I'm pretty sure we hard code it to a particular date - it doesn't become the date you build your project
> 
> https://github.com/nuxt/nuxt/blob/1f6e6d3f64463c98729aa2a816f1d910fb898887/packages/nuxt/src/core/nuxt.ts#L143

Seems like I have mistaken the date.

But I think it's still a good idea to keep this PR since some people may not need to care about setting and updating the date so it's maybe a good idea to just ask them once.

PR #31726, opened 2025-04-07
ci: add notify-nuxt-website workflow

atinux:
The goal is to trigger a new deployment once the docs for v3 is updated.

### DIFF
--- a/.github/workflows/notify-nuxt-website.yml
+++ b/.github/workflows/notify-nuxt-website.yml
@@ -1,7 +1,6 @@
 name: notify-nuxt-website
 on:
-  pull_request:
-    types: [closed]
+  push:
     branches:
       - 3.x
     paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-nuxt-website.yml
```diff
@@ -0,0 +1,26 @@
+name: notify-nuxt-website
+on:
+  pull_request:
+    types: [closed]
+    branches:
```
danielroe, line 5:
```suggestion
  push:
    branches:
```

.github/workflows/notify-nuxt-website.yml
```diff
@@ -0,0 +1,26 @@
...
+      - 3.x
+    paths:
+      - "docs/**"
+
+permissions: {}
+
+jobs:
+  notify:
+    if: github.event.pull_request.merged == true
```
atinux, line 14: Based on your suggestion, I guess we should remove this line too @danielroe ?

```suggestion
```

danielroe (approved):

PR #31728, opened 2025-04-08
docs: warning fetch onRequest to set params

xjccc:
### 📚 Description

From: https://github.com/unjs/ofetch/issues/477

And https://github.com/unjs/ofetch/issues/378

It might be deprecated in the next version. So add a warning for users.


danielroe (commented):
I think it would be better not to warn against using something before it is deprecated. (And when it _is_ deprecated then users will get the warning in their IDE.)

It is better to fix the issue (https://github.com/unjs/ofetch/pull/478) rather than adding a warning to the docs, I think.

PR #31747, opened 2025-04-09
docs: adjust callOnce fn description

xjccc:
### 📚 Description

It's a bit confusing with “return a Promise and a value”. Because `callOnce` not return the value.


docs/3.api/3.utils/call-once.md
```diff
@@ -86,7 +86,7 @@ type CallOnceOptions = {
 ## Parameters
 
 - `key`: A unique key ensuring that the code is run once. If you do not provide a key, then a key that is unique to the file and line number of the instance of `callOnce` will be generated for you.
-- `fn`: The function to run once. This function can also return a `Promise` and a value.
+- `fn`: The function to run once. This function could be async or not.
```
danielroe, line 89:
```suggestion
- `fn`: The function to run once. It can be asynchronous.
```

PR #31760, opened 2025-04-10
docs: use `vs.` consistently

iiio2:
I think the correct form is  "vs." with a period.


danielroe:
I prefer `vs` personally. (both are valid forms, but vs is used more in british english and `vs.` in american). as we aim for american english in the docs, so we'll move to `vs.` as you've suggested.

I've also updated other instances.

iiio2:
Thanks @danielroe . 😊

danielroe (approved):

PR #31766, opened 2025-04-10
feat(nuxt): add showError option to `useAsyncData`

huang-julien:
### 📚 Description

This can be quite useful when users think that some `useAsyncData` is mandatory for a page. Instead of having to watch an error, we can directly trigger a `showError`


huang-julien:
yes, if the handler throws, `showError` will be triggered

Aareksio:
You probably want to modify `useFetch` too, as it shares the options with `useAsyncData`, but currently catches `showError` option and forwards it to fetch instead. https://github.com/nuxt/nuxt/pull/31827/files#diff-3ace8ffd6a6f0f4eb7979e99189c489e15fa917b74b48c26ee7e12ad13dfb866

PR #31791, opened 2025-04-14
fix(nuxt): wait for suspense to resolve before handling `NuxtErrorBoundary` error

### LINKED ISSUE #25869: NuxtErrorBoundary does not catch all client-side errors
### Environment

nuxt 3.10.2

### Reproduction

Repository: https://stackblitz.com/edit/github-tobvzw-jwk7ax

Just load the application, you will be presented with the "500" error view of nuxt. The expected behavior is that the text "This is the error rendering" is rendered. 

If one would comment out the `onMounted` hook in `my-broken-component.vue` and click on the rendered paragraph, the `NuxtErrorBoundary` works as expected.

### Describe the bug

It seems that client-side thrown errors will pass through the `NuxtErrorBoundary` component if they are thrown right after being mounted.
In the stackblitz example, the error is thrown in the `onMounted` lifecycle hook. This is also the case if one would watch a value coming from `useAsyncData` (with `{server:false, immediate:true}` as configuration) and throw based on that. So it appears to be a timing issue when throwing errors in the client too early.

This only happens when SSR is enabled.

### Additional context

_No response_

### Logs

_No response_

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/25869

### 📚 Description

Currently, throwing an error in an `onMounted` hook doesn't get caught by the `<NuxtErrorBoundary>`. This PR delays handling an error until the suspense resolves (if the error occurs during hydration).


danielroe (commented):
this looks good! any chance you could add some tests?

danielroe:
added some basic runtime tests in https://github.com/nuxt/nuxt/commit/3a027b356b71522786c1d14532b951538ba21798.

danielroe (approved):

PR #31792, opened 2025-04-14
docs: capitalize title

iiio2:
SEO and Meta. Here I think Meta should be with capital `M`. Because I think in other place the title is given like `SEO and Meta`.


danielroe (commented):
this doesn't need to be capitalised with title case because it is used within a sentence

iiio2:
Oh, I see. Ok @danielroe . Thanks.

PR #31804, opened 2025-04-15
docs: best practices security

Baroshem:
Hey there!

This PR aims to add Best Practices about Security for Nuxt.

Let me know about the feedback! :)

### DIFF
--- a/docs/2.guide/5.best-practices/security.md
+++ b/docs/2.guide/5.best-practices/security.md
@@ -1,0 +1,306 @@
+---
+navigation.title: 'Nuxt Security'
+title: Nuxt security
+description: Best practices for making Nuxt apps more secure.
+---
+
+Nuxt applications can be affected by the same security risks as any other web application. To protect against them, it is crucial to be aware what kind of attacks and risks are out there. This guide outlines best practices to make your Nuxt application more secure.
+
+## Risks
+
+The best source of knowledge for most popular security risks is the [OWASP Top 10](https://owasp.org/www-project-top-ten/) that is a standard awareness document for developers and web application security. Let's talk about few of them that we can handle on Nuxt.
+
+### Broken Access Control
+
+Access control ensures users can only perform actions within their permissions. Failures can lead to unauthorized data access, changes, deletion, or misuse of business functions.
+
+A simple example attack scenarion can be:
+```
+An attacker tries to access admin page when unauthenticated or not-admin user and gets the access to
+https://example.com/admin/sensitive-data
+```
+
+To protect against such attack, you could implement a proper authorization. Check out Nuxt Authorization [below](#nuxt-authorization).
+
+:read-more{title="Broken Access Control" to="https://owasp.org/Top10/A01_2021-Broken_Access_Control/"}
+
+### Denial of Service
+
+A Denial of Service (DoS) attack focuses on making a resource (such as a site, application, or server) unavailable for its intended purpose. This can be achieved through various methods, including manipulating network packets, exploiting programming or logical vulnerabilities, or mismanaging resources, among others.
+
+A simple example attack scenarion can be:
+```
+An attacker tries to overload the application server with too many or too heavy requests to make it unable to respond and stop working properly.
+```
+
+To protect against such attack, you could implement a proper rate limiting solution. Check out Nuxt Security [below](#nuxt-security).
+
+:read-more{title="Denial of Service" to="https://owasp.org/www-community/attacks/Denial_of_Service"}
+
+### Injections
+
+An application is vulnerable when user input isn't properly validated or sanitized, dynamic queries lack parameterization or escaping, and untrusted data is used in ORM queries or concatenated into SQL, commands, or stored procedures. Some of the more common injections are SQL, NoSQL, OS command, Object Relational Mapping (ORM), LDAP, and Expression Language (EL) or Object Graph Navigation Library (OGNL) injection.
+
+A simple example attack scenarion can be:
+
+```
+An application uses untrusted data in the construction of the following vulnerable SQL call:
+SELECT \* FROM accounts WHERE custID='" + request.getParameter("id") + "';
+```
+
+To protect against such attack, you could implement a proper sanitization and escaping. Check out Nuxt Security [below](#nuxt-security).
+
+:read-more{title="Injection" to="https://owasp.org/Top10/A03_2021-Injection/"}
+
+### Outdated components
+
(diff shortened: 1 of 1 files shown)


huang-julien (commented):
Do we want to focus on everything ? 

I was thinking about only having Nuxt/SSR frameworks specific security issues 🤔 .  Like cross-requets state pollution.

Baroshem:
Hey @huang-julien 

It is still not everying - it is a tip of the iceberg honestly :D 

My idea when writing it was to focus on the aspects that can affect Nuxt apps and developers who are developing them. So basically, I focused on risks and how Nuxt devs can solve them - by either using a module such as Nuxt Security, Nuxt Auth Utils or Nuxt Authorization, or by configuring Nuxt core itself (useState, runtimeConfig, etc).

The idea of this page in my head was to have is a simpler source of knowledge about security than what I have in Nuxt Security docs and what is included in the OWASP Top 10 page. So developers are more aware about the risks and how to solve them in Nuxt ecosystem.

I think it is similar with Performance, we didn't just include the Nuxt specific things (such as Lazy Hydration for example) but also things like assets - images, fonts, etc - that are not Nuxt specific but websites in general.

This is also one of the things that I think Nuxt docs can do better than Next.js - they focus on CSP only AFAIK.

Let me know what you think about it :)

docs/2.guide/5.best-practices/security.md
```diff
@@ -0,0 +1,306 @@
...
+
+export function useSafeRef() {
+  const safeRef = useState<number>('safeRef', () => 2);
+
+  return {
+    safeGlobal,
+    safeRef
+  }
+}
+```
```
huang-julien, line 267: that's not true. Having useState in a global var will make all request using this file fail because `useState` is called at file import -> nuxt will not be available

docs/2.guide/5.best-practices/security.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Common problems
+
+When building more complex Nuxt applications, you will probably encounter some of the problems listed below. Understanding these problems and fixing them will help make your website more secure.
+
+### Returning entities with sensitive data
+
+**Problem**: When fetching the data from external source (i.e. Database) and returning it in the Nuxt API endpoint, you may return the whole table entity (such as User) to the browser and handle it there. This can lead to unintentional leaking of sensitive information that may not even be needed. 
+
+**Solution**: Utilize `pick` or use `interceptors` concepts for both `useAsyncData` and `useFetch` to sanitize what you get from the server without leaking confidential data.
```
huang-julien, line 225: This isn't considered as a security improvement but more of a performance improvement. Because it allows to reduce the amount of payload in the SSR HTML
Sensitive data must always be filtered by the API and not by the consumer.

huang-julien (changes requested):

docs/2.guide/5.best-practices/security.md
Baroshem: Agree, better to have everything filtered in the API/backend side. But if it is not there (maybe a mistake on the API/Backend side) we can always pick just the information we need. That's what I had in mind when I was adding this point. But we can also remove it if you think that it won't bring value and cause confusion instead :)

docs/2.guide/5.best-practices/security.md
huang-julien: Are you talking about sharing something among all SSR requests ? 
If so, i'd use a nitro plugin to inject into the event. But i'm always against the usage of objects in the global scope anyway. It"s way too easy to cause a memleak.

For global state management, within a single NuxtApp. There's a few solutions. 
Use `useState` in a composable. UseState is a state within the NuxtApp. 
Use a NuxtPlugin to provide data
Use Pinia (which has a plugin then provide composables)

docs/2.guide/5.best-practices/security.md
huang-julien: :+1: You can move it to the performance part. but it still need to rephrase/remove the next line.

> This can lead to unintentional leaking of sensitive information that may not even be needed. 

The API returning a whole table entity as a response is still a data leak. Even tho the Users may not see it, people with bad intentions will look at the response. An API is and should always  be considered as something public and accessible.

docs/2.guide/5.best-practices/security.md
```diff
@@ -0,0 +1,306 @@
...
+
+Nuxt Security helps you increase the security of your site by enabling Strict CSP support for both SSR and SSG applications.
+
+1. For `SSR` applications, Nuxt Security implements strict CSP via nonces. A one-time cryptographically-generated random nonce is generated at runtime by the server for each request of a page. In this rendering mode, you can also use all security headers in a regular form of a response header (including headers such as Cross Origin Embeder Policy, Cross Origin Opener Policy, etc).
+2. For `SSG` applications, Nuxt Security implements strict CSP via hashes. At static build-time, Nuxt Security computes the SHA hashes of the elements that are allowed to execute on your site. The CSP header value is then delivered as `<meta http-equiv>` tag and you can also implement Subresource Integrity that enables the browser to verify that the static assets that your application is loading have not been altered.
+
+:read-more{title="Strict Content Security Policy" to="https://nuxt-security.vercel.app/advanced/strict-csp"}
+
+
+## Third-party modules
```
huang-julien, line 105: btw, IMO it is probably not a good idea to mention specifics modules that are not part of the official modules.

This can fracture the community as we're showing a kind of favoritism toward some maintainers/authors.

Baroshem: Agree about favoritism but the problem is that there is no official module such as Nuxt Image, Nuxt Scripts, etc. All these modules are community - not official modules and I thought that we should at least recommend something that will help users find answers for their questions or needs.

@danielroe what are your thoughts on that? Both about that and maybe about the general direction of this particular docs. Should we focus on security issues coming from OWASP and how to solve them in Nuxt? Or maybe just about Content Security Policy? But then, we will be missing content about other risks that are equally or even more important. 

I would love to get your feedback on that to better understand how to write this properly :)

huang-julien: > the problem is that there is no official module such as Nuxt Image, Nuxt Scripts, etc.

That's an excellent point you made here.
I personnally think that authorizations and auth-utils should remain outside nuxt core and nuxt official modules as there's different ways to implement authorizations/auth. We could always document anti-patterns tho. 
Some solutions provided by nuxt-security should indeed be handled, not by Nuxt but probably by Nitro since things like CSP could be built-in within Nitro.

Let's wait for @danielroe 's opinion :eyes:

Baroshem:
Closing this PR for now as we discussed that the different approach should be selected here.

I will focus on https://github.com/nuxt/nuxt/discussions/30404 and then in this security docs we will focus the best practices for content security policy.

Thanks for the comments!

PR #31812, opened 2025-04-16
feat(nuxt,vite): add experimental support for `rolldown-vite`

### LINKED ISSUE #30654: Support Rolldown
### Describe the feature

Rolldown is compatible with Rollup, so I expect to be able to switch with a little changes.
But because there are some differences with Rollup, it is not possible to build Nuxt currently.

`Nuxt Build Error: Failed to convert JavaScript value function anonymous(..)  into rust type String on BindingOutputOptions.assetFileNames on BindingBundlerOptions.outputOptions`

Rolldown does not allow `function` in `assetFileNames`(https://github.com/rolldown/rolldown/issues/3253)

### Additional information

### Final checks

TheAlexLichter:
Resolves https://github.com/nuxt/nuxt/issues/30654
[PR running with the actual rolldown-vite dep for testing](https://github.com/nuxt/nuxt/pull/31880)

Current issues:

* SSR Inline Styles seem to be not working yet due to rolldown/rollup differences I assume. 
	* Needs further investigation.
* ~~Watch mode for `build` [is not implemented in `rolldown-vite` at the moment](https://github.com/vitejs/rolldown-vite/pull/84). We ignore the watch options in this PR for now.~~
  * ~~Partially available (but of course no chokidar watching). Should be worked around in the future if possible~~
  * ~~Needs to be implemented in the PR when available.~~
* ~~Used Rollup plugins throw a type error (https://github.com/vitejs/rolldown-vite/issues/117)~~
	* ~~This is available upstream now but pending a release (likely 6.3.13). ~~
* ~~Can't easly detect Rolldown in an unplugin vite `config` hook yet~~
	* ~~workaround exists but would lead to implicitly bundling vite (meh!)~~
	* ~~Experimental option is the way to go now I'd say~~
* ~~esbuild options should be moved/merged/ignored (as only rollupOptions are used)~~
	* ~~This is just a TODO and can be implemented in this PR already~~
* Certain Rollup options are not supported at the moment
  * ~~`preserveModules` - https://github.com/rolldown/rolldown/issues/2622~~ implemented
  * `generatedCode` - https://github.com/rolldown/rolldown/issues/206


TheAlexLichter:
image
That seems new 👀🤔

Tests are failing at the moment (at least locally).

We should be able to ship it with an experimental flag, yes!

huang-julien:
I got an issue with `vite:json` when using `@nuxtjs/i18n`.
I haven't debugged it yet. But @BobbieGoede and @kazupon are probably interested about that.

kazupon:
related issue
https://github.com/intlify/bundle-tools/issues/483#issuecomment-2811554778

@BobbieGoede 
I'v just commented that issue.
We can do the same thing by using the hook's `order` and `handler`

lansi951:
How about using the `experimental.enableNativePlugin` option?
It helps improve performance.
However, it doesn't work properly in dev mode currently, so should use the `resolver` in dev mode.

nuxt.config.ts:
```ts
$development: {
  vite: {
    experimental: {
      enableNativePlugin: 'resolver',
    },
  },
},
$production: {
  vite: {
    experimental: {
      enableNativePlugin: true,
    },
  },
}

**Comment by TheAlexLichter** [comment]
> @TheAlexLichter do you think we could ship under an experimental flag? users would still need to manually set a resolution for `vite`, I think...

I wonder if we could even ship it unflagged as we have rolldown detection in place everywhere ☺️

File: packages/nuxt/src/components/plugins/islands-transform.ts [hunk 1]
```diff
@@ -198,9 +198,12 @@
         } else if (typeof rollupOptions.input === 'object' && Array.isArray(rollupOptions.input)) {
           rollupOptions.input = rollupOptions.input.reduce<{ [key: string]: string }>((acc, input) => { acc[input] = input; return acc }, {})
         }
-
-        // don't use 'strict', this would create another "facade" chunk for the entry file, causing the ssr styles to not detect everything
-        rollupOptions.preserveEntrySignatures = 'allow-extension'
+        const isRolldown = !!(await import('vite')).rolldownVersion
```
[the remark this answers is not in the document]
TheAlexLichter: nope, it should not. That's on purpose 😋

packages/nuxt/src/components/module.ts
```diff
@@ -253,6 +253,7 @@ export default defineNuxtModule<ComponentsOptions>({
             config.plugins.push(ComponentsChunkPlugin.vite({
               getComponents,
               buildDir: nuxt.options.buildDir,
+              isRolldownCompatEnabled: nuxt.options.experimental.rolldownCompat,
```
danielroe, line 256: is it possible to detect this without requiring the user to set a flag?

danielroe (commented):
this looks great - though I wonder if it might be easier to ship as a standalone module which would make it easier to iterate upon things.

packages/nuxt/src/components/module.ts
TheAlexLichter: I tried before (see PR description [and this commit](https://github.com/nuxt/nuxt/pull/31812/commits/4c95619eb5cedacf4677ec2130ff50a2a0e62ebf#diff-87425e8bfe7538c62da5fe7af431df0b50a5e46704cfcdb6fb525c9607ba03caR201)) but this would've led to implicitly bundling vite. If we can avoid that (externalizing?) I think it'd be possible.

TheAlexLichter:
> this looks great - though I wonder if it might be easier to ship as a standalone module which would make it easier to iterate upon things.

Fair point. That might be an option - though ideally not many changes should be necessary. The changes are even optional at the moment and mainly relevant when not using `ROLLDOWN_OPTION_VALIDATION=loose`.

If a module is the way, we should then make sure it is documented & the module runs for the `rolldown-vite` ecosystem CI as well.

lansi951:
`preserveModules` is supported now.

packages/vite/src/server.ts
```diff
@@ -116,7 +123,17 @@ export async function buildServer (ctx: ViteBuildContext) {
   } satisfies vite.InlineConfig, ctx.nuxt.options.vite.$server || {}))
 
   if (serverConfig.build?.rollupOptions?.output && !Array.isArray(serverConfig.build.rollupOptions.output)) {
-    delete serverConfig.build.rollupOptions.output.manualChunks
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore Rolldown-Vite specific
+    if (!vite.rolldownVersion) {
+      // Rolldown has no support for `output.manualChunks`
+      delete serverConfig.build.rollupOptions.output.manualChunks
```
AaronDewes: Maybe I'm misunderstanding something, but isn't this code only triggered when Rolldown is **not** active? So the comment seems wrong 🤔

TheAlexLichter: Yes, this should only apply for Rollup. As there was no support for `manualChunks` in Rolldown, it should only be removed / reset when using Rollup. This part is in "vanilla nuxt" already (unsetting all manual chunk settings for the server build).

But now that Rolldown *does* support it in a later version, we can even remove the check 😋

danielroe:
@TheAlexLichter do you think we can get this in the next minor?

TheAlexLichter:
Yes!

danielroe (approved):

AaronDewes:
@danielroe Did you do that intentionally? :upside_down_face: 

grafik

danielroe:
🤯

Issue #31743, opened 2025-04-08
useAsyncData - do not swallow errors


Aareksio:
## Feature Request
Allow `useAsyncData` to **optionally** throw error, instead of returning as ref. This should be configurable in nuxt config. With an option to still receive a ref when requested on usage level. Error refs are still crazy useful when needed. The request is not about removing them.

## Explanation
`useAsyncData` is a general purpose composable for anything that needs to be asynchronously loaded. It is recommended and also fairly natural to use for things we need to do only once in SSR application. Its reactiveness works great on happy path. Being resilient by default is beneficial for a lot of projects on the surface. 

However, once you dig deeper, you realize:
- each `useAsyncData` must handle error at component level, it is swallowed otherwise
- handling `error` is not trivial, as it is a reactive value, which may cause over time (eg. via `refresh`)
- throwing proper 404 when external (to nuxt app) API returns error requires a PhD in nuxt, javascript and SEO
- properly handling errors makes detecting issues easier - interpreting empty data as generic "no content" hides problems
- most components do not need custom error handlers, not by default, if they do, developers will add them

Opt-in throwing saves a ton of repetitive, often forgotten or invalid boilerplate. It could only be enabled by projects valuing the points mentioned above, perhaps recommended as a good practice. Casual users can still be resilient-by-default.

## Motivation (yapping, non technical)
Hi! Error handling in nuxt 3 has always been a pain point of mine. 

In the early days it was quite disturbing. Then we got `showError`, it was still terrible, but at least errors got acknowledged. Later on `showError` was recommended to be avoided in favor of `throw`s. Which still has issues (#28557), but mostly works. Nowadays the apps can do mostly fine. As long as they remember `useAsyncData` returns an error, check it and throw. And maybe remember it is reactive... Did I mention it is returned, not thrown? The fact developers have to remember all this is daunting. Every time we need a simple fetch, there are pitfalls waiting for the developer to step in.

I have been tracking many issues and discussions. Developers have hard time understanding how to handle errors in Nuxt 3. This includes the newcomers who try nuxt for the first time and have trouble throwing 404; experienced folk who still struggle with reactiveness (let's not pretend, everyone does!); and business people, who see their page not working without any error message anywhere (including console / error trackers).

---

I am not pointing fingers, I started by inspecting own projects, they are much worse. However, let's use nuxt.com as an example. I think it is fair to assume the developers responsible for the website have above the average access to Nuxt knowledge. Yet:

[...]

PR #31827, opened 2025-04-17
feat(nuxt): add `throwOnError` option to `useAsyncData`

Aareksio:
### 🔗 Linked issue

#31743

### 📚 Description

Inspired by #31766. I am not sure whether the features are conflicting or complimentary. Main benefit over `showError`:

```ts
// Before - optional-chaining, if checks when not feasible
const { data } = await useFetch<Post>('/blog/404')
const upvote = () => data.value ? $fetch(`/${data.value.id}/upvote`) : {}
const loadAuthor = () => {
  if (!data.value) { return }
  return $fetch(`/author/${data.value.meta.authorId}`)
}

// After - data is always available, safe to bang, types could further be improved
const { data } = await useFetch<Post>('/blog/404', { throwOnError: true })
const upvote = () => $fetch(`/${data.value!.id}/upvote`) 
const loadAuthor = () => $fetch(`/author/${data.value!.meta.authorId}`)
```

Ideally, we would get the return type as `Post`, not `Post | undefined`.
Draft PR, because I think this is the clue: https://github.com/nuxt/nuxt/issues/31743#issuecomment-2811312234

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -97,6 +97,15 @@ export interface AsyncDataOptions<
    * @default 'cancel'
    */
   dedupe?: 'cancel' | 'defer'
+
+  /**
+   * Throw an error if handler function fails
+   * Accepts a boolean value or a function that returns a NuxtError object.
+   * If set to true, the returned promise will reject when an error occurs.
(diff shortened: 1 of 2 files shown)


daniluk4000:
This is a long, long needed one

PR #31840, opened 2025-04-18
ci: bump node v20

iiio2:
Actually node v18 EOL is April 30, 2025, which is within in 2 weeks I guess. I mean can we update it to v20 actually.


danielroe:
we can do so after that date but not until then

in general, there is no need to raise PRs updating dependencies or versions as this is normally something I do. (feel free to check with me directly if you think I've missed a version bump).

iiio2:
Oh, I see. Ok @danielroe . Thanks. 😊

PR #31847, opened 2025-04-19
feat(nuxt): convert `<NuxtErrorBoundary>` to SFC; expose `error` and `clearError`

### LINKED ISSUE #31846: NuxtErrorBoundary: expose `clearError`
### Environment

- Operating System: Linux
- Node Version:     v20.19.0
- Nuxt Version:     3.16.2
- CLI Version:      3.25.0
- Nitro Version:    2.11.9
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-5kudj1oz

### Describe the bug

Currently there is no (easy) way to clear the error from `<nuxt-error-boundary>` in script. This makes it complicated to use when component is re-used. For example when nesting pages, see reproduction.

https://github.com/user-attachments/assets/37cc340b-cfe1-41e5-a378-8ce4fa80ab47

### Additional context

I do not think nuxt should attempt to be clever here. `defineExpose({ clearError })` seem to be enough. Especially as error boundary is rarely used component anyway, it does not require anything but the bare minimum effort to make it work.

### Logs

```shell-script

```

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31846

### 📚 Description

- converts `<NuxtErrorBoundary>` to SFC
- exposes `error` and `clearError`
- changes `error` from `ref` to `shallowRef`
- updates docs with an example (and adds missing `clearError` to section showcasing `error` slot usage)


packages/nuxt/src/app/components/nuxt-error-boundary.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <slot
+    v-if="error"
+    :error="error"
+    :clear-error="clearError"
```
danielroe, line 5:
```suggestion
    v-bind="{ error, clearError }"
```

huang-julien (approved):

danielroe (approved):

Tofandel:
FIY

This is a breaking change

Previously error was exposed as a ref to the template, meaning you could (and had to) do this

 ```vue
<template #error="{ error }">
  <ErrorPageContent :error="error.value" @clear="clearErr(error)"></ErrorPageContent>
</template>
         
<script setup>
function clearErr(error: Ref<unknown>) {
  nextTick(() => {
    error.value = null;
  });
}
</script>
```

Of course the new way is better, but this just broke my error page

danielroe:
what is breaking about it? the existing slot props should still be present

Tofandel:
`error.value` is undefined, it's only accessible as `error` now

Before this error was a ref in the template and you needed to use `error.value` to clear the error

danielroe:
ah! my mistake.

I'm sorry about that! I wonder if we could add a compatibility layer that somehow defines the ref _on_ the error? 🤔

Tofandel:
I guess defining a getter setter on value as a non enumerable property if it doesn't already exist should be fine

PR #31868, opened 2025-04-21
refactor: opt in to the new unplugin filter option

### LINKED ISSUE #31837: opt in to new unplugin filter options
https://unplugin.unjs.io/guide/#filters

Saeid-Za:
### 🔗 Linked issue

resolves #31837

### 📚 Description

Adding transform [`filter`](https://rolldown.rs/guide/plugin-development#plugin-hook-filters) option based on `id` or `code` patterns.

There is a limitation in how the include option is interpreted:
> If multiple values are passed to include, the filter matches if any of them match.

This means there is no simple way to combine multiple regular expressions with an AND operator (unless we use regex lookaheads to require both patterns). In some cases, a plugin needed to match multiple combined regexes with an AND relationship. As a workaround, I previously picked the most restrictive pattern as a micro-optimization.

AFAIK, combining regexes with lookaheads would require modifying both patterns, and given the fragile nature of regexes (and my limited experience with them), I avoided that approach.


Saeid-Za:
The failing test is because of the manual plugin mocking in unit tests.

Should I mock the new filter option on these tests?

Saeid-Za:
The test `component names > should ignore files without extension` in `packages/nuxt/test/component-names.test.ts` was originally intended to verify the transformation filter logic. However, since this logic is now bypassed in the unit test and delegated to the underlying bundler, the test may no longer be necessary. I believe it could be removed, but this should be confirmed first.

danielroe:
could you look at the test failure? seems like a quick fix

Saeid-Za:
Oh, yes, I don't know if I should just remove the test, since the test is testing if a regrex filter is applied to file inputs.

This logic was previously in the `transform` function, now moved to `transform.filter` which is not being run when we run the unit test without a bundler actually being there.

I guess my question is, based on that, could we just remove the test?

danielroe:
ah, good point. we can remove that test item indeed.

danielroe (approved):
great work ❤️

PR #31880, opened 2025-04-23
test: run against actual `rolldown-vite` dep

TheAlexLichter:
This is just a PR to see [how the integration](https://github.com/nuxt/nuxt/pull/31812) runs with `rolldown-vite`!

(Might be redundant when a consistent `rolldown-vite` ecosystem CI is in place)


packages/ui-templates/test/__snapshots__/templates.spec.ts.snap
TheAlexLichter, line 1: Probably minifcation-related.

test/bundle.test.ts
```diff
@@ -38,18 +38,22 @@ describe.skipIf(process.env.SKIP_BUNDLE_SIZE === 'true' || process.env.ECOSYSTEM
...
         "_nuxt/a.js",
         "_nuxt/client-component.js",
         "_nuxt/default.js",
         "_nuxt/entry.js",
-        "_nuxt/index.js",
+        "_nuxt/nuxt.js",
+        "_nuxt/pages.js",
+        "_nuxt/runtime-core.js",
         "_nuxt/server-component.js",
+        "_nuxt/utils.js",
```
TheAlexLichter, line 56: Wondering why these files are added 🤔

TheAlexLichter:
https://github.com/vitejs/vite-ecosystem-ci/blob/34744e35900dfa30383132c2a2a23d3b3895bec7/tests-patches/nuxt/nuxt.patch#L79-L80 regarding inline styles (will investigate later on)

danielroe:
I think we can close this now.

PR #31894, opened 2025-04-24
refactor(schema): do not generate types for `ConfigSchema`

danielroe:
### 📚 Description

this moves the generated config schema directly into the source, which will allow us to version/curate it better in future, such as maintaining 'userinput' and 'normalisedconfig' options more explicitly.

we will still generate the schema.json itself of course, and I've not stripped the `@type` fields from there yet.

however, I think the schema.json is [only consumed in `nuxtr`](https://github.com/search?q=%40nuxt%2Fschema%2Fschema%2Fconfig.schema.json&type=code) so we might want to think about whether it's worth continuing to generate/ship it in v4 or whether there might be a better solution - what do you think @adhamfarrag @antfu?

### DIFF
--- a/packages/kit/src/module/define.ts
+++ b/packages/kit/src/module/define.ts
@@ -51,17 +51,14 @@ function _defineNuxtModule<
   module.meta.configKey ||= module.meta.name
 
   // Resolves module options from inline options, [configKey] in nuxt.config, defaults and schema
-  async function getOptions (
-    inlineOptions?: Partial<TOptions>,
-    nuxt: Nuxt = useNuxt(),
-  ): Promise<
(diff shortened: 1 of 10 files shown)


adhamfarrag:
Hey @danielroe, will is still possible to create an adaptor layer to strip the schema with V4?

danielroe:
@adhamfarrag it very much depends what you need! how much of the schema data do you want to consume?

adhamfarrag:
@danielroe I only rely on the schema to offer autocompletion for `.nuxtrc` files. If possible, I can just strip what's needed depending on the Nuxt version a user uses directly through the extension.

PR #22192, opened 2023-07-18
fix(nuxt): don't redirect if initial path has trailing slash

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/22182

### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Because nitro will always render routes without trailing slash, this means Nuxt will currently redirect to non-trailing slash option on initial load. This removes that behaviour to allow users to configure nginx/etc as desired.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #31902, opened 2025-04-25
fix(nuxt): don't redirect when route has trailing slash

### LINKED ISSUE #31521: [SSG] The trailing "/" is missing from the browser address bar link
### Environment

- Node Version: `20.12.2`
- Nuxt Version: `3.16.1`

### Reproduction

N/A

### Describe the bug

I'm using the SSG feature in my Nuxt project. I generate the static pages using the following commands and serve them with:
```
npx nuxi generate
npx serve ./.output/public/
```
The pages are generally accessible. However, when accessing certain pages (for example, accessing http://127.0.0.1:3000/test/), the trailing slash ("/") in the browser's address bar is sometimes removed, turning the URL into http://127.0.0.1:3000/test. This behavior doesn't occur every time; after regenerating the static pages with `npx nuxi generate` and re-accessing the pages, the trailing slash might be retained occasionally.
I want to always retain the trailing slash at the end of the URL in the browser's address bar.

### Additional context

_No response_

### Logs

```shell-script

```

kyohei-23:
### 🔗 Linked issue
resolves #31521 

### 📚 Description
The idea with #22192 was likely to make URLs the same with or without a trailing slash, but it didn't.
This is due to using `isEqual` with `trailingSlash:true` option.

As I mentioned in [my comment on the issue](https://github.com/nuxt/nuxt/issues/31521#issuecomment-2828100969), there are two ways to tackle this problem, but I figured using `isSamePath`, with its simpler arguments and clearer usage, would be better.


test/nuxt/nuxt-time.test.ts
```diff
@@ -15,6 +15,7 @@ describe('<NuxtTime>', () => {
             month: 'long',
             day: 'numeric',
             second: 'numeric',
+            timeZone: 'UTC',
```
danielroe, line 18: what was the issue this solves?

kyohei-23: That is to resolve this test failure.
https://github.com/nuxt/nuxt/actions/runs/14669031895/job/41171091112 

This may not be due to my changes, and not related with issue subject, but I thought all test passing is necessary to merge. 
I will revert this commit if it is not necessary.

danielroe: the main branch isn't failing - I would find it strange if this change had to be made in order to merge this PR.

would you revert? 🙏

kyohei-23: Sure!

The failure in test was maybe due to timezone of my region...😓

PR #31903, opened 2025-04-25
fix(nuxt): do not include components of key in `useFetch` watch sources

### LINKED ISSUE #31896: bug: <no response> Request aborted as another request to the same endpoint was initiated.
### Environment

-

### Reproduction

https://stackblitz.com/edit/github-cwy9wkvf?file=app.vue

### Describe the bug

Regression introduced in https://github.com/nuxt/nuxt/pull/31373

Before:

https://github.com/user-attachments/assets/ac917f7d-5926-4b76-a67d-c2ce410207d4

After:

https://github.com/user-attachments/assets/103ac446-2087-48b9-b402-9b117efff833

### Additional context

_No response_

### Logs

```shell-script

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31896

### 📚 Description

For some fetch options, they both affect the generation of the hashed key _and_ are manually added via `fetchOptions`.

This PR:

1. ~~removes auto-generating keys from `useFetch` (as this was never being used)~~
3. creates a subset of fetch options to be watched (as the rest are already watched via the now-reactive `key`


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -93,20 +93,18 @@ export function useFetch<
...
-  const [opts = {}, autoKey] = typeof arg1 === 'string' ? [{}, arg1] : [arg1, arg2]
+  const [opts = {}] = typeof arg1 === 'string' ? [{}, arg1] : [arg1, arg2]
 
   const _request = computed(() => toValue(request))
 
-  const _key = computed(() => toValue(opts.key) || hash([autoKey, typeof _request.value === 'string' ? _request.value : '', ...generateOptionSegments(opts)]))
-  if (!_key.value || typeof _key.value !== 'string') {
-    throw new TypeError('[nuxt] [useFetch] key must be a string: ' + _key.value)
+  const key = computed(() => toValue(opts.key) || ('$f' + hash([typeof _request.value === 'string' ? _request.value : '', ...generateOptionSegments(opts)])))
+  if (!key.value || typeof key.value !== 'string') {
```
huang-julien, line 101: is this still needed ? key should be always ` Ref<string>`

danielroe: probably not!

PR #31904, opened 2025-04-25
feat(nuxt,schema): make build cache location configurable

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt/issues/31899

### 📚 Description

by default, we use the workspace dir to store the cache of `experimental.buildCache`. This PR makes it configurable

PR #31907, opened 2025-04-26
fix(nuxt): use first existing modulesDir to store build cache files

huang-julien:
### 🔗 Linked issue
closes https://github.com/nuxt/nuxt/pull/31904 

### 📚 Description

Do we want to make it configurable ?


danielroe (approved):

danielroe:
hm. rethinking this - we use `<rootDir>/.cache` elsewhere, and that feels possibly safer - wdyt?

huang-julien:
https://github.com/nuxt/nuxt/pull/31907#issuecomment-2832539552

I think storing it in node_modules shoudl be good too. Users usually cache node_modules in CI. Do you think we could give users the possibility to customize the location of the build cache ? (https://github.com/nuxt/nuxt/pull/31904)

PR #31945, opened 2025-04-29
fix(nuxt): skip view transition if user agent is providing one before defining `transition`

### LINKED ISSUE #31928: Double View Transition in Safari
### Reproduction
Trigger a swipe back gesture in Safari with any [View Transition](https://nuxt.com/docs/getting-started/transitions#view-transitions-api-experimental):
https://jch.github.io/turbo-cancel-view-transition/

### Describe the bug

The new View Transition API became available in Safari 18. However, Safari’s default swipe gesture for back and forward navigation triggers the animation after the page swipe gesture, causing a double transition effect.

[This blog post](https://discuss.hotwired.dev/t/demo-skip-view-transition-in-favor-of-default-user-agent-animation/6043) highlights this issue and provides a workaround for Turbo.

It would be great if a similar fix could be implemented for Nuxt, by detecting swipe gestures user-agent transition and disabling the transition in such cases.

lorenzofiamingo:
### 🔗 Linked issue

Resolves #31928

https://github.com/user-attachments/assets/b7be0212-7ec8-4224-be4a-f3dffa044c26

### 📚 Description

This PR is a follow-up to #31938, where the transition is skipped when `PopStateEvent.hasUAVisualTransition` is `true` by calling `transition.skipTransition()`. However, after a page reload, it can happen that the `popstate` event of the subsequent navigation is fired before `transition` is created, so calling `transition.skipTransition()` has no effect.  
This PR covers this edge case by setting a flag and checking it when the `transition` is created.

### 📱 Recording

https://github.com/user-attachments/assets/d5342ea7-ba6e-478a-9de6-8be6add7a49d

### DIFF
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -23,7 +23,7 @@ export default defineNuxtPlugin((nuxtApp) => {
 
   window.addEventListener('popstate', (event) => {
     hasUAVisualTransition = event.hasUAVisualTransition
-    if (hasUAVisualTransition) transition?.skipTransition()
+    if (hasUAVisualTransition) { transition?.skipTransition() }
   })
 
   const router = useRouter()
--- a/packages/nuxt/src/app/plugins/view-transitions.client.ts
+++ b/packages/nuxt/src/app/plugins/view-transitions.client.ts
@@ -9,8 +9,15 @@ export default defineNuxtPlugin((nuxtApp) => {
     return
   }
 
+  let transition: undefined | ViewTransition
   let finishTransition: undefined | (() => void)
   let abortTransition: undefined | (() => void)
+
+  window.addEventListener('popstate', (event) => {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/plugins/view-transitions.client.ts
```diff
@@ -26,7 +33,12 @@ export default defineNuxtPlugin((nuxtApp) => {
     const prefersReducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)').matches
     const prefersNoTransition = prefersReducedMotion && viewTransitionMode !== 'always'
 
-    if (viewTransitionMode === false || prefersNoTransition || !isChangingPage(to, from)) {
+    if (
+      viewTransitionMode === false ||
+      prefersNoTransition ||
+      !isChangingPage(to, from) ||
+      hasUAVisualTransition
```
danielroe, line 40:
```suggestion
      hasUAVisualTransition ||
      !isChangingPage(to, from)
```

danielroe (approved):

PR #31950, opened 2025-04-29
fix(nuxt): preserve params/meta/matched when navigating with Route object

ivansky:
### 📚 Description

This change modifies the `getRouteFromPath` utility function to correctly handle cases where a `Partial<Route>` object is passed to navigation methods (`router.push`, `router.replace`, `navigateTo`).

Previously, properties like `params`, `meta`, and `matched` from the input object were discarded and overwritten with empty values in the returned route object, limiting the ability to pass rich, pre-defined route state.

With this update, `getRouteFromPath` now preserves `params`, `meta`, and `matched` if they are present in the input `Partial<Route>`.

This enhancement allows developers to pass more complete `Route` objects, ensuring custom context (especially via `meta`) and pre-calculated route information persists through the routing pipeline (middleware, navigation guards).

This opens up opportunities for more advanced routing strategies and state management. A key use case this enables is the ability to define navigation-specific `meta` context (e.g., to signal state restoration) which can be reliably accessed in middleware/guards, addressing scenarios where differentiating navigation types (forward vs. back-like) is necessary for specific page behaviors.

This is a backward-compatible change; navigating with just paths or simple location objects behaves as before.

### DIFF
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -40,7 +40,7 @@ function getRouteFromPath (fullPath: string | Partial<Route>) {
   let params: Route['params'] = {}
   let matched: Route['matched'] = []
   let meta: Route['meta'] = {}
-  
+
   if (typeof fullPath === 'object') {
     params = fullPath.params || {}
     matched = fullPath.matched || []
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -37,7 +37,14 @@ interface Route {
 }
 
 function getRouteFromPath (fullPath: string | Partial<Route>) {
+  let params: Route['params'] = {}
+  let matched: Route['matched'] = []
+  let meta: Route['meta'] = {}
+  
   if (typeof fullPath === 'object') {
+    params = fullPath.params || {}
+    matched = fullPath.matched || []
+    meta = fullPath.meta || {}
     fullPath = stringifyParsedURL({
       pathname: fullPath.path || '',
       search: stringifyQuery(fullPath.query || {}),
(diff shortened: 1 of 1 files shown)


danielroe:
would you provide a regression test, or share a reproduction of a bug or issue?

the reason I ask is that the plugin you are editing is not actually what provides routing for most nuxt applications - it's a fallback 'stub' router for when users _are not using_ the `pages/` directory or a vue-router integration.

instead, when the `pages/` directory is present, [this file](https://github.com/nuxt/nuxt/blob/6977bb3ad96ec9936579c44c9ef20c6ddcc5c65d/packages/nuxt/src/pages/runtime/plugins/router.ts) sets up the router

ivansky:
@danielroe is there a good regression test related to routing that I can use as an example?

danielroe:
@ivansky because it's just a stub for compatibility, we don't currently test this routing implementation

I can help with that, though, if can you provide a bug reproduction.

ivansky:
@danielroe will it be helpful ? https://stackblitz.com/edit/github-gy8ea2t4?file=middleware%2Ftype-set.global.ts expectatio is that the nav-type is set, because we set meta property and read it in either middleware or regular router hook.

danielroe:
@ivansky so you are not using a `pages/` directory in your project?

ivansky:
@danielroe we are not using `/pages` because it does not fit our SEO urls requirements, to know what exact page to draw depending on URL we need to make a request, in normal `/pages` approach you need to know all routes before hand, the list of different routes we could have is not definitive. I wish there is async function to match specific page. Thats why we are using `app.vue` to draw specific  page component after URL is processed by our backend service.

danielroe:
Ahhh, I see. Forgive me for doubting you. This looks like a good fix then.

danielroe (approved):

PR #31967, opened 2025-05-01
fix(nuxt): short circuit middleware when validate returns false

### LINKED ISSUE #29250: Route Page validate not working on SSR
### Environment

Nuxt project info:                                                                                                                                                                                                              11:36:52 AM

- Operating System: Linux
- Node Version:     v20.9.0
- Nuxt Version:     3.13.0
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://codesandbox.io/p/devbox/wandering-field-rjwg87
open `/posts/nan-id`  with ssr 

expected: Page not exist. The middleware is not called 

### Describe the bug

In SSR mode, validate does not work properly. And after it return false, the middleware still called. In spa mode, nothing will happen if you try to go to an invalid page.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/29250

### 📚 Description

this updates middleware behaviour to short circuit when `validate` returns false

I'd like to confirm this isn't a breaking change but wasn't able to replicate 'falling through' to other routes if validate returned `false`

### DIFF
--- a/packages/nuxt/src/pages/runtime/validate.ts
+++ b/packages/nuxt/src/pages/runtime/validate.ts
@@ -1,12 +1,11 @@
 import { createError, showError } from '#app/composables/error'
 import { useNuxtApp } from '#app/nuxt'
-import { defineNuxtRouteMiddleware, useRouter } from '#app/composables/router'
+import { defineNuxtRouteMiddleware } from '#app/composables/router'
 
-export default defineNuxtRouteMiddleware(async (to) => {
+export default defineNuxtRouteMiddleware(async (to, from) => {
   if (!to.meta?.validate) { return }
 
   const nuxtApp = useNuxtApp()
-  const router = useRouter()
 
   const result = await Promise.resolve(to.meta.validate(to))
   if (result === true) {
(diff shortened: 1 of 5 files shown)


huang-julien (approved):

packages/nuxt/src/pages/runtime/validate.ts
```diff
@@ -1,41 +1,29 @@
...
-      // We stop the navigation immediately before it resolves
-      // if there is no other route matching it.
-      return false
-    }
-  })
+
+  // We pretend to have navigated to the invalid route so
+  // that the user can return to the previous page with
+  // the back button.
+  if (typeof window !== 'undefined') {
```
Mini-ghost, line 24: Would there be any difference between using `import.meta.client` here versus `typeof window !== 'undefined'`?

danielroe: not sure where that line originally came from, but `import.meta.client` should be just as safe 👍  (they are both tree-shakeable)

PR #31968, opened 2025-05-01
feat(nuxt): add ssr debug prompt with link

CodeMan62:
### 🔗 Linked issue

resolves #21633 

### 📚 Description
In this PR I am trying to add an ssr debug prompt whenever an error is encountered rendering a page with vite-node or nitro, so it will show a prompt in console or a special link that allows opening the page in browser with ssr: false (dev mode only) for debugging


danielroe:
This is looking great. I think I'm going to merge this into some work I plan to do to display errors on the client side with Youch, so I'll complete that first and then merge this PR (in case you're wondering on the status).

harlan-zw:
It could be more integrated into existing error flows more but it does  solve the problem, so I'd happily see it merged, nice work :+1:

PR #31980, opened 2025-05-02
feat(nuxt): ignore files & folders prefixed with an underscore

### LINKED ISSUE #31979: feat(pages): Private Folders Prefixed with an Underscore (`_[folderName]`)
### Describe the feature

Similar to [private folders in Next.js (App Router)](https://nextjs.org/docs/app/getting-started/project-structure#private-folders), prefixing folders inside the `pages` directory with an underscore (`_`) should opt them (and all their child segments) out of routing—visiting them from the browsers shows the 404 page.

A scenario in which this comes pretty handy is when you want to do FDD (Feature-Driven Development) right from the `pages` directory, co-locating all routes, components, composables, utils, etc. of a feature in a single folder.

For example, given the following folder strcuture
```md
pages
└── profile
    ├── _components
    │   ├── ChangePassword.vue
    │   └── UserPreferences.vue
    ├── _composables
    │   └── useAccountForm.ts
    ├── _utils
    │   ├── parseAccountData.ts
    │   └── prepareAccountPayload.ts
    ├── _schema
    │   └── account.schema.ts
    │
    ├── account.vue
    └── index.vue
```
The only generated route should be `/profile` and `/profile/account`.

<br />

I implemented this feature as a module by leveraging the `pages:extend` Nuxt hook (_modules/privateRoutes.ts_):
```ts
import type { NuxtPage } from '@nuxt/schema';

import { defineNuxtModule } from 'nuxt/kit';

  hooks: {
    'pages:extend': removePrivateRoutes,
  },
});

function removePrivateRoutes(routes: NuxtPage[]): void {
  for (const route of routes) {
    if (isRoutePrivate(route.path)) {
      const routeIndex = routes.indexOf(route);
      routes.splice(routeIndex, 1);
    }
  }
}

function isRoutePrivate(path: string): boolean {
  return path.includes('/_');
}
```

### Additional information

- [x] Could this feature be implemented as a module?

### Final checks

ShayanTheNerd:
### 🔗 Linked issue

Resolves #31979

### 📚 Description

Details and use case are outlined in the linked issue.

### DIFF
--- a/docs/2.guide/2.directory-structure/1.pages.md
+++ b/docs/2.guide/2.directory-structure/1.pages.md
@@ -15,6 +15,10 @@ Pages are Vue components and can have any [valid extension](/docs/api/configurat
 
 Nuxt will automatically create a route for every page in your `~/pages/` directory.
 
+::important
+Nuxt ignores files and directories prefixed with an underscore (`_`) and will not generate routes for them. This is useful for components, composables, utilities, or other files you want to co-locate within the `pages` directory without them becoming routes.
+::
+
 ::code-group
 
 ```vue [pages/index.vue]
@@ -40,6 +44,18 @@ export default defineComponent({
     return <h1>Index page</h1>
   }
 })
+```
+
+```vue [pages/profile/_components/Card.vue]
+<!-- Nuxt will not generate a route for this comopnent. -->
+<template>
+  <div>Profile Card</div>
+</template>
+```
+
+```ts twoslash [pages/_utils/helper.ts]
+// Nuxt will not generate a route for this utility.
+export const helper = () => { /* ... */ }
 ```
 
 ::
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -102,6 +102,11 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
       .replace(new RegExp(`${escapeRE(extname(file.relativePath))}$`), '')
       .split('/')
 
+    // Ignore files and directories starting with an underscore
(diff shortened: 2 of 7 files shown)


danielroe:
I believe this feature already exists see: https://github.com/nuxt/nuxt/issues/31979#issuecomment-2847606867

ShayanTheNerd:
> I believe this feature already exists - see: [#31979 (comment)](https://github.com/nuxt/nuxt/issues/31979#issuecomment-2847606867)

@danielroe, the behavior is configurable using the code you provided in the aforementioned issue. However, I still think this should be built-in because it's useful and is based on a well-known convention in the community.
In addition, people coming from a Next.js background will probably be looking for things like this. So, as I see it, having such features makes the framework look more mature and featureful.

Apart from that, I recommend adding the `pages.pattern` configuration option as a side note directly in [Pages • Nuxt Directory Structure](https://nuxt.com/docs/guide/directory-structure/pages).

danielroe:
I think it might make sense to add the `nuxt.options.ignore` patterns to pages.pattern by default (and `_` is ignored by default).

And definitely worth documenting in the meantime.

danielroe:
checking, we had a regression with user-provided ignore patterns (resolved in https://github.com/nuxt/nuxt/pull/32020).

Nuxt also ignores files beginning with `-` by default, which I think is what you want.

To customise this to an underscore you can set `ignorePrefix: '_'`.

I think that should resolve the issue 🙏

danielroe:
... and just to add - thank you so much for the PR!

if this wasn't a feature we already had I would have been delighted to merge it. (and I would still welcome a PR improving the docs, or calling out this feature if it isn't already there)

❤️

PR #31982, opened 2025-05-02
feat(nuxt): add View Transitions Types support

### LINKED ISSUE #31972: Add types support to View Transitions
### Describe the feature

It would be nice to introduce the possibility to define support for view transition types in the experimental view transitions, as described in the [Google View Transition Update](https://developer.chrome.com/blog/view-transitions-update-io24?hl=it#view-transition-types).  
I suggest introducing `ViewTransitionOptions` in `nuxt.config.ts` to allow defining a callback for the types:

```ts
experimental: {
  viewTransition: {
    enabled: true,
    defineTypes: (from, to) => [`${from.name?.toString()}-to-${to.name?.toString()}`],
  },
},
```

This way, you can define animations based on `:active-view-transition-type()`. In this way you could be able to define a slide animations or, more generally, asymmetric animations:

```ts
:active-view-transition-type(store-to-store-slug) {
  &::view-transition-old(selected-image) {
    // ...
  }

  &::view-transition-new(selected-image) {
    // ...
  }
}

:active-view-transition-type(store-slug-to-store) {
  &::view-transition-old(selected-image) {
    // ...
  }

  &::view-transition-new(selected-image) {
    // ...
  }
}
```

To implement this, it should be enough to define `ViewTransitionOptions` and add the following to `view-transitions.client.ts`:

```ts
transition = document.startViewTransition!({
  // @ts-expect-error update is not in the type
  update: async () => {
    changeRoute()
    return promise
  },
  types: appViewTransitionTypes(from, to),
})
```

P.S. maybe could be useful this as default type for the transitions:
```ts
`${from.name?.toString().replace(/___[^_]*$/, '')}-to-${to.name?.toString().replace(/___[^_]*$/, '')}`
```
where `.replace(/___[^_]*$/, '')}` is to handle localized routes.

### Additional information

### Final checks

lorenzofiamingo:
### 🔗 Linked issue

Resolves #31972

### 📚 Description

This PR introduce the possibility to define support for view transition types in the experimental view transitions, as described in the [Google View Transition Update](https://developer.chrome.com/blog/view-transitions-update-io24?hl=it#view-transition-types).

https://github.com/nuxt/nuxt/issues/31972#issuecomment-2846888626
> Normally we try to avoid passing functions from nuxt config -> runtime code.
> 
> Maybe instead we could have a hook that allows users to configure this at runtime instead? For example, we currently have this one that allows access to the `transition`:
> 
> [nuxt/packages/nuxt/src/app/plugins/view-transitions.client.ts](https://github.com/nuxt/nuxt/blob/b74f2a64efc8376d8a41392f57c8248395049817/packages/nuxt/src/app/plugins/view-transitions.client.ts#L60)
> 
> Line 60 in [b74f2a6](/nuxt/nuxt/commit/b74f2a64efc8376d8a41392f57c8248395049817)
> 
>  await nuxtApp.callHook('page:view-transition:start', transition)

I haven’t implemented the hook approach.  If feasible, we should definitely move in that direction to better align with Nuxt’s conventions and avoid passing functions from config to runtime.


danielroe (approved):
goodness, I'm sorry this took so long! but thank you - this is great ❤️

PR #31987, opened 2025-05-03
fix(nuxt): ensure `useAsyncData` reactive to `key` changes when `immediate: false`

### LINKED ISSUE #31976: URL is not watch in useFetch when `immediate` is set to `false`
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     3.17.1
- CLI Version:      3.25.0
- Nitro Version:    2.11.11
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/nuxt-starter-4ptgsiht?file=app.vue

### Describe the bug

Hi I noticed this little bug,

If you set params `immediate` to `false` with useFetch composable, URL is not watch correctly (a new fetch is not perform when url change).

I don't know if it's due to new Data Fetching system introduce with Nuxt 3.17.

Thx in advance if you can do something about that ! 😄

#### Workaround
You can add URL to watch params like that:

```JS

const url = computed(() => ...)  // your computed url
const { data: results } = useFetch(url, {
  immediate: false,
  watch: [url]
});
```

### Additional context

_No response_

### Logs

```shell-script

```

Mini-ghost:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/31976

### 📚 Description

This PR addresses an issue where `useAsyncData` does not react to changes in the `key` parameter when `immediate: false`. As a result, data is not re-fetched upon key changes, leading to potential inconsistencies.

Thank you very much for your time and review! If there’s anything I might have overlooked, please feel free to let me know.

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -362,15 +362,11 @@ export function useAsyncData<
       if (oldKey) {
         unregister(oldKey)
       }
-
-      let nextAsyncData = nuxtApp._asyncData[newKey]
-
-      if (!nextAsyncData?._init) {
-        nuxtApp._asyncData[newKey] = nextAsyncData = createAsyncData(nuxtApp, newKey, _handler, options, options.getCachedData!(newKey, nuxtApp, { cause: 'initial' }))
+      if (!nuxtApp._asyncData[newKey]?._init) {
+        nuxtApp._asyncData[newKey] = createAsyncData(nuxtApp, newKey, _handler, options, options.getCachedData!(newKey, nuxtApp, { cause: 'initial' }))
       }
-
-      nextAsyncData._deps++
-      nextAsyncData.execute({ cause: 'initial', dedupe: options.dedupe })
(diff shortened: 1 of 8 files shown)


danielroe (commented):
I think we should probably only immediately execute asyncData when the first `execute` had already been called.

for useFetch, we need to execute regardless, because that was the behaviour previously. we can make the behaviour the same for v4, so we should export a flag for now to equalise the data.

we can use the `.immediate` trick in fetch.ts only when the flag is not defined

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -359,15 +359,16 @@ export function useAsyncData<
       }
     }
     const unsubKey = watch(key, (newKey, oldKey) => {
+      const hasRun = nuxtApp._asyncData[oldKey]?.data.value !== asyncDataDefaults.value
```
danielroe, line 362:
```suggestion
      const hasRun = oldKey && nuxtApp._asyncData[oldKey]?.data.value !== asyncDataDefaults.value
```

PR #32013, opened 2025-05-06
fix(nuxt): handle File within `FormData`

### LINKED ISSUE #32000: useFetch FormData handling crashes v3.17.2
We are stuck on 3.16.2 and cannot upgrade to the latest version because of this issue (and #32006) and we really need to upgrade for there are some other fixes in the latest version that we can benefit from. 

The same as #31962 

Reproduction

https://github.com/raminjafary/nuxt-upload-formdata-repro

The error is `Cannot serialize File`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32000

### 📚 Description

this adds additional handling for `File` within `FormData`

not sure if we should implement upstream instead in `ohash` (cc: @pi0)


pi0 (approved):
Looks good to me. We had a discussion before about adding customizable serializer in ohash but it has not been moved forward yet. (Supporting DOM APIs can add a lot to the baseline bundle)

PR #32045, opened 2025-05-08
refactor(nuxt,schema): use oxc for `onPrehydrate` transform

TheAlexLichter:
### 📚 Description

Simple refactor to replace esbuild with oxc for transforming onPrehydrate


packages/schema/package.json
```diff
@@ -79,6 +79,7 @@
     "@vue/shared": "^3.5.13",
     "consola": "^3.4.2",
     "defu": "^6.1.4",
+    "oxc-transform": "^0.68.1",
```
danielroe, line 82: this should be a dev-dependency

packages/schema/src/types/schema.ts
```diff
@@ -1867,6 +1867,12 @@ export interface ConfigSchema {
     options: import('esbuild').TransformOptions
   }
 
+  oxc: {
+    transform: {
+      options: import('oxc-transform').TransformOptions
```
danielroe, line 1872: here, and above in `esbuild`, we could probably directly import the options

danielroe (commented):
we might also be able to replace the usage of esbuild in `packages/vite/src/plugins/analyze.ts`

TheAlexLichter:
@danielroe will check!

PR #32049, opened 2025-05-08
chore(nuxt): lint with oxlint besides eslint

TheAlexLichter:
### 📚 Description

Runs `oxlint` besides `eslint` for all rules that are available in the rust-based linter already.

I've had to disable three rulesets (see 7f730389417fcf8a1a1c5b5c38ab9228d9cfadac) as they were not behaving as before.

Related issues: https://github.com/oxc-project/oxc/issues/10912 https://github.com/oxc-project/oxc/issues/10911 https://github.com/oxc-project/oxc/issues/10910


packages/vite/src/runtime/vite-node.mjs
```diff
@@ -11,7 +11,7 @@ const runner = createRunner()
 /** @type {(ssrContext: import('#app').NuxtSSRContext) => Promise<any>} */
 let render
 
-/** @param ssrContext {import('#app').NuxtSSRContext} */
```
TheAlexLichter, line 14: We had a few JSdocs wrong here actually!

.oxlintrc.json
TheAlexLichter, line 1: This is a pretty big config file. It was defined based on our eslint config (via oxlint's migration script). The most part is related to globals though

.oxlintrc.json
```diff
@@ -0,0 +1,329 @@
...
+    "onUnmounted": "readonly",
+    "reactive": "readonly",
+    "ref": "readonly",
+    "shallowReactive": "readonly",
+    "shallowRef": "readonly",
+    "toRef": "readonly",
+    "toRefs": "readonly",
+    "watch": "readonly",
+    "watchEffect": "readonly"
+  },
```
danielroe, line 103: is there any way to get oxlint to be aware of this from the TS environment?

.oxlintrc.json
```diff
@@ -0,0 +1,329 @@
...
+    "**/test-results/",
+    "**/playwright-report",
+    "**/temp",
+    "**/public"
+  ],
+  "globals": {
+    "$fetch": "readonly",
+    "NodeJS": "readonly",
+    "__dirname": "readonly",
+    "__filename": "readonly",
```
danielroe, line 84: I don't think we use these anywhere in our source - maybe just in jiti stubs?

danielroe (commented):
interestingly this increases the time it takes to run `pnpm eslint . --cache`, which increases from 1.6s on my computer to 1.8s with this PR ...

is it feasible at some point to fully replace eslint, including any type aware rules, or is this additive only? are there benefits to using oxlint in addition to eslint?

.oxlintrc.json
TheAlexLichter: Most likely. We can also remove it. The values were set through the migrate script based on the existing eslint config 😋

.oxlintrc.json
TheAlexLichter: Will check

TheAlexLichter:
> interestingly this increases the time it takes to run `pnpm eslint . --cache`, which increases from 1.6s on my computer to 1.8s with this PR ...

That's weird, it should not 🤔 I'll do some more measurements and check.
 
> is it feasible at some point to fully replace eslint, including any type aware rules, or is this additive only? are there benefits to using oxlint in addition to eslint?

That's the goal of `oxlint` eventually, yes. Type aware linting is still wip though (e.g. through a [tsslint](https://github.com/johnsoncodehk/tsslint) integration or other ways), as well as further rules, especially those that aren't focusing JavaScript (so all vue-related rules).

.oxlintrc.json
danielroe: ahhh, makes sense.

TheAlexLichter:
Scenario | Avg Time (s) | vs ESLint only (main)
-- | -- | --
ESLint only (main) | 4.638 | –
ESLint only (this branch) | 4.721 | +1.8% (~+0.08s)
ESLint + Oxlint (this branch) | 5.869 | +26.6% (~+1.23s)

This is strange. I'll double check the results on a non-WSL machine. Right now, my suspicion is:

* There are still ESLint rules that have to visit most files, meaning that ESLint time doesn't go down (but I wonder that it didn't even move down a tiny bit)
* Due to I/O and/or cores, parallelization didn't work too well (as Oxlint was taking ~1-1.4s, so that'd explain the perf slowdown). This might be my test setup though

<details><summary>See the full benchmark on my WSL machine</summary>

### Testing env

Measured on my WSL instance. YMMV depending on the system.

### ESLint only (main branch, 0b55bf729e6555e408ae5f50efd42762b093bee3)

#### Hyperfine

`hyperfine --warmup 3 'pnpm run lint'`

Benchmark 1: pnpm run lint
  Time (mean ± σ):      4.638 s ±  0.059 s    [User: 4.804 s, System: 0.695 s]
  Range (min … max):    4.579 s …  4.740 s    10 runs

#### Timing for ESLint

`TIMING=1 pnpm lint`

Rule                              | Time (ms) | Relative
:---------------------------------|----------:|--------:
@stylistic/indent                 |  1173.991 |     7.4%
@typescript-eslint/no-unused-vars |  1116.320 |     7.0%
import/no-duplicates              |   829.774 |     5.2%
import/no-restricted-paths        |   663.368 |     4.2%
regexp/confusing-quantifier       |   630.324 |     4.0%
jsdoc/check-access                |   604.903 |     3.8%
vue/no-child-content              |   499.101 |     3.1%
import/order                      |   434.692 |     2.7%
import/no-self-import             |   362.659 |     2.3%
jsdoc/require-property-name       |   323.619 |     2.0%

### With Oxlint & ESLint in parallel (this branch, f726783)

#### Hyperfine

`hyperfine --warmup 3 'pnpm run lint'`

Benchmark 1: pnpm run lint
  Time (mean ± σ):      5.869 s ±  0.189 s    [User: 7.722 s, System: 1.120 s]
  Range (min … max):    5.618 s …  6.206 s    10 runs

#### Timing for ESLint

`TIMING=1 pnpm eslint .`

Rule                              | Time (ms) | Relative
:---------------------------------|----------:|--------:
@typescript-eslint/no-unused-vars |   983.146 |    10.2%
@stylistic/indent                 |   887.404 |     9.2%
import/order                      |   809.769 |     8.4%
import/no-restricted-paths        |   632.879 |     6.6%
regexp/confusing-quantifier       |   541.757 |     5.6%
jsdoc/check-types                 |   488.335 |     5.1%
vue/no-child-content              |   337.990 |     3.5%
jsdoc/no-multi-asterisks          |   223.100 |     2.3%
jsdoc/check-tag-names             |   222.450 |     2.3%
@stylistic/no-extra-parens        |   206.921 |     2.2%

### ESLint only, without the Oxlint timing (this branch, f726783)

#### Hyperfine

`hyperfine --warmup 3 'pnpm run lint:es'`

Benchmark 1: pnpm run lint:es
  Time (mean ± σ):      4.721 s ±  0.340 s    [User: 4.907 s, System: 0.633 s]
  Range (min … max):    4.330 s …  5.266 s    10 runs

</details>

Sysix:
> interestingly this increases the time it takes to run pnpm eslint . --cache, which increases from 1.6s on my computer to 1.8s with this PR ...

The `oxlint.buildFromOxlintConfigFile()` will read a file from system, parse the JSON and transform the config to ESLint. 
I guess the problem is there. Happy to help and improve the `eslint-plugin-oxlint` integration.

danielroe:
We've discussed this as a team (with alex too!) and in general our take-away (for the benefit of anyone else who might look at this) is:

- stylistic lint plugins are important to us.
- we don't like the prettier defaults as we feel they make the code less human-readable (for the benefit of speed), and currently it seems it's not possible to configure oxlint to perform formatting, nor is there any replacement in the oxc ecosystem. 
   - **further reading**: https://eslint.style/guide/why#linters-vs-formatters and https://antfu.me/posts/why-not-prettier
- running a two-stage process is not necessarily worth it. we'd rather have a single command that lets us lint (including stylistically lint) our code.

I hope that in future it will be possible to adopt oxlint when/if it is a drop-in replacement for our current rules (including stylistic/formatting rules).

PS. I'll cherry-pick out the fixes that oxlint picked up 🙏

.oxlintrc.json
tobiasdiez: @TheAlexLichter did you ever found a solution for this?

PR #32052, opened 2025-05-08
perf(nuxt): pre-calculate extension glob before app resolution

moshetanzer:


packages/nuxt/test/app.test.ts
```diff
@@ -94,27 +94,8 @@ describe('resolveApp', () => {
...
-          "file": "<rootDir>/layouts/default/index.vue",
-          "name": "default",
-        },
-        "other": {
-          "file": "<rootDir>/layouts/other.vue",
-          "name": "other",
-        },
-      }
-    `)
+    expect(app.middleware.filter(m => m.path.startsWith('<rootDir>'))).toMatchInlineSnapshot(`[]`)
```
TheAlexLichter, line 97: Looks like it isn't working at the moment (snapshots should stay the same)

danielroe (commented):
thank you ❤️ 

I don't think this improves performance, except perhaps by sharing the `extGlob` and the micro-optimisation of reducing property access.

(In fact it decreases performance by converting a sync operation to async (in L151), and eagerly running it, when before it was only executed if `app.mainComponent` was unset)

danielroe (approved):

PR #32054, opened 2025-05-09
docs: update `useNuxtData` default return to `undefined`

xjccc:
### 📚 Description

1. Correct type

https://github.com/nuxt/nuxt/blob/76b300eddae6d755948239c305bfcc587dd74197/packages/nuxt/src/app/composables/asyncData.ts#L488

### DIFF
--- a/docs/3.api/2.composables/use-nuxt-data.md
+++ b/docs/3.api/2.composables/use-nuxt-data.md
@@ -108,5 +108,5 @@ async function addTodo () {
 ## Type
 
 ```ts
-useNuxtData<DataT = any> (key: string): { data: Ref<DataT | null> }
+useNuxtData<DataT = any> (key: string): { data: Ref<DataT | undefined> }
 ```


docs/3.api/2.composables/use-nuxt-data.md
```diff
@@ -108,5 +108,5 @@ async function addTodo () {
 ## Type
 
 ```ts
-useNuxtData<DataT = any> (key: string): { data: Ref<DataT | null> }
+useNuxtData<DataT = any> (key: string): { data: Ref<DataT | undefined> }
```
danielroe, line 111: this differs between `3.x` + `main` branches, I think ...

Would you split this into two PRs, so I can cherry pick the removal of the unneeded JSdoc comments to 3.x after merging?

xjccc: Yes. Thank you for your feedback❤

danielroe (approved):

PR #32064, opened 2025-05-10
fix(nuxt): do not suppress chunk import error

wattanx:
### 📚 Description
This PR does not completely handle  https://github.com/nuxt/nuxt/issues/29624 .
`Error: undefined is not an object (evaluating 'A.default').` occurs because the chunk error is `preventDefault` in `app:chunkError`, allowing execution to continue.
With this PR, the error message will change to `Failed to fetch dynamically imported module:`.

Additionally, the error `Couldn't resolve component "default" at "xxx"` will also change to `Failed to fetch dynamically imported module:`

With `emitRouteChunkError: "automatic-immediate"`, I confirmed that the page reloads even when the error occurs.
https://github.com/wattanx/middleware-chunk-error/


wattanx:
@danielroe 
I think this makes the error message clearer than `Error: undefined is not an object (evaluating 'A.default')`. What do you think?

danielroe (approved):

PR #32077, opened 2025-05-12
fix(nuxt): asyncData default would be same as createAsyncData default value when clearNuxtData

xjccc:
### 📚 Description

https://github.com/nuxt/nuxt/blob/23f57e9ff58815729f84e3f32d0613ce00281548/packages/nuxt/src/app/composables/asyncData.ts#L563

When use `clearNuxtData`, I think it would be same as `createAsyncData().data` with the default `cache` or `default`.

Is it right? Thank you ❤

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -703,7 +703,7 @@ function createAsyncData<
       return nuxtApp._asyncDataPromises[key]!
     },
     _execute: debounce((...args) => asyncData.execute(...args), 0, { leading: true }),
-    _default: options.default!,
+    _default: () => hasCachedData ? initialCachedData : options.default!(),
     _deps: 0,
     _init: true,
     _hash: isDev ? createHash(_handler, options) : undefined,


danielroe:
could you share a reproduction or some code that isn't working at the moment?

xjccc:
> could you share a reproduction or some code that isn't working at the moment?

Oh.. I don't have the scene. Just see that at `clearNuxtData` to reset `data.value` to the `options.default`, and think if this will be same with the `data.value` when provide the cache?

xjccc:
I think maybe like this?

```
  experimental: {
    defaults: {
      useAsyncData: {
        value: { message: 'hello nuxt' },
      },
    },
  },
```

```ts
<script setup lang="ts">
const { data } = useAsyncData('nuxt-playground', () => {
  return new Promise((resolve) => {
    setTimeout(() => {
      resolve({ message: 'Hello from Nuxt Playground!' })
    }, 1000)
  })
}, {
  getCachedData: () => {
    return { message: 'hello' }
  },
})
await clearNuxtData('nuxt-playground')
console.log(data.value) // must be { message: 'hello' } not { message: 'hello nuxt' }
</script>
```

PR #32081, opened 2025-05-12
test: add reproduction watch params

OrbisK:


danielroe:
this should have been solved in https://github.com/nuxt/nuxt/pull/32019.

OrbisK:
> this should have been solved in #32019.

you are right. did not saw that 🫠

PR #32094, opened 2025-05-13
docs: document NuxtLink query param option

Buzut:
### 📚 Description

In addition to url parameters, NuxtLink can provide query parameters. Although very useful, it wasn't documented so, hence this PR.


danielroe (commented):
would it be better to link to the vue router documentation instead? or maybe just add a line commenting that you can also pass `query`, `hash`, `path`, `name` in this object? I'm not sure we need an example.

Buzut:
Makes sense, documenting `query` without mentioning the others doesn't feel right and we're not gonna add an exemple for each of these. 

I think a simple mention is the perfect balance.

PR #32096, opened 2025-05-13
fix(nuxt): clear async data after a tick

### LINKED ISSUE #32093: Broken transition if component has useAsyncData / useFetch
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     3.17.3
- CLI Version:      3.25.1
- Nitro Version:    2.11.11
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

❌ `nuxt@3.17.3`: https://stackblitz.com/edit/nuxt-starter-sngxdoeg?file=app.vue
✅ `nuxt@3.16.2`: https://stackblitz.com/edit/nuxt-starter-tymfyu7d?file=app.vue

### Describe the bug

If the component uses `useAsyncData` or `useFetch`, the `<transition>` only works on the first render and freezes on next updates.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32093

### 📚 Description

when clearing nuxt data _inside_ a transition this can break the transition totally (likely a vue upstream bug in here that would be worth investigating and reporting)

a workaround is to defer this until the next tick

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -1,4 +1,4 @@
-import { computed, getCurrentInstance, getCurrentScope, inject, isShallow, onBeforeMount, onScopeDispose, onServerPrefetch, onUnmounted, ref, shallowRef, toRef, toValue, unref, watch } from 'vue'
+import { computed, getCurrentInstance, getCurrentScope, inject, isShallow, nextTick, onBeforeMount, onScopeDispose, onServerPrefetch, onUnmounted, ref, shallowRef, toRef, toValue, unref, watch } from 'vue'
 import type { MaybeRefOrGetter, MultiWatchSources, Ref } from 'vue'
 import { captureStackTrace } from 'errx'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -712,8 +712,12 @@ function createAsyncData<
       asyncData._init = false
       // TODO: disable in v4 in favour of custom caching strategies
       if (purgeCachedData && !hasCustomGetCachedData) {
-        clearNuxtDataByKey(nuxtApp, key)
-        asyncData.execute = () => Promise.resolve()
+        nextTick(() => {
+          if (!asyncData._init) {
```
balek, line 716: Hi @danielroe! I guess it was meant to be:
`if (!nuxtApp._asyncData[key]?._init) {`

I'm not sure, but I don't see any place where _init could be switched from `false` to `true`. And it seems like there can be a bug if an asyncData is going to be cleared and a new asyncData with the same key is created at this moment.

PR #32100, opened 2025-05-13
fix(nuxt): clear previous head in island-renderer

Norbiros:
### 📚 Description

This bug has been mentioned multiple times, such as in [this comment](https://github.com/nuxt/nuxt/issues/19772#issuecomment-2776775639). Essentially, the logic for clearing entries in head used a deprecated method that didn't function properly.

As a result, every server component in project that uses @nuxt/ui includes the entire css variables declaration, significantly impacting performance. The difference is striking: current code, the head size is 7778 bytes; after my changes, it's only 394 bytes—nearly a 20x difference, for a very simple markdown component.

### DIFF
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -1023,7 +1023,7 @@ describe('head tags', () => {
     // should render charset by default
     expect(indexHtml).toContain('<meta charset="utf-8">')
     // should render <Head> components
-    expect(indexHtml).toContain('<title>Basic fixture - Fixture</title>')
+    expect(indexHtml).toContain('<title>Basic fixture</title>')
   })
 
   it('SSR script setup should render tags', async () => {
@@ -2045,6 +2045,7 @@ describe('server components/islands', () => {
     const html = await $fetch<string>('/server-page')
     // test island head
     expect(html).toContain('<meta name="author" content="Nuxt">')
+    expect(html).toContain('plugin-style')
   })
 
   it('/server-page - client side navigation', async () => {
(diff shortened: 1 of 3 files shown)


danielroe (approved):

test/basic.test.ts
```diff
@@ -1023,7 +1023,7 @@ describe('head tags', () => {
     // should render charset by default
     expect(indexHtml).toContain('<meta charset="utf-8">')
     // should render <Head> components
-    expect(indexHtml).toContain('<title>Basic fixture - Fixture</title>')
+    expect(indexHtml).toContain('<title>Basic fixture</title>')
```
danielroe, line 1026: this seems like a regression

Norbiros: It was added in https://github.com/nuxt/nuxt/pull/31169 and I believe it was a regression.

On production my website it works properly:
image

danielroe: the title template for the index page is set here:

https://github.com/nuxt/nuxt/blob/f209158352b09d1986aa320e29ff36353b91c358/test/fixtures/basic/plugins/my-plugin.ts#L2-L4

Shouldn't it be respected?

Norbiros: Interesting... So it was a bug before updating to unhead v2? `head.vue` has specified:
```ts
      titleTemplate: title => `${title} - Title Template Fn Change`,
```
So plugin should override it?

danielroe: the title is an input to -> title template, so I don't see this conflicting.

Norbiros: @danielroe It's overridden here:

https://github.com/nuxt/nuxt/blob/f209158352b09d1986aa320e29ff36353b91c358/test/fixtures/basic/pages/index.vue#L118-L121

danielroe: 😱  you are quite right.

thank you for the spot...

danielroe (approved):

huang-julien (approved):

PR #32103, opened 2025-05-14
fix(schema): change `pendingWhenIdle` to opt-in

wattanx:
### 📚 Description
`pendingWhenIdle` was introduced in https://github.com/nuxt/nuxt/pull/25864, but because it is currently opt-out, it breaks existing applications.
`pendingWhenIdle` should only be enabled when the Nuxt 4.

### DIFF
--- a/packages/schema/src/config/experimental.ts
+++ b/packages/schema/src/config/experimental.ts
@@ -637,7 +637,7 @@ export default defineResolvers({
      */
     pendingWhenIdle: {
       $resolve: async (val, get) => {
-        return typeof val === 'boolean' ? val : (await get('future')).compatibilityVersion !== 4
+        return typeof val === 'boolean' ? val : (await get('future')).compatibilityVersion === 4
       },
     },
   },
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/experimental.ts
```diff
@@ -637,7 +637,7 @@ export default defineResolvers({
      */
     pendingWhenIdle: {
       $resolve: async (val, get) => {
-        return typeof val === 'boolean' ? val : (await get('future')).compatibilityVersion !== 4
+        return typeof val === 'boolean' ? val : (await get('future')).compatibilityVersion === 4
```
xjccc, line 640: Emm.. It's always `true` is better?

```
return typeof val === 'boolean' ? val : true
```

wattanx: If it is always set to true, existing applications will break.
Other features are opt-in so they don’t affect existing applications, so I think this feature should be opt-in as well.

wattanx:
Sorry, I misunderstood this.
Setting it to true retains the previous behavior.

PR #32112, opened 2025-05-14
perf(nuxt): dont create shallow copy in `createNuxt` if same as prev

moshetanzer:

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -98,10 +98,11 @@ export function createNuxt (options: NuxtOptions): Nuxt {
             return
           }
           let value = applyData?.args ?? newValue
+
           // Make a shallow copy of the value
-          if (Array.isArray(value)) {
+          if (Array.isArray(value) && !Object.is(value, previousValue)) {
             value = [...value]
-          } else if (typeof value === 'object') {
+          } else if (typeof value === 'object' && !Object.is(value, previousValue)) {
             value = { ...(value as any) }
           }
           nuxt._debug!.moduleMutationRecords!.push({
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -98,10 +98,11 @@ export function createNuxt (options: NuxtOptions): Nuxt {
             return
           }
           let value = applyData?.args ?? newValue
+
           // Make a shallow copy of the value
-          if (Array.isArray(value)) {
+          if (Array.isArray(value) && !Object.is(value, previousValue)) {
```
danielroe, line 103: isn't this covered by the L97 - `if (newValue === previousValue)`?

in what situations will this trigger?

moshetanzer: Hey,

Looking at it again not 100% sure it is nessacry but it will trigger in cases where we have different array objects with the same content when apply data exists.

danielroe: There is no difference in checking Object.is and === in this case - they are equivalent for arrays and objects

so the only scenario is based on  applyData, and I'm not sure when that might realistically occur, so I was wondering if you encountered this or did any testing

danielroe: oh, I misread it - newValue is compared with previous Value and then value, so this comparison hasn't already been done 

but still, I would like to know how to test this

moshetanzer: Am driving and on phone so haven't been able to test byt pretty sure something like this should pass the first and fail the second
```

let newValue = [1, 2, 3]
let previousValue = [1, 2, 3]
let applyData = { args: [1, 2, 3] }

```

danielroe:
with such small sets, I don't think it is a meaningful difference.

PR #32115, opened 2025-05-15
fix(rspack): use `ts-checker-rspack-plugin`

watsonhaw5566:
### 🔗 Linked issue

No issue

### 📚 Description

Because before "fork-ts-checker-webpack-plugin" Webpack plugin dependencies Webpack when the  user using  @nuxt/rspack-builder must be install the Webpack5 . 

So I used the "ts-checker-rspack-plugin" replace "fork-ts-checker-webpack-plugin".

But now I found the Type error in  "webpack/configs/client.ts" and "webpack/configs/service.ts". I used the @ts-expect-error to solve them.


danielroe (approved):

danielroe:
Thank you ❤️

PR #32119, opened 2025-05-15
fix(nuxt)!: remove public and assets alias to fix relative imports from other sources

### LINKED ISSUE #26164: Monaco volar language worker not working in Nuxt
### Environment

https://codesandbox.io/p/devbox/floral-shape-g3jqfk?workspaceId=e5b65ea6-731c-41c7-b94a-05025ff341b6

### Reproduction

https://codesandbox.io/p/devbox/floral-shape-g3jqfk?workspaceId=e5b65ea6-731c-41c7-b94a-05025ff341b6

### Describe the bug

The monaco volar package is not working correctly in Nuxt. 
The same implementation works fine in a Vite + Vue setup, but in a Nuxt environment the monaco hover content is not displayed correctly. 

There are no errors so it's a bit hard to debug. Setting up custom hover content works so I assume there is a problem with the Worker or Vite configuration, something related to Nuxt.

Nuxt environment(not working)
https://codesandbox.io/p/devbox/floral-shape-g3jqfk

Vue+Vite environment(working)
https://codesandbox.io/p/devbox/nifty-drake-pkr7jz

What is expected to work:
2024-03-09 14_29_06-Window

But does not work in Nuxt:
2024-03-09 14_35_50-Clipboard
 

### Additional context

_No response_

### Logs

_No response_

OrbisK:
### 🔗 Linked issue

should fix https://github.com/nuxt/nuxt/issues/26164

### 📚 Description

[Reference](https://discord.com/channels/473401852243869706/897487139888062506/1371904076232785951)

This PR removes the explicit `assets` and `public` alias. This should fix for example any false positive resolved asset from node modules.


OrbisK:
Looks like removing the aliases resulted in 

```html
  <img
    src="~/assets/logo.svg"
    class="h-20 mb-4"
  >
```
not working

OrbisK:
> Looks like removing the aliases resulted in
> 
> ```
>        src="~/assets/logo.svg"
>     class="h-20 mb-4"
>   >
> ```
> 
> not working

Reason: https://github.com/vitejs/vite-plugin-vue/issues/15

danielroe:
We likely need to add a vite plugin of our own to add support for this until/unless it's fixed upstream. I think there should be no similar issue with the `public` alias, however.

OrbisK:
> I think there should be no similar issue with the `public` alias, however.

Do you mean keeping the `public` alias? I thought there might be some edge cases where a package or assets are built to `@org/pkg/public`, but maybe that's too much of an edge case.

danielroe:
I mean, there is no problem with getting rid of it

packages/schema/src/config/vite.ts
```diff
@@ -34,6 +34,16 @@ export default defineResolvers({
     },
     resolve: {
       extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue'],
+      alias: {
+        $resolve: async (val, get) => {
+          const [srcDir, assetsDir, publicDir] = await Promise.all([get('srcDir'), get('dir.assets'), get('dir.public')])
+          return {
+            [basename(assetsDir)]: resolve(srcDir, assetsDir),
+            [basename(publicDir)]: resolve(srcDir, publicDir),
```
danielroe:
```suggestion
```

packages/schema/src/config/vite.ts
```diff
@@ -34,6 +34,16 @@ export default defineResolvers({
     },
     resolve: {
       extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue'],
+      alias: {
+        $resolve: async (val, get) => {
+          const [srcDir, assetsDir, publicDir] = await Promise.all([get('srcDir'), get('dir.assets'), get('dir.public')])
```
danielroe:
```suggestion
          const [srcDir, assetsDir] = await Promise.all([get('srcDir'), get('dir.assets')])
```

packages/schema/src/config/webpack.ts
```diff
@@ -1,8 +1,25 @@
...
+          const [srcDir, assetsDir, publicDir] = await Promise.all([
+            get('srcDir'),
+            get('dir.assets'),
+            get('dir.public'),
+          ])
+          return {
+            [basename(assetsDir)]: resolve(srcDir, assetsDir),
+            [basename(publicDir)]: resolve(srcDir, publicDir),
+            ...typeof val === 'object' ? val : {},
+          }
```
danielroe:
```suggestion
          const [srcDir, assetsDir] = await Promise.all([
            get('srcDir'),
            get('dir.assets'),
          ])
          return {
            [basename(assetsDir)]: resolve(srcDir, assetsDir),
            ...typeof val === 'object' ? val : {},
          }
```

danielroe (approved):

PR #32121, opened 2025-05-15
test: use vitest workspaces

danielroe:
### 📚 Description

using new nuxt/test-utils v3.19, we can now adopt vitest workspaces for simpler configuration


package.json
```diff
@@ -26,12 +26,12 @@
     "play:preview": "nuxi preview playground",
     "test": "pnpm test:fixtures && pnpm test:fixtures:dev && pnpm test:fixtures:webpack && pnpm test:unit && pnpm test:runtime && pnpm test:types && pnpm typecheck",
```
OrbisK, line 27: I think it shoud be possible to combine at least `test:runtime`, `test:unit`, `test:fixtures` into a single `vitest run` (or `vitest run --project fixtures --project unit --project nuxt`) command. I think this can speedup the tests because vitest can decide the order :)

danielroe: it is indeed!

OrbisK: This might also speed up ci here: https://github.com/nuxt/nuxt/blob/9c45060335ee1514a4e34c9278d100ff707041a2/.github/workflows/ci.yml#L185-L189

danielroe: there are a lot of variants and I'm not sure they we should add them all as separate projects

Issue #12830, opened 2021-11-15
Polyfill cjs globals for vite


pi0:
Ref nuxt/framework#1835, nuxt/bridge#190

Vite considers source input as native ESM while some dependencies might depend on Node.js/CJS globals like `module` and `global`. Doing polyfill can mitigate this potential issue (also making it more consistent with webpack that handles CJS at some level)

We can do it with unenv (#1502) or before that...

Ibochkarev:
+1

bmulholland:
FYI, this is preventing nuxt/storybook from adding Nuxt 3 support: https://github.com/nuxt-community/storybook/pull/385

PR #32130, opened 2025-05-16
fix(vite): do not replace `global` with `globalThis`

### LINKED ISSUE #13941: setting `global` in `vite.define` breaks build

### LINKED ISSUE #32127: Nuxt rewrites inline WAT instruction `global.set` to `globalThis.set`
### Environment

- Operating System: Linux
- Node Version:     v22.15.0
- Nuxt Version:     3.9.3
- CLI Version:      3.10.0
- Nitro Version:    2.8.1
- Package Manager:  yarn@1.22.22
- Builder:          -
- User Config:      devServer, devtools, modules, pwa, ssr, typescript
- Runtime Modules:  @pinia/nuxt@0.5.1, @vite-pwa/nuxt@0.8.1
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-lsfqdb9q?file=app.vue

### Describe the bug

https://github.com/nuxt/nuxt/commit/17f9c54a7913917c009c16e4e3f4cadb922aad4e (v3.1.1) added a replacement of `global` to `globalThis`. It's unclear why. It doesn't take multline strings with backticks into consideration, so trying to embed (in my case) WebAssembly text code as a multiline string sees weird replacements happening if the line starts with `global.`:

```ts
const text = `global.set 1
global.set 2
global.set 3
(global.set 4)
\tglobal.set 5
;global.set 6
 global.get 7`;
```

becomes

```ts
const text = `global.set 1
globalThis.set 2
globalThis.set 3
(global.set 4)
\tglobal.set 5
;global.set 6
 global.get 7`;
```

Note that even though the replacement code is meant to replace all of these, only the `\nglobal.` seems to be affected. Perhaps due to the [delimiters](https://github.com/rollup/plugins/tree/master/packages/replace#word-boundaries) setting.

Workarounds include adding indentation, not using multiline strings, adding parentheses (in the WAT case.) However, that something rewrites my strings seems dangerous.

### Additional context

_No response_

### Logs

```shell-script

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32127

### 📚 Description

previously, we added this workaround (https://github.com/nuxt/framework/pull/4958) to resolve https://github.com/nuxt/nuxt/issues/13941 and help resolve issues with dependencies that used CJS (https://github.com/nuxt/nuxt/issues/12830).

I think we _might_ be able to drop it now. https://github.com/nuxt/nuxt/issues/13941 is no longer reproducible with the current version of Vite.

(Note that Nitro will also [do the same thing](https://github.com/unjs/nitro/blob/eab51cf32b43b31d4b8880eedab94e3e5f767ef4/src/build/config.ts#L57-L62) though, so this might not totally resolve the issue.)

### DIFF
--- a/packages/nuxt/src/app/entry.ts
+++ b/packages/nuxt/src/app/entry.ts
@@ -4,6 +4,8 @@ import type { App } from 'vue'
 // This file must be imported first as we set globalThis.$fetch via this import
 // @ts-expect-error virtual file
 import '#build/fetch.mjs'
+// @ts-expect-error virtual file
+import '#build/global-polyfills.mjs'
 
 import { applyPlugins, createNuxtApp } from './nuxt'
 import type { CreateOptions } from './nuxt'
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -520,6 +520,17 @@ export const publicPathTemplate: NuxtTemplate = {
   },
 }
 
+export const globalPolyfillsTemplate: NuxtTemplate = {
+  filename: 'global-polyfills.mjs',
+  getContents () {
+    // Node.js compatibility
+    return `
+if (!("global" in globalThis)) {
+  globalThis.global = globalThis;
+}`
+  },
+}
+
 export const dollarFetchTemplate: NuxtTemplate = {
   filename: 'fetch.mjs',
   getContents () {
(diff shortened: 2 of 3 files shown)


pi0:
https://github.com/nitrojs/nitro/pull/3364

For Nuxt **Client** builds, you might also add [this](https://github.com/unjs/unenv/blob/main/src/runtime/polyfill/globalthis-global.ts)) to footer since any lib might be working because of this and will be broken without in browser.

PR #32137, opened 2025-05-17
fix(nuxt): suppress client-side errors by crawlers

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/issues/29624

### 📚 Description

the linked issue is _extremely_ pesky

the issue is reproducible across more than just nuxt websites, and seems to be related to a bug in Google's search index which leads to either:
1. storing HTML and then trying to visit it with JS later on, when some of the JS chunks may no longer be available after a deploy
2. exhausting a JS 'budget'

either of these cases can result in a full 500 error page being rendered on the client.

This PR is a relatively safe hotfix for this issue - it will simply suppress the full-page error when triggered by crawlers/bots.

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-root.vue
+++ b/packages/nuxt/src/app/components/nuxt-root.vue
@@ -56,8 +56,12 @@ if (import.meta.dev && results && results.some(i => i && 'then' in i)) {
 const error = useError()
 // render an empty <div> when plugins have thrown an error but we're not yet rendering the error page
 const abortRender = import.meta.server && error.value && !nuxtApp.ssrContext.error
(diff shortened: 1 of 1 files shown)


TheAlexLichter (approved):

moshetanzer:
You sure it's not a good idea to add a log or some sort of telemetry here to track the occurrence? Oof this is a painful bug.

florian-strasser:
This is a super important update - I were already going insane since I thought this has something to do with my hoster... Hoping for a fast next release :-)

harlan-zw:
Looks good

The only minor thing is if someone is using a bug tracker that just watches for console errors (as opposed to a Vue integration using  `onErrorCaptured`), they won't have visibility on any errors thrown for these bots as they're silenced because we don't run the `app:error` hook (which does the console.error).

PR #32139, opened 2025-05-17
fix(nuxt): add watch paths outside `srcDir` to parcel strategy

### LINKED ISSUE #29244: nuxt.config.watch is not watching files
### Environment

- Node Version:     v20.13.1
- Nuxt Version:     3.13.2
- CLI Version:      3.13.2
- Nitro Version:    2.9.7
- Package Manager:  npm@10.5.2
- Builder:          -
- User Config:      future, compatibilityDate, devtools, nitro, experimental, devServer, imports, hooks, runtimeConfig, plugins, extends, alias, typescript, modules, watch
- Runtime Modules:  @nuxtjs/tailwindcss@6.12.1
- Build Modules: 

### Reproduction

https://github.com/nuxt/nuxt/blob/e6c527c9dbf61b61c9b9c6a3be0f691191f10c60/packages/nuxt/src/core/builder.ts#L123

### Describe the bug

The watch property should let you define patterns that will restart the Nuxt dev server when changed.
I was not able to get this to work.

In the source code I see that every path provided in` nuxt.config.watch` is tested against the `pahtsToWatch` array. This array includes `sourceDir` by default. Therefore all paths that are within the source Directory will be skipped and not watched.

I think changing the evaluation to `==` instead of `.startsWith` would fix this issue. Watcher are defined with a depth of 0, so `==` would be save.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/29244
follow-up on https://github.com/nuxt/nuxt/pull/32138

### 📚 Description

we were not watching paths outside `srcDir` with parcel, because of its different API.

this fixes that issue and also adds unit tests (and refactors resolving paths in a slightly more performant way)


Anoesj:
Nice work, thanks @danielroe! Looks like the watchers are working now.

What I noticed since upgrading Nuxt to 3.17.4 is that I now get the `WARN Falling back to chokidar-granular as @parcel/watcher cannot be resolved in your project.` error, which made me think something was wrong with resolving `@parcel/watcher`, but in fact resolving it wasn't the problem in my case. It was other logic in that try/catch block (the `isDirectory` check) that was throwing an error. The watch paths I provided were globs, and that resulted in aforementioned warning, because the `isDirectory` check throws `ENOENT: no such file or directory, lstat '/censored/for/privacy'` on further inspection. When I turned the globs in paths (by removing `/**` or `/**/*`), the warning went away 🙂. 

It might be nice to try/catch parts of that current try/catch block separately and log warnings about "wrong" paths. The Nuxt docs are a bit unclear about how to change your Nuxt config file to work with the parcel watcher.

Sidenote: I was surprised to see Nuxt configuring parcel to watch the parent dirs of passed watch paths, instead of just that path itself. Why is that?

PR #32144, opened 2025-05-18
perf(nuxt): use hash for process template comparison

moshetanzer:


moshetanzer:
is the empty file’s hash causing an expected or unexpected behavior in tests - a bit confused?

danielroe:
I'm closing this for now 🙏

will happily revisit later on - but let's do so with data (e.g. benchmark) to show that this is worth it

PR #32148, opened 2025-05-19
docs: use node v20 for minimum requirement for nuxt setup

iiio2:
Actually I have seen vite has moved to minimum 20+ requirement, Nitro is also I guess. That is why I thought we can go it for Nuxt also because Nuxt is built with Vite + Nitro.


danielroe (approved):

iiio2:
@danielroe , Thanks.

PR #32152, opened 2025-05-19
perf(nuxt): rewrite all `ref` with primitive values to be `shallowRef`

OrbisK:
### 📚 Description

This PR refactors every ref with a primitive value so that it becomes an explicit shallowRef. This should not cause any issues and should provide a slight performance boost.


packages/nuxt/src/app/components/nuxt-layout.ts
```diff
@@ -68,7 +68,7 @@ export default defineComponent({
       return layout
     })
 
-    const layoutRef = ref()
+    const layoutRef = shallowRef()
```
OrbisK, line 71: I am not sure if we can use `useTemplateRef` 🤔

danielroe (approved):

PR #32162, opened 2025-05-20
perf: deallocate out-of-scope references

GalacticHypernova:
### 🔗 Linked issue

A reiteration of #29661 on the latest main branch.
This iteration goes further to optimize related internally-used functions in core functions that execute only once, and reverts previously "optimized" parts that do not benefit from the optimization due to global reference dependencies/closures (hooks, watchers, plugins, etc)


GalacticHypernova:
@danielroe 
I wanted to get your opinion on this because it is a trade-off:

This PR would optimize memory utilization of dev server starts and production builds, because it will allow references to be GC'ed before the process exits (as only then the modules would be unloaded). Especially for the dev server, which runs indefinitely until explicitly shut down, this can be significant for constrained environments, as otherwise the data would be held in memory indefinitely.

However, in case of a `nuxt.config.ts` change in dev mode, which triggers a full dev server restart, it will re-allocate the references for the repeated execution and then once again free them. This includes initial setup of infrastructure (nuxt/nitro/vite/etc), internal modules, and more.

A `nuxt.config` change in development is relatively infrequent compared to regular HMR, which doesn't appear to be affected by it, but if it happens it might have a little slowdown in dev server re-starts.

I was wondering what your thoughts are on this trade-off? Would you still like this PR to be pushed? If we consider this objectively, all the core parts, such as setting

danielroe:
closing as this has a significant number of conflicts - I would love to make changes like this based on hard data - e.g. memory profiling of an actual nuxt server 🙏

PR #32167, opened 2025-05-20
perf(nuxt): use Intl.Collator instead of localeCompare

moshetanzer:

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -42,6 +42,7 @@ interface ScannedFile {
   absolutePath: string
 }
 
+const enUSComparator = new Intl.Collator('en-US')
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/utils.ts
```diff
@@ -133,7 +136,10 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
         continue
       }
 
-      const segmentName = tokens.map(({ value, type }) => type === SegmentTokenType.group ? '' : value).join('')
+      const segmentName = tokens
+        .filter(token => token.type !== SegmentTokenType.group)
+        .map(token => token.value)
+        .join('')
```
danielroe, line 142: I think the previous option is slightly faster as it has fewer iterations
```suggestion
      const segmentName = tokens.map(({ value, type }) => type === SegmentTokenType.group ? '' : value).join('')
```

danielroe (approved):

Issue #32177, opened 2025-05-21
`useAsyncData` state is reset after HMR from 3.17.4


IsraelOrtuno:
### Environment

------------------------------
- Operating System: Darwin
- Node Version:     v22.14.0
- Nuxt Version:     3.17.4
- CLI Version:      3.25.1
- Nitro Version:    2.11.12
- Package Manager:  pnpm@10.11.0
- Builder:          -
- User Config:      compatibilityDate, devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-3hi861fv?file=pages%2Findex.vue

1. Add a new element to the list or change the "test" class name to something else.
2. See how the list does not render anything.
3. No errors.
4. Refresh page, working as it should.

### Describe the bug

After upgrading to 3.17.4, whenever HMR takes place, any `useAsyncData` state seems to be flushed. I've attached some screenshots from DevTools.

Not sure if #32165 or #32096 could be related.

I could not reproduce this in 3.17.3.

### Additional context

<img width="304" alt="Image" src="https://github.com/user-attachments/assets/0e04b66d-d9ec-411e-a85e-1f64d5fbed52" />

<img width="356" alt="Image" src="https://github.com/user-attachments/assets/ef92cbab-410a-4f92-8ec9-a9fb1a5fa545" />

Had to click _Refresh View_ icon to update.

### Logs

```shell-script

```

danielroe:
This regression was caused by https://github.com/nuxt/nuxt/pull/32096.

woldtwerk:
I use ` lazy: import.meta.env.DEV,` as a workaround

Ebube540:
I'm interested in fixing #32177. The `useAsyncData` state resetting after HMR in Nuxt 3.17.4 is a significant developer experience regression. My experience in Nuxt.js, debugging reactivity systems, and understanding of HMR mechanisms makes me well-suited. I'll investigate the core changes in 3.17.4 related to `useAsyncData` and HMR to identify and patch the state persistence issue effectively.

danielroe:
@Ebube540 please do not use LLMs to generate responses, particularly without actually looking at the issue. (This one, for example, already has a PR open to resolve it.)

PR #32182, opened 2025-05-22
test: add regression test for asyncData HMR

OrbisK:
### 📚 Description

This PR add failing test (with `.fail` vairant to pass in ci) for #32177. 

To run this test:

1. add only modifier to test `test/e2e/hmr.test.ts` (L177) -> `test.fail` > `test.only`
1. `pnpm test:e2e hmr`

If we want this PR should be ready to merge, because the test is marked to fail


test/e2e/hmr.test.ts
```diff
@@ -173,5 +173,14 @@ if (process.env.TEST_ENV === 'built' || isWindows) {
       // Verify no unexpected errors
       expect(filteredLogs).toStrictEqual([])
     })
+
+    test.fail('#32177', async ({ page, goto }) => {
```
danielroe:
```suggestion
    test('should allow hmr with useAsyncData (#32177)', async ({ page, goto }) => {
```

danielroe (approved):
this was fixed in https://github.com/nuxt/nuxt/pull/32232 but it's good to have a test - I should have merged this sooner.... ❤️

PR #32223, opened 2025-05-27
chore: clarify runtimeConfig and process.env usage in nuxt.config

### LINKED ISSUE #32222: Docs: process.env usage within nuxt.config is unclear and unintuitive (needs clarification)
This article does not explain updating `.env` variables in production to the extend that I would like it to.

https://nuxt.com/docs/guide/directory-structure/env

paranoidPhantom:
### 🔗 Linked issue

Should resolve #32222 

### 📚 Description

Clarified how runtime config works (I may be wrong in my clarification).


docs/2.guide/2.directory-structure/2.env.md
```diff
@@ -55,6 +55,10 @@ Since `.env` files are not used in production, you must explicitly set environme
 
 * Many cloud service providers, such as Vercel, Netlify, and AWS, provide interfaces for setting environment variables via their dashboards, CLI tools or configuration files.
 
+::warning
+If you are looking to update your environment variables in production, refer to [`runtimeConfig` docs](https://nuxt.com/docs/guide/going-further/runtime-config#environment-variabes) as using `process.env` in an attempt to read environment variables at runtime will not work.
+::
+
```
OrbisK, line 61: I'm not sure if this might be misleading.

Some thoughts:
- Runtime config cannot be updated in production when using generate/static builds.
- There are some places where you can't use runtime config (outside of Nuxt context), where you have to use process.env.
- I think you should always use the runtimeConfig inside the Nuxt context.

Maybe we can simplify this to:
- Always use the runtimeConfig.
- If you are not in a Nuxt context, you can only use process.env.

pi0: In confirmation to @huang-julien yes, `process.env` is guaranteed by nitro to be accessible at least in all server contexts we can also sync same env with context of nuxt config and modules.

Static build is exception.

cybercoder-naj (commented):
I think it is worth adding Alex's [video](https://www.youtube.com/watch?v=_FYV5WfiWvs) on handling environment variables. There are more videos on his channel regarding nuxt's runtimeConfig.

paranoidPhantom:
> Maybe just link to that page instead?
> 
> I am not sure the current warning makes sense in the page about the .env file. 

Sounds like the best option to me, now that I think about it.

docs/2.guide/2.directory-structure/2.env.md
```diff
@@ -55,6 +55,14 @@ Since `.env` files are not used in production, you must explicitly set environme
 
 * Many cloud service providers, such as Vercel, Netlify, and AWS, provide interfaces for setting environment variables via their dashboards, CLI tools or configuration files.
 
+::important
+`nuxt.config.ts` only runs at build time, therefore you can not change anything in it using environment variables at runtime (including runtimeConfig). Refer to Alexamder Lichter's video
```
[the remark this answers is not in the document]
paranoidPhantom: Added an ::important block with a link to runtimeConfig docs. Not sure if it's necessary and will get merged, but I lean towards it being beneficial.

paranoidPhantom:
@danielroe Would love some feedback 🙏

danielroe (approved):
thank you ❤️

PR #32237, opened 2025-05-29
docs: recommend `nuxt` command consistently

danielroe:
### 📚 Description

this simplifies the usage of nuxi/nuxt throughout the project, which should hopefully make things more straightforward for new users


antfu (approved):
Does this mean we are gradually deprecating the usage of `nuxi` (I agree it can be confusing two have two CLI)? If so should we have a section somewhere in our docs to explain the difference between `nuxi` and `nuxt` and tell users whether they should migrate/change over?

BobbieGoede (approved):

blouflashdb:
Will there be a migration script?

OrbisK:
> Will there be a migration script?

`nuxi` will still work. It's just to avoid confusing new users.

cjpearson:
@danielroe Is there any chance we can get a "nuxi vs nuxt" sequel? 😀 https://www.youtube.com/watch?v=R1YLXgCVSec 

I was searching for a quick explainer, but I think this recommendation is now out of date?

cjpearson:
One more follow-up question: For programmatic usage, is it preferable to import `runCommand` from `@nuxt/cli` ([example](https://github.com/nuxt/module-builder/blob/5ba973d1491bcdd507a1a78140b80617f8917025/src/commands/prepare.ts#L22)) or `nuxi` ([example](https://github.com/nuxt/scripts/blob/fb2d62b7e8938ac9f38679b9fba4f0ed1804f8c7/src/kit.ts#L78))?

PR #32238, opened 2025-05-29
fix(nuxt): remove query and params from `useAsyncData` watch list for `useFetch`

huang-julien:
### 🔗 Linked issue

#32233

### 📚 Description

This PR fixes a perf issue and a bug. `query`/`params` are sent to `useAsyncData` for watch but `useFetch` also depends on theses to generate the key that `useAsyncData` also watches to re-trigger the handler.


UfukUstali:
i can also observe the same issue as #32233 when using body.
i think it is a more general bug that is not just limited to query or params

Issue #32053, opened 2025-05-09
💥 Nuxt Page Transition causes a Flash / Blank when navigating — possible layout/rendering issue


kevinkenfack:
## Description

Hi, I’m working on a Nuxt 3 project deployed on Vercel. I’m experiencing an unwanted **white flash** or **layout flicker** on navigation between pages, especially when using `<NuxtLink>` inside the navbar. The transition feels broken or unpolished.

I already use default Nuxt page transitions like this:

```css
.page-enter-active,
.page-leave-active {
  transition: all 0.4s;
}
.page-enter-from,
.page-leave-to {
  opacity: 0;
  filter: blur(1rem);
}
```
My app layout includes the navbar and footer outside of `<NuxtPage />` like this:

```vue
<template>
  <NuxtLoadingIndicator />
  <AppNavbar />
  <div class="h-32" />
  <UContainer>
    <NuxtLayout>
      <NuxtPage />
    </NuxtLayout>
  </UContainer>
  <div class="h-32" />
  <AppFooter />
</template>
```
my index page on page/

```vue
   <template>
  <main class="min-h-screen">
    <div class="space-y-24">
      <HomeIntro />
      <HomePremiumService />
      <HomeWork />
      <HomeSocialLinks />
      <HomeFeaturedProjects />
      <HomeFeaturedArticles />
      <HomeTestimonials />
      <HomeContactCTA />
    </div>
  </main>
</template>
```

I’ve created a video demonstrating the issue I'm encountering with Nuxt. The video shows the flashing behavior when navigating between pages, even though I have set up transitions as expected.

https://github.com/user-attachments/assets/50acdc47-bb18-4f89-b186-f6f220cd731c

Even though the transition is defined, navigation causes a brief white screen before the new content appears. It seems like the layout or components are unmounted/re-mounted unexpectedly.

## What I’ve tried

- Using static assets (`<NuxtImg>`) with placeholder and blur.
- Moving layout elements inside Nuxt layouts.
- Ensuring no SSR hydration errors.
- Checked for large images or fonts causing flashes.
- Tried with `nuxt@3.8.1`, but got package warnings on `glob@7.x`.

## Possible cause

It seems like `<NuxtPage />` renders **before** layout is fully hydrated, or there's a delay with dynamic components (like `useSeoMeta`, runtime config, or `<NuxtImg>`).  
Maybe the `transition` hook isn't synced with the component load.

## Questions

- Is this a known issue with page transitions in Nuxt 3?
[...]

PR #31914, opened 2025-04-27
fix(nuxt): respect the scroll behavior set by `scrollToTop`

Mini-ghost:
### 🔗 Linked issue

- related #31654
- related #31638
- resolves #25241
- Resolves #31650

### 📚 Description

Fix an issue where setting `scrollToTop: false` did not preserve the current scroll position when navigating within the same page component.  

After this PR, #31654, #31638 and #25241 can now be addressed by using the following approach to prevent scrolling to the top when navigating within the same page component:

```vue
<script setup lang="ts">
definePageMeta({
  scrollToTop: false,
})
</script>
```

- [#31654 fixed reproduction](https://stackblitz.com/edit/github-c5oaac3b-ifb1papv?file=pages/foo/%5Bid%5D.vue,package.json)
- [#31638 fixed reproduction](https://stackblitz.com/edit/github-prrsurkb-efbkjy6b?file=pages/schedule/%5Bday%5D.vue)
- [#25241 fixed reproduction](https://stackblitz.com/edit/nuxt-starter-qtuymnfi?file=pages/index.vue,pages/index/dialog.vue)

Additionally, this PR also corrects the registration and invocation order of the `page:loading:end` hook.

In current, when navigating within the same page component, the first navigation triggers `nuxtApp.callHook('page:loading:end')` inside `page.ts` **before** `nuxtApp.hooks.hookOnce('page:loading:end', () => { ... })` is registered in `router.options.ts`.

This explains why, in the provided minimal reproduction, the page does not scroll to the top on the first navigation but does on the second navigation.

https://stackblitz.com/edit/github-c5oaac3b-jwythaqo?file=pages/foo/%5Bid%5D.vue

Lastly, this PR also updates the `hookToWait` condition in `router.options.ts` to ensure that any method of enabling `<Transition>` is properly recognized.

- [#31650 fixed reproduction](https://stackblitz.com/edit/nuxt-starter-lbcyvm6q?file=pages/index.vue)

Thank you very much for your time and review! If there’s anything I might have overlooked, please feel free to let me know.

PR #32239, opened 2025-05-29
fix(nuxt): align scroll behavior with page transition completion

### LINKED ISSUE #32193: Scroll behaviour timing is not properly aligned to page transition duration
### Environment

- Operating System: `Darwin`
- Node Version:     `v18.20.7`
- Nuxt Version:     `3.17.4`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `npm@10.8.2`
- Builder:          `-`
- User Config:      `app`, `compatibilityDate`, `devtools`, `site`, `css`, `ssr`, `modules`, `ogImage`, `seoExperiments`, `linkChecker`, `fonts`, `image`, `vite`
- Runtime Modules:  `@nuxt/image@1.10.0`, `@nuxtjs/seo@3.0.3`, `@nuxt/fonts@0.11.2`, `lenis/nuxt@1.3.3`
- Build Modules:    `-`

### Reproduction

Follow the documentation to apply page transitions: https://nuxt.com/docs/getting-started/transitions

Add this to your nuxt.config.ts file:
```
  app: {
    pageTransition: { name: 'page', mode: 'out-in' }
  },
})
```
As well as this to your app.vue:
```
<template>
  <NuxtPage />
</template>

<style>
.page-enter-active,
.page-leave-active {
  transition: all 0.4s;
}
.page-enter-from,
.page-leave-to {
  opacity: 0;
  filter: blur(1rem);
}
</style>
```

When you switch between pages and you have already scrolled a bit to the bottom you will the jump between the pages.

Thanks to @awfulcode there is already a minimal reproduction for the issue: https://codesandbox.io/p/devbox/festive-cohen-jg4w3d

### Describe the bug

As already mentioned by @awfulcode and others in the Issue https://github.com/nuxt/nuxt/issues/32053 there is a weird bug when using page transitions according to the documentation. There is a problem that the scroll to function is being executed before the page transition finished, results in a jiggy looking page transition since you see the scroll position applied before transition is completed.

I'm currently using a workaround to get around this, but this is not perfect since there is a defined timeout. It would be better if this is properly aligned to the end of the page transition.

### Additional context

_No response_

### Logs

```shell-script

```

Mini-ghost:
### 🔗 Linked issue

Related(?) #32053
Resolve #32193

### 📚 Description

In #31914, we used the `<Transition>` component’s `onBeforeLeave` hook to record whether a page transition is in progress, allowing `router.options.ts` to determine which hook to wait for before performing scroll behavior.

Originally, the sequence of events was as follows:

```
Suspense onPending → Transition onBeforeLeave → router.options.ts → Suspense onResolve → Transition onAfterLeave
```

However, when a page component uses `<script setup>` with top-level `await` expressions or an `async setup()` function, the event order changes to:

```
Suspense onPending → router.options.ts → Transition onBeforeLeave → Suspense onResolve → Transition onAfterLeave
```

In this case, since `<Transition>`’s `onBeforeLeave` is triggered **after** `router.options.ts`, make scroll logic cannot correctly determine whether a transition is in progress.

This PR moves the transition state tracking into the `onPending` hook of `<Suspense>`. I hope this adjustment ensures that scroll behavior remains correctly.

👉 [#32193 fixed reproduction](https://codesandbox.io/p/devbox/stupefied-firefly-m5qpwy?file=/app/pages/%5B...slug%5D.vue:12,17&workspaceId=ws_DxA7J68jekhxuwgY9N69vj)


florian-strasser:
Looks promising - hope it will finally fix the issue :-)

danielroe (commented):
I would love to have a test for this... (as there is a matrix of scroll behaviour and I'd like this not to break again)

ideally a nuxt runtime test rather than an e2e test (which can be notoriously flaky on scroll).

danielroe (approved):

PR #32250, opened 2025-05-31
refactor(nuxt): migrate to `oxc-walker`

cernymatej:
### 📚 Description
This PR removes the old implementation of the parse utils and migrates to the new ones from `oxc-walker`.

_The dependency wasn't included yet, since [this](https://github.com/oxc-project/oxc-walker/pull/68) needs to be released first._


danielroe:
@cernymatej looks like a few tests fail owing to differences in scope tracker - might warrant a look

cernymatej:
noticed that too… I'll investigate

Locally, some tests related to logging failed too - specifically the ones that require the vite plugin defined in root-level nuxt.config.ts (intoduced in the async data layer rewrite).
How should I run the tests to include it, so that it is easier to see which tests really fail? 🙏
Thanks!

danielroe:
you should be able to run `pnpm vitest <name of test file>` now that we have a single vitest file with environments

cernymatej:
That's what I've been using, but for example, `pnpm vitest test/nuxt/composables.test.ts -t="should warn if incompatible options are used"` always fails

image

and if I change the `isDev` check in `asyncData.ts`, it passes
```ts
const isDev = import.meta.dev /* and in test */ || true
```

cernymatej:
The problem was that this PR didn't include the last-minute changes made in `oxc-walker` because I had been waiting for the package to be released.
Should be all good now. (apart from the failing tests - at least locally - I mentioned earlier)

danielroe (approved):

danielroe (approved):

PR #32263, opened 2025-06-02
docs: Update 5.nuxt-loading-indicator.md

1411430556:
### 📚 Description

❶The link in the red box. If the user does not move the mouse here, they will not notice that there is a link that can be clicked here (because it has no style difference from the surrounding text).
Now a prominent clickable logo has been added to it
<details><summary>Details</summary>
<p>
The styles before and after placing the mouse on it

image
</p>
</details> 

image

❷The redundant selected text within the green box has been removed
image


danielroe:
for accessibility, the content of a link should make sense on its own, and describe the destination/link.

so the current form of these links is more correct than this PR change.

instead, may I suggest we make changes to the _styling_ of links in the CSS of https://github.com/nuxt/nuxt.com ? 🙏

PR #32281, opened 2025-06-03
fix(webpack): use plugin for rollup-compatible dynamic imports

danielroe:
### 📚 Description

webpack generates a chunk loading mechanism which doesn't place nicely with nitro.

in nitro v2 there's a built in dynamic-require plugin in the rollup build which handles this, but it has periodic problems when webpack changes its output format, and it also relies on a webpack manifest.

this changes switches to a webpack-native plugin to implement this directly in the webpack build, which allows us to handle changes/bugfixes in nuxt directly


packages/webpack/src/webpack.ts
```diff
@@ -35,29 +35,19 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
...
-        'server.mjs',
-        'server.manifest.mjs',
-      ],
-    })
-    const prerenderRollupPlugins = nitro.options._config.rollupConfig!.plugins as InputPluginOption[]
-    const rollupPlugins = nitro.options.rollupConfig!.plugins as InputPluginOption[]
+    nitro.hooks.hook('rollup:before', (nitro, config) => {
+      const rollupCompatPlugin = dynamicRequire({
+        dir: resolve(nuxt.options.buildDir, 'dist/server'),
+      })
```
danielroe:
```suggestion
      const dynamicRequirePlugin = dynamicRequire({
	      dir: resolve(nuxt.options.buildDir, 'dist/server'),
	      inline:
	      nitro.options.node === false || nitro.options.inlineDynamicImports,
	      ignore: [
	        'client.manifest.mjs',
	        'server.js',
	        'server.cjs',
	        'server.mjs',
	        'server.manifest.mjs',
	      ],
	    })
```

packages/webpack/src/webpack.ts
```diff
@@ -35,29 +35,19 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
...
+      const plugins = config.plugins as InputPluginOption[]
 
-    for (const plugins of [prerenderRollupPlugins, rollupPlugins]) {
       const existingPlugin = plugins.findIndex(i => i && 'name' in i && i.name === 'dynamic-require')
       if (existingPlugin >= 0) {
-        plugins.splice(existingPlugin, 1, dynamicRequirePlugin)
+        plugins.splice(existingPlugin, 1, rollupCompatPlugin)
```
danielroe:
```suggestion
        plugins.splice(existingPlugin, 1, dynamicRequirePlugin)
```

packages/webpack/src/webpack.ts
```diff
@@ -35,29 +35,19 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
...
       } else {
-        plugins.push(dynamicRequirePlugin)
+        plugins.push(rollupCompatPlugin)
```
danielroe:
```suggestion
        plugins.push(dynamicRequirePlugin)
```

PR #32289, opened 2025-06-04
feat(kit): support single import in `addServerImports`

BobbieGoede:
### 📚 Description
Noticed that `addServerImports` does not support passing a single import like [`addImports`](https://nuxt.com/docs/api/kit/autoimports#addimports) does. I haven't actually tested the code changes, but I matched the logic of `addServerImportsDir`.

While adding this I noticed the documentation was missing `addServerImports` entirely so I added that as well, this is basically the same as the documentation for `addImports`, feels a bit weird to duplicate it...

### DIFF
--- a/docs/3.api/5.kit/11.nitro.md
+++ b/docs/3.api/5.kit/11.nitro.md
@@ -305,16 +305,16 @@ import { defineNuxtModule, createResolver, addServerImports } from '@nuxt/kit'
 export default defineNuxtModule({
   setup(options) {
     const names = [
-      "useStoryblok",
-      "useStoryblokApi",
-      "useStoryblokBridge",
-      "renderRichText",
-      "RichTextSchema"
-    ];
+      'useStoryblok',
+      'useStoryblokApi',
+      'useStoryblokBridge',
+      'renderRichText',
+      'RichTextSchema'
+    ]
 
     names.forEach((name) =>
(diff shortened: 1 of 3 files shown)


docs/3.api/5.kit/11.nitro.md
```diff
@@ -293,6 +293,53 @@ function addPrerenderRoutes (routes: string | string[]): void
...
+      "useStoryblok",
+      "useStoryblokApi",
+      "useStoryblokBridge",
+      "renderRichText",
+      "RichTextSchema"
+    ];
+
+    names.forEach((name) =>
+      addServerImports({ name, as: name, from: "@storyblok/vue" })
+    );
```
DamianGlowala:
```suggestion
  setup(options) {
    const names = [
      'useStoryblok',
      'useStoryblokApi',
      'useStoryblokBridge',
      'renderRichText',
      'RichTextSchema'
    ]

    names.forEach((name) =>
      addServerImports({ name, as: name, from: '@storyblok/vue' })
    )
```

BobbieGoede: This is copied from the `addImports` docs, so I have updated the formatting for both.

huang-julien (approved):

danielroe (approved):

PR #32300, opened 2025-06-04
feat(nuxt,kit,schema): add a factory function for `useFetch` and `useAsyncData`

### LINKED ISSUE #14936: Ability to set global options on data fetching composabkes
### Describe the feature

Often it is needed that you need to add a specific header or some other specific thing on all requests made to a specific domain.

So this feature would allow a user to set these global options as part of the nuxt config. Similar to how axios allows you set global options and interceptors.

The level of granularity or complexity of these options would be a matter discussion. 

### Additional information

### Final checks

### LINKED ISSUE #24157: add global interceptors for `useFetch` without creating `customUseFetch`
### Describe the feature

# Description:
The proposal is to implement request and response interceptors for the `useFetch` composable in Nuxt 3, similar to how Axios interceptors function. These interceptors would allow users to globally manage, modify, or intercept HTTP requests before they are sent and responses before they are handled.

# Use Case:
Integrating request and response interceptors within the `useFetch` composable would be highly beneficial for various scenarios. For instance, users could add headers, perform pre-processing or transformations on outgoing requests, log requests or responses, handle errors, or modify responses in a uniform and centralized manner across their application.

# Expected Behavior:
The request interceptors would enable users to intercept and modify the outgoing request before it is sent, while the response interceptors would allow users to intercept and modify the response before it is further processed. These interceptors should seamlessly integrate with the existing useFetch functionality, ensuring flexibility and ease of use.

# Examples or Use Scenarios:
```javascript
// Global request interceptor example
onRequest((config) => {
  // Modify the request config before sending
  const token = useCookie('token');
  config.headers['Authorization'] = `Bearer ${token.value}`
  return config
});

// Global response interceptor example
onResponse((response) => {
  // Log the response or modify it before further processing
  console.log('Response:', response)
  return response
});
```

# Potential Impact:
This feature would significantly enhance the flexibility and control for developers using the `useFetch` composable in Nuxt 3. It would streamline the process of globally handling requests and responses, making it easier to manage authentication, logging, error handling, and other common functionalities across the application.

---

This feature would significantly enhance the flexibility and control for developers using the `useFetch` composable in Nuxt 3. It would streamline the process of globally handling requests and responses, making it easier to manage authentication, logging, error handling, and other common functionalities across an application.

### Additional information

### Final checks

### LINKED ISSUE #28485: useFetch returns null used in component when on parent data is loaded via composable useFetch on route watch
### Environment

```
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/nuxt-starter-rt2zig?file=pages%2Findex.vue

### Describe the bug

On Page Load `Response` is null. But when changing something in code to trigger client refresh (i provided the console.log to trigger update), Response is given. 
It works on page reload when using the doSuccess with another composable method.

While debugging, it seems like that causes it:

In the described behaviour asyncData use `_nuxtOnBeforeMountCbs` to process the initialFetch

https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L372

But, what seems to be a bug, the line ([395](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L395)) resolves in the given case the Promise where it isn´t processed ([359](https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/composables/asyncData.ts#L359)) yet.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #28792: Add a `resetAfter` option to `useFetch` to automatically reset the state after a specified time.
### Describe the feature

I propose to add a new option called `resetAfter` to the `useFetch` composable in Nuxt 3. This option would take a number (in milliseconds) and automatically reset the state of the `useFetch` instance after the specified amount of time has passed since the last fetch.

This feature is particularly useful when you need to fetch data not immediately, but after a certain delay, and possibly several times during the lifecycle of a component. By automatically resetting the state, it reduces the need for manual state management, making it easier to work with time-dependent fetch operations.

#### Suggested changes:
- Add a new `resetAfter` parameter to the `useFetch` composable.
- Automatically clear and reset the state after the specified milliseconds.
- Optionally allow the reset timer to restart on each new fetch.

### Example use case
Here's an example of a Vue 3 component using the `resetAfter` option:

```vue
<template>
  <div>
    <p>Data: {{ data }}</p>
    <p>Error: {{ error }}</p>
    <p>Status: {{ pending ? 'Loading...' : 'Fetched' }}</p>
    <button @click="refetch">Refetch Data</button>
  </div>
</template>

<script setup>
import { ref } from 'vue'
import { useFetch } from '#imports'

const url = ref('https://api.example.com/data')
const { data, error, pending, refresh } = useFetch(url, {
  resetAfter: 5000,  // Reset state 5 seconds after the last fetch
  lazy: true,
  immediate: false
})

const refetch = () => {
  refresh()
}
</script>
```

In this example, the data is automatically reset after 5 seconds, making the component ready for a new fetch if required.

### Additional information

### Final checks

### LINKED ISSUE #33008: Type error on documentation example about Custom useFetch in Nuxt
### Environment

- Node Version:     `v22.18.0`
- Nuxt Version:     `4.0.3`
- CLI Version:      `3.28.0`
- Nitro Version:    `2.12.4`
- Package Manager:  `pnpm@9.15.2`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `css`, `vite`, `modules`, `runtimeConfig`
- Runtime Modules:  `@nuxt/icon@2.0.0`, `@nuxt/devtools@2.6.2`, `@nuxt/image@1.11.0`, `@nuxt/fonts@0.11.4`, `@nuxt/eslint@1.9.0`, `vue3-carousel-nuxt@1.1.6`
- Build Modules:    `-`

### Reproduction

app\composables\useApi.ts

```
import type { UseFetchOptions } from 'nuxt/app'

export function useAPI<T>(
  url: string | (() => string),
  options?: UseFetchOptions<T>,
) {
  return useFetch<T>(url, {
    ...options,
    $fetch: useNuxtApp().$api
  })
}
```

### Describe the bug

I'm using the example on the documentation page [https://nuxt.com/docs/4.x/guide/recipes/custom-usefetch](url) because i'm trying to make my own custom useApi composable and this error show up.

> No overload matches this call.
>   Overload 1 of 2, '(request: NitroFetchRequest | Ref<NitroFetchRequest, NitroFetchRequest> | (() => NitroFetchRequest), opts?: UseFetchOptions<...> | undefined): AsyncData<...>', gave the following error.
>     Argument of type '{ $fetch: $Fetch<unknown, NitroFetchRequest>; key?: MaybeRefOrGetter<string>; watch?: MultiWatchSources | false; default?: (() => globalThis.Ref<...> | undefined) | undefined; ... 38 more ...; onResponseError?: FetchHook<...> | ... 3 more ... | undefined; }' is not assignable to parameter of type 'UseFetchOptions<T extends void ? unknown : T, T extends void ? unknown : T, KeysOf<T extends void ? unknown : T>, undefined, NitroFetchRequest, T extends void ? "get" : AvailableRouterMethod<...>>'.
>       Types of property 'getCachedData' are incompatible.
>         Type '((key: string, nuxtApp: NuxtApp, context: { cause: AsyncDataRefreshCause; }) => NoInfer<T> | undefined) | undefined' is not assignable to type '((key: string, nuxtApp: NuxtApp, context: { cause: AsyncDataRefreshCause; }) => NoInfer<T extends void ? unknown : T> | undefined) | undefined'.
>           Type '(key: string, nuxtApp: NuxtApp, context: { cause: AsyncDataRefreshCause; }) => NoInfer<T> | undefined' is not assignable to type '(key: string, nuxtApp: NuxtApp, context: { cause: AsyncDataRefreshCause; }) => NoInfer<T extends void ? unknown : T> | undefined'.
>             Type 'NoInfer<T> | undefined' is not assignable to type 'NoInfer<T extends void ? unknown : T> | undefined'.
>               Type 'NoInfer<T>' is not assignable to type 'NoInfer<T extends void ? unknown : T> | undefined'.
>                 Type 'T' is not assignable to type 'NoInfer<T extends void ? unknown : T> | undefined'.
>   Overload 2 of 2, '(request: NitroFetchRequest | Ref<NitroFetchRequest, NitroFetchRequest> | (() => NitroFetchRequest), opts?: UseFetchOptions<...> | undefined): AsyncData<...>', gave the following error.
>     Argument of type '{ $fetch: $Fetch<unknown, NitroFetchRequest>; key?: MaybeRefOrGetter<string>; watch?: MultiWatchSources | false; default?: (() => globalThis.Ref<...> | undefined) | undefined; ... 38 more ...; onResponseError?: FetchHook<...> | ... 3 more ... | undefined; }' is not assignable to parameter of type 'UseFetchOptions<T extends void ? unknown : T, T extends void ? unknown : T, KeysOf<T extends void ? unknown : T>, T extends void ? unknown : T, NitroFetchRequest, T extends void ? "get" : AvailableRouterMethod<...>>'.
>       Types of property 'default' are incompatible.
>         Type '(() => Ref<undefined, undefined> | undefined) | undefined' is not assignable to type '(() => (T extends void ? unknown : T) | Ref<T extends void ? unknown : T, T extends void ? unknown : T>) | undefined'.
>           Type '() => Ref<undefined, undefined> | undefined' is not assignable to type '() => (T extends void ? unknown : T) | Ref<T extends void ? unknown : T, T extends void ? unknown : T>'.
>             Type 'Ref<undefined, undefined> | undefined' is not assignable to type '(T extends void ? unknown : T) | Ref<T extends void ? unknown : T, T extends void ? unknown : T>'.
>               Type 'undefined' is not assignable to type '(T extends void ? unknown : T) | Ref<T extends void ? unknown : T, T extends void ? unknown : T>'.

### Additional context

The reason why im using the <T> on the useFetch its because otherwise the errors appears when i use the transform option like this 
```
const { data: brands } = await useApi(`car_brands/`, {
  key: "brands",
  transform: (data: responseData) => data.results,
});
```

### Logs

```shell-script

```

cernymatej:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/14736
resolves https://github.com/nuxt/nuxt/issues/14936
resolves https://github.com/nuxt/nuxt/issues/28792
resolves https://github.com/nuxt/nuxt/issues/28485
resolves https://github.com/nuxt/nuxt/issues/24157
resolves https://github.com/nuxt/nuxt/issues/33008

### 📚 Description

This PR introduces a way to create custom `useFetch` and `useAsyncData` functions with user-defined default options.

These custom instances can be configured with either static options or a dynamic getter, which can be used to access values from `runtimeConfig`, for example.

```ts
// composables/custom-fetch.ts

// ✏️ providing simple defaults
export const useClientFetch = createUseFetch({
  server: false,
})

// ✨ full control over option merging
export const useApiFetch = createUseFetch((currentOptions) => {
  const runtimeConfig = useRuntimeConfig()

  return {
    ...currentOptions,
    baseURL: currentOptions.baseURL ?? runtimeConfig.public.baseApiUrl,
  }
})
```

In the case of a getter, the user must manually handle merging with the usage options, as the returned options object will replace them. When a plain object is passed into the factory, the usage options will automatically override the options set in the factory function.

#### How it works
Both composables rely on automatic key injection. For this reason, it was necessary to create a mechanism to automatically register these new custom instances to `optimization.keyedComposables`. This registration occurs during Nuxt initialization in a new `nuxt:compiler` module, where specified directories are scanned for usages of these factory functions. Each export is processed, and the corresponding composable identifier is registered for key injection.

The factory functions should be treated more like compiler macros. Because of this, there are only a few places where they can be used - as exports from files in the directories scanned by the new Nuxt compiler. To warn users about incorrect usage, the runtime version of these factory functions simply throws an error. Only when they are correctly recognized and processed during compilation are they replaced with the actual factory functions.

> Factories are created using the `defineKeyedFunctionFactory` helper function provided by `@nuxt/kit`.
> Example:
> ```ts
> export const createUseFetch = defineKeyedFunctionFactory({
>   name: 'createUseFetch',
>   factory<...>(...) { ... }
> ```

### 🔦 Nuxt compiler
This is a new core module I added to make it easier to hook into file scanning, should it be needed for something else in the future. The module ensures that files are read only once, and all its plugins share the contents (as well as the AST, if necessary).

It is possible to configure which directories should be scanned by the module, as well as which files within those directories:
```ts
interface CompilerScanDir {
  /**
   * Path (absolute or relative) to the directory to scan for files.
   * Relative paths are resolved against the Nuxt source directory of the project.
   *
   * You can use Nuxt aliases (~ or @) to refer to directories inside the project or directly use an NPM package path similar to `require()`.
   */
  path: string
  /**
   * The file extensions to scan in the specified path.
   *
   * This has no effect if `pattern` is specified.
   */
  extensions?: (JavascriptExtension | (string & {}))[]
  /**
   * Accept Pattern that will be run against the specified path.
   */
  pattern?: string | string[]
  /**
   * Ignore patterns that will be run against the specified path.
   */
  ignore?: string[]
}
```

### 💬 Additional points to discuss
~~- Should provided options like `onRequest` completely override the defaults when calling the composable, or should they merge with them? We could potentially separate these two strategies under different keys in the factory configuration object. Currently, everything is grouped under `defaults`, but I'm considering introducing a second key (e.g., `merge`) that would merge the options. This would allow users to define an `onRequest` handler when calling the composable, while still executing the one defined in the factory function. (similar issue https://github.com/unjs/ofetch/issues/319)~~
- Some time ago, I recall there was a discussion about having `useFetch` pass cookies from the server to the client (by default?). Is this something we want to tackle in this PR, or should we revisit it later? (https://github.com/nuxt/nuxt/issues/32306)
- https://github.com/nuxt/nuxt/issues/28169
- Should we make `UseFetchOptions` extendable? This would allow users to add custom configuration keys to their `useFetch` instances, which they could then handle in the dynamic getter version of the factory function (by applying different logic in request hooks, etc.) - related: https://github.com/nuxt/nuxt/issues/28792

### 🚧 TODO
- [x] decide on the final public API and function names
- [x] tidy up the implementation
- [x] decide whether to expose `@nuxt/kit` utilities for creating factories
- [x] implement `@nuxt/kit` utilities for creating keyed factories
- [ ] ~~export the internal `useFetch` utilities like `generateOptionSegments` and user-friendly types for better customization~~
- [ ] ~~come up with a guide for manual creation of the custom instances (without auto imports and compiler scanning)~~
- [x] improve the AST walking performance with https://github.com/oxc-project/oxc-walker/issues/106
- [x] check auto-imports for keyed composable specifiers too
- [x] add a factory for `useAsyncData()`
- [x] add tests
- [x] add docs (and fix the documented types of useFetch)


maximepvrt:
@cernymatej It would also be great to support creating a custom `$fetch` instance, for example to use when submitting a form (with different headers, base URL, …).

cernymatej:
@maximepvrt that is already possible in using `$fetch.create(/* defaults */)` - or am I missing something?

https://nuxt.com/docs/guide/recipes/custom-usefetch#custom-fetch

maximepvrt:
@cernymatej Yes, you're right. Tt's possible via `$fetch.create`, but currently we have to define it as a plugin and access it through `const { $api } = useNuxtApp()`, which isn’t ideal for DX compared to auto-imported composables.

It would be great if we could also easily create a custom `useFetch` from a `$fetch` instance, something like:

```ts
export const useClientFetch = createUseFetch({ fetch: $api })
```

or

```ts
export const useClientFetch = createUseFetch($api)
```

cernymatej:
You can already register the custom `$fetch` instance to be auto-imported too. Either by putting it inside the `utils` folder, or by configuring the auto-import manually.
```ts
// utils/myFetch.ts
export const $myFetch = $fetch.create()  // will be auto-imported
```

as for providing a custom `$fetch` instance to these custom `useFetch` instances, that is also already possible in this PR:
```ts
// composables/myUseFetch.ts
import { $myFetch } from '../utils/myFetch'

export const myUseFetch = createUseFetch({ $fetch: $myFetch })
```

maximepvrt:
Perfect 🥰🥰

MuhammadM1998:
> Should provided options like onRequest completely override the defaults when calling the composable, or should they merge with them? We could potentially separate these two strategies under different keys in the factory configuration object. Currently, everything is grouped under defaults, but I'm considering introducing a second key (e.g., merge) that would merge the options. This would allow users to define an onRequest handler when calling the composable, while still executing the one defined in the factory function. (similar issue https://github.com/unjs/ofetch/issues/319)

In numerous Nuxt projects, I always need to **merge** these hooks. For example, I log the error in `onResponseError` by default in my `useCustomFetch` factory function, then in upstream calls to `useCustomFetch`, I sometimes need to do something in `onResponseError` too.

This can be granular and on the hook level too! 

 -  Be able to _merge_ **or** _override_ all hooks in a `useCustomFetch` call
 -  Be able to _merge_ some hooks **but** _override_ other hooks in the same `useCustomFetch` call

Maybe this can be achieved by adding a `merge` key that accepts an array of objects, where each object contains the hook name and the desired behavior (merge, override)?

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -266,3 +82,163 @@ function generateOptionSegments<_ResT, DataT, DefaultT> (opts: UseFetchOptions<_
...
+      }, _asyncDataOptions)
+
+      return asyncData
+    }
+
+    return useFetch
+  },
+})
+
+export const useFetch = (createUseFetch as unknown as { __nuxt_factory: typeof createUseFetch }).__nuxt_factory()
```
cernymatej, line 232: not sure how to make this API more user-friendly for module authors who want to manually define the keyed function entry (e.g., `{ name: 'useFetch', argumentLength: ... }` in this case), in order to avoid scanning the module's source code.

this feels weird:
```ts
export function unwrapKeyedFunctionFactory<T extends Function> (fn: T): T {
  return (fn as unknown as { __nuxt_factory: T }).__nuxt_factory
}

export const useFetch = unwrapKeyedFunctionFactory(createUseFetch)()
```

packages/nuxt/src/compiler/module.ts
```diff
@@ -0,0 +1,218 @@
...
+        if (!plugin.afterScan) { return }
+        try {
+          await plugin.afterScan(nuxt)
+        } catch (e) {
+          logger.error(`[nuxt:compiler] Error in \`afterScan\` hook of plugin \`${plugin.name}\``, e)
+        }
+      }))
+    }
+
+    // TODO: add support for HMR
```
cernymatej: HMR support is also a little tricky... the scan plugins adjust `nuxt.options.optimization.keyedComposables` 🤔

```ts
    afterScan: (nuxt) => {
      nuxt.options.optimization.keyedComposables.push(...keyedFunctionsCreatedByFactories)
    },
```

Is there a clever way to handle this without having to restart Nuxt?

packages/nuxt/src/core/nuxt.ts
```diff
@@ -584,6 +577,8 @@ async function initNuxt (nuxt: Nuxt) {
 
   await nuxt.callHook('modules:done')
 
+  await nuxt.callHook('compiler:ready')
```
cernymatej: can we use a different hook? the scanning can't run in `modules:done` because that's the hook where `imports:context` is called, and the unimport context is needed for scanning

MuhammadM1998:
> defineKeyedFunctionFactory({
  name: 'createUseFetch',
  factory<...>(...) { ... }

Does this mean we'll be able to pass/augment default types for body, error, and options ?

maximepvrt:
@cernymatej it's ready ?

atinux:
I don't want to be that guy, but could you start documenting it?

I think we also need to update https://nuxt.com/docs/4.x/guide/recipes/custom-usefetch

cernymatej:
@atinux Sure, I was just waiting on @danielroe's feedback on the API, as I didn't want to rewrite the docs multiple times.  
The rewrite will be quite significant, so I'll do it all at once.

danielroe (approved):

maximepvrt:
@cernymatej & @danielroe 🙌💚

In this discussion https://github.com/nuxt/nuxt/pull/32300#issuecomment-2943962509, custom `$fetch` was presented as a util rather than a plugin. However, in the current documentation, I noticed that the examples still refer to it as a plugin.

I previously opened a PR (#32305) regarding this, which I’ve since closed as obsolete. Would you be open to me reapplying those changes to the new documentation? Specifically, there was an important point regarding the use of `nuxtApp.runWithContext` for handling redirects.

Additionally, I noticed that the reference to unjs/ofetch#319 has been struck through in the issue description. Is this still something you plan to address soon? This is a major pain point for me when managing large-scale applications with complex, multi-level error handling.

danielroe:
yes, thank you @maximepvrt 🙏

PR #32305, opened 2025-06-05
docs: update custom `$fetch` declaration

maximepvrt:
### 🔗 Linked issue

This is linked to the discussion in https://github.com/nuxt/nuxt/pull/32300#issuecomment-2943962509 @cernymatej

### DIFF
--- a/docs/2.guide/4.recipes/3.custom-usefetch.md
+++ b/docs/2.guide/4.recipes/3.custom-usefetch.md
@@ -24,11 +24,11 @@ Let's pretend here that:
 - If the API responds with a `401` status code, we redirect the user to the `/login` page
 
 ```ts [utils/api.ts]
-const { session } = useUserSession()
-
 export const $api = $fetch.create({
     baseURL: 'https://api.nuxt.com',
     onRequest({ request, options, error }) {
+      const { session } = useUserSession()
+
       if (session.value?.token) {
         // note that this relies on ofetch >= 1.4.0 - you may need to refresh your lockfile
         options.headers.set('Authorization', `Bearer ${session.value?.token}`)
@@ -36,7 +36,7 @@ export const $api = $fetch.create({
     },
     async onResponseError({ response }) {
       if (response.status === 401) {
-        await nuxtApp.runWithContext(() => navigateTo('/login'))
+        await useNuxtApp().runWithContext(() => navigateTo('/login'))
       }
     }
 })
--- a/docs/2.guide/4.recipes/3.custom-usefetch.md
+++ b/docs/2.guide/4.recipes/3.custom-usefetch.md
@@ -12,7 +12,7 @@ However, Nuxt provides a way to create a custom fetcher for your API (or multipl
 
 ## Custom `$fetch`
 
-Let's create a custom `$fetch` instance with a [Nuxt plugin](/docs/guide/directory-structure/plugins).
+Let's create a custom `$fetch` instance with a [Nuxt utility](/docs/guide/directory-structure/utils).
 
 ::note
 `$fetch` is a configured instance of [ofetch](https://github.com/unjs/ofetch) which supports adding the base URL of your Nuxt server as well as direct function calls during SSR (avoiding HTTP roundtrips).
@@ -23,11 +23,10 @@ Let's pretend here that:
 - We are storing the JWT token in a session with [nuxt-auth-utils](https://github.com/atinux/nuxt-auth-utils)
 - If the API responds with a `401` status code, we redirect the user to the `/login` page
 
-```ts [plugins/api.ts]
-export default defineNuxtPlugin((nuxtApp) => {
-  const { session } = useUserSession()
+```ts [utils/api.ts]
+const { session } = useUserSession()
 
-  const api = $fetch.create({
+export const api = $fetch.create({
     baseURL: 'https://api.nuxt.com',
     onRequest({ request, options, error }) {
       if (session.value?.token) {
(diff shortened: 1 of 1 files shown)


MuhammadM1998:
Thanks for this 🙏

I've been confused about this for a while. I thought the benefit of creating a custom fetch instance in a plugin is that it's the first thing loaded in a Nuxt app, and otherwise, it won't work correctly.

However, I have a bunch of projects where I'm creating custom `$fetch` and `useFetch` in a **composable** way, and it's working as expected while also having the auto-import benefit. Why can't we just do that? More info explaining this would be useful

maximepvrt:
@MuhammadM1998 I had the same questions. That’s exactly why I opened the PR, to clarify the best practice and improve DX.

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -40,22 +39,13 @@ export default defineNuxtPlugin((nuxtApp) => {
         await nuxtApp.runWithContext(() => navigateTo('/login'))
```
danielroe: nuxtApp is not defined anywhere.

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -23,11 +23,10 @@ Let's pretend here that:
 - We are storing the JWT token in a session with [nuxt-auth-utils](https://github.com/atinux/nuxt-auth-utils)
 - If the API responds with a `401` status code, we redirect the user to the `/login` page
 
-```ts [plugins/api.ts]
-export default defineNuxtPlugin((nuxtApp) => {
-  const { session } = useUserSession()
+```ts [utils/api.ts]
+const { session } = useUserSession()
```
danielroe: composables shouldn't be called at the top level ('ambiently' in JS) - this relies on being able to access the Nuxt context and will fail on the server.

danielroe (commented):
sorry for the delay in reviewing this! 🙏 

(I've marked as a draft just for my own triage - if you can unmark it whenever you want me to have a look at it again, that would be very helpful!)

docs/2.guide/4.recipes/3.custom-usefetch.md
maximepvrt: fixed

docs/2.guide/4.recipes/3.custom-usefetch.md
maximepvrt: fixed

docs/2.guide/4.recipes/3.custom-usefetch.md
```diff
@@ -23,39 +23,29 @@ Let's pretend here that:
...
+
       if (session.value?.token) {
         // note that this relies on ofetch >= 1.4.0 - you may need to refresh your lockfile
         options.headers.set('Authorization', `Bearer ${session.value?.token}`)
       }
     },
     async onResponseError({ response }) {
       if (response.status === 401) {
-        await nuxtApp.runWithContext(() => navigateTo('/login'))
+        await useNuxtApp().runWithContext(() => navigateTo('/login'))
```
Mini-ghost: I was wondering whether we officially recommend using `useNuxtApp` in any context. Some composables explicitly mention that they must be used within the Nuxt context—such as Nuxt plugins, Nuxt hooks, middleware, or setup functions.

However, this example uses `useNuxtApp` inside `onResponseError`, and it appears to work without any issues. That said, it’s a bit confusing to determine when `useNuxtApp` must strictly be used within those specific Nuxt contexts, and when it’s safe to use it more freely.

maximepvrt: @Mini-ghost I made this change because @cernymatej suggested using it as a utils to benefit from auto-import rather than declaring it as a plugin : https://github.com/nuxt/nuxt/pull/32300#issuecomment-2943962509.

I’ll admit I didn’t really consider the question of whether it was being used within the proper Nuxt context.

danielroe: (this will only work, I believe, on the client.)

maximepvrt: `runWithContext` was added in relation to this issue: https://github.com/nuxt/nuxt/issues/27991, but if `$api` is now defined as a utility, it’s probably no longer needed.

Mini-ghost: Sorry for the late reply! ~This looks great to me — I think using a utility is more convenient than a plugin in practice.~

Mini-ghost (approved):

docs/2.guide/4.recipes/3.custom-usefetch.md
Mini-ghost: I apologize for the earlier incorrect information — I overlooked some details. This block of code can cause issues on the **server** side. While calling `useNuxtApp` from almost anywhere is fine on the **client**, on the **server** it must only be used inside setup scripts, Nuxt plugins, or Nuxt router middleware, because those are the places processed by `UnctxTransformPlugin`.

截圖 2025-09-13 晚上11 20 12

maximepvrt: @Mini-ghost According to the docs (https://nuxt.com/docs/4.x/api/composables/use-nuxt-app#runwithcontext), it's useful to add `runWithContext` only in a plugin or middleware. I have removed its usage in my PR.

Mini-ghost: In this example, the code still results in an error because `navigateTo` internally relies on `useNuxtApp`. 🥲

maximepvrt: @cernymatej can you help us ?

Mini-ghost: I did some digging and it seems the key difference is whether `useAsyncData` is awaited.

When `await useAsyncData` is used, `navigateTo` works as expected:

```ts
// ✅ This works
const { data } = await useAsyncData(() => $fetch('/api/test', {
  async onResponseError () {
    await navigateTo('/')
  },
}))
```

Without `await useAsyncData`, `navigateTo` throws an error:

```ts
 // ❌ This does not work
const { data } = useAsyncData(() => $fetch('/api/test', {
  async onResponseError () {
    await navigateTo('/')
  },
}))
```

After looking into Nuxt’s internals, this behavior makes sense, but I’m not sure it’s intuitive for users. Aside from that, the documentation itself seems fine.

The original approach is a bit more verbose but less error-prone. The approach proposed in this PR makes the code cleaner, but it requires following a specific pattern, which may be harder to grasp.

I’m comfortable with either direction—@danielroe, I’d be grateful for your thoughts.

Issue #32211, opened 2025-05-26
Missing `defineNuxtPlugin` documentation


BobbieGoede:
While the function is partially/indirectly documented on https://nuxt.com/docs/guide/directory-structure/plugins#creating-plugins, there is no dedicated `API > Utils > defineNuxtPlugin` page which describes the function signature/parameters in full.

This is problematic as searching for `defineNuxtPlugin` in the documentation has the [migration guide for `@nuxtjs/composition-api`](https://nuxt.com/docs/bridge/bridge-composition-api#definenuxtplugin) as the first result, which describes it as a helper function which was removed (related https://github.com/nuxt/nuxt/issues/22433).

<img width="801" alt="Image" src="https://github.com/user-attachments/assets/ba237867-2fe7-4876-84f5-5b87e07b6237" />

Though there is a notice that there is a new function with the same name at the bottom, with a broken read more link.

<img width="890" alt="Image" src="https://github.com/user-attachments/assets/a91e8a18-cdc9-4377-805a-f9d6794b1b25" />

Might be a good first issue if anyone is up for it, might be a while until I have time to work on this myself.

PR #32321, opened 2025-06-06
fix(nuxt): use augmented `NuxtApp` as plugin argument type

BobbieGoede:
### 📚 Description
The `_NuxtApp` type used for the argument of `defineNuxtPlugin` does not include augmentations, if this is intended I am curious to know why.

Every now and then an issue opens on nuxt-i18n that claim that the types are incorrect, this is the case when trying to access `nuxtApp.$i18n` rather than `useNuxtApp().$i18n`. But the same applies to other augmentations like `nuxtApp.$router`, but my guess is that users probably work around that by using the `useRouter()` composable.

Reproduction https://stackblitz.com/edit/github-z4qhnxvg?file=plugins%2Fmy-plugin.ts

The type issue
image

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -224,7 +224,7 @@ export interface ResolvedPluginMeta {
 }
 
 export interface Plugin<Injections extends Record<string, unknown> = Record<string, unknown>> {
-  (nuxt: _NuxtApp): Promise<void> | Promise<{ provide?: Injections }> | void | { provide?: Injections }
+  (nuxt: NuxtApp): Promise<void> | Promise<{ provide?: Injections }> | void | { provide?: Injections }
   [NuxtPluginIndicator]?: true
   meta?: ResolvedPluginMeta
 }


danielroe:
This is intended as otherwise it creates a circular reference. If required, you can use `const nuxtApp = useNuxtApp()` in the plugin, but in general it's advised to avoid as this indicates a dependency _betweeen_ plugins.

You can point people to https://nuxt.com/docs/guide/directory-structure/plugins#typing-plugins.

BobbieGoede:
> This is intended as otherwise it creates a circular reference. If required, you can use `const nuxtApp = useNuxtApp()` in the plugin ...
> You can point people to [nuxt.com/docs/guide/directory-structure/plugins#typing-plugins](https://nuxt.com/docs/guide/directory-structure/plugins#typing-plugins).

My guess is that the confusion stems from the argument name and the [docs](https://nuxt.com/docs/guide/directory-structure/plugins#creating-plugins) referring (at least linking) to it being the same as `useNuxtApp`.

Having `$config` typed on `_NuxtApp` could give the impression that injections are supposed to be on there as well.

These details will be useful for https://github.com/nuxt/nuxt/issues/32211

Issue #32213, opened 2025-05-27
Setting `experimental.defaults.useAsyncData.value` does not update TS types


xak2000:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v20.10.0`
- Nuxt Version:     `3.17.4`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `yarn@1.22.19`
- Builder:          `-`
- User Config:      `future`, `compatibilityDate`, `experimental`, `features`, `unhead`, `modules`, `eslint`, `components`, `primevue`, `shadcn`, `colorMode`, `dayjs`, `veeValidate`, `app`, `nitro`, `runtimeConfig`
- Runtime Modules:  `@nuxt/eslint@1.4.1`, `@vueuse/nuxt@13.2.0`, `@nuxtjs/tailwindcss@6.14.0`, `@headlessui-float/nuxt@0.15.0`, `@primevue/nuxt-module@4.3.4`, `dayjs-nuxt@2.1.11`, `shadcn-nuxt@2.1.0`, `@nuxtjs/color-mode@3.5.2`, `@vee-validate/nuxt@4.15.0`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-glf1dfbb?file=nuxt.config.ts

Open `useMyData.ts` and check the `res` type. It should be `AsyncData<string | null, ...>`, but it is `AsyncData<string | undefined, ...>` desipte the fact that `experimental.defaults.useAsyncData.value` is set to `'null'` in `nuxt.config.ts.`.

Also please check `.nuxt/types/app.defaults.d.ts` file, that shows that the type is really `undefined` (`type DefaultAsyncDataValue = undefined`).

### Describe the bug

When `future.compatibilityVersion` is set to `4` and `experimental.defaults.useAsyncData.value` is set to `'null'` then `useAsyncData` and `useFetch` use `null` as the default. This works in runtime as expected.

But the type of `DefaultAsyncDataValue` generated in `.nuxt/types/app.defaults.d.ts` is still `undefined`. I thought the whole point of introduction of `DefaultAsyncDataValue` type was to make this change backward compatible.

I even found a template in #27294 that supposedly should be responsible for this: https://github.com/nuxt/nuxt/pull/27294/files#diff-2176c7ab69dcd5ffc94e79963693b007471788f25591aeb72e7533240ebad2c2

[...]

PR #32324, opened 2025-06-06
fix(nuxt): generate correct types for async data defaults based on `nuxt.config`

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32213

### 📚 Description

Correctly generates types for chosen async data/data error value when v4 compat mode is enabled.


packages/nuxt/src/core/templates.ts
```diff
@@ -105,9 +105,9 @@ export const appDefaults: NuxtTypeTemplate = {
     const isV4 = ctx.nuxt.options.future.compatibilityVersion === 4
     return `
 declare module 'nuxt/app/defaults' {
-  type DefaultAsyncDataErrorValue = ${isV4 ? 'undefined' : 'null'}
-  type DefaultAsyncDataValue = ${isV4 ? 'undefined' : 'null'}
-  type DefaultErrorValue = ${isV4 ? 'undefined' : 'null'}
+  type DefaultAsyncDataErrorValue = ${isV4 ? ctx.nuxt.options.experimental.defaults.useAsyncData.errorValue : null}
+  type DefaultAsyncDataValue = ${isV4 ? ctx.nuxt.options.experimental.defaults.useAsyncData.value : null}
+  type DefaultErrorValue = ${isV4 ? undefined : null}
```
danielroe:
```suggestion
  type DefaultAsyncDataErrorValue = ${ctx.nuxt.options.experimental.defaults.useAsyncData.errorValue}
  type DefaultAsyncDataValue = ${ctx.nuxt.options.experimental.defaults.useAsyncData.value}
  type DefaultErrorValue = ${isV4 ? 'undefined' : 'null'}
```

packages/nuxt/src/core/templates.ts
```diff
@@ -105,8 +105,8 @@ export const appDefaults: NuxtTypeTemplate = {
     const isV4 = ctx.nuxt.options.future.compatibilityVersion === 4
     return `
 declare module 'nuxt/app/defaults' {
-  type DefaultAsyncDataErrorValue = ${isV4 ? 'undefined' : 'null'}
-  type DefaultAsyncDataValue = ${isV4 ? 'undefined' : 'null'}
+  type DefaultAsyncDataErrorValue = ${ctx.nuxt.options.experimental.defaults.useAsyncData.errorValue}
+  type DefaultAsyncDataValue = ${ctx.nuxt.options.experimental.defaults.useAsyncData.value}
```
OrbisK, line 109: what if there is some object here?

danielroe: it's disallowed by the types here:

packages/schema/src/types/schema.ts

PR #32356, opened 2025-06-10
feat(nuxt,schema): add top-level `prerender` alias for `nitro.prerender`

### LINKED ISSUE #15440: Introduce `nuxi prerender` and top level `prerender` in `nuxt.confing`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/15440

### 📚 Description

this adds a new top-level `prerender` optioni, which is an alias for `nitro.prerender`, much as nuxt also has a top-level `runtimeConfig` and `routeRules`. It will also be a place we can add nuxt-specific prerendering options in future.

### DIFF
--- a/docs/1.getting-started/15.prerendering.md
+++ b/docs/1.getting-started/15.prerendering.md
@@ -38,7 +38,7 @@ You can now deploy the `.output/public` directory to any static hosting service 
 
 Working of the Nitro crawler:
 
-1. Load the HTML of your application's root route (`/`), any non-dynamic pages in your `~/pages` directory, and any other routes in the `nitro.prerender.routes` array.
+1. Load the HTML of your application's root route (`/`), any non-dynamic pages in your `~/pages` directory, and any other routes in the `prerender.routes` array.
 2. Save the HTML and `payload.json` to the `~/.output/public/` directory to be served statically.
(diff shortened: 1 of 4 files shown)


pi0:
Looks good for top level only i was thinking we could simplify top level usge, `prerender: boolean | string[]` (some advanced options from nitro are not always best also to be set by users and specific to deployment preset)

danielroe:
I almost always need to configure more than just `true/string[]`. It would also be difficult down the line if we expand options + module authors use it - they would need to check.

(so they couldn't easily push a string, for example)

the benefit of keeping it as an alias (+ future additional nuxt options) is that the documentation + jsdoc types flow through from nitro

pi0:
I believe module authors should prefer hooks to extend prerender routes.

My main concern is that advanced options like `autoSubfolderIndex`, `ignoreUnprefixedPublicAssets` would be exposed alone, so I very much prefer if it were a cherry-picked subset at least. (We might also make Nitro v3 config safer by relocating advanced options, but that also can cause more breaking changes surface for Nuxt)

(+ future additional nuxt options) This also makes sense that we don't simply portal same interface but map it to nitro.

danielroe:
a cherry-picked subset works well for me 👌

PR #32361, opened 2025-06-11
docs: mention that NuxtLink encodes query params

lazercaveman:
📚 Description
This PR adds a new “Query Parameters & URL Encoding” section to the documentation. By explicitly documenting NuxtLink’s built-in encoding behavior, this change closes an important gap in the reference and prevents developers from manually using encodeURI/encodeURIComponent.


danielroe (approved):
thank you ❤

PR #32373, opened 2025-06-12
feat(nuxt,schema)!: allow setting global defaults for custom page meta

### LINKED ISSUE #32363: Allow defining defaults for definePageMeta()
## Problem:
[`definePageMeta()`](https://nuxt.com/docs/api/utils/define-page-meta) is super helpful to set certain (custom) route meta values.
[The docs also explain how to type this custom meta data](https://nuxt.com/docs/guide/directory-structure/pages#typing-custom-metadata).

The only point I'm really missing in this concept is being able to define a default / fallback value.

For example when a few pages (login/registration/password reset/...) should not need auth, it would be great to be able to just write:
```
<!-- pages/example.vue -->

### Override the default in single pages:
```

<script setup>
definePageMeta({
  // Override value for this page:
 auth: false
})
</script>
```
And set `auth: true` as default somewhere centrally.

Current workaround only seems to be to check the value for _undefined_ and treat it like `true`?


## Related documentation:
- https://nuxt.com/docs/api/utils/define-page-meta
- https://nuxt.com/docs/guide/directory-structure/pages#page-metadata
- https://nuxt.com/docs/guide/directory-structure/pages#typing-custom-metadata

## Proposed solution:

### a) Define the default(s) in app.vue
```

<script setup>
definePageMeta({
 // Define default: auth: true
 auth: true
})
</script>
```

### b) Define the default(s) in nuxt.config:
```

  [...],
  app: {
    pageMeta: {
       // Define default: auth: true
      auth: true,
    }
  },

### Override the default in single pages:
```
<!-- pages/example.vue -->
<script setup>
definePageMeta({
  // Override value for this page:
 auth: false
})
</script>
```

What do you think about this concept? Or is it already possible and not yet documented (or I'm too blind to find it)?

Thank you!

### Additional information

### Final checks

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32363

### 📚 Description

This allows setting global defaults for custom page meta in `nuxt.config`. Couldn't find a way to type `app.customPageMeta` in a non-breaking way (unless some TS wizardry could make it possible).


xjccc:
I don't know if `customPageMeta` will be more, and if this needs to disable all `customPageMeta` in `definePageMeta`?
```
definePageMeta({
   customPageMeta: false
})
```

danielroe:
I think we should be able to type this in a non-breaking way.

PR #32374, opened 2025-06-12
chore: install corepack globally

iiio2:
@danielroe , actually I am not sure, but when we use `corepack enable`, shouldn't we need to install it first globally


danielroe (commented):
not anymore, as the fix shipped in the latest node versions

iiio2:
oh I see. Thanks. @danielroe

PR #32376, opened 2025-06-12
fix(nuxt): scroll to anchor if present when changing page without saved position

### LINKED ISSUE #32366: NuxtLink with anchor (hash) doesn't work unless defined as external
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     3.17.5
- CLI Version:      3.25.1
- Nitro Version:    2.11.12
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-d1n2dope?file=pages%2Findex.vue

Click a NuxtLink that links to a specific element on another page, like `/example#two`. The section is only scrolled into view when reloading the page.

### Describe the bug

When using a link to a specific element, that element is not scrolled into view.
Adding `external` works, but seems to trigger server-side rendering.

### Additional context

I'm not sure when the problem appeared, but navigating straight to a specific section used to work.
Our end-to-end tests failed after upgrading from 3.16.0 to 3.17.5, though I could reproduce the error on 3.16.0 in StackBlitz.

### Logs

```shell-script

```

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32366

### 📚 Description

Moved logic which decides on scroll position to a single function for better readability. It now scrolls to hash if present when changing page without saved position. Removed explicit `instant` behavior as this appears to be the default one anyway.


packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -68,23 +60,23 @@ function _getHashElementScrollMarginTop (selector: string): number {
...
 ): ScrollPosition {
-  // Handle saved position for backward/forward navigation
-  if (position) {
-    return position
+  if (!savedPosition && isChangingPage(to, from) && !to.hash) {
+    return { left: 0, top: 0 }
+  }
+
+  if (savedPosition) {
+    return savedPosition
```
danielroe: I guess these can be flipped?

packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -68,23 +60,23 @@ function _getHashElementScrollMarginTop (selector: string): number {
...
   }
 
-  // Scroll to the element specified in the URL hash, if present
   if (to.hash) {
     return {
       el: to.hash,
       top: _getHashElementScrollMarginTop(to.hash),
-      behavior: scrollBehaviorType,
```
danielroe, line 84: I think the 'instant' behaviour is necessary when user has specified smooth scroll behaviour, to avoid smooth scroll to a saved position. in that case (in initial load), it should be instant.

Issue #31915, opened 2025-04-27
NUXT_APP_BASE_URL with Dynamic Routes: Base URL Stacking Issue on Non-Existent Pages


webrsb:
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     3.16.2
- CLI Version:      3.25.0
- Nitro Version:    2.11.9
- Package Manager:  npm@10.8.2
- Builder:          -
- User Config:      compatibilityDate
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-4na34dtw?file=.env

### Describe the bug

When setting NUXT_APP_BASE_URL alongside a Dynamic Routes page, accessing a non-existent page may result in the Base URL continuously stacking.

### Additional context

https://github.com/user-attachments/assets/ab92e20c-46ce-47f2-8a8d-85f5c527863b

### Logs

```shell-script

```

huang-julien:
that's strange I can't reproduce it locally nor in stackblitz

webrsb:
I have created a [repository](https://github.com/webrsb/nuxt-base-bug) that can reproduce this issue. You can give it a try.

DamianGlowala:
@webrsb I can reproduce in Nuxt v3.16.2, thanks! However, as this appears to be fixed in latest Nuxt (v3.17.5), let me close this issue.

PR #32382, opened 2025-06-13
fix(nuxt): account for app's base URL when reloading app in `nuxt:chunk-reload-immediate`

### LINKED ISSUE #32312: chunk-reload-immediate client composable does not take into account the base path, thus can reload and send users to a 404
### Environment

Nuxt 3.17.5

### Reproduction

A reproduction has not been provided, though I have tried to be very clear in my example below.

## Description 

On statically generated sites that have a base app path set, if we set the experimental flag `emitRouteChunkError: 'automatic-immidiate'` then when a new build is deployed and nuxt fires the composable the app reloads and is sent to an incorrect URL that doesn't take into account app base path. 

## Example 

(In the GitLab CI/CD Pipeline):

`NUXT_APP_BASE_URL="/project-dashboard/"`

The user will be on this URL: 

```
https://[redacted].ca/project-dashboard/en/projects/dsfintake-678/
```
And then, after a new build is pushed out and the nuxt plugin (emit chunk immediate) fires the user gets forwarded here

```
https://[redacted].ca/en/projects/dsfintake-678/
```

In the console log you can see the `GET` request is missing the base url. 

Image

And this is validated in the Browser Inspector Network tab. After the app gets the latest manifest, it tries to redirect / reload the page at the wrong URL and gets a 404. 

Image

## Proposed fix 

By modifying the `nuxt:chunk-reload-immediate` to take into account the base app folder similar to https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/plugins/chunk-reload.client.ts where a helper function calculates the full path, I believe this could be easily fixed. 

```ts
import { joinURL } from 'ufo'
import { defineNuxtPlugin, useRuntimeConfig } from '../nuxt'
import { reloadNuxtApp } from '../composables/chunk'
import { addRouteMiddleware } from '../composables/router'

const reloadNuxtApp_ = (path: string) => { 
  reloadNuxtApp({ persistState: true, path }) 
}

// See https://github.com/nuxt/nuxt/issues/23612 for more context
  name: 'nuxt:chunk-reload-immediate',
  setup (nuxtApp) {
    
    // Helper function to construct full path with base URL
    function getFullPathWithBase(route: { fullPath: string, href?: string }) {
      const isHash = 'href' in route && (route.href as string)[0] === '#'
      return isHash ? config.app.baseURL + route.href : joinURL(config.app.baseURL, route.fullPath)
    }
    
    // Remember the full path (including base) when navigating to a new path
    let currentlyNavigationTo: null | string = null
    addRouteMiddleware((to) => {
      currentlyNavigationTo = getFullPathWithBase(to)
    })
    
    // Reload when a `chunkError` is thrown
    nuxtApp.hook('app:chunkError', () => 
      reloadNuxtApp_(currentlyNavigationTo ?? getFullPathWithBase(nuxtApp._route))
    )
    
    // Reload when the app manifest updates
    nuxtApp.hook('app:manifest:update', () => 
      reloadNuxtApp_(getFullPathWithBase(nuxtApp._route))
    )
  },
})
```

## Additional information 

https://github.com/nuxt/nuxt/issues/23612

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32312

### 📚 Description

Takes into account app's base URL when reloading an app on chunk error/manifest update.

### DIFF
--- a/packages/nuxt/src/app/plugins/chunk-reload-immediate.client.ts
+++ b/packages/nuxt/src/app/plugins/chunk-reload-immediate.client.ts
@@ -1,23 +1,32 @@
-import { defineNuxtPlugin } from '../nuxt'
+import type { RouteLocationNormalized } from 'vue-router'
+import { joinURL } from 'ufo'
+import { defineNuxtPlugin, useRuntimeConfig } from '../nuxt'
 import { reloadNuxtApp } from '../composables/chunk'
 import { addRouteMiddleware } from '../composables/router'
-
-const reloadNuxtApp_ = (path: string) => { reloadNuxtApp({ persistState: true, path }) }
 
 // See https://github.com/nuxt/nuxt/issues/23612 for more context
 export default defineNuxtPlugin({
   name: 'nuxt:chunk-reload-immediate',
   setup (nuxtApp) {
     // Remember `to.path` when navigating to a new path: A `chunkError` may occur during navigation, we then want to then reload at `to.path`
-    let currentlyNavigationTo: null | string = null
+    let currentlyNavigationTo: RouteLocationNormalized | null = null
+
     addRouteMiddleware((to) => {
-      currentlyNavigationTo = to.path
+      currentlyNavigationTo = to
     })
 
+    const config = useRuntimeConfig()
+
+    function reloadAppAtPath (to: RouteLocationNormalized) {
+      const isHash = 'href' in to && (to.href as string)[0] === '#'
+      const path = isHash ? config.app.baseURL + (to as any).href : joinURL(config.app.baseURL, to.fullPath)
+      reloadNuxtApp({ path, persistState: true })
+    }
+
     // Reload when a `chunkError` is thrown
-    nuxtApp.hook('app:chunkError', () => reloadNuxtApp_(currentlyNavigationTo ?? nuxtApp._route.path))
+    nuxtApp.hook('app:chunkError', () => reloadAppAtPath(currentlyNavigationTo ?? nuxtApp._route))
 
     // Reload when the app manifest updates
-    nuxtApp.hook('app:manifest:update', () => reloadNuxtApp_(nuxtApp._route.path))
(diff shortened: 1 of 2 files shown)


OrbisK:
@DamianGlowala I wonder it this can fix https://github.com/nuxt/nuxt/issues/31915 too?

DamianGlowala:
@OrbisK as this PR introduces changes which take effect only behind an experimental flag, I am pretty confident it doesn't fix mentioned issue

UPDATE: https://github.com/nuxt/nuxt/issues/31915 is no longer relevant in latest Nuxt (v3.17.5)

packages/nuxt/src/app/plugins/chunk-reload-immediate.client.ts
```diff
@@ -1,23 +1,32 @@
...
+
     addRouteMiddleware((to) => {
-      currentlyNavigationTo = to.path
+      currentlyNavigationTo = to
     })
 
+    const config = useRuntimeConfig()
+
+    function reloadAppAtPath (to: RouteLocationNormalized) {
+      const isHash = 'href' in to && (to.href as string)[0] === '#'
```
danielroe: why are we checking `href`? when would this trigger?

DamianGlowala: this is based on https://github.com/nuxt/nuxt/blob/9c7398ba1a8bbc2a01e64fbbc9621cc14328b4a2/packages/nuxt/src/app/plugins/chunk-reload.client.ts#L19

added first here:
https://github.com/GalacticHypernova/nuxt/commit/7dce07653c9c789dd914edb0e75f44f58634bedd#diff-04bd18e1bfe9c11c2c833477d4de246b7e0b13d4f1d721b1a540a9f3c98ea412R19

Might be worth revisiting this logic in both places?

danielroe (approved):

PR #32393, opened 2025-06-14
fix(nuxt): when `<NuxtLink>` renders native anchor tag, handle click with `navigateTo`

### LINKED ISSUE #32329: NuxtLink hashtag and scrollBehavior smooth
### Environment

- Operating System: `Darwin`
- Node Version:     `v20.19.0`
- Nuxt Version:     `3.17.5`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `npm@10.8.2`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`, `router`, `srcDir`, `dir`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/~/github.com/yuriystasiv/nuxt-link-hashtag
https://github.com/yuriystasiv/nuxt-link-hashtag

### Describe the bug

After updating Nuxt from 3.11.2 to 13.17.5, I noticed that `scrollBehavior: 'smooth'` no longer works correctly when using `NuxtLink` with a hash (#). Navigation to the element still occurs, but instead of a smooth scroll, it jumps immediately to the target and only applies the smooth scroll animation at the very end, which feels abrupt.

This seems to be caused by NuxtLink now rendering a plain <a> tag instead of RouterLink. This change happened in [this PR](https://github.com/nuxt/nuxt/pull/30190). When I switch it back to use RouterLink (like in earlier versions), smooth scrolling works correctly again. Here’s a demo with patch:
https://stackblitz.com/~/github.com/yuriystasiv/nuxt-link-hashtag-patch
https://github.com/yuriystasiv/nuxt-link-hashtag-patch

**Expected behavior**
Smooth scrolling should work with NuxtLink and hash links, even when it’s rendered as an anchor tag.

### Additional context

Is there something that needs to be handled differently in NuxtLink to make smooth scrolling work as expected when using anchor tags?

As a workaround, you can prevent the default behavior and use navigateTo or router.push manually, like this:
```
<NuxtLink
  to="#test"
  @click.prevent="navigateToSection('#test')"
  >example
</NuxtLink>
<script setup lang="ts">
const route = useRoute();
const navigateToSection = (section: string) => {
  navigateTo(`${route.path}${section}`);
};
</script>
```

### Logs

```shell-script

```

DamianGlowala:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32329

### 📚 Description

This allows for rendering a plain anchor tag without breaking smooth scroll behavior.

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -513,7 +513,9 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
           onClick: (event) => {
             event.preventDefault()
 
-            navigateTo(href.value)
+            return props.replace
+              ? router.replace(href.value)
+              : router.push(href.value)
           },
         }, slots.default?.())
       }
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -505,8 +505,17 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
           } satisfies NuxtLinkDefaultSlotProps<true>)
         }
 
-        // converts `""` to `null` to prevent the attribute from being added as empty (`href=""`)
-        return h('a', { ref: el, href: href.value || null, rel, target }, slots.default?.())
+        return h('a', {
+          ref: el,
+          href: href.value || null, // converts `""` to `null` to prevent the attribute from being added as empty (`href=""`)
+          rel,
+          target,
+          onClick: (event) => {
+            event.preventDefault()
+
+            navigateTo(href.value)
+          },
+        }, slots.default?.())
       }
     },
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -505,8 +505,17 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
-        return h('a', { ref: el, href: href.value || null, rel, target }, slots.default?.())
+        return h('a', {
+          ref: el,
+          href: href.value || null, // converts `""` to `null` to prevent the attribute from being added as empty (`href=""`)
+          rel,
+          target,
+          onClick: (event) => {
+            event.preventDefault()
+
+            navigateTo(href.value)
```
danielroe: should we just use scrollTo? is this any different from rendering a NuxtLink?

danielroe (approved):

KazariEX:
This change breaks navigation functionality of external links. As @coderabbitai says:

> onClick handler breaks external links, target-based navigation and modifier-key behaviour.

PR #32397, opened 2025-06-15
feat(kit): add `onInstall` and `onUpgrade` module hooks

### LINKED ISSUE #31070: module install hooks
modules should be able to run steps on first install, upgrade, etc. (and ideally track this in a version-controlled way in `.nuxtrc`):

`.nuxtrc`

```
setups.eslint=v3.15.2
```

romhml:
### 🔗 Linked issue
Resolves #31070 

### 📚 Description

Adds `onInstall` and `onUpgrade` module hooks to let modules perform additional setup steps when installed or upgraded. These hooks are defined in the module definition:
```ts
  meta: {
    name: 'my-module',
    version: '1.0.0',
  },

  onInstall () {
    // This will be run when the module is first installed.
  },

  onUpgrade () {
    // This will be run when the module is upgraded.
  }
})
```

Note that the `name` and `version` fields must be provided in the module's metadata for these hooks to be executed.

The implementation uses the `.nuxtrc` file internally to detect when hooks must be executed. It logs the module's version as `setups.my-module="1.0.0"` after a hook is executed, ensuring that the correct hook is triggered when the module is updated or installed.

Possible follow-ups:
- We could add a `onUninstall` hook which would be run when a module is removed (if it is still available).
- We could add support for `rc`  files into `@unjs/c12`'s [`updateConfig` function ](https://github.com/unjs/c12/blob/main/src/update.ts) to make the implementation here a bit more straightforward.
- The `nuxi module add` command could also be updated to run the `onInstall` hook.


packages/kit/src/module/install.ts
```diff
@@ -29,6 +31,24 @@ export async function installModule<
...
+
+  if (meta?.name && meta?.version) {
+    const rc = readRc({ dir: nuxt.options.rootDir, name: '.nuxtrc' })
+    const previousVersion = rc?.setups?.[meta.name]
+
+    if (!previousVersion) {
+      await nuxtModule?.onInstall?.(nuxt)
+    } else if (semver.gt(meta.version, previousVersion)) {
+      await nuxtModule?.onUpgrade?.(inlineOptions, nuxt, previousVersion)
+    }
```
HigherOrderLogic: Maybe try catching error here and tell the user what went wrong, then exit later?

PR #32413, opened 2025-06-16
fix(nuxt): normalize segment catchall pattern before checking for parent

### LINKED ISSUE #32365: Nested catch-all routes flattened instead of grouped as children after upgrading from Nuxt 3.15.4 to 3.17.5
### Environment

- Operating System: `Linux`
- Node Version:     `v22.13.1`
- Nuxt Version:     `3.17.4`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `npm@10.9.2`
- Builder:          `-`
- User Config:      `devtools`, `colorMode`, `compatibilityDate`, `experimental`, `routeRules`, `app`, `icon`, `modules`, `fonts`, `image`, `tailwindcss`, `gtm`, `router`, `components`, `srcDir`, `schemaOrg`, `runtimeConfig`, `build`, `nitro`, `vite`, `headlessui`, `imports`, `$production`
- Runtime Modules:  `./src/modules/logger/module`, `./src/modules/error-pages/module`, `floating-vue/nuxt@5.2.2`, `@nuxtjs/eslint-module@4.1.0`, `nuxt-headlessui@1.1.4`, `@vueuse/nuxt@10.9.0`, `nuxt-esi-tags@1.0.8`, `@nuxt/test-utils/module@3.14.2`, `radix-vue/nuxt@1.9.14`, `@nuxt/ui@2.21.0`, `nuxt-auth-utils@0.5.15`, `@zadigetvoltaire/nuxt-gtm@0.0.13`, `@nuxt/image@1.8.0`, `nuxt-schema-org@3.5.0`, `@nuxt/fonts@0.10.3`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-9xvsdsz9?file=pages%2F%5B...layout%5D%2Fpage-%5Bpage%5D.vue

I think it can be  easily  reproduced by having the following folder structure in `/app/pages/` : 
pages/
├── [...layout].vue
├── [...layout]/
│ ├── index.vue
│ └── page-[page].vue

### Describe the bug

TL;DR  With this [PR](https://github.com/nuxt/nuxt/pull/31450) a bug was introduced, and  nested catch-all routes are no logger properly grouped as children.

After upgrading from Nuxt 3.15.4 to 3.17.5, nested catch-all routes are no longer being grouped as children but are instead being flattened into separate top-level routes, breaking our existing URL structure.

I went in to more details [here](https://github.com/nuxt/nuxt/pull/31450#issuecomment-2945341080) and [here](https://github.com/nuxt/nuxt/pull/31450#issuecomment-2945601874). 

Let me know if you need more info.

### Additional context

_No response_

### Logs

```shell-script

```

Eazash:
### 🔗 Linked issue

resolves #32365

### 📚 Description

When a catch-all route is the last segment, it uses the greedy regexp (#31528). However, if it is used in middle of a path, it uses a more performant regexp (see #31450).

This change normalizes segment paths to the greedy regexp during parent checks to handle cases where both might be present (eg. if`[...id].vue` and `[...id]/[slug].vue` are pages) so that child routes are correctly nested.

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -144,7 +144,7 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
       // ex: parent.vue + parent/child.vue
       const routePath = getRoutePath(tokens, segments[i + 1] !== undefined && segments[i + 1] !== 'index')
       const path = withLeadingSlash(joinURL(route.path, routePath.replace(INDEX_PAGE_RE, '/')))
-      const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path)
+      const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path.replace('([^/]*)*', '(.*)*'))
 
       if (child && child.children) {
         parent = child.children
--- a/packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
+++ b/packages/nuxt/test/__snapshots__/pages-override-meta-disabled.test.ts.snap
@@ -157,10 +157,10 @@
       "children": [
         {
           "alias": "mockMeta?.alias || []",
-          "component": "() => import("pages/[...slug]/[foo].vue")",
+          "component": "() => import("pages/[...slug]/[id].vue")",
           "meta": "mockMeta || {}",
-          "name": "mockMeta?.name ?? "slug-foo"",
-          "path": "mockMeta?.path ?? ":foo()"",
+          "name": "mockMeta?.name ?? "slug-id"",
+          "path": "mockMeta?.path ?? ":id()"",
           "props": "mockMeta?.props ?? false",
           "redirect": "mockMeta?.redirect",
         },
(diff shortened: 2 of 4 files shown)


packages/nuxt/src/pages/utils.ts
```diff
@@ -144,7 +144,7 @@ export function generateRoutesFromFiles (files: ScannedFile[], options: Generate
       // ex: parent.vue + parent/child.vue
       const routePath = getRoutePath(tokens, segments[i + 1] !== undefined && segments[i + 1] !== 'index')
       const path = withLeadingSlash(joinURL(route.path, routePath.replace(INDEX_PAGE_RE, '/')))
-      const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path)
+      const child = parent.find(parentRoute => parentRoute.name === route.name && parentRoute.path === path.replace('([^/]*)*', '(.*)*'))
```
danielroe, line 147: perf: we can likely hoist the `path.replace` out of the arrow function to save it from having to be rerun as often.

danielroe (approved):

danielroe:
thank you ❤️

PR #32427, opened 2025-06-18
chore: mention `bun.lock` for lockfile

iiio2:
bun generates `bun.lock` by default from bun v1.2. That is why I think may be here we may add `bun.lock` with `bun.lockb` as well.
https://bun.sh/docs/install/lockfile


danielroe (approved):

iiio2:
Thanks @danielroe . :blush:

PR #32431, opened 2025-06-18
feat(kit,nuxt)!: defer installing modules with `installModule`

### LINKED ISSUE #31674: order-dependent modules
Ideally Nuxt modules should not be order-dependent but there are a number of situations where users or modules want to change the options of a module (or hook into its hooks).

I would advise that modules which expose one-time hooks should call those hooks in `modules:done` to ensure other modules have a chance to register callbacks. (And we need to document this pattern in the module author guide.)

However, there may be modules which want to modify the _options_ of another module, and this is a tricky pattern we're looking to solve.

We are considering:

- a module dependency graph
- tracking module options modification by other modules (and perhaps calling back to that module?)
- ...

However, we would really like feedback from the community and module authors on whether you have modules which _are_ dependent on order. I'd like to know which modules they are and what the use-cases are. This would be very helpful for us as we find a generic solution that will make for a more stable solution (and don't require users to specify modules in a certain order, which is very fragile as an approach).

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/31674

### 📚 Description

this is a draft PR to look at the effect of deferring modules registered with `installModule` until the end of the module install cycle. The kit utility is backwards compatible with previous versions of Nuxt (which won't set `_modulesToDefer`).

If `installedModule` is called more than once, it will merge all sets of options into resolved configuration - and, of course, respect the user's configuration.

(This will not solve the issue where a layer adds a module but another module in the user's project modifies the config or calls `installModule` itself.)


BobbieGoede:
@danielroe 
Here's a repo with an [attempt at demonstrating some issues](https://github.com/BobbieGoede/nuxt-i18n-starter/tree/nuxt-install-module-defer) 
* the module installation order differs from layer priority order (maybe intended?)
* nuxt-i18n has a `langDir` option which gets overwritten by each module installation, for layers we can retrieve it from `_layers`
* one way we try to work around `langDir` being lost is by forcing users to pass absolute paths for locale files instead

Here is a repo with a [variant of this setup using layers instead](https://github.com/BobbieGoede/nuxt-i18n-starter/tree/nuxt-install-module-defer-layer-order)
* shows the difference in merging order for the options passed to nuxt-i18n
* in nuxt-i18n we use `_layers` to merge things manually, we ignore relative file paths from the module options (and assume absolute paths are provided by `installModule`)

The setups can get more complicated when locales have file arrays instead of a single file path, it's hard to demonstrate but you can imagine that the merging/overwrite order is important in these setups.

BobbieGoede:
But to be clear, the changes in this PR are probably good for the majority of use cases

danielroe:
~> https://github.com/nuxt/nuxt/pull/33063

PR #32455, opened 2025-06-23
perf(nuxt): decrease if checks when prerendering

djixadin:

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts
@@ -248,6 +248,8 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
 
   // 6. Scripts
   if (!routeOptions.noScripts) {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/handlers/renderer.ts
```diff
@@ -248,14 +248,16 @@ export default defineRenderHandler(async (event): Promise<Partial<RenderResponse
 
   // 6. Scripts
   if (!routeOptions.noScripts) {
+    const tagPosition = (_PAYLOAD_EXTRACTION && !process.env.NUXT_JSON_PAYLOADS) ? 'bodyClose' : 'head'
```
danielroe, line 251: I think this is worth doing but there should not be any very significant benefit of doing so.

in production this is optimised to a constant, and when prerendering it's a simple ternary.

was there an issue you were hoping to solve, or did you just spot the possible optimisation?

danielroe (approved):

Issue #32109, opened 2025-05-14
`useFetch` with `immediate: false` causes watch to stop triggering in Nuxt 3.17.3


maximepvrt:
### Environment

nuxt: 3.17.3

### Reproduction

https://github.com/user-attachments/assets/c5839828-9408-49d7-90e4-72ae38d8015f

https://stackblitz.com/edit/github-ycxmi8dl-ou7o3zwv?file=pages%2Findex.vue

### Describe the bug

When `immediate: false` is set, the watch that should trigger the fetch stops working after a while. I haven't been able to fully isolate the cause, but it seems to happen when I change the watched value before the previous fetch has completed. After that, subsequent changes to the value don't trigger any new request.

### Logs

```shell-script

```

OrbisK:
With a hard coded key it does work: https://stackblitz.com/edit/github-ycxmi8dl-d6v8z7z3?file=pages%2Findex.vue

OrbisK:
Or removing compat version 4:

https://stackblitz.com/edit/github-ycxmi8dl-b3xjx9ee?file=pages%2Findex.vue

rafinhacuri:
> Ou removendo a versão compatível 4:
> 
> https://stackblitz.com/edit/github-ycxmi8dl-b3xjx9ee?file=pages%2Findex.vue

If I don't want to remove compatibility with nuxt 4, will it only work using the key?

OrbisK:
> > Or removing compat version 4:
> > https://stackblitz.com/edit/github-ycxmi8dl-b3xjx9ee?file=pages%2Findex.vue
> 
> If I don't want to remove compatibility with nuxt 4, will it only work using the key?

I think you can downgrade to 3.16.x, until this is fixed

maximepvrt:
@OrbisK Nuxt v4 should be released as a stable version by the end of the month. if it's possible to address the issue before then, that would be great.

OrbisK:
I think I have a fix for this!

OrbisK:
Another workaround: https://github.com/nuxt/nuxt/pull/32466#issuecomment-3001153087

maximepvrt:
@OrbisK Thank you for looking into this and for your work

maximepvrt:
@OrbisK @danielroe I've just tested with the new version of Nuxt 3.17.6.
Here is the reproduction code: https://stackblitz.com/edit/github-ycxmi8dl-uugn8jzs?file=package.json
And I'm seeing an error in the console.

[...]

PR #32464, opened 2025-06-24
WIP: test: reproduction of #32109

OrbisK:
### 📚 Description

This PR adds a reproduction runtime test. 

The bug only occurs when `immediate: false` and a parallel execute while the internal one is still pending.


OrbisK:
EDIT:
maybe these tests are not recreating this bug.

PR #32466, opened 2025-06-24
fix(nuxt): ensure newKey asyncData runs when oldKey asyncData is still running

### LINKED ISSUE #32109: `useFetch` with `immediate: false` causes watch to stop triggering in Nuxt 3.17.3

OrbisK:
### 🔗 Linked issue

fixes #32109

### 📚 Description

This PR should fix #32109. It ensures that the newKey asyncdata is **not** only executed once the request for the old key has finished.

I'll still mark it as a draft so that I can check whether there is a better solution.

~~TODO: add tests~~

> [!NOTE]
> It was not possible for me to reproduce this test as runtime or fixtures test. Only with full e2e test.


OrbisK:
Maybe this should be solved here instead:

https://github.com/nuxt/nuxt/blob/8668208e1c16da7e7514384d2be7ac3f5e7c9617/packages/nuxt/src/app/composables/fetch.ts#L144-L145

OrbisK:
Alternatively, setting `experimental.alwaysRunFetchOnKeyChange` to true does solve the issue too.

EDIT: now I wonder if this was intended https://nuxt.com/docs/4.x/api/nuxt-config#alwaysrunfetchonkeychange

test/e2e/composables.test.ts
```diff
@@ -0,0 +1,45 @@
+import { fileURLToPath } from 'node:url'
+import { isWindows } from 'std-env'
+import { join } from 'pathe'
+import { expect, test } from './test-utils'
```
danielroe: is there any chance we can use a nuxt runtime test rather than an e2e test?

huang-julien: @OrbisK you should be able to do it by mocking `_asyncPromises`. I can try to do it if you want

OrbisK: @huang-julien that would be nice

OrbisK: @huang-julien I have it! 💚

test/nuxt/composables.test.ts
```diff
@@ -923,6 +923,32 @@ describe('useFetch', () => {
     await nextTick()
     expect.soft(status.value).toBe('pending')
   })
+
+  it.skipIf(process.env.NUXT_LEGACY !== '1')('should handle parallel execute with `immediate: false`', async () => {
```
danielroe: I think this is not running at all at the moment. I updated the env variable before merging the other PR:

```suggestion
  it.runIf(process.env.PROJECT === 'nuxt-legacy')('should handle parallel execute with `immediate: false`', async () => {
```

OrbisK: even better :)

danielroe (approved):
thank you ❤

test/e2e/composables.test.ts
huang-julien: great :fire: !

PR #32473, opened 2025-06-25
test: add case for key only changes with `immediate: false`

OrbisK:
### 📚 Description

This PR introduces a small test for asyncData to track the `immediate: false` behaviour when key is changed, without altering any watch sources.

### DIFF
--- a/test/nuxt/composables.test.ts
+++ b/test/nuxt/composables.test.ts
@@ -743,6 +743,20 @@ describe('useAsyncData', () => {
     expect.soft(handler).toHaveBeenCalledTimes(1)
     expect.soft(getCachedData).toHaveBeenCalledTimes(1)
   })
+
+  it('should not execute if immediate is false and only the key changes', async () => {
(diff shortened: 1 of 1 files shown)


test/nuxt/composables.test.ts
```diff
@@ -898,6 +912,8 @@ describe('useFetch', () => {
     expect(status.value).toBe('error')
     expect(error.value).toMatchInlineSnapshot(`[Error: [GET] "[object Promise]": <no response> Failed to parse URL from [object Promise]]`)
   })
+
+  it('should refetch ')
```
danielroe:
```suggestion
```

OrbisK: I wanted to add the same test for useFetch, but since we are using `experimental.alwaysRunFetchOnKeyChange: true` for tests, so I had to test the opposite. 👍🏽

danielroe (approved):

PR #32481, opened 2025-06-26
test: separate nuxt legacy runtime tests

OrbisK:
### 📚 Description

This PR separates the nuxt runtime tests to be able to disable some v3/v4 related features.


vitest.config.ts
```diff
@@ -73,6 +73,35 @@ export default defineConfig({
...
+                },
+                experimental: {
+                  appManifest: process.env.TEST_MANIFEST !== 'manifest-off',
+                },
+                imports: {
+                  polyfills: false,
+                },
+              },
+            },
+          },
```
OrbisK: @danielroe please check here if you want to add/remove some features here

vitest.config.ts
```diff
@@ -73,6 +73,35 @@ export default defineConfig({
...
+          environmentOptions: {
+            nuxt: {
+              overrides: {
+                routeRules: {
+                  '/specific-prerendered': { prerender: true },
+                  '/pre/test': { redirect: '/' },
+                  '/pre/**': { prerender: true },
+                },
+                experimental: {
+                  appManifest: process.env.TEST_MANIFEST !== 'manifest-off',
```
danielroe: feels like maybe we could move the unit tests outside of the test matrix entirely, in favour of just creating an additional project (you can leave this with me if you want)

OrbisK: I think the runtime (nuxt, nuxt-universal, nuxt-legacy) tests are not even executed in the matrix ci tests. only fixtures and e2e.

OrbisK: maybe we should add "not legacy" tests here: https://github.com/nuxt/nuxt/blob/a3f3a8e52b294fb62292bb0cacfa3444a5347d41/.github/workflows/ci.yml#L189




EDIT: nvm done with [2b949dd](https://github.com/nuxt/nuxt/pull/32481/commits/2b949dd4ba456157772ed5d05af65c127313c463)

danielroe (approved):
looks great!

PR #32483, opened 2025-06-26
fix(nuxt): append `set-cookie` headers in error handler

### LINKED ISSUE #32472: Error page discards all but the last `Set-Cookie` header from response
### Environment

- Operating System: `Darwin`
- Node Version:     `v22.14.0`
- Nuxt Version:     `3.17.5`
- CLI Version:      `3.25.1`
- Nitro Version:    `2.11.12`
- Package Manager:  `pnpm@10.11.0`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Couldn't do it in StackBlitz because it's a cookie thing but here is a minimal reproduction to run locally: [https://github.com/pgraug/nuxt-repro-forgets-set-cookie-on-error](https://github.com/pgraug/nuxt-repro-forgets-set-cookie-on-error)

### Describe the bug

When setting multiple cookies during SSR and encountering an error, only the last cookie you set will be included in the response headers instead of one `set-cookie` header per cookie.

This is due to one use of `setResponseHeader` in Nuxt's Nitro error handler ([permalink to line](https://github.com/nuxt/nuxt/blob/d9e57c6839beda91d99b78782fb043aef060356e/packages/nuxt/src/core/runtime/nitro/handlers/error.ts#L77)) that overwrites the previous `set-cookie` headers each time it sees a new one.

I propose two possible solutions:
- Using `appendResponseHeader` inside the loop instead since it handles multiple cookies with the same by converting them to an array. Only thing I'm not sure about with this is whether there are some unintended side-effects like `content-type` becoming an array as well because the response already had one.
- Doing an `Array.reduce` on `Array.from(res.headers.entries())` to group headers of the same name into an array and then using `setResponseHeaders` to set them all at once. This should be safe from any unintended side-effects but is more complex.

The first solution seems the simplest but I would love your thoughts on whether it would have any unintended consequences. I can make the PR for it if you're okay with my solution.

### Additional context

_No response_

### Logs

```shell-script

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32472

### 📚 Description

this uses `appendResponseHeader` so set-cookie can be merged (by browser).

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/handlers/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/handlers/error.ts
@@ -1,6 +1,6 @@
 import { joinURL, withQuery, withoutBase } from 'ufo'
 import type { NitroErrorHandler } from 'nitropack/types'
-import { getRequestHeaders, send, setResponseHeader, setResponseHeaders, setResponseStatus } from 'h3'
+import { appendResponseHeader, getRequestHeaders, send, setResponseHeader, setResponseHeaders, setResponseStatus } from 'h3'
 
 import { useNitroApp, useRuntimeConfig } from 'nitropack/runtime'
 import { isJsonRequest } from '../utils/error'
@@ -74,6 +74,10 @@ export default <NitroErrorHandler> async function errorhandler (error, event, { 
 
   const html = await res.text()
   for (const [header, value] of res.headers.entries()) {
+    if (header === 'set-cookie') {
+      appendResponseHeader(event, header, value)
+      continue
+    }
     setResponseHeader(event, header, value)
   }
   setResponseStatus(event, res.status && res.status !== 200 ? res.status : defaultRes.status, res.statusText || defaultRes.statusText)


packages/nuxt/src/core/runtime/nitro/handlers/error.ts
```diff
@@ -74,6 +74,10 @@ export default <NitroErrorHandler> async function errorhandler (error, event, {
 
   const html = await res.text()
   for (const [header, value] of res.headers.entries()) {
+    if (header === 'set-cookie') {
+      appendResponseHeader(event, header, value)
```
pi0, line 78: Parse + using [setCookie](https://github.com/h3js/h3/blob/dab856757c4cc92b6dfe6e02e7e80119f5e4327a/src/utils/cookie.ts#L45) might be much safer for merging.

(append does not merges but sends duplicate headers, clients/browser should respect and merge)

danielroe: this is an already-serialised 'set-cookie' header, not a raw value. what we're doing here is very similar to the mergeHeaders function in h3:

https://github.com/unjs/h3/blob/67af0575eb2078ba4c2ed1c535dfc168a4e578e6/src/response.ts#L101-L111

danielroe: the issue is that we want to _preserve_ the existing headers, as they are, rather than flattening them into a single `set-cookie` header.

merging to iterate, but let me know if you have any concerns 🙏

PR #32497, opened 2025-06-29
docs:  bunx command example typo in Upgrading Nuxt page

cram0:
### 📚 Description

There's an extra space in the bunx command example to upgrade Nuxt : 

`bun x` -> `bunx`


danielroe:
It's not a typo.

CleanShot 2025-06-29 at 12 08 51

> https://bun.sh/docs/cli/bunx

PR #32503, opened 2025-06-30
docs: adjust wording to reduce confusion in lifecycle section

### LINKED ISSUE #32303: Nuxt/Vue lifecycle data load mismatch in layout and page components
### Environment

Nuxt version 3.17.5

### Reproduction

Playground: https://stackblitz.com/edit/layout-page-hydrate-test

There you can see the hydration problem, and also visually you can see how the data in the \<pre\> tag is flickering. This is precisely because on the server the data `settings`, when the TheHeader component is executed, is empty.

Image

### Describe the bug

The problem is that I have a layout that contains TheHeader component. However, the request to get data for the header theme is in the page component. And it turns out that the data for the header is not substituted into TheHeader component, which is in the layout.

I know that Vue  _compute_ computed properties only once, when it encounters them for the first time on server.
However, the [Nuxt Lifecycle](https://nuxt.com/docs/guide/concepts/nuxt-lifecycle#step-3-initialize-nuxt-and-execute-nuxt-app-plugins) documentation says that in step 6, Nuxt initializes all components and makes a request to get data for display and only then, in step 7, the entire page is rendered.

It seems that the layout execution happens somewhere between step 5 and 6.

Can I somehow change the logic using Nuxt Lifecycle to force Nuxt to re-render TheHeader component only when the page data has already loaded via page component?

I can't execute the request to load the page data somewhere externally, in the layout or middleware.
Putting TheHeader component in each of the pages is also not a solution.

### Additional context

_No response_

### Logs

```shell-script

```

Mini-ghost:
### 🔗 Linked issue

Resolve #32303

### 📚 Description

This PR clarifies the wording in the Nuxt lifecycle documentation to make it easier to understand and reduce potential misunderstandings.

If there’s anything I may have overlooked, please feel free to let me know. Thank you very much!

### DIFF
--- a/docs/2.guide/1.concepts/10.nuxt-lifecycle.md
+++ b/docs/2.guide/1.concepts/10.nuxt-lifecycle.md
@@ -76,17 +76,27 @@ Any redirection on the server will result in a `Location:` header being sent to 
 
 :read-more{to="/docs/guide/directory-structure/middleware"}
 
-### Step 6: Setup Page and Components
+### Step 6: Render Page and Components
 
-Nuxt initializes the page and its components during this step and fetches any required data with `useFetch` and `useAsyncData`. Since there are no dynamic updates and no DOM operations occur on the server, Vue lifecycle hooks such as `onBeforeMount`, `onMounted`, and subsequent hooks are **NOT** executed during SSR.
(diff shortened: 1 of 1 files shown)


danielroe (approved):

danielroe:
thank you ❤️

Issue #32494, opened 2025-06-27
Nuxt 3.16.0 and 3.17.0 build error: `"appSpaLoaderAttrs" is not exported by virtual:nuxt`


gabrielduete:
Hi everyone! Hope you're doing well 

I'm currently working on updating project dependencies due to security alerts, and I'm trying to upgrade Nuxt to version 3.16.0 or 3.17.0. However, I'm running into issues during the build process.

❗ Problem with version 3.16.0+

When trying to build the project after upgrading to Nuxt 3.16.0+, I get this error:

[nuxi]  ERROR  Nuxt Build Error: ../../node_modules/nuxt/dist/app/entry.js (8:16): "appSpaLoaderAttrs" is not exported by "virtual:nuxt:/.../.nuxt/nuxt.config.mjs", imported by "../../node_modules/nuxt/dist/app/entry.js"

⚠️ Problem with version 3.17.0

When updating to 3.17.0, I get a different error, which seems to be solvable with a workaround — but the original error from 3.16.0 still happens afterwards:

ERROR  nuxt.options.experimental.extraPageMetaExtractionKeys is not iterable

Temporary workaround:
We can bypass the above by adding the following to our Nuxt config:

```
experimental: {
  extraPageMetaExtractionKeys: []
}
```

However, after that, the build still fails with the same appSpaLoaderAttrs error from 3.16.0.

🔎 What we’ve tried
- Updated only nuxt, or all related @nuxt/* libraries to 3.16.0 / 3.17.0
- Tried both versions on a branch that uses an updated @nx/nuxt (18.3.5)
- Tried installing or updating the vite builder
- Cleaned .nuxt, node_modules, yarn.lock, reinstalled everything
- Tried both locally and in CI (same result)

We’re using Nx monorepo with @nx/nuxt, and this project includes multiple apps and shared libs.

💭 Suspicions

I believe this may be due to internal conflicts between different versions of @nuxt/kit or @nuxt/schema, especially with how Nx wraps the Nuxt config and injects the builder.

We've seen @nuxt/kit@3.15.4 and @nuxt/schema@3.14.x being pulled transitively even when we explicitly define 3.17.5 — possibly by @nuxt/test-utils, @nx/nuxt, or vitest-environment-nuxt.

🙏 What I need help with

- Has anyone else experienced this "appSpaLoaderAttrs" virtual module error when using Nuxt with Nx?
- Is this a known incompatibility with @nx/nuxt and Nuxt 3.16+?
[...]

PR #32510, opened 2025-06-30
fix(nuxt): safe-guard `extraPageMetaExtractionKeys`

s1gr1d:
### 🔗 Linked issue

This issue came up while checking the canary tests in Sentry Nuxt: https://github.com/getsentry/sentry-javascript/issues/16762

Building the application resulted in this error: `nuxt.options.experimental.extraPageMetaExtractionKeys is not iterable`

### 📚 Description

This PR creates a safeguard (`Array.isArray()`) before spreading the experimental value.

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -489,7 +489,12 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extractedKeys = [...defaultExtractionKeys, 'middleware', ...nuxt.options.experimental.extraPageMetaExtractionKeys]
+    const extraPageMetaExtractionKeys = nuxt.options.experimental.extraPageMetaExtractionKeys
+    const extractedKeys = [
+      ...defaultExtractionKeys,
+      'middleware',
+      ...(Array.isArray(extraPageMetaExtractionKeys) ? extraPageMetaExtractionKeys : []),
+    ]
 
     nuxt.hook('modules:done', () => {
       addBuildPlugin(PageMetaPlugin({
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -489,7 +489,7 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extraPageMetaExtractionKeys = nuxt.options.experimental.extraPageMetaExtractionKeys
+    const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys
     const extractedKeys = [
       ...defaultExtractionKeys,
       'middleware',
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -489,11 +489,10 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys
+    const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys || []
     const extractedKeys = [
       ...defaultExtractionKeys,
-      'middleware',
-      ...(Array.isArray(extraPageMetaExtractionKeys) ? extraPageMetaExtractionKeys : []),
+      ...extraPageMetaExtractionKeys,
     ]
 
     nuxt.hook('modules:done', () => {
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/pages/module.ts
```diff
@@ -489,7 +489,12 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extractedKeys = [...defaultExtractionKeys, 'middleware', ...nuxt.options.experimental.extraPageMetaExtractionKeys]
+    const extraPageMetaExtractionKeys = nuxt.options.experimental.extraPageMetaExtractionKeys
+    const extractedKeys = [
+      ...defaultExtractionKeys,
+      'middleware',
+      ...(Array.isArray(extraPageMetaExtractionKeys) ? extraPageMetaExtractionKeys : []),
```
Mini-ghost: I was wondering if it might be recommended to use `toArray` here?
https://github.com/nuxt/nuxt/blob/main/packages/nuxt/src/app/utils.ts

s1gr1d: Wouldn't this result to `[undefined]` if there is no value set? I'm not sure if this is wanted 🤔

Mini-ghost: You’re absolutely right! I overlooked this part.

packages/nuxt/src/pages/module.ts
```diff
@@ -489,7 +489,12 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extractedKeys = [...defaultExtractionKeys, 'middleware', ...nuxt.options.experimental.extraPageMetaExtractionKeys]
+    const extraPageMetaExtractionKeys = nuxt.options.experimental.extraPageMetaExtractionKeys
+    const extractedKeys = [
+      ...defaultExtractionKeys,
+      'middleware',
```
xjccc: emmm.. `middleware` is also in `defaultExtractionKeys`, does it need to remove this line?

s1gr1d: True, does it make sense to use `defaultExtractionKeys` here as well?
It only contains `'middleware'`
https://github.com/nuxt/nuxt/pull/32510/files#diff-c2594c96f1f7e8279fb8940560db036a9efa49e39522509ee02193dbac213092R77

xjccc: Does it seem that it maybe could also be deleted? But it doesn't matter with this.

https://github.com/nuxt/nuxt/blob/d39f540132da9b6fa4444006b8c2642278df0c4d/packages/nuxt/src/pages/utils.ts#L241

danielroe:
This is a nice check, but also highlights an issue with the wrong version of `@nuxt/schema` being hoisted into projects (e.g. https://github.com/nuxt/nuxt/issues/32494).

I have plans to fix/improve that (but for a start, `@nuxt/schema` should not be directly depended upon by libraries - this case, two core nuxt modules.)

packages/nuxt/src/pages/utils.ts
```diff
@@ -70,8 +70,13 @@ export async function resolvePagesRoutes (pattern: string | string[], nuxt = use
     return pages
   }
 
+  const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys
+
   const augmentCtx = {
-    extraExtractionKeys: new Set(['middleware', ...nuxt.options.experimental.extraPageMetaExtractionKeys]),
+    extraExtractionKeys: new Set([
+      'middleware',
```
s1gr1d, line 77: @danielroe maybe you know if `defaultExtractionKeys` can be used here instead of just adding `'middleware'`? Came up [here](https://github.com/nuxt/nuxt/pull/32510/files#r2176243781).

danielroe: No, it can't - we treat them slightly separately in terms of whether we remove them from the route metadata that's accessible at runtime from `route.meta`.

gabrielduete (approved):

gabrielduete (approved):
Awesome!!

packages/nuxt/src/pages/module.ts
```diff
@@ -489,7 +489,12 @@ export default defineNuxtModule({
     }
 
     // Extract macros from pages
-    const extractedKeys = [...defaultExtractionKeys, 'middleware', ...nuxt.options.experimental.extraPageMetaExtractionKeys]
+    const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys
```
danielroe: I think we could probably just do:

```suggestion
    const extraPageMetaExtractionKeys = nuxt.options?.experimental?.extraPageMetaExtractionKeys || []
```

... as the issue is presumably that it's not defined, rather than set to a string or something like that

xjccc (approved):

danielroe (approved):

danielroe:
thank you ❤️

PR #32522, opened 2025-07-02
docs: update fetch type

xjccc:
### 📚 Description

Update type of `useFetch` & `useAsyncData`


docs/3.api/2.composables/use-async-data.md
```diff
@@ -194,11 +195,12 @@ type AsyncDataRequestContext = {
 }
 
 type AsyncData<DataT, ErrorT> = {
-  data: Ref<DataT | null>
+  data: Ref<DataT | undefined>
+  pending: Ref<boolean>
```
OrbisK: I think we soft deprecated pending. (IMO thats why it is missing)

xjccc: Thank you for your explanation.❤

danielroe (approved):

PR #32531, opened 2025-07-03
feat(nuxt): add abortController option to `useAsyncData`

### LINKED ISSUE #32530: `clear` does not cancel `useFetch` requests
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     3.12.3
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Builder:          -
- User Config:      future, devtools, compatibilityDate, modules, experimental
- Runtime Modules:  @nuxt/eslint@0.3.13, @nuxt/ui@2.17.0, @vueuse/nuxt@10.11.0
- Build Modules:    -

### Reproduction

Stackblitz to reproduce the issue

https://stackblitz.com/edit/nuxt-starter-tfaddtjt?file=app%2Fapp.vue

### Describe the bug

Upon using the `clear` function after executing an `useFetch` request, the pending request does not get cancelled.

### Additional context

This can be checked in the network tab of the issues.

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

fix #32530

### 📚 Description

I don't really find this PR satisfying as we need to overwrite `clear` from `asyncData` in `useFetch` 🤔 should we instead add a `onClear` in `useAsyncData` ?


danielroe:
maybe we can add controller as an option to useAsyncData instead, or implement in useAsyncData rather than useFetch?

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -361,7 +365,10 @@ export function useAsyncData<
     error: writableComputedRef(() => nuxtApp._asyncData[key.value]?.error as Ref<NuxtErrorDataT extends Error | NuxtError ? NuxtErrorDataT : NuxtError<NuxtErrorDataT>>),
     refresh: (...args) => nuxtApp._asyncData[key.value]!.execute(...args),
     execute: (...args) => nuxtApp._asyncData[key.value]!.execute(...args),
-    clear: () => clearNuxtDataByKey(nuxtApp, key.value),
+    clear: () => {
+      options.abortController?.abort?.(new DOMException('Request aborted as the async data was cleared.', 'AbortError'))
```
danielroe: I think if an abortController is triggered, then we should also track this and avoid doing anything with the finished result of an asyncData. so it shouldn't update the data/status, just as if it were cancelled.

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -361,7 +365,15 @@ export function useAsyncData<
...
     refresh: (...args) => nuxtApp._asyncData[key.value]!.execute(...args),
     execute: (...args) => nuxtApp._asyncData[key.value]!.execute(...args),
-    clear: () => clearNuxtDataByKey(nuxtApp, key.value),
+    clear: () => {
+      if (options.abortController) {
+        options.abortController.abort(new DOMException('Request aborted as the async data was cleared.', 'AbortError'))
+      }
+      if (nuxtApp._asyncDataPromises[key.value]) {
+        (nuxtApp._asyncDataPromises[key.value] as any).cancelled = true
+      }
```
OrbisK: I'm just asking, but don't we need something like this, too?

```ts
options.abortController.signal.addEventListener('abort', () => {
  if (nuxtApp._asyncDataPromises[key.value]) {
        (nuxtApp._asyncDataPromises[key.value] as any).cancelled = true
      }
})
```

What if someone cancels the signal from outside?

huang-julien: make sense

danielroe:
would you mind checking the test failures?

OrbisK:
> would you mind checking the test failures?

https://github.com/nuxt/nuxt/pull/33083 should fix it. Maybe you can have a look :)

OrbisK:
Ohhh. Tests failing too 🤔

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -687,16 +747,21 @@ function createAsyncData<
           asyncData.status.value = 'success'
         })
         .catch((error: any) => {
-        // If this request is cancelled, resolve to the latest request.
-          if ((promise as any).cancelled) { return nuxtApp._asyncDataPromises[key] }
+          // If the promise was replaced by another one, we do not update the asyncData
+          if (nuxtApp._asyncDataPromises[key] && nuxtApp._asyncDataPromises[key] !== promise) {
+            return promise
```
OrbisK:
```suggestion
            return
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -687,16 +747,21 @@ function createAsyncData<
...
+          }
+
+          // If the asyncData was explicitly aborted internally (dedupe or clear), we do not update the asyncData
+          if (asyncData._abortController?.signal.aborted) {
+            return promise
```
OrbisK:
```suggestion
            return
```

OrbisK (commented):
@danielroe this solves the test issue. not sure why I retured the promise there.

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -112,6 +118,10 @@ export interface AsyncDataExecuteOptions {
 
   /** @internal */
   cachedData?: any
+
+  signal?: AbortSignal
+
+  timeout?: number
```
danielroe, line 123: we should probably not add additional options to `execute` - what about passing an AbortController when creating useAsyncData in the first place?

(btw, there are lots of places we call `execute` at the moment, where we do not proxy the signal/timeout from the parent options)

OrbisK: >  what about passing an AbortController when creating useAsyncData in the first place?

I think this was @huang-julien s initial PR.

In my opinion, this does not feel right + once the AbortController signal is aborted, it cannot be restored without recreating the controller. (Maybe i am wrong?)

> we should probably not add additional options to execute

The `timeout` and `signal` here are optional + they are merged with the options from `asyncData`. Having a `signal` for `execute()` makes sense to me as it behaves more like `fetch(url, { signal })`. 

> (btw, there are lots of places we call execute at the moment, where we do not proxy the signal/timeout from the parent options)

This is not needed for existing implementations. However, it does allow for usage like this.

```ts
const {execute: execute1} = useAsyncData()
const {execute: execute2} = useAsyncData()

function abortableFoo(signal){
  execute1({signal})
  const p1 = execute1({signal}) // this will dedupe the first execute1()
  const p2 = execute2({signal})
  const p3 = axios(url, {signal})
  const p4 = barAbortablePromise({signal}) 


  await Promise.all([p1,p2,p3,p4])
}
const controller = AbortController()

function handleUpdate() {
  abortableFoo(controller.signal)
}

function handleClick(){
  controller.abort()
}
```

huang-julien: > AbortController signal is aborted, it cannot be restored without recreating the controller. (Maybe i am wrong?)

:+1: Yes once an AbortController is aborted, we need another one. That's why we need to have one per handler call

bwshnaftjmal2025:
تحياتي للجميع بالتوفيق ان شاء الله

PR #32534, opened 2025-07-05
docs: improve Nuxt doc

iiio2:


huang-julien (commented):
Capitalization are done on prupose for theses titles

danielroe:
@iiio2 I think it might be more helpful to refine the linting so we don't need to have individual PRs addressing case. Do you think perhaps (instead) you could investigate the markdownlint settings and see why the headings aren't being capitalised consistently? 🙏

iiio2:
@danielroe , I have done this work for markdown lint. Can you check it ?

PR #32547, opened 2025-07-07
feat(nuxt): add `<NuxtTime>` export props interface

### LINKED ISSUE #32546: Expose `<NuxtTime>` props
### Describe the feature

I want to make a wrapper component over Next Time, I don't want to copy all the Props. 

Exporting props would be a good idea. For example, as in NuxtLink

### Additional information

### Final checks

hywax:
### 🔗 Linked issue

Resolves #32546

### 📚 Description

The props export has been added

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "test:e2e:ui": "playwright test --ui",
     "test:e2e:dev": "TEST_ENV=dev playwright test",
     "test:e2e:webpack": "TEST_BUILDER=webpack playwright test",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "vue-tsc --noEmit",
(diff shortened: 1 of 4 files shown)


danielroe (approved):

hywax:
@danielroe there is currently an issue (which caused the CI to fail): it’s not possible to import types from `.vue` files into `.ts` files.

I can refactor the component to use a render function, which would avoid this problem. However, I think it would be better to configure the project to allow importing from `.vue` files, as this might be needed for other components in the future.

danielroe:
Yes, let's configure it to support importing from vue files, if you're happy to tackle that. It might require running `vue-tsc` instead of `tsc` as a type-checking step.

hywax:
@danielroe done, migrated to `vue-tsc`

danielroe:
thank you!

Issue #26565, opened 2024-03-29
Failed to fetch dynamically imported module "/_nuxt/OpeningHours.CTqYYKGj.js"


mikeCrafted:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v21.6.1
- Nuxt Version:     3.11.1
- CLI Version:      3.11.1
- Nitro Version:    2.9.5
- Package Manager:  npm@9.6.4
- Builder:          -
- User Config:      runtimeConfig, sourcemap, app, css, modules, i18n, alias, auth, build, site, schemaOrg, ogImage, sitemap, devtools      
- Runtime Modules:  @sidebase/nuxt-auth@0.6.7, @nuxtjs/i18n@8.2.0, @pinia/nuxt@0.5.1, nuxt3-leaflet@1.0.12, @nuxtjs/seo@2.0.0-rc.10
- Build Modules:    -
------------------------------

### Reproduction

This is an indeterministic bug in production, I cannot find commonalities when it occurs. Apple users (Safari browser) tend to have it more often. I get a lot of sentry errors and see it in the server logs, but mixed with a lot of correct requests, it works on all my devices and on ones of my colleagues, but according to sentry there are users who experience crashes.

### Describe the bug

According to our sentry report, there are regular crashes of the app on many different pages (sometimes on the homepage, sometimes during search etc.), when nuxt fails to resolve a file path and the app crashes. 

The majority of logs are fine, but still we see a regular issues popping up in sentry. The components that fail to resolve are also different. They work on all my devices, so I cannot reproduce the error in production. Nor does it occur in development. 

We host the app on Azure App Service.

I recently upgraded to the latest nuxt version, where the naming of the chunks was changed, still I saw some logs today that a component with the old naming style could not be imported. I am not opting in for the old one, I switched to the default new one.

Can setting the `builder: 'webpack'` help?

I feel like this issue might be related @danielroe ? #21780 

![nuxt-error](https://github.com/nuxt/nuxt/assets/63555352/94b45b55-6fe8-409c-9cca-83e051dc0c46)

danielroe:
I suspect this is due to new deployments and failing to load a dynamic chunk after a new deployment occurs, which might have invalidating existing public assets. We currently handle chunk loading errors when navigating across routes but we could probably think about whether there is anything we could do when that is a dynamic _component_ import.

mikeCrafted:
[...]

PR #32563, opened 2025-07-08
Set automatic immediate as default

MikeBellika:
### 🔗 Linked issue

Had a quick discussion in this PR about it: https://github.com/nuxt/nuxt/pull/28160#issuecomment-3037036793
Should hopefully also help with https://github.com/nuxt/nuxt/issues/26565

### 📚 Description

When using dynamically imported components, users risk getting chunk loading errors when the app is updated. With this change, the app is reloaded when any chunk errors happen.

This could potentially cause a unwanted reload if the component isn't needed for rendering the page. However I still think this is preferable to breaking the app, which can happen if the component is needed. 

At work we’ve seen the same error even with non‑lazy components pulled from other packages in the monorepo and `'automatic-immediate'` seems to have fixed it.

### DIFF
--- a/docs/2.guide/3.going-further/1.experimental-features.md
+++ b/docs/2.guide/3.going-further/1.experimental-features.md
@@ -59,16 +59,18 @@ This feature will likely be removed in a near future.
 
 ## emitRouteChunkError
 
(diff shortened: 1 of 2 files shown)


danielroe:
I think we won't make this the default - it is a destructive action to automatically reload the app, and can lose user state, which is why we want people to opt-in.

PR #32578, opened 2025-07-10
perf: optimize git repository size

productdevbook:
## Summary
- Applied git repository optimization techniques to reduce repository size
- Used `git gc --aggressive --prune=now` and `git repack -Ad` commands
- Achieved approximately 42% reduction in repository size (from 120MB to 70MB)
- All commit history is preserved
- Added documentation for future maintenance

## Test plan
- [ ] Verify repository functionality after optimization
- [ ] Confirm all git operations work correctly
- [ ] Test clone performance improvements


danielroe:
I think this probably doesn't need to be in the Nuxt repo - those are good commands you can indeed run locally, however.

PR #32587, opened 2025-07-11
docs: provide `async` keyword

iiio2:
since we are using `await $fetch`, I think the function should be `async` also.


TheAlexLichter (approved):

danielroe (approved):

iiio2:
Thanks @danielroe , @TheAlexLichter :blush:

PR #32592, opened 2025-07-11
fix(kit): add layer `app/` and `server/` folders into tsconfigs

### LINKED ISSUE #32590: [v4] New tsconfig.json style: Importing props type in layer component via layer alias fails in [@vue/compiler-sfc]
### Environment

```
Working directory: /home/projects/github-upm6owvq                        nuxi 19:24:52
Nuxt project info:                                                       nuxi 19:24:52

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     4.0.0-rc.0
- CLI Version:      3.25.1
- Nitro Version:    2.11.13
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -
```

### Reproduction

https://stackblitz.com/edit/github-upm6owvq

### Describe the bug

Hello,
I am currently trying out the v4.0.0 RC.0 and noticed an issue when using the new tsconfig.json structure (using references as described [here](https://nuxt.com/docs/4.x/guide/directory-structure/tsconfig)).

### Setup
- Define a component in a Nuxt layer folder
- Import a type from another file of that layer via the autogenerated layer alias (e.g. `#layers/my-layer/composables/my-composable`)
- Use that type as the type for the props of that component

If I now use that component (see `TestLayerComponent3` in the Stackblitz Repro) in my app, Nuxt breaks with:
```
[@vue/compiler-sfc] Failed to resolve import source "#layers/test-layer/composables/sub-folder/use-test-props"
/home/projects/github-upm6owvq/layers/test-layer/components/TestLayerComponent3.vue
import type { TestProps } from '#layers/test-layer/composables/sub-folder/use-test-props';
defineProps<TestProps>();
^^^^^^^^^ 
</script>
at new H3Error (/home/projects/github-upm6owvq/node_modules/h3/dist/index.mjs:170:5) at Module.createError (/home/projects/github-upm6owvq/node_modules/h3/dist/index.mjs:196:15) at eval (/home/projects/github-upm6owvq/node_modules/@nuxt/vite-builder/dist/index.mjs:420:25) at async processMessage (/home/projects/github-upm6owvq/node_modules/@nuxt/vite-builder/dist/index.mjs:403:30)
```

### Further Info
It only breaks in this exact setup. The following similar cases all work for me:
- Importing the props type from that layer in a component of the main /app folder (see `TestComponent`)
- Using a relative import path to import the type inside the layer (see `TestLayerComponent1`)
- Importing something else - e.g. a composable - via the layer alias inside the layer (see `TestLayerComponent2`)

Using the old tsconfig.json structure with `"extends": "./.nuxt/tsconfig.json"` also fixes the problem.

### Additional context

_No response_

### Logs

```shell-script

```

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32590

### 📚 Description

this ensures that vue can correctly resolve types in `layers/*/app/` folders.

note that this hard-codes the need to use the `app/` directory within layers, if types are going to work seamlessly.

### DIFF
--- a/packages/kit/src/template.ts
+++ b/packages/kit/src/template.ts
@@ -166,10 +166,12 @@ export function resolveLayerPaths (dir: Nuxt['options']['dir'], buildDir: string
     nuxt: [
       join(relativeSrcDir, '**/*'),
       join(relativeModulesDir, `*/runtime/**/*`),
+      join(relativeRootDir, `layers/*/app/**/*`),
       join(relativeRootDir, `layers/*/modules/*/runtime/**/*`),
     ],
     nitro: [
       join(relativeModulesDir, `*/runtime/server/**/*`),
+      join(relativeRootDir, `layers/*/server/**/*`),
       join(relativeRootDir, `layers/*/modules/*/runtime/server/**/*`),
     ],
     node: [
--- a/packages/kit/test/generate-types.spec.ts
+++ b/packages/kit/test/generate-types.spec.ts
@@ -129,6 +129,7 @@ describe('resolveLayerPaths', () => {
         ],
         "nitro": [
           "../custom-modules/*/runtime/server/**/*",
+          "../layers/*/server/**/*",
           "../layers/*/modules/*/runtime/server/**/*",
         ],
         "node": [
@@ -142,6 +143,7 @@ describe('resolveLayerPaths', () => {
         "nuxt": [
           "../app/**/*",
           "../custom-modules/*/runtime/**/*",
+          "../layers/*/app/**/*",
           "../layers/*/modules/*/runtime/**/*",
         ],
         "shared": [


yschroe:
Thanks for the quick fix. 👍
Is there a possibility to get this to work without the need for the `/app` subfolder inside the layer or is this a limitation of TypeScript? For DDD it is quite nice to use nuxt layers to group feature code together and separate it from other layers. The forced `/app` subfolder would add a little bit of bloat to the folder structure and import paths.

Or is the recommendation to just use the old tsconfig format in this case?

danielroe:
i did struggle with that and would be happy to rethink. here's my worry:

- we can of course exclude server/ and modules/ folders, but what about other folders, like content/ or i18n/? they would wrongly think we could auto import vue composables there, for example
- we could try to _detect_ the folder structure of each layer but this requires fs calls (and a server restart when new layers are added)

i'm thinking a better pattern would be to allow module authors or users to include files in particular contexts via config, so you could add your own pattern easily

wdyt?

yschroe:
What exactly controls where the auto imports are available? Is there some kind of whitelist for allowed folders in the nuxt internals? Or is this what's controlled via the "include" property in the tsconfig files? Because I think it would only be a problem if there is a mismatch in what the user sees in their IDE and what is actually happening when they run the project. If they want to use a composable in the `/content` folder and nuxt makes it work internally, would that be a problem? Or am I misunderstanding the problem?

Also I have been trying to manually test the changes you did in this PR to see if this fixes the problem and I can't get it to work. But maybe I am doing something wrong. If I keep the folder structure of the repro in the original issue and manually add `"../layers/**/*"` (as a more general glob) to the "include" array in `tsconfig.app.json` after the file is generated, should this work? Because I still see the same error.

Do you know why in the repro the `TestLayerComponent2` which imports a _composable_ via the layer alias works but `TestLayerComponent3` which imports a _type_ from the same file breaks? I would have assumed that either nothing or everything works in that folder based on the tsconfig. 🤔

danielroe:
@yschroe Auto-imports are available everywhere in your project. The issue isn't related to auto-imports, it's how the vue compiler knows how to resolve aliases specifically for _props_, and it does so via tsconfig. (Everywhere else aliases are resolved via vite, which doesn't care about your tsconfig.)

yschroe:
Ah, so aliases for props and aliases for other data structures are handled differently? This explains the different behaviour for the different test components in the repro, thanks for the explanation.

My question regarding auto imports was directed at what you wrote in your first bullet point ("they would wrongly think we could auto import vue composables there, for example"). If auto imports are available everywhere, wouldn't it be correct to signal this to the custom folders like content, i18n as well (even if it maybe doesn't make sense to use composables etc there)?

Lastly I have another question regarding the original issue, if you've got the time. 😅 I still don't fully understand why the props import/resolution works when the original v3-style tsconfig is used since in there the `layers` root folder is also not part of the "include" array. Is this due to using "extends" vs. "references"?

PR #32598, opened 2025-07-13
fix(nuxt): remove `assets` alias for all resolvers

OrbisK:
### 🔗 Linked issue

fixes https://github.com/nuxt/nuxt/issues/32657

### 📚 Description

This PR follows up on https://github.com/nuxt/nuxt/pull/32119 and is wating for https://github.com/vuejs/core/pull/13462

### DIFF
--- a/packages/schema/src/config/app.ts
+++ b/packages/schema/src/config/app.ts
@@ -7,11 +7,14 @@ import type { NuxtAppConfig } from '../types/config'
 export default defineResolvers({
   vue: {
     transformAssetUrls: {
-      video: ['src', 'poster'],
-      source: ['src'],
-      img: ['src'],
-      image: ['xlink:href', 'href'],
-      use: ['xlink:href', 'href'],
+      preserveTilde: true,
+      tags: {
+        video: ['src', 'poster'],
+        source: ['src'],
+        img: ['src'],
+        image: ['xlink:href', 'href'],
+        use: ['xlink:href', 'href'],
+      },
     },
     compilerOptions: {},
     runtimeCompiler: {
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -1,6 +1,6 @@
 import { existsSync } from 'node:fs'
 import * as vite from 'vite'
-import { basename, dirname, join, normalize, resolve } from 'pathe'
+import { dirname, join, normalize, resolve } from 'pathe'
 import type { Nuxt, NuxtBuilder, ViteConfig } from '@nuxt/schema'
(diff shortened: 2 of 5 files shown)


packages/vite/src/vite.ts
```diff
@@ -89,7 +89,6 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
         },
         resolve: {
           alias: {
-            [basename(nuxt.options.dir.assets)]: resolve(nuxt.options.srcDir, nuxt.options.dir.assets),
```
danielroe: I think we need to keep this entry here and in the webpack presets until https://github.com/vuejs/core/pull/13462 merges.

see https://github.com/nuxt/nuxt/pull/32119/commits/03d7c6e4fd14df072c4f5945e0ede7f399f58367

OrbisK: Okay. Looks like https://github.com/vuejs/core/pull/13462 comes with next minor

OrbisK:
@danielroe I have prepared this PR to be compatible when https://github.com/vuejs/core/pull/13462 gets merged.

I moved the legacy `transformAssetsUrls` tags to the `tags` key, because it does not let me set the new tilde option because of this check: https://github.com/vuejs/core/blob/main/packages/compiler-sfc/src/template/transformAssetUrl.ts#L52 


Tests are passing locally with:

```
  "pnpm": {
    "overrides": {
      "vue": "https://pkg.pr.new/vue@13462",
      "@vue/compiler-sfc": "https://pkg.pr.new/@vue/compiler-sfc@13462"
    }
  },
```

packages/schema/src/config/app.ts
```diff
@@ -7,11 +7,14 @@ import type { NuxtAppConfig } from '../types/config'
...
-      image: ['xlink:href', 'href'],
-      use: ['xlink:href', 'href'],
+      preserveTilde: true,
+      tags: {
+        video: ['src', 'poster'],
+        source: ['src'],
+        img: ['src'],
+        image: ['xlink:href', 'href'],
+        use: ['xlink:href', 'href'],
+      },
```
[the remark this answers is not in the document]
OrbisK: > Lift all tag entries out of the `tags` object into the root of `transformAssetUrls`

From vue changelog [3.0.0-beta.9](https://github.com/vuejs/core/blob/main/changelogs/CHANGELOG-3.0.md#300-beta9-2020-05-04)

```
compiler-sfc: @vue/compiler-sfc's transformAssetUrlsBase option has been removed. It is merged into trasnformAssetUrls which now also accepts the format of

{
  base?: string
  includeAbsolute?: string
  tags?: { [name: string]: string[] }
}
```

PR #32601, opened 2025-07-13
fix(nuxt): allow modules to add to `typescript.hoist`

danielroe:
### 📚 Description

https://github.com/nuxt-modules/storybook/issues/920#issuecomment-3065340572

this allows modules to push to `typescript.hoist` for key dependencies which they expect users to be able to import from or augment.


BobbieGoede (approved):
Great! Tested with storybook reproduction project.

BobbieGoede:
Tests flaky or did something actually break?

danielroe:
flaky tests for some reason

PR #32613, opened 2025-07-14
docs: fix abbreviation

iiio2:


huang-julien:
Hey 👋 I noticed you’ve opened several PRs to fix grammar and typos in the documentation.

Thanks for your contributions! Would you mind merging them into a single PR so we can review all the changes at once?

iiio2:
@huang-julien , Hey. 
Actually when I study about Nuxt or learn it, If I find any mistake then I try to fix it. I do not know how many things are there. Also reading the full doc is slightly tough for me. Merging all in a single PR it might also be tough for me because it may take much time. I think I can create many PRs. When you plan to make a new release, then you can add those in your branch or commit whatever you say. I think @danielroe can help for it if you prefer. Thanks.

danielroe (approved):

danielroe:
I find regexp based searching is very helpful for finding and spotting these kind of things once you find one instance.

and ultimately I would like lint configuration that fully covers everything (and then we take the decision either not to care about things that aren't caught by the linter, or improve the lint config).

iiio2:
Thanks @danielroe . :blush:

PR #32628, opened 2025-07-16
fix(nuxt): Improper code sanitization on JSON.stringify

odaysec:
https://github.com/nuxt/nuxt/blob/232b14e2f9708463148b08d9124e98303c12e9f3/packages/nuxt/src/pages/utils.ts#L551-L551
https://github.com/nuxt/nuxt/blob/232b14e2f9708463148b08d9124e98303c12e9f3/packages/nuxt/src/components/plugins/transform.ts#L77-L77

Using string concatenation to construct JavaScript code can be error-prone, or in the worst case, enable code injection if an input is constructed by an attacker.

fix the issue ensure that all potentially dangerous characters in values used to construct JavaScript code are escaped. This can be achieved by introducing a utility function (e.g., `escapeUnsafeChars`) that escapes characters like `<`, `>`, `\`, and others that could break out of JavaScript contexts or introduce vulnerabilities. The `serializeRouteValue` function should be updated to apply this escaping to its output. The changes require the following steps:
1. Define an `escapeUnsafeChars` function based on the provided safe escaping logic.
2. Update the `serializeRouteValue` function to apply `escapeUnsafeChars` to the result of `JSON.stringify`.

### DIFF
--- a/packages/nuxt/src/components/plugins/transform.ts
+++ b/packages/nuxt/src/components/plugins/transform.ts
@@ -1,3 +1,22 @@
+
+const charMap = {
+  '<': '\\u003C',
+  '>': '\\u003E',
+  '/': '\\u002F',
+  '\\': '\\\\',
+  '\b': '\\b',
+  '\f': '\\f',
+  '\n': '\\n',
+  '\r': '\\r',
+  '\t': '\\t',
+  '\0': '\\0',
+  '\u2028': '\\u2028',
+  '\u2029': '\\u2029',
+};
+
+function escapeUnsafeChars(str: string): string {
+  return str.replace(/[<>\b\f\n\r\t\0\u2028\u2029]/g, x => charMap[x] || x);
+}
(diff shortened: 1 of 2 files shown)


danielroe (commented):
1. security issues should be disclosed responsibly via our security policy
2. this is not untrusted input and is running on your source code at build time
3. it looks like you have used AI to generate the PR notes and likely also the PR, which is probably why you have missed the context of the functions

I always appreciate security issues but next time please disclose via https://github.com/nuxt/nuxt/security/advisories/new.

odaysec:
hi @danielroe sure its not a security issue there is a fixing a `feat` the real original poc will on advisory but there already closed

danielroe:
i'm not sure i understand what you are saying.

could you explain?

would you give a POC or an example of an issue that this PR solves?

odaysec:
hi @danielroe sure i will be eable in `advisory:new` for more discuss

PR #32643, opened 2025-07-16
docs: reference `noUncheckedIndexedAccess` rule change in v4 guide

Flamenate:
### 🔗 Linked issue

None.

### 📚 Description

The `noUncheckedIndexedAccess` rule change makes the new typescript configuration not 100% backward compatible.
I added the required migration step, but left the "Backward compatibility" bullet point because I think it's still relevant.

### DIFF
--- a/docs/1.getting-started/18.upgrade.md
+++ b/docs/1.getting-started/18.upgrade.md
@@ -985,10 +985,12 @@ Nuxt now generates separate TypeScript configurations for different contexts to 
 
 1. **New TypeScript configuration files**: Nuxt now generates additional TypeScript configurations:
    * `.nuxt/tsconfig.app.json` - For your app code (Vue components, composables, etc.)
+   * * `noUncheckedIndexedAccess` rule was changed from `false` to `true`.
    * `.nuxt/tsconfig.server.json` - For your server-side code (Nitro/server directory)  
    * `.nuxt/tsconfig.node.json` - For your build-time code (modules, `nuxt.config.ts`, etc.)
    * `.nuxt/tsconfig.shared.json` - For code shared between app and server contexts (like types and non-environment specific utilities)
    * `.nuxt/tsconfig.json` - Legacy configuration for backward compatibility
+   * * `noUncheckedIndexedAccess` rule was changed from `false` to `true`.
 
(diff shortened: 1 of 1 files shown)


docs/1.getting-started/18.upgrade.md
```diff
@@ -985,10 +985,12 @@ Nuxt now generates separate TypeScript configurations for different contexts to
 
 1. **New TypeScript configuration files**: Nuxt now generates additional TypeScript configurations:
    * `.nuxt/tsconfig.app.json` - For your app code (Vue components, composables, etc.)
+   * * `noUncheckedIndexedAccess` rule was changed from `false` to `true`.
    * `.nuxt/tsconfig.server.json` - For your server-side code (Nitro/server directory)  
    * `.nuxt/tsconfig.node.json` - For your build-time code (modules, `nuxt.config.ts`, etc.)
    * `.nuxt/tsconfig.shared.json` - For code shared between app and server contexts (like types and non-environment specific utilities)
    * `.nuxt/tsconfig.json` - Legacy configuration for backward compatibility
+   * * `noUncheckedIndexedAccess` rule was changed from `false` to `true`.
```
OrbisK: twice? 😅

Flamenate: I thought it's relevant to mention in both the new tsconfig and the old one. feel free to remove if you think it's unnecessary

danielroe (commented):
this change isn't related to the separate type contexts and should be a new section instead.

(here's the PR that introduced the change: https://github.com/nuxt/nuxt/pull/27485)

Flamenate:
> this change isn't related to the separate type contexts and should be a new section instead.

I was going to add it as a separate section at first, but I'm not sure what the reason behind the change is so I decided against it.

If you're going to add it as a new section, I suggest renaming the current one to something like `Typescript Configuration Splitting`, since there are actual configuration changes.

danielroe (approved):

danielroe:
thank you ❤️

docs/1.getting-started/18.upgrade.md
OrbisK: Sorry, Git diff looks weird on mobile.🫠

PR #32648, opened 2025-07-16
docs: add missing import in alias config example

buglavecz:

### DIFF
--- a/docs/3.api/6.nuxt-config.md
+++ b/docs/3.api/6.nuxt-config.md
@@ -37,7 +37,9 @@ your alias by prefixing it with `~`.
 
 **Example**:
 ```js
-export default {
+import { fileURLToPath, URL } from "node:url";
+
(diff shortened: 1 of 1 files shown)


docs/3.api/6.nuxt-config.md
```diff
@@ -37,13 +37,15 @@ your alias by prefixing it with `~`.
 
 **Example**:
 ```js
-export default {
+import { fileURLToPath, URL } from "node:url";
```
danielroe:
```suggestion
import { fileURLToPath } from "node:url";
```

buglavecz: @danielroe for the `new URL()` import the `URL` is not required? 🤔

danielroe: it is globally available

danielroe (approved):
thank you!

PR #32654, opened 2025-07-17
fix: compat bun oven-sh/bun#18737

typed-sigterm:
### 🔗 Linked issue

oven-sh/bun#18737

### 📚 Description

After upgrading to v4, Nuxt dev server could't work. When I visit the page:

```
[14:17:48]  ERROR  [unhandledRejection] { [Error: Malformed_HTTP_Response fetching "http://127.0.0.1:51684/_nuxt/?token=Gje1rGQ5rAeF". For more information, pass verbose: true in the second argument to fetch()]
  code: 'Malformed_HTTP_Response',
  path: 'http://127.0.0.1:51684/_nuxt/?token=Gje1rGQ5rAeF',
  errno: 0 }
```

So I spent 3 hrs working on this ✅

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -21,6 +21,12 @@ import { ModulePreloadPolyfillPlugin } from './plugins/module-preload-polyfill'
 import { ViteNodePlugin } from './vite-node'
 import { createViteLogger } from './utils/logger'
 
+// https://github.com/oven-sh/bun/issues/18737
+const removeTrailingSlash = (s: string) => {
+  const parsed = new URL(s, 'http://example.com').pathname
+  return parsed.length > 1 && parsed.at(-1) === '/' ? parsed.slice(0, -1) : parsed
+}
+
 export async function buildClient (nuxt: Nuxt, ctx: ViteBuildContext) {
   const nodeCompat = nuxt.options.experimental.clientNodeCompat
     ? {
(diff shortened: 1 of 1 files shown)


packages/vite/src/client.ts
```diff
@@ -32,7 +38,7 @@ export async function buildClient (nuxt: Nuxt, ctx: ViteBuildContext) {
   const clientConfig: ViteConfig = vite.mergeConfig(ctx.config, vite.mergeConfig({
     configFile: false,
     base: nuxt.options.dev
-      ? joinURL(nuxt.options.app.baseURL.replace(/^\.\//, '/') || '/', nuxt.options.app.buildAssetsDir)
+      ? removeTrailingSlash(joinURL(nuxt.options.app.baseURL.replace(/^\.\//, '/') || '/', nuxt.options.app.buildAssetsDir))
       : './',
```
typed-sigterm, line 37: I'm not sure if `./` could work?

packages/vite/src/client.ts
```diff
@@ -21,6 +21,12 @@ import { ModulePreloadPolyfillPlugin } from './plugins/module-preload-polyfill'
 import { ViteNodePlugin } from './vite-node'
 import { createViteLogger } from './utils/logger'
 
+// https://github.com/oven-sh/bun/issues/18737
+const removeTrailingSlash = (s: string) => {
```
danielroe: we have a withoutTrailingSlash utility we can use from ufo

danielroe (commented):
sounds like a wild ride! thank you for diving in

this seems safe to me

it could probably also be reported/hotfixed in vite

typed-sigterm:
> it could probably also be reported/hotfixed in vite

Node.js can handle the upgrade request of `ws://localhost/_nuxt/?token=xxx` correctly, it should be a Bun bug. I'm not sure if the hack should be sent to upstream.

typed-sigterm:
Hmm, doesn't work... 🤔

[`__HMR_BASE__` of `node_modules/vite/dist/client/client.mjs`](https://github.com/vitejs/vite/blob/587bdf700c8943ab330b1a41d914598adb51ac95/packages/vite/src/client/client.ts#L28-L39) is still replaced to `/_nuxt/`.

Could anyone help? I searched the identifiers but found nothing.

danielroe:
i would check to see if this is something vite is doing.

i would expect that vite normalises this value in any case, which is why i suggested that a hotfix should take place in vite rather than nuxt

typed-sigterm:
Looks good now...

PR #32660, opened 2025-07-17
feat: add buildAssetsFileNamePrefix option

maxmaxme:
Hi team! ▲▲

### 📚 Description

This PR adds a new configuration option `buildAssetsFileNamePrefix` that allows users to specify a prefix for built asset file names. This feature is useful for:

- Adding version identifiers to asset files (e.g., `v1-[hash].js`)
- Distinguishing assets from different environments (e.g., `prod-[hash].js`, `staging-[hash].js`)
- Implementing custom naming conventions for organization or caching purposes

_Note: I know about the `vite.build.rollupOptions.output`, but I don't want to override Nuxt's generation of filename because it can be changed in the future on the Nuxt side._

Example usage:
```ts
// nuxt.config.ts
  app: {
    buildAssetsFileNamePrefix: 'custom-'
  }
})
```

Or via environment variable:
```bash
NUXT_APP_BUILD_ASSETS_FILE_NAME_PREFIX=custom- node .output/server/index.mjs
```

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -157,7 +157,7 @@ export async function buildClient (nuxt: Nuxt, ctx: ViteBuildContext) {
   }
 
   // We want to respect users' own rollup output options
-  const fileNames = withoutLeadingSlash(join(nuxt.options.app.buildAssetsDir, '[hash].js'))
+  const fileNames = withoutLeadingSlash(join(nuxt.options.app.buildAssetsDir, `${nuxt.options.app.buildAssetsFileNamePrefix || ''}[hash].js`))
   clientConfig.build!.rollupOptions = defu(clientConfig.build!.rollupOptions!, {
     output: {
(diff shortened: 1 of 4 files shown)


packages/vite/src/client.ts
```diff
@@ -157,7 +157,7 @@ export async function buildClient (nuxt: Nuxt, ctx: ViteBuildContext) {
   }
 
   // We want to respect users' own rollup output options
-  const fileNames = withoutLeadingSlash(join(nuxt.options.app.buildAssetsDir, '[hash].js'))
+  const fileNames = withoutLeadingSlash(join(nuxt.options.app.buildAssetsDir, `${nuxt.options.app.buildAssetsFileNamePrefix || ''}[hash].js`))
```
huang-julien, line 160: hey :wave: I don't really see what this resolves as you can already customize filenames with `vite.build.rollupOptions.output` in your nuxt.config

PR #32671, opened 2025-07-18
feat(nuxt): add type declarations for shared auto-imports

### LINKED ISSUE #32646: fix(imports): enable intra-directory auto-imports for shared utilities
### Problem
Nuxt 4's auto-import system has a limitation where functions exported from shared directories (e.g., `shared/utils/`) are globally auto-imported for Vue components, but TypeScript/editors don't recognize these functions when used within the same directory. This causes "Cannot find name" errors in development despite the code working correctly at runtime.

### Example
In a typical setup with shared utilities:

```typescript
// shared/utils/typeGuards.ts
export const isArray = (value: any): value is any[] => Array.isArray(value)

// shared/utils/numberHelpers.ts
export const toNumber = (v: unknown) => 
  isArray(v) ? v.map(item => toNumber(item)) : Number(v) // ❌ Cannot find name 'isArray'
```

The `isArray` function is available globally in Vue components but not recognized within the same `shared/utils/` directory.

### Root Cause
The auto-import scanning in `packages/nuxt/src/imports/module.ts` only makes exports globally available but doesn't handle intra-directory imports. The `scanDirExports` function creates imports that work globally but not within the same directory scope.

### Solution
Enhanced the import scanning logic to create duplicate import entries for shared directories:

1. **Identify shared directories**: Filter `composablesDirs` for paths containing `/shared/`
2. **Create duplicate imports**: Generate additional import entries with special metadata for intra-directory access
3. **Maintain compatibility**: Preserve existing global auto-import functionality

### Implementation Details
The fix adds this logic to the `regenerateImports` function:

```typescript
// Add intra-directory imports for shared directories
const intraDirectoryImports: Import[] = []
const sharedDirs = composablesDirs.filter(dir => dir.includes('/shared/'))
for (const sharedDir of sharedDirs) {
  const dirImports = scannedImports.filter(imp => imp.from.startsWith(sharedDir))
  for (const imp of dirImports) {
    intraDirectoryImports.push({
      ...imp,
      meta: { ...imp.meta, intraDirectory: sharedDir }
    })
  }
}
imports.push(...scannedImports, ...intraDirectoryImports)
```

### Benefits
- ✅ **TypeScript Support**: Eliminates "Cannot find name" errors in editors
- ✅ **No Breaking Changes**: Maintains existing global auto-import behavior  
- ✅ **Better DX**: Enables seamless cross-file usage within shared directories
- ✅ **Performance**: Minimal overhead, only affects shared directories

stephenjason89:
### 📋 **Summary**

This PR introduces a dedicated TypeScript declaration file for shared auto-imports, solving the issue where shared utilities are globally auto-imported but not recognized within the same shared directory by TypeScript/editors.

### 🎯 **Problem**

As described in [#32646](https://github.com/nuxt/nuxt/issues/32646), Nuxt 4's auto-import system has a limitation where:

```typescript
// shared/utils/typeGuards.ts
export const isArray = (value: unknown): value is unknown[] => Array.isArray(value)

// shared/utils/numberHelpers.ts
export const toNumber = (v: unknown) => 
  isArray(v) ? v.map(item => toNumber(item)) : Number(v) // ❌ Cannot find name 'isArray'
```

- ❌ Shared imports were duplicated across multiple declaration files
- ❌ **Not available in shared directory files themselves** (main issue from #32646)
- ❌ Missing from server-side (Nitro) type declarations
- ❌ Causing "Cannot find name" TypeScript errors in editors

### ✨ **Solution**

Instead of the complex duplicate import approach proposed in the issue, this PR implements a cleaner solution using TypeScript's declaration file system:

- ✅ Creates `types/shared.d.ts` with **only** shared auto-imports
- ✅ Adds it to shared TypeScript context via `{ shared: true }`
- ✅ References it from all relevant declaration files
- ✅ **Eliminates TypeScript errors in shared directories** 🎯

#### 1. **New Declaration File**
- Added `types/shared.d.ts` template that filters and generates declarations for shared imports only
- Configured with `{ shared: true }` context to include in `nuxt.shared.d.ts`

#### 2. **Updated Existing Templates**
- **`imports.d.ts`**: Excludes shared imports, references `types/shared.d.ts`
- **`types/imports.d.ts`**: Excludes shared imports, references `shared.d.ts`
- **`types/nitro-nuxt.d.ts`**: References `shared.d.ts` for server-side access

### 📁 **File Structure**

```
.nuxt/
├── imports.d.ts           # Non-shared imports + ref to shared.d.ts
├── nuxt.shared.d.ts       # ✅ Includes reference to types/shared.d.ts
├── types/
│   ├── imports.d.ts       # Non-shared imports + ref to shared.d.ts  
│   ├── shared.d.ts        # 🆕 ONLY shared imports (solves #32646)
│   └── nitro-nuxt.d.ts    # Server types + ref to shared.d.ts
```

#### ✅ **Now Works Without Errors:**
```typescript
// shared/utils/typeGuards.ts
export const isArray = (value: unknown): value is unknown[] => Array.isArray(value)

// shared/utils/numberHelpers.ts
export const toNumber = (v: unknown) => 
  isArray(v) ? v.map(item => toNumber(item)) : Number(v) // ✅ Works perfectly!
```

### 🔍 **Benefits**

1. **Solves #32646**: ✅ TypeScript recognizes shared imports within shared directories
2. **No Duplication**: Shared imports declared once, referenced everywhere
3. **Universal Access**: Available in client, server, and shared contexts
4. **Type Safety**: Full TypeScript intellisense across all contexts
5. **Performance**: Zero runtime overhead, pure TypeScript solution

### ⚠️ **Breaking Changes**

None. This is a purely additive feature that enhances existing functionality.

### ✅ **Testing**

- [x] **Fixes #32646**: Shared imports work within shared directory files
- [x] Shared imports work in client components
- [x] Shared imports work in server API routes  
- [x] No TypeScript compilation errors
- [x] No duplication in generated declaration files
- [x] Build process completes successfully

**Closes #32646** 🎉


packages/nuxt/src/imports/module.ts
```diff
@@ -228,19 +234,66 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
 
   addTypeTemplate({
     filename: 'imports.d.ts',
-    getContents: async ({ nuxt }) => toExports(await ctx.getImports(), nuxt.options.buildDir, true),
+    getContents: async ({ nuxt }) => {
```
danielroe: we shouldn't exclude them from this file as it is still possible to import from #imports (which is what this file represents)

packages/nuxt/src/imports/module.ts
```diff
@@ -228,19 +234,66 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
...
-      await cacheImportPaths(imports)
-      return '// Generated by auto imports\n' + (
+      const allImports = await ctx.getImports()
+      await cacheImportPaths(allImports)
+
+      // Exclude shared imports to avoid duplication
+      const nonSharedImports = allImports.filter(imp => !isSharedImport(imp))
+
+      // Create a temporary context with only non-shared imports
+      const nonSharedCtx = createUnimport({
```
danielroe: this is likely to be non performant, as it creates a new unimport context every time the file is generated

packages/nuxt/src/imports/module.ts
```diff
@@ -228,19 +234,66 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
...
+
+      // Filter imports that come from shared directories
+      const sharedImports = allImports.filter(isSharedImport)
+
+      if (sharedImports.length === 0) {
+        return '// No shared imports found'
+      }
+
+      // Create a temporary context with only shared imports
+      const sharedCtx = createUnimport({
```
danielroe: likewise

packages/nuxt/src/core/templates.ts
```diff
@@ -366,6 +366,7 @@ export const nitroSchemaTemplate: NuxtTemplate = {
     return /* typescript */`
 ${lines.join('\n')}
 /// <reference path="./runtime-config.d.ts" />
+/// <reference path="./shared.d.ts" />
```
danielroe:
```suggestion
/// <reference path="./types/shared.d.ts" />
```

i would also recommend renaming to shared-imports.d.ts for clarity

packages/nuxt/src/imports/module.ts
stephenjason89: I’ve removed the exclusion. Thanks for the heads-up!

packages/nuxt/src/imports/module.ts
```diff
@@ -243,4 +243,19 @@ function addDeclarationTemplates (ctx: Unimport, options: Partial<ImportsOptions
       )
     },
   })
+
+  // Add shared imports declaration file
+  addTypeTemplate({
+    filename: 'types/shared-imports.d.ts',
+    getContents: async () => {
+      const imports = await ctx.getImports()
+      await cacheImportPaths(imports)
```
danielroe: this doesn't look right to me - we need to generate only the types for imports that are available in _both_ nitro + nuxt auto-importable context.

stephenjason89: You're absolutely right! I have now filtered imports from the /shared/ directory. Would we want other imports other than the ones inside shared to be included here as well?

danielroe: this doesn't seem to be the right way - what if there are other imports added to both nuxt + nitro auto-imports? Or what if the shared directory is named differently (it's configurable)? is it possible to iterate through both unimport instances used in nitro + nuxt?

stephenjason89: Sorry, i totally missed that! The current approach has several issues:

- Hardcoded directory name: I'm assuming the shared directory is always named "shared", but it's configurable via `nuxt.options.dir.shared`

- Missing other imports: This doesn't account for other imports that might be added to both Nuxt and Nitro auto-imports

- Better approach needed: We should iterate through both unimport instances used in Nitro and Nuxt to find the intersection
	- Instead of filtering by directory path, we should:
		- Get all imports from the Nuxt unimport context
		- Get all imports from the Nitro unimport context
		- Find the intersection (imports available in both contexts)
		- Generate types for that intersection

This would be more robust and handle configurable directory names and additional imports properly. Should I implement this approach instead?

danielroe: are you there? or is this an ai response?

stephenjason89: I'm here 😂 Had ai format my response to be clearer though 😂

stephenjason89: Hello @danielroe, i have addressed the last issues. Please let me know if this needs further revisions. Thank you so much 😊

danielroe:
thank you for your work on this ❤️ 

I've merged a slightly different implementation, and marked you as a co-author 🙏

stephenjason89:
> thank you for your work on this ❤️ 
> 
> 
> 
> I've merged a slightly different implementation, and marked you as a co-author 🙏 

Thanks Daniel! Appreciate your time and the co-author credit 🙏

The final implementation looks great🔥

PR #32723, opened 2025-07-22
docs: set correct default value for deep option in asyncdata

mmeester:
### 📚 Description

This pull request updates the documentation for the deep option in useAsyncData/useFetch to set the default value to `true` instead of `false`. This change aligns the documentation with the actual implementation in the source code and ensures consistency for users referencing the docs.

#### Motivation
Previously, there was a discrepancy between the Nuxt documentation and the codebase regarding the default value of the deep option. The documentation stated the default was `false`, while the implementation and expected behavior is `true`. This update corrects the documentation, helping prevent user confusion and improving clarity.

Summary of changes
- Set the documented default value of deep to `true`.
- Clarified related explanations to reflect this default.


danielroe:
it defaults to `false` in nuxt 4

PR #32724, opened 2025-07-22
docs: set correct default value for deep option in usefetch

mmeester:
### 📚 Description

This pull request updates the documentation for the deep option in useFetch to set the default value to `true` instead of `false`. This change aligns the documentation with the actual implementation in the source code and ensures consistency for users referencing the docs.

Motivation
Previously, there was a discrepancy between the Nuxt 3.x documentation and the codebase regarding the default value of the `deep` option. The documentation stated the default was `false`, while the implementation and expected behavior is true. This update corrects the documentation, helping prevent user confusion and improving clarity.

Summary of changes

- Set the documented default value of deep to `true`.
- Clarified related explanations to reflect this default.


danielroe:
as with the other PR, the default is false for v4

what made you think otherwise?

mmeester:
@danielroe I know, I made the misstake of adjusting it to v4 first, this is a change for v3 docs

danielroe (approved):

danielroe:
thank you!

PR #32744, opened 2025-07-24
fix(nuxt): execute all plugins after error rendering error.vue

### LINKED ISSUE #32743: Plugin error is obfuscated by errors caused by skipped plugins
### Environment

(stackblitz)
- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     4.0.1
- CLI Version:      3.26.4
- Nitro Version:    2.12.3
- Builder:          -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-1wnis5cb?file=app%2Fplugins%2F1.first-plugin.ts,app%2Fplugins%2F2.second-plugin.ts,app%2Ferror.vue

### Describe the bug

If a project has a custom `error.vue` page, and this page (or its layout) is using methods/properties provided by other plugins, the error visible to the user will obfuscate the actual source error.

The visible error would be ` ERROR  [unhandledRejection] _ctx.$formatString is not a function`, because the plugin was skipped due to an erroring plugin before it.

While the original error thrown by the broken plugin would be much more helpful
```
Bad plugin!
at setup (/home/projects/nbwioavjwg.github/app/plugins/1.first-plugin.ts:7:11)
```

While this might be an edge case, it's not uncommon to have localized error pages, so users with broken plugins end up opening issues at nuxt-i18n.

Maybe we should allow other plugins to run if a previous plugin throws an error? 🤔 The current behavior might encourage plugin creators to enforce order which is probably not something we want.

### Additional context

Sometimes the full error is shown, which really would suggest this method is the source of the error.

Image

### Logs

```shell-script

```

Related
* https://github.com/vuejs/vuefire/issues/1633
* https://github.com/nuxt-modules/i18n/issues/3755

BobbieGoede:
### 🔗 Linked issue
* #32743

### 📚 Description
Resolves #32743

This way the order of plugins won't break the setup of unrelated plugins.


danielroe (commented):
This makes sense to me. I think originally this might have been because we wanted to avoid other plugins erroring if they _depended_ on the broken original plugin.

... but as we are already handling an array of errors, I think it makes sense.

maybe we could even move the `.catch` to where the promise is defined to avoid duplicating on L448

BobbieGoede:
> maybe we could even move the `.catch` to where the promise is defined to avoid duplicating on L448

Done!

>This makes sense to me. I think originally this might have been because we wanted to avoid other plugins erroring if they depended on the broken original plugin.
>... but as we are already handling an array of errors, I think it makes sense.

We're collecting errors in an array but only throwing the first error on L474, do we plan to make those other errors visible someday? Hiding the errors from plugins depending on a broken plugin makes sense to me though, but (with this change) it's possible unrelated plugins will error too.

danielroe (commented):
what about we only short-circuit running the plugins if we _aren't_ rendering an error page?

packages/nuxt/src/app/nuxt.ts
```diff
@@ -425,6 +425,11 @@ export async function applyPlugins (nuxtApp: NuxtApp, plugins: Array<Plugin & Ob
   let promiseDepth = 0
 
   async function executePlugin (plugin: Plugin & ObjectPlugin<any>) {
+    // Short-circuit when not rendering error page
+    if (!nuxtApp.payload.error && errors.length) {
+      throw errors[0]
```
BobbieGoede: @danielroe 
Not sure if this is the right way to check if we're rendering an error page, is this what you had in mind?

danielroe (approved):

huang-julien (approved):

danielroe:
sorry for the back and forth! 🤦

hopefully this is still good!

packages/nuxt/src/app/nuxt.ts
```diff
@@ -442,10 +442,16 @@ export async function applyPlugins (nuxtApp: NuxtApp, plugins: Array<Plugin & Ob
             }
           }))
         }
+      }).catch((e) => {
+        // short circuit if we are not rendering `error.vue`
+        if (!plugin.parallel && !nuxtApp.payload.error) {
+          throw error
```
BobbieGoede, line 448: @danielroe 
This should probably be `throw e`, right? 👀

PR #32759, opened 2025-07-25
fix(nuxt): update `unbuild` and use absolute path in stubs

BobbieGoede:
### 📚 Description
This should avoid jiti not found errors during stub build. This change should only affect development, not sure if `fix:` prefix is appropriate.

Not sure if there are any related open issues, the edge case this resolves is a bit specific.

For context to those unaware, this should prevent the following error when an error is thrown while building a stub (screenshot is module builder but same applies to nuxt stub builds):
image


danielroe (approved):
thank you! ❤️

PR #32762, opened 2025-07-25
fix(nuxt): ensure getCachedData function is returning cache if possible

Aietes:
Unless there is manual refresh or refresh() has been explicitly called, return the cache if available. use payload data first, fallback to static if that is not available. Returns undefined when no cache has been found, triggering a fetch. Previous default function only returned cache when hydrating, which is not desired in most scenarios, therefore the default is changed.

### 🔗 Linked issue

Resolves #31816 and #32658

### 📚 Description

When using `useAsyncData` our expectation is that for the same id cached data is returned, as long as there is no manual refresh of the page or `refresh()` is used. Currently this does not work as expected, a new request is made every time. This is due to the implementation of the default getCachedData function:

```js
const getDefaultCachedData = (key, nuxtApp, ctx) => {
  if (nuxtApp.isHydrating) {
    return nuxtApp.payload.data[key];
  }
  if (ctx.cause !== "refresh:manual" && ctx.cause !== "refresh:hook") {
    return nuxtApp.static.data[key];
  }
};
```

This function only returns cached data when `isHydrating` is `true`. As this is only true on the initial page load, you only get cached data when the page loads initially. However, our expectation would be to also get cached data beyond the point of hydration, in fact caching is typically a lot more important beyond that point in a SPA with many components sharing data. For example when using forms on tabs, the data is now fetched on every tab change, which is certainly not desired.

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -736,12 +736,8 @@ function createAsyncData<
 // Used to get default values
 const getDefault = () => undefined
 const getDefaultCachedData: AsyncDataOptions<any>['getCachedData'] = (key, nuxtApp, ctx) => {
-  if (nuxtApp.isHydrating) {
-    return nuxtApp.payload.data[key]
-  }
-
   if (ctx.cause !== 'refresh:manual' && ctx.cause !== 'refresh:hook') {
-    return nuxtApp.static.data[key]
+    return nuxtApp.payload.data[key] || nuxtApp.static.data[key] || undefined
   }
 }
 


danielroe:
this is very similar to https://github.com/nuxt/nuxt/pull/32447.

note that it has been a deliberate decision since the release of Nuxt v3 not to use cached data after hydration, and changing this behaviour would be a breaking change.

Aietes:
If this is conscious decision, are you willing to revisit it? Right now in Nuxt there is basically NO caching after hydration. I don't think this is expected by developers, and is a wasted opportunity with all the caching mechanisms in place. The way Nuxt is used for most applications, effectively it's like **there is no caching**, or am I missing something here?

In real world use cases caching would be important and should be the default. As it stands now, on many trivial UI interactions, like changing a Nuxt UI tab, data is being refetched, which is counterintuitive. Caching should be on by default, as it's straightforward to trigger a manual refresh if needed, but it is not trivial to change the default caching behavior for less experienced users.

danielroe:
I'd be willing to revisit it, but bear in mind:

- an infinite cache = a memory leak so your app _probably_ needs to deal with refetching data (at least as an edge case)
- we don't want to break users so this would be a major version change or an option (global perhaps)
- we already effectively have an infinite cache via the browser when using payload extraction (because the `payload.json` can be cached). support for non-static site payload extraction is coming soon

Aietes:
I would be willing to work on an approach for more sustainable caching. One of our applications has many asyncData fetches across many components and pages, so we need to develop an approach anyway. I also think this is a critical feature for Nuxt. Do you have a preferred way to steer the development of such a feature? I can propose an approach and mechanics for caching before we implement it, it would be great to have input from the core team and community.

PR #32767, opened 2025-07-25
chore(deps): update vite + vite plugins to v7

danielroe:
### 📚 Description

remake of https://github.com/nuxt/nuxt/pull/32470 for 3.x


test/bundle.test.ts
TheAlexLichter, line 1: Nice to see that this *decreases* the bundle size!

danielroe: 💯

Issue #32751, opened 2025-07-24
Server components (nuxt island) urls are relative and can't be resolved from other domains even with BASE_URL


maximelebreton:
### Environment

Hello everybody, and first of all, thanks to all the team for your incredible job!

Here is the context:
I have a Nuxt app hosted on a **root domain**.
Some parts of the app are also accessible via **subdomains** (or even different domains).

To make assets and API calls work properly across all domains, I’ve set:

`.env`
```env
BASE_URL="http://labaxxxxxxxxx.fr"
```

`nuxt.config.ts`
```ts
{
//...
  app: {
    cdnURL: process.env.BASE_URL
  },
  runtimeConfig: {
    public: {
      baseURL: process.env.BASE_URL,
    }
 }
//...
}
```

### Reproduction

No reproduction

### Describe the bug

Everything works as expected on the **root domain** and even from **subdomains** for assets and API calls.

However, server components break when accessed from **subdomains** because their requests seem to be **relative**, not **absolute**.

✅ If the page is open from my **root domain**, everything works as excepted.
```
https://labaxxxxxxxxx.fr
``` 

But if I the page is open from a subdomain:

```
https://labrasive.labaxxxxxxxxx.fr
``` 

✅ The `assets` or `api` calls correctly targets the **root domain**, like this `/band`:   

<img width="419" height="85" alt="Image" src="https://github.com/user-attachments/assets/2d43dd7a-edd4-4ad0-8d8e-74fa33d72efd" />

❌ But the `Markdown.server.vue` component call is broken because it doesn't target the **root domain**...  

<img width="630" height="67" alt="Image" src="https://github.com/user-attachments/assets/6bf99ef7-d57b-4694-b3b7-faf3a7fbe1ab" />  

So is there a way to set a `BASE_URL` to Server Components, or is it a bug?

Thank you for your help!

### Logs

```shell-script

```

maximelebreton:
Maybe duplicate with, but not sure: https://github.com/nuxt/nuxt/issues/32637

jonitiso:
Hi, i have the same issue but directly with pages, i have a page with .server.

I have

    app: {
        baseURL: '<MyURL>',
    },

but not is used


 @danielroe

danielroe:
@jonitiso would you provide a reproduction? 🙏

danielroe:
closing as likely resolved in https://github.com/nuxt/nuxt/pull/32772. if not, please provide a reproduction 🙏

PR #32772, opened 2025-07-26
fix(nuxt): use joinURL when remote island is provided

### LINKED ISSUE #32637: Applications with a base URL cannot be used as a source for remote component islands
### Environment

- Operating System: Darwin
- Node Version:     v23.9.0
- Nuxt Version:     3.16.2
- CLI Version:      3.25.0
- Nitro Version:    2.11.9
- Package Manager:  bun@1.2.4
- Builder:          -
- User Config:      extends, modules, imports, devtools, app, experimental, css, sourcemap, compatibilityDate, nitro, vite, logLevel, eslint, security
- Runtime Modules:  @nuxt/devtools@1.7.0, @nuxt/image@1.10.0, @vee-validate/nuxt@4.15.0, nuxt-security@2.2.0, @nuxt/eslint@0.6.2
- Build Modules:    -

### Reproduction

CodeSandbox: https://codesandbox.io/p/sandbox/component-island-remote-with-base-url-repro-yw25f9

This is a bit tricky to reproduce as it requires two Nuxt applications where one has a base URL, and the other is consuming a server island from the application with the base URL. I have managed to reproduce this locally using Caddy to reverse proxy and get around CORS issues. However, multiple Nuxt sites on the same domain is also reflective of my production environment.

### Describe the bug

I have a Nuxt application that has a `baseURL` configured in `nuxt.config`. When attempting to use component islands remotely in a second Nuxt application, the base URL is not taken into account.

For example, say the host of my component island, `MyIsland` is running on `http://localhost:3000/base/`. When trying to use `MyIsland` via remote component islands on another Nuxt application, like the following:

```tsx
<NuxtIsland name="MyIsland" source="http://localhost:3000/base/" />
```

A request is made to `http://localhost:3000/__nuxt_island/MyIsland[hash].json` instead of `http://localhost:3000/base/__nuxt_island/MyIsland[hash].json`. I have narrowed it down to [this line of code](https://github.com/nuxt/nuxt/blob/5af115f0eb2cfa21b5ad5bb86eeac2d6ebee3d9c/packages/nuxt/src/app/components/nuxt-island.ts#L198C1-L199C1) where the URL is constructed. Since `/__nuxt_island` is provided as the first argument of the `URL` constructor, the base will always be removed from `source`.

huang-julien:
### 🔗 Linked issue

fix [#32637](https://github.com/nuxt/nuxt/issues/32637)

### 📚 Description

This PR makes nuxtisland use joinURL instead of cosntructing the URL with new URL which removes any base path from the provided remote source

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-island.ts
+++ b/packages/nuxt/src/app/components/nuxt-island.ts
@@ -195,8 +195,7 @@ export default defineComponent({
 
       if (!force && nuxtApp.payload.data[key]?.html) { return nuxtApp.payload.data[key] }
 
-      const url = remoteComponentIslands && props.source ? new URL(`/__nuxt_island/${key}.json`, props.source).href : `/__nuxt_island/${key}.json`
-
+      const url = remoteComponentIslands && props.source ? joinURL(props.source, `/__nuxt_island/${key}.json`) : `/__nuxt_island/${key}.json`
       if (import.meta.server && import.meta.prerender) {
         // Hint to Nitro to prerender the island component
         nuxtApp.runWithContext(() => prerenderRoutes(url))
--- a/test/nuxt/nuxt-island.test.ts
+++ b/test/nuxt/nuxt-island.test.ts
@@ -94,6 +94,34 @@ describe('runtime server component', () => {
     await server.close()
   })
 
+    it('expect remote island with baseURL to be rendered', async () => {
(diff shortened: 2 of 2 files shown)


danielroe (approved):

danielroe:
does this also resolve https://github.com/nuxt/nuxt/issues/32751?

PR #32797, opened 2025-07-28
fix: make devServer.url dynamic based on port and host configuration

cheezone:
- Replace hardcoded 'http://localhost:3000' with dynamic URL generation
- Use  to access devServer.port, devServer.host, and devServer.https
- Support custom ports, hosts, and HTTPS configuration
- Maintain backward compatibility when url is explicitly set
- Fix getClientManifest to use correct port in development mode

### DIFF
--- a/packages/schema/src/config/dev.ts
+++ b/packages/schema/src/config/dev.ts
@@ -6,7 +6,19 @@ export default defineResolvers({
     https: false,
     port: Number(process.env.NUXT_PORT || process.env.NITRO_PORT || process.env.PORT || 3000),
     host: process.env.NUXT_HOST || process.env.NITRO_HOST || process.env.HOST || undefined,
-    url: 'http://localhost:3000',
+    url: {
+      $resolve: async (val, get) => {
+        if (typeof val === 'string') {
+          return val
+        }
+        const port = await get('devServer.port')
+        const host = await get('devServer.host')
+        const https = await get('devServer.https')
+        const protocol = https ? 'https' : 'http'
+        const hostname = host || 'localhost'
+        return `${protocol}://${hostname}:${port}`
+      },
+    },
     loadingTemplate,
     cors: {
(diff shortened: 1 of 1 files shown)


danielroe:
is this AI-generated? (i'm asking because the description doesn't seem to match the content of the PR.)

if not, would you create a reproduction of the bug that this PR solves?

cheezone:
> is this AI-generated? (i'm asking because the description doesn't seem to match the content of the PR.)
> 
> if not, would you create a reproduction of the bug that this PR solves?

Yes, because I'm not very familiar with English. Simply put, when I set the port to something other than 3000, HMR sometimes throws errors. In my case, it tries to access http://localhost:3000/__nuxt_vite_node__/manifest, even though I’ve already configured a non-3000 port. I can provide a minimal reproduction tomorrow if needed (though I don’t think it’s necessary — hardcoding like that probably isn’t correct anyway).

danielroe:
i don't mind ai-generated things, but please make sure they are accurate

nuxt doesn't actually hardcode this value although it might seem that way.

this value is in fact set by nuxt/cli on startup after the url is resolved.

cheezone:
> i don't mind ai-generated things, but please make sure they are accurate
> 
> nuxt doesn't actually hardcode this value although it might seem that way.
> 
> this value is in fact set by nuxt/cli on startup after the url is resolved.

Oh, that’s a bit troublesome. I’ll take a closer look tomorrow to figure out what’s really causing the issue. Thanks for your review!

cheezone:
image
image

After HMR, `process.env.NUXT_VITE_NODE_OPTIONS` will changed sometimes.

cheezone:
image

Oh..... I use my pr in pr new, but also get error.

danielroe:
why don't you start by creating and issue and a reproduction, and we can brainstorm?

cheezone:
> why don't you start by creating and issue and a reproduction, and we can brainstorm?

I tried modifying the official template, setting a custom port of 3008, and then writing a script to modify the text in the component to a random number to trigger HMR. However, this seems to have nothing to do with `getClientManifest`, so the `bug` can't be triggered. So sad.

PR #32820, opened 2025-07-30
docs: mention `bun.lock` for lockfile

iiio2:
I think here we can also mention bun.lock for lockfile. bun generates default `bun.lock` from v1.2. 
https://bun.sh/docs/install/lockfile


danielroe (approved):

iiio2:
Thanks @danielroe . :blush:

PR #32827, opened 2025-07-31
fix(nuxt): trigger call once navigation even when no suspense

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32811

### 📚 Description

`page:start` is only called when suspense initiates, meaning the previous implementation did not reset `callOnce` when synchronous navigation took place.

was there a reason we didn't use the router originally @TheAlexLichter?


TheAlexLichter:
The only scenario I could think of that could be different would be triggering when a `pageKey` is set - but I've never tried it together with `callOnce`.

thenetwrx:
@danielroe this doesn't solve the issue unless the behavior you expect is different from what i thought in the original issue. The repro is still the same but with nuxt dependency set at 4.0.3. the callback is still not executed on client side routing

danielroe:
callOnce does not automatically re-call the handler, and you called it in app.vue setup, which only runs once.

you could recall it in a watcher.

thenetwrx:
I'm trying not to sound passive aggressive, but the docs say it will, specifically on client side navigation. What is the difference between the modes and importantly what should occur?

danielroe:
the docs may be unclear

the idea is _if you call it in a page_  it will only run again if mode: navigation is set

it won't automatically run on navigation - you have to call it again

PR #32832, opened 2025-07-31
fix(nuxt): add warning for lazy-hydration missing prefix

### LINKED ISSUE #32736: Warn when using hydration directive without `<Lazy` prefix.
### Describe the feature

Right now, there is no warning when people use hydration strategies but forgetting to add the `<Lazy` prefix. This should give a warning during dev. 

### Additional information

### Final checks

OskarLebuda:
### 🔗 Linked issue

resolves #32736

### 📚 Description
Add a missing warning for components that use the `lazy-hydration` props but aren't prefixed with `<Lazy`. This helps catch cases where developers intend to lazy-load a component but forget the naming convention, potentially leading to unexpected behavior.

Example:
```vue
<template>
  <SomeComp hydrate-never/>
</template>
```

Warning:
obraz

### DIFF
--- a/packages/nuxt/src/components/plugins/lazy-hydration-transform.ts
+++ b/packages/nuxt/src/components/plugins/lazy-hydration-transform.ts
@@ -76,7 +76,8 @@ export const LazyHydrationTransformPlugin = (options: LoaderOptions) => createUn
             }
             if (!/^(?:Lazy|lazy-)/.test(node.name)) {
               if (node.name !== 'template') {
-                logger.warn(`Component ${node.name} has lazy-hydration props but it's not declared as lazy component. \nChange it to <Lazy${node.name} /> or remove the lazy-hydration props.`)
+                logger.warn(`Component <${node.name}> has lazy-hydration props but is not declared as a lazy component.\n` +
+                  `Rename it to <Lazy${node.name} /> or remove the lazy-hydration props to avoid unexpected behavior.`)
               }
               return
             }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/components/plugins/lazy-hydration-transform.ts
```diff
@@ -75,6 +75,10 @@ export const LazyHydrationTransformPlugin = (options: LoaderOptions) => createUn
               return
             }
             if (!/^(?:Lazy|lazy-)/.test(node.name)) {
+              if (node.name !== 'template') {
+                logger.warn(`Component <${node.name}> has lazy-hydration props but is not declared as a lazy component.\n` +
```
TheAlexLichter: Do we have any chance to point at the file (and ideally LOC) here? That'd be even better DX

OskarLebuda: I'll try to check that!

OskarLebuda: @TheAlexLichter switched it to run only in `dev` and `test` mode :) I also added `test` explicitly so it doesn’t get skipped during test runs.

OskarLebuda: @TheAlexLichter by “point the file” - do you mean the file where the issue actually occurs? Like, where the component is being **used**, not where it’s defined?

TheAlexLichter: Yes, ideally where it is used so I can fix it with a click in my console 👍🏻

OskarLebuda: Added 👍🏻

OskarLebuda: @TheAlexLichter is that what you want?

danielroe (approved):

PR #32835, opened 2025-08-01
fix(vite): include tsconfig references during `typeCheck`

### LINKED ISSUE #32823: [v4]: typecheck on build does not work
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     4.0.2
- CLI Version:      3.27.0
- Nitro Version:    2.12.4
- Builder:          -
- User Config:      compatibilityDate, devtools, typescript
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

[reproduction](https://stackblitz.com/edit/github-qvzrtdvv)

### Describe the bug

Nuxt will not typecheck the codebase when setting the `typescript.typeCheck` to `build` on nuxt config.

I have placed both typescript errors in `app.vue` and `server/routes/health.ts` files and no errors are thrown on build. Running nuxt typecheck directly will only throw the app.vue error and not the one on the server directory.

### Additional context

_No response_

### Logs

```shell-script

```

BobbieGoede:
### 🔗 Linked issue
* #32823

### 📚 Description
Resolves #32823

@danielroe 
The type checking is not using the `-b` argument, so no type errors are reported in projects using the new tsconfig structure. 

We should probably check if the project is using tsconfig references and enable `buildMode` based on that.

This seems to work, but it seems to report type errors in server routes twice 🤔, it also adds `tsconfig.*.tsbuildinfo` files in `.nuxt` for each tsconfig file it contains which may be undesirable.

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -8,7 +8,7 @@ import type { RollupReplaceOptions } from '@rollup/plugin-replace'
 import { sanitizeFilePath } from 'mlly'
 import { withTrailingSlash, withoutLeadingSlash } from 'ufo'
 import { filename } from 'pathe/utils'
-import { resolveTSConfig } from 'pkg-types'
+import { readTSConfig, resolveTSConfig } from 'pkg-types'
 import { resolveModulePath } from 'exsolve'
 
 import { buildClient } from './client'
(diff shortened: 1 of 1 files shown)


danielroe:
we probably need to do this only when the user has the new `references` format `tsconfig`.

BobbieGoede:
> we probably need to do this only when the user has the new `references` format `tsconfig`.

I figured this was checked somewhere already but couldn't find it so I adapted the checking logic from nuxt-cli.

Type checking (in projects with tsconfig references) seems to report server route errors twice, not sure how to prevent that from happening 🤔 still better than no errors at all though

BobbieGoede:
Nevermind, the double error reporting may have been caused by a stray `tsconfig.tsbuildinfo` in the root of the project (may have been created during earlier tests). It looks like this is working fine

packages/vite/src/vite.ts
```diff
@@ -184,10 +184,13 @@ export const bundle: NuxtBuilder['bundle'] = async (nuxt) => {
 
   // Add type-checking
   if (!nuxt.options.test && (nuxt.options.typescript.typeCheck === true || (nuxt.options.typescript.typeCheck === 'build' && !nuxt.options.dev))) {
+    const tsconfigPath = await resolveTSConfig(nuxt.options.rootDir)
+    const supportsProjects = await readTSConfig(tsconfigPath).then(r => !!(r.references?.length))
```
danielroe, line 188:
```suggestion
    const supportsProjects = await readTSConfig(nuxt.options.rootDir).then(r => !!(r.references?.length))
```

danielroe, line 188:
```suggestion
    const supportsProjects = await readTSConfig(tsconfigPath).then(r => !!(r.references?.length)).catch(() => false)
```

... or maybe instead we could log if we can't resolve the tsconfig

danielroe: actually, we were already not handling an error, so I think it's fine to leave as is.

danielroe (approved):
❤️

PR #32843, opened 2025-08-03
docs: add a section about augmenting types with TS project references

cernymatej:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt/issues/32727

### 📚 Description
This adds a section about needing to augment types in the correct context with the new Project Referneces TS setup and also mentions it in the upgrade guide to Nuxt 4.

----
I wasn't able to able to get augmentations recognized outside the default contexts (i.e. in `/types/` which would be in the same directory as `nuxt.config.ts`) by doing this:
```ts
typescript: {
    nodeTsConfig: {
        include: [
            resolve('./types/**/*.ts'),
        ],
    },
},
```
So I didn't mention it yet.

It did pick up the augmentations when I added the directory to `typescript.tsConfig.include` - but if I'm not mistaken, that only includes it in the `app` context. It also didn't work when I tried defining it in `typescript.sharedTsConfig.include`.
Is there something I'm missing? https://stackblitz.com/edit/github-1zim1ccd?file=project%2Fnuxt.config.ts

### DIFF
--- a/docs/1.getting-started/18.upgrade.md
+++ b/docs/1.getting-started/18.upgrade.md
@@ -1071,15 +1071,15 @@ However, to take advantage of improved type checking, you can opt in to the new 
    - "typecheck": "nuxt prepare && vue-tsc --noEmit"
    + "typecheck": "nuxt prepare && vue-tsc -b --noEmit"
    ```
-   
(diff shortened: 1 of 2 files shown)


danielroe (approved):
👌

PR #32844, opened 2025-08-03
fix(nuxt): remove unused `enabled` flag from directory options

cernymatej:
### 📚 Description

This PR removes an unused `enabled` flag in the components directory options.

It was hard-coded to `true` in the `app:resolve` hook in:
https://github.com/nuxt/nuxt/blob/5f6cc151328b694d174d20e94958c265eea5c78b/packages/nuxt/src/components/module.ts#L105-L115

So it shouldn't have had any effect here either:
https://github.com/nuxt/nuxt/blob/5f6cc151328b694d174d20e94958c265eea5c78b/packages/nuxt/src/components/scan.ts#L34-L36

Since the only time the `scanComponents` function is called, it uses the array with the hard-coded value:
https://github.com/nuxt/nuxt/blob/5f6cc151328b694d174d20e94958c265eea5c78b/packages/nuxt/src/components/module.ts#L181-L182


cernymatej:
One test failed because it passed component entries with `enabled: false` to the `scanComponents` function. I don't think this is possible to do "normally" since the `components:dirs` hook receives the `allDirs` array, which gets later transformed to the hard-coded `enabled: true`.

Looking forward to receiving feedback on this.

PR #32846, opened 2025-08-04
perf(kit): get absolute path from `tinyglobby` in `resolveFiles`

cernymatej:
### 📚 Description

This should bring a small performance improvement to the `resolveFiles` funcion by getting the absolute path directly from `tinyglobby` instead of performing additional resolution.

From my simple local testing on MacOS (M2 Max) with 500 files in a single directory, it has reduced the execution time from ~4.35ms to ~3.2ms.


danielroe (approved):

TheAlexLichter (approved):
Nice finding! 👏🏻

PR #32853, opened 2025-08-04
fix(kit,nuxt): use `reverseResolveAlias` for better errors

danielroe:
### 📚 Description

this PR updates a number of places where we hard-code `~/${somePath}` for more consistency and fixes bugs resolving paths that have a more precise path (such as in layers)

### DIFF
--- a/packages/nuxt/src/utils.ts
+++ b/packages/nuxt/src/utils.ts
@@ -1,5 +1,6 @@
 import { promises as fsp } from 'node:fs'
-import { useLogger } from '@nuxt/kit'
+import { tryUseNuxt, useLogger } from '@nuxt/kit'
+import { reverseResolveAlias } from 'pathe/utils'
 
 /** @since 3.9.0 */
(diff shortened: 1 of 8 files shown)


packages/nuxt/src/utils.ts
```diff
@@ -11,3 +12,12 @@ export async function isDirectory (path: string) {
...
 export const logger = useLogger('nuxt')
+
+export function resolveToAlias (path: string, nuxt = tryUseNuxt()) {
+  return reverseResolveAlias(path, { ...nuxt?.options.alias || {}, ...strippedAtAliases }).pop() || path
+}
+
+const strippedAtAliases = {
+  '@': '',
+  '@@': '',
+}
```
cernymatej, line 23: wouldn't it be better to put this into `packages/kit/src/resolve.ts` ?

danielroe: you're thinking of creating + exposing a new utility? good idea - we can refactor this  in the next minor to add a new export ...

PR #32855, opened 2025-08-04
docs: improve explanation of global middleware

RDistinct:
…eware is executed automatically on initial page load

This is an improvement to the statement: "Nuxt automatically executes global middleware for first time enter to the application" which is not very clear or grammatically correct.

### DIFF
--- a/docs/2.guide/1.concepts/10.nuxt-lifecycle.md
+++ b/docs/2.guide/1.concepts/10.nuxt-lifecycle.md
@@ -68,7 +68,7 @@ In Nuxt, there are three types of middleware:
 - **Named route middleware**
 - **Anonymous (or inline) route middleware**
 
-Nuxt automatically executes global middleware for first-time entry(initial page load) to the application and every time before route navigation. Named and anonymous middleware are executed only on the routes specified in the middleware property of the page(route) meta defined in the corresponding page components.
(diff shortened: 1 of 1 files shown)


docs/2.guide/1.concepts/10.nuxt-lifecycle.md
```diff
@@ -68,7 +68,7 @@ In Nuxt, there are three types of middleware:
 - **Named route middleware**
 - **Anonymous (or inline) route middleware**
 
-Nuxt automatically executes global middleware for first time enter to the application and every time before route navigation. Named and anonymous middleware are executed only on the routes specified in the middleware property of the page(route) meta defined in the corresponding page components.
+Nuxt automatically executes global middleware for first-time entry(initial page load) to the application and every time before route navigation. Named and anonymous middleware are executed only on the routes specified in the middleware property of the page(route) meta defined in the corresponding page components.
```
danielroe:
```suggestion
Nuxt executes all global middleware on the initial page load (both on server and client) and then again before any client-side navigation. Named and anonymous middleware are executed only on the routes specified in the middleware property of the page(route) meta defined in the corresponding page components.
```

danielroe (approved):
thank you! ❤️

PR #32880, opened 2025-08-06
perf(nuxt): stop watching app manifest once a change has been detected

### LINKED ISSUE #32839: Stop watching app manifest once a change has been detected
### Describe the feature

I just watched [this video](https://www.youtube.com/watch?v=D9gWQ1x6Qg8) by @TheAlexLichter which covers the experimental [checkOutdatedBuildInterval](https://nuxt.com/docs/4.x/api/nuxt-config#checkoutdatedbuildinterval) setting. It's my opinion that once an updated manifest has been detected and the hook has been called, there is no need to keep polling for a new build. If the Nuxt team agree, I'd be happy to open a PR to cancel the timeout after the [hook as been called](https://github.com/nuxt/nuxt/blob/3a66515903bfb6761f3535b5533f4b805adf4aa9/packages/nuxt/src/app/plugins/check-outdated-build.client.ts#L23).  

### Additional information

### Final checks

luc122c:
### 🔗 Linked issue

Fixes #32839

### 📚 Description

Once a change to the app manifest has been detected, the hook is called. It is then no longer necessary to keep polling the server for an updated app manifest. 

I have also updated the type of the timeout. As this is a client only plugin, it will be using [window.setTimeout](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout) not [NodeJS.setTimeout](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args).

### DIFF
--- a/packages/nuxt/src/app/plugins/check-outdated-build.client.ts
+++ b/packages/nuxt/src/app/plugins/check-outdated-build.client.ts
@@ -21,7 +21,7 @@ export default defineNuxtPlugin((nuxtApp) => {
       if (meta.id !== currentManifest.id) {
         // There is a newer build which we will let the user handle
         nuxtApp.hooks.callHook('app:manifest:update', meta)
-        if(timeout) { clearTimeout(timeout) }
+        if (timeout) { clearTimeout(timeout) }
       }
     } catch {
       // fail gracefully on network issue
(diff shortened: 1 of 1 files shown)


luc122c:
I don't think the test failures are related to this PR.

danielroe (approved):
nice catch! 🙌

PR #32882, opened 2025-08-06
feat(nuxt): support `routeRules` through `definePageMeta`

### LINKED ISSUE #32810: Support setting `routeRules` through `definePageMeta`
### Describe the feature

While we do support setting route rules from a page component with `experimental.inlineRouteRules` + `defineRouteRules`, it would be nice to also/instead be able to do this on the meta object of a route.

An example of what this looks like
```html
<script setup lang="ts">
definePageMeta({
  routeRules: {
    prerender: true
    // other rules can be added here
  }
})

// would be equivalent to
defineRouteRules({
  prerender: true,
  // other rules can be added here
})
</script>
```

This would also imply that it's possible to set it on pages through hooks:
```ts
nuxt.hook('pages:extend', pages => {
  pages.push({
    path: '/my-page',
    meta: {
      routeRules: { prerender: true }
    }
  })
})
```

The motivation behind this feature request comes from i18n, localizing routes involves duplicating routes and changing their paths (not just adding language code prefixes).

If `routeRules` are part of route meta, and extracted before `pages:resolved`, then these rules would be preserved when making copies of the route objects.

### Additional information

### Final checks

BobbieGoede:
### 🔗 Linked issue
* #32810

### 📚 Description
Resolves #32810

Adds support for setting `routeRules` in `definePageMeta`:
```html
<script setup lang="ts">
definePageMeta({
  routeRules: {
    prerender: true
  }
})
</script>
```

This also adds support for setting the `routeRules` on pages `meta.routeRules` from `pages:extend` or `pages:resolved` hooks directly.

The `meta.routeRules` property is removed after extraction to prevent exposing this in the final build.

---

I'm not entirely sure where the logic fits, suggestions on reorganization welcome 🙏

BTW, wrapping `resolvePageRoutes` rather than using `pages:extend` or `pages:resolved` hooks is intentional since the `routeRules` are processed and removed from the routes after those hooks.

I was studying how `inlineRouteRules` works and noticed some edge cases/bugs:
* Does not support multiple route paths
  * `pageToGlobMap` maps files to 1 path, this is an easy fix
* Route rules linger on route removal/rename 
  * `pageToGlobMap` is updated in `pages:extend` before `builder:watch` removes the routeRules from the `inlineRules` object, not sure how to fix that cleanly

### DIFF
--- a/docs/2.guide/2.directory-structure/1.app/1.pages.md
+++ b/docs/2.guide/2.directory-structure/1.app/1.pages.md
@@ -337,7 +337,15 @@ You may define a path matcher, if you have a more complex pattern than can be ex
 
 #### `props`
 
-Allows accessing the route `params` as props passed to the page component. See[the `vue-router` docs](https://router.vuejs.org/guide/essentials/passing-props) for more information.
+Allows accessing the route `params` as props passed to the page component. See [the `vue-router` docs](https://router.vuejs.org/guide/essentials/passing-props) for more information.
+
+#### `routeRules`
+
+You can define Nitro [route rules](/docs/guide/concepts/rendering#route-rules) for this page's route, this allows you to set prerendering, caching, and other rules.
+
+These are applied to routes matching a glob pattern based on the page's path, for example:
+  - A rule defined in `~/pages/foo/bar.vue` will be applied to `/foo/bar` requests.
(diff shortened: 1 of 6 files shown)


BobbieGoede:
Closing in favor of #32897

PR #32891, opened 2025-08-07
fix(nuxt): forward response type to ofetch options

### LINKED ISSUE #32890: `useFetch` response hooks lose generic type information, `response._data` typed as `any`
### Description
When using `useFetch` with a generic type parameter, the `response._data` property in callback hooks like `onResponse` loses its type information and falls back to `any`, instead of using the specified generic type.

### Reproduction

```typescript
type ApiResponse = {
  message: string;
  data: { id: number; name: string }[];
};

useFetch<ApiResponse>("/api/v1/users", {
  onResponse({ response }) {
    if (!response.ok) return;

    // ❌ response._data has type `any` instead of `ApiResponse`
    doSomethingWithMyData(response._data);
    //                    ^^^^^^^^^^^^^
    //                    Type is `any`, no IntelliSense or type safety
  }
});
```

### Expected Behavior
`response._data` should be typed as `ApiResponse` (the generic type passed to `useFetch`), providing proper type safety and IntelliSense.

### Actual Behavior
`response._data` has type `any`, losing all type information and safety.

### Root Cause

The issue stems from how Nuxt handles type propagation to the underlying ofetch hooks:

1. **ofetch typing**: The `_data` property type is defined in ofetch's `FetchResponse` interface ([source](https://github.com/unjs/ofetch/blob/2d3af5eb7e5ec38cf4b6f5d9b38f56fc3b9ba244/src/types.ts#L136-L138))

2. **Type propagation gap**: Nuxt extends ofetch's `FetchOptions` type in [`NitroFetchOptions`](https://github.com/nuxt/nuxt/blob/8ea3fc34dedb423f8685afc199ce983bf02d71eb/packages/nuxt/src/app/composables/fetch.ts#L25-L27), but doesn't forward the response type generics to the underlying ofetch type system

3. **Fallback to `any`**: Without proper type information, ofetch's response hooks default to `any` for the `_data` property

### Environment
- Nuxt version: 4.0.2
- TypeScript version: 5.8.3

### Solution
I have already opened a pull request solving this in #32891  as the fix is super small, but I thought I'd open an issue too just to follow protocol and get feedback.

> *Note: I had Claude rewrite my issue for clarity*

### Additional information

### Final checks

TheColorman:
### 🔗 Linked issue

resolves #32890 

### 📚 Description

Fixes typing of the `useFetch` response hooks so that `response._data` does not have the `any` type, but instead inherits the generic type parameter given to `useFetch`.

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -1,4 +1,4 @@
-import type { FetchError, FetchOptions } from 'ofetch'
+import type { FetchError, FetchOptions, ResponseType as _ResponseType } from 'ofetch'
 import type { $Fetch, H3Event$Fetch, NitroFetchRequest, TypedInternalResponse, AvailableRouterMethod as _AvailableRouterMethod } from 'nitropack/types'
 import type { MaybeRef, MaybeRefOrGetter, Ref } from 'vue'
 import { computed, reactive, toValue, watch } from 'vue'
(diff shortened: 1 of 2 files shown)


danielroe (approved):
thank you! 💪

PR #32899, opened 2025-08-08
fix(nuxt): compare final matched routes when syncing `route` object

cernymatej:
### 🔗 Linked issue
fix https://github.com/nuxt/nuxt/issues/32898

### 📚 Description
This should fix an issue with syncing the `route` object during navigation between 2 child routes.


danielroe:
do you think you could add a unit/runtime test? 🙏

PR #32901, opened 2025-08-08
refactor(nuxt): simplify page segment parsing

BobbieGoede:
### 📚 Description
This is mostly to remove the long ternaries, they're a bit difficult to read when syncing the segment parsing in i18n. 

This was also the only place in the codebase using enums, I'm not sure if we would like to work towards using erasable typescript syntax only, I replaced these in case we do.


danielroe (approved):
this is great! 

ultimately we'll use https://github.com/unjs/unrouting and I think some of these changes have already been made there. I'll try to refactor nuxt to depend on it shortly - and do check unrouting to see if there are any improvements to be made there.

PR #32902, opened 2025-08-08
fix(nuxt): handle TSSatisfiesExpression in page augmentation

### LINKED ISSUE #32889: Cannot use `satisfies` in definePageMeta name
### Environment

- Operating System: Linux
- Node Version:     v20.19.1
- Nuxt Version:     4.0.3
- CLI Version:      3.27.0
- Nitro Version:    2.12.4
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -

### Reproduction

https://stackblitz.com/edit/github-fbdwqjp3-eywjzzca?file=pages%2F%5Bslug%5D%2Fdashboard.vue

### Describe the bug

Using `satisfies` for page name throws 500 error for route match not found.

This breaks naming:
```
<script setup lang="ts">
type PageName = 'name-from-page-meta' | 'whatever';

definePageMeta({
  name: 'name-from-page-meta' satisfies PageName,
});
</script>
```

### Additional context

_No response_

### Logs

```shell-script

```

huang-julien:
### 🔗 Linked issue

fix https://github.com/nuxt/nuxt/issues/32889

### 📚 Description

this PR allows to handle TSSatisfiesExpression when extracting page meta data

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -675,6 +675,10 @@ export function isSerializable (code: string, node: Node): { value?: any, serial
     }
   }
 
+  if (node.type === 'TSSatisfiesExpression') {
+    return isSerializable(code, node.expression)
+  }
+
   return {
     serializable: false,
   }
--- a/packages/nuxt/test/page-metadata.test.ts
+++ b/packages/nuxt/test/page-metadata.test.ts
@@ -135,6 +135,23 @@ definePageMeta({ name: 'bar' })
       }
     `)
   })
+  it('should extract metadata with TS satisfies', () => {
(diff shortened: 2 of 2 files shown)


danielroe (approved):

BobbieGoede:
This handles `satisfies` for top-level meta properties but extraction will still silently fail when using other TS syntax like `as` or when using TS in nested properties.

In i18n we transform/transpile the argument object for `defineI18nRoute` before eval/extraction to solve this, not sure about the performance implications but maybe this is something we could do here as well?  🤔

huang-julien:
nice catch @BobbieGoede  . I'll open another PR

PR #32913, opened 2025-08-09
fix(nuxt): handle uncached current build manifests

jackylamhk:
### 🔗 Linked issue

Closes #32912

### 📚 Description

Check if the `getAppManifest()` call returned a 404 or 403 (usually returned when the CDN does not have access to list files). If so, proceed anyway so that the `app:manifest:update` hook can still be called.

This issue affects both Nuxt 3 and 4.

### DIFF
--- a/packages/nuxt/src/app/plugins/check-outdated-build.client.ts
+++ b/packages/nuxt/src/app/plugins/check-outdated-build.client.ts
@@ -1,4 +1,4 @@
-import { FetchError } from 'ofetch'
+import type { FetchError } from 'ofetch'
 import { defineNuxtPlugin } from '../nuxt'
 import { getAppManifest } from '../composables/manifest'
 import type { NuxtAppManifestMeta } from '../composables/manifest'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/plugins/check-outdated-build.client.ts
```diff
@@ -13,12 +14,20 @@ export default defineNuxtPlugin((nuxtApp) => {
   let timeout: NodeJS.Timeout
 
   async function getLatestManifest () {
-    const currentManifest = await getAppManifest()
+    let currentManifest: NuxtAppManifestMeta | undefined
+    try {
+      currentManifest = await getAppManifest()
+    } catch (e) {
+      // The build is already outdated but the manifest is not cached
+      if (!(e instanceof FetchError && (e.status === 404 || e.status === 403))) {
```
danielroe: perhaps we can avoid verifying it's an instance of FetchError and just check the status?

packages/nuxt/src/app/plugins/check-outdated-build.client.ts
```diff
@@ -1,3 +1,4 @@
+import { FetchError } from 'ofetch'
```
luc122c:
```suggestion
import type { FetchError } from 'ofetch'
```

danielroe: I think the instanceof test wouldn't work

luc122c: Ahh, you're probably right. My mistake!

packages/nuxt/src/app/plugins/check-outdated-build.client.ts
jackylamhk: Sure, just thought it'd be more robust and we'd get the type annotation for free.

danielroe (approved):

PR #32916, opened 2025-08-10
fix(nuxt): add warning for nuxt 3 dir structure (#32638)

xc6pm:
### 🔗 Linked issue

#32638

### 📚 Description

Added a warning for nuxt v3 dir structure, as per @florian-strasser suggestion - as the other solution (going for a strict mode), may break things for existing projects.
The warning shows up every time the user runs `nuxt dev` as I believe that to be the most visible place.


florian-strasser:
I was wondering if this warning will only be executed in dev mode. Otherwise I would recommend adding such if condition:

```javascript
if(import.meta.dev)
```

danielroe:
I am a bit torn on this one.

We deliberately support backwards-compatibility on the directory structure, and I fear that adding logging about it will be poor DX if people are choosing not to change the structure.

xc6pm:
I see. I imagined that future versions will fully move to the new structure, so a log at this point would be a logical step.

florian-strasser:
I also think that at some point this warning should just be removed and the new structure should be forced to use.
Personally I don't use the old structure at all, it's just easier for the moment to migrate like this.

danielroe:
Maybe we can make the message informational instead.

packages/nuxt/src/core/builder.ts
```diff
@@ -19,6 +19,11 @@ export async function build (nuxt: Nuxt) {
   await generateApp()
 
   if (nuxt.options.dev) {
+    const noAppFolder = nuxt.options.srcDir === nuxt.options.rootDir
+    if (noAppFolder && !nuxt.options.test) {
+      logger.warn('You are using the Nuxt v3 directory structure. We recommend moving your application code inside an /app directory in Nuxt v4. https://nuxt.com/blog/v4#%EF%B8%8F-new-project-structure')
```
danielroe: What about something like:

```suggestion
      logger.info('Using legacy directory structure. https://nuxt.com/blog/v4#%EF%B8%8F-new-project-structure')
```

... and we can use a short url instead of the blog link.

wdyt @atinux?

xc6pm: The url seems ok to me and is consistent with the rest of the project. I would've used a url shortener if the docs came with one.

packages/nuxt/src/core/builder.ts
```diff
@@ -19,6 +19,11 @@ export async function build (nuxt: Nuxt) {
   await generateApp()
 
   if (nuxt.options.dev) {
+    const noAppFolder = nuxt.options.srcDir === nuxt.options.rootDir
+    if (noAppFolder && !nuxt.options.test) {
+      logger.info('Using legacy directory structure. https://nuxt.com/blog/v4#%EF%B8%8F-new-project-structure')
```
atinux:
```suggestion
      logger.info('Using legacy directory structure. https://go.nuxt.com/v4-structure')
```

packages/nuxt/src/core/builder.ts
atinux: Agree, see my last code suggestion

PR #32920, opened 2025-08-10
fix(nuxt): transform ts before page meta extraction

BobbieGoede:
### 📚 Description
@huang-julien this is what I meant in https://github.com/nuxt/nuxt/pull/32902#issuecomment-3170662491

By transforming the script code first, we won't have to add logic to handle typescript expressions. This will also allow nested properties to use typescript syntax since we did not cover that yet.


huang-julien (commented):
I also thought of doing this but wasn't sure about if that would be worth it in terms of performances. I'm pretty sure most projects will use TS in every single pages anyways. 

I don't think there is many TS specific expressions we need to handle.

Any thoughts @danielroe ?

BobbieGoede:
@huang-julien 
I have no clue on the performance impact and not sure how to measure it, but I can try to reduce the amount of transformed code. 

The alternative would be to traverse object properties so we handle ts expressions in nested properties too, we won't be able to serialize by passing to `JSON.stringify` directly, it'll complicate things a bit.

packages/nuxt/src/pages/utils.ts
```diff
@@ -255,17 +257,33 @@ export function getRouteMeta (contents: string, absolutePath: string, extraExtra
...
+      let code = script.code
+      let pageMetaArgument = node.expression.arguments[0]
       if (pageMetaArgument?.type !== 'ObjectExpression') {
         logger.warn(`\`definePageMeta\` must be called with an object literal (reading \`${absolutePath}\`).`)
         return
       }
 
+      // when using ts we slice, transform and parse the `definePageMeta` node to avoid parsing the whole file
+      if (/tsx?/.test(script.loader)) {
+        const transformed = oxcTransform(absolutePath, script.code.slice(node.start, node.end), { lang: script.loader })
```
BobbieGoede, line 270: @huang-julien now only the `definePageMeta` node is transformed, but will need to be parsed again because of that 🤔

I guess this would be faster on pages with lots of code (less to transform) but slower for smaller pages (parsing twice)

huang-julien:
wait... we may actually want to transform the whole code 🤔 this would also allow the usage of `const enum`

BobbieGoede:
>wait... we may actually want to transform the whole code 🤔 this would also allow the usage of const enum

@huang-julien could you share a snippet of what that usage looks like?

danielroe (approved):

